PR #2725, opened 2018-11-22
Adding spatial features to content (Lucene indexed)

petedavis:
Adding spatial features to orchard core. This will allow content items to have a location and enable spatial searches for content.

Read PR documentation for details.

Fixes #2658, Fixes #4568, Fixes #9009, Fixes #3550

# TODO

- [x] Make a function that will retrieve (top, left, bottom, right) coordinates from a specific lat/long and distance(radius). This will allow to automatically find the coordinates of the `geo_bounding_box` for a `geo_distance` query.
- [x] Make the leaflet editor behave better (optional)
- [x] Move Query filters documentation to Lucene module documentation.
- [ ] Make latitude and longitude precision length configurable per Lucene Index.
- [ ] Make latitude and longitude precision length configurable per GeoPointField. Also, don't allow to have a higher length than the highest maxlength of all it's Lucene indexes.
- [x] Decide if we use a Decimal or Double to represent Latitude and Longitude.
- [ ] Lucene Spatial Queries Unit Tests
- [ ] (Idea) Use the CalcBoxByDistFromPtDEG in the GeoDistance Query to evaluate the Geo Bounding Box and make it implicit when we use a GeoDistance Query. So basically always use a GeoBoundingBox Query when using a GeoDistance Query. This way we mitigate the records on which we will evaluate a distance (perf adjustment). You can already do this by using a Nested Boolean Query for now like shown in the documentation. See http://janmatuschek.de/LatitudeLongitudeBoundingCoordinates

See Spatial4n.Core for first task : https://github.com/synhershko/Spatial4n/blob/44d01c4ce96e5be9000e9d38bbcb3d7745d7bc90/Spatial4n.Core/Distance/DistanceUtils.cs#L283

PR #5579, opened 2020-02-18
Lucene spatial

### LINKED ISSUE #2658: Lucene and API (REST & GraphQL) Geo distance search
Hi,

Is it possible at the moment to do a geolocation search with Lucene/Elastic Search (https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-geo-distance-query.html)  against some content items that would have a long/lat properties and access either via the REST or GraphQL endpoints.

Thanks

Steve

### LINKED ISSUE #4568: "filter" is not implemented in the Lucene module
In Lucene you can do a "filter", which is like "must", but does not affect the score and can also be cached which isn't the case for a "must".

https://lucene.apache.org/solr/guide/8_1/json-query-dsl.html#filter-queries

https://github.com/apache/lucenenet/blob/master/src/Lucene.Net.Queries/BooleanFilter.cs

Skrypt:
See #2725

Fixes #2658, Fixes #4568

image

image

image


Skrypt:
@petedavis please merge this branch on top of your PR to get the credits of the PR.

petedavis:
Thanks @Skrypt. Sorry I have not been looking at OC much lately. I'll get this done soon.

petedavis:
All done @Skrypt

sebastienros:
Should we get a demo of that branch?

sebastienros:
Also are we missing anything from yessql to be able to index spatial fields?

petedavis:
@sebastienros It would be good to be able to create spatial index in the data store and then use spatial queries to find the documents, but I'm not too sure if all databases will have support (eg Sqlite).

sebastienros:
We can implement it as extension methods in the query methods (like `IsIn`), and use custom dialect implementations based on the database. I am just wondering if there is something that should/could be done.

petedavis:
I think it should be done. If spatial content items were able to be searched from the database, I think you would be able to almost build a web page like: https://www.realestate.com.au/buy/in-adelaide,+sa+5000/map-1?includeSurrounding=false

And use more specific query where Lucene might not be the best solution.

sebastienros:
At least right now we can do it from lucene? Do we have the specific queries handled in the json elasticsearch language (in this PR)

PR #5580, opened 2020-02-18
Azure container name must be lower case

### LINKED ISSUE #5578: Container creation fails in Azure Media Storage if name contains uppercase chars
When creating folders or uploading images with Azure Media Storage an error is thrown: "the specified resource name contains invalid characters". 

My settings are as follows:
```
"OrchardCore.Media.Azure": 
{ "ContainerName": "{{ ShellSettings.Name }}", 
"BasePath": "media", 
"CreateContainer": true },
```

I have found out the issue is with the storage container creation. When creating a tenant I will name the tenant like MyCompany but containers names have to be lower case in Azure.

Azure Data Protection also fails with the same issue if your template for container has uppercase.

scleaver:
Fixes #5578

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.DataProtection.Azure/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.DataProtection.Azure/Startup.cs
@@ -58,7 +58,8 @@ namespace OrchardCore.DataProtection.Azure
 
                 var template = FluidTemplate.Parse(containerName);
 
-                containerName = template.Render(templateContext, NullEncoder.Default);
+                // container name must be lowercase
+                containerName = template.Render(templateContext, NullEncoder.Default).ToLower();
                 containerName.Replace("\r", String.Empty).Replace("\n", String.Empty);
             }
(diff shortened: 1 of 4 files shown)


scleaver:
Ok cool... yes, I agree that we wouldn't want to implement the whole container naming rules but I thought to lower would help someone else save the 30 mins I spent today trying to work out why it wouldn't work. I usually use camel case for tenant names myself you see.

In my modifications to docs, I have put the comment "A container name must be a valid DNS name and conform to Azure container naming rules eg. lowercase only" as this applies not only to templating but also any string you might place here. Do you think this is enough or do you want me to change the wording a little more?

deanmarcussen:
Would you mind just adding to the section
`Templating Configuration`

something like
!!! note
    When templating from the `ShellSettings.Name` property the tenants name will be automatically lowercased, however you must also make sure it conforms to other Azure Blob naming conventions.

Then everyone is saved half an hour, and we are covered

scleaver:
Done and yet to go!

PR #5596, opened 2020-02-19
Make ContentPicker default to displaying all content types when none are selected

Craige:
There is no reasonable use-case for a content picker to be configured with no DisplayedContentTypes other than it just hasn't been configured yet. On the other hand, it seems reasonable that some circumstances may necessitate picking a content item of an undeterminable ContentType at setup.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Settings/ContentPickerFieldSettingsDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Settings/ContentPickerFieldSettingsDriver.cs
@@ -22,7 +22,7 @@ namespace OrchardCore.ContentFields.Settings
 
             if (model.DisplayAllContentTypes)
             {
-                model.DisplayedContentTypes = new string[0];
+                model.DisplayedContentTypes = Array.Empty<String>();
             }
 
             context.Builder.WithSettings(model);
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Settings/ContentPickerFieldSettingsDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Settings/ContentPickerFieldSettingsDriver.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Threading.Tasks;
 using OrchardCore.ContentFields.Fields;
 using OrchardCore.ContentManagement.Metadata.Models;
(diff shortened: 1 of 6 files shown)


Craige:
@sebastienros had mentioned restricting the lookup to Documents without a stereotype, but I didn't see an easy way to do that without adding the stereotype to the ContentItemIndex, and I'm not sure we want to pollute that without a good reason.

Open to suggestions though.

sebastienros:
```
            var noStereotype = _contentDefinitionManager.ListTypeDefinitions()
                .Where(ctd => String.IsNullOrEmpty(ctd.GetSettings<ContentTypeSettings>().Stereotype));
```

Also it would be better if that was an explicit flag on the field settings: `Only content items` such that it won't conflict later on if we decide to add more things that could conflict with not selecting any content type.

And then the editor can just hide the list of content types when this option is checked.

Sorry this is asking for more work.

Craige:
> Sorry this is asking for more work.

Don't worry about it. I reckon your right anyhow. Explicit is probably better than magic. I just didn't see a valid use case for an empty DisplayedContentTypes.

I'll add an explicit flag to trigger this behavior.

Craige:
I'm really more of a backend guy, so you may want to look over what I did with the frontend toggle on the ContentPickerFieldSettings.Edit.cshtml. This is the first time I've even really touched Vue if I'm being perfectly honest, but it seems to work.

I made the `UpdateAsync` action remove any `DisplayedContentTypes` when the toggle is activated, because at this point in time I don't think it makes sense from a data-integrity point of view to allow both of those to be set at once.

Craige:
Any chance of having this looked at in the near future?

src/OrchardCore.Modules/OrchardCore.ContentFields/Settings/ContentPickerFieldSettingsDriver.cs
```diff
@@ -20,6 +20,11 @@ public override async Task<IDisplayResult> UpdateAsync(ContentPartFieldDefinitio
 
             await context.Updater.TryUpdateModelAsync(model, Prefix);
 
+            if (model.DisplayAllContentTypes)
+            {
+                model.DisplayedContentTypes = new string[0];
```
sebastienros:
```suggestion
                model.DisplayedContentTypes = Array.Empty<String>();
```

Just for your knowledge

sebastienros:
I only noticed the update after your comment. As a general rule don't hesitate to write a comment when you have added something that requires another look.

Issue #5532, opened 2020-02-12
Reduce allocations on IAsyncResultFilters


deanmarcussen:
While working on the display options for parts and fields I noticed how many requests (more so to the admin, than a content item display request) are creating a lot of unnecessary allocations, particularly around the use of `IAsyncResultFilters` 

An admin request to edit a page will often make 3-4 requests, the first to retrieve the content, then subsequent to retrieve media from the MediaController. These later request are not `ViewResult` or `PageResult` so the `Layers` filter and the `AdminMenuFilter` passes on these almost immediately.

In the case of the `AdminMenuFilter` it injects the `INavigationManager` which in turn injects all the `INavigationProviders` etc.

However these filters are instance per request, so the constructor injected dependencies are instantiated for every request, even though for a lot of requests, these filters will no-op very quickly, and never use the dependencies

It might an over optimisation, but suggest we consider refactoring to resolve necessary dependencies from the service provider, after the filter has chosen to execute, rather than no-op

jtkech:
Yes, looks like a good idea, so we would have to resolve them dynamically through a scoped provider, and maybe cache them in local variables of the related scoped service, if it may be called multiple times.

We already do this in some services where we cache the scoped resolutions, as i remember particularly when we are using `ShellScope.Services` for the scoped service provider.

PR #5600, opened 2020-02-20
Resolve asyncfilter scoped services lazily

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/5532

Probably the biggest win here is actually on the `AdminMenuFilter` which was resolving the `INavigationManager` for no reason at all on all requests, not just admin.

@jtkech just resolved through the filter context services, assume that is best?

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Admin/AdminMenuFilter.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Admin/AdminMenuFilter.cs
@@ -16,8 +16,8 @@ namespace OrchardCore.Admin
     /// </summary>
     public class AdminMenuFilter : IAsyncResultFilter
     {
-        private IShapeFactory shapeFactory;
-        private ILayoutAccessor layoutAccessor;
+        private IShapeFactory _shapeFactory;
+        private ILayoutAccessor _layoutAccessor;
 
         public async Task OnResultExecutionAsync(ResultExecutingContext filterContext, ResultExecutionDelegate next)
         {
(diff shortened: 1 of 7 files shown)


sebastienros:
Design suggestion, create a new interface like `IAdminAsyncResultFilter` that our filters would implement instead (inherits from IAsynResultFilter) and are registered as this interface. Then an implementation of `IAsyncReultfilter` that will check if we are in the admin (or not) and/or in a View, before resolving the services implementing IAdminAsyncResultFilter, or IAdminViewAsyncResultFilter, or IFrontEndAsyncResultFilter ...

jtkech:
Good idea for the `IAsyncReultfilter` wrapper but if all `IAdminAsyncResultFilter` inherits from `IAsynResultFilter`, i think they will be still resolved by MVC if resolved as `IEnumerable<>`.

Then, just to say that when you use multiple time `HttpContext.RequestServices` because somewhere we access our `AsyncLocal<ShellScope>`, better to use an intermediate services variable.

    var services = context.HttpContext.RequestServices;
    var service1 = services.GetRequiredService<IService1>();
    var service2 = services.GetRequiredService<IService2>();

Same remark if we were using `ShellScope.Services`.

deanmarcussen:
Decided not worth the risk of interfering with the `IAsyncResultFilter` `await next.Invoke()` pattern if placed in a wrapper.

Closing, and will do another small pr, just to improve the `AdminMenuFilter` which was the biggest win

PR #5601, opened 2020-02-20
Correct logging

deanmarcussen:


src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/UserInfoController.cs
```diff
@@ -120,8 +120,7 @@ public async Task<IActionResult> Me()
                     }
                     else
                     {
-                        _logger.LogWarning($"{OpenIdConnectConstants.Claims.UpdatedAt} claim value '{updatedAtClaimValue}' is invalid. Should be the time the end-user's" +
-                                           " information was last updated, as number of seconds since the Unix epoch (1970-01-01T0:0:0Z) as measured in UTC until the date/time. ");
+                        _logger.LogWarning("{UpdatedAt} claim value '{UpdatedAtClaimValue}' is invalid. Should be the time the end-user's information was last updated, as number of seconds since the Unix epoch (1970-01-01T0:0:0Z) as measured in UTC until the date/time. ", OpenIdConectConstants.Claims.UpdatedAt, updatedAtClaimvalue);
```
agriffard: > _logger.LogWarning("{0} claim value '{1}' is invalid. Should be the time the end-user's information was last updated, as number of seconds since the Unix epoch (1970-01-01T0:0:0Z) as measured in UTC until the date/time.", OpenIdConnectConstants.Claims.UpdatedAt, updatedAtClaimValue);

I was about to do the change but I don't get it why is has to be changed.
I understand it for strings that have to be translated but why for this kind of log?

kevinchalet: Personally, I'd prefer if we had an exception instead of a log message. Or we could just use `long.Parse` like we do for booleans (BTW, we should use an overload accepting a `CultureInfo`).

src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/UserInfoController.cs
```diff
@@ -114,15 +115,7 @@ public async Task<IActionResult> Me()
...
-                    if (long.TryParse(updatedAtClaimValue, out var epoch))
-                    {
-                        claims[OpenIdConnectConstants.Claims.UpdatedAt] = epoch;
-                    }
-                    else
-                    {
-                        _logger.LogWarning($"{OpenIdConnectConstants.Claims.UpdatedAt} claim value '{updatedAtClaimValue}' is invalid. Should be the time the end-user's" +
-                                           " information was last updated, as number of seconds since the Unix epoch (1970-01-01T0:0:0Z) as measured in UTC until the date/time. ");
-                    }
+                    claims[OpenIdConnectConstants.Claims.UpdatedAt] = long.Parse(updatedAtClaimValue, CultureInfo.CurrentCulture);
```
kevinchalet:
```suggestion
                    claims[OpenIdConnectConstants.Claims.UpdatedAt] = long.Parse(updatedAtClaimValue, CultureInfo.InvariantCulture);
```

src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/UserInfoController.cs
deanmarcussen: Changed to an exception rather than a log warning @kevinchalet 

@agriffard for all logs, shouldn't use the string interpolation syntax of `$"{some-format}"`

src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/UserInfoController.cs
deanmarcussen: done

sebastienros:
If we expect this value to not be a long, then I would instead return a better error message than just a generic 500 with a "parse exception".

kevinchalet:
The `updated_at` claim is defined by the OIDC spec as a JSON number (so in theory, it could be a decimal value, but in practice, it's pretty much never the case). Having a non-numeric value should **never** happen, except if the developer explicitly added an `updated_at` claim with a non-standard value. It's a corner case and I'm not sure it deserves a nice error page.

PR #5604, opened 2020-02-20
LocalizationSetContentPicker missing translated content items

### LINKED ISSUE #5603: LocalizationSetContentPicker returning items from other culture
https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.ContentFields/Controllers/LocalizationSetContentPickerAdminController.cs#L65

Here we can't do a Take(40) because we never specify which culture we want in this query. It should take all the contentItems of all cultures which means that if we do a Take(40) we will end up with missing contentitems that we filter per culture later on in the DefaultContentLocalizationManager.

Skrypt:
Fixes #5603

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Controllers/LocalizationSetContentPickerAdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Controllers/LocalizationSetContentPickerAdminController.cs
@@ -62,7 +62,7 @@ namespace OrchardCore.ContentFields.Controllers
                 dbQuery.With<ContentItemIndex>(x => x.DisplayText.Contains(query) || x.ContentType.Contains(query));
             }
 
-            var contentItems = await dbQuery.Take(40).ListAsync();
+            var contentItems = await dbQuery.ListAsync();
 
             // if 2 search results have the same set, select one based on the current culture
             var cleanedContentItems = await _contentLocalizationManager.DeduplicateContentItemsAsync(contentItems);


Skrypt:
To be honest I'm not sure why we don't use directly the `_session.QueryIndex<LocalizedContentItemIndex>` directly in the controller here. I need to read that code more. Waiting on @jptissot explanation.

src/OrchardCore.Modules/OrchardCore.ContentFields/Controllers/LocalizationSetContentPickerAdminController.cs
```diff
@@ -62,7 +62,7 @@ public async Task<IActionResult> SearchLocalizationSets(string part, string fiel
                 dbQuery.With<ContentItemIndex>(x => x.DisplayText.Contains(query) || x.ContentType.Contains(query));
             }
 
-            var contentItems = await dbQuery.Take(40).ListAsync();
+            var contentItems = await dbQuery.ListAsync();
```
sebastienros: This can not be merged as-is as this will load all content items on every lookup. This action should accept a culture (optionally?) and then prefilter with this culture, so you only get 40 items (or whatever fixed number) of this specific culture.

sebastienros: Using the localization set index is probably related to the solution.

jptissot:
If I remember correctly, the issue is that you want to include the top items that have a LocalizationSet but are not necessarily translated in the current locale.

Skrypt:
But why would I want to pick a content item that is in a different culture than the one your are creating content for?

Here, the issue is that it takes the first 40 content items which one of them could have a translated content item but only at the 41'st record for example. Then when we are filtering the content items in a later method which takes the content items that are from that current content item culture or pick the one from the default culture. So we end up with a list of content items who are from multiple different cultures and also not accurate because we could have had that 41'st record which would have been in the proper culture.

sebastienros:
> But why would we want to pick a content item that is in a different culture than the one we are creating content for?

Why couldn't we? At least it would be a setting if you really want to make that decision on a per field level.

Skrypt:
I don't mind, but yes, make it a setting then because it's not the typical use case scenario. Though, the issue would be the same with the actual code because of that limit(40).

Let's say we have 3 different cultures. This limit(40) is an even number which would return in all cases missing content items. This limit needs then to be divisible by the number of cultures that are supported.

jptissot:
If you want to pick an item in a specific culture, you should use the ContentPicker. The only reason to use the LocalizationSetContentPicker is when you are referencing a localized parent ContentItem from a non Localized ContentItem and want your template to resolve the ContentItem in the right culture when displaying it, not when selecting it.

jptissot:
In my projects, when I reference a Localized ContentItem from another Localized ContentItem, I don't use the LocalizationSetContentPicker but instead use the ContentPicker and Content for each culture picks the content in it's culture.

Skrypt:
I don't get this result on my side. When I use a content picker with localized content items it will list content items of any culture.

jptissot:
Yes, maybe it lists all items but the one you select is to a single item of the correct culture.

hishamco:
Any updates on this?

agriffard:
There are 2 approvals but many comments after.
Can this be merged?

hishamco:
I don't think so, coz there's something that Seb asked @Skrypt to do

deanmarcussen:
Yes, maybe @jptissot can explain more clearly, I withdraw my approval ;)

Skrypt:
Guys I added that dontmerge flag 5 days ago for that matter.

hishamco:
Ya, I was asking for any updates if there is ;)

agriffard:
What is the status of this PR @Skrypt?

Skrypt:
It's low priority in my list as people did not agree on the proposed solution. I need to think about a different one.

agriffard:
Closing the dontmerge labels.

Issue #4950, opened 2019-12-04
Media file store cache management


TFleury:
When the `DefaultMediaFileStoreCacheFileProvider` is enabled (with Azure Media Storage) all accessed media are stored in 'ms-cache' directory.

It's very efficient for performance, but I see some issues here :
- If a media is changed or removed, it's still available unless you manually purge cache.
- When using multiple instances, you need to manually purge cache on each of them.
- There is no disk space management, and unless you purge cache periodically the storage will grow and it will potentially end up with all medias stored in cache folder.
One interest (the most important to me) of Azure Media Storage is to have a decent application directory size even if you have lot of medias.

Shouldn't we use `IDistributedCache` for that ? Probably with a maxFileSize above which the file is not cached but served directly from `IMediaFileStore`.

Skrypt:
//cc @deanmarcussen

sebastienros:
When this feature was designed we realized these limitations, and postponed these on purpose for when we would find some time to improve it. So definitely a yes to find solution to these issues.

The IDistributedCache is not a valid solution IMO as it will just move the static files in memory (even if it's stored on redis for instance). The goal here is to use the fastest way to serve the content by using the FS. 

I think we talked that we could just define more options in the feature with disk quotas, and a background task to ensure the quotas are respected. And also create an handler interface when files are updated/removed such that this feature could also remove the file from the cache.

This should not be hard to implement.

deanmarcussen:
Yes, also agree that there are more improvements that can be done to reduce the limitations.

Here's a few thoughts on the limitations mentioned, plus a couple of others.

- Multiple instances, and purging. This should probably be handled by a change token, see what @jtkech has been working on here https://github.com/OrchardCMS/OrchardCore/pull/4001 and then when the distributed support is completed https://github.com/OrchardCMS/OrchardCore/pull/2247 it should work in the same way as the other cached items.

[...]

PR #5617, opened 2020-02-23
Remove items from media cache when events fired

deanmarcussen:
Relates to https://github.com/OrchardCMS/OrchardCore/issues/4950

Deletes files from cache when deleted / moved

Also removes unused interfaces from the unused `Media.Services` project I removed last week.

//cc @Skrypt This should resolve the items left in cache from the media attached temp directory that you showed me last week.


jtkech:
Then we will need to adpat it to a distributed environment, knowing that we prefer to not use a distributed signal using a message bus. I have different ideas that we will need to discuss. Maybe a distributed implementation through #5249 or in a separate PR.

Is it so expensive to get from the blob the Tree of media files? If not maybe just a bg Task that grabs these infos periodically and delete the cached files that don't exist anymore. Note: Or just a bg Task that just check the files that has not been accessed for a while.

Otherwise grab these infos once (by the 1st running instance) from the blob to built a shared Document (not persisted i think but shared through the multi level cache) that the handler will update.

We would have to check that a blob action doesn't fail before updating the cached Document, as i did with the file content definition store in #5249 by using an `IFileDocumentStore`.

Skrypt:
Yes, we would require a cache of fileInfos to make sure these files gets removed from the cache on each instance.

Skrypt:
I don't think this PR should cover this scenario so it is fine to merge it as it at least fix the issue on a single instance tenant website.

jtkech:
I agree, i only suggested to do it later, in my other pr or a separate one, but yes not in this one

Also we try to not use a message bus => no distributed events => rather use distributed states

deanmarcussen:
Yes @jtkech I agree this will need another implementation, of a handler, for distributed.

My thoughts were to have a second `MediaEventHandler` as a feature that depends on `Distributed` to manage that.

Options include
- Background task for period checking of accessed dates (that is already suggested on the issue anyway, to limit the cache size) and I think it is a good idea, regardless of distributed, but I intended it for another pr (don't want to make them too big!), and as an opt-in feature

- For distributed, I was thinking maybe of a volatile state document, which maintained the last 24 hours (configurable, maybe 1 week), of updated media files, knowing that if a server is out of the instance for longer than the period, there would be a few instant state related things to clear, lucene rebuild possibly, as well, so it would potentially always be something of a process, to bring a server back online, after an extended period of outage.

- They could also be persisted permanently, but overtime could grow.

- Also there is the ImageSharp cache to resolve at some point

- For sitemaps cache too.

- Also the cache clearing (from the admin) would need to raise an event / volatile state across servers.

jtkech:
Okay will see, maybe we would have to do separate PRs for Lucene, Azure Media and SiteMaps.

> For distributed, I was thinking maybe of a volatile state document, which maintained the last 24 hours (configurable, maybe 1 week), of updated media files

Document cache options are configurable per document type. That said here i was thinking about a document representing the list / tree of the existing blob media files, not the list of the updated media files which is more like a kind of event, will see ;)

PR #5618, opened 2020-02-23
Produce cleaner image resize query string

deanmarcussen:
While working on something else realised that the `resize` liquid filter is producing quite an ugly query string

`media/about/2.jpg?&width=160&height=160&rmode=stretch` with an ampersand directly after the `?`, which works, but is unnecessary, and ugly.

This just uses the Mvc `QueryHelper` to add it in a cleaner way so you get
`media/about/2.jpg?width=160&height=160&rmode=stretch`

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Media/Filters/MediaFilters.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Filters/MediaFilters.cs
@@ -1,8 +1,8 @@
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.WebUtilities;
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.Media/Filters/MediaFilters.cs
```diff
@@ -1,6 +1,8 @@
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using Fluid;
 using Fluid.Values;
+using Microsoft.AspNetCore.WebUtilities;
```
hishamco: Move it up before System.*

PR #5624, opened 2020-02-24
RetrieveContent workflow task added outcome and fixed parsing

mawendel:
Thix PR fixes a couple of things:

1. The workflow task RetrieveContent was missing a "Not found" outcome instead of simply failing the workflow.
2. There were also problems when trying to parse a JObject property from requestFormAsJson(), where the JObject would return a JArray instead of a text string thus returning null.
3. In the RetrieveContent input if you for some reason need to append a text string such as `property("FormInput").ContentItemId + "GIG"`, it would return a malformed text string that had to be cleaned.


sebastienros:
Please don't merge yet, I need some more time to understand the changes. Thanks.

mawendel:
If you have any questions let me know.

Issue #5334, opened 2020-01-21
Disabled Workflow with HttpRequestEvent still triggers


jptissot:
Disabling the workflow does not stop the HttpRequestEvent to trigger.

Skrypt:
@sfmskywalker Please give me some feedback on this. Is it related to the fact that we create a route for this HttpRequestEvent and that it remains even when we disable the start event from the Workflow?

jptissot:
I think we just need to add a if here: 

https://github.com/OrchardCMS/OrchardCore/blob/cb4f0ec17188bc0ab20c241b36609726b62b0902/src/OrchardCore.Modules/OrchardCore.Workflows/Http/Controllers/HttpWorkflowController.cs#L82-L83
that checks the IsEnabled flag of the workflow.

Skrypt:
Do a PR and ask @sfmskywalker for review. Quickly done on diner time 😉

PR #5625, opened 2020-02-24
Add check to see if http workflows are enabled before action

JoeBerkley:
I'd added a check to see if Http workflows are enabled before continuing to attempt to start any actions.

This closes OrchardCMS/OrchardCore#5334

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Workflows/Http/Controllers/HttpWorkflowController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Workflows/Http/Controllers/HttpWorkflowController.cs
@@ -91,6 +91,16 @@ namespace OrchardCore.Workflows.Http.Controllers
                 return NotFound();
             }
 
+            if(!workflowType.IsEnabled)
+            {
+                if (_logger.IsEnabled(LogLevel.Warning))
+                {
+                    _logger.LogWarning("The provided workflow type with ID '{WorkflowTypeId}' is not enabled.", payload.WorkflowId);
+                }
+
+                return NotFound();
+            }
+
             // Get the activity record using the activity ID provided by the token.
             var startActivity = workflowType.Activities.FirstOrDefault(x => x.ActivityId == payload.ActivityId);
 


src/OrchardCore.Modules/OrchardCore.Workflows/Http/Controllers/HttpWorkflowController.cs
```diff
@@ -91,6 +91,16 @@ public async Task<IActionResult> Invoke(string token)
                 return NotFound();
             }
 
+            if(!workflowType.IsEnabled)
+            {
+                if (_logger.IsEnabled(LogLevel.Warning))
+                {
+                    _logger.LogWarning("The provided workflow type with ID '{WorkflowTypeId}' is not enabled.", payload.WorkflowId);
```
hishamco: The message need to be localized

JoeBerkley: It doesn't look like any of these warnings are localized, I'll get them all updated while I'm here 👍

src/OrchardCore.Modules/OrchardCore.Workflows/Http/Controllers/HttpWorkflowController.cs
```diff
@@ -21,6 +22,7 @@ public class HttpWorkflowController : Controller
         private readonly IWorkflowStore _workflowStore;
         private readonly IActivityLibrary _activityLibrary;
         private readonly ISecurityTokenService _securityTokenService;
+        private readonly IStringLocalizer T;
```
hishamco: Please use `S` instead, we already use this convention everwhere

JoeBerkley: For my own reference, when I contribute in future, if I see a location using `T` should I convert it to `S`?

hishamco: We use `S` for `IStringLocalizer`, `H` for `IHtmlLocalizer` and `T` for localizers in views only

src/OrchardCore.Modules/OrchardCore.Workflows/Http/Controllers/HttpWorkflowController.cs
```diff
@@ -143,7 +157,7 @@ public async Task<IActionResult> Invoke(string token)
                 {
                     if (_logger.IsEnabled(LogLevel.Warning))
                     {
-                        _logger.LogWarning("No workflow found that is blocked on activity '{ActivityId}'", startActivity.ActivityId);
+                        _logger.LogWarning(S["No workflow found that is blocked on activity '{ActivityId}'"], startActivity.ActivityId);
```
hishamco: I think we should use params instead, but wait for @sebastienros to confirm

src/OrchardCore.Modules/OrchardCore.Workflows/Http/Controllers/HttpWorkflowController.cs
deanmarcussen: Good to see your internet is working again @hishamco :)

I'm confused why you have asked @JoeBerkley for logger messages to be localized? 

We don't do that anywhere else that I'm aware, and logging correctly has been a much discussed topic?

JoeBerkley: > We don't do that anywhere else that I'm aware, and logging correctly has been a much discussed topic?

Seems like you're right there @deanmarcussen - [Theres only one other place that LogWarning is used with a localized message](https://github.com/OrchardCMS/OrchardCore/blob/6cddd9f9904ff117ba82587f39fc76f49fac3cc3/src/OrchardCore.Modules/OrchardCore.Users/Controllers/AccountController.cs#L419) - and even then its done in a weird way

deanmarcussen: Yeah sorry to give you more work @JoeBerkley will see what @hishamco has to say.

You're right that other one is done in a weird way, but shares same message for both the logger, and the UI, so maybe ok.

Found another one here as well https://github.com/OrchardCMS/OrchardCore/blob/5ed3e9c3ccc4517399e16c569da4a6cd101c7de3/src/OrchardCore/OrchardCore.ReCaptcha.Core/TagHelpers/ReCaptchaTagHelper.cs#L99

sebastienros: I don't think we should start localizing exception messages, it would create a bad precedent that others could follow. If that was like this before please remove it.

src/OrchardCore.Modules/OrchardCore.Workflows/Http/Controllers/HttpWorkflowController.cs
sebastienros: The code before localization was perfect.

src/OrchardCore.Modules/OrchardCore.Workflows/Http/Controllers/HttpWorkflowController.cs
JoeBerkley: I'll reverse the Localization and push up the fix for the original issue 👍

sebastienros:
If we think we need to localize the logs, which might be a good thing, let's not use this PR. We need to talk about it first.

hishamco:
> I'm confused why you have asked @JoeBerkley for logger messages to be localized?

> If we think we need to localize the logs, which might be a good thing, let's not use this PR. We need to talk about it first.

Sorry, I though it's localized everywhere

hishamco:
@JoeBerkley you can revert the changes for the localization stuff, seem I'm outdated ;) for almost month and half during the internet issue in my country

We also need to remove the localized logs that mentioned by @JoeBerkley too

hishamco:
I just filed a new issue to discuss the [ability to support log messages localization](https://github.com/OrchardCMS/OrchardCore/issues/5634)

PR #5626, opened 2020-02-24
Localize AutoroutePart pattern

### LINKED ISSUE #5599: Autoroute Pattern and {{ | t }} filter
We cannot use {{ "something" | t }} in an autoroute pattern because it will always use the CultureInfo.CurrentUICulture. I tried forcing a Culture on the LiquidTemplateContext and it doesn't take it into consideration.

hishamco:
Fixes #5599

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
@@ -131,7 +131,7 @@ namespace OrchardCore.Autoroute.Handlers
                     ContentItem = part.ContentItem
                 };
 
-                var cultureAspect = await _contentManager.PopulateAspectAsync(contentItem, new CultureAspect());
+                var cultureAspect = await _contentManager.PopulateAspectAsync(part.ContentItem, new CultureAspect());
                 if (cultureAspect.Culture != null)
                 {
                     LiquidViewTemplate.Context.CultureInfo = CultureInfo.GetCultureInfo(cultureAspect.Culture);
--- a/src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
@@ -8,6 +8,7 @@ using Microsoft.Extensions.Options;
 using OrchardCore.Autoroute.Drivers;
 using OrchardCore.Autoroute.Models;
 using OrchardCore.Autoroute.ViewModels;
+using OrchardCore.ContentLocalization;
 using OrchardCore.ContentLocalization.Models;
 using OrchardCore.ContentManagement;
 using OrchardCore.ContentManagement.Handlers;
@@ -133,10 +134,7 @@ namespace OrchardCore.Autoroute.Handlers
                 };
 
                 var cultureAspect = await _contentManager.PopulateAspectAsync(part.ContentItem, new CultureAspect());
-                if (cultureAspect.Culture != null)
-                {
-                    LiquidViewTemplate.Context.CultureInfo = CultureInfo.GetCultureInfo(cultureAspect.Culture);
-                }
+                LiquidViewTemplate.Context.CultureInfo = cultureAspect.Culture;
 
                 part.Path = await _liquidTemplateManager.RenderAsync(pattern, NullEncoder.Default, model,
                     scope => scope.SetValue("ContentItem", model.ContentItem));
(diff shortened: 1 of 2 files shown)


src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
```diff
@@ -126,6 +129,12 @@ public override async Task UpdatedAsync(UpdateContentContext context, AutorouteP
                     ContentItem = part.ContentItem
                 };
 
+                var localizationPart = part.ContentItem.As<LocalizationPart>();
```
sebastienros: It's better to use the `GetContentCultureAsync` method which uses the `CultureAspect1`. Such that it will work without a part that is not `LocalizationPart`.

Skrypt: ok for that

hishamco: > It's better to use the `GetContentCultureAsync` method

I was thinking about this before I started on this PR, but I'm not sure if access the required services from `LiquidViewTemplate` is the right way

hishamco:
FYI I tried this

```csharp
var orchardHelper = LiquidViewTemplate.Context.Services.GetService<IOrchardHelper>();
var contentItemCulture = await orchardHelper.GetContentCultureAsync(model.ContentItem);

LiquidViewTemplate.Context.CultureInfo = contentItemCulture;
```

it doesn't work as expected

sebastienros:
I did the change, please check

src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
```diff
@@ -126,6 +131,12 @@ public override async Task UpdatedAsync(UpdateContentContext context, AutorouteP
                     ContentItem = part.ContentItem
                 };
 
+                var cultureAspect = await _contentManager.PopulateAspectAsync(contentItem, new CultureAspect());
```
hishamco: Aha you 're using `PopulateAspectAsync` instead, but I'm not sure wh y `GetContentCultureAsync` did work as expected?!!

I'm expected the `LocalizationPart` the one who responsible for populating the aspect

hishamco:
Hope to check it ASAP, also I will check why the build is fail

src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
```diff
@@ -126,6 +131,12 @@ public override async Task UpdatedAsync(UpdateContentContext context, AutorouteP
                     ContentItem = part.ContentItem
                 };
 
+                var cultureAspect = await _contentManager.PopulateAspectAsync(part.ContentItem, new CultureAspect());
+                if (cultureAspect.Culture != null)
```
hishamco: I don't think the `cultureAspect.Culture` will be `null` while it set to the `CurrentUICulture` in `CultureAspect`

Skrypt:
I pushed one more commit but the PR did not get updated with it. Strange one. Also, the changes doesn't work. The code looks fine but it doesn't work.

Skrypt:
The code looks fine but it doesn't work.

hishamco:
> The code looks fine but it doesn't work.

That's why I asked before when I used `GetContentAspectAsync()` too

hishamco:
If you tried the original commit it work perfectly

hishamco:
@Skrypt did you remove the permlink and add it again after the build?

Skrypt:
I'm using the refresh permalink checkbox option of the AutoroutePart. So I've set it to be updatable.

Skrypt:
orchard

Skrypt:
And here it works when I change the CultureInfo.CurrentUICulture : 

orchard

hishamco:
> I'm using the refresh permalink checkbox option of the AutoroutePart

I didn't know there's a checkbox, but when I remove and add the permlink again it works with my me at the first time before the other commits have been added

Skrypt:
@hishamco I tried also by clearing the permalink input of course. Does the same.
And I also changed the code to your inital first commit to test if it work and it does not either.

orchard

Skrypt:
Also tried the previous post code with removing this line in LiquidViewTemplate.
And it doesn't work either.

image

hishamco:
Strange!! Could you please clean the solution and be sure that the translations are there, I can try this too if you want

Skrypt:
Translations are there don't worry about that. You can see it's working on a previous gif.

hishamco:
I will try this again with m initial commit with `IContentManager` injection

Skrypt:
Cleaned up the solution, same result.

hishamco:
Nooo .. let me try it now

hishamco:
Oops, I'm getting an exception whenever I create or edit content item .. I'm checking why this happing, I didn't change anything since then

hishamco:
Why this is close? I will test it again once I fixed the exception I got

PR #5628, opened 2020-02-25
Use sender field when testing SMTP settings

petedavis:


src/OrchardCore.Modules/OrchardCore.Email/Controllers/AdminController.cs
```diff
@@ -92,6 +92,11 @@ private MailMessage CreateMessageFromViewModel(SmtpSettingsViewModel testSetting
                 ReplyTo = testSettings.ReplyTo
             };
 
+            if (!String.IsNullOrWhiteSpace(testSettings.Sender))
```
hishamco: If I'm not wrong this already set in the `SmtpService`

PR #5636, opened 2020-02-26
Remove a static localizer

jtkech:
Is it worth trying to provide a more specific type for the localizer?


deanmarcussen:
@hishamco hopefully you will know the answer to this question.

Will the poextractor find this translation, when it is not in the `Startup` class, and the localizer requested is `IStringLocalizer<Startup>`, i.e. of `Startup`

hishamco:
@deanmarcussen at the begging I thought the `PoExtractor` look for the localizers with names `S`, `H` & `T` but after I dig into the code what I see is the code scans the code tree and look for such names everywhere

I can try this change with `PoExtractor` to ensure that it works, which it should

I still have a doubt about the po extractor, coz if you have an indexer or array with the above names, the extractor will extract them as well which is totally wrong. I may need to look into the PoExtractor again and do some PRs to enhance this behavior

hishamco:
Wait a minute, while the name is `localizer` I don't think it be extracted at all, @jtkech we need to change the localizer name to make it discoverable

Let me check this again with PoExtractor

jtkech:
I would need more time to test it but following the readme of PoExtractor, it seems that the type is not so important (we are not using the mvc resource manager), and as @hishamco said we need to have a field named `S`.

So in my last commit i added an `IStringLocalizer` field named `S` but that is dynamically resolved from the current validation context.

To make the unit tests working, we need to resolve the localizer from the service provider of the current validation context. The only way i found is to use an `AsyncLocal<>` that is always set in the flow of a given scope so that it is not shared across tenants.

sebastienros:
Or just remove this extension method and use the code directly where it needs to be. Or externalize the localization logic somehow if this method is called multiple times.

jtkech:
Okay, i will try but no luck, i thought i could do this one in 5 minutes ;)

sebastienros:
Found the solution probably in another class that implements this interface:
`var S = validationContext.GetService<IStringLocalizer<SmtpSettings>>();`

jtkech:
Okay i'll take a look

jtkech:
Hmm seems that this is what i did in my last commit but by still using a field in place of a local variable, as it seems to be required by the po extractor

jtkech:
Okay, i used a local variable in place of a field / property, as in other places.

But maybe this is these other places that would need to use a field named S, not a local var

src/OrchardCore.Modules/OrchardCore.OpenId/ViewModels/EditOpenIdApplicationViewModel.cs
```diff
@@ -30,13 +32,29 @@ public class EditOpenIdApplicationViewModel : IValidatableObject
...
+        public IEnumerable<ValidationResult> Validate(ValidationContext validationContext) => ValidateUrls(validationContext, nameof(RedirectUris), RedirectUris)
+                .Union(ValidateUrls(validationContext, nameof(PostLogoutRedirectUris), PostLogoutRedirectUris));
 
         public class RoleEntry
         {
             public string Name { get; set; }
             public bool Selected { get; set; }
         }
+
+        private IEnumerable<ValidationResult> ValidateUrls(ValidationContext context, string memberName, string member)
```
hishamco: This will become hard to manage, once you removed the extension method

jtkech: I removed it as suggested by @sebastienros 

So now it is as in some other classes as `SmtpSettings`, resolved with the right type and by using a local `S` variable, so i removed the related field.

The only question i had is that maybe it would have been better to use a field for poextractor, if so we would have to update these other classes instead.

Issue #4529, opened 2019-10-14
Workflow javascript function requestForm() does not work


mawendel:
I have been trying to figure out how to convert form input fields to an array and then use the **foreach** task to iterate over them all, but so far i have not been able to find anything that could satisfy the requirement. Maybe a pure array would work? But problem is it is bugged.

Documentation says:
```
requestForm(): String
requestForm(name: String): String or Array
```

Looking at the code 

```
_requestFormMethod = new GlobalMethod
            {
                Name = "requestForm",
                Method = serviceProvider => (Func<string, object>)(field =>
                {
                    object result;
                    if (httpContextAccessor.HttpContext.Request.Form.TryGetValue(field, out var values))
                    {
                        if (values.Count == 0)
                        {
                            result = null;
                        }
                        else if (values.Count == 1)
                        {
                            result = values[0];
                        }
                        else
                        {
                            result = values.ToArray();
                        }
                    }
                    else
                    {
                        result = null;
                    }
                    return result;
                })
            };
```

It is clear that requestForm() was supposed to return an array, but when no parameter is given a null pointer exception is thrown instead.

mawendel:
I made a quick fix for it

```
_requestFormMethod = new GlobalMethod
            {
                Name = "requestForm",
                Method = serviceProvider => (Func<string, object>)(field =>
                {
                    object result;
                    if (field != null && httpContextAccessor.HttpContext.Request.Form.TryGetValue(field, out var values))
                    {
                        if (values.Count == 0)
                        {
                            result = null;
                        }
                        else if (values.Count == 1)
                        {
                            result = values[0];
                        }
                        else
                        {
                            result = values.ToArray();
                        }
                    }
                    else if (httpContextAccessor.HttpContext.Request.Form.Count > 0)
                    {
                        result = httpContextAccessor.HttpContext.Request.Form.ToArray();
                    }
                    else
                    {
                        result = null;
                    }
                    return result;
                })
            };

```

This returns the array as i need.

jptissot:
@mawendel want to make a PR ?

mawendel:
@jptissot  i'll try once im back from vacation.

PR #5645, opened 2020-02-27
Fix requestForm() null exception and clarify docs.

mawendel:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/4529

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Workflows/Http/Scripting/HttpMethodsProvider.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Workflows/Http/Scripting/HttpMethodsProvider.cs
@@ -118,6 +118,10 @@ namespace OrchardCore.Workflows.Http.Scripting
                             result = values.ToArray();
                         }
                     }
+                    else if (httpContextAccessor.HttpContext.Request.Form.Count > 0)
+                    {
+                        result = httpContextAccessor.HttpContext.Request.Form.ToArray();
+                    }
                     else
                     {
                         result = null;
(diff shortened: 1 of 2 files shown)


src/OrchardCore.Modules/OrchardCore.Workflows/Http/Scripting/HttpMethodsProvider.cs
```diff
@@ -118,6 +118,10 @@ public HttpMethodsProvider(IHttpContextAccessor httpContextAccessor)
                             result = values.ToArray();
                         }
                     }
+                    else if (httpContextAccessor.HttpContext.Request.Form.Count > 0)
+                    {
+                        result = httpContextAccessor.HttpContext.Request.Form.ToArray();
```
sebastienros: An array of what? internally is an IEnumerable<KeyValue>>. That's why the document states it would return a string instead, which I agree is still broken without any change.

mawendel: It returns an array of the form fields and their values. 

At the time i was trying to use the workflow task for loop to iterate through the form array, but i was not able to get the form array since it just returned null. If you would rather return a text string of all the form values then you can delete the pr.

sebastienros: Could you paste an example with what it looks like as an array, and as a string?
We still need to fix it, I am just wondering what is the most appropriate option for the scenario.

mawendel: I'm not sure what you mean with what it looks like? Its just an array of strings from the form. It works and returns in the same way, as the `_queryStringMethod` does it.

What i'm using today however, is requestFormAsJson and then new Array(json.MyArray) when i need to iterate. But im sure there are still cases where iterating form values could be useful or/and easier for some users.

PR #5649, opened 2020-02-29
Customize ViewContext property name

### LINKED ISSUE #5648: Setting ViewContext on Liquid TagHelper
In LiquidTagHelperActivator it should check for `[ViewContext]` attribute not the property name.

https://github.com/OrchardCMS/OrchardCore/blob/6cddd9f9904ff117ba82587f39fc76f49fac3cc3/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/TagHelpers/LiquidTagHelperActivator.cs#L58

As its possible that developer have given some different name. for example.

```
[ViewContext]
public ViewContext Vc { get; set; }
```

hishamco:
Fixes #5648

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/TagHelpers/LiquidTagHelperActivator.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/TagHelpers/LiquidTagHelperActivator.cs
@@ -38,6 +38,15 @@ namespace OrchardCore.DisplayManagement.Liquid.TagHelpers
 
             foreach (var property in accessibleProperties)
             {
+                var setter = MakeFastPropertySetter(type, property);
+                var viewContextAttribute = property.GetCustomAttribute<ViewContextAttribute>();
+
+                if (viewContextAttribute != null && property.PropertyType == typeof(ViewContext))
+                {
+                    _viewContextSetter = (helper, context) => setter(helper, context);
+                    continue;
+                }
+
                 var allNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase) { property.Name };
                 var htmlAttribute = property.GetCustomAttribute<HtmlAttributeNameAttribute>();
 
@@ -53,15 +62,6 @@ namespace OrchardCore.DisplayManagement.Liquid.TagHelpers
 
                 foreach (var propertyName in allNames)
                 {
-                    var setter = MakeFastPropertySetter(type, property);
-                    var viewContextAttribute = property.GetCustomAttribute<ViewContextAttribute>();
-
-                    if (viewContextAttribute != null && property.PropertyType == typeof(ViewContext))
-                    {
-                        _viewContextSetter = (helper, context) => setter(helper, context);
-                        continue;
-                    }
-
                     _setters.Add(propertyName, (h, v) =>
                     {
                         object value = null;
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/TagHelpers/LiquidTagHelperActivator.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/TagHelpers/LiquidTagHelperActivator.cs
@@ -52,15 +52,15 @@ namespace OrchardCore.DisplayManagement.Liquid.TagHelpers
                 }
 
                 var setter = MakeFastPropertySetter(type, property);
+                var viewContextAttribute = property.GetCustomAttribute<ViewContextAttribute>();
+                if (viewContextAttribute != null && property.PropertyType == typeof(ViewContext))
+                {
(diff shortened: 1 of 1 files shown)


src/OrchardCore/OrchardCore.DisplayManagement.Liquid/TagHelpers/LiquidTagHelperActivator.cs
```diff
@@ -52,15 +52,15 @@ public LiquidTagHelperActivator(Type type)
...
+                var viewContextAttribute = property.GetCustomAttribute<ViewContextAttribute>();
+                if (viewContextAttribute != null && property.PropertyType == typeof(ViewContext))
+                {
+                    _viewContextSetter = (helper, context) => setter(helper, context);
+                    continue;
+                }
 
                 foreach (var propertyName in allNames)
                 {
-                    if (propertyName == "ViewContext")
```
ns8482e: Code stays here just add condition to compare propertyName ==  viewContentProperty

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/TagHelpers/LiquidTagHelperActivator.cs
```diff
@@ -52,15 +52,15 @@ public LiquidTagHelperActivator(Type type)
                 }
 
                 var setter = MakeFastPropertySetter(type, property);
+                var viewContextAttribute = property.GetCustomAttribute<ViewContextAttribute>();
+                if (viewContextAttribute != null && property.PropertyType == typeof(ViewContext))
+                {
+                    _viewContextSetter = (helper, context) => setter(helper, context);
```
ns8482e: just assign property name to new variable
`var viewContentProperty`

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/TagHelpers/LiquidTagHelperActivator.cs
hishamco: That's why I asked before `allNames` members of  type `HtmlAttributeNameAttribute` I'm not sure if it is by intention or not

ns8482e: Ideally property with `[ViewContext]` should also have `[HtmlAttributeNotBound]` so that there is no html binding for such property. So guess your code seems correct.

hishamco:
@jtkech your review please

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/TagHelpers/LiquidTagHelperActivator.cs
jtkech: Yes, regarding the existing code @ns8482e is right.

But because viewContext can't be html bound => Looks good to me

Maybe a new line after line 54 and / or line 55

I would have put line 54 to 60 at the beginning of the loop

jtkech:
I did a little review, you can just put your additional check at the beginning of the loop.

Otherwise LGTM because viewContext can't be html bound

@ns8482e regarding your comments, any other feedback?

I would like if you or @ns8482e could try it and confirm that it works

Then i will approve it if there is no other feedback

ns8482e:
It would also be nice if you could also check datatype of the property to be always `ViewContext` so that following property def be ignored.

```
[ViewContext]
public string Vc { get; set; }
```

ns8482e:
My bad - deleted prev comment - LGTM

hishamco:
I will update the PR soon, hope @ns8482e check it before final approval

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/TagHelpers/LiquidTagHelperActivator.cs
hishamco: > I would have put line 54 to 60 at the beginning of the loop

Why? I think it will be executed more than what expected, unless there's a proper reason, I will change the code for what you are suggesting, but hopefully you answer my questions, so everything will be clear for me ;)

jtkech:
@hishamco 

> I would have put line 54 to 60 at the beginning of the loop

I was talking about the beginning of the `foreach (var property in accessibleProperties)` loop ;)

Not the `foreach (var propertyName in allNames)` coz here you will call `MakeFastPropertySetter` for all names which is unnecessary, the setter needs to be built once for a given property.

hishamco:
> I was talking about the beginning of the foreach (var property in accessibleProperties) loop ;)

Not the foreach (var propertyName in allNames) coz here you will call MakeFastPropertySetter for all names which is unnecessary, the setter needs to be built once for a given property.

I see, coz I confused little bit from the original code ;)

Issue #3435, opened 2019-04-04
Register PAge


aghili371:
I want to have register Page With Extra field like(name, address, phone,...)
i should get a clone from OrchardCore.Users Module and redesign controller and views or you have a better solution?

sebastienros:
You can already change the view by copying it in your theme (or module) in the `Views\OrchardCore.Users\Registration` folder.

Then there is an interface in the `OrchardCore.Users.Abstractions` that is named `IRegistrationFormEvents` that will let you execute some custom code when the form is validated and the user is created. You will need to resolve the current http request and access the form by yourself. Also you will need to query the user from the user store and then update its data (or store it somewhere else). If you use the `User` class it has a `Properties` object that you can use to store custom data.

Maybe we should add more properties to these events to simplify your work but it looks totally doable.

Let us know how it goes.

For my own reference, this is how it's done in WP: https://codex.wordpress.org/Customizing_the_Registration_Form

sebastienros:
I think this issue will also help you: https://github.com/OrchardCMS/OrchardCore/issues/3353

aghili371:
this projct it's good for redesign the user pages.
[ModernBusiness.OC.RazorPages](https://github.com/psijkof/ModernBusiness.OC.RazorPages/tree/master/src/Modules.Pages/ModernBusiness.Pages.Users)

just I have a problem , how I can tell the page to use the default layout in my theme ("Layout.liquid")?

aghili371:
i used "properties " to save extra fields, it works fine but when i want to have listed and edit the information, it was difficult to use.
i created new content type "user info" and i added the field here.

the problem is how to connect this content type and "user register info".

i you set an extra field in the identity of the user to get the user id, it will be great.

i can get the username by :
 User.Identity.Name;

do you have any field to get userid?

PR #5650, opened 2020-02-29
Create a abstract service, UserPropertiesService,  for user

infofromca:


infofromca:
fixed #3435

deanmarcussen:
I'm not sure I understand how this is fixing the issue you mention?

The `Properties` update methods you have added are already available as extensions methods on `EntityExtensions`

It looks like the `UserManager` already does what is necessary to get a user, so I wonder why a wrapper service is helpful?

infofromca:
it is still diff.
1-- mine deal with user, for exaple, null
2--update the use directly.

for get , just benefit from the injection of _userManager and make it clearly.  if not necessary, we can delete it.

deanmarcussen:
From your issue here https://github.com/OrchardCMS/OrchardCore/issues/5916 I don't see how this service would work?

PR #5651, opened 2020-02-29
ViewComponents should be searched in Views & Pages

### LINKED ISSUE #5642: Cannot find Razor Pages ViewComponent View
When using a ViewComponent in a razor page, the module isn't detected as ComponentView containing module.

It appears that the ComponentViewLocationExpanderProvider only checks for razor files in "/Views/Shared/Components". I think the location for views of ViewComponents in Razor Pages is: "/Pages/Components".

When I add a dummy cshtml file in the /Views/Shared/Components folder, the module is included to the search paths, but only "/Views/Shared/Components" and not "/Pages/Components".

The screenshot displays the list of paths. The view component from the module Idella.Cms.Module is displayed in the Idella.ExternalProfiles module.

image

ps. The culture variants are also not added to the search paths.

hishamco:
Fixes #5642

### DIFF
--- a/src/OrchardCore/OrchardCore.Mvc.Core/LocationExpander/ComponentViewLocationExpanderProvider.cs
+++ b/src/OrchardCore/OrchardCore.Mvc.Core/LocationExpander/ComponentViewLocationExpanderProvider.cs
@@ -23,6 +23,8 @@ namespace OrchardCore.Mvc.LocationExpander
         private readonly ShellDescriptor _shellDescriptor;
         private readonly IMemoryCache _memoryCache;
 
+        private bool _initialized;
+
         public ComponentViewLocationExpanderProvider(
             RazorCompilationFileProviderAccessor fileProviderAccessor,
             IExtensionManager extensionManager,
@@ -33,20 +35,19 @@ namespace OrchardCore.Mvc.LocationExpander
             _shellDescriptor = shellDescriptor;
             _memoryCache = memoryCache;
 
-            if (_modulesWithComponentViews != null)
+            if (_modulesWithComponentViews != null && _modulesWithPagesComponentViews != null)
             {
                 return;
             }
 
             lock (_synLock)
             {
+                var orderedModules = _extensionManager.GetExtensions()
+                        .Where(e => e.Manifest.Type.Equals("module", StringComparison.OrdinalIgnoreCase))
+                        .Reverse();
                 if (_modulesWithComponentViews == null)
                 {
                     var modulesWithComponentViews = new List<IExtensionInfo>();
-
-                    var orderedModules = _extensionManager.GetExtensions()
-                        .Where(e => e.Manifest.Type.Equals("module", StringComparison.OrdinalIgnoreCase))
-                        .Reverse();
 
                     foreach (var module in orderedModules)
                     {
@@ -62,30 +63,18 @@ namespace OrchardCore.Mvc.LocationExpander
 
                     _modulesWithComponentViews = modulesWithComponentViews;
                 }
-            }
 
-            if (_modulesWithPagesComponentViews != null)
-            {
-                return;
-            }
-
-            lock (_synLock)
-            {
                 if (_modulesWithPagesComponentViews == null)
                 {
                     var modulesWithPagesComponentViews = new List<IExtensionInfo>();
 
-                    var orderedModules = _extensionManager.GetExtensions()
-                        .Where(e => e.Manifest.Type.Equals("module", StringComparison.OrdinalIgnoreCase))
-                        .Reverse();
-
                     foreach (var module in orderedModules)
                     {
-                        var moduleComponentsViewFilePaths = fileProviderAccessor.FileProvider.GetViewFilePaths(
+                        var moduleWithPagesComponentsViewFilePaths = fileProviderAccessor.FileProvider.GetViewFilePaths(
                             module.SubPath + "/Pages/Shared/Components", RazorExtensions,
                             viewsFolder: null, inViewsFolder: true, inDepth: true);
 
-                        if (moduleComponentsViewFilePaths.Any())
+                        if (moduleWithPagesComponentsViewFilePaths.Any())
                         {
                             modulesWithPagesComponentViews.Add(module);
                         }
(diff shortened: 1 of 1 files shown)


sebastienros:
I am ok with it if @jtkech agrees.

jtkech:
Maybe not necessary but I'm okay because of the mvc doc saying

    The runtime searches for the view in the following paths:

    /Views/{Controller Name}/Components/{View Component Name}/{View Name}
    /Views/Shared/Components/{View Component Name}/{View Name}
    /Pages/Shared/Components/{View Component Name}/{View Name}

But after a quick look seems that the PR is incomplete. It adds modules defining views under `/Pages/Shared/Components`, **but it doesn't add the related locations as done for the `SharedViewsPath`** that is equal to `"/Views/Shared/{0}" + RazorViewEngine.ViewExtension`.

                    moduleComponentViewLocations = _modulesWithComponentViews
                        .Where(m => enabledExtensionIds.Contains(m.Id))
                        .Select(m => '/' + m.SubPath + SharedViewsPath)
                        .ToArray();

Would need to also define e.g a `PageSharedViewsPath = "/Pages/Shared/{0}" + RazorViewEngine.ViewExtension` and then use it as above to add the related locations.

Also maybe use 2 collections `_modulesWithComponentViews` and `_modulesWithPageComponentViews` just to not add useless locations (but less important).

And then try it.

src/OrchardCore/OrchardCore.Mvc.Core/LocationExpander/ComponentViewLocationExpanderProvider.cs
```diff
@@ -56,6 +56,15 @@ public class ComponentViewLocationExpanderProvider : IViewLocationExpanderProvid
...
+
+                        moduleComponentsViewFilePaths = fileProviderAccessor.FileProvider.GetViewFilePaths(
+                            module.SubPath + "/Pages/Shared/Components", RazorExtensions,
+                            viewsFolder: null, inViewsFolder: true, inDepth: true);
+
+                        if (moduleComponentsViewFilePaths.Any())
+                        {
+                            modulesWithComponentViews.Add(module);
+                        }
                     }
```
jtkech: Looks good but then you need to add the related locations as done for the SharedViewsPath ...

See my last comment in the thread

src/OrchardCore/OrchardCore.Mvc.Core/LocationExpander/ComponentViewLocationExpanderProvider.cs
```diff
@@ -56,18 +58,40 @@ public class ComponentViewLocationExpanderProvider : IViewLocationExpanderProvid
...
+                    }
+
+                    _modulesWithComponentViews = modulesWithComponentViews;
+                }
+            }
 
-                        moduleComponentsViewFilePaths = fileProviderAccessor.FileProvider.GetViewFilePaths(
+            if (_modulesWithPagesComponentViews != null)
+            {
+                return;
```
jtkech: So, because at the beginning line 36 we still have the following, we may return if `_modulesWithComponentViews` is initialized while `_modulesWithComponentViews` may not be initalized yet.

            if (_modulesWithComponentViews != null)
            {
                return;
            }

So the suggestion is to use another `private bool _initialized` field to do the double check around the lock, then initialize both collections in the same lock and where you can re-use the same `orderedModules` computed once, then set `_initialized` to true at the end after having initialized both static collections.

src/OrchardCore/OrchardCore.Mvc.Core/LocationExpander/ComponentViewLocationExpanderProvider.cs
```diff
@@ -112,6 +136,27 @@ public void PopulateValues(ViewLocationExpanderContext context)
                 }
 
                 result.AddRange(moduleComponentViewLocations);
+
+                if (!_memoryCache.TryGetValue(CacheKey, out string[] moduleWithPagesComponentViewLocations))
+                {
+                    var enabledIds = _extensionManager.GetFeatures().Where(f => _shellDescriptor
```
jtkech: So here i would have used only one cache entry so that we can re-use the same `enabledIds` and `enabledExtensionIds` to still build 2 local location collections by using `_modulesWithComponentViews` and `_modulesWithPagesComponentViews`, but then concat them to update an unique cache entry.

ns8482e:
> Maybe not necessary but I'm okay because of the mvc doc saying

If ViewComponent is located at `[module]/Pages/Shared/Components` and used in shape - that executes from RazorPages will work fine. However I doubt that same shape execution would continue to give error reported on #5642 if executed from controller view. IMHO It needs better documentation.

src/OrchardCore/OrchardCore.Mvc.Core/LocationExpander/ComponentViewLocationExpanderProvider.cs
hishamco: I was thinking about that, but I kept everything as its to avoid undesired results

src/OrchardCore/OrchardCore.Mvc.Core/LocationExpander/ComponentViewLocationExpanderProvider.cs
hishamco: > but then concat them to update an unique cache entry.

I didn't get this, l will update the PR, feel free to add your commits then after if not meet your suggestions

hishamco:
@jtkech I updated the PR, feel free to add your comments if I miss understand you in some of your suggestions

jtkech:
Okay no problem i'll take a look tonight

jtkech:
I updated your PR to show you what i meant

ns8482e:
@hishamco @jtkech  is this working for you?

 It's giving me compilation error when I add component at `/Pages/Shared/Components` and try to use `@Model`

`The name 'Model' does not exist in the current context`

jtkech:
Just tried to move the `OC.ContentTypes` `SelectContentTypes` component from the `Views` folder to the `Pages` folder, and tried to add a content picker and edit its setting that uses this component.

It works but i also needed to copy paste the `_ViewImports.cshtml` e.g under `Pages/Shared`.

ns8482e:
I already have  `/Pages/_ViewImports.cshtml` with following
```

@inherits OrchardCore.DisplayManagement.RazorPages.Page
@addTagHelper *, OrchardCore.Contents.TagHelpers

```

Do you mean I need to create another at `/Pages/Shared/_ViewImports.cshtml`

jtkech:
Yes or in any subfolder e.g `/Pages/Shared/Components` and that contains

Issue #5633, opened 2020-02-25
Updating a ContentItem with bag items, causes all bag ContentItemId's to be changed


mawendel:
For some reason all bag items have their ContentItemId's updated if you save through the admin UI. 

If i save through updateContentItem() from a workflow script task they are not changed which is exactly what i need and would expect.

sebastienros:
> For some reason

The reason is that when you update the content item in the editor, the server will re-create the content item completely. When you update the json by yourself, you are only changing the parts you care about.

Why is it an issue? Eventually we could remove all the content item ids from the payload, it should not break anything. So the next question is why are you using these ids?

mawendel:
Because i'm referencing the ContentItemId in the bag from another separate content item. 

ContentItem
- Bag1
-- Bag1ContentItemId: asdf234wrefg5
-- Bag2ContentItemId: dsfghdrth3443
-- Bag3ContentItemId: sdfghhw45624

Then from another content item i reference the ContentItemId + the Bag2ContentItemId to get the subselection.

sebastienros:
Did you create your own "picker"

You should use an alias part instead. This is how we will solve routing for items in bag parts too.

@deanmarcussen but I have a doubt now, do we need ids to be constant when there is no alias?

mawendel:
Ok, a little context.

- For the top content item, there can be several thousands. 
- The content items in the bag part, is completely controlled by the user (from a specific content type of course). So they can setup how many they want and config these with typical 2-10.
- Other users can then view these options and select one of them (through front-end Views).

Normally you would have a FK to this table. But here i would just iterate through all the bag items to find the correct one based on its ContentItemId.

However, i guess i could use a generated alias (custom id) if it was possible to reference nested bag items? Maybe i misunderstand the implementation, but i dont see a way to use the alias though. It saves just a ContentItemId in the AliasPartIndex, but is that the parent content item or the bag part content item? 

This also only assumes one level of bag items. If you have nested two bag items then it would need to save three ContentItemId's. ParentId, FirstBagId, FinalBagSelectionId. Unless of course an alias could be searched iteratively:

```
var parentItem = getItemByAlias("UserPage1234");
var firstBagItem = getBagItemByAlias(parentItem, "SubSelection1");
var secondBagItem = getBagItemByAlias(firstBagItem, "SubSelection2");
```

[...]

PR #5653, opened 2020-03-02
Maintain content item id across BagPart.ContentItems

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/5633

Only done for BagParts, not Flows or WidgetList.

Necessary for contained content item routing.

Plus not hard to achieve.

For me, it makes sense for the content item id to either remain consistent, or not be there.

So as it is useful to have it...

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Flows/Drivers/BagPartDisplay.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Flows/Drivers/BagPartDisplay.cs
@@ -68,7 +68,7 @@ namespace OrchardCore.Flows.Drivers
             for (var i = 0; i < model.Prefixes.Length; i++)
             {
                 var contentItem = await _contentManager.NewAsync(model.ContentTypes[i]);
-                var existing = part.ContentItems.FirstOrDefault(x => x.ContentItemId == model.Prefixes[i]);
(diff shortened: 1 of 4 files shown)


src/OrchardCore.Modules/OrchardCore.Flows/Drivers/BagPartDisplay.cs
```diff
@@ -62,16 +62,25 @@ public override async Task<IDisplayResult> UpdateAsync(BagPart part, UpdatePartE
...
             await context.Updater.TryUpdateModelAsync(model, Prefix);
 
-            part.ContentItems.Clear();
+            //part.ContentItems.Clear();
+            var contentItems = new List<ContentItem>();
 
             for (var i = 0; i < model.Prefixes.Length; i++)
             {
                 var contentItem = await _contentManager.NewAsync(model.ContentTypes[i]);
+                var existing = part.ContentItems.FirstOrDefault(x => x.ContentItemId == model.Prefixes[i]);
```
sebastienros: Case Compare.OrdinalIgnoreCase ?

deanmarcussen: Done

PR #5659, opened 2020-03-03
Placement documentation small fixes

scleaver:
Fix incorrect field differentiator format needs - (dash) instead of . (dot)
Make the placement rules definition a little clearer.


hishamco:
I was looking for that issue Dean, thanks that you mention it here

Issue #5662, opened 2020-03-03
UserTaskEvent Localiser Unset


duncanhoggan:
Currently the UserTaskEvent task is throwing Object not set to instance of Object error as the `IStringLocalizer T` is not set.

```
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=OrchardCore.Workflows
  StackTrace:
   at OrchardCore.Workflows.UserTasks.Activities.UserTaskEvent.<GetPossibleOutcomes>b__18_0(String x) in C:\Developement\RandD\CMS\OrchardCore\src\OrchardCore.Modules\OrchardCore.Workflows\UserTasks\Activities\UserTaskEvent.cs:line 45
```
It appears that someone came in to do some localisation updates but missed swapping the localiser that is used.

agriffard:
Fixed by #5664

PR #5663, opened 2020-03-03
Fixed string localiser

duncanhoggan:
Looks like the reference to the string localiser was never swapped for the UserTaskEvent.
There may be more of these, but this is a blocker on a project at the moment.
#5662


duncanhoggan:
Want to rebase before this goes

duncanhoggan:
Rebase done, nice and clean now

jtkech:
Yes i agree, there is a remaining T that has not been replaced with S

Yes, open a new PR then i will approve it

hishamco:
Strange!! I scan the entire source few months ago, seems we need a roslyn analyzer to do this in out behalf ;)

jtkech:
Yes because the T property is still defined and used in one place, so ...

Issue #5654, opened 2020-03-02
Taxonomy Term validation errors fail to maintain route data


deanmarcussen:
When a Term being edited has a validation error, the route data is not maintained correctly.

This results in a not found error on the next save (when validation has been corrected).

PR #5666, opened 2020-03-03
Maintain taxonomy / term hidden inputs when validation fails

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/5654

These values are applied via a hidden input, so must always be present, or model binding fails.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Taxonomies/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Taxonomies/Controllers/AdminController.cs
@@ -98,10 +98,13 @@ namespace OrchardCore.Taxonomies.Controllers
 
             var contentItem = await _contentManager.NewAsync(id);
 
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.Taxonomies/Controllers/AdminController.cs
```diff
@@ -98,10 +98,13 @@ public async Task<IActionResult> CreatePost(string id, string taxonomyContentIte
 
             var contentItem = await _contentManager.NewAsync(id);
 
-            var model = await _contentItemDisplayManager.UpdateEditorAsync(contentItem, _updateModelAccessor.ModelUpdater, true);
+            dynamic model = await _contentItemDisplayManager.UpdateEditorAsync(contentItem, _updateModelAccessor.ModelUpdater, true);
```
hishamco: Why you didn't use `var` as it was before?

deanmarcussen: because the properties are added to a dynamic shape (i.e. to the properties dict). refer code above this

Issue #4107, opened 2019-08-28
Guidance on stateless configuration


mattcrossmarc:
I'm working on building an OrchardCore based web application that will be hosted in AWS on autoscaling containers where state is not preserved.

I understand the need to store data protection keys centrally and plan to use this provider to store them in AWS SSM:

https://aws.amazon.com/blogs/developer/aws-ssm-asp-net-core-data-protection-provider/

And, from the comment linked below, it sounds like it's possible to get to a stateless configuration for other types of data:

https://github.com/OrchardCMS/OrchardCore/issues/3381#issuecomment-477351856

Of the 3 requirements listed in the comment above, I understand how to do 1 & 2, but it's unclear to me how to cause OrchardCore to do 3.

**1. Use a server database like SQL Server or PostgresQL instead of sqlite.** 
I plan to use PostgreSQL in AWS RDS.

**2. Use Azure blob storage for storing media.** 
I plan to use AWS S3 for this by writing a custom media storage module like that done for Azure in `OrchardCore.FileStorage.AzureBlob`.

**3. Using the database to store content types**
How is this enabled/configured? My apologies if this is in the documentation or source and I missed it.

Also, I see a couple of other things currently stored in App_Data and wwwroot that may need to be stored elsewhere. Can you please advise either way?

**4. App_Data/tenants.json**
Can this be stored in the database or elsewhere?

**5. wwwroot/is-cache**
I have seen references to this being a cache only that does not need to be shared across instances. Is that correct?

Thanks for your guidance.

Matt

deanmarcussen:
Orchard Core will work well on AWS, I've had it running on EC2 instances for a POC.

However Distributed services for Orchard Core is still a work in progress by @jtkech . See https://github.com/OrchardCMS/OrchardCore/pull/2247 for some details on the wip, and what issues multiple instances have (namely caching redistribution, often for the admin)

2. S3 will work well as a abstracted media storage module, there is work towards making Azure Blob serve media from blob through Orchard Core, which will get resizing with ImageSharp working, and would also work with S3, refer https://github.com/OrchardCMS/OrchardCore/pull/4025
 
3. To use the database to store content type definitions disable the File Content Definition module. I use this on Azure with different slots. The trick if you already have content type definitions, is to do an export first, then disable, then import to the database.

4. I'm not sure about what is planned for `tenants.json` with distributed services. One thing that can be done with AWS is mounting a shared volume for App_Data, and pointing Orchard Core at it. 

[...]

PR #5551, opened 2020-02-15
ReleaseShellContext() and ReleaseAllShellContexts() methods

jtkech:

PR #5678, opened 2020-03-06
Database and Azure blob shells configuration

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/4107

Proposal for a direct replacement option for the current `IShellsConfigurationSources` and IShellsSettingsSources` for an Azure blob container, rather than `App_Data`.

These services are registered at Application Level, so this is the easy replacement, rather than creating a default shell to resolve through the default tenants database.

It also means you can pick the configuration content (`tenants.json` and each tenants `appsettings.json` straight out of the `App_Data` folder, copy it to a **private** blob container, and start using it.

It's configured by
```
    "OrchardCore.Shells.Azure": {
      "ConnectionString": "UseDevelopmentStorage=true",
      "BasePath": "",                                       // optional
      "ContainerName": "myhost"
    }
```

and adding this to the main host `Startup`

```
services.AddOrchardCms()
    .AddAzureShellsConfiguration();
```

Pros
Simple, and easy to configure

Cons
Because each individual tenants file is stored in its own `appsettings.json` a http request for each file will be incurred on tenant startup. We could store all in one file, and load once, and cache in memory, but that would make transitioning to this feature not as easy (and recipes are not a simple option here, as the site will not be running to import them, although we could have a step to export them, but we would be exported passwords unprotected so not ideal)

Currently this is similar in concept to `Microsoft.Extensions.Configuration.AzureAppConfiguration ` or `Microsoft.AspNetCore.DataProtection.AzureStorage` as it would be a standalone `NuGet` package, which would be referenced separately as and when required. (we could integrate it somewhere by taking a reference on it, but it is an application option, not a module...)

We could also consider implementing something similar to `Microsoft.AspNetCore.DataProtection.AzureKeyVault` as an option to encrypt the configuration files while at rest (although we don't yet offer that feature for our `DataProtection` module. 
It would also make the files non editable, as they would be encrypted in the container.

cc @jtkech


jtkech:
> We could store all in one file, and load once, and cache in memory

Maybe this is from the beginning that it would have been better to not have settings, only configs in one file that would be **cached for us** when building the shared configuration. Could be changed (I did it at some point), otherwise better to keep them separate when we only change the sources.

---

One concern i have is with the json stream provider that uses the json stream parser to populate its data but then **doesn't release the stream**. Maybe because the config is intended to be built once, but in our case the config can be re-built multiple times (e.g when reloading a tenant).

The json file config provider also uses the json stream parser to populate its data, but then it releases the stream. So, maybe worth to do a custom config source / provider that releases the stream.

---

I did a **testing version** of a `DocumentStoreShellsSettingsSources` that right now also uses the json stream provider (so same remark as above). Here the options coming from the main config provides the needed data to connect to the database, then the only difference is to get / update the json data.

Here how i create a shell context with an empty shell descriptor (without any features)

    private Task<ShellContext> GetContextAsync()
    {
        if (_options.DatabaseProvider == null)
        {
            return null;
        }

        var settings = new ShellSettings()
        {
            Name = ShellHelper.DefaultShellName,
            State = Models.TenantState.Running
        };

        settings["DatabaseProvider"] = _options.DatabaseProvider;
        settings["ConnectionString"] = _options.ConnectionString;
        settings["TablePrefix"] = _options.TablePrefix;

        return _shellContextFactory.CreateDescribedContextAsync(settings, new ShellDescriptor());
    }

Then a light scope (not our ShellScope) just to resolve the session and so on.

        using (var scope = context.ServiceProvider.CreateScope())
        {
            var session = scope.ServiceProvider.GetRequiredService<ISession>();

            var document = await session.Query<ShellsSettingsDocument>()
                .FirstOrDefaultAsync() ?? new ShellsSettingsDocument();
            ...

Need to be discussed e.g on skipe but seems that we are not awake at the same times ;)

deanmarcussen:
> Maybe this is from the beginning that it would have been better to not have settings, only configs in one file that would be **cached for us** when building the shared configuration. 

Yes, this could be better. I was concerned about migrating to the feature, but anyone migrating would know enough to merge the existing `appsettings.json` into one file (or document).

> One concern i have is with the json stream provider that uses the json stream parser to populate its data but then **doesn't release the stream**. 

That is naive of me to assume aspnetcore would release the stream! 
Yes we will need a provider that releases the stream. Or a custom builder. I have yet to look and understand what the options are.

> I did a **testing version** of a `DocumentStoreShellsSettingsSources` that right now also uses t

So not actually much more difficult to create a Document, rather than a blobstore 

> Need to be discussed e.g on skipe but seems that we are not awake at the same times ;)

Yes, we need to chat. I'm 50/50 as to whether database or blob is better, and not especially bothered either way, so if anyone else has any preference, feel free to chime in :)

jtkech:
Let's do both ;)

I think we can do both in your PR because only the way to get the json stream is different

I started a custom `DisposableJsonStreamConfigurationProvider` that we could use for both

Need to leave, i will continue this night

As a reminder: Need to be in sync with my shells branch where i update the shell settings in a deferred task, thinking about having to wait a commit if there are stored in database. But if we use a temporary and standalone scope to update them, we don't need a deferred task anymore.

deanmarcussen:
> Let's do both ;)

Why not indeed! Worth noting that for many users looking for distributed they are only running the `Default` tenant anyway, so that suits a `Document` store better.

> I think we can do both in your PR because only the way to get the json stream is different
> 
> I started a custom `DisposableJsonStreamConfigurationProvider` that we could use for both

Sounds exactly what we need.

jtkech:
Just pinged you on skipe, is it too late to have a call?

jtkech:
- I did `LoadSettingsAsync()` and `AddSourcesAsync()`.

- I made some settings / config sources using the database, i followed the same names conventions so that it will be easier if we want to change them.

    Configuration (for testing we can use `Sqlite` but idem with a shared sql database)

      "OrchardCore.Shells.Database": {
        "DatabaseProvider": "Sqlite",
        "ConnectionString": "",
        "TablePrefix": ""
      }

    Usage

        services.AddOrchardCms()
            .AddDatabaseShellsConfiguration();


- I made `DatabaseShellsSettingsSources` related to `tenants.json`.

- I made `DatabaseShellConfigurationSources` related to the `appsettings.json` under the tenant folder. In fact we have only one document for all tenants that we could cache, but when reloading a tenant we really need to reload this document to get the updated settings of this tenant.

    Caching may be useful on startup but we would need two `AddSourceAsync()`, one that would use the cached document and one that would reload it (when we reload a tenant). That said on startup the document will not be loaded multiple times immediately because for each tenant its specific configuration is lazily loaded on a 1st request on this tenant, so will see.

- I didn't make `DatabaseShellsConfigurationSources` related to the `appsettings.json` directly under `App_Data`. When using the blob, do we really need it? Hmm, maybe useful if it's easy to edit a blob file. But when using the database, we would need to edit it through the admin UI, so will see. Hmm, but maybe people would like to add configs by editing the database directly, let me know.

- **Todo:** Find a strategy to migrate from the file system to the blob / database storage. I have some ideas but didn't have the time to implement it.

- **Note:** I'm thinking about removing a tenant, when using a shared database we already have the problem to remove its tables. We also have to remove it from `tenants.json` by editing the file. Idem if using the blob store but maybe less easy with the database. Maybe we could provide a command through the admin UI, and in all cases even if we are using the `tenants.json` file ...

jtkech:
@deanmarcussen sorry to monopolize your PR ;)

I'm gonna leave soon so you can re-work on it

I will come back in around 5 hours

deanmarcussen:
@jtkech no problem! changes look good at first glance. I have no time today, but will take a proper look tomorrow hopefully

jtkech:
- I made a `MigrateFromFiles` option so that if settings or a given configuration is not in the database we try to get it from the related file and then we update the database.

    As we discussed, not so good to cache data as we update the store, better to reload them and / or have an integrity check. But because it is just for a migration path i kept things simple.

- So, you can use `AddDatabaseShellsConfiguration()` from the beginning and setup multiple tenants without creating the settings / config files.

    But for existing tenants, you can use the `MigrateFromFiles` option. The settings will be migrated automatically, but because tenant specific configurations are lazily built, you have to request each tenant (or just wait) or reload them from the admin. Then you can remove the files (or not).

deanmarcussen:
> * I made a `MigrateFromFiles` option so that if settings or a given configuration is not in the database we try to get it from the related file and then we update the database.

I like this (especially as an `Option`)

> * **Update:** Finally i started to do `DatabaseShellsConfigurationSources` related to the `appsettings.json` and `appsettings.{_environment}.json` just under `App_Data`, so that we can migrate even if we were using these files.
>   So here the configuration for "AllEnvironment" and for different environments are all in the same document. Maybe good to have a feature allowing to edit this document but too lazy to do it
>   But still not sure that it is worth to have this one for the database storage, will be easy to revert.

Yes, tricky this one. I wonder if anyone is even using the feature, rather than just `appsettings` in their `Startup`. I wouldn't think to have a feature to edit this. Feels like too much of a super admin privilege who should have access to the database.

I think in the one document is fine.

For info I checked the `.AddJsonStream()` and underlying it uses a `JsonFileParserHelper` which uses a `StreamReader` in a using block, so the `StreamReader` will dispose of the stream when it has finished reading.

So I think it is looking good, knowing that there are other pieces from the other pr's, and discussions we have had to finalize.

I think the migrate option is useful, and we can reuse for blob as well, so I will borrow, and add to the blob configuration.

jtkech:
> ... I wonder if anyone is even using the feature, rather than just appsettings in their Startup

Yes, i will just keep `DatabaseShellsSettingsSources` and `DatabaseShellConfigurationSources`.

> For info I checked the .AddJsonStream() ... so the StreamReader will dispose of the stream ...

Yes as we saw together on skipe

> I think the migrate option is useful, and we can reuse for blob as well

Okay i let you do it for the blob so that you can review what i did for the database.

> So I think it is looking good

Me too ;)

ruisdias:
Morning,
Is there any expectation to this to be released? 
Right now, I'm trying to deploy my CMS to a server which I always get `System.IO.IOException: Read-only file system` because it always tries to create the App_Data/Sites in the server. 

I think this would solve my problem, because even though I enabled the data protection keys Azure feature but it stills passes in the `ShellConfigurationSources` to create the folder and I get the exception.

With the feature enabled, it writes the data protection in the blob storage but always tries to create this folder.

Thanks in the advance,

Rui Dias

deanmarcussen:
@ruisdias yes, this would be expected, and I would hope this feature solves that problem.

In the short term the solution is to create the `Sites` folder in your readonly file system before deployment, or mounting, or allow it as read/write during the setup process, and set it readonly afterwards

jtkech:
One idea on the fly, we could have used a common implementation, even for the one using json files, the extension point being to use a service that just provide a concrete way to get / save a json stream.

deanmarcussen:
> One idea on the fly, we could have used a common implementation, even for the one using json 

JT as I am looking at it now, with all that you have done, I agree.

The only reason I'd kept it separate was to avoid a reference on `Orchard.FileStore.Blob`.

If you agree that it is ok for `OrchardCore.Infrastructure` to have a reference to `Orchard.FileStore.Blob` I will move it there and see if we can have a common interface.

jtkech:
What i meant is that each default `IShellXxxxSources` implementation could have used a new `IShellXxxxJsonStreamSources` interface whose default implementation would expose e.g a `GetJsonStream()` from a json file, and a `SaveJsonStream()` to save data to a json file.

So that for the blob you would have just to override `IShellXxxxJsonStreamSources` to get / save json streams but from the blob, and in my case from the database.

jtkech:
@deanmarcussen 

- Finally i think that's okay to let it as is and not try to make a common interface.

- So maybe you would just have to implement the `MigrateFromFiles` option for the blob, as you want. Maybe just for `tenants.json` and the `appsettings.json` under the tenant folder, maybe not so useful for the more rarely used `appsettings.json` just under `App_Data`.

- Maybe good to try it again for the blob since we have made some methods async.

Otherwise LGTM

deanmarcussen:
@jtkech 
- Agreed 
- Will implement
- Definitily needs a test
- Best update the PR title too ;)

Hope to tomorrow

deanmarcussen:
@jtkech I managed to get there :)

- Implemented `MigrateFromFiles` option. I did for the rarely used as well, as it fitted but maybe not. See what you think.

- Tested, and tested auto migrate to blob.

- Added some default settings to the `appsettings.json` for both.

See what you think, perhaps I should remove the `App_Data` `appsetings.json` ?

jtkech:
@deanmarcussen 

Okay cool

> See what you think, perhaps I should remove the App_Data appsetings.json ?

As you want, that's okay for me

- I merged last dev

- I commented out the configurations examples

- I integrated a little changed related to a race condition in the `_configBuilderFactory`, as done in #5551, but here the factory is now async so i adapted the fix so that if #5551 get merged 1st it will be easier to solve the conflicts.

nuno-tiago-reis:
@deanmarcussen  just a simple question, does this "replace" the azure media blob storage feature in the sense that if I use this I don't need to use the storage specifically for the media?

nuno-tiago-reis:
It also seems to keep a folder called "IdentityModel-Signing-Certificates" in the App_Data folder, is there any way to move that aswell?

deanmarcussen:
@nuno-tiago-reis no it doesn't replace any other storage features.

It **only** moves the shell settings, `tenants.json` and `appsettings.json` out of `App_Data`

Everything else remains.

So you still need media blob, and data protection for azure features

Not sure where `IdentityModel-Signing-Certificates` comes from? 
Are you using AzureAD?

nuno-tiago-reis:
No I'm not, that folder doesn't exist in our cloud environments, it is only being created locally (which is weird).

Either way, that works fine for me, I'll just define the remaining environment variables and tick the features on.

Thank you :)

sonny-kava-up:
> @nuno-tiago-reis no it doesn't replace any other storage features.
> 
> It **only** moves the shell settings, `tenants.json` and `appsettings.json` out of `App_Data`
> 
> Everything else remains.
> 
> So you still need media blob, and data protection for azure features
> 
> Not sure where `IdentityModel-Signing-Certificates` comes from?
> Are you using AzureAD?

@deanmarcussen I'm interested in using OrchardCore for an application that may require scaling out.  This update pushes the files to azure blob storage, but there is no mechanism for syncing settings between independently running services. Do you have any suggestions on how to get this working in a multi service env? Thanks

PR #5680, opened 2020-03-07
Code Mirror editor for TextField

### LINKED ISSUE #5087: Feature request: CodeMirror editor for TextField
Implement a TextField editor (with settings) for CodeMirror

agriffard:
Fixes #5087

Editor option:
image

Result in Edit:
image

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Views/TextField-CodeMirror.Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Views/TextField-CodeMirror.Edit.cshtml
@@ -6,7 +6,7 @@
 }
 
 <div class="form-group">
-    <label asp-for="Text">@Model.PartFieldDefinition.DisplayName() code</label>
+    <label asp-for="Text">@Model.PartFieldDefinition.DisplayName()</label>
     <input asp-for="Text" class="form-control content-preview-text" dir="@culture.GetLanguageDirection()" />
     @if (!String.IsNullOrEmpty(settings.Hint))
     {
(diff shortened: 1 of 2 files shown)


src/OrchardCore.Modules/OrchardCore.ContentFields/Views/TextField-CodeMirror.Edit.cshtml
```diff
@@ -0,0 +1,33 @@
...
+    $(function () {
+        var optionsTexArea = document.getElementById('@Html.IdFor(x => x.Text)');
+        var editor = CodeMirror.fromTextArea(optionsTexArea, {
+            lineNumbers: false,
+            lineWrapping: true,
+            matchBrackets: true,
+            mode: { name: "htmlmixed" }
+        });
+        setTimeout(function() { editor.refresh(); },1);
+    });
```
deanmarcussen: @agriffard I am still confused why we need the `setTimeout` here, but not other places?
is it because the editor option is initially hidden, and codemirror doesn't draw initially when unhidden?

agriffard: If the editor is not refreshed, it is not resized to show the whole text inside.
Uncomment this line and load the page with a long text to see what it does.

deanmarcussen: `autoRefresh: true,` is the code mirror option we use elsewhere, and works for `Autoroute` (I will see later when I have time to see if it works here)

agriffard:
There is a pb: The newlines disappear when you re-edit an item.

PR #5362, opened 2020-01-25
Users Change email feature

agriffard:
Adds a feature and settings to allow the users to change their email.

Fixes #1630

PR #5688, opened 2020-03-09
Change Email review

agriffard:
React to @sebastienros's comments.

Related to #5362


agriffard:
Reminder: If you just check with MimeKit `!MailboxAddress.TryParse(Email, out var emailAddress)`, an email without any `@` would work, so I also returned un invalid email if `Email.IndexOf('@') == -1`.

https://github.com/jstedfast/MimeKit/issues/465

hishamco:
@agriffard you could rebase on dev after merging #5710

agriffard:
ok, dev merged and IEmailAddressValidator used in ChangeEmailViewModel.

src/OrchardCore.Modules/OrchardCore.Users/ViewModels/ChangeEmailViewModel.cs
```diff
@@ -1,13 +1,21 @@
+using System;
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Localization;
-using MimeKit;
+using Microsoft.Extensions.Localization;
```
hishamco: Remove extra whitespace

src/OrchardCore.Modules/OrchardCore.Users/ViewModels/ChangeEmailViewModel.cs
```diff
@@ -1,13 +1,21 @@
...
 using System.ComponentModel.DataAnnotations;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Localization;
-using MimeKit;
+using Microsoft.Extensions.Localization;    
+using OrchardCore.Email;
 
 namespace OrchardCore.Users.ViewModels
 {
     public class ChangeEmailViewModel : IValidatableObject
```
hishamco: Do we really need `IValidatableObject`? `IEmailAddressValidator` can handle the validation logic, so you can check directly in the `POST`

src/OrchardCore.Modules/OrchardCore.Users/Controllers/ChangeEmailController.cs
```diff
@@ -49,7 +55,7 @@ public async Task<IActionResult> Index()
 
             var user = await _userService.GetAuthenticatedUserAsync(User);
 
-            return View(new ChangeEmailViewModel() { Email = ((User)user).Email });
+            return View(new ChangeEmailViewModel(_emailAddressValidator) { Email = ((User)user).Email });
```
sebastienros: The service needs to be resolved in the Vallidate method

src/OrchardCore.Modules/OrchardCore.Users/Controllers/ChangeEmailController.cs
```diff
@@ -38,7 +38,7 @@ public class ChangeEmailController : Controller
             _userService = userService;
             _userManager = userManager;
             _siteService = siteService;
-            _emailAddressValidator = emailAddressValidator ?? throw new ArgumentNullException(nameof(emailAddressValidator));
+            _emailAddressValidator = emailAddressValidator;
```
hishamco: I'm not sure why we don't check from `null` in constructor which is useful for `public` APIs, if the intend to prevent DI from throws, it will if the service is not there

deanmarcussen: The DI container will throw a DI exception before it ever reaches this null check so a null check here is unreachable code

hishamco: How? I think the check need to be in services public APIs, not in controller as you said

src/OrchardCore.Modules/OrchardCore.Users/Controllers/ChangeEmailController.cs
agriffard: OK: Resolve EmailAddressValidator from Validate method

Issue #5689, opened 2020-03-09
GraphiQL exception when using a malformed Query Schema 


Fermain:
When defining [Lucene Queries](https://docs.orchardcore.net/en/dev/docs/reference/modules/Queries/), the Schema field can cause GraphiQL to return an exception:

`Uncaught (in promise) Error: Invalid or incomplete introspection result. Ensure that you are passing "data" property of introspection response and no "errors" was returned alongside: undefined`

Issue can be reproduced by inserting a schema with an inappropriate content item name:

`{
    "type": "ContentItem/SomeContentItemType"
}`

mkgn:
This problem still happens. Refer my last comments to #5782. After that, I did a brand new installation with empty Postgres db. Lucene & GraphQL modules enabled but I haven't added any Lucene queries. Only the index is created. Still when I go to GraphQL section I get the error.

PR #5690, opened 2020-03-09
Added null checks around Lucene queries linked to content types.

duncanhoggan:
When building the GraphQl schema specifically for License queries we now ignore content types that don't exist.

#5689


deanmarcussen:
I think this is probably good to go (I fixed the build after we created an issue with it during triage)

Issue #5694, opened 2020-03-09
"Attached" Media Field No Longer Able to Upload File


jeffolmstead:
I am trying to determine what changed and will add an update here when available. As of right now though (code updated a couple of days ago) a Media Field set up as "Attached" will throw an error of 
```
{Field Name}: There was an error handling the files.
```

hishamco:
@deanmarcussen any idea?

deanmarcussen:
@jeffolmstead I can't repro.

Can you look at your logs, as upload issues are generally well logged.

jeffolmstead:
@deanmarcussen sorry for delay, my computer was turning into a brick so I had to clean it off - back to performance now. No logs are being thrown unfortunately. 

I did some playing around and, if I turn off Azure Media Storage, it functions properly. If I turn it back on, it doesn't work. Azure Media Storage is working fine for the "normal" uploading of media, just seems to be this field.

Here is my text example:

* ContentType: TestUpload
* ContentParts: none
* ContentFields: MediaField - ContentPartFieldSettings: Editor: "Attached"

When my **Azure Media Storage** is enabled, it throws the error.

jeffolmstead:
Interesting, "MediaFieldDriver.cs" on line 72 has it's own try / catch block so none of the issues would seem to make it to a log file:

```
try
{
     await _attachedMediaFieldFileService.HandleFilesOnFieldUpdateAsync(items, context.ContentPart.ContentItem);
}
catch (Exception)
{
     updater.ModelState.AddModelError(Prefix, S["{0}: There was an error handling the files.", context.PartFieldDefinition.DisplayName()]);
}
```

Does that make sense or should the exception be passed back to the log? 

The error I have is:
ErrorCode: BlobAlreadyExists
```
The specified blob already exists.
RequestId:e5c6a756-937a-464b-a077-f539d6d1e44b
Time:2020-03-10T13:46:28.5813902Z
Status: 409 (The specified blob already exists.)

ErrorCode: BlobAlreadyExists

Headers:
Server: Windows-Azure-Blob/1.0,Microsoft-HTTPAPI/2.0
x-ms-request-id: e5c6a756-937a-464b-a077-f539d6d1e44b
x-ms-version: 2019-02-02
x-ms-error-code: BlobAlreadyExists
Date: Tue, 10 Mar 2020 13:46:28 GMT
Content-Length: 220
Content-Type: application/xml
```

The stack trace is:
```
[...]

PR #5696, opened 2020-03-10
Fix Azure error when using TryCreateDirectory

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/5694

Tests properly on `TryCreateDirectoryAsync` to see if either
- a file exists with the name of the directory, or
- the directory already exists

Also adds exception logging to the `MediaFieldDisplayDriver`


DariuS231:
Who can approve this PR? It would be good to have it as soon as possible as it is currently blocking us.

PR #5703, opened 2020-03-11
Add tenant description

### LINKED ISSUE #5638: Add a description to the tenant list
Add a description to the tenant list

截图录屏_选择区域_20200226201124

hishamco:
Fixes #5638

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
@@ -91,7 +91,7 @@ namespace OrchardCore.Tenants.Controllers
                 {
                     var entry = new ShellSettingsEntry
                     {
-                        Description = x.Description,
+                        Description = x.["Description"],
                         Name = x.Name,
                         ShellSettings = x,
                         IsDefaultTenant = string.Equals(x.Name, ShellHelper.DefaultShellName, StringComparison.OrdinalIgnoreCase)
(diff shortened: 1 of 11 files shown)


jtkech:
@hishamco 

The `ShellSettings` uses 2 underlying `ShellConfiguration` => `_settings` and `_configuration`. The 1st one is for the minimum set of properties needed to run a tenant and that are saved in `tenants.json` (if not pre-configured from another config source), the 2nd is used for other configuration values.

And for a given `shellSettings` when you are using `shellSettings["SomeName"]` it uses the 2nd one.

So here it's better to not define a new `shellSettings.Description` and in your code just use `shellSettings["Description"]`, as an example see how it is already done for `shellSettings["Secret"]`.

hishamco:
> So here it's better to not define a new shellSettings.Description and in your code just use shellSettings["Description"]

I just asked myself at the beginning why there's two options ;) I will update the PR now

hishamco:
@jtkech your could merge if it looks good to you now ;)

agriffard:
PR build failed:
`Controllers\AdminController.cs(94,41): error CS1001: Identifier expected [C:\projects\orchardcore\src\OrchardCore.Modules\OrchardCore.Tenants\OrchardCore.Tenants.csproj]`

hishamco:
Oops, thanks Antoine

sebastienros:
@jtkech can you approve (or not)

jtkech:
As i can see LGTM so i will aprove it

But do we need to wait for someone to try it?
did somebody already try it?

If needed i'll try to take the time to try it this w-end

hishamco:
I tried creating and editing a tenant with description, someone could try the workflow if it's necessary coz I didn't deal with it so far

PR #5238, opened 2020-01-09
Adding unit tests for SmtpService

sebastienros:

PR #5710, opened 2020-03-12
Add EmailAddressValidator service

hishamco:
Address the comment https://github.com/OrchardCMS/OrchardCore/pull/5238#discussion_r364932265


hishamco:
I think it's better to move `IEmailAddressValidator` to `Infrastructure` or `Core.Abstraction` and provide a default implementation should be activated without enable email service, your thoughts @sebastienros

sebastienros:
Please move the implementation to OrhardCore.Infrastructure, , fix Validate, and add one little unit test.

hishamco:
> Please move the implementation to OrhardCore.Infrastructure,

That's what I suggested in my above comment ;)

hishamco:
One last question shall I register `IEmailAddressValidator` on `AddOrchardCms` extension message is there a better way to register such service?

agriffard:
Error during build:
`Email\EmailAddressValidator.cs(6,42): error CS0246: The type or namespace name 'IEmailAddressValidator' could not be found (are you missing a using directive or an assembly reference?) [C:\projects\orchardcore\src\OrchardCore\OrchardCore.Infrastructure\OrchardCore.Infrastructure.csproj]`

You also need to move IEmailAddressValidator to OrchardCore.Infrastructure.Abstractions

hishamco:
I will fix the build, but I'm asking where the registration of the services should be

agriffard:
Test has failed during build.

hishamco:
It needs a `null` check operator

hishamco:
@agriffard your final review please

src/OrchardCore.Modules/OrchardCore.Users/ViewModels/RegisterExternalLoginViewModel.cs
```diff
@@ -1,13 +1,21 @@
...
 
 namespace OrchardCore.Users.ViewModels
 {
     public class RegisterExternalLoginViewModel : IValidatableObject
     {
+        private readonly IEmailAddressValidator _emailAddressValidator;
+
+        public RegisterExternalLoginViewModel(IEmailAddressValidator emailAddressValidator)
+        {
+            _emailAddressValidator = emailAddressValidator ?? throw new ArgumentNullException(nameof(emailAddressValidator));
```
deanmarcussen: @hishamco you can't pass the `IEmailAddressValidator` into a constructor of a view model.

You need to resolve it from the `ValidationContext`

src/OrchardCore.Modules/OrchardCore.Users/ViewModels/EditUserViewModel.cs
```diff
@@ -1,13 +1,21 @@
...
+using OrchardCore.Email;
 
 namespace OrchardCore.Users.ViewModels
 {
     public class EditUserViewModel : IValidatableObject
     {
+        private readonly IEmailAddressValidator _emailAddressValidator;
+
+        public EditUserViewModel(IEmailAddressValidator emailAddressValidator)
+        {
```
deanmarcussen: Same here, not in the constructor

src/OrchardCore/OrchardCore.Infrastructure/Email/OrchardCoreBuilderExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+        /// <summary>
+        /// Adds e-mail address validator service.
+        /// </summary>
+        /// <param name="builder">The <see cref="OrchardCoreBuilder"/>.</param>
+        public static OrchardCoreBuilder AddEmailAddressValidator(this OrchardCoreBuilder builder)
+        {
+            builder.ConfigureServices(services =>
+            {
+                services.AddTransient<IEmailAddressValidator, EmailAddressValidator>();
+            });
```
deanmarcussen: `Scoped` I think not transient

src/OrchardCore.Modules/OrchardCore.Users/ViewModels/RegisterExternalLoginViewModel.cs
hishamco: It's passed from where it called, no problem I will address your feedback now, I think I did it two times

src/OrchardCore/OrchardCore.Infrastructure/Email/OrchardCoreBuilderExtensions.cs
jtkech: Transient is okay as it has no state and it allows to be injected by a singleton service. It could also have been a singleton, and if we are sure that we always use this implementation, maybe a static component that don't need to be injected.

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs
```diff
@@ -65,7 +67,7 @@ public class ApiController : Controller
             _databaseProviders = databaseProviders;
             _currentShellSettings = currentShellSettings;
             _notifier = notifier;
-
+            _emailAddressValidator = emailAddressValidator ?? throw new ArgumentNullException(nameof(emailAddressValidator));
```
sebastienros: Don't throw exceptions in constructors. Here the DI will fail if the dependency is null

src/OrchardCore.Modules/OrchardCore.Users/ViewModels/EditUserViewModel.cs
```diff
@@ -1,13 +1,21 @@
...
-using MimeKit;
+using OrchardCore.Email;
 
 namespace OrchardCore.Users.ViewModels
 {
     public class EditUserViewModel : IValidatableObject
     {
+        private readonly IEmailAddressValidator _emailAddressValidator;
+
+        public EditUserViewModel(IEmailAddressValidator emailAddressValidator)
```
sebastienros: Don't inject services in these models, instead resolve it in the Validate method like it's done for the localizer

src/OrchardCore/OrchardCore.Infrastructure/Email/OrchardCoreBuilderExtensions.cs
sebastienros: Transient and Singleton are good, but from DI as we need devs to be able to customize the logic.

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs
hishamco: I think throwing exception is good to ensure the services should be registered in DI, otherwise weird things may happen. Also it good for testability

PR #5718, opened 2020-03-13
CodeMirror autorefresh

### LINKED ISSUE #5716: Bug - html body text box in bag
In a bag I enter the htmlbody content.
Then I save the item.
If I refresh and come back to the page and open the bag item the htmlbody is empty.
But as soon as a I click inside the text box the text will show up.

agriffard:
Remove remaining `setTimeout(function() { editor.refresh(); },1);` and use autorefresh with CodeMirror.

Fixes #5716


deanmarcussen:
Hmm. Can't see why the build is failing. 

@agriffard did you try this against the actual issue? if so, great the fix is simple ;)

One day we should make these code mirrors into shapes. Then we could easily implement the Monaco editor as an option

sebastienros:
Cool find

PR #5722, opened 2020-03-14
Fix blob prefix when there is no container base path

deanmarcussen:
Fixes issue reported by @Skrypt with blob file store when there is no base path supplied for the container

### DIFF
--- a/src/OrchardCore/OrchardCore.FileStorage.AzureBlob/BlobFileStore.cs
+++ b/src/OrchardCore/OrchardCore.FileStorage.AzureBlob/BlobFileStore.cs
@@ -305,11 +305,18 @@ namespace OrchardCore.FileStorage.AzureBlob
     internal static class BlogFileStoreExtensions
     {
         /// <summary>
-        /// Blob prefix requires a trailing slash.
+        /// Blob prefix requires a trailing slash except when loading the root of the container.
         /// </summary>
         internal static string NormalizePrefix(this BlobFileStore blobFileStore, string prefix)
         {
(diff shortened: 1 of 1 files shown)


deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/5724

deanmarcussen:
Yes works without blob storage. But you remind me to ask why I made it an extension method when it is not used anywhere else. So have made it a private method so it is clear it will have no impact on normal storage

src/OrchardCore/OrchardCore.FileStorage.AzureBlob/BlobFileStore.cs
```diff
@@ -300,20 +300,18 @@ private async Task CreateDirectoryAsync(string path)
...
         }
-    }
 
-    internal static class BlogFileStoreExtensions
-    {
         /// <summary>
         /// Blob prefix requires a trailing slash except when loading the root of the container.
         /// </summary>
-        internal static string NormalizePrefix(this BlobFileStore blobFileStore, string prefix)
+        private string NormalizePrefix(string prefix)
```
hishamco: That's what I wanna asked you before, I'm wondering why it's an extension method

jtkech:
Just for the fun i would have used a little variant ;)

        private string NormalizePrefix(string prefix)
        {
            prefix = prefix.Trim('/');
            if (prefix.Length == 0)
            {
                return String.Empty;
            }
            else
            {
                return prefix + '/';
            }
        }

PR #5730, opened 2020-03-15
Fix localizer names

hishamco:


hishamco:
@agriffard I think there's no issue to merge this

PR #5734, opened 2020-03-15
Add culture picker to setup page

### LINKED ISSUE #5726: Add Cultures Picker in Setup page
I saw a request from someone in gitter yesterday, then @Skrypt remind me with one of m old PR to add such feature, so I decide to file an issue for this

IMHO setup has some usability issue, so no one will know that the setup is localized until he looked into the code, then the user needs to add a querystring to make this happen, which is weird!! adding such feature will add some value

hishamco:
Fixes #5726

/cc @Skrypt

### DIFF
--- a/src/OrchardCore.Cms.Web/appsettings.json
+++ b/src/OrchardCore.Cms.Web/appsettings.json
@@ -9,7 +9,7 @@
   "OrchardCore": {
     "OrchardCore.Setup": {
       "DefaultCulture": "",
-      "SupportedCultures": [ "en" ]
+      "SupportedCultures": [ "ar", "cs", "da", "de", "el", "es", "fa", "fi", "fr", "he", "hr", "hu", "id", "it", "ja", "ko", "lt", "mk", "nl", "pl", "pt", "ru", "sk", "sl", "sr-CS", "sr-SP", "sv", "tr", "uk", "vi", "zh-CN", "zh-TW" ]
     }
     // Uncomment to configure media storage in Azure Blob Storage.
     //"OrchardCore.Media.Azure":
--- a/src/OrchardCore.Cms.Web/appsettings.json
+++ b/src/OrchardCore.Cms.Web/appsettings.json
@@ -9,7 +9,7 @@
   "OrchardCore": {
     "OrchardCore.Setup": {
       "DefaultCulture": "",
-      "SupportedCultures": [ "ar", "cs", "da", "de", "el", "es", "fa", "fi", "fr", "he", "hr", "hu", "id", "it", "ja", "ko", "lt", "mk", "nl", "pl", "pt", "ru", "sk", "sl", "sr-CS", "sr-SP", "sv", "tr", "uk", "vi", "zh-CN", "zh-TW" ]
+      "SupportedCultures": [ "ar", "cs", "da", "de", "el", "en", "es", "fa", "fi", "fr", "he", "hr", "hu", "id", "it", "ja", "ko", "lt", "mk", "nl", "pl", "pt", "ru", "sk", "sl", "sr-CS", "sr-SP", "sv", "tr", "uk", "vi", "zh-CN", "zh-TW" ]
     }
     // Uncomment to configure media storage in Azure Blob Storage.
     //"OrchardCore.Media.Azure":
--- a/src/OrchardCore.Cms.Web/appsettings.json
+++ b/src/OrchardCore.Cms.Web/appsettings.json
@@ -8,7 +8,7 @@
   },
   "OrchardCore": {
     "OrchardCore.Setup": {
-      "DefaultCulture": "",
+      "DefaultCulture": "", // When using "" the system OS culture will be used
       "SupportedCultures": [ "ar", "cs", "da", "de", "el", "en", "es", "fa", "fi", "fr", "he", "hr", "hu", "id", "it", "ja", "ko", "lt", "mk", "nl", "pl", "pt", "ru", "sk", "sl", "sv", "tr", "uk", "vi", "zh-CN", "zh-TW" ]
     }
     // Uncomment to configure media storage in Azure Blob Storage.
--- a/src/OrchardCore.Cms.Web/appsettings.json
+++ b/src/OrchardCore.Cms.Web/appsettings.json
@@ -9,7 +9,7 @@
   "OrchardCore": {
     "OrchardCore.Setup": {
       "DefaultCulture": "",
-      "SupportedCultures": [ "ar", "cs", "da", "de", "el", "en", "es", "fa", "fi", "fr", "he", "hr", "hu", "id", "it", "ja", "ko", "lt", "mk", "nl", "pl", "pt", "ru", "sk", "sl", "sr-CS", "sr-SP", "sv", "tr", "uk", "vi", "zh-CN", "zh-TW" ]
+      "SupportedCultures": [ "ar", "cs", "da", "de", "el", "en", "es", "fa", "fi", "fr", "he", "hr", "hu", "id", "it", "ja", "ko", "lt", "mk", "nl", "pl", "pt", "ru", "sk", "sl", "sv", "tr", "uk", "vi", "zh-CN", "zh-TW" ]
     }
     // Uncomment to configure media storage in Azure Blob Storage.
     //"OrchardCore.Media.Azure":
--- a/src/OrchardCore.Cms.Web/appsettings.json
+++ b/src/OrchardCore.Cms.Web/appsettings.json
@@ -35,6 +35,8 @@
         "ru",
         "sk",
         "sl",
+        "sr-cyrl-rs",
+        "sr-latn-rs",
         "sv",
         "tr",
         "uk",
--- a/src/OrchardCore.Cms.Web/appsettings.json
+++ b/src/OrchardCore.Cms.Web/appsettings.json
@@ -9,7 +9,39 @@
   "OrchardCore": {
     "OrchardCore.Setup": {
       "DefaultCulture": "", // When using "" the system OS culture will be used
-      "SupportedCultures": [ "ar", "cs", "da", "de", "el", "en", "es", "fa", "fi", "fr", "he", "hr", "hu", "id", "it", "ja", "ko", "lt", "mk", "nl", "pl", "pt", "ru", "sk", "sl", "sv", "tr", "uk", "vi", "zh-CN", "zh-TW" ]
+      "SupportedCultures": [
+        "ar",
+        "cs",
+        "da",
+        "de",
+        "el",
+        "en",
+        "es",
+        "fa",
+        "fi",
+        "fr",
+        "he",
+        "hr",
+        "hu",
+        "id",
+        "it",
+        "ja",
+        "ko",
+        "lt",
+        "mk",
+        "nl",
+        "pl",
+        "pt",
+        "ru",
+        "sk",
+        "sl",
+        "sv",
+        "tr",
+        "uk",
+        "vi",
+        "zh-CN",
+        "zh-TW"
+      ] // "" value (InvariantCulture) is not supported for these
     }
     // Uncomment to configure media storage in Azure Blob Storage.
     //"OrchardCore.Media.Azure":
--- a/src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
@@ -16,7 +16,7 @@ namespace OrchardCore.Setup
     {
         private readonly string _defaultCulture;
         private string[] _supportedCultures = new string[] {
-            "ar", "cs", "da", "de", "el", "en", "es", "fa", "fi", "fr", "he", "hr", "hu", "id", "it", "ja", "ko", "lt", "mk", "nl", "pl", "pt", "ru", "sk", "sl", "sv", "tr", "uk", "vi", "zh-CN", "zh-TW"
+            "ar", "cs", "da", "de", "el", "en", "es", "fa", "fi", "fr", "he", "hr", "hu", "id", "it", "ja", "ko", "lt", "mk", "nl", "pl", "pt", "ru", "sk", "sl", "sr-cyrl-rs", "sr-latn-rs", "sv", "tr", "uk", "vi", "zh-CN", "zh-TW"
         };
 
         public Startup(IShellConfiguration shellConfiguration)
--- a/src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
@@ -22,13 +22,8 @@ namespace OrchardCore.Setup
         {
             var configurationSection = shellConfiguration.GetSection("OrchardCore.Setup");
 
-            _defaultCulture = configurationSection["DefaultCulture"] ?? "";
+            _defaultCulture = configurationSection["DefaultCulture"];
             _supportedCultures = configurationSection.GetSection("SupportedCultures").Get<string[]>();
-            var results = configurationSection.GetSection("SupportedCultures").Get<List<string>>();
-            if (Array.IndexOf(_supportedCultures, null) > -1) {
-                results.Add("");
-            }
-            _supportedCultures = results.ToArray();
         }
 
         public override void ConfigureServices(IServiceCollection services)
--- a/src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
@@ -15,14 +15,16 @@ namespace OrchardCore.Setup
     public class Startup : StartupBase
     {
         private readonly string _defaultCulture;
-        private string[] _supportedCultures;
+        private string[] _supportedCultures = new string[] {
+            "ar", "cs", "da", "de", "el", "en", "es", "fa", "fi", "fr", "he", "hr", "hu", "id", "it", "ja", "ko", "lt", "mk", "nl", "pl", "pt", "ru", "sk", "sl", "sv", "tr", "uk", "vi", "zh-CN", "zh-TW"
+        };
 
         public Startup(IShellConfiguration shellConfiguration)
         {
             var configurationSection = shellConfiguration.GetSection("OrchardCore.Setup");
 
             _defaultCulture = configurationSection["DefaultCulture"];
-            _supportedCultures = configurationSection.GetSection("SupportedCultures").Get<string[]>();
+            _supportedCultures = configurationSection.GetSection("SupportedCultures").Get<string[]>() ?? _supportedCultures;
         }
 
         public override void ConfigureServices(IServiceCollection services)
--- a/src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Routing;
@@ -21,8 +22,8 @@ namespace OrchardCore.Setup
         {
             var configurationSection = shellConfiguration.GetSection("OrchardCore.Setup");
 
-            _defaultCulture = configurationSection["DefaultCulture"];
-            _supportedCultures = configurationSection.GetSection("SupportedCultures").Get<string[]>();
+            _defaultCulture = configurationSection["DefaultCulture"] ?? "";
+            _supportedCultures = configurationSection.GetSection("SupportedCultures").Get<List<string>>().ToArray();
         }
 
         public override void ConfigureServices(IServiceCollection services)
--- a/src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Globalization;
 using System.Linq;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Routing;
@@ -23,7 +24,7 @@ namespace OrchardCore.Setup
         {
             var configurationSection = shellConfiguration.GetSection("OrchardCore.Setup");
 
-            _defaultCulture = configurationSection["DefaultCulture"];
+            _defaultCulture = configurationSection["DefaultCulture"] ?? CultureInfo.InstalledUICulture.Name;
             _supportedCultures = configurationSection.GetSection("SupportedCultures").Get<string[]>() ?? _supportedCultures;
         }
 
--- a/src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Routing;
--- a/src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
@@ -23,7 +23,12 @@ namespace OrchardCore.Setup
             var configurationSection = shellConfiguration.GetSection("OrchardCore.Setup");
 
             _defaultCulture = configurationSection["DefaultCulture"] ?? "";
-            _supportedCultures = configurationSection.GetSection("SupportedCultures").Get<List<string>>().ToArray();
+            _supportedCultures = configurationSection.GetSection("SupportedCultures").Get<string[]>();
+            var results = configurationSection.GetSection("SupportedCultures").Get<List<string>>();
+            if (Array.IndexOf(_supportedCultures, null) > -1) {
+                results.Add("");
+            }
+            _supportedCultures = results.ToArray();
         }
 
         public override void ConfigureServices(IServiceCollection services)
--- a/src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
@@ -15,7 +15,8 @@ namespace OrchardCore.Setup
 {
     public class Startup : StartupBase
     {
-        private readonly string _defaultCulture;
+        private readonly string _defaultCulture = CultureInfo.InstalledUICulture.Name;
+
         private string[] _supportedCultures = new string[] {
             "ar", "cs", "da", "de", "el", "en", "es", "fa", "fi", "fr", "he", "hr", "hu", "id", "it", "ja", "ko", "lt", "mk", "nl", "pl", "pt", "ru", "sk", "sl", "sr-cyrl-rs", "sr-latn-rs", "sv", "tr", "uk", "vi", "zh-CN", "zh-TW"
         };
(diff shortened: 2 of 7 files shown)


Skrypt:
Ok I will adjust the design on this one and yes normally we should use DisplayName as it is localized, the Name is technical.

Skrypt:
orchard

hishamco:
Antoine regarding the drop down I tried to make it above the jumbotron but when I set use`float-right` it becomes inside. I will try find away

agriffard:
@hishamco : It seems that @Skrypt managed to do it. He just added a margin-top.

src/OrchardCore.Modules/OrchardCore.Setup/Views/_Layout.cshtml
```diff
@@ -21,5 +57,22 @@
...
+            var url = window.location.origin;
+
+            if (url.indexOf('?') > -1) {
+                url = url.split('&culture=')[0] + '&';
+            }
+            else {
+                url += '?';
+            }
+
+            url += 'culture=' + culture + '&ui-culture=' + culture;
```
Skrypt: Not sure if this is required that we set CurrentCulture and CurrentUICulture when we do setup things. I think actually that we should only use the CurrentUICulture for this.

hishamco: No problem `ui-culture` is enought

hishamco:
> It seems that @Skrypt managed to do it. He just added a margin-top.

I expect it will work without adding anything ;)

agriffard:
> I expect it will work without adding anything ;)

I meant a class on the jumbotron:

`    <div class="jumbotron mt-5">`

hishamco:
Ya I know what you meant, but I expecting it should work without `mt-5`

Skrypt:
No you can't just use float like that because the entire layout is using display: flex (flexboxes).

hishamco:
I see, seem the new layout system using FlexBox, I'm little outdated in bootstrap underlying layout system ;)

src/OrchardCore.Modules/OrchardCore.Setup/Views/_Layout.cshtml
```diff
@@ -21,5 +57,22 @@
...
+
+            if (url.indexOf('?') > -1) {
+                url = url.split('&ui-culture=')[0] + '&';
+            }
+            else {
+                url += '?';
+            }
+
+            url += 'ui-culture=' + culture;
+            window.location = url;
```
jtkech: i don't remember all details and maybe not so important, so just for infos and as i remember.

Generating / updating url is not so simple, the url may have other parameters (not only `ui-culture`) the url may end with a fragment string `#...` and so on. And i don't talk about encoding because i assume that the culture keys / values don't need to be encoded.

Take a look e.g at what the `QueryHelpers.AddQueryString()` [does here](https://github.com/dotnet/aspnetcore/blob/116799fa709ff003781368b578e4efe2fa32e937/src/Http/WebUtilities/src/QueryHelpers.cs#L85).

hishamco: FYI I didn't ignore the currents params if there is

Skrypt: I agree, but as of now, we don't have any other params on this url ; so I don't think it will be an issue.

Skrypt:
The only thing I can see is that people choosing a specific culture for the setup won't automatically install this culture as a SupportedCulture. This is not related. We should eventually add on the setup a list of culture to install from the setup that will be totally independent of that culture picker.

src/OrchardCore.Modules/OrchardCore.Setup/Views/_Layout.cshtml
```diff
@@ -1,4 +1,17 @@
+﻿@using Microsoft.Extensions.Configuration
+@using System.Globalization
+@using OrchardCore.Environment.Shell.Configuration
+@inject Microsoft.AspNetCore.Mvc.Localization.IViewLocalizer T
 @inject OrchardCore.IOrchardHelper Orchard
+@inject IShellConfiguration ShellConfiguration
+
+@{
+    var configurationSection = ShellConfiguration.GetSection("OrchardCore.Setup");
+    var currentCulture = Orchard.HttpContext.Request.Query.ContainsKey("ui-culture")
```
sebastienros: Can't we use the default culture providers to pick the correct culture, and select the drop down based on the "CurrentCulture" ?

src/OrchardCore.Modules/OrchardCore.Setup/Views/_Layout.cshtml
jtkech: > FYI I didn't ignore the currents params if there are

Again maybe not so important but the part of your code that checks an existing `ui-culture` query parameter and then replaces it, looks wrong for me.

Parsing / updating a query string is not so simple.

E.g you are splitting on `&ui-culture=` so seems that you only keep the previous parameter(s), and what happens if `ui-culture=` is the 1st parameter `...?ui-culture=...` without starting with the `&` separator.

src/OrchardCore.Modules/OrchardCore.Setup/Views/_Layout.cshtml
hishamco: Probably we can

src/OrchardCore.Modules/OrchardCore.Setup/Views/_Layout.cshtml
hishamco: > what happens if ui-culture= is the 1st parameter ...?ui-culture=... without starting with the & separator.

Probably I forgot this case

> I agree, but as of now, we don't have any other params on this url ; so I don't think it will be an issue.

There will be a params if you 're use the setup from tenants page

jtkech: Maybe the [`URLSearchParams` interface might be helpful](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)

Skrypt: Ok you guys fix that part. I added the required changes requested by @sebastienros

hishamco: Noo, that's why my push refuse ;)

src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
```diff
@@ -21,8 +22,8 @@ public Startup(IShellConfiguration shellConfiguration)
         {
             var configurationSection = shellConfiguration.GetSection("OrchardCore.Setup");
 
-            _defaultCulture = configurationSection["DefaultCulture"];
-            _supportedCultures = configurationSection.GetSection("SupportedCultures").Get<string[]>();
+            _defaultCulture = configurationSection["DefaultCulture"] ?? "";
+            _supportedCultures = configurationSection.GetSection("SupportedCultures").Get<List<string>>().ToArray();
```
hishamco: Why?

src/OrchardCore.Modules/OrchardCore.Setup/Views/_Layout.cshtml
```diff
@@ -21,5 +58,16 @@
...
         </footer>
     </div>
+    <script>
+        function setLocalizationCookie() {
+            var culturesList = document.getElementById('culturesList');
+            var culture = culturesList.options[culturesList.selectedIndex].value
+            var queryString = window.location.search;
+            var searchParams = new URLSearchParams(queryString);
+
+            searchParams.set("ui-culture", culture);
```
hishamco: @jtkech the queryString becomes much simpler with `URLSearchParams` ;)

Thanks

src/OrchardCore.Modules/OrchardCore.Setup/Views/_Layout.cshtml
```diff
@@ -1,4 +1,18 @@
...
+@using Microsoft.AspNetCore.Localization
+@using Microsoft.Extensions.Options
+@using System.Globalization; 
+@inject Microsoft.AspNetCore.Mvc.Localization.IViewLocalizer T
 @inject OrchardCore.IOrchardHelper Orchard
+@inject IOptions<RequestLocalizationOptions> LocOptions
+
+@{
+    var requestCulture = Context.Features.Get<IRequestCultureFeature>();
+
```
hishamco: Extra line

src/OrchardCore.Modules/OrchardCore.Setup/Views/_Layout.cshtml
```diff
@@ -1,4 +1,18 @@
...
+    var currentCulture = CultureInfo.GetCultureInfo(Orchard.HttpContext.Request.Query.ContainsKey("ui-culture")
```
hishamco: @Skrypt I'm not sure why we need the check again, while the `LocOptions.Value.DefaultRequestCulture.Culture.Name` will pick up the current culture automatically from `QueryString`

Skrypt: No it's not. It takes the default culture that we set in the startup.cs file.

hishamco: Sorry, it's my bad I mean `requestCulture`

Skrypt: Ok you are right I removed this double check

src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
Skrypt: Here one thing I noticed is the fact that we can't add an invariant culture as a supported culture. I think it is not necessary though.

sebastienros:
There is something I don't understand. The culture providers are defined as I see it's used directly. But why read the Request one, and not just let the middleware update the CurrentCulture. That would also take it from the cookie automatically.

Skrypt:
We could probably just get it from the Thread.CurrentThread.CurrentUICulture too.

Skrypt:
Just tested and using Thread.CurrentThread.CurrentUICulture works too.

src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
hishamco: Again I'm not sure why we need to call `ToArray()` while it was an array before you convert it to `List <string>`

Skrypt: If you set the supported culture like this with an empty string for InvariantCulture : 

```
    "OrchardCore.Setup": {
      "DefaultCulture": "en",
      "SupportedCultures": [ "", "en", "fr" ]
    }
```

And you do this : 

`_supportedCultures = configurationSection.GetSection("SupportedCultures").Get<List<string>>()` 

The array will contain a value that is converted to "null". 
{ null, en, fr }

You can't pass a null value to SupportedCulture as it will fail with an exception.

`_supportedCultures = configurationSection.GetSection("SupportedCultures").Get<List<string>>()` Will remove the string.empty value from the configuration and return only the 2 values that are not null.

Skrypt:
Moved back to use .Name instead of TwoLetterISOLanguageName as it is more accurate. If someone wants to add "fr-CA" as a setup culture he can do as it will fallback to "fr" naturally if he doesn't provide any .po file.

src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
```diff
@@ -23,7 +23,12 @@ public Startup(IShellConfiguration shellConfiguration)
             var configurationSection = shellConfiguration.GetSection("OrchardCore.Setup");
 
             _defaultCulture = configurationSection["DefaultCulture"] ?? "";
-            _supportedCultures = configurationSection.GetSection("SupportedCultures").Get<List<string>>().ToArray();
+            _supportedCultures = configurationSection.GetSection("SupportedCultures").Get<string[]>();
+            var results = configurationSection.GetSection("SupportedCultures").Get<List<string>>();
```
hishamco: Can I know what are you trying to do? If you want to add the `InvariantCulture` we can do it in the registration of the service

Skrypt: Is it not what we do here?

hishamco: So what's you intend to do here

Skrypt: We take the list of cultures in the appsettings.json file and we register those cultures in the service registration in the configure() method. I don't understand your comment. I explained yesterday why I was juggling with the `.Get<string[]>` and `.Get<List<string>>` that's because of the binding of the `.Get()` method. So I'm looking if someone wants to pass a string.empty in the SupportedCulture section of the appsetting.json file. If it's found then we add the string.empty value to the List<string> so that we support the InvariantCulture that is associated with it.

Skrypt: .Get<string[]> will return an array containing { null, "fr", "en" } if we set a "" value in the appsettings.json file... that how the .Get() method binds data. We can't change that.

Skrypt: https://github.com/aspnet/Configuration/issues/864

Skrypt: Here an other one : https://github.com/aspnet/Configuration/issues/788

Skrypt: Ok if I put { null, "fr", "en" } in the appsettings.json file it binds it to null when we do the `.Get<string[]>`  so it's not a possible workaround.

sebastienros:
After the triage we decided to:
- Remove all the changes in Startup, the current code will switch to the server's system culture as the default culture
- Add all supported cultures to the Setup supported cultures list
- Use System.Threading.Thread.CurrentThread.CurrentUICulture in the view, or `@Orchard.CultureName`, and not resolve the feature.
- Rename the JavaScript function

Jasmin is applying these changes

Skrypt:
image

Here is one thing though if we use the DisplayName it will render it in the System OS culture.

src/OrchardCore.Modules/OrchardCore.Setup/Views/_Layout.cshtml
```diff
@@ -21,5 +42,16 @@
...
     </div>
+    <script>
+        function setLocalizationUrl() {
+            var culturesList = document.getElementById('culturesList');
+            var culture = culturesList.options[culturesList.selectedIndex].value
+            var queryString = window.location.search;
+            var searchParams = new URLSearchParams(queryString);
+
+            searchParams.set("ui-culture", culture);
+            window.location = window.location.origin + '?' + searchParams.toString();
```
hishamco: `window.location.origin` breaks the tenant setup, I will handle this soon

hishamco:
@agriffard final review please, if it's good then you can merge

src/OrchardCore.Cms.Web/appsettings.json
```diff
@@ -8,8 +8,8 @@
   },
   "OrchardCore": {
     "OrchardCore.Setup": {
-      "DefaultCulture": "",
-      "SupportedCultures": [ "en" ]
+      "DefaultCulture": "", // When using "" the system OS culture will be used
+      "SupportedCultures": [ "ar", "cs", "da", "de", "el", "en", "es", "fa", "fi", "fr", "he", "hr", "hu", "id", "it", "ja", "ko", "lt", "mk", "nl", "pl", "pt", "ru", "sk", "sl", "sv", "tr", "uk", "vi", "zh-CN", "zh-TW" ]
     }
```
deanmarcussen: I think the intention here is to provide a list of all the supported cultures.
While this works if someone is building and running from the source code, it doesn't provide any defaults for anyone using NuGet packages, because this `appsettings.json` file is not copied as part of the templates.

With media we use provide default settings in the source code if nothing is configured in `appsettings.json`, and if it is configured in `appsettings.json` that will override the defaults.

Maybe I am wrong but it seems like it would make more sense to do the same here, otherwise this configuration we have set here is useless for most users.

src/OrchardCore.Modules/OrchardCore.Setup/Views/_Layout.cshtml
```diff
@@ -21,5 +42,16 @@
...
+    <script>
+        function setLocalizationUrl() {
+            var culturesList = document.getElementById('culturesList');
+            var culture = culturesList.options[culturesList.selectedIndex].value
+            var queryString = window.location.search;
+            var searchParams = new URLSearchParams(queryString);
+
+            searchParams.set("ui-culture", culture);
+            window.location = window.location.href.split('?')[0] + '?' + searchParams.toString();
+        }
```
deanmarcussen: the tenant url (if present) might be better retrieved from the `HttpContext.PathBase`

src/OrchardCore.Cms.Web/appsettings.json
Skrypt: I think we want to support the culture that we have in Crowdin. So it makes sense to have it there for that matter. If someone wants a different culture than what we support in Crowdin then he can add it in it's own appsettings.json file set in his default tenant for example. And we could add this appsettings.json file in the web app template.

src/OrchardCore.Modules/OrchardCore.Setup/Views/_Layout.cshtml
```diff
@@ -21,5 +42,16 @@
...
     </div>
+    <script>
+        function setLocalizationUrl() {
+            var culturesList = document.getElementById('culturesList');
+            var culture = culturesList.options[culturesList.selectedIndex].value
+            var queryString = window.location.search;
+            var searchParams = new URLSearchParams(queryString);
+
+            searchParams.set("ui-culture", culture);
+            window.location = window.location.origin + '@(!String.IsNullOrWhiteSpace(Context.Request.PathBase.Value) ? Context.Request.PathBase.Value : "")' + '?' + searchParams.toString();
```
jtkech: For infos `PathSting.HasValue` (here `PathBase.Value`) does the same so that's okay

        /// <summary>
        /// True if the path is not empty
        /// </summary>
        public bool HasValue
        {
            get { return !string.IsNullOrEmpty(_value); }
        }

There are some places where we fully encode all uri components, but yes there are some places where we know or want to work on an unencoded url string and where we use `Request.PathBase.Value` e.g to prevent double encoding. But here we are updating `window.location.origin` directly, that maybe needs to be correctly escaped.

If so we would need to use `PathBase.ToUriComponent()` or `PathBase.ToString()` that does the same

        /// <summary>
        /// Provides the path string escaped in a way which is correct for combining into the URI representation.
        /// </summary>
        /// <returns>The escaped path value</returns>
        public override string ToString()
        {
            return ToUriComponent();
        }

src/OrchardCore.Cms.Web/appsettings.json
deanmarcussen: Thanks @Skrypt updating the templates will work well for new projects.

I would still say that for most existing users they will not see these support cultures without knowing to update their `appsettings.json` in their existing project.

Would it make sense for this to be a `DefaultSupportedCultures` property in 
https://github.com/OrchardCMS/OrchardCore/blob/b3bb908c9abd414b528626f9135fe4efb3d70c9d/src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs#L25

Similar to what is done with Media/AllowedFileExtensions
https://github.com/OrchardCMS/OrchardCore/blob/6cddd9f9904ff117ba82587f39fc76f49fac3cc3/src/OrchardCore.Modules/OrchardCore.Media/Services/MediaOptionsConfiguration.cs#L70

Then as more cultures are available via crowdin we just update the default and they show for everyone?

Maybe is not the intention though?

src/OrchardCore.Modules/OrchardCore.Setup/Views/_Layout.cshtml
```diff
@@ -21,5 +46,16 @@
...
     </div>
+    <script>
+        function setLocalizationUrl() {
+            var culturesList = document.getElementById('culturesList');
+            var culture = culturesList.options[culturesList.selectedIndex].value
+            var queryString = window.location.search;
+            var searchParams = new URLSearchParams(queryString);
+
+            searchParams.set("ui-culture", culture);
+            window.location = '@(fullUri)' + '?' + searchParams.toString();
```
sebastienros: I think the relative url could be created directly in the html as a `data` attribute. Then this even would just read it. And I assume that should make creating the url easier too, even tenant aware, as we'd just need to change the querystring of the url, and there are helpers for this in dotnet.

src/OrchardCore.Modules/OrchardCore.Setup/Views/_Layout.cshtml
```diff
@@ -1,12 +1,13 @@
...
 @inject Microsoft.AspNetCore.Mvc.Localization.IViewLocalizer T
 @inject OrchardCore.IOrchardHelper Orchard
 @inject IOptions<RequestLocalizationOptions> LocOptions
 @{
-    var fullUri = Context.Request.Scheme + "://" + Context.Request.Host.ToUriComponent();
-    fullUri += Context.Request.PathBase.HasValue ? Context.Request.PathBase.ToUriComponent() : "";
-    fullUri += Context.Request.Path.ToUriComponent();
+    Uri currentUrl = new Uri(Context.Request.GetEncodedUrl());
+    var urlWithTenant = String.Format("{0}{1}{2}{3}", Context.Request.Scheme,
+                    Uri.SchemeDelimiter, currentUrl.Authority, currentUrl.AbsolutePath);
```
jtkech: @Skrypt 

I think you just need to set an uri string with `var uri = Context.Request.GetEncodedUrl()`

It already takes into account the `Scheme` and the `Host` ans so on, [see here](https://github.com/dotnet/aspnetcore/blob/master/src/Http/Http.Extensions/src/UriHelper.cs#L177)

Skrypt: Yes but the javascript changes the ui-culture=culturecode part of the url and Context.Request.GetEncodedUrl() returns the entire thing. So I just need to remove the querystring part so that it gets appended through javascript.

sebastienros:
I simplified url building and javascript (I think)

Skrypt:
oh well nice merge it please I'm tired. 😄

src/OrchardCore.Modules/OrchardCore.Setup/Views/_Layout.cshtml
```diff
@@ -48,12 +49,7 @@
     <script>
         function setLocalizationUrl() {
             var culturesList = document.getElementById('culturesList');
-            var culture = culturesList.options[culturesList.selectedIndex].value
-            var queryString = window.location.search;
-            var searchParams = new URLSearchParams(queryString);
-
-            searchParams.set("ui-culture", culture);
-            window.location = culturesList.dataset.url + '?' + searchParams.toString();
+            window.location = culturesList.options[culturesList.selectedIndex].dataset.url;
```
jtkech: I thought that window loction would need to be set with an absolute uri, not a relative url

But didn't work on javascript for a while ;)

src/OrchardCore.Modules/OrchardCore.Setup/Views/_Layout.cshtml
```diff
@@ -27,10 +26,12 @@
                 {
                     <div class="float-right form-inline">
                         <label for="culturesList" class="mr-2">@T["Change setup culture: "]</label>
-                        <select id="culturesList" class="form-control" data-url="@uriString" onchange="setLocalizationUrl()">
+                        <select id="culturesList" class="form-control" onchange="setLocalizationUrl()">
                             @foreach (var culture in LocOptions.Value.SupportedUICultures.OrderBy(x => x.NativeName))
                             {
-                                <option value="@culture" selected="@(culture.Name == Orchard.CultureName())">@(culture.NativeName ?? culture.Name)</option>
+                                var uriString = UriHelper.BuildRelative(Context.Request.PathBase, Context.Request.Path, new QueryString().Add("ui-culture", culture.Name));
```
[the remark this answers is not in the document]
jtkech: @sebastienros 

Just tried, it fails when changing the culture if it is not the default tenant, we lose the token parameter of the current query string

src/OrchardCore.Modules/OrchardCore.Setup/Views/_Layout.cshtml
jtkech: Okay just tried, it works with the relative url, the only problem, as said above, is that we lose the token parameter of the current query string.

hishamco: I'm not sure what's the wrong with `URLSearchParams` that I did earlier?!!

Skrypt:
@hishamco The serbian culture that you added back don't have a corresponding folder that match them so they won't be loaded. If we do that then @agriffard need to fix Crowdin to get them to match.

hishamco:
I'm sure they already added to Crowdin after someone in the community asking for it

hishamco:
FYI https://github.com/OrchardCMS/OrchardCore.Translations/issues/39

Skrypt:
Ok then if that will eventually work.

hishamco:
Exactly, seems you added the old cultures before

Skrypt:
Is there any more things to add or change?

jtkech:
> The CI doesn't build

It is not specific to this PR

> Is there any more things to add or change?

I suugest to do something like this in the startup ctor but not sure of the value to use

    _defaultCulture = configurationSection["DefaultCulture"] ?? CultureInfo.InstalledUICulture.Name;

So that it is added to the selectable cultures even no default culture is defined through the configuration

Skrypt:
> _defaultCulture = configurationSection["DefaultCulture"] ?? CultureInfo.InstalledUICulture.Name;

Not necessary because if you pass an empty string like it is right now the configurationSection databinding will return null and thus will not set anything. Then if you set nothing on the defaultCulture of the middleware it will take the CultureInfo.InstalledUICulture.

src/Templates/OrchardCore.ProjectTemplates/content/OrchardCore.Templates.Cms.Web/appsettings.json
```diff
@@ -36,6 +36,44 @@
         }
       }
     ]
-  }
+  },
   //#endif
+  "OrchardCore": {
```
sebastienros: I think this should be reverted since it's now part of the code as defaults.

hishamco: I think it's useful if we run the source, while the defaults are useful if we run out of the source

Correct me if I'm wrong?

jtkech: We always run partially from the source, the application itself ;) What @sebastienros says is that it is not necessary when building an application by using the `Templates` so that we don't have to maintain 2 `appsettings.json`, and then people can still update it as they want.

sebastienros: We changed this file when we added the cultures to the appsettings. Then we removed the changes in appsettings so we should also remove it there.

jtkech: Oups i missed that it has also been removed under the web app ;)

jtkech: Just checked, `SupportedCultures` are still defined in the appsettings of the web app.

PR #5746, opened 2020-03-17
Fix IEmailAddressValidator regression bug

### LINKED ISSUE #5745: Exception registering external user
Hi,

I'm getting an exception when registering an external user.

Exception:
```
InvalidOperationException: Could not create an instance of type 'OrchardCore.Users.ViewModels.RegisterExternalLoginViewModel'. Model bound complex types must not be abstract or value types and must have a parameterless constructor. Alternatively, give the 'model' parameter a non-null default value.
Microsoft.AspNetCore.Mvc.ModelBinding.Binders.ComplexTypeModelBinder.CreateModel(ModelBindingContext bindingContext)
```

Repro:

1. go to login page
2. choose the external IDP
3. login with a user that is not registered yet
4. on returning to OC (/OrchardCore.Users/Account/ExternalLoginCallback) set the user name and click Register

Logfile: [orchard-log-2020-03-17.log](https://github.com/OrchardCMS/OrchardCore/files/4344206/orchard-log-2020-03-17.log)

Arjan

hishamco:
Fixes #5745

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Users/Controllers/AccountController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Controllers/AccountController.cs
@@ -69,7 +69,6 @@ namespace OrchardCore.Users.Controllers
             _distributedCache = distributedCache;
             _dataProtectionProvider = dataProtectionProvider;
             _externalLoginHandlers = externalLoginHandlers;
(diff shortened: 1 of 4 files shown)


hishamco:
Unfortunately seems some changes doesn't saved since the first commit

PR #5749, opened 2020-03-18
Update package.json dependencies versions and fix vulnerabilities

agriffard:
Run `ncu -u` and `npm audit fix` tu upgrade the packages to latest version and fix some security vulnerabilities

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,29 +12,29 @@
       }
     },
     "@babel/compat-data": {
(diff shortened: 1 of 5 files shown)


agriffard:
@jptissot There is an error during AppVeyor build:
`Jest did not exit one second after the test run has completed.`

Would you mind checking if it is caused by the update of the Jest package?

PR #5765, opened 2020-03-20
Update ContentLinkTagHelper for general actions

flew2bits:
Added option to ContentLinkTagHelper to be called by liquid without any of the following options: edit_for, create_for, remove_for, display_for, or admin_for. Currently the {% a ... %} option can't be used in Liquid pages without one of the content item options. However, the "a" tag helper supports options for "asp-route-" attributes, and these are unusable from Liquid without this update.


src/OrchardCore/OrchardCore.Contents.TagHelpers/ContentLinkTagHelper.cs
```diff
@@ -15,6 +15,7 @@ namespace OrchardCore.Contents.TagHelpers
     [HtmlTargetElement("a", Attributes = ContentLinkEdit)]
     [HtmlTargetElement("a", Attributes = ContentLinkRemove)]
     [HtmlTargetElement("a", Attributes = ContentLinkCreate)]
+    [HtmlTargetElement("a")]
```
ns8482e: No need of ContentLinkTagHelper to Target `<a>`

Use asp.net tag helper instead

flew2bits:
You're correct. I'm fairly new to tag helpers and didn't think about that.

PR #5772, opened 2020-03-22
Update health checks docs for core 3.1

deanmarcussen:
Version agnostic link, will redirect to the latest (?view=aspnetcore-3.1)


hishamco:
Is there an yproblem with Travis, I saw couple of issues failing on Travis with unknown reason

PR #5785, opened 2020-03-23
Lockout on wrong password

agriffard:
Try to use LockoutOptions
https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.identity.lockoutoptions?view=aspnetcore-3.1
https://code-maze.com/user-lockout-aspnet-core-identity/

Related to #2936


MatthijsKrempel:
As an alternative, you can activate ReCaptcha features. Instead of locking out users, we slow down the attack on the account.

hishamco:
> As an alternative, you can activate ReCaptcha features

This is another option, but Lockout is essential if I'm not wrong

src/OrchardCore.Modules/OrchardCore.Users/Startup.cs
```diff
@@ -118,7 +118,12 @@ public override void ConfigureServices(IServiceCollection services)
 
             // Adds the default token providers used to generate tokens for reset passwords, change email
             // and change telephone number operations, and for two factor authentication token generation.
-            services.AddIdentity<IUser, IRole>().AddDefaultTokenProviders();
+            services.AddIdentity<IUser, IRole>(opt =>
+            {
+                opt.Lockout.AllowedForNewUsers = true;
+                opt.Lockout.DefaultLockoutTimeSpan = TimeSpan.FromMinutes(2);
+                opt.Lockout.MaxFailedAccessAttempts = 3;
```
hishamco: Why these options can't come from the settings?

agriffard: This is the goal @hishamco .
I just pushed the PR to show what I tried so far.
No need to review right now.

I am working on it with @Skrypt .

Skrypt:
@MatthijsKrempel yes but sometimes you just want to lock the account too. Even more secure. Because Recaptcha is not 100% safe. 😉

src/OrchardCore/OrchardCore.Users.Core/Indexes/UserIndex.cs
```diff
@@ -9,6 +10,9 @@ public class UserIndex : MapIndex
         public string NormalizedUserName { get; set; }
         public string NormalizedEmail { get; set; }
         public bool IsEnabled { get; set; }
+        public bool LockoutEnabled { get; set; }
+        public DateTimeOffset? LockoutEnd { get; set; }
```
Skrypt: I'm not sure, but I think we can't use DateTimeOffset with SQLite. Unless we find a fix for YesSQL. Normally a DateTime should be fine enough else we will need to save this as a string timestamp.

Skrypt:
I think that we should also add an option to edit the LockoutEnabled param on each user. By default on the recipe it makes the Admin user to have this param set to false so that the admin user can't be locked out. Would it not be a security issue if we allow to brute force on this user account? Or maybe I did not understand properly the usage of LockoutEnabled.

src/OrchardCore/OrchardCore.Users.Core/Models/User.cs
```diff
@@ -17,6 +17,9 @@ public class User : Entity, IUser
         public string SecurityStamp { get; set; }
         public bool EmailConfirmed { get; set; }
         public bool IsEnabled { get; set; } = true;
+        public bool LockoutEnabled { get; set; }
```
sebastienros, line 20:
```suggestion
        public bool LockedOut { get; set; }
```

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
```diff
@@ -664,5 +665,87 @@ private static UserToken GetUserToken(IUser user, string loginProvider, string n
...
+        }
+
+        public Task SetLockoutEndDateAsync(IUser user, DateTimeOffset? lockoutEnd, CancellationToken cancellationToken)
+        {
+            if (user == null)
+            {
+                throw new ArgumentNullException(nameof(user));
+            }
+
+            ((User)user).LockoutEnd = lockoutEnd;
```
sebastienros, line 744:
```suggestion
            ((User)user).LockoutEnd = lockoutEnd.ToUniversalDateTime();
```

src/OrchardCore/OrchardCore.Users.Core/Indexes/UserIndex.cs
```diff
@@ -10,6 +11,9 @@ public class UserIndex : MapIndex
...
```
sebastienros, line 15:
```suggestion
        public DateTime? LockoutEndUtc { get; set; }
```

src/OrchardCore.Cms.Web/appsettings.json
```diff
@@ -1,4 +1,19 @@
...
+    //"Password": {
+    //  "RequireDigit": true,
+    //  "RequireLowercase": true,
+    //  "RequireUppercase": true,
+    //  "RequireNonAlphanumeric": true,
+    //  "RequiredUniqueChars": 3,
+    //  "RequiredLength": 8
+    //},
+    "Lockout": {
+      "AllowedForNewUsers ": true,
```
sebastienros, line 12: Remove spaces in string

sebastienros: I think it should be disabled by default.

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
agriffard: How can I convert a DateTimeOffset to a DateTime then?

jptissot:
@agriffard can we revive this PR?

agriffard:
Conflict solved and latest version merged.

src/OrchardCore.Cms.Web/appsettings.json
```diff
@@ -1,4 +1,11 @@
 {
+  "IdentityOptions": {
```
deanmarcussen, line 2: This isn't bound by configuration, so does nothing.
It would need to be in at least the `OrchardCore` section, and bound using `service.ConfigureOptions` in the users project to work

src/OrchardCore.Modules/OrchardCore.Users/Migrations.cs
```diff
@@ -189,5 +190,22 @@ public int UpdateFrom9()
...
+                .AddColumn<bool>(nameof(UserIndex.IsLockedOut), c => c.NotNull().WithDefault(false)));
+
+            SchemaBuilder.AlterIndexTable<UserIndex>(table => table
+                .AddColumn<DateTimeOffset?>(nameof(UserIndex.LockoutEnd), c => c.Nullable()));
+
+            SchemaBuilder.AlterIndexTable<UserIndex>(table => table
+                .AddColumn<int>(nameof(UserIndex.AccessFailedCount), c => c.NotNull().WithDefault(0)));
+
+            SchemaBuilder.AlterIndexTable<UserIndex>(table => table
+                .CreateIndex("IDX_UserIndex_IsLockedOut", "IsLockedOut"));
```
deanmarcussen, line 206: This all needs to be in `Create` as well, and adjust create to skip this update.
The indexed column needs to be added to one of the existing indexes to be of some use ;)

src/OrchardCore.Cms.Web/appsettings.json
agriffard: https://docs.microsoft.com/en-us/aspnet/core/security/authentication/identity-configuration?view=aspnetcore-5.0#lockout

agriffard:
About `DateTimeOffset`: When you inherit from `IUserLockoutStore`, you have to implement `SetLockoutEndDateAsync`:
`Task SetLockoutEndDateAsync(TUser user, DateTimeOffset? lockoutEnd, CancellationToken cancellationToken);`

https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.identity.iuserlockoutstore-1?view=aspnetcore-5.0#methods

src/OrchardCore.Modules/OrchardCore.Users/Migrations.cs
agriffard: Are my latest changes in Migrations correct?

jtkech: Even you add these columns to an existing index table, better to also add them when defining this index table in `Create()`, so that on new install this index table is created in one pass. Then in `Create()` return 10 + 1 = 11.

To be okay with all providers, we should not use `DateTimeOffset` in the database, use `DateTime` instead

In the `OpenId` look at the `OpenIdAuthorizationIndexProvider` with a `DateTime` in the index / provider, and a `DateTimeOffset` in the `OpenIdAuthorizationStore` API, where we use the UTC part of the `DateTimeOffset`.

agriffard: @jtkech The version you are looking at might be outdated, I already did the changes you mention.

agriffard: @jtkech I can't use this syntax:
image
instead of:
`return Task.FromResult(new DateTimeOffset?(((User)user).LockoutEndUtc.Value));`
if it is what you mean.

agriffard: ok, @jtkech done.

src/OrchardCore.Cms.Web/appsettings.json
deanmarcussen: The above would roughly work.

But because we don't bind any other identity options automatically from `appsettings.json` I think it would be easier, and better if we didn't provide automatic binding.

You have already written docs on how to do the binding https://docs.orchardcore.net/en/dev/docs/guides/password-configuration/ 

Why not just add it more docs there? and remove it from the `appsettings.json`

agriffard: Documentation added.

src/OrchardCore/OrchardCore.Users.Core/Indexes/UserIndex.cs
```diff
@@ -10,6 +11,9 @@ public class UserIndex : MapIndex
         public string NormalizedUserName { get; set; }
         public string NormalizedEmail { get; set; }
         public bool IsEnabled { get; set; }
+        public bool IsLockedOut { get; set; }
+        public DateTime? LockoutEnd { get; set; }
```
sebastienros, line 15:
```suggestion
        public DateTime? LockoutEndUtc { get; set; }
```

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
```diff
@@ -664,5 +665,101 @@ private static UserToken GetUserToken(IUser user, string loginProvider, string n
...
+        public Task<DateTimeOffset?> GetLockoutEndDateAsync(IUser user, CancellationToken cancellationToken)
+        {
+            if (user == null)
+            {
+                throw new ArgumentNullException(nameof(user));
+            }
+
+            if (((User)user).LockoutEndUtc.HasValue)
+            {
+                return Task.FromResult<DateTimeOffset?>(DateTime.SpecifyKind(((User)user).LockoutEndUtc.Value, DateTimeKind.Utc));
```
sebastienros, line 700:
```suggestion
                return Task.FromResult<DateTimeOffset?>(((User)user).LockoutEndUtc.Value.ToUniversalTime());
```

src/docs/guides/lockout-configuration/README.md
```diff
@@ -0,0 +1,56 @@
...
+      "DefaultLockoutTimeSpan ": "00:05:00",
+      "MaxFailedAccessAttempts ": 5
+    }
+  }
+}
+```
+
+## Summary
+
+You just learnt how to configure the Lockout settings from an `appsettings.json` file.
```
sebastienros, line 56:
```suggestion
You just learned how to configure the Lockout settings from an `appsettings.json` file.
```

deanmarcussen:
I looked at the options docs, is there actually a way to disable it from those options?

Or is it on permanently?

Do we need to be able to disable it? I'm not sure, can it be used as a vector to lock the admin out?

src/OrchardCore.Modules/OrchardCore.Users/Controllers/AdminController.cs
```diff
@@ -506,5 +506,42 @@ public async Task<IActionResult> EditPassword(ResetPasswordViewModel model)
...
+
+            var user = await _userManager.FindByIdAsync(id) as User;
+
+            if (user == null)
+            {
+                return NotFound();
+            }
+
+            await _userManager.ResetAccessFailedCountAsync(user);
+            var result = await _userManager.SetLockoutEnabledAsync(user, false);
```
sebastienros, line 526: This should not disable/enable lock out, but clear the `LockedOutUtc` of this user.

src/OrchardCore.Modules/OrchardCore.Users/Controllers/AdminController.cs
```diff
@@ -506,5 +506,42 @@ public async Task<IActionResult> EditPassword(ResetPasswordViewModel model)
 
             return View(model);
         }
+
+        [HttpPost]
+        public async Task<IActionResult> Unlock(string id)
```
sebastienros, line 511: And maybe add `DisableLockOut` and `EnableLockOut` to change the feature for each user.

src/OrchardCore.Modules/OrchardCore.Users/Migrations.cs
```diff
@@ -23,6 +24,9 @@ public int Create()
                 .Column<string>("NormalizedUserName") // TODO These should have defaults. on SQL Server they will fall at 255. Exceptions are currently thrown if you go over that.
                 .Column<string>("NormalizedEmail")
                 .Column<bool>("IsEnabled", c => c.NotNull().WithDefault(true))
+                .Column<bool>("IsLockedOut", c => c.NotNull().WithDefault(false))
```
sebastienros, line 27: Also add `.Column<bool>("IsLockOutEnabled", c => c.NotNull().WithDefault(false))

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
```diff
@@ -664,5 +665,101 @@ private static UserToken GetUserToken(IUser user, string loginProvider, string n
...
+                throw new ArgumentNullException(nameof(user));
+            }
+
+            if (((User)user).LockoutEndUtc.HasValue)
+            {
+                return Task.FromResult<DateTimeOffset?>(((User)user).LockoutEndUtc.Value.ToUniversalTime());
+            }
+            else
+            {
+                return Task.FromResult(new DateTimeOffset?());
```
sebastienros, line 704: Try returning `null`

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
```diff
@@ -664,5 +665,101 @@ private static UserToken GetUserToken(IUser user, string loginProvider, string n
...
+        }
+
+        public Task SetLockoutEnabledAsync(IUser user, bool enabled, CancellationToken cancellationToken)
+        {
+            if (user == null)
+            {
+                throw new ArgumentNullException(nameof(user));
+            }
+
+            ((User)user).IsLockedOut = enabled;
```
sebastienros, line 739:
```suggestion
            ((User)user).IsLockOutEnabled = enabled;
```

src/OrchardCore.Modules/OrchardCore.Users/Controllers/AdminController.cs
agriffard: The `Unlock` button in the Admin is only displayed when the user is locked (the number of failed access has been reached).

src/OrchardCore.Modules/OrchardCore.Users/Controllers/AdminController.cs
agriffard: Added `SetLockoutEndDateAsync(user, null);`, Kept `SetLockoutEnabledAsync`

src/OrchardCore.Modules/OrchardCore.Users/Controllers/AdminController.cs
sebastienros: Yes, and this is why calling this should "unlock" the user, by setting the LockedEndUtc date to null.

src/OrchardCore.Modules/OrchardCore.Users/Controllers/AdminController.cs
sebastienros: This permanently disables lock out for this user, while it should just unlock it.

src/OrchardCore.Modules/OrchardCore.Users/Controllers/AdminController.cs
```diff
@@ -506,5 +506,43 @@ public async Task<IActionResult> EditPassword(ResetPasswordViewModel model)
...
+            }
+            else
+            {
+                await _session.CancelAsync();
+
+                _notifier.Error(H["Could not unlock the user."]);
+
+                foreach (var error in result.Errors)
+                {
+                    _notifier.Error(H[error.Description]);
```
sebastienros, line 541: Are these message meant to be translated dynamically?

src/OrchardCore.Modules/OrchardCore.Users/Controllers/AdminController.cs
```diff
@@ -506,5 +506,43 @@ public async Task<IActionResult> EditPassword(ResetPasswordViewModel model)
...
+            {
+                return NotFound();
+            }
+
+            await _userManager.ResetAccessFailedCountAsync(user);
+            var result = await _userManager.SetLockoutEnabledAsync(user, false);
+
+            if (result.Succeeded)
+            {
+                await _userManager.SetLockoutEndDateAsync(user, null);
```
sebastienros, line 530: Should it also call `ResetAccessFailedCountAsync(0)` ?

agriffard: `ResetAccessFailedCountAsync` is called 5 lines above.

src/OrchardCore.Modules/OrchardCore.Users/Views/UserButtons.cshtml
```diff
@@ -16,6 +20,10 @@
         <a class="btn btn-danger btn-sm" asp-action="Delete" asp-route-id="@user.UserId" data-url-af="RemoveUrl UnsafeUrl" >@T["Delete"]</a>
     }
     <a asp-action="EditPassword" asp-route-id="@user.UserId" class="btn btn-secondary btn-sm">@T["Edit Password"]</a>
+    if (isLockedOut)
+    {
+        <a asp-action="Unlock" asp-route-id="@user.Id" class="btn btn-danger btn-sm" data-url-af="RemoveUrl UnsafeUrl" data-title="@T["Unlock user"]" data-message="@T["Are you sure you want to unlock this user?"]">@T["Unlock"]</a>
+    }
```
deanmarcussen, line 26: I think this view (and the driver) is the one that needs the editor (checkbox) for `IsLockoutEnabled`

It should be wrapped in a `if (hasManageUser)` check so that a user editing their own profile can't update that value.

src/OrchardCore.Modules/OrchardCore.Users/Views/UserButtons.cshtml
```diff
@@ -16,6 +20,10 @@
         <a class="btn btn-danger btn-sm" asp-action="Delete" asp-route-id="@user.UserId" data-url-af="RemoveUrl UnsafeUrl" >@T["Delete"]</a>
     }
     <a asp-action="EditPassword" asp-route-id="@user.UserId" class="btn btn-secondary btn-sm">@T["Edit Password"]</a>
+    if (isLockedOut)
```
sebastienros, line 23:
```suggestion
    if (isLockedOut && hasManageUser)
```

src/OrchardCore.Modules/OrchardCore.Users/Controllers/AdminController.cs
sebastienros: Logger.Error(error.Description)

sebastienros (approved):

src/OrchardCore.Modules/OrchardCore.Users/Views/UserButtons.cshtml
agriffard: The code is already inside a `@if (hasManageUser)`

Issue #5712, opened 2020-03-12
Test Chrome samesite changes


sebastienros:

hishamco:
@MatthijsKrempel could please have a look?

MatthijsKrempel:
Sure!

MatthijsKrempel:
Tested it with igocnito mode, seems to login just fine.
So what seems to be what Chrome considers "safe mode".
If that is not the issue, I need more details.

sebastienros:
Did you switch the chrome feature on before?

MatthijsKrempel:
@sebastienros got a link I can follow, because I feel like I am chasing a rabbit.

sebastienros:
Sorry, I should have said "SameSite":

https://www.chromium.org/updates/same-site/test-debug

deanmarcussen:
@MatthijsKrempel am receiving this console warning on a non http site during login which fails (latest chrome / mac will get the version tomorrow)

A cookie associated with a resource at ... was set withSameSite=Nonebut withoutSecure. It has been blocked, as Chrome now only delivers cookies markedSameSite=Noneif they are also markedSecure. You can review cookies in developer tools under Application>Storage>Cookies and see more details at https://www.chromestatus.com/feature/5633521622188032.

MatthijsKrempel:
![image](https://user-images.githubusercontent.com/1598506/76893177-3610c400-688c-11ea-8f05-f9f93d29307b.png)
The login cookie is secure, so need a bit more info on your situation.

deanmarcussen:
try with non http. then it is not secure :)

MatthijsKrempel:
I reproduced it when creating a non https site :)

Thanks @deanmarcussen , will have a look.

MatthijsKrempel:
@deanmarcussen I think the only option is to move to HTTPS if you want to be able to log in to your site. It seems google is actively pushing towards this goal.

The options we have is to set the samesite mode to strict, but if your site was ever hosted on https the new cookie will not be overwritten by chrome. Since it values secure cookies over non secure ones.
Then the difficulty would be that we need to set it to samesite none when activating the OpenId module, to support the prompt=none.

Since we have secure cookies during development time, it is ultimately a decision if we want to support http production scenarios for Orchard Core sites.

deanmarcussen:
It is indeed a conundrum...

Chrome is of course not the only browser in the world ;)

[...]

PR #5786, opened 2020-03-23
Moved SameSite.None to OpenId Module at startup

MatthijsKrempel:
PR for https://github.com/OrchardCMS/OrchardCore/issues/5712

Since we need SameSite cookie behavior set to None to support the OpenId prompt feature, the default setting should be Strict and None when OpenId Server feature is turned on.

I've added a critical log statement that you definitely must run your OpenId server on Https, otherwise Chrome will reject the cookie.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.OpenId/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.OpenId/Startup.cs
@@ -82,7 +82,9 @@ namespace OrchardCore.OpenId
                     // protected by antiforgery checks, that are enforced with or without this setting being changed.
                     // 2020-03-23; Moved the SameSiteNode.None here, this will require that the site runs on HTTPS;
                     if (cookieContext.CookieName.StartsWith("orchauth_"))
+                    {
                         cookieContext.CookieOptions.SameSite = SameSiteMode.None;
+                    }
                 };
             });
         }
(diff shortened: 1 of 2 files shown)


src/OrchardCore.Modules/OrchardCore.OpenId/Startup.cs
```diff
@@ -66,10 +72,29 @@ public override void ConfigureServices(IServiceCollection services)
                 ServiceDescriptor.Scoped<IPermissionProvider, Permissions>(),
                 ServiceDescriptor.Scoped<INavigationProvider, AdminMenu>(),
             });
+
+            services.Configure<CookiePolicyOptions>(options =>
+            {
```
deanmarcussen: you will probably tell me it doesn't work @MatthijsKrempel but was it not possible to just add

```
services.ConfigureApplicationCookie(options => options.Cookie.SameSite = SameSiteMode.None)
```

here to alter the settings all ready configured by users, rather than having to use the cookie middleware to write it?

MatthijsKrempel: @deanmarcussen sounds good, doesnt work.

MatthijsKrempel: What I do need to do though, is move all the backwards compatibility into this module.
As we only need it here.

jptissot:
This pr caused a regression for Azure AD external login provider. #6049

PR #5791, opened 2020-03-24
Add missing tenant permission check in admin

### LINKED ISSUE #5790: A user without the ManageTenants permission can access to the tenants
Repro: Create a user with the Moderator role and go to the admin, you will be able to see the Tenants menu.

agriffard:
Check that the user has the `ManageTenants` permission

Fixes #5790

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
@@ -74,6 +74,11 @@ namespace OrchardCore.Tenants.Controllers
 
         public async Task<IActionResult> Index(TenantIndexOptions options, PagerParameters pagerParameters)
         {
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -74,6 +74,11 @@ public class AdminController : Controller
 
         public async Task<IActionResult> Index(TenantIndexOptions options, PagerParameters pagerParameters)
         {
+            if (!await _authorizationService.AuthorizeAsync(User, Permissions.ManageTenants))
+            {
+                return Forbid();
+            }
+
```
deanmarcussen: can you add this check in as well, (we do it on the other methods) so I'm guessing it's relevant
```
            if (!IsDefaultShell())
            {
                return Forbid();
            }
```

PR #5793, opened 2020-03-24
Add data annotations default error messages

hishamco:
During the lack of Crowdin, because it requires that localization string should have sort of context `msgctxt` me & @agriffard decided to add new class that contains the default data annotations error messages. This will let the `PoExtractor` to extract the localization strings as well as adding a proper context

### DIFF
--- a/src/OrchardCore/OrchardCore.Localization.Core/DataAnnotations/DataAnnotationsDefaultErrorMessages.cs
+++ b/src/OrchardCore/OrchardCore.Localization.Core/DataAnnotations/DataAnnotationsDefaultErrorMessages.cs
@@ -1,0 +1,118 @@
+using Microsoft.Extensions.Localization;
+
+namespace OrchardCore.Localization.Core.DataAnnotations
+{
+    internal class DataAnnotationsDefaultErrorMessages
+    {
+        private readonly IStringLocalizer<DataAnnotationsDefaultErrorMessages> S;
+
+        public DataAnnotationsDefaultErrorMessages(IStringLocalizer<DataAnnotationsDefaultErrorMessages> localizer)
+        {
+            S = localizer;
+        }
+
+        public string AssociatedMetadataTypeTypeDescriptorMetadataTypeContainsUnknownProperties => S["The associated metadata type for type '{0}' contains the following unknown properties or fields: {1}. Please make sure that the names of these members match the names of the properties on the main type."];
+
+        public string AttributeStoreUnknownProperty => S["The type '{0}' does not contain a public property named '{1}'."];
+
+        public string CommonPropertyNotFound => S["The property {0}.{1} could not be found."];
+
+        public string CompareAttributeMustMatch => S["'{0}' and '{1}' do not match."];
+
+        public string CompareAttributeUnknownProperty => S["Could not find a property named {0}."];
+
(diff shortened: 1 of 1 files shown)


hishamco:
Here is a sample of how the data annotation message will look like

```
#: OrchardCore.Localization.Core\DataAnnotations\DataAnnotationsDefaultErrorMessages.cs:84
#. public string RequiredAttributeValidationError => S["The {0} field is required."];
msgctxt "OrchardCore.Localization.Core.DataAnnotations.DataAnnotationsDefaultErrorMessages"
msgid "The {0} field is required."
msgstr ""

#: OrchardCore.Localization.Core\DataAnnotations\DataAnnotationsDefaultErrorMessages.cs:86
#. public string StringLengthAttributeInvalidMaxLength => S["The maximum length must be a nonnegative integer."];
msgctxt "OrchardCore.Localization.Core.DataAnnotations.DataAnnotationsDefaultErrorMessages"
msgid "The maximum length must be a nonnegative integer."
msgstr ""

#: OrchardCore.Localization.Core\DataAnnotations\DataAnnotationsDefaultErrorMessages.cs:88
#. public string StringLengthAttributeValidationError => S["The field {0} must be a string with a maximum length of {1}."];
msgctxt "OrchardCore.Localization.Core.DataAnnotations.DataAnnotationsDefaultErrorMessages"
msgid "The field {0} must be a string with a maximum length of {1}."
msgstr ""

#: OrchardCore.Localization.Core\DataAnnotations\DataAnnotationsDefaultErrorMessages.cs:90
#. public string StringLengthAttributeValidationErrorIncludingMinimum => S["The field {0} must be a string with a minimum length of {2} and a maximum length of {1}."];
msgctxt "OrchardCore.Localization.Core.DataAnnotations.DataAnnotationsDefaultErrorMessages"
msgid "The field {0} must be a string with a minimum length of {2} and a maximum length of {1}."
msgstr ""
```

src/OrchardCore/OrchardCore.Localization.Core/DataAnnotations/DataAnnotationsDefaultErrorMessages.cs
```diff
@@ -0,0 +1,118 @@
...
+{
+    internal class DataAnnotationsDefaultErrorMessages
+    {
+        private readonly IStringLocalizer<DataAnnotationsDefaultErrorMessages> S;
+
+        public DataAnnotationsDefaultErrorMessages(IStringLocalizer<DataAnnotationsDefaultErrorMessages> localizer)
+        {
+            S = localizer;
+        }
+
```
deanmarcussen: Is this class actually used? Or is it just a marker class that the POExtractor will use with reflection to find and add the strings?

If it is just a marker, does it need the constructor? 
Or does the PO extractor instantiate an instance of it?

And could it have a comment explaining that it is a marker, and is not for use?

hishamco: Ya it's sort of marker, that's why I declare it as `internal`, perhaps you are right no need for the constructor, but it may be useful in some use cases when you need to get a specific localized error message

+1 for adding a comment, so let me create a PR and ping you there

Thanks

PR #5816, opened 2020-03-29
Don't pre-create the DataProtection directory

jtkech:
Working on a Redis DataProtection i saw that the related directory is still created on the file system.

Here, if the `FileSystemXmlRepository` is not used the `DataProtection-Keys` directory is not created


Skrypt:
I've not tested but looking good to merge.

Issue #5812, opened 2020-03-28
Console.Log regression


deanmarcussen:
The `{{ Model.Content | console_log }}` filter is no longer serializing the alternates collection as a collection.

The regression is from https://github.com/OrchardCMS/OrchardCore/pull/4933 where the `AlternateCollection` no longer implements `IEnumerable<string>` so the Json Serializer thinks of it as an object, rather than an enumerable.

The result is now just the last alternate, and a count, rather than the list of alternates

@jtkech was there any particular need to remove the `IEnumerable<string>` interface or can I add it back?

jtkech:
You would have also to re-add

        IEnumerator IEnumerable.GetEnumerator()
        {
            if (_collection == null)
            {
                return Array.Empty<string>().GetEnumerator();
            }

            return _collection.GetEnumerator();
        }

I think we did this with @sebastienros to avoid some `Linq` pitfalls, but because our code has been updated accordingly, i'm not against to re-intoduce it if there is no other solution.

PR #5817, opened 2020-03-29
Allow AlternatesCollection to serialize to json array

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/5812

@jtkech I just made the `IEnumerator IEnumerable.GetEnumerator()` return the `GetEnumerator()` method as I assume that is well optimized.

If you'd rather not have the `IEnumerable<string>` interface, I could do a custom json converter for it


jtkech:
LGTM

PR #4654, opened 2019-10-26
Make alert dismissable

hishamco:
Fix #4236

Alert

PR #5820, opened 2020-03-29
apply message styles

giannik:
apply bootstrap styles to message based on notification type

### DIFF
--- a/src/OrchardCore.Themes/TheAdmin/Views/Message.cshtml
+++ b/src/OrchardCore.Themes/TheAdmin/Views/Message.cshtml
@@ -1,4 +1,4 @@
-<div class="alert alert-warning alert-dismissible fade show" role="alert">
+<div class="alert @(Model.Type.ToString() == "Error" ? "alert-danger" : Model.Type.ToString() == "Success"? "alert-success": Model.Type.ToString() == "Information" ? "alert-info" : "alert-warning") alert-dismissible fade show" role="alert">
     @Model.Message
     <button type="button" class="close" data-dismiss="alert" aria-label="Close">
         <span aria-hidden="true">&times;</span>
--- a/src/OrchardCore.Themes/TheAdmin/Views/Message.cshtml
+++ b/src/OrchardCore.Themes/TheAdmin/Views/Message.cshtml
@@ -1,4 +1,7 @@
-<div class="alert @(Model.Type.ToString() == "Error" ? "alert-danger" : Model.Type.ToString() == "Success"? "alert-success": Model.Type.ToString() == "Information" ? "alert-info" : "alert-warning") alert-dismissible fade show" role="alert">
+@{
+    string type = Model.Type.ToString().ToLowerInvariant();
+}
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Themes/TheAdmin/Views/Message.cshtml
```diff
@@ -1,4 +1,4 @@
-<div class="alert alert-warning alert-dismissible fade show" role="alert">
+<div class="alert @(Model.Type.ToString() == "Error" ? "alert-danger" : Model.Type.ToString() == "Success"? "alert-success": Model.Type.ToString() == "Information" ? "alert-info" : "alert-warning") alert-dismissible fade show" role="alert">
```
hishamco: Could we use the pattern matching?

giannik: an example ?

deanmarcussen:
This shape was done incorrectly in the first instance.

It needs to use the `TagBuilder` pattern, then it would take the correct classes (for which we have the correct css already) from the Core `Message` shape, and the `alert-dismissable` etc classes could be added to the existing `Message` shape. 

@hishamco I am sure I explained this on the original pr, can you find the example from there for @giannik please

hishamco:
Ya, I remember this one, let me shoot the original PR that I made

hishamco:
@deanmarcussen I think you mean this one https://github.com/OrchardCMS/OrchardCore/pull/4654

deanmarcussen:
@giannik I have updated this to use the existing css styles we have for the message-type (they just extend the bootstrap ones).

If you agree with the change I will approve the pr

src/OrchardCore.Themes/TheAdmin/Views/Message.cshtml
```diff
@@ -1,4 +1,7 @@
-<div class="alert alert-warning alert-dismissible fade show" role="alert">
+@{
+    string type = Model.Type.ToString().ToLowerInvariant();
+}
+<div class="alert alert-dismissible fade show message-@type" role="alert">
```
Skrypt: Why not using `Model.Type.ToString().ToLowerInvariant();` directly instead?

giannik:
> @giannik I have updated this to use the existing css styles we have for the message-type (they just extend the bootstrap ones).
> 
> If you agree with the change I will approve the pr

sure @deanmarcussen fine by me.

src/OrchardCore.Themes/TheAdmin/Views/Message.cshtml
deanmarcussen: Yes probably could be @Skrypt I did it with `string type` as it is dynamic, and a bit more explicit. will check tomorrow

Skrypt: Well I tried with my website and it works fine. You cast it as a String when you do .ToString() anyways 😄

PR #5827, opened 2020-03-31
Remove unnecessary constructor from DataAnnotationsDefaultErrorMessages

hishamco:
Address @deanmarcussen note https://github.com/OrchardCMS/OrchardCore/pull/5793/files/c658972e60b217c23d3c91cb06ea53454759cc9a#r397679244

### DIFF
--- a/src/OrchardCore/OrchardCore.Localization.Core/DataAnnotations/DataAnnotationsDefaultErrorMessages.cs
+++ b/src/OrchardCore/OrchardCore.Localization.Core/DataAnnotations/DataAnnotationsDefaultErrorMessages.cs
@@ -1,6 +1,6 @@
 using Microsoft.Extensions.Localization;
 
-namespace OrchardCore.Localization.Core.DataAnnotations
+namespace OrchardCore.Localization.DataAnnotations
 {
     /// <summary>
     /// This is just a marker class to allow the POExtractor to extract the default error messages for data annotations attributes.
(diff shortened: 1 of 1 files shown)


hishamco:
I will assign it to `null` to make the build pass

sebastienros:
This is very dangerous as any code that would somehow reference this class, not even use it, would trigger the static constructor and the properties would trigger a Null Reference Exception.

I would suggest to keep the non-static properties, and make this class abstract (and sealed if that compiles) such that we are sure it's not instantiated.

hishamco:
This safe right now ..

src/OrchardCore/OrchardCore.Localization.Core/DataAnnotations/DataAnnotationsDefaultErrorMessages.cs
```diff
@@ -1,118 +1,121 @@
...
+        public string CustomValidationAttributeTypeMustBePublic =>  S["The custom validation type '{0}' must be public."];
 
-        public string CustomValidationAttributeValidationError => S["{0} is not valid."];
+        public string CustomValidationAttributeValidationError =>  S["{0} is not valid."];
 
-        public string CustomValidationAttributeValidatorTypeRequired => S["The CustomValidationAttribute.ValidatorType was not specified."];
+        public string CustomValidationAttributeValidatorTypeRequired =>  S["The CustomValidationAttribute.ValidatorType was not specified."];
 
-        public string DataTypeAttributeEmptyDataTypeString => S["The custom DataType string cannot be null or empty."];
+        public string DataTypeAttributeEmptyDataTypeString =>  S["The custom DataType string cannot be null or empty."];
```
sebastienros, line 45: Extrace space added on every line before `S[]`

hishamco: Good catch, I will fix it now ..

PR #5829, opened 2020-04-01
add line break to make the section easy to see

infofromca:


hishamco:
Could please provide a screen shoot?

deanmarcussen:
Can you provide a before and after screenshot please?

I am fine with the idea, but we use this style in many places, so it would be good to see the difference, so we can decide if it should be everywhere or not.

infofromca:
no br
no_br

infofromca:
with br
with_br

infofromca:
no br
image

infofromca:
with br
image

deanmarcussen:
Thanks for the screenshots @infofromca 

What does everyone else think?

I am all good with it, but if we did it here, we would have to do it everywhere we use the alert

i.e. on Site Settings, and the other places that have an alert notification in the top of the view

This one is common throughout, and also placed in the same location
`<p class="alert alert-warning">@T["The current tenant will be reloaded when the settings are saved."]</p>`

ns8482e:
No <hr /> please.

agriffard:
Closing it. Many people don't seem to agree on this addition, including me.

PR #5838, opened 2020-04-02
Update footer dates in theme recipes

deanmarcussen:


hishamco:
This will be changed every year, can we use JavaScript to generate years instead

Skrypt:
`{{ "now" | date: "%Y" }}`

Skrypt:
This is a snippet added from a recipe so I think it will add it to the database as a Widget. So it will parse the Liquid and render it fine.

sebastienros:
Is it really worth creating a template for that? Will add more buffering ... and technically this date doesn't have to vary with time. Also if we can just have to copy it from the templates when we update them ... But I won't prevent you from doing it.

Skrypt:
The template was already there I just renamed it to .liquid instead of .html to remove confusion. If you mean that you don't want it to be dynamic because it adds Liquid parsing then let's revert the change and make the date static but I don't see how it can slow down the website generally.

hishamco:
I like your change nothing but I thought it can be used in Razor somehow

deanmarcussen:
I'm fine with your change @Skrypt The `HtmlBodyPart` on the widget will render the content with liquid whether there is liquid in the html or not, so I can't see any perf impact.

I would probably prefer keeping the name as `.html` rather than `.liquid` as it suggests it's an actual template (which maybe where @sebastienros comment comes from), rather than a snippet, but I'm not picky

sebastienros:
> any perf impact

The parser will be a no-op if there is no liquid tag. There is a lookup for { before it does anything. Here it will have to parse the block and render it.

That's also making me think about all the things liquid is allowing in an html part... Might be an issue, I'll file an issue.

deanmarcussen:
> The parser will be a no-op if there is no liquid tag. There is a lookup for { before it does anything.  Here it will have to parse the block and render it.

I understand the issue now.

sebastienros:
I know it was already there. It was not a Template, is was a bunch of HTML.

I don't know how much impact if will have, maybe not significant, just saying it's definitely parsing and rendering the footer now, on each request. If there was only HTML it would not use any liquid logic.

Skrypt:
Let's use a caching block for the whole thing then

PR #5852, opened 2020-04-03
Themes standardization

agriffard:
Add missing HeadMeta section in all themes

I would like to configure the zones by default in all themes and add to this one this kind of meta in a RawHtml widget:
`<link type="image/x-icon" rel="shortcut icon" href="/TheAgencyTheme/favicon.ico">`
but it would require to remove the wrappers around the widget in that zone.

### DIFF
--- a/src/OrchardCore.Themes/TheAgencyTheme/Views/Layout.liquid
+++ b/src/OrchardCore.Themes/TheAgencyTheme/Views/Layout.liquid
@@ -3,8 +3,8 @@
 <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+    <title>{{ "PageTitle" | shape_new | shape_stringify }}</title>
     {% render_section "HeadMeta", required: false %}
-    <title>{% page_title Site.SiteName %}</title>
     {% resources type: "Meta" %}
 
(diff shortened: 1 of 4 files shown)


agriffard:
May be that there is another way to add these metas in the head, like what is suggested in this issue #5723

agriffard:
How do I use the PageTitle shape on a reazor page for TheTheme ?

agriffard:
An easy way to do it would be to have a Head property on the Site, editable in the General settings, also available in Liquid and that we'd use in the themes.

agriffard:
Or: Would it be possible to create a Taghelper and a LiquidFilter called 'Template' with a parameter name that would call and render the corresponding template?

jtkech:
In liquid we have the `page_title` tag and in a razor page the `@RenderTitleSegments` that both use `IPageTitleBuilder`. And, as i remember, to make it working when rendering a page, we need at least to render the `ContentsMetadata` of the related content item.

ns8482e:
@agriffard IMO     ` {% render_section "HeadMeta", required: false %}` should be  the last line just before `</head>` as in Admin and must be after  all   `{% resources type: "XXX" %}` entries

Skrypt:
This should also work normally : 
`Title.AddSegment(Html.Raw(Html.Encode(Model.ContentItem.DisplayText)));`

PR #5853, opened 2020-04-04
Cleanup unused things

jtkech:
To be continued


sebastienros:
Good to see you are alive

jtkech:
Good to see you are alive too ;)

hishamco:
@jtkech please let me know if you need any help on this, seems the changed files are HUGE ;)

jtkech:
@hishamco 

That's okay it's finished, only simple things but many of them

One thing is that we inject `ILogger<SomeType>` so that the factory returns the right implementation, but then we don't need an `ILogger<SomeType> _logger` field, we just need an `ILogger _logger` field

Idem with the string / html localizers.

hishamco:
I can help you on this, there are many places where `IStringLocalizer <T>` and `IHtmlLocalizer <T>` are used

jtkech:
Thanks but already done this w-end ;)

Also changed all the remaining logger / localizer that was still using property getter

PR #5854, opened 2020-04-04
Site Meta

agriffard:
Add a Meta property to ISite.


agriffard:
When the site settings are set with the recipe, they are correctly set in the database but it seems thta they are not refreshed because the Meta is empty when you edit the General Settings in the admin after the setup is done.

/cc @jtkech

src/OrchardCore.Modules/OrchardCore.Settings/SiteSettings.cs
```diff
@@ -28,5 +28,6 @@ public class SiteSettings : Entity, ISite
         public string CdnBaseUrl { get; set; }
         public RouteValueDictionary HomeRoute { get; set; } = new RouteValueDictionary();
         public bool AppendVersion { get; set; } = true;
+        public string Meta { get; set; }
     }
```
deanmarcussen, line 32: This should be added to the `ISite` interface as well.

And then an update to both the deployment step and the content step is required with the matching property.
(see the comment at the top of the file)

src/OrchardCore.Modules/OrchardCore.Settings/Views/Settings.Edit.cshtml
```diff
@@ -86,3 +86,29 @@
         <span class="hint">@T["Determines whether scripts and stylesheets load in their debuggable or minified form."]</span>
     </div>
 </div>
+
+<div class="form-group" asp-validation-class-for="Meta">
+    <label asp-for="Meta">@T["Meta"]</label>
+    <input asp-for="Meta" class="form-control" />
+    <span asp-validation-for="Meta"></span>
+    <span class="hint">@T["The meta added in the head."]</span>
```
deanmarcussen, line 94:
```suggestion
    <span class="hint">@T["The meta to render in the head section of the current theme."]</span>
```

jtkech:
@agriffard

Yes `GetSiteSettingsAsync()` exposes an `ISite` that also needs to include this new property

agriffard:
Added tabs to split the settings.
Site:

image

Resources:
image

deanmarcussen:
I like the tab to `Resources`, and concept of site meta.

The site meta, however is providing a `favicon` which seems related to the current active Theme?

What happens if a user switches themes, then they must update the site.meta, with the new theme settings?

Do we then need to make it more of a theme settings? Or have a concept of settings that are related to themes?

but I would not block this pr if you disagree @agriffard

agriffard:
OK, changed the recipe to import the favicon in the media and use it.

My goal is to provide the simplest way to add a global meta to your site.

It won't prevent anyone from doing a custom module and manage some specific theme settings.

deanmarcussen:
Ok, but now we have `liquid` again in the layout, which adds a performance cost.

Sorry, I would prefer it how it was. It is, as you say a simple way to set meta on the site, and if the theme is changed, the site meta can be updated

agriffard:
Ok @deanmarcussen reversed commit to previous change.

PR #5856, opened 2020-04-04
Update Readme.md table

cristian-clamsen:
Made changes so the table of Parameter/Example/Description can be easily read without Markdown plugin without affecting Markdown view for Github

### DIFF
--- a/src/docs/reference/modules/Lucene/README.md
+++ b/src/docs/reference/modules/Lucene/README.md
@@ -47,11 +47,11 @@ Executes a query with the specified name and returns the corresponding content i
 
 Verbs: `POST` and `GET`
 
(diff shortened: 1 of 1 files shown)


hishamco:
Could please provide a snap shoot before & after to see the differences?

cristian-clamsen:
its just a manual formatting change. If you have a look at file File Changes you can see the difference.
If you view https://github.com/cristian-clamsen/OrchardCore/blob/patch-1/src/docs/reference/modules/Lucene/README.md GITHUB will not show any difference. 
The purpose for the change is that when users Read.md from visual studio/vscode they will not need a markdown viewer.

agriffard:
Sorry but this kind of change doesn't bring much value.

PR #5861, opened 2020-04-05
Tab placement documentation

### LINKED ISSUE #1822: Tabbed UI for editor views
Not only for widgets displayed in Flows, but for any content item, we should have a way to designate a shape to a tab. The `PlacementInfo` type already has support for this (via the `GetTab` method), but as far as I could tell, there's no UI yet that takes advantage of this. So that would be the scope of this issue.

agriffard:
Explains how to use placement file to position a part in a different tab.

Fixes #1822


hishamco:
It would be nice if add the tabs in one of our templates, seems this feature is hidden from others included me ;)

Issue #5348, opened 2020-01-23
Export a Content in json


agriffard:
Goal: Provide an Export action in the content items list that generates a {contentid}.recipe.json file with the appropriate structure and deployment step that you can then import thanks to the Recipes page.

agriffard:
If you can select multiple content items and export them in one file with a batch action, that is even better.

deanmarcussen:
It would be great to be able to add content to a deployment queue, so you could batch up multiple selections into one queue (deployment destination, be it file or a remote deployment source).

PR #5487, opened 2020-02-08
Make Content Recipe Step Idempotent

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/5230 
Fixes https://github.com/OrchardCMS/OrchardCore/issues/3089
Fixes https://github.com/OrchardCMS/OrchardCore/issues/4694

Putting this forward as a draft to demonstrate the issue that occurs when using deployment plans to manage content from a stage to prod environment.

The problem that occurs then, when importing a new version of an existing content item, is that the old content remains listed as published or latest, because the old content is not queried and updated.

~~I've done all this in the recipe step, without adding an import handler to the content manager, as if we did so it should be designed.~~

~~This makes recipe imports idempotent, but does not include any verb patching as has been suggested in https://github.com/OrchardCMS/OrchardCore/issues/1970. This is because deployment sources don't currently include any option to mutate a content item, they only care about published items.~~

~~technically a recipe could be modified by hand to try and patch an item, but right now that wouldn't work, and is not currently designed to do so, they are currently designed to operate solely on whether the contentitemversion is recognized or not.~~

~~So this also makes some arbitrary decisions about what content should remain published.
e.g. if it's in the recipe and a new version id, it takes priority over any existing drafts, and becomes published.~~

~~Otherwise the possibility exists that an item is drafted on a prod server, a new version published on the stage server, and if the prod draft item remained `latest` then it would still potentially be behind what was imported from the stage server. But we can design that better if we want to be less arbitrary.~~

~~One of the issues that remains is potential for problems because this is not done in a `contentmanager.import` (but I understand why that was removed after all the O1 issues around it) is issue https://github.com/OrchardCMS/OrchardCore/issues/3405.
The potential issues exists that if a scoped index provider tries to retrieve an imported version from the content manager it could end up with a new entity/query, because of the `ContentManagerSession` cache.~~

~~So we may want to design an `Import` method, that covers that better, than using the existing `Create` method~~

PR #5862, opened 2020-04-05
Export to deployment target / Add to deployment plan

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/5348

wip. and for after https://github.com/OrchardCMS/OrchardCore/pull/5487

Still to do: bulk actions

To use, add a deployment plan, then add a `Click to deploy` step (and other steps first, like content definition, if you want to)

Running the plan by itself will produce no outcome, but when ran from the admin list, it will export the current content item (either the published or the draft version, are available as options if both are present)

The feature will find the first deployment plan with a `Click to deploy` step, or no-op if there is non


deanmarcussen:
after discussion with @agriffard have moved the feature to the Admin List.

deploydraft

deploypublished

sebastienros:
Make it a feature, with a dependency on Deployment. Then create a setting to select which deployment plan to chose, and optionally create a new deployment plan during the migration and set it in the setting.

It's a nice tool, but kind of extreme to just be able to see a content item, a link to the API endpoint would just work. And we could easily change this endpoint to support bulk actions. Maybe the two fearures should be available, as running a deployment with single click from this view might be dangerous. And you need Deployment permissions to see a content item payload (write vs read).

agriffard:
All the ClickToDeployContent... classes should also be renamed into ExportContent...

deanmarcussen:
@sebastienros I started it as a feature, then changed my mind. So I will change my mind back.

I will do a second simpler feature for `Export`, if I can base it on the ApiController (not sure cookie permissions will work), or in some way. 

And yes, already planning to add the selected deployment plan via a settings

So I will change the text here back to `Deploy` @agriffard and the other feature will be `Export`. Sound ok?

sebastienros:
Should the link for deployment plan just be called "Add to deployment plan"? And maybe it could just popup the list if plans then. If we have a dedicated Export action.

deanmarcussen:
I did think of doing that, but then there are two modals, one to select a deployment plan, then another to select a destination, so it becomes `ClicksToDeploy`, which was more steps than I was hoping for.

deanmarcussen:
Feature renamed as per @agriffard feedback ;)

Features now included in this pr
- View /Download as JSON
- Export To Deployment Target
- Add To Deployment Plan

some screenshots 

Bulk Actions Menu
bulkactions

Actions Menu (without a draft)

withoutdrafts

Actions Menu (with a draft)

withdrafts

Note: Add to deployment plan doesn't handle drafts (because it now just references a content item id, it would be weird to have an action that is related to a **specific** draft actually reference a content item id, and say export **any** draft of this content item id)

src/OrchardCore/OrchardCore.Deployment.Core/CommonPermissions.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace OrchardCore.Deployment
+{
+    /// <summary>
+    /// This class contains the source references to the OrchardCore.Deployment module permissions so they can be used in other modules
+    /// without having to reference the OrchardCore.Deployment by itself.
+    /// </summary>
+    public class CommonPermissions
+    {
+        public static readonly Permission ManageDeploymentPlan = new Permission("ManageDeploymentPlan", "Manage deployment plan");
```
agriffard, line 11: `ManageDeploymentPlan` => `ManageDeploymentPlans` may be

deanmarcussen: Hmm, could be, I haven't changed this (just moved it). Could just update the description, without changing the permission

src/OrchardCore.Modules/OrchardCore.Contents/Deployment/ExportContentToDeploymentTarget/ExportContentToDeploymentTargetAdminMenu.cs
```diff
@@ -0,0 +1,40 @@
...
+            if (!String.Equals(name, "admin", StringComparison.OrdinalIgnoreCase))
+            {
+                return Task.CompletedTask;
+            }
+
+            builder
+                .Add(S["Configuration"], configuration => configuration
+                    .Add(S["Import/Export"], S["Import/Export"], import => import
+                        .Add(S["Settings"], settings => settings
+                            .Add(S["Export Content To Deployment Target"], "am-" + S["Export Content To Deployment Target"], deployment => deployment
```
agriffard, line 28: Please merge dev and use .PrefixPosition() instead of `"am-" +`.

deanmarcussen: Done

agriffard:
Conflict to resolve.

src/OrchardCore.Modules/OrchardCore.Contents/Views/Admin/List.cshtml
```diff
@@ -82,10 +82,7 @@
                             @T["Actions"]
                         </button>
                         <span class="dropdown-menu dropdown-menu-right" aria-labelledby="bulk-action-menu-button">
-                            @foreach (var item in Model.Options.ContentsBulkAction)
-                            {
-                                <a class="dropdown-item" href="javascript:void(0)" data-action="@item.Value" data-title="@T["Bulk Action"]" data-message="@T["Are you sure you want to {0} these items?", @item.Text.ToLower()]">@item.Text</a>
-                            }
+                            @await DisplayAsync(await New.Contents__BulkActions(BulkActions: Model.Options.ContentsBulkAction))
```
deanmarcussen, line 85: Looking for reviews on this. The idea is to use a shape, and then `IShapeTableProviders` to extend the Contents Admin List so features such as these, can provide `BulkActions` via an `IShapeTableProvider` (see comment above).

If this is a `good` technique for extensibility, I would on another pr, a `Shape` that can be used to extend the content list filters (to filter by localization etc), noting that we already have query filter providers, that can mutate the content items query, we just need a `Shape` to display the extendable options. (refer `IContentAdminFilters`)

src/OrchardCore.Modules/OrchardCore.Contents/Deployment/ExportContentToDeploymentTarget/ExportContentToDeploymentTargetShapes.cs
```diff
@@ -0,0 +1,75 @@
...
+                        var siteSettings = await _siteService.GetSiteSettingsAsync();
+                        var exportContentToDeploymentTargetSettings = siteSettings.As<ExportContentToDeploymentTargetSettings>();
+                        if (exportContentToDeploymentTargetSettings.ExportContentToDeploymentTargetPlanId != 0)
+                        {
+                            displaying.Shape.Metadata.Wrappers.Add("ExportContentToDeploymentTarget_Wrapper__ActionDeploymentTarget");
+                        }
+                    }
+                });
+
+            builder.Describe("Contents__BulkActions")
```
deanmarcussen, line 45: This is the extension point for the `Contents__BulkActions` shape, as discussed below

agriffard:
Ready to merge during next triage meeting?

deanmarcussen:
Need to update this to use the new zones for the admin list, then it will be ready again @agriffard

deanmarcussen:
@agriffard updated to use the new admin list zones.

Could do with a test if you have any time, but I think ready to merge

agriffard:
I have multiple deployment plans with the same name so this error occurs after enabling the feature:
`ArgumentException: An item with the same key has already been added. Key: `

deanmarcussen:
The bug is deployment plans themselves. The `Name` should be unique, because the deployment plan step is expecting to be able to export them keyed by `Name` (so it can either create or update them).

But it is not checking that, nor is nor actually trying to create or update them properly (it will recreate them every time, and you will end up with multiples named the same, rather than actually updating an existing plan)

I will open a seperate issue

Issue #5843, opened 2020-04-02
Workflow editor (slow)


Skrypt:
When editing a Workflow it became slower recently if you want to unlink tasks/events.

PR #5872, opened 2020-04-06
Remove font awesome pseudoSearch config

### LINKED ISSUE #5864: Culture vuejs select is slow
The vuejs Culture Select is slow to select new cultures.

To select an item from the cultures list when localization is enabled takes around 25 seconds for the add method to complete, and the culture to become selectable.

I haven't had time for a good look, but I can see from the code we have fallen into a `v-for` and `v-if` trap. https://vuejs.org/v2/guide/list.html#v-for-with-v-if

However I suspect it's the add method that is running slowly.

Must be a regression as it wasn't running this slowly last time I used it.

agriffard:
Do not use 'search pseudo elements' in font awesome anymore because it slows the pages.

Fixes #5864

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Flows/Views/ContentCard-FlowPart.Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Flows/Views/ContentCard-FlowPart.Edit.cshtml
@@ -28,7 +28,8 @@
                 @if (Model.CanInsert != false)
                 {
                     <div class="btn-group">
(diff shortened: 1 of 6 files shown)


sebastienros:
Will the issue with the icon still be fixed?

Skrypt:
Probably fixes #5843 too

deanmarcussen:
Is the icon on the flow part still an issue with this @agriffard or can we merge it?

I didn't catch the details of what was happening with the icon, but if you need a hand let me know.

Was the icon working on a dev build, but not when used from the myget packages?

agriffard:
The icon is not toggling anymore but at least this is not using pseudosearch anymore.

I don't know how to switch the icon classes because it is using the .js and loading a svg and commenting the icon.

PR #5877, opened 2020-04-07
Update Themes package dependencies

agriffard:
Bootstrap


deanmarcussen:
I thought this would need a gulp rebuild as the css will change with the new bootstrap version?

Or is bootstrap just used for the variables here? which wouldn't need a rebuild then

agriffard:
I did a gulp rebuild and didn't have any changes.

PR #5830, opened 2020-04-01
Fixes Workflow Timer

jtkech:
Fixes #5810 
Fixes #5231 
Fixes #5223 

I think that it fixes #3436 

As i commented in the source code: The 'TimerEvent' is a one minute signal and in fact this is the timer activity that checks if it has expired. So, because a timer event can't occur again before it has previously expired, if there is already a workflow instance that is halted on a timer activity, we should not start a new instance, we just have to resume it.

Otherwise, if you have a timer event as a starting activity in a workflow type that is not marked as a singleton. E.g if the timer is set to expire every 5 minutes, you will create 5 instances that will get called, one after the other in the same scope.

PR #5878, opened 2020-04-07
Tenant setup event

### LINKED ISSUE #5858: Need event after a tenant is setup
I need a workflow event triggered when a tenant setup is done.

agriffard:
Add a Workflow event when a tenant is setup.

Fixes #5858

@jtkech Please help to figure out how it do it right. 
There is an issue when I want to setup a tenant:
`Unable to resolve service for type 'OrchardCore.Setup.Events.ISetupEventHandler' while attempting to activate 'OrchardCore.Setup.Services.SetupService'.`

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Tenants/Handlers/SetupEventHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Tenants/Handlers/SetupEventHandler.cs
@@ -1,0 +1,36 @@
+using System;
+using System.Threading.Tasks;
+using OrchardCore.Environment.Shell;
+using OrchardCore.Setup.Events;
+using OrchardCore.Tenants.Workflows.Activities;
+using OrchardCore.Workflows.Services;
+
+namespace OrchardCore.Tenants.Workflows.Handlers
+{
+    public class SetupEventHandler : ISetupEventHandler
+    {
+        private readonly IWorkflowManager _workflowManager;
+
+        public SetupEventHandler(IWorkflowManager workflowManager)
+        {
+            _workflowManager = workflowManager;
+        }
+
+        public Task Setup(string siteName,
+            string userName,
+            string email,
+            string password,
+            string dbProvider,
+            string dbConnectionString,
+            string dbTablePrefix,
+            string siteTimeZone,
+            Action<string, string> reportError)
+        {
+            return _workflowManager.TriggerEventAsync(nameof(TenantSetupEvent),
+                input: new { AdminUserName = userName, AdminEmail = email, AdminPassword = password, DatabaseProvider = dbProvider, DatabaseConnectionString = dbConnectionString, DatabaseTablePrefix = dbTablePrefix, SiteTimeZone = siteTimeZone },
+                correlationId: siteName
+            );
+        }
+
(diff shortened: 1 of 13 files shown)


jtkech:
During a setup, as other features, `OC.Tenants` is not enabled, so there is no `ISetupEventHandler` that's why it fails to resolve one. We can add features to the setup context as we do for the setup one.

                .AddSetupFeatures("OrchardCore.Setup")

`ISetupEventHandler` are already resolved and called further but after the tenant recipes have been executed and in the context of a tenant / shell scope. So, if your recipe enables `OC.Tenants` and defines a WF type using the setup event, it may work.

Note: Here `ISetupEventHandler` are resolved as an IEnumerable so it doesn't fail even if there is no registered handlers.

But do you want to execute a WF in the context of the tenant that has been setup? Or to execute a WF e.g always in the context of the Default tenant after any other tenant has been setup?

I will think about it.

agriffard:
I want to execute a Workflow from the default tenant: After a setup, I want to use a timer event to disable the tenant after N days.

jtkech:
Okay

One "problem" is that we are in the context of the tenant that has been setup, so we need to create a scope on the Default tenant to resolve the related service (another interface).

Hmm, i did something similar in my old distributed branch, to send any tenant events to the Default tenant, by creating a scope on the Default tenant and resolving `IShellEvents` that is registered only for the Default tenant, [e.g see here](https://github.com/OrchardCMS/OrchardCore/blob/3f08530f5d761a8341d966ed0a027a3a32d26983/src/OrchardCore/OrchardCore.Abstractions/Shell/Extensions/ShellHostExtensions.cs#L98).

So, seems that you need

1. An `ISetupEventHandler` but tied to a feature that needs to be enabled e.g through the recipe, so maybe better to define it in the setup module itself. It will get called after the setup.

2. In this handler, resolve the default tenant and create a scope on it, then resolve another interface and call on it a method where you can also pass the shell settings / name of the tenant that has been setup.

3. In the implementation of the last interface (only registered on the default tenant) do what you did.

Can be done later but i think that would need to be done more globally for any tenant / shell event, when released, reloaded and so on. And by the core e.g through an `IShellHost` extension that would create for us the scope on the default tenant and resolve e.g `IShellEvents` handlers.

    _shellHost.ShellEventAsync(e => e.SetupAsync(_shellSettings.Name, ...))

But can be done later, otherwise point 1. and 2. could be done directly in the setup service.

Note: And in a distributed env some tenant events may need to be propagated e.g by a message bus OR by sharing some distributed data / states, e.g through the database + distributed cache.

---

Maybe another solution by using a bgTask on the default tenant that would get all shell settings through the shell host and cache all previous states, and then trigger the setup event when a given one change to `Running`. Here again, it's a states checking solution in place of using events, but yes maybe better to send an event at the end of the setup.

Note: There is an issue with the timer event, see #5830, but maybe okay if it is not used as a **starting** activity.

Last funny idea, a WF on each tenant using a timer starting activity, and marked as a singleton because of the #5830 issue. The WF will be automatically executed when the tenant has been setup, then it will disable itself. Hmm, but someone would be able to edit the WF and stop it ;)

PR #5896, opened 2020-04-09
[doc] Correct shape tags usage with arrays

TFleury:
`shape_add_alternates`, `shape_add_wrappers` and `shape_add_classes` support one argument only.

### DIFF
--- a/src/docs/reference/modules/Liquid/README.md
+++ b/src/docs/reference/modules/Liquid/README.md
@@ -483,8 +483,10 @@ Adds alternates to a shape.
 Input
 
 ```liquid
-{% shape_add_alternates my_shape "alternate1", "alternate2" %}
 {% shape_add_alternates my_shape "alternate1 alternate2" %}
+
+{% assign my_alternates = "alternate1,alternate2" | split: "," %}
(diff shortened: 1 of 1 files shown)


deanmarcussen:
Was it not working with multiple arguments?

The code suggests it should, but I see the split (in code) is done on `Split(' ')`, not `Split(', ')` as it says in the docs

TFleury:
TLDR: no it only takes one argument.

Here is the code for `shape_add_classes`, the two others are similar :
```C#
var classes = arguments["classes"].Or(arguments.At(0));

if (classes.Type == FluidValues.String)
{
	var values = classes.ToStringValue().Split(' ', StringSplitOptions.RemoveEmptyEntries);

	foreach (var value in values)
	{
		shape.Classes.Add(value);
	}
}
else if (classes.Type == FluidValues.Array)
{
	foreach (var value in classes.Enumerate())
	{
		shape.Classes.Add(value.ToStringValue());
	}
}
```

So it takes "classes" named argument and fallback to first argument. Argument can be string or array.
`Split(' ')` is used for string argument, when you use this syntax :
```liquid
{% shape_add_classes my_shape "class1 class2" %}
```

PR #5899, opened 2020-04-09
Allow registering style block

### LINKED ISSUE #5897: Feature : Allow registering style block from shapes
The intent is to have something similar to custom scripts feature :

Razor
```
<style>
    .my-class {
        /* some style */
    }
</style>
```

Liquid
```
{% styleblock %}
    .my-class {
        /* some style */
    }
{% endstyleblock %}
```

To do :
- [ ] Allow inline declaration with `StyleTagHelper`
- [ ] Add 'styleblock' Liquid `NamedHelperBlock`
- [ ] Document feature

TFleury:
Closes #5897 : Allow registering style block

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewTemplate.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewTemplate.cs
@@ -87,7 +87,8 @@ namespace OrchardCore.DisplayManagement.Liquid
             Factory.RegisterBlock<HelperBlock>("block");
             Factory.RegisterBlock<NamedHelperBlock>("a");
             Factory.RegisterBlock<NamedHelperBlock>("zone");
-            Factory.RegisterBlock<NamedHelperBlock>("scriptblock");            
+            Factory.RegisterBlock<NamedHelperBlock>("scriptblock");
+            Factory.RegisterBlock<NamedHelperBlock>("styleblock");
 
             // Dynamic caching
             Factory.RegisterBlock<CacheBlock>("cache");
--- a/src/OrchardCore/OrchardCore.ResourceManagement/TagHelpers/StyleTagHelper.cs
+++ b/src/OrchardCore/OrchardCore.ResourceManagement/TagHelpers/StyleTagHelper.cs
@@ -1,14 +1,18 @@
 using System;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc.Rendering;
 using Microsoft.AspNetCore.Razor.TagHelpers;
 
 namespace OrchardCore.ResourceManagement.TagHelpers
 {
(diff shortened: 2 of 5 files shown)


src/OrchardCore/OrchardCore.ResourceManagement/TagHelpers/StyleTagHelper.cs
```diff
@@ -1,14 +1,18 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc.Rendering;
 using Microsoft.AspNetCore.Razor.TagHelpers;
 
 namespace OrchardCore.ResourceManagement.TagHelpers
 {
     [HtmlTargetElement("style", Attributes = NameAttributeName)]
     [HtmlTargetElement("style", Attributes = SrcAttributeName)]
+    [HtmlTargetElement("style", Attributes = AtAttributeName)]
     public class StyleTagHelper : TagHelper
```
deanmarcussen, line 11: I'm surprised we even support the `At` attribute for styles. I can't think why or how it could eve be anything but placed at the `Head`. Maybe just to be explicit?

A hint for when you get to looking at how the dependencies work

They require a `Name` attribute to work, so I would use that as your defining `HtmlTargetElement` rather than `At`

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewTemplate.cs
```diff
@@ -87,7 +87,8 @@ static LiquidViewTemplate()
             Factory.RegisterBlock<HelperBlock>("block");
             Factory.RegisterBlock<NamedHelperBlock>("a");
             Factory.RegisterBlock<NamedHelperBlock>("zone");
-            Factory.RegisterBlock<NamedHelperBlock>("scriptblock");            
+            Factory.RegisterBlock<NamedHelperBlock>("scriptblock");
+            Factory.RegisterBlock<NamedHelperBlock>("styleblock");
```
ns8482e, line 91: How about creating new "stylesheet" tag instead of "styleblock" for inline style that support both CSS and SCSS?.

Some of VS code extension already support syntax highlighting for "stylesheet" for liquid syntax. https://shopify.dev/tutorials/develop-theme-use-sections

e.g. https://marketplace.visualstudio.com/items?itemName=sissel.shopify-liquid and https://marketplace.visualstudio.com/items?itemName=killalau.vscode-liquid-snippets

src/OrchardCore/OrchardCore.ResourceManagement/TagHelpers/StyleTagHelper.cs
ns8482e: IMHO add new StylesheetTagHelper for inline styles that supports both css/less/scss

TFleury:
Closed in favor of #5903

TFleury:
Merged into #5903

PR #5507, opened 2020-02-09
Added new method RegisterEditorResources to register widget editing resources instead of calling BuildEditorAsync

ns8482e:
Continuation with PR #5499 

Fixes #5149 
Fixes #3250 
Fixes #4223

If content item contains parts like `FlowPart`, `BagPart`, `WidgetListPart` mock editor is created by calling `BuildEditorAsync` for all the widgets to register resources (scripts and styles required for widget editing) with the Page.

`BuildEditorAsync `Invokes all the Display Handlers registered with Service collection. 
`ContentPreviewDriver `and `ContentsDriver `are such display handlers that invoked during `BuildEditorAsync`, that are responsible for creating action buttons like `Preview`, `Save Draft`

These mock editor for widgets are not intended to render But when `placement.json` or `<zone>` tag helper is used to place shapes on specific zone, shapes from mock editor are also placed on these zones and rendered.

If mock editor for widget contains content specific initialization `<script>` which are not intended to be included as resources for widget editing but are included in the Page unnecessarily  and in some cases throws java script error or adhoc fix like below is required. https://github.com/OrchardCMS/OrchardCore/blob/a5e11d392ac191b6b9d1baa99a1753fb1fdbf4fa/src/OrchardCore.Modules/OrchardCore.Liquid/Views/LiquidPart.Edit.cshtml#L20-L24

Also it increases the response size.

The solution is, 

Added new Method `Task RegisterEditorResources(ContentItem content);` 
Added property `public bool IsMockRendering ` on RazorPage

New method, `RegisterEditorResources `is similar to `BuildEditor`, except it renders everything using `MockLayout` instead of `Layout` from LayoutAccessor, so that any mock UI rendered are not placed on Page `Layout` zones 

Widget can now use `if(!IsMockRendering)` to include `DOM` `<script>`, only if it's not mock rendering 
Also added `IsMockRendering` condition for `BagPart.Edit.cshtml`, `FlowPart.Edit.cshtml`, `WidgetsListPart.Edit.cshtml`, that reduced the file size to `376 KB` from `505 KB` in my case.

image

PR #5413, opened 2020-01-31
Added support for @section syntax for Views

ns8482e:
Fixes #5412

Zones can be defined using @ section in Views and Shapes 

```
@section Header
{
     <h1>My Header</h1> 
}
```

PR #5903, opened 2020-04-10
Allow registering named style and script resources with inline content

### LINKED ISSUE #5897: Feature : Allow registering style block from shapes

TFleury:
Closes #5897 : Allow registering named style and script resources with inline content

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewTemplate.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewTemplate.cs
@@ -89,6 +89,7 @@ namespace OrchardCore.DisplayManagement.Liquid
             Factory.RegisterBlock<NamedHelperBlock>("zone");
             Factory.RegisterBlock<NamedHelperBlock>("form");
             Factory.RegisterBlock<NamedHelperBlock>("scriptblock");
+            Factory.RegisterBlock<NamedHelperBlock>("styleblock");
 
             // Dynamic caching
             Factory.RegisterBlock<CacheBlock>("cache");
--- a/src/OrchardCore/OrchardCore.ResourceManagement.Abstractions/IResourceManager.cs
+++ b/src/OrchardCore/OrchardCore.ResourceManagement.Abstractions/IResourceManager.cs
@@ -31,15 +31,21 @@ namespace OrchardCore.ResourceManagement
         RequireSettings RegisterResource(string resourceType, string resourceName);
 
         /// <summary>
-        /// Registers a custom script tag on at the head.
+        /// Registers a custom script tag at the head.
         /// </summary>
         void RegisterHeadScript(IHtmlContent script);
 
         /// <summary>
-        /// Registers a custom script tag on at the foot.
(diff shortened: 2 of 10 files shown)


TFleury:
I think it could replace the existing [custom script feature](https://orchardcore.readthedocs.io/en/latest/docs/reference/modules/Resources/#custom-scripts). It just requires an additional name attribute.
So, we will be able to remove `RegisterHeadScript` and `RegisterFootScript` from `ResourceManager`.

#### Usage :
Razor
```razor
<script asp-name="Test" at="Head" depends-on="jQuery">
    console.log("Hello");
</script>

<style asp-name="custom-style" depends-on="thetheme-bootstrap-oc">
    .my-class {
        /* Some css rules */
    }
</style>
```
Liquid
```liquid
{% scriptblock name:"Test", at:"Head", depends-on:"jQuery" %}
    console.log("Hello");
{% endscriptblock %}

{% styleblock name:"custom-style", depends-on:"thetheme-bootstrap-oc" %}
    .my-class {
        /* Some css rules */
    }
{% endstyleblock %}
```

deanmarcussen:
I like this option two @TFleury and I would be happy with both this, and your other pr.

For me it makes using an inline script easier, because you don't need to also specify the script you are depending on, earlier in the template.

I wouldn't think to remove the `RegisterHeadScript` and `RegisterFootScript` though, as that would be a breaking change, but over time possibly deprecate?

TFleury:
Merged into #5903

ns8482e:
@deanmarcussen @TFleury comments on general concept - not on PR. 

I didn't like the original implementation of inline script combined with resource but it works because normally `<script>` are placed at `Foot` (works with some bugs that I am trying to fix in PR #5507 ) , So I fear that having in-line style on similar implementation would add similar issues. 

In my opinion inline scripts and inline styles are `content` and not `resources`  hence it should be kept outside resource manager, `zones` are perfect for inline script/style and you can manage order using placement order.

For naming in liquid, it's prefer to use `javascript` instead of `scriptblock` and `stylesheet` instead of `styleblock` as those are known to liquid developers. or add both names.

ns8482e:
FYI with PR #5413 for `@section` It would be even easier for asp.net developer to define scripts with known method, without needing to know orchard core provided implementations of script/style tag helper

```
@section styles {
.my-class {
        /* Some css rules */
    }
}
@section scripts {
    <script>
      alert('foo');
    </script>
}
```

TFleury:
Yes, this is what I used so far.

But, assume you have a shape that injects a custom script/style, and you have n instance of your shape on your page. Unless you use a trick with `HttpContext.Items`, you will end up with n instances of your script/style.

That's what I wanted to solve with this PR.

Issue #5700, opened 2020-03-11
Bug? No template for bag item?


JoshTango:
When you render a bunch of items in a bag it seems to use the standard content template which adds a header to the top of the page for each piece of content.  So if you have 20 items in your bag you will have 20 headers?  Seems like a template just for a bag item is needed where it does not add a header.  This happened with the clean blog theme and the agency theme.

sebastienros:
Can you share a piece of html with the issue, I don't really understand what headers you are talking about, like admin, front-end, ...

JoshTango:
{% zone "Header" %}
<header class="masthead">
    <div class="container">
        <div class="intro-text">
            <div class="intro-lead-in"></div>
            <div class="intro-heading text-uppercase">{{ Model.ContentItem | display_text }}</div>
      </div>
    </div>
</header>
{% endzone %}



<section>
  <div class="container">
      {{ Model.Content | shape_render }}
  </div>
</section>

JoshTango:
that's from content.liquid
it injects something into the Header zone

JoshTango:
regardless I am getting extra headers at top of page

ns8482e:
> {% zone "Header" %}
> 
> {{ Model.ContentItem | display_text }}
> {% endzone %}
> {{ Model.Content | shape_render }}

Is this your custom template? 

Btw, All content shape fallback to content.liquid. So if you want to move title to header zone either use placement for the content type or create custom template for your content type instead of content.liquid

JoshTango:
No this is not my custom template.  Its the code in clean blog and agency theme.

JoshTango:
use clean blog and make a bag and add 5 items to the bag.  Then view it.  You will see 5 headers at top of page.

deanmarcussen:
This is not a bug.

You need to make a content template for your bag items that doesn't render the title in the header

Something like `Content-MyBagItem.liquid` or with the Templates feature

As @ns8482e has said the fallback will always be `Content.liquid`, but when the default doesn't suit your needs, you need to make a template.

JoshTango:
I did make a template for my specific content type in the bag

JoshTango:
let me look further into it

JoshTango:
Ok I still think it is a bug.  there should be a template fall back for bag items that is not the content.liquid template.

JoshTango:
[...]

PR #5499, opened 2020-02-09
Do not create action buttons for contained content item.

ns8482e:
Partially Fixes #5149 - Continuation with PR #5339 

If content item contains other content items (widget) and when `BuildEditorAsync` is called on contained content item (widget) then do not create action buttons.

 When content item is edited, Content Display manager creates Editor shape by calling `_contentItemDisplayManager.BuildEditorAsync`.  `BuildEditorAsync` Invokes all the Display Handlers registered with Service collection.

`ContentPreviewDriver` and `ContentsDriver` are such display handlers that invoked during `BuildEditorAsync`, that are responsible for creating action buttons like `Preview`, `Save Draft`

The problems starts when the content item contains Parts that contains child content items - like `FlowPart`, `BagPart`, `WidgetListPart`. 

Edit shape of these parts again calls `BuildEditorAsync` for every contained content item (widget), that also invokes ContentsDisplayDrivers which creates action buttons which are not required as action buttons are already created by ContentsDisplayDrivers of primary content item being edited.

To differentiate between primary content item and contained content item inside ContentsDisplayDrivers - Added new field `IsContentUpdater` in `IUpdateModel`  that would return `false` for `ModelStateWrapperUpdater`

Issue #5149, opened 2019-12-27
In admin theme, changing Location of buttons using placement.json shows lots of buttons in UI


ns8482e:
In Admin theme, 

When "Actions" zone is added in `Layout.cshtml `  as following

 ```
<div class="ta-content">
            @await RenderSectionAsync("Messages", required: false)
            @await RenderSectionAsync("Breadcrumbs", required: false)

            @await RenderBodyAsync()

            @await RenderSectionAsync("Actions", required: false)


        </div>
```

And use `placement.json` as below

**placement.json**
```
{
  "Content_PublishButton": [
    {
      "place": "/Actions:5"
    }
  ],
  "Content_SaveDraftButton": [
    {
      "place": "/Actions:6"
    }
  ],
  "ContentPreview_Button": [
    {
      "place": "/Actions:6"
    }
  ]
}
```

Edit Content shows lots of buttons in UI.  Publish, Save Draft and Preview button would be seen only once. 

![image](https://user-images.githubusercontent.com/23270244/71535430-b21df000-28cb-11ea-8bb2-2df6aa10b9c4.png)

hishamco:
OMG what's that?!!

ns8482e:
`ContentsDriver` that responsible for creating these button shapes is creating shapes for every widgets.

Skrypt:
Wow that's a strange one 😄

sebastienros:
Your placement file is actually resetting the rule that hides these for widgets. Is there a way to filter by stereotype in the placement file, such that your new rules could just apply to "Content" ?

ns8482e:
These buttons are placed by `ContentsDriver` while Rendering a mock widgets. Display of mock widget itself is not rendered but if  a child shape (i.e. Actions ) is placed on layout zone using placement then its visible regardless of display of that child shape (i..e. mock widget )

ns8482e:
> such that your new rules could just apply to "Content"

I tried to differentiate between "Content" and "Widget" by trying to add `IsPrimaryContent` property in `IUpdateModel`  but due to #5253, It forced me to define that property in every AdminController

sebastienros:
There could be a `stereotype` constraint in the placement file, we had that in O1.

ns8482e:
> There could be a `stereotype` constraint in the placement file, we had that in O1.

At first, the action button shapes should only be generated for primary content item being edited not for mock widgets

sebastienros:
[...]

Issue #4372, opened 2019-09-23
Rendering of Widget ContentItem containing Bag throws exception 


ns8482e:
Error while rendering "widget" content item containing Bag Part. 
`
An unhandled exception occurred while processing the request.
Exception: Shape type 'Widget_Summary' not found
OrchardCore.DisplayManagement.Implementation.DefaultHtmlDisplay.ExecuteAsync(DisplayContext context) in DefaultHtmlDisplay.cs, line 125
`
To reproduce,

1. Create a widget content type "News"
2. Create a widget content type "News Roll", Add Bag Part and select "Contained Content Types" News.
3. Create New "News Roll" publish and click View.

Default display type should be Detail if display type is null

https://github.com/OrchardCMS/OrchardCore/blob/f832db952faf5c55e7f1d43ef74226214675dc88/src/OrchardCore.Modules/OrchardCore.Flows/Views/BagPart.cshtml#L9

jptissot:
I had the same bug, you can mitigate by implementing in your theme: 

Shape name: `BagPart`

```liquid
{% for item in Model.BagPart.ContentItems %}
{{ item | shape_build_display | shape_render }}
{% endfor %}
```

sebastienros:
@jptissot can you provide a fix as a PR if you agree with the change? I would take it for rc, seems low risk and an easy bug to get into

deanmarcussen:
The other fix for this is just the set the display type you want, i.e. `Detail` in the `BagPartSettings`.

Probably we could consider a WidgetSummary shape, otherwise changing the default might affect all the other BagPart default displays

sebastienros:
We can adda Summray template that has the same content as the DEtails one. @deanmarcussen you want to do it?

ns8482e:
@deanmarcussen BagPart uses  BagPartSettings but it fallback to "Summary" only in event of null that resulted in this exception.

deanmarcussen:
I’m not near the computer tonight so if someone else can?

sebastienros:
I will then.

PR #5904, opened 2020-04-10
Template for types contained in BagPart.

microposmp:
Implement overrides for types contained in BagParts as suggested in #5700


src/OrchardCore.Modules/OrchardCore.Flows/Views/BagPart.cshtml
```diff
@@ -6,7 +6,7 @@
 <section class="flow">
     @foreach (var item in Model.BagPart.ContentItems)
     {
-        var itemContent = await ContentItemDisplayManager.BuildDisplayAsync(item, Model.BuildPartDisplayContext.Updater, Model.Settings.DisplayType ?? "Detail", Model.BuildPartDisplayContext.GroupId);
+        var itemContent = await ContentItemDisplayManager.BuildDisplayAsync(item, Model.BuildPartDisplayContext.Updater, Model.Settings.DisplayType ?? "Contained", Model.BuildPartDisplayContext.GroupId);
```
deanmarcussen, line 9: I see what you are trying to do here @microposmp but I don't think you can introduce a new display type `Contained` here.

The drivers (part and field) only render content in a location of `Detail` or `Summary` (by default) so a new display type would likely render empty content here.

We suggested on the issue to simply add a couple of alternates at this point.

It would just need to be done in such a way that the alternates do not take precedence over the normal content item alternates (probably add to `Displaying`)

microposmp: Changed to add alternates in this file. I see the issue you describe with the drivers (part and field). I was not able to figure out how to see if the content type being rendered was contained in a BagPart.

microposmp:
Added a change to filter out empty titles.
Updated the documentation for the liquid sample to add these alternates.

src/OrchardCore.Modules/OrchardCore.Flows/Views/BagPart.cshtml
```diff
@@ -8,6 +8,9 @@
     {
         var itemContent = await ContentItemDisplayManager.BuildDisplayAsync(item, Model.BuildPartDisplayContext.Updater, Model.Settings.DisplayType ?? "Detail", Model.BuildPartDisplayContext.GroupId);
 
+        itemContent.Metadata.Alternates.Add("Content_Contained");
+        itemContent.Metadata.Alternates.Add($"Content_Contained__{item.ContentItem.ContentType}");
+
```
deanmarcussen, line 13: Doing it like this makes the `Content_Contained` alternate takes precedence over all the normal alternates, like `Content_MyBagItem`, which we can't do.

@jtkech may have some more advice on how the alternate precedence works.

Possibly
```
itemContent.Metadata.OnProcessing(shape => {
         shape.Metadata.Alternates.Add("Content_Contained");
        shape.Metadata.Alternates.Add($"Content_Contained__{item.ContentItem.ContentType}");
        return Task.FromResult(shape);
        });
```

Might work, but I would need @jtkech to confirm this will make it last, in terms of precedence

microposmp:
I think this render correctly.

If rendered inside a `BagPart` it will take the following precedence:
1. `Content_Contained__[ContentType]`
2. `Content_Contained`
3. ...

If rendered outside a BagPart these alternates will never be used if not explicitly added.

If `[ContentType]` is also used as a root content type and an alternate for `Content__[ContentType]` is created, it will probably render with a header the same way as in `Content.cshtml`. If `Content__[ContentType]` takes precedence over `Content_Contained` and no type specific alternate is provided it will render the header also when rendered as a child in a `BagPart`.

I can do a short demo and try to explain my thinking.

src/OrchardCore.Modules/OrchardCore.Flows/Views/BagPart.cshtml
jtkech: So, `Contained` is a kind of **display mode**,

@deanmarcussen so you would want the following reverse precedence (alternates order)

    Content` // <= in fact not an alternate, the actual shape type on which we fallback

    // The alternate we add but with a lower precedence than 'Content__MyBagItem'
    Content_Contained

    // Normal alternate added in 'ContentItemDisplayManager.BuildDisplayAsync()'
    Content__MyBagItem

    // The normal alternates added in 'Shapes.cs' in '("Content").OnDisplaying()'
    Content_Detail
    Content__0 // <= here 0 is the contentItem.Id, equals to 0 as it is in a bag
    Content_Detail__MyBagItem
    Content_Detail__0 <= here 0 is the contentItem.Id, equals to 0 as it is in a bag

    // The other alternate we add with the higher precedence
    Content_Contained__MyBagItem <= the more specialized one

So specifying the content type would have an higher weight than specifying the `Contained` "display mode". Notice that this is not the case for the display type, `Content_Detail` is higher than `Content__MyBagItem`. So maybe normal that, if we already have `Content-MyBagItem.cshtml`, we would also need `Content-MyBagItem.Contained.cshtml`.

But makes sense that `Content_Contained` only overrides `Content` but not `Content__MyBagItem`.

---

**Do we want to use the display type?**  If so do you remember what we had to do with this kind of alternate

    {fieldType}_{ctx.Shape.Metadata.DisplayType}__{displayMode}{DisplayToken}

If so we would need to add e.g

    Content_Contained__ContentType` // <= if no display type or == "Detail"
    Content_DisplayType__ContentType_Contained` // <= use the display type and _Contained at the end

    Content_Contained` // <= if no display type or == "Detail"
    Content_DisplayType_Contained` // <= use the display type and _Contained at the end

Because only the term after the 1st `_` is moved at the end of the file name and by using a `.`

**Do we want to use the stereotype?** Do we want to use the content item id? But here always equals to 0 ;)

Maybe we would need to do what is done in `ContentItemDisplayManager` that checks / uses both **the display type and the stereotype to compute the actual shape type**. And then add the same alternates, as done in `Shape.cs` in `("Content").OnDisplaying()`, but with the `Contained` display mode.

---

Maybe we miss a display mode metadata property, so that the right alternates could be added in the right order by the existing `Content` shape. Maybe **we could check if the contentitem id** is equal to 0, hmm not a good idea.

But to keep it simple, maybe **just only add these 2 alternates and in the right order**, hmm but the display type may be important, and also the stereotype that may implies another shape type if not null, will see tomorrow.

The `.OnProcessing()` event is a good idea because it executes after the `("Content").OnDisplaying()` so we can see all the normal alternates which is not the case before in the bagpart view. But it doesn't solve yet the order issue because we can't insert alternates, just add them at the end if they don't already exist.

So without checking the stereotype and the display type, with what we have i would suggest

    @using OrchardCore.DisplayManagement.Shapes;
    ...

        var alternates = new AlternatesCollection("Content_Contained");
        alternates.AddRange(itemContent.Metadata.Alternates);
        alternates.Add($"Content_Contained__{item.ContentItem.ContentType}");

        itemContent.Metadata.Alternates.Clear();
        itemContent.Metadata.Alternates.AddRange(alternates);

Which results with the following alternates, looks good

    "Content_Contained"
    "Content__MyBagItem"
    "Content_Detail"
    "Content__0"
    "Content_Detail__MyBagItem"
    "Content_Detail__0"
    "Content_Contained__MyBagItem"

Okay just tried by specifying `Summary` in the bag part settings, which results with

    "Content_Contained" // here we would need "Content_Summary_Contained"
    "Content_Summary__BagItem" // <= only the normal alternate use the display type
    "Content_Contained__BagItem" // here we would need "Content_Summary__BagItem_Contained"

And tried by specifying the `RawHtml` widget type for the contained type, which results with


    "Content_Contained" // here we would need "Widget_Contained"
    "Widget__RawHtml" // <= only the normal alternate use the stereotype
    "Content_Contained__RawHtml" // here we would need "Widget_Contained__RawHtml"

So as said above, to be consistent maybe we would have to check the display type and the stereotype

deanmarcussen: Thank you @jtkech you have both cleared my question up, and given me another hint about what the actual issue is.

> **Do we want to use the display type?**

What I realise is that the issue we are trying to fix here is a regression from a previous change to fix another issue. The original issue was https://github.com/OrchardCMS/OrchardCore/issues/4372 and the fix went in quickly just before RC1 https://github.com/OrchardCMS/OrchardCore/pull/4376/files

I think the fix we put in has created two issues, i.e. the wrong template picked up by `TheAgencyTheme` and the `PageTitle` being messed up.

By changing the default display type to `Detail` we are now, by default, treating items that are contained as it they are standalone (Detail) content items. When actually they are not `Detail` items, they are `Summary` items.

By treating them as `Detail` items the `Content.liquid` template in `TheAgencyTheme` is being picked up as the default template.

Because of this the title is being rendered in the `Header` zone as defined by that template.

In addition to that the `ContentsDriver` is rendering the `ContentsMetadata` shape for the `Detail` display type. This causes the issue that @microposmp has fixed by changing the `PageTitleBuilder`

If the default display type for the `BagPart` is changed back to `Summary` then all of the problems go away.

The default `Content.Summary` template renders the content item, the `ContentsMetadata` is not rendered, so does not impact the `PageTitle`, and there is no need for us to try and hack in some alternates at a template level, which when we discussed in triage, was seen as a last resort.

Using the display type of `Summary` (as a default) also makes sense from a `BagPart` routing point of view, as the item displayed in the page, is a `Summary` and the item when routed to is a `Detail` 

Sorry @microposmp I am not trying to sabotage your pr, just trying to figure out what is going on, and what is the best solution :)

microposmp: No problem. I learned a lot about the templating feature when working with this.

I agree to what you suggest about changing back to Summary display type.

jtkech: @deanmarcussen cool!

So we would need a Widget_Summary shape, right?

Or maybe detect that the contained item has a widget stereotype and then display it in the "Detail" display type

deanmarcussen: Yes @jtkech thanks for reminding me to think about display types! 

I think we just need a `Widget_Summary` shape (copied exactly from the existing `Widget` shape), and to change the display type to `Summary` for the `BagPart` as the default.

And then I must argue my case with @sebastienros and @ns8482e !

@microposmp Cool, I'm always learning too. Would you like to do another pr and I am happy to help you with it, or shall I just do one? Perhaps leave this open until the case has been argued ;)

microposmp: @deanmarcussen Thanks, I can do a new PR. Will let you know when it's available.

ns8482e: It's not good idea hard-code  alternate inside BagPart

ns8482e:
@jtkech @deanmarcussen IMO "Contained" should be implicit stereotype 

https://github.com/OrchardCMS/OrchardCore/blob/e3d75d463f35d29777b2478613eb5a3180edb204/src/OrchardCore/OrchardCore.ContentManagement.Display/ContentDisplayManager.cs#L68-L70

Line 70, will show either "Contained" or "Content" depending on whether Content Item is inside BagPart or not ( After  PR #5499)

```
var actualShapeType = stereotype ?? "Content";
if (!updater.IsContentUpdater)
{
      actualShapeType = "Contained";
}  
```

jtkech:
@ns8482e not easy to remember all details here ;)

So in #5499 you check the updater `IsContentUpdater` to display or not some action buttons, and it is related to all collection parts as FlowPart, BagPart, ListPart.

Here it is more specific to BagPart on the front end and where each item would add an header to the page, this because ListPart can't be configured to render their items in the `Detail` diplay type, and widgets in a flow part have a different stereotype and then don't use the Content templates.

Hmm, but i agree we could have allowed to configure a ListPart to display its items in the `Detail` display type, and we can do it for a future collection part.

But here my feeling is that, in place of an implicit stereotype acting on the actual shape type, it is more like a "display mode" so that if no template is defined **we fallback to the regular Content templates**.

ns8482e:
> implicit stereotype acting on the actual shape type,

I meant If Content type is Primary then it will use Content.cshtml and related Alternates defined. However if Content type is child content then ContainedContent.cshtml take precedence over Content.cshtml. So Ideal place to define alternate is ShapeTableProvider of OrchardCore.Content not to hard-code in BagPart shape.

In #5499 `IsContentUpdater ` it's not just for Buttons, It's flag that identifies whether BuildDisplay or BuildEditor is called from Controller or Called from within parent content for child contents ( e.g. BagPart or FlowPart). 

The original issue reported for Title on front-end which is same as multiple buttons for Edit in Admin ( In both issues #5700 and #5149 renders specific shape of child items on Zone which is not intended ). So IMO #5499 is needed for #5700 too

ns8482e:
if #5499 is in-place we could fix this without having added contained alternate just using if/else in content shape

jtkech:
Okay i understand

> then ContainedContent.cshtml take precedence

But as i understand when you do

    var actualShapeType = stereotype ?? "Content";
    if (!updater.IsContentUpdater)
    {
        actualShapeType = "Contained";
    }

The templates will be instead `Contained.cshtml` and `Contained-{ContentType}.cshtml` and so on in `Detail` display type. So if someone already have a `Content-MyBagItemType.cshtml` but not a `Contained-MyBagItemType.cshtml`, it would no be taken into account.

> So Ideal place to define alternate is ShapeTableProvider of OrchardCore.Content not to hard-code in BagPart shape.

But before you said: `should be implicit stereotype` checked in `ContentDisplayManager`.

So here you may be right if we can check your flag, and maybe you meant `Content.Contained.cshtml` in place of `ContainedContent.cshtml`. And then add the alternates in the right order so that if someone already have a `Content-MyBagItemType.cshtml`, as mentioned by @deanmarcussen,  it would need to still have precedence on `Content.Contained.cshtml`, but not on `Content-MyBagItemType.Contained.cshtml`.

I'm lost with template names ;)

So if i now understand correctly, you are suggesting what we were doing here at some point, with the above alternates ordering concern, but not in the `BagPart.cshtml`, rather in the `ShapeTableProvider` where we could check your flag.

Through another PR #5933, the idea is to display this kind of contained item in the `Summary` display type by default. Note: Hmm, but knowing that we can still select `Detail` in the BagPart settings.

But i understand your idea ;) And you could add some clarifications in #5933.

ns8482e:
> But before you said: `should be implicit stereotype` checked in `ContentDisplayManager`.

Sorry I confused you by referring code at  ContentDisplayManager that creates alternates but rather the actual place is Shape.cs in OrchardCore.Contents. 

What I meant by implicit stereotype ( may be I chose wrong word) is that adding an alternate that start with just `Contained` and not `Content_Contained`.

So If default Shape  and Alternates are as following

```
Content
Content__Page
Content_Detail
Content__3267
Content_Detail__Page
Content_Detail__3267
```

Then, when ContentType is an child item, the `IsContentUpdater ` is false,  then it will add alternates as following.


```
Content
Contained
Content__Page
Contained__Page
Content_Detail
Contained_Detail
Content__3267
Contained__3267
Content_Detail__Page
Contained_Detail__Page
Content_Detail__3267
Contained_Detail__3267
```

jtkech:
Okay it's clearer thanks

But above in `Shape.cs` how do you get the related updater to check `updater.IsContentUpdater`?

Note: Hmm, as i remember, sometimes we call build display methods by passing a null updater

ns8482e:
> But above in `Shape.cs` how do you get the related updater to check `updater.IsContentUpdater`?

Yes that is available in `BuildDisplayContext` but not in `ShapeDisplayContext` we need to figure out. 
May be we should add Contained = true/flase in ShapeMetadata property?

jtkech:
Yes make sense but @sebastienros doesn't like to add new shape metadata ;)

ns8482e:
> Yes make sense but @sebastienros doesn't like to add new shape metadata ;)

Another option is, In `ContentDisplayManager.cs`
```
if (!updater.IsContentUpdater)
{
      metadata.Alternates.Add("Contained");
}
```

and in `Shape.cs`

```
// Content__[DisplayType] e.g. Content-Summary
displaying.Shape.Metadata.Alternates.Add("Content_" + EncodeAlternateElement(displaying.Shape.Metadata.DisplayType));

if (displaying.Shape.Metadata.Alternates.Contains("Contained"))
{
      displaying.Shape.Metadata.Alternates.Add("Contained_" + EncodeAlternateElement(displaying.Shape.Metadata.DisplayType)
}
```

ns8482e:
@jtkech suggested changes  https://github.com/OrchardCMS/OrchardCore/compare/dev...surevelox:ns8482e/%235700

jtkech:
Okay i see the concept

Hmm, maybe in place of an `update.IsContentUpdater`, use an `updater.IsContained` to be consistent with what we would do for editing and for rendering in the front end.

But as i remember there are some places where we call `BuildDisplayAsync` with a null updater, some places where we resolve `IUpdateModelAccessor` before and here i think we would get the controller updater, or a `NullModelUpdater` if in a bg task or through graphQL. So even if it works for editing, not sure it would work in any rendering scenarios.

Currently we can use only one display type, and we use some tricks to use display mode as `.Edit`, `.Options`, `.Display`, and another trick that allows e.g `TextField-Header.Display.Summary` ;)

Maybe we would need a more complete management display type and / or display mode and / or display options, where templates are automatically enlisted with automatic fallbacks and so on.

Maybe display options would not need additional templates and we would let the templates checking the related state. E.g when editing that would be the action buttons templates that would need to render themselves or not

Hmm, maybe the best would be through shape metadata as you suggested, @sebastienros doesn't like to add new properties, but conversely maybe good to add any other metadata property type through e.g `SetMetadata<SomeType>()` and `GetMetadata<SomeType>()` as aspnetcore does when there are many scenarios and data need to be shared.

So we could have a `DisplayOptions` class / struct with additional display infos. And these are the parent collections that would need to set these metadata to its child shapes. Maybe through the `BuildDisplay` methods e.g allowing to pass a delegate to act on the shape e.g to add a metadata.

Or, just for the cases related to displaying, maybe just allow to pass to the `BuildDisplay` methods a more complex `DisplayOptions` e.g including the display type, in place of just a display type string.

But i'm lost, would need more time to see it in details ;)

deanmarcussen:
I think the other solution @microposmp did on https://github.com/OrchardCMS/OrchardCore/pull/5933 was probably the better option to solve the issue.

Does anyone disagree?

microposmp:
Yes, we did the change in #5933. For me it's ok to close this PR.

PR #5918, opened 2020-04-12
Make config keys compatible with Linux OS

### LINKED ISSUE #3766: Azure Blob Storage AppSettings Configurations
I've been trying to configure the Azure Media Storage module and from what I could see in the source code these are the app settings that I should define:

````
[
  {
    "name": "Modules__OrchardCore.DataProtection.Azure__ConnectionString",
    "value": "***"
  },
  {
    "name": "Modules__OrchardCore.DataProtection.Azure__ContainerName",
    "value": "***"
  },
  {
    "name": "Modules__OrchardCore.Media.Azure__ConnectionString",
    "value": ***"
  },
  {
    "name": "Modules__OrchardCore.Media.Azure__ContainerName",
    "value": "***"
  }
]
````

My issue here is that apparently azure (at least web apps on linux) can't read these keys. I have had this problem before when using the '.' character in the app settings keys. Is this the correct configuration? If so, is there any way I can add these entries programmatically?

**Edit:**
The error I'm getting is "Azure Media Storage is enabled but not active because some settings are missing or invalid. Check the error log for details, then correct the settings and restart the application.

````
2019-06-05T18:24:33.897312326Z [41m[1m[37mcrit[39m[22m[49m: OrchardCore.DataProtection.Azure.Startup[0]
2019-06-05T18:24:33.907626096Z       No connection string was supplied for OrchardCore.DataProtection.Azure. Ensure that an application setting containing a valid Azure Storage connection string is available at `Modules:OrchardCore.DataProtection.Azure:ConnectionString`.
2019-06-05T18:24:33.919774979Z [41m[30mfail[39m[22m[49m: OrchardCore.Media.Azure.Startup[0]
2019-06-05T18:24:33.919802479Z       Azure Media Storage is enabled but not active because ConnectionString is missing or empty in application configuration.
2019-06-05T18:24:33.941332026Z [41m[30mfail[39m[22m[49m: OrchardCore.Media.Azure.Startup[0]
2019-06-05T18:24:33.976649867Z       Azure Media Storage is enabled but not active because ContainerName is missing or empty in application configuration.
````

### PR DESCRIPTION
**Pull request by sfmskywalker**
Fixes #3766

### CODE REVIEW

**Comment by sfmskywalker** [comment]
I just realized there might be more modules using the dot-syntax for retrieving their settings - I will update them too as I find them.

**Comment by deanmarcussen** [comment]
Seems like a good plan to me. 

Smallest possible change, and points for the backwards compatibility ;)

There are a lot of modules using this dot-syntax

File: src/OrchardCore.Modules/OrchardCore.Admin/Startup.cs [hunk 1]
```diff
@@ -54,7 +54,11 @@ public override void ConfigureServices(IServiceCollection services)
             services.AddScoped<IPermissionProvider, PermissionsAdminSettings>();
             services.AddScoped<INavigationProvider, AdminMenu>();
 
-            services.Configure<AdminOptions>(_configuration.GetSection("OrchardCore.Admin"));
+            var adminSection =
+                _configuration.GetSection("OrchardCore_Admin") ??
```
jtkech, line 58: A returned section is never null but may have no value and no children

This can be checked by using `section.Exists()`

src/OrchardCore.Modules/OrchardCore.Media.Azure/MediaBlobStorageOptionsConfiguration.cs
```diff
@@ -27,7 +27,9 @@ ILogger<MediaBlobStorageOptionsConfiguration> logger
 
         public void Configure(MediaBlobStorageOptions options)
         {
-            var section = _shellConfiguration.GetSection("OrchardCore.Media.Azure");
+            var section =
+                _shellConfiguration.GetSection("OrchardCore_Media_Azure") ??
```
jtkech, line 31: Idem

src/OrchardCore.Modules/OrchardCore.Media/Services/MediaOptionsConfiguration.cs
```diff
@@ -60,7 +60,9 @@ public MediaOptionsConfiguration(IShellConfiguration shellConfiguration)
 
         public void Configure(MediaOptions options)
         {
-            var section = _shellConfiguration.GetSection("OrchardCore.Media");
+            var section =
+                _shellConfiguration.GetSection("OrchardCore_Media") ??
```
jtkech, line 64: Idem

jtkech:
LGTM

Just saw many of them in the `OrchardCore.Apis.GraphQL` startup

Hmm, in this one, in place of using multiple times `.GetValue<>()`, better to get the section once `.GetSection("OrchardCore.Apis.GraphQL")`. But in that case we would need to also check if `section.Exists()` because a returned section is never null but may have no value and no children, so maybe not the simplest way.

But as you want, can be part of another PR ;)

sfmskywalker:
@deanmarcussen Indeed :)

On a slightly different note: I was wondering if instead of doing the single underscore, would it perhaps be a little bit more elegant to treat each segment as an actual section perhaps?

For example, `"OrchardCore:OrchardCore_Media_Azure:ConnectionString"` would become `"OrchardCore:OrchardCore:Media:Azure:ConnectionString"`, which as an ENV var would become:

`"OrchardCore__OrchardCore__Media__Azure__ConnectionString``, and in JSON it would become:

```json
{
   "OrchardCore": {
      "OrchardCore": {
         "Media": {
            "Azure": {
               "ConnectionString": ""
            }
         }
      }
   }
}
```

Obviously the double "OrchardCore" element isn't as elegant. I do get that we want a section representing the module using its full name, but maybe if we treated the module's conventional "OrchardCore" prefix as implicit, and imagine we have settings for both "OrchardCore.Media" and "OrchardCore.Media.Azure", then the following JSON might be considered a little bit nicer (combining settings between OrchardCore.Media and OrchardCore.Media.Azure, both having a shared section):

```json
{
   "OrchardCore": {
      "Media": {
         "AllowedFileExtensions": [".jpg", ".png"],
         "MaxFileSize": 4096
         "Azure": {
            "ConnectionString": ""
         }
      }
   }
}
```

Notice that all media related settings now fall under "Media", and more specific ones under "Azure".

```json
{
   "OrchardCore": {
      "OrchardCore_Media": {
         "AllowedFileExtensions": [".jpg", ".png"],
         "MaxFileSize": 4096
      },
      "OrchardCore_Media_Azure": {
          "ConnectionString": ""
       }
   }
}
```
Although these field names are less attractive IMO, a real benefit is that it's immediately clear what settings belong to which modules.

I don't know. Maybe nesting these settings and mixing multiple modules like that is just really bad practice. Just thinking out loud really.

sfmskywalker:
@jtkech Good feedback, thanks. I will double-check on the section thing and the module you mentioned. I'm happy to make all of these changes as part of this PR so we get as much fixed related to these setting names with periods as possible.

src/OrchardCore.Modules/OrchardCore.Admin/Startup.cs
sfmskywalker: Good catch. I'll double-check this everywhere.

sfmskywalker:
@jtkech So the following would work as expected:

```csharp
var adminSection = _configuration.GetSection("OrchardCore_Admin");
adminSection = adminSection.Exists() ? adminSection : _configuration.GetSection("OrchardCore.Admin"); // For backward compatibility.
```

It's a little bit much because I would need to do this everywhere. Normally I would create a reusable extension method to be less repetitive, but I'm a little bit hesitant to do so, since this pattern is not something to be continued with new code. On the other hand, it would make the code more readable if it could be like e.g. so:

```csharp
var adminSection = _configuration.GetSection("OrchardCore_Admin", "OrchardCore.Admin");
```

The extension of course would look like this:

```csharp
public static IConfigurationSection GetSection(this IConfiguration configuration, string key, string compatibilityKey)
{
   var section = configuration.GetSection(key);
   return section.Exists() ? section : configuration.GetSection(compatibilityKey);
}
```

I'm leaning towards adding this extension method, but I'm open to suggestions.

sfmskywalker:
@deanmarcussen Unrelated to this PR, but something I noticed while working on it, looking at [this line](https://github.com/OrchardCMS/OrchardCore/pull/5918/files#diff-3f6987219074ad8bec0708010c86dda0L55):

```csharp
options.ContainerName.Replace("\r", String.Empty).Replace("\n", String.Empty);
```

Unless I'm misunderstanding its purpose, but from the looks of it this does nothing, since the `Replace` method does not have any side effects - instead, the result is returned to the caller, which in this case is discarded. Should it be as follows instead?

```csharp
options.ContainerName = options.ContainerName.Replace("\r", String.Empty).Replace("\n", String.Empty);
```

Same question for [this line](https://github.com/OrchardCMS/OrchardCore/pull/5918/files#diff-3f6987219074ad8bec0708010c86dda0L71).

jtkech:
Yes i was also thinking about extensions ;)

I thought about a table initialized on startup so that e.g a module could have a fallback name. But this is not only related to module names, so having a `compatibilityKey` parameter looks good.

We also have sections that are not a module name, e.g `OrchardCore.Shells.Azure` and `OrchardCore.Shells.Database` that we use in `DatabaseShellConfigurationSources`.

            _options = configuration
                .GetSection("OrchardCore")
                .GetSection("OrchardCore.Shells.Database")
                .Get<DatabaseShellsStorageOptions>()
                ?? new DatabaseShellsStorageOptions();

Note: Here we also get a section on "OrchardCore" as it is an app level service.

Maybe an extension that fallbacks automatically by replacing the `.` to `_`.
Also maybe an extension for the `GetValue<>()`

Not so simple, I'm thinking about it ;)

sfmskywalker:
I considered doing `GetValue<>` too, but it's a bit tricky because the type can be nullable or non-nullable, and there doesn't seem to be a method to see if a key exists on an `IConfiguration` other than trying to retrieve a value and checking if it's null (which is awkward for e.g. primitives such as `int`).

jtkech:
Hmm, to make it working at the tenant level through the `IShellConfiguration`

Maybe we could just update our `ShellConfiguration` class wich implement `IConfiguration`

Here the part of the code in `ShellConfiguration` that could be updated to fallback automatically to a key where we replace all `.` with `_`.


        public string this[string key]
        {
            get => Configuration[key];
            set
            {
                EnsureConfiguration();
                _updatableData.Set(key, value);
            }
        }

        public IConfigurationSection GetSection(string key)
        {
            return Configuration.GetSection(key);
        }

sfmskywalker:
@jtkech Regarding this:

```csharp
_options = configuration
            .GetSection("OrchardCore")
            .GetSection("OrchardCore.Shells.Database")
            .Get<DatabaseShellsStorageOptions>()
            ?? new DatabaseShellsStorageOptions();
```

The following should work, right?

```csharp
_options = configuration
            .GetSection("OrchardCore")
            .GetSection("OrchardCore_Shells_Database", "OrchardCore.Shells.Database")
            .Get<DatabaseShellsStorageOptions>()
            ?? new DatabaseShellsStorageOptions();
```

I also considered doing an automatic replacement of "." with "_" in case the key with "." doesn't exist, but decided it might be best to be explicit. But I'm happy to change it if someone feels strongly about that.

jtkech:
> The following should work, right?

Yes

jtkech:
> I also considered doing an automatic replacement of "." with "_"

Maybe some perf impact, but because it is intended to be used only on startup, maybe okay

sfmskywalker:
> Hmm, to make it working at the tenant level through the `IShellConfiguration`
> 
> Maybe we could just update our `ShellConfiguration` class wich implement `IConfiguration`
> 
> Here the part of the code in `ShellConfiguration` that could be updated to fallback automatically to a key where we replace all `.` with `_`.
> 
> ```
>     public string this[string key]
>     {
>         get => Configuration[key];
>         set
>         {
>             EnsureConfiguration();
>             _updatableData.Set(key, value);
>         }
>     }
> 
>     public IConfigurationSection GetSection(string key)
>     {
>         return Configuration.GetSection(key);
>     }
> ```

If we do the automatic replacement here (upon a miss), we would cover all modules trying to read a section from `ShellConfiguration`. I like that.

So we update the extension method I just added to do an auto-replacement, and update `ShellConfiguration.GetSection` to use that extension?

jtkech:
I think that when we use a key with a `:`

    configuration.GetValue<T>("OrchardCore.Module:ConfigValue");

Somewhere the `GetSection()` is called

Also there is the case where we use `configuration["OrchardCore.Module:ConfigValue"]`

So maybe doing the change at the `ShellConfiguration` level (updating its implementation) could fix all this cases

sfmskywalker:
Interesting. So then all we need to update is `ShellConfiguration.GetSection`?

sfmskywalker:
And then this:

```csharp
var containerName =
                _configuration.GetValue<string>("OrchardCore_DataProtection_Azure:ContainerName") ??
                _configuration.GetValue<string>("OrchardCore.DataProtection.Azure:ContainerName") ?? // For backward compatibility.
                "dataprotection";
```

Can be changed back to the way it was:

```csharp
var containerName = _configuration.GetValue<string>("OrchardCore.DataProtection.Azure:ContainerName", "dataprotection")
```

Since `_configuration` is a `ShellConfiguration`, and internally it will call `GetSection` for `"OrchardCore.DataProtection.Azure"`.

I like it!

jtkech:
> Interesting. So then all we need to update is ShellConfiguration.GetSection?

Also the indexer access `public string this[string key]`

But as you just thinking out loud ;)

Need to be tried with get section, get value, and the indexer way

jtkech:
Hmm, if the key contains a `:`, even we use the indexer access, maybe our GetSection get called, i will try it soon as i have time

jtkech:
Okay, just did some quick tests

Not fully tried with real values but seems that `GetValue<>()` triggers our `GetSection()` but not when using `_configuration["Test.Tutu:Toto"]`, so we would need to also update

        public string this[string key]
        {
            get => Configuration[key];

Here it is only intended to return a string, so we could check if it is null and then just try to fallback.

sfmskywalker:
Nice. OK so I updated `ShellConfiguration`'s indexer as well as its `GetSection` method.

* For the indexer: if no value was found for the given key, it checks to see if the key contains a `_`. If so, it replaces that with a `.`, because the configuration might still be using the "old" notation using `.`.
* For `GetSection`: same logic, but instead if checking to see if the returned section was null (which it never is), it uses its `Exists()` method.

deanmarcussen:
> @deanmarcussen Unrelated to this PR, but something I noticed while working on it, looking at [this line](https://github.com/OrchardCMS/OrchardCore/pull/5918/files#diff-3f6987219074ad8bec0708010c86dda0L55):
> 
> ```cs
> options.ContainerName.Replace("\r", String.Empty).Replace("\n", String.Empty);
> ```

Yes, well spotted, it's definitely a mistake by me ;) Easily fixed on this pr if you don't mind

deanmarcussen:
> @deanmarcussen Indeed :)
> 
> On a slightly different note: I was wondering if instead of doing the single underscore, would it perhaps be a little bit more elegant to treat each segment as an actual section perhaps?
> 
> For example, `"OrchardCore:OrchardCore_Media_Azure:ConnectionString"` would become `"OrchardCore:OrchardCore:Media:Azure:ConnectionString"`, which as an ENV var would become:
> 
> `"OrchardCore__OrchardCore__Media__Azure__ConnectionString``, and in JSON it would become:
> 
> ```json
> {
>    "OrchardCore": {
>       "OrchardCore": {
>          "Media": {
>             "Azure": {
>                "ConnectionString": ""
>             }
>          }
>       }
>    }
> }
> ```
> 
> Obviously the double "OrchardCore" element isn't as elegant. I do get that we want a section representing the module using its full name, but maybe if we treated the module's conventional "OrchardCore" prefix as implicit, and imagine we have settings for both "OrchardCore.Media" and "OrchardCore.Media.Azure", then the following JSON might be considered a little bit nicer (combining settings between OrchardCore.Media and OrchardCore.Media.Azure, both having a shared section):
> 
> ```json
> {
>    "OrchardCore": {
>       "Media": {
>          "AllowedFileExtensions": [".jpg", ".png"],
>          "MaxFileSize": 4096,
>          "Azure": {
>             "ConnectionString": ""
>          }
>       }
>    }
> }
> ```
> 
> Notice that all media related settings now fall under "Media", and more specific ones under "Azure".
> For comparison:
> 
> ```json
> {
>    "OrchardCore": {
>       "OrchardCore_Media": {
>          "AllowedFileExtensions": [".jpg", ".png"],
>          "MaxFileSize": 4096
>       },
>       "OrchardCore_Media_Azure": {
>           "ConnectionString": ""
>        }
>    }
> }
> ```
> 
> Although these field names are less attractive IMO, a real benefit is that it's immediately clear what settings belong to which modules.
> 
> I don't know. Maybe nesting these settings and mixing multiple modules like that is just really bad practice. Just thinking out loud really.

Yes, I wondered the same when I first saw the issue.

I suspect (and I didn't write the code) that when it was done with `OrchardCore.Media.Azure` it was chosen so it was both immediately clear which module the section belongs to, and to make sure the sections were disambiguous for other module authors also producing and wanting to consume `IShellConfiguration`. i.e. if another author produces a module that provides some `Media` functionality then what would their `IShellConfiguration` naming convention be? 

Regardless the extension approach looks good to me, and keeps the process of migrating simple

sfmskywalker:
@deanmarcussen Makes sense. I'll leave the current structure as-is then. Thanks for confirming the newline thing, I pushed a fix 👍

src/OrchardCore.Modules/OrchardCore.Admin/Startup.cs
```diff
@@ -45,18 +47,15 @@ public override void ConfigureServices(IServiceCollection services)
...
-
             services.AddScoped<IDisplayDriver<ISite>, AdminSiteSettingsDisplayDriver>();
             services.AddScoped<IPermissionProvider, PermissionsAdminSettings>();
             services.AddScoped<INavigationProvider, AdminMenu>();
-
-            services.Configure<AdminOptions>(_configuration.GetSection("OrchardCore.Admin"));
-
             services.AddSingleton<IPageRouteModelProvider, AdminPageRouteModelProvider>();
+
+            services.Configure<AdminOptions>(_configuration.GetSection("OrchardCore_Admin"));
```
jtkech, line 58: So if `_` is the new convention we would need to update the docs and appsettings samples

But as you want, maybe part of another doc PR ;) Just as a reminder

And we fallback to the `.` so for perf people would have to update their appsettings

**OR maybe**

Kepp the `.` convention and then just update the doc to say that under linux it is better to use a `_` separator

src/OrchardCore/OrchardCore.Shells.Azure/Extensions/BlobShellsOrchardCoreBuilderExtensions.cs
```diff
@@ -27,7 +28,7 @@ public static OrchardCoreBuilder AddAzureShellsConfiguration(this OrchardCoreBui
             services.AddSingleton<IShellsFileStore>(sp =>
             {
                 var configuration = sp.GetRequiredService<IConfiguration>();
-                var blobOptions = configuration.GetSection("OrchardCore:OrchardCore.Shells.Azure").Get<BlobShellStorageOptions>();
+                var blobOptions = configuration.GetSectionCompat("OrchardCore:OrchardCore_Shells_Azure").Get<BlobShellStorageOptions>();
```
jtkech, line 31: We need the same in `DatabaseShellConfigurationSources.cs` and in `DatabaseShellsSettingsSources`

src/OrchardCore.Modules/OrchardCore.Admin/Startup.cs
sfmskywalker: I found the `docs` folder as part of this repo - will make changes there as well 👍 .

jtkech:
Just updated the samples that are commented out in appsettings.json

deanmarcussen:
Looking good, but will you do the GraphQL ones as well?

I think they are the last of them :)

sfmskywalker:
Thanks! Certainly, good find 👍

PR #5919, opened 2020-04-12
Add icons to Configuration Settings menus

agriffard:
image


agriffard:
Users and Roles icons:

image

PR #5920, opened 2020-04-12
Admin Settings

agriffard:
Allow to customize the logo and favicon url and add some code into the head of admin pages.

image

People could for example insert a top banner by using js code.

Tell me if it is useful or if can cause a security problem.


hishamco:
IMHO there are few things we could add to settings:

- Logo
- Favorite Icon

injecting styles it might useful, but I prefer to be set from within the views

hishamco:
How the setting will look like after your latest modifications?

agriffard:
> How the setting will look like after your latest modifications?

Look at the firt co!!ent, I updated the screenshot.

hishamco:
Nice!! just little tiny comment, it would be nice if we use a media picker instead? What do you think?

agriffard:
Added media pickers:

image

but I don't understand why the selected image is not reloaded after a save while the path is well stored in db.

@Skrypt Can you please tell me how to finalize it and check if the way I deduct the paths property is correctly implemented ?

Skrypt:
There's a Liquid and Razor helper for the resolving the assets path correctly. `@Orchard.AssetUrl(DbPath)` or `{{ dbPath | asset_url }}`
I think I saw that you append the "/media" path somewhere you might need to remove that. The Path is relative to the current tenant.

src/OrchardCore.Modules/OrchardCore.Admin/Admin/AdminSettings.cs
```diff
@@ -1,7 +1,56 @@
...
+
+        public string BrandImage
+        {
+            get
+            {
+                string path = null;
+
+                if (!string.IsNullOrEmpty(BrandImageUrl))
+                {
+                    var media = JArray.Parse(BrandImageUrl);
```
Skrypt, line 22: I don't like the fact that you parse the BrandImageUrl here and make use of 2 different string vars to return the proper unique value from the json set in BrandImageUrl. You should parse this JSON directly in the Driver instead.

agriffard:
@Skrypt Ok, done. 
Now, do you know how the paths can be reloaded after saving and the thumbnails displayed?

src/OrchardCore.Modules/OrchardCore.Admin/Views/AdminSettings.Edit.cshtml
```diff
@@ -1,9 +1,144 @@
...
-<fieldset class="form-group" asp-validation-class-for="Zones">
+<script asp-src="~/OrchardCore.Media/Scripts/media.js" asp-name="media" at="Foot" depends-on="admin, vuejs, sortable, vuedraggable"></script>
+<style asp-src="~/OrchardCore.Media/Styles/media.min.css" debug-src="~/OrchardCore.Media/Styles/media.css"></style>
+
+<script at="Foot">
+    initializeMediaApplication(false,  '@Url.Action("MediaApplication", "Admin", new { area = "OrchardCore.Media" })');
+    initializeMediaField(
+        document.getElementById('@Html.IdFor(m => m.BrandImage)'),
+        document.getElementById('@Html.IdFor(m => m.BrandImage)-ModalBody'),
+        '@Url.Action("GetMediaItem", "Admin", new { area = "OrchardCore.Media" })',
```
Skrypt, line 11: This method is what loads the asset(s) on the media picker

Skrypt: I'm debugging to see what you mean.

agriffard: I tried this but it doesn't work:

`<input asp-for="BrandImage" type="hidden" class="paths" :value="paths" data-init="[{ 'Paths': '@Model.BrandImage' }]" />`

Skrypt: Ok there is a bug in resolving the assets in the media library.

Skrypt: I have the same issue here. I'm trying to find why `:value="paths"` doesn't work. If we test with a Field it makes a GET call to `GetMediaItem?path=yourfile.jpg` but it never calls it on this admin settings view. We pass now the proper JSON to the component in the `data-init` html attribute so if it's not working there's something happening in the Vue.js component.

agriffard:
Thank you for your changes @Skrypt .
Do you know why it is working for the MediaField and not here?

Skrypt:
You can revert your last commit it's not necessary with what I did in the driver.
I'm not sure why it's not working. Probably 2 different bugs that needs to be fixed. 
First the media library is not listing the images that are uploaded to the /media folder directly.
And, second issue, the `:value="paths"` never gets any data from the Vue.js component.
I'm going to try to find why later on.

src/OrchardCore.Themes/TheAdmin/Assets/scss/TheAdmin.scss
```diff
@@ -90,16 +90,15 @@ body {
...
-    background-image: url(../logo.png);
-    background-repeat: no-repeat;
-    background-position: left center;
-    background-size: 28px auto;
     white-space: nowrap;
     font-size: 14px;
     transition: width 0.2s ease-in-out;
 
+    img {
+        width: 28px;
```
Skrypt, line 99: What was the intent here? We should probably make this less global and use a class selector to wrap this.

orchardc:
@agriffard , this setting doesn't exisit any more in the latest build.

Can you suggest??

image

agriffard:
This PR has never been merged. 
You can use Custom Settings to do this and use them in your own theme.

orchardc:
Could you provide an example? I have a custom theme but no idea how to replace those icon. 
It's sad to see the feature hadn't been merged. :(

agriffard:
https://github.com/agriffard/TheAdminTheme
It has not been updated since RC2.

orchardc:
Thanks @agriffard , are you refering to TheAdminTheme/src/TheAdminTheme/Views/Layout.cshtml 

` @string logoUrl = (adminSettings?.Logo?.Paths != null && adminSettings?.Logo?.Paths.Count > 0) ? Orchard.AssetUrl((string)adminSettings?.Logo?.Paths[0], 32) : "/TheAdminTheme/logo.png";`

and then

`  <style>.ta-navbar-brand {background-image: url(@logoUrl) !important;}</style>`


I tried the the above code with replace 'TheAdminTheme' to 'MyTheme', however the admin logo isn't replaced.

orchardc:
In my site, it is still poit to the **\TheAdmin\logo.png**,  where does that picture stored?? I couldn't find it from folders or database.

image

agriffard:
The files are in TheAdmin\wwwroot.
You don't seem to have changed the admin theme. 
It has to be done in the database and restarted.

orchardc:
Any idea which table is it? Maybe I can just manually replace logo.png with my site's logo.png

Still I'm not sure how can I use the custom setting in my theme to replace the admin logo...

jtkech:
@orchardc as a side note the right usage is to take into account the PathBase to be fully tenant aware, and that may contain a virtual folder, e.g. `Url.Content("~/TheAdminTheme/logo.png")` with a leading tilde. I will do a PR to fix the current usage

orchardc:
@agriffard  and @jtkech , I read this article about Custom Settings, https://docs.orchardcore.net/en/dev/docs/reference/modules/CustomSettings/

But not sure how to **replace the logo on the Admin panel** with Custom Settings?
Appreciate your input.

jtkech:
@orchardc 

I think you got the idea with @Skrypt to make an admin theme based on `TheAdmin`

About Custom Setting, show the video in the doc, as i remember you need to create a dedicated content type with the stereotype set to `CustomSettings`, then you can edit these settings through the admin, and finally you can access them through the `Site.Properties` object e.g. through liquid

If you generate a logo url from a setting, remember to take into account the `PathBase` as done in our image / asset liquid filters or our `Orchard` razor helpers. If in a razor view you can also use an `UrlHelper` extension, e.g. `Url.Content(~/path)`, but it depends on the source of your logo, e.g. if you manage it as a media file, because the media file store may have different storage, you would need to use somewhere `IMediaFileStore`, there is a liquid filter to generate an public image url.

orchardc:
Thanks @jtkech solved now with a new theme.

PR #5927, opened 2020-04-13
Implement IQueryableUserStore<IUser> in UserStore

### LINKED ISSUE #5916: IQueryableUserStore
Hi,
when I try to get all users :
var users = _userManager.Users;
I get this error
NotSupportedException: Store does not implement IQueryableUserStore.
get_Users
thank you

hishamco:
Fixes #5916


src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
```diff
@@ -524,5 +534,17 @@ public async Task<IList<IUser>> GetUsersForClaimAsync(Claim claim, CancellationT
...
 
         #endregion IUserClaimStore<IUser>
+
+        #region IQueryableuserStore
+        public IQueryable<IUser> Users => GetUsersAsync().GetAwaiter().GetResult().Roles.AsQueryable();
+        #endregion
+
+        private async Task<UsersDocument> GetUsersAsync()
+        {
+            return await _scopedDistributedCache.GetOrSetAsync(() =>
```
sebastienros, line 544: don't cache that, it's already bad to call it as it's an unbounded query. what if there are 1 million or more users?

src/OrchardCore/OrchardCore.Users.Abstractions/Services/IUserService.cs
```diff
@@ -60,6 +61,8 @@ public interface IUserService
         /// <returns>The <see cref="IUser"/> represents the retrieved user.</returns>
         Task<IUser> GetUserAsync(string userName);
 
+        Task<IEnumerable<IUser>> GetUsersAsync();
```
sebastienros, line 64: I would not expose it in the service. It's a bad method (unbounded).

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
hishamco: What about `Roles`? I saw similar behavior

src/OrchardCore/OrchardCore.Users.Abstractions/Services/IUserService.cs
hishamco: I saw similar thing in `Roles`, I will try to remove this for now

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
jtkech: you may have a million of users but not a million of roles

src/OrchardCore/OrchardCore.Users.Abstractions/Services/IUserService.cs
sebastienros: I thought about it, but the probability of having that many roles is smaller

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
jtkech: In the related issue i said `Maybe we should add this`, thinking that anyway somebody can already do an unbounded query through the session, but yes maybe better to not add it explicitly to the interface.

hishamco: Ok, I just update the PR and I will need to test the admin UI to make sure everything work as expected

src/OrchardCore.Modules/OrchardCore.Users/Controllers/AdminController.cs
```diff
@@ -82,7 +82,7 @@ public async Task<ActionResult> Index(UserIndexOptions options, PagerParameters
                 options = new UserIndexOptions();
             }
 
-            var users = _session.Query<User, UserIndex>();
+            var users = _userManager.Users.Select(u => u as User);
```
deanmarcussen, line 85: I would not change this file. What you are doing here is resolving the entire list of users then reducing the list with linq for the paging. Makes no sense.

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
```diff
@@ -524,5 +526,9 @@ public async Task<IList<IUser>> GetUsersForClaimAsync(Claim claim, CancellationT
         }
 
         #endregion IUserClaimStore<IUser>
+
+        #region IQueryableuserStore
+        public IQueryable<IUser> Users => _session.Query<User>().ListAsync().GetAwaiter().GetResult().AsQueryable();
+        #endregion
```
deanmarcussen, line 532: also here should be `.Query<User,UserIndex>()`

deanmarcussen:
I agree about the unbounded nature of exposing the entire list of users like this

Better to throw an exception there, which says, make the query with `ISession`

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
hishamco: This mean I should use what I suggested earlier [here](https://github.com/OrchardCMS/OrchardCore/issues/5916#issuecomment-612624587)

jtkech: @deanmarcussen @hishamco 

Sorry, this is me that i wrongly suggested to use `_session.Query<User>()` if there is no predicate ;)

But yes, using the indexed table is still much more better as it holds all the related ids, maybe i was too tired ;)

jtkech:
**For Users**

So maybe just do nothing as in the aspnetcore `UserManager` there is


        /// <summary>
        ///     Returns an IQueryable of users if the store is an IQueryableUserStore
        /// </summary>
        public virtual IQueryable<TUser> Users
        {
            get
            {
                var queryableStore = Store as IQueryableUserStore<TUser>;
                if (queryableStore == null)
                {
                    throw new NotSupportedException(Resources.StoreNotIQueryableUserStore);
                }
                return queryableStore.Users;
            }
        }

In fact for users we don't use it as in the admin controller we use directly

    var users = _session.Query<User, UserIndex>();

---
**For Roles**

We have the same in the aspnetcore `Rolemanager` that checks if it is an `IQueryableRoleStore<TRole>`

But for roles, the admin controller use `_roleService.GetRolesAsync()` that uses `_roleManager.Roles`.

So maybe the `RoleService` could use directly the `RoleStore`

    await _roleStore.GetRolesAsync();

So that we could also remove the `IQueryableRoleStore<IRole>` in RoleStore. Okay we don't have so many roles but it's still a non async call.

    IQueryable<IRole> Roles => GetRolesAsync().GetAwaiter().GetResult().Roles.AsQueryable();

But that's okay as it seems that it is only used to list all roles in the related admin index page.

hishamco:
So, what should we do here, coz I'm confused little bit ;)

jtkech:
I suggest that for Users we do nothing

Then for Roles i just added some infos on my findings, but i think that's okay, even if the call is non async as Roles are managed through a Document that is cached.

So for me, just nothing to do.

hishamco:
> I suggest that for Users we do nothing

Is that mean nothing to do at all or nothing to do after the latest changes?

jtkech:
nothing to do at all

deanmarcussen:
concur with @jtkech

hishamco:
So need for this PR ;)

agriffard:
Closing it, according to the comments.

PR #5933, opened 2020-04-13
Render bagpart items with summary displaytype.

### LINKED ISSUE #5700: Bug? No template for bag item?

microposmp:
@deanmarcussen New PR fixes #5700 that uses summary display type instead of detail.

Added Widget_Summary template to prevent issue #4372 when changing to summary display type.
 
A different solution was started in #5904 that provides 2 new alternates:
Content_Contained
Content_Contained__[ContentType]

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Flows/Views/BagPart.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Flows/Views/BagPart.cshtml
@@ -6,7 +6,7 @@
 <section class="flow">
     @foreach (var item in Model.BagPart.ContentItems)
     {
-        var itemContent = await ContentItemDisplayManager.BuildDisplayAsync(item, Model.BuildPartDisplayContext.Updater, Model.Settings.DisplayType ?? "Detail", Model.BuildPartDisplayContext.GroupId);
+        var itemContent = await ContentItemDisplayManager.BuildDisplayAsync(item, Model.BuildPartDisplayContext.Updater, Model.Settings.DisplayType ?? "Summary", Model.BuildPartDisplayContext.GroupId);
 
         @await DisplayAsync(itemContent)
     }
--- a/src/OrchardCore.Modules/OrchardCore.Flows/Views/BagPartSettings.Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Flows/Views/BagPartSettings.Edit.cshtml
@@ -9,7 +9,7 @@
 <div class="form-group">
     <div class="row col-sm-6">
         <label asp-for="DisplayType">@T["Display Type"]</label>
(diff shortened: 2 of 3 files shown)


deanmarcussen:
Thanks @microposmp I shall make the case for it ;)

What I realise is that the issue we are trying to fix on the other pr is a regression from a previous change to fix another issue. The original issue was https://github.com/OrchardCMS/OrchardCore/issues/4372 and the fix went in quickly just before RC1 https://github.com/OrchardCMS/OrchardCore/pull/4376/files where we changed the `BagPart` default display type from `Summary` to `Detail`

I think the fix we put in has created two issues, i.e. the wrong template picked up by TheAgencyTheme and the PageTitle being messed up (i.e for each item in the bag it adds to the page title).

By changing the default display type to Detail we are currently, by default, treating items that are contained as it they are standalone (Detail) content items. When actually they are not Detail items, they are Summary items.

By treating them as Detail items the Content.liquid template in TheAgencyTheme is being picked up as the default template.

Because of this the title is being rendered in the Header zone as defined by that template.

In addition to that the ContentsDriver is rendering the ContentsMetadata shape for the Detail display type. This causes the issue that @microposmp has fixed by changing the PageTitleBuilder

If the default display type for the BagPart is changed back to Summary then all of the problems go away.

The default Content.Summary template renders the content item, the ContentsMetadata is not rendered, so does not impact the PageTitle, and there is no need for us to try and hack in some alternates at a template level, which when we discussed in triage, was seen as a last resort.

Using the display type of Summary (as a default) also makes sense from a BagPart routing point of view, as the item displayed in the page, is a Summary and the item when routed to is a Detail

@microposmp one last thing we need to do here is also update the edit display for the default in https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.Flows/Views/BagPartSettings.Edit.cshtml


/cc @sebastienros and @ns8482e as you argued for `Detail` as the display type originally

jtkech:
Hmm, not sure but maybe a mix of the 2 PRs, this one that adds the missing `Widget_Summary` template and set the display type to `Summary` by default

Then, if the current display type is `Detail`, add the 2 alternates `Content_Contained` and `Content_Contained__ContentType`, but without adding these templates in our themes.

sebastienros:
BagPart will render content items as Summary display type.
So if any content type template was customized for Detail (Content-Foo.cshtml) in bag parts, it will need t be changed to  the summary display type (Content-Foo.Summary.cshtml)

sebastienros:
Let's merge it after Tuesday so people are not surprised by the change.

PR #5934, opened 2020-04-13
Fix RSS items description

### LINKED ISSUE #5915: There is a bug in RSS feeds
There is a bug in RSS feeds, all the posts description are the same as the first post.

mmexport1586669690916

@agriffard @sebastienros @deanmarcussen

hishamco:
Fix #5915

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Feeds/CommonFeedItemBuilder.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Feeds/CommonFeedItemBuilder.cs
@@ -26,7 +26,7 @@ namespace OrchardCore.Contents.Feeds.Builders
             {
                 var contentItem = feedItem.Item;
                 var contentItemMetadata = await _contentManager.PopulateAspectAsync<ContentItemMetadata>(contentItem);
-                var body = contentItem.Content.MarkdownBodyPart.Markdown.Value;
+                var bodyAspect = await _contentManager.PopulateAspectAsync<BodyAspect>(contentItem);
                 var routes = contentItemMetadata.DisplayRouteValues;
 
                 // author is intentionally left empty as it could result in unwanted spam
@@ -49,7 +49,7 @@ namespace OrchardCore.Contents.Feeds.Builders
                     feedItem.Element.SetElementValue("title", WebUtility.HtmlEncode(contentItem.DisplayText));
                     feedItem.Element.Add(link);
 
-                    feedItem.Element.SetElementValue("description", body != null ? $"<![CDATA[{body?.ToString()}]]>" : String.Empty);
+                    feedItem.Element.SetElementValue("description", bodyAspect.Body != null ? $"<![CDATA[{bodyAspect.Body?.ToString()}]]>" : String.Empty);
 
                     if (contentItem.PublishedUtc != null)
                     {
@@ -75,7 +75,7 @@ namespace OrchardCore.Contents.Feeds.Builders
                     });
 
                     context.Builder.AddProperty(context, feedItem, "title", WebUtility.HtmlEncode(contentItem.DisplayText));
-                    context.Builder.AddProperty(context, feedItem, "description", body != null ? $"<![CDATA[{body?.ToString()}]]>" : String.Empty);
+                    context.Builder.AddProperty(context, feedItem, "description", bodyAspect.Body != null ? $"<![CDATA[{bodyAspect.Body?.ToString()}]]>" : String.Empty);
 
                     if (contentItem.PublishedUtc != null)
                     {
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Feeds/CommonFeedItemBuilder.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Feeds/CommonFeedItemBuilder.cs
@@ -26,7 +26,7 @@ namespace OrchardCore.Contents.Feeds.Builders
             {
                 var contentItem = feedItem.Item;
                 var contentItemMetadata = await _contentManager.PopulateAspectAsync<ContentItemMetadata>(contentItem);
-                var bodyAspect = await _contentManager.PopulateAspectAsync<BodyAspect>(contentItem);
+                var body = contentItem.Content.MarkdownBodyPart.Markdown.Value;
                 var routes = contentItemMetadata.DisplayRouteValues;
 
                 // author is intentionally left empty as it could result in unwanted spam
@@ -49,7 +49,7 @@ namespace OrchardCore.Contents.Feeds.Builders
                     feedItem.Element.SetElementValue("title", WebUtility.HtmlEncode(contentItem.DisplayText));
                     feedItem.Element.Add(link);
 
-                    feedItem.Element.SetElementValue("description", bodyAspect.Body != null ? $"<![CDATA[{bodyAspect.Body?.ToString()}]]>" : String.Empty);
+                    feedItem.Element.SetElementValue("description", body != null ? $"<![CDATA[{body?.ToString()}]]>" : String.Empty);
 
                     if (contentItem.PublishedUtc != null)
                     {
@@ -75,7 +75,7 @@ namespace OrchardCore.Contents.Feeds.Builders
                     });
 
                     context.Builder.AddProperty(context, feedItem, "title", WebUtility.HtmlEncode(contentItem.DisplayText));
-                    context.Builder.AddProperty(context, feedItem, "description", bodyAspect.Body != null ? $"<![CDATA[{bodyAspect.Body?.ToString()}]]>" : String.Empty);
+                    context.Builder.AddProperty(context, feedItem, "description", body != null ? $"<![CDATA[{body?.ToString()}]]>" : String.Empty);
 
                     if (contentItem.PublishedUtc != null)
                     {
--- a/src/OrchardCore.Modules/OrchardCore.Html/Handlers/HtmlBodyPartHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Html/Handlers/HtmlBodyPartHandler.cs
@@ -1,4 +1,3 @@
-using System.Collections.Generic;
 using System.Text.Encodings.Web;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Html;
@@ -30,7 +29,6 @@ namespace OrchardCore.Html.Handlers
                 if (bodyAspect != null && part.ContentItem.Id == _contentItemId)
                 {
                     bodyAspect.Body = _bodyAspect;
-                    part.ContentItem.Id = _contentItemId;
 
                     return;
                 }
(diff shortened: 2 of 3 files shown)


src/OrchardCore.Modules/OrchardCore.Contents/Feeds/CommonFeedItemBuilder.cs
```diff
@@ -26,7 +26,7 @@ public async Task PopulateAsync(FeedContext context)
             {
                 var contentItem = feedItem.Item;
                 var contentItemMetadata = await _contentManager.PopulateAspectAsync<ContentItemMetadata>(contentItem);
-                var bodyAspect = await _contentManager.PopulateAspectAsync<BodyAspect>(contentItem);
```
hishamco, line 29: I'm not sure if the `BodyAspect` works well with `MarkdownBodyPart`, because what I had notice is the `bodyAspect` contains the body of the last item all the time?!!

This fix probably not efficient because it requires `MarkdownBodyPart` on the `BlogPost` content type

@deanmarcussen any idea?

hishamco:
> I'm not sure if the BodyAspect works well with MarkdownBodyPart

Seems it should works, I need to dig into `MarkdownBodyPartHandler`

hishamco:
The issue comes from the `MarkdownBodyPartHandler`, the `_bodyAspect` caches the value after it runs for the first time. I'm not sure if this by intention

src/OrchardCore.Modules/OrchardCore.Markdown/Handlers/MarkdownBodyPartHandler.cs
```diff
@@ -25,12 +24,6 @@ public override Task GetContentItemAspectAsync(ContentItemAspectContext context,
         {
             return context.ForAsync<BodyAspect>(async bodyAspect =>
             {
-                if (_bodyAspect != null)
-                {
-                    bodyAspect.Body = _bodyAspect;
-                    return;
-                }
-
```
deanmarcussen, line 33: yes, this is cached for the request, so that we are not having to generate the markdown/liquid rendering multiple times on the same request.

What you could do is move the cached body aspect to a `Dictionary<string, string>()` keyed by `ContentItemId`

Also there are other instances of this usage, for the `HtmlBodyPartHandler` (and others, you will find) which would want to be updated as well.

hishamco: Ya, I was thinking about other handlers .. I will try to update the PR by introducing a dictionary cache as you suggested 

Thanks

hishamco:
@deanmarcussen you can :shipit: if it looks good for you

src/OrchardCore.Modules/OrchardCore.Markdown/Handlers/MarkdownBodyPartHandler.cs
```diff
@@ -25,9 +26,11 @@ public override Task GetContentItemAspectAsync(ContentItemAspectContext context,
         {
             return context.ForAsync<BodyAspect>(async bodyAspect =>
             {
-                if (_bodyAspect != null)
```
sebastienros, line 28: This is the only line to change. It's a cached value for performance reasons I assume, but apparently this can be called for different content items. So we should store the contentitemid too and check for both value before assigning it. Otherwise just recompute it.

hishamco: > but apparently this can be called for different content items

How? With old implementation once the content item is computed the others will read the old computed value, this why we introduce the dictionary

sebastienros:
```
if (_bodyAspect != null && contentItem.Id == _contentItemId) 
{
  // use cache here
}

// otherwise compute a new value and cache it FOR THIS CONTENT ITEM

_bodyAspect = ... // compute body aspect for this content item
_contentItemId = contentItem.Id; // store that the cache is for this content item only
```

sebastienros:
Or just remove the cache completely and wait for some benchmarking to show that there is a perf improvement to make here. At least run in debug mode on some pages to see if we come back to this handler many times.

hishamco:
Seems no need for introducing a dictionary as you said before, tested and it works as expected

src/OrchardCore.Modules/OrchardCore.Markdown/Handlers/MarkdownBodyPartHandler.cs
```diff
@@ -25,12 +28,17 @@ public override Task GetContentItemAspectAsync(ContentItemAspectContext context,
...
+                if (bodyAspect != null && part.ContentItem.Id == _contentItemId)
                 {
                     bodyAspect.Body = _bodyAspect;
+                    part.ContentItem.Id = _contentItemId;
+
```
deanmarcussen, line 35: why are you setting this here? You've already compared it as equal, so no need to set?

src/OrchardCore.Modules/OrchardCore.Markdown/Handlers/MarkdownBodyPartHandler.cs
```diff
@@ -13,7 +14,9 @@ public class MarkdownBodyPartHandler : ContentPartHandler<MarkdownBodyPart>
     {
         private readonly ILiquidTemplateManager _liquidTemplateManager;
         private readonly HtmlEncoder _htmlEncoder;
-        private HtmlString _bodyAspect;
+        private readonly IDictionary<string, HtmlString> _bodiesAspectDictionary = new Dictionary<string, HtmlString>();
```
deanmarcussen, line 17: the dictionary can go now right?

src/OrchardCore.Modules/OrchardCore.Html/Handlers/HtmlBodyPartHandler.cs
```diff
@@ -25,12 +27,17 @@ public override Task GetContentItemAspectAsync(ContentItemAspectContext context,
         {
             return context.ForAsync<BodyAspect>(async bodyAspect =>
             {
-                if (_bodyAspect != null)
+                if (bodyAspect != null && part.ContentItem.Id == _contentItemId)
                 {
                     bodyAspect.Body = _bodyAspect;
+                    part.ContentItem.Id = _contentItemId;
```
deanmarcussen, line 33: why are you setting this here? You've already compared it as equal, so no need to set?

src/OrchardCore.Modules/OrchardCore.Markdown/Handlers/MarkdownBodyPartHandler.cs
```diff
@@ -13,7 +14,9 @@ public class MarkdownBodyPartHandler : ContentPartHandler<MarkdownBodyPart>
     {
         private readonly ILiquidTemplateManager _liquidTemplateManager;
         private readonly HtmlEncoder _htmlEncoder;
-        private HtmlString _bodyAspect;
```
deanmarcussen, line 16: why change from `HtmlString`?

src/OrchardCore.Modules/OrchardCore.Html/Handlers/HtmlBodyPartHandler.cs
```diff
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
```
deanmarcussen, line 1: and no need for this using anymore?

src/OrchardCore.Modules/OrchardCore.Markdown/Handlers/MarkdownBodyPartHandler.cs
hishamco: Ya, seems I forgot to delete it from here

src/OrchardCore.Modules/OrchardCore.Markdown/Handlers/MarkdownBodyPartHandler.cs
hishamco: Seem I added accidentally ;)

src/OrchardCore.Modules/OrchardCore.Markdown/Handlers/MarkdownBodyPartHandler.cs
hishamco: Unable to cast `IHtmlContent` to `HtmlString`

deanmarcussen: Please change this back to `HtmlString`. When you fix where it is it being cached it will not require changing

src/OrchardCore.Modules/OrchardCore.Html/Handlers/HtmlBodyPartHandler.cs
```diff
@@ -25,12 +26,16 @@ public override Task GetContentItemAspectAsync(ContentItemAspectContext context,
...
-                if (_bodyAspect != null)
+                if (bodyAspect != null && part.ContentItem.Id == _contentItemId)
                 {
                     bodyAspect.Body = _bodyAspect;
+
                     return;
                 }
 
+                _bodyAspect = bodyAspect.Body;
+                _contentItemId = part.ContentItem.Id;
```
deanmarcussen, line 37: You are setting this cache value here, before it has even been generated. 
So cannot work

Please refer to line 51 and 55 where the bodyAspect is actually being set.

add `_contentItemId = part.ContentItem.Id;` to after the catch block, after line 56

src/OrchardCore.Modules/OrchardCore.Markdown/Handlers/MarkdownBodyPartHandler.cs
```diff
@@ -25,12 +27,16 @@ public override Task GetContentItemAspectAsync(ContentItemAspectContext context,
...
-                if (_bodyAspect != null)
+                if (bodyAspect != null && part.ContentItem.Id == _contentItemId)
                 {
                     bodyAspect.Body = _bodyAspect;
+
                     return;
                 }
 
+                _bodyAspect = bodyAspect.Body;
+                _contentItemId = part.ContentItem.Id;
```
deanmarcussen, line 38: as above for the same problem

src/OrchardCore.Modules/OrchardCore.Html/Handlers/HtmlBodyPartHandler.cs
```diff
@@ -13,7 +13,8 @@ public class HtmlBodyPartHandler : ContentPartHandler<HtmlBodyPart>
     {
         private readonly ILiquidTemplateManager _liquidTemplateManager;
         private readonly HtmlEncoder _htmlEncoder;
-        private HtmlString _bodyAspect;
+        private IHtmlContent _bodyAspect;
```
deanmarcussen, line 16: please change back to `HtmlString`

hishamco: You could try it, it will raise invalid casting issue

hishamco: Revert the change, it was a weird error ;)

PR #5938, opened 2020-04-14
Fixes #5921 shape table providers

### LINKED ISSUE #5921: ShapeTableProvider from feature not recognized
When adding an `IShapeTableProvider` as a feature the descriptor is not recognized by the `ShapeTableManager` until the site is hard reset.

I think the static  Shape Descriptor Dictionary already has an entry for the Feature, that the `IShapeTableProvider` is altering, so the new alteration is skipped on rebuild of the shape table.

Clearing the `_shapeDescriptors.Clear()` resolves the issue, but not in a good way.

For repro refer https://github.com/OrchardCMS/OrchardCore/pull/5862

jtkech:
Fixes #5921 

- So we use `excludedFeatures` to not run twice (accross tenants) the providers of a given feature. The issue is that the one that harvests template files, associate the files of a given module to all of its features (we can't tie a file to a feature). So, if one of its feature defines an `IShapeTableProvider` and is enabled afterwards, it may be already excluded and its provider not executed.

- So the fix would just be to only associate the files of a given module to its main feature (the 1st). It works and doing this i could **reduce the `ShapeTable` entries from 921 to 525**.

- But it introduces a little constraint, we can't have a feature that is not the main one (the 1st), that needs at least a template file, and that doesn't depend on the main feature meaning that the feature can be enabled without the main one enabled.

- That's okay for 90% of the module manifests but, because of the remaining 10%, here we associate the files of a given module to its main feature (the 1st) and also to all of its features that doesn't depend on it. Here we can still **reduce the `ShapeTable` entries from 921 to 638**. Note: For perf we only check the 1st level of the dependency graph.

- But this re-introduces the original issue, only for 10% of the modules and only if they define an `IShapeTableProvider`, but still a potential issue.

- So finally, here the `excludedFeatures` is no more used to filter the execution of a given shape provider. This because before the main providers (the harversters e.g of template files) was not filtered, and the other ones are much more faster. But the `excludedFeatures` are still used by the harversters themselves, this is where there is the most perf gain So there is **no significant impact, e.g only 1ms for a whole around 80ms** to build a full table. And it fixes again the original issue.

**So here we fix the issue and we reduce the `ShapeTable` entries from 921 to 638**.

Note: To reduce again the shape entries to 525, we could try to update the 10% of manifests that don't follow the new constraint. But a good example is the Templates module where we can enable the Admin Templates but not enable the main feature being the front end Templates. **Update**: Bad example as it fails if we do that, so i added the missing dependency.

Hmm, maybe i have an idea.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Manifest.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Manifest.cs
@@ -7,6 +7,7 @@ using OrchardCore.Modules.Manifest;
     Version = "2.0.0",
     Category = "Content Management"
 )]
+
 [assembly: Feature(
     Id = "OrchardCore.ContentFields",
     Name = "Content Fields",
@@ -19,5 +20,6 @@ using OrchardCore.Modules.Manifest;
     Id = "OrchardCore.ContentFields.Indexing.SQL",
     Name = "Content Fields Indexing (SQL)",
     Category = "Content Management",
-    Description = "Content Fields Indexing module adds database indexing for content fields."
+    Description = "Content Fields Indexing module adds database indexing for content fields.",
+    Dependencies = new[] { "OrchardCore.ContentFields" }
 )]
--- a/src/OrchardCore.Modules/OrchardCore.ContentLocalization/Manifest.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentLocalization/Manifest.cs
@@ -8,6 +8,7 @@ using OrchardCore.Modules.Manifest;
     Description = "Provides a part that allows to localize content items.",
     Category = "Internationalization"
 )]
+
 [assembly: Feature(
     Id = "OrchardCore.ContentLocalization",
     Name = "Content Localization",
@@ -28,6 +29,6 @@ using OrchardCore.Modules.Manifest;
     Id = "OrchardCore.ContentLocalization.Sitemaps",
     Name = "Localized Content Item Sitemaps",
     Description = "Provides support for localized content item sitemaps.",
-    Dependencies = new[] { "OrchardCore.Sitemaps" },
+    Dependencies = new[] { "OrchardCore.ContentLocalization", "OrchardCore.Sitemaps" },
     Category = "Internationalization"
 )]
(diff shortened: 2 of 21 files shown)


jtkech:
@deanmarcussen 

So i updated some manifest dependencies where it makes sense, so now we only have 3 or 4 manifest files that contains a feature that doesn't depend on the main one (the 1st one) and that uses template files. **So now we can still reduce the shape table entries from 921 to 570**.

So we could have just associated the files of a given module only to its main feature. But because of these rare cases i kept my implementation that also associate them to all features that doesn't depend on the main one.

You can try it to see if it fixes the issue in your PR #5862, just copy `ShapeTemplateBindingStrategy` and `DefaultShapeTableManager`. But because you are not in one of the rare case, you can also try it by just copying the 1st one OR just copying the 2nd one.

deanmarcussen:
Cool @jtkech I’ll try it tomorrow. Thank you

jtkech:
Updated more dependencies

E.g for `OrchardCore.Microsoft.Authentication` where there are 2  independent features with their own services and startups

But but but they share the same `IPermissionProvider` that is registered through a shared startup (not decorated). But because none of the 2 feature names match the module id, this shared startup is never executed and the authentication permissions was never registered.

So i created a 1st feature to manage authentication permissions and added a dependency on this one for the 2 other features, then i tried it to check that the shared permissions are registered.

jtkech:
I reverted the changes for `OC.Microsoft.Authentication` i removed the new 1st feature to manage permissions. In place, **i register the shared `Permissions` in the 2 specific startups but with the `TryAddEnumerable` pattern** that check the implementation type (not only the service type.) It's cleaner like this, it fixes the issue without an additional feature just for the permissions.

jtkech:
I improved the implementation

In other words, when a module is active (at least one enabled feature) its template files need to be associated at least with one feature that is currently enabled, most of the time the 1st one.

So we don't process an extension if not active or that already have an associated feature (in the excluded list) and that is still enabled. This prevents unecessary template files harvesting.

Then when we process the extension, we associate the templates only to the 1st enabled feature, most of the time the 1st one. This limit the entries due to the rare cases.

E.g if there are only 2 independent feature1 and feature2 that are enabled, we tie the templates to only feature1. This is when we disable feature1 that we also tie the templates to feature2 (if still enabled).

PR #5946, opened 2020-04-15
Remove Site.Meta

agriffard:
Discussed during meeting. 
Remove Meta from SiteSettings, it is now some Theme settings called `Head` and `Foot`.


agriffard:
Just kept the removal of Site.Meta.

Discussed with @deanmarcussen , will do a seperate PR using CustomSettings.

PR #5950, opened 2020-04-15
Returning null in case of not finding the entry in Workflow's dictionaries

siyamandayubi:
Since in Javascript, we don't have a method to check availability of an input entry or property entry, in case of not having the value in the dictionary, we got an exception. A simple solution is
returning null in case of not having the item there.


jptissot:
LGTM

deanmarcussen:
I think this makes sense, so rather than having an exception which would break the workflow, it returns null, which means in javascript you can check the response for null.
My only concern would be if it makes it harder to debug, by not throwing the exception, but workflow exceptions are not that easy to debug anyway, and there may be instances of workflows where the properties vary, so an exception is not helpful then

So LGTM

PR #5951, opened 2020-04-15
Supporting "not in (select)" in SQL Query

siyamandayubi:


sebastienros:
Why not just add the `in` operator and let the current NOT handle it?

siyamandayubi:
'In' is a little tricky, as it currently supports "in (val1,... valn)", so it needs changing that. In this way, we need a more sophisticated parser that can diffrentiate between list of values and a select statement.

On the other hand, In(select ... ) can be done using joins easily, but not in (select..) is tricky using joins

sebastienros:
I think this should be handled with another rule that will be able to dsitinguishd between 'SELECT' or a list of literals. i can try to give it a shot. We can still use your PR if I don't manage.

siyamandayubi:
Would be great to have that, I didn't try, as I tried to avoid having side effects.

sebastienros:
This PR handles the extra cases
https://github.com/OrchardCMS/OrchardCore/pull/5952

PR #5955, opened 2020-04-16
[doc] Correct text field usage in liquid example

mguma:
The property name should be "Text" and not "Value"


agriffard:
`{{ Model.ContentItem.Content.Article.MyField.Value }}`

It is just an example.
It depends on the field.

mguma:
But in this specific example it says that the MyField of the content type Article is a Text Field. Both samples (liquid and razor) should be using the same property otherwise it's confusing. 
When I was studying the example I was thinking about why the liquid syntax uses another property as the sample in the razor syntax.

PR #5969, opened 2020-04-18
Add UseCdn Picker to Setup Page

stantoxt:
The cdn is not working for all areas around the world,so making usecdn as an option is necessary.


Skrypt:
If the issue is that the CDN is not working then I'd rather add an option so that we can change the CDN source.

stantoxt:
That's right,but some special scenarios even don't have Wan,Only UseCdn Resources would cause the page load failed.

deanmarcussen:
I understand the issue regarding cdn's. It is hard to provide perfect coverage.

However I'm not sure if this needs to be a setup setting.

You can adjust the default site settings by using a setup recipe, if you don't want to engage a cdn, and in those circumstances, it could be handled by using your own recipe.

Or you can simply turn the setting off afterwards in the admin?

Or are you unable to reach the admin because of the lack of cdn?

I would be interested to understand where this issue is being experienced, is it a development setup issue / tenant setup issue?

stantoxt:
> You can adjust the default site settings by using a setup recipe, if you don't want to engage a cdn, and in those circumstances, it could be handled by using your own recipe.
That is great.

> Or are you unable to reach the admin area because of the lack of cdn?

Not all network can access the cdn,at least I can not access the cdn in shanghai.

> I would be interested to understand where this issue is being experienced, is it a development setup issue / tenant setup issue?
Some enterprise-level apps those need secured network environment,like erp,bank system,etc

deanmarcussen:
> Or are you unable to reach the admin area because of the lack of cdn?

What I was asking is if you are unable to access `jQuery` does the whole admin site break so much that you cannot access the settings to disable the cdn setting?

stantoxt:
@deanmarcussen jquery,jquery-ui and jquery-ui-i18n can not be accessed.
But I think some special scenarios only have lan,so site setup is impossible.

deanmarcussen:
I tested the issue with loading `jQuery` from a CDN.

If the CDN is not available, or you are on a LAN without internet access, the admin menu system will not work, and there is no way to get into the admin area to change the CDN value, without going into the database, or knowing the correct url.

So I change my mind, maybe we should add this option to the setup page?

@Skrypt thoughts?

Skrypt:
I need to check but it should basically work the same as what we did in the admin then.
If you are on dev, you should be able to load assets locally.
If you are on prod, you should be able to define if you want to use a CDN or not.
If you use a CDN you should be able to define which one you should use.
Though, here as these are assets that are defined in our ResourceManifests we are only using one unique CDN for most of them. So basically, if this CDN doesn't work for them then they are "screwed" to only use local assets... which defeats the purpose of having the settings in the beginning. I'm not sure, at least this PR patches the issue.

stantoxt:
> So basically, if this CDN doesn't work for them then they are "screwed" to only use local assets... which defeats the purpose of having the settings in the beginning. I'm not sure, at least this PR patches the issue.

This is great.

sebastienros:
I also think we should have this option in the setup screen _somehow_. I checked and it's not a tri-state like for minified version of scripts (base on prod/dev) and it shouldn't be as where you are located, or even more important, where your users are, doesn't rely on dev/prod. 

Not even sure what should be the default value for this option. Maybe we should make it an app setting. 

At the same time last time an issue was open for the CDNs, we found some sources that were available in china. @stantoxt do you think you could find some CDN sources that would work in  shangai for the resources that are blocked?

Should we have a script that would check all our CDN urls and their availability in specific regions? We could ask community members to run these to be sure it works fine.

But overall I am ok with this PR. I will change some texts though.

src/OrchardCore.Modules/OrchardCore.Setup/Views/Setup/Index.cshtml
```diff
@@ -153,6 +155,15 @@
             <span asp-validation-for="SiteTimeZone"></span>
             <span class="text-muted form-text small">@T["Determines the default time zone used when displaying and editing dates and times."]</span>
         </div>
+        <div class="form-group col-md-6" asp-validation-class-for="UseCdn">
+            <label asp-for="UseCdn">@T["UseCdn"]</label>
```
sebastienros, line 159:
```suggestion
            <label asp-for="UseCdn">@T["Content Delivery Network (CDN)"]</label>
```

src/OrchardCore.Modules/OrchardCore.Setup/Views/Setup/Index.cshtml
```diff
@@ -153,6 +155,15 @@
...
+            <select asp-for="UseCdn" class="form-control">
+                <option value="true" selected>@T["Enable"]</option>
+                <option value="false">@T["Disable"]</option>
```
sebastienros, line 162:
```suggestion
                <option value="true" selected>@T["Enabled — Serve common scripts from their public CDN"]</option>
                <option value="false">@T["Disabled —  Serve common scripts directly from this site"]</option>
```

src/OrchardCore.Modules/OrchardCore.Setup/Views/Setup/Index.cshtml
```diff
@@ -153,6 +155,15 @@
...
+            </select>
+            <span asp-validation-for="UseCdn"></span>
+            <span class="text-muted form-text small">@T["Determine whether use cdn."]</span>
```
sebastienros, line 165:
```suggestion
            <span class="text-muted form-text small">@T["Change this option if your users can't access public CDNs."]</span>
```

stantoxt:
> do you think you could find some CDN sources that would work in shangai for the resources that are blocked?

@sebastienros  I will search about this.

deanmarcussen:
@stantoxt can you resolve the merge conflicts and feedback?

stantoxt:
> @stantoxt can you resolve the merge conflicts and feedback?

@deanmarcussen I will deal with it later

hyzx86:
@stantoxt   亲。。求更新啊 🤣

hyzx86:
@deanmarcussen Or , could you merge this PR first? 
https://github.com/OrchardCMS/OrchardCore/pull/7164

stantoxt:
Sorry,I will update this pr this week.

hyzx86:
Great , thank you .

agriffard:
Closing, too much conflicts.
Make a new PR if needed.

hyzx86:
@stantoxt 你那边研究的怎样了？用作生产了吗？ 抽空加个群唠唠呗？ 877196442

sebastienros:
The solution, as talked on this PR https://github.com/OrchardCMS/OrchardCore/pull/7599 is to have the CDN disabled by default.

Issue #5943, opened 2020-04-14
Theme settings: Head, Foot


agriffard:
Replace the Meta SiteSetting by a ThemeSetting.
Rename it into Head and also provide a Foot property that allows to inject script in the body footer.

Properties with Liquid support and Codemirror editor.

Use them in the recipes, initialize with a meta for the favicon, pointing to a file imported in media.

PR #5972, opened 2020-04-18
Theme settings with custom settings

agriffard:
Fixes: #5943


src/OrchardCore.Themes/TheTheme/Recipes/themeSettings.recipe.json
```diff
@@ -0,0 +1,109 @@
...
+      "enable": [
+        "OrchardCore.CustomSettings"
+      ]
+    },
+    {
+      "name": "ContentDefinition",
+      "ContentTypes": [
+        {
+          "Name": "ThemeSettings",
+          "DisplayName": "Theme",
```
deanmarcussen, line 24: I would name this `TheThemeSettings`. Then it would be unique to `TheTheme`, and can contain the reference to `TheTheme/favicon.ico` and then when someone changes themes, the `TheBlogThemeSettings` would activate and apply

src/OrchardCore.Themes/TheTheme/Recipes/themeSettings.recipe.json
```diff
@@ -0,0 +1,109 @@
...
+            }
+          ]
+        }
+      ]
+    },
+    {
+      "name": "custom-settings",
+      "ThemeSettings": {
+        "ContentItemId": "40zx9155k4kbpstjb13zn6htwz",
+        "ContentItemVersionId": null,
```
deanmarcussen, line 88: could you generate with js like the other recipes do?

src/OrchardCore.Themes/TheTheme/Views/Layout.cshtml
```diff
@@ -1,9 +1,19 @@
+﻿@inject OrchardCore.Settings.ISiteService siteService
+@using OrchardCore.Entities
+@using OrchardCore.ContentManagement
+@{ 
+    var siteSettings = await siteService.GetSiteSettingsAsync();
+        var themeSettings = siteSettings.As<ContentItem>("ThemeSettings");
```
deanmarcussen, line 6: I think `SiteSettings` should be available automatically through `@Site` (comes from the base razor page)

agriffard:
@deanmarcussen Please review.

agriffard:
@deanmarcussen Can you please review?

deanmarcussen:
When we talked about it in Triage wasn't the conclusion that it would be ok for one theme (at a stretch), but not for all themes?

agriffard:
So, keep it only for TheTheme (in Razor) and remove it from TheBlog and comingsoon themes (in Liquid)?

deanmarcussen:
I would say not `TheTheme` at all - that is only used for the Saas recipe, and isn't really a `sample theme` at all. 

I'm not super convinced we need this at all (but I know you want it, I just don't know why) so a great argument from you on why we want it would be great :)

agriffard:
My arguments are:
- There will be examples in Razor and Liquid to use CustomSettings.
- It will be nice to have it out of the box, on try.orchardcore.net for example.
- People will always need a way to customize head or foot (favicon, custom script, ...) on every pages.

@sebastienros and you don't seem to agree and want to keep the themes as simple as possible so I close the PR and the issue.

PR #5979, opened 2020-04-19
Resolve potential media key conflicts

deanmarcussen:
As pointed out by @jtkech when keyed by the path, this might result in conflicts across tenants.

For media, I use PathBase + Path (not querystring)

For sitemaps, I use the sitemapId (guid), and also a `OrdinalIgnoreCase`

And I remove my old comment, that is not relevant to do (IMO).

It's something the ASP.NET Core StaticFileMiddleware does, because it listens on every route, but we validate against the `/media` path


jtkech:
Good idea to use the `sitemapId`

Then for the `MediaFileStoreResolverMiddleware` i would suggest to just use `PathBase.Value + subPathValue` to have shorter keys.

But here the middleware is a tenant singleton, so why not having a concurrent dictionary per tenant without having to prefix the keys, **this by not declaring it as static**.

src/OrchardCore.Modules/OrchardCore.Media/Services/MediaFileStoreResolverMiddleware.cs
```diff
@@ -65,12 +63,14 @@ public async Task Invoke(HttpContext context)
             // subpath.Value returns an unescaped path value, subPath returns an escaped path value.
             var subPathValue = subPath.Value;
 
+            var workerKey = context.Request.PathBase.Value + context.Request.Path.Value;
+
```
jtkech, line 67: maybe just `context.Request.PathBase.Value + subPathValue`
Or as before if we use a concurrent dictionary per tenant

The middleware is a singleton, so why not having a non static concurrent dictionary per tenant.

src/OrchardCore.Modules/OrchardCore.Sitemaps/Controllers/SitemapController.cs
```diff
@@ -65,7 +65,7 @@ public async Task<IActionResult> Index(CancellationToken cancellationToken, stri
             }
             else
             {
-                var work = await Workers.GetOrAdd(sitemap.Path, x => new Lazy<Task<Stream>>(async () =>
+                var work = await Workers.GetOrAdd(sitemapId, x => new Lazy<Task<Stream>>(async () =>
```
jtkech, line 68: You also need to use `sitemapId` line 106

Note: There is a bad formatting line 95

deanmarcussen: Done. Good spotting :)

src/OrchardCore.Modules/OrchardCore.Media/Services/MediaFileStoreResolverMiddleware.cs
deanmarcussen: No reason why not, so updated to use a per instance (i.e. tenant) concurrent dictionary. Probably better as less keys per tenant to lookup

src/OrchardCore.Modules/OrchardCore.Sitemaps/Controllers/SitemapController.cs
```diff
@@ -15,7 +15,7 @@ namespace OrchardCore.Sitemaps.Controllers
 {
     public class SitemapController : Controller
     {
-        private static readonly ConcurrentDictionary<string, Lazy<Task<Stream>>> Workers = new ConcurrentDictionary<string, Lazy<Task<Stream>>>();
+        private static readonly ConcurrentDictionary<string, Lazy<Task<Stream>>> Workers = new ConcurrentDictionary<string, Lazy<Task<Stream>>>(StringComparer.OrdinalIgnoreCase);
```
jtkech, line 18: Why using `StringComparer.OrdinalIgnoreCase` as the keys are unique identifiers.

src/OrchardCore.Modules/OrchardCore.Media/Services/MediaFileStoreResolverMiddleware.cs
```diff
@@ -14,7 +14,7 @@ namespace OrchardCore.Media.Services
     /// </summary>
     public class MediaFileStoreResolverMiddleware
     {
-        private static readonly ConcurrentDictionary<string, Lazy<Task>> Workers = new ConcurrentDictionary<string, Lazy<Task>>();
+        private readonly ConcurrentDictionary<string, Lazy<Task>> Workers = new ConcurrentDictionary<string, Lazy<Task>>();
```
jtkech, line 17: This here, where keys are paths, that i would have used `StringComparer.OrdinalIgnoreCase`

jtkech: Hmm, but here because related to concrete files (path are not only url path), maybe better to not use `OrdinalIgnoreCase` e.g if the file system is case sensitive.

But then afterwards you use `StartsWithNormalizedSegments` with `OrdinalIgnoreCase`.

src/OrchardCore.Modules/OrchardCore.Sitemaps/Controllers/SitemapController.cs
jtkech: I think you did a mistake with `_sitemapPaths` and `_sitemapIds`

E.g in `_sitemapPaths` this is the keys that are paths

I would want to fix it in my distributedCache branch but maybe a breaking change

But this is just a convention to tie the name with the keys, not the values as in AutotouteEntries.

So at least, as it is right now, in my branch this is `SitemapPaths` that would need to use `OrdinalIgnoreCase`, not `SitemapIds`. Here you may need to add `OrdinalIgnoreCase` to `_sitemapPaths`.

jtkech: Okay, tested in my distributed branch, good to not use here `OrdinalIgnoreCase` as the keys are unique ids

And good to use `OrdinalIgnoreCase` for the `SitemapPaths` dictionary

Note: First it was not working on my branch because `MessagePack` doesn't preserve the `OrdinalIgnoreCase` comparison when deserializing.

src/OrchardCore.Modules/OrchardCore.Media/Services/MediaFileStoreResolverMiddleware.cs
jtkech: Okay the `OrdinalIgnoreCase` on `StartsWithNormalizedSegments` just acts on the `_assetsRequestPath`. Then this is the `_mediaFileStoreCache` that will return something or not depending on the case sensitivity of the file system.

**So finally i think it is good to use `OrdinalIgnoreCase` for the worker dictionary. This to prevent 2 concurrent works on the same file if the current file system is not case sensitive.**

Then if the file system is case sensitive, if you have 2 requests `/sitemap1.xml` and `/Sitemap1.xml`, even there are 2 different files they will be built one after the other, which is not a problem.

jtkech:
@deanmarcussen not urgent at all just to be sure you saw my review

deanmarcussen:
@jtkech yes, thank you. I just ran out of time yesterday to open the code and check, because sitemaps is so long ago I have forgotten the details. 

I have a memory that sitemaps urls need to be lowercased and possibly not url encoded (also but do not think this is in the code, so I have to check), to be valid for google, but also that may just be a convention. So I will have to check again and see.

idem same for recipe idempotancy

jtkech:
@deanmarcussen okay no problem

About `MediaFileStoreResolverMiddleware`

- This is the _mediaFileStoreCache that will return something or not depending on the case sensitivity of the file system. So finally **i think it is good to use `OrdinalIgnoreCase` for the worker dictionary**. This to prevent 2 concurrent works on the same file if the current file system is not case sensitive.

- Then if the file system is case sensitive, if you have 2 requests `/media1.jpg` and `/Media1.jpg`, even there are 2 different files they will be built one after the other, which is not a problem, less important than the potential above issue.

About `SitemapController`

- Why using `OrdinalIgnoreCase` for the worker dictionary where the keys are unique identifiers? But see below if we re-use the sitemap paths for the keys.

- I think you did a mistake with `_sitemapPaths` and `_sitemapIds`, e.g in _sitemapPaths this is the keys that are paths, e.g in Autoroute this is `_contentItemIds` whose key are paths, but that's okay.

- Then good to use `OrdinalIgnoreCase` for the `_sitemapPaths` dictionary (as done for the related one in Autoroute) so that beyond the google convention the requested urls are not case sensitive.

- But if 2 sitemaps have the following paths `sitemap1.xml` and `Sitemap1.xml`. Because here the worker keys are sitemap ids, 2 concurrent works may act on the same file if the file system is case insensitive. Anyway, even if we don't use `OrdinalIgnoreCase` for `_sitemapPaths` and / or even if the file system is case sensitive, **nothing prevents 2 sitemaps from having the exact same path**.

- **So finally i think it is better to re-use sitemap paths for the workers dictionary keys** this by prefixing the keys with the tenant prefix. And then still use  `OrdinalIgnoreCase` for the worker dictionary as done for cached media files.

deanmarcussen:
> @deanmarcussen okay no problem
> 
> About `MediaFileStoreResolverMiddleware`

Cool, let me check (it will not be until I have time tomorrow)

I think what I am trying to do here, is stay inline with the ASP.NET Core Static File Middleware, which (and now I need to recheck ;) ) is case sensitive

So `/Image.jpg` and `/image.jpg` are considered different files, and one will 404.

**But maybe, as you say, this is dependant on the file system** 

I haven't looked, but I am not aware of a file system, that is not case senstive?

> About `SitemapController`
> 
>   **Update**: I did the change in my `distributedCache` branch so that you can see what i mean.

Will check here also to see what you mean

jtkech:
> I think what I am trying to do here, is stay inline with the ASP.NET Core Static File Middleware, which (and now I need to recheck ;) ) is case sensitive

They do `ValidatePath(context, _matchUrl, out var subPath)`
=> `Helpers.TryMatchPath(context, matchUrl, forDirectory: false, out subPath)`
=> path.StartsWithSegments(matchUrl, out subpath)

But in fact by default `OrdinalIgnoreCase` is used in `PathString.cs`

        public bool StartsWithSegments(PathString other)
        {
            return StartsWithSegments(other, StringComparison.OrdinalIgnoreCase);
        }


Then they do `TryServeStaticFile(context, contentType, subPath)`
=> `StaticFileContext.LookupFileInfo()`
=> `_fileProvider.GetFileInfo(_subPath.Value)` that relies on the file system.

So they are case insensitive, just tried with a `test.img` under wwwroot, i can get it with `/test.jpg` and `Test.jpg` coz i'm under windows. But under linux i think that with the last url `TryServeStaticFile` that relies on the file system returns false.

> I haven't looked, but I am not aware of a file system, that is not case senstive?

On windows it is not case sensitive, you can't have both `Image.jpg` and `image.jpg` files.
On linux it is case sensitive

deanmarcussen:
> On windows it is not case sensitive, you can't have both `Image.jpg` and `image.jpg` files.
> On linux it is case sensitive

Ok @jtkech I had some time to look at this, and clearly I spend too much time on my mac, and only write code on windows ;)

This reveals another problem then. Because the Azure Blob File System is case sensitive.

Serving from the cache on a windows file system, would mean that `Image.jpg` and `image.jpg` would be different files on Blob Storage, but the same file when retrieved from cache. So whichever **Image/image arrived first in the cache would be served**, regardless of the incoming request path.

Maybe we would need to hash the filename, and rewrite the request path inside the cache handler, or lowercase all files going into blob storage (that would be invasive for users though), and still not prevent uploading files to blob storage directly.

Re: Sitemaps

> Why using OrdinalIgnoreCase for the worker dictionary where the keys are unique identifiers? But see below if we re-use the sitemap paths for the keys.

As discussed on the other pr, I think OrdinalIgnoreCase is ok here, but up for debate on it :)

> I think you did a mistake with _sitemapPaths and _sitemapIds, e.g in _sitemapPaths this is the keys that are paths, e.g in Autoroute this is _contentItemIds whose key are paths, but that's okay.

(bad naming from me)

> Then good to use OrdinalIgnoreCase for the _sitemapPaths dictionary (as done for the related one in Autoroute) so that beyond the google convention the requested urls are not case sensitive.

Yes, and I think you have done this on your branch?

> But if 2 sitemaps have the following paths sitemap1.xml and Sitemap1.xml. Because here the worker keys are sitemap ids, 2 concurrent works may act on the same file if the file system is case insensitive. Anyway, even if we don't use OrdinalIgnoreCase for _sitemapPaths and / or even if the file system is case sensitive, **nothing prevents 2 sitemaps from having the exact same path.**

This is bad, maybe I lost something in one of the refactors here, there should be something to prevent it.

> So finally i think it is better to re-use sitemap paths for the workers dictionary keys this by prefixing the keys with the tenant prefix. And then still use OrdinalIgnoreCase for the worker dictionary as done for cached media files.

> Update: I did the change in my distributedCache branch so that you can see what i mean.

Ok, so looked through your distributed branch, and there are many changes there :)
Rather than creating more merge conflicts, I will revert the changes here, to sitemaps, and I will look on another pr when I have time, to address the path uniqueness

deanmarcussen:
ok @jtkech changes to sitemaps reverted in favour of the changes on your branch.

I will investigate duplicates further separately. 

And consider the mismatch between blob storage, and windows file storage.

Was really hoping to avoid having to use a hash with this. But may need to

jtkech:
@deanmarcussen 

Just realized that the unique identifiers we generate are always lower case characters, so if they can be entered manually e.g to define an url, i agree it's good to use `OrdinalIgnoreCase`. But in the `SitemapController` we always get the sitemapId that is set in the route values by the sitemap transformer. Anyway, here we will re-use paths (but tenant prefixed) for the worker keys and then with `OrdinalIgnoreCase`.

> ok @jtkech changes to sitemaps reverted in favour of the changes on your branch.

In my branch i updated `SitemapController` just to show you what i meant and to be up to date on what we want to do. But i think it's better to update it here, as i did in my branch, because your PR will get merged before mine, no problem if there are conflicts i will fix them on my branch afterwards.

> This reveals another problem then. Because the Azure Blob File System is case sensitive.

At least the app will not fail and i think it is a very rare case and we could just say in the docs that if we are using the blob that is case sensitive, and we are running on a system that is not case sensitive, we don't serve `Image.jpg` and `image.jpg` as different images.

Beyond blob files, it is the same than having an app that doesn't use azure media and that runs on a case sensitive system with 2 `Image.jpg` and `image.jpg` local files, and wanting to deploy this app to a case insensitive system?

deanmarcussen:
> In my branch i updated `SitemapController` just to show you what i meant and to be up to date on what we want to do. But i think it's better to update it here, as i did in my branch, because your PR will get merged before mine, no problem if there are conflicts i will fix them on my branch afterwards.

Ok, misunderstood there. Will bring bits from your branch over tomorrow.

Found the problem with the `Sitemap.xml` and `sitemap.xml` being possible.

```
routeExists = (await _sitemapManager.ListSitemapsAsync())
                    .Any(p => p.Path == path);
```

This time I am not using `OrdinalIgnoreCase` here, where I should be.
I will fix it here.

But also I think currently not using `OrdinalIgnoreCase` on the `SitemapEntries`. Did you already change that behaviour in your branch? 

I think I will change it here as well, as those paths should be unique (in my mind), as per `Autoroute`

> At least the app will not fail and i think it is a very rare case and we could just say in the docs that if we are using the blob that is case sensitive, and we are running on a system that is not case 

Ok, with me (no one has reported it!)

> Beyond blob files, it is the same than having an app that doesn't use azure media and that runs on a case sensitive system with 2 `Image.jpg` and `image.jpg` local files, and wanting to deploy this app to a case insensitive system?

Yes, I suspect that would create the same scenario. 
Not related to the media cache itself, but I imagine one would overwrite the other, during the import to the media folder (unchecked)

jtkech:
> But also I think currently not using OrdinalIgnoreCase on the SitemapEntries. Did you already change that behaviour in your branch?

In my branch i have a `SitemapRouteDocument` where `SitemapPaths` is a dictionary using `OrdinalIgnoreCase`, the related one in the `AutorouteDocument` is `ContentItemIds` that also uses `OrdinalIgnoreCase`, as currently in `AutorouteEntries` where `_contentItemIds` is initialized with `OrdinalIgnoreCase`.

Okay for using `OrdinalIgnoreCase` to see if `routeExists`. The related code in Autoroute is the following where we now know that through the query predicate it is not case sensitive.

    private async Task<bool> IsPathUniqueAsync(string path, AutoroutePart context)
    {
        return (await _session.QueryIndex<AutoroutePartIndex>(o =>
            o.ContentItemId != context.ContentItem.ContentItemId && o.Path == path)
            .CountAsync()) == 0;
    }

deanmarcussen:
ok @jtkech hopefully we are on track

Sorted the equality comparison, so `Sitemap.xml` and `sitemap.xml` are considered the same (for duplicates)
Controller updated, to match, and as I look at your explanation, it makes sense.

> I think you did a mistake with _sitemapPaths and _sitemapIds
E.g in _sitemapPaths this is the keys that are paths
I would want to fix it in my distributedCache branch but maybe a breaking change

I have changed this, to match the convention in `Autoroute`, so we are all on the same page. It was done the opposite, with keys, rather than values, however makes more sense for them both to be the same convention. 
I don't see this as being a major breaking change, I do not think anyone is using `SitemapEntries` directly themselves. 

I feel like we could probably use the same OrdinalIgnoreCase on both dictionaries, but I have not done this here. 
Will keep it the same as per `Autoroute` for the moment, and if in your branch you feel it's worthwhile changing then...

jtkech:
@deanmarcussen 

> I have changed this, to match the convention in Autoroute

Okay i will update my branch accordingly

> I feel like we could probably use the same OrdinalIgnoreCase on both dictionaries

As you said i suggest to keep it the same for the moment, here it seems that ids are never manually populated and for Autoroute it is like this since a long time. And then i will not have to add another deserializer handler because MessagePack doesn't preserve a non default comparison type ;)

jtkech:
@deanmarcussen 

I found a little issue in `SitemapIndexController` that you may want to update here

When you delete a sitemap index you redirect to the `Edit` action so with a null sitemapId and it fails, in my branch i updated it to redirect to the `List` action.

On my side i also swapped the sitemap dictionaries

deanmarcussen:
@jtkech 

Thank you. You see everything ;)

Done

PR #5984, opened 2020-04-19
Lucene numHits must be > 0

### LINKED ISSUE #5983: ArgumentException: numHits must be > 0; please use TotalHitCountCollector if you just need the total hit count
image

infofromca:
fix #5983

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Lucene/Services/SearchQueryService.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lucene/Services/SearchQueryService.cs
@@ -21,15 +21,16 @@ namespace OrchardCore.Lucene.Services
 
             await _luceneIndexManager.SearchAsync(indexName, searcher =>
             {
-                var collector = TopScoreDocCollector.Create(end, true);
-
-                searcher.Search(query, collector);
-                var hits = collector.GetTopDocs(start, end);
-
-                foreach (var hit in hits.ScoreDocs)
+                if (end > 0)
                 {
-                    var d = searcher.Doc(hit.Doc, IdSet);
-                    contentItemIds.Add(d.GetField("ContentItemId").GetStringValue());
+                    var collector = TopScoreDocCollector.Create(end, true);
+                    searcher.Search(query, collector);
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.Lucene/Services/SearchQueryService.cs
```diff
@@ -21,15 +21,16 @@ public async Task<IList<string>> ExecuteQueryAsync(Query query, string indexName
 
             await _luceneIndexManager.SearchAsync(indexName, searcher =>
             {
-                var collector = TopScoreDocCollector.Create(end, true);
-
-                searcher.Search(query, collector);
-                var hits = collector.GetTopDocs(start, end);
-
-                foreach (var hit in hits.ScoreDocs)
+                if (end > 0)
```
sebastienros, line 24: How do you get `end` to be `0` ?

Skrypt: When there is no index files found

Skrypt: This PR needs more work. I'm not sure this is the proper fix. It's a nice patch but that does not fix the underlying issue. If you expect to be able to search in an index there should be an index available.

Skrypt:
This PR needs more work. I'm not sure this is the proper fix. It's a nice patch but that does not fix the underlying issue. If you expect to be able to search in an index there should be an index available.

infofromca:
Ithinkwe can do 2 parts. 
1--merge this to make the code stronger.
2--fix the underlying part by your code.

src/OrchardCore.Modules/OrchardCore.Lucene/Services/SearchQueryService.cs
Skrypt: Example : 

```JSON
{
  "from" : 0,
  "size": 0,
  "query": {
    "match_all": { }
  }
}
```

PR #6000, opened 2020-04-21
Fixes #5761 : Move HtmlField's media modal in footer

### LINKED ISSUE #5761: FlowPart: layout for second HTML widget is broken
When adding two widgets with an HTML field to a FlowPart, layout is broken for the second widget:

flow-layout

It's working fine when adding the widget, this happens when you edit an existing item.

TFleury:
Having the modal in the content breaks the `div.form-group:only-of-type` CSS rule.

It's also cleaner to have the modal body in the footer.


TFleury:
It also fixes an unreferenced issue : If you remove the first HtmlWidget (the one hosting the modal) from FlowPart, you lose media picker.

deanmarcussen:
This looks like a great idea (I need to use the same idea for a modal on one of my pr's)

Would you need to do the Markdown Field / HtmlBodyPart / MarkdownBodyPart modal wrappers as well?

TFleury:
Thank you for your comment.
I didn't know there was the same trick in other part/fields. I will apply the same for consistency.

PR #6007, opened 2020-04-21
Admin menus positions

agriffard:
Prefix all position parameters by `am-` in admin menus to differenciate them from the title.


sebastienros:
Maybe we should use a special prefix for localized strings that should not be added to po files. That would mitigate the issues with this technique. Then for "Content Types" and "Content Parts" you can use `T["Content Types"] + ".1"` and `T["Content Types"] + ".2"` so it's still order relatively to the local language, but parts is always right after types.

agriffard:
> This will duplicate the localization strings. Am I right?
Yes but people don't really need to translate them

agriffard:
For `Content Types` and `Content Parts`, I called them `am1-Content Types` and `am2-Content Parts`.

agriffard:
> Maybe we should use a special prefix for localized strings that should not be added to po files

I like the idea.

Issue #6008, opened 2020-04-21
Workflow: Content not available when create event is fired


duncanhoggan:
I have had a strange problem that straight after a content item is created and the event triggered in a workflow the content item hasn't actually been created and I get the following exception.

```
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=OrchardCore.ContentManagement.Abstractions
  StackTrace:
   at OrchardCore.ContentManagement.ContentItemExtensions.Merge(ContentItem contentItem, Object properties, JsonMergeSettings jsonMergeSettings) in C:\Developement\R&D\OrchardCore\src\OrchardCore\OrchardCore.ContentManagement.Abstractions\ContentItemExtensions.cs:line 88

```

Tracking the exception up the stack it seem like the event is fired before the content creation is saved.

https://github.com/OrchardCMS/OrchardCore/blob/cac3c48ebd21ff03ab44414dceecbd3520ecb34d/src/OrchardCore/OrchardCore.ContentManagement/DefaultContentManager.cs#L383-L392

The sample workflow is as seen below for reproduction steps.
![image](https://user-images.githubusercontent.com/6050438/79906819-cc4e8180-8418-11ea-8787-24ec392cb960.png)

I'll add the PR and reference shortly. 😎

sebastienros:
It will never be fired before it's saved, because the db commit happens at the end of the request, and the workflow is running inline, before that. But the instance should be available in one the arguments of the task. And doing a sql query using YesSql should still make it availble to queries.

duncanhoggan:
@sebastienros I tested both cases, the previous order throws the exception and after re-ordering its fine.

sebastienros:
What exactly is in the task that triggers the exception ?

duncanhoggan:
As seen in the workflow image.
When the `Content Created` event is fired the content item doesn't exist.
Subsequently any task that that tries to access the Content Item looks like its failing, I'll have a look and see how the workflow accesses the content, but i assume its through the Content Manager.

duncanhoggan:
var contentItemId = await GetContentItemIdAsync(workflowContext);
            var contentItem = await ContentManager.GetAsync(contentItemId, VersionOptions.Latest);

The ContentItem queried in the `ContentManager.GetAsync` returns null as is currently.

Its using the `DefaultContentManager` underneath and looks like the transaction hasn't been committed to the YesSql `_session`

[...]

PR #6009, opened 2020-04-21
Fixed content create event order

duncanhoggan:
Set content create method to fire when content has been created as currently the content item doesn't exist when the event is fired.
#6008


src/OrchardCore/OrchardCore.ContentManagement/DefaultContentManager.cs
```diff
@@ -380,6 +380,9 @@ public async Task CreateAsync(ContentItem contentItem, VersionOptions options)
                 contentItem.Published = false;
             }
 
+            _session.Save(contentItem);
+            _contentManagerSession.Store(contentItem);
+
```
Skrypt, line 385: Should be between creating and created event

duncanhoggan:
@Skrypt Swapped the order around as per our discussion.

jtkech:
@duncanhoggan @Skrypt 

> Should be between creating and created event. In fact it doesn't make any difference though since _session.Save() here is no op. It will execute at the end of the Scope.

Yes the _session.Save() is no-op in the sense that it just enlists the item in the saved items collection of yessql that will be used when commiting at the end of the scope. And it is really no-op if the item is already in the collection.

Then if in an handler (or any other places) we do a query, there is an auto flush that clears the saved items but that saves the entities, so here a session save, that would re-enlist the item, is still useless as the entities have been saved. This is where i was wrong in one of my comments in #6008.

But then if after the query **there is another mutation** (no matter where), **another session save is needed** to take into account the last mutation.

**That's why in some places we do a session save after the handlers**. @Skrypt we already saw this issue together when you tried to mutate an item after `PublishAsync()` that does a session save before the handlers, and when one handler was doing a query.

But on another side if we want to allow any handler to see a current mutation, **we would need to do a session save before any handler**. Note: And `_contentManagerSession.Store` if there is one and that may help an handler to retrieve a cached item without triggering a session query.

So we would need a more global solution, see my last comment in #6008 where i suggested one.

**Meanwhile here i would suggest** to call session save (and store here) **before any handlers** as @duncanhoggan originally did (e.g if in the future we also have a WF `ContentCreatingTask` that also needs to get the item). **And then, as before, still call session save after any handler**.

duncanhoggan:
@jtkech Thanks I can see you have a much better understanding of the data interaction trust you guidance here. I have a project currently that relies on this so i think i will have to implement my own ContentManager where I'll make the changes you suggested, I'll update the pr to include the session save in both places for now and eagerly await the changes you are making. Thank you!

Skrypt:
My suggestion was to make the session.save() in between those 2 events because I assumed that handlers could/would potentially be used for doing a session.cancel(). If you call a session.save() after the CreatedAsync event then you remove the ability to cancel a session in a handler.

If we do : 

session.save()
CreatingAsync() --> can't do session.cancel() in Handler as session.save() will be called later on.
session.save()
CreatedAsync() --> can't do session.cancel() in Handler as session.save() will be called later on.
session.save()

Here it is just an example. But the logic should be that on a CreatingAsync event Handler we should only need to mutate the data. So if we considerate this factor then we don't need to call a session.save() before... since the data could be mutated and we know that we will call session.save() later on.

CreatedAsync() logic should be that it is called when the data is persisted. Here we can't because the SQL CommitAsync call needs to be cancellable from Handlers through session.Cancel(). So here again if a handler tries to cancel the session we can't because we do a session.save() after the event call.

So this is why I was suggesting adding it only between those 2 events so that at least we could still cancel the session on a CreatedAsync() event handler.

duncanhoggan:
I've managed to run into some more interesting situations which is highlighting the need for a robust reliable solution to Content Item access through the workflow.

If a `User Task` is acted on this triggers a content update through the admin controller, if this content is updated as a follow on of the user action a new version of the item is created and the state update is not saved. 

I'm going to retract my PR as I'm pretty certain this isn't the correct fix for this. 
interested to see #6021 in action potentially that will solve our problems.

jtkech:
@Skrypt 

Yes i totally understand what you mean but you can do a session cancel at any time before committing, can be before / after any session save. Then when committing at the end of the scope, if canceled the transaction is rollbacked.

                if (!_cancel)
                {
                    if (_transaction != null)
                    {
                        _transaction.Commit();
                    }
                }
                else
                {
                    if (_transaction != null)
                    {
                        _transaction.Rollback();
                    }
                }

Having an event triggered when an item has been really persisted is a good idea. Maybe good also in some context to be able to check if the session has already been canceled.

@duncanhoggan okay thanks, will let you know when i think the PR is ready

Hmm, with user activities maybe the problem is that the `UserStore` does explicit `CommitAsync()`, i don't remember if there is any reason to do that. So the session may be committed before the end of the scope, so after an user task, a next activity may not be able to do any other updates.

duncanhoggan:
@jtkech I've spent the better part of 2 days fighting that, it ends up locking the DB and creating duplicate Content Items but not updating them. I tried running Commit Transaction tasks in between and didn't have and joy with that. 

I'm going to have to abandon this as a viable option for the current project that requires this and figure out another way to get this done.

Thanks for you help and I eagerly await your #5249 and #6021 contributions. 
If i have no possible way of getting this done I'll potentially revisit this.

jtkech:
@duncanhoggan 

> If a User Task is acted on this triggers a content update through the admin controller, if this content is updated as a follow on of the user action a new version of the item is created and the state update is not saved.

Could you give me more infos on your WF when you saw this issue? Do you mean that you still used a `ContentCreatedEvent` for starting the WF or a `ContentUpdatedEvent`? Then an user task, if so a `RegisterUserTask` or an `AssignUserRoleTask`? Then an `UpdateContentTask` as in your 1st WF?

And did you tried it with the last dev branch, or with your changes? If so with session save only before the handlers or also after?

duncanhoggan:
@jtkech I've created a simple example of what happens using the WF below.

image

Before executing the WF my Content Item appears as follows.

image

Then i select `Deny` on the user task
image

Then the result is two instances neither of which have the updated content.
image

I've tried to insulate the calls using the `Commit Transaction` Task but it has no effect.

jtkech:
@duncanhoggan thanks a lot for these infos

I will try this scenario tomorrow. on my side i started to do real tests on different scenarios, I'm starting to get good results and i think i will open another PR soon.

But need to sleep now, will see tomorow ;)

Issue #2609, opened 2018-11-01
Rebuilding Lucene Indices after import


Skrypt:
When we reimport content through the deployment module ; content items are not added to the Indexing table unless you go back and re-publish them one by one. Either we need to add an option so that all published content items be added back to the index table when we rebuild the Lucene index or that we add those records when we reimport data.

I believe that the index table was introduced for performance reasons. @jtkech

Skrypt:
Here is a SQL snippet that I used for this after reimporting data with import/export feature.

```SQL
INSERT INTO IndexingTask ( ContentItemId, CreatedUtc, Type )
SELECT ContentItemId, date('now'), 0 FROM ContentItemIndex WHERE ContentType = 'ContentTypeName' AND Published = 1 AND ContentItemId NOT IN (
SELECT ContentItemIndex.ContentItemId FROM ContentItemIndex INNER JOIN IndexingTask ON ContentItemIndex.ContentItemId = IndexingTask.ContentItemId WHERE ContentType = 'ContentTypeName' AND Published = 1
)
```

agriffard:
This is an important fix to make.
When I Import a Blog and some associated Posts with a ListPart, the posts won't appear in the blog admin page or on front because the indexing hasn't been triggered.

sebastienros:
Maybe we need a step to rebuild an index as part of the recipe.

Another option would be to have an event handler when a recipe is executed (we might already have it) and have the Lucene feature handle it to update the index, not rebuild (or reset).

sebastienros:
I think this should work now, but maybe too well, in a way that each import would trigger a content item to be indexed instantaneously. This might slow down the import process.

A solution would be to make this automatic indexation an option in the lucene service state, have it on by default, and then create new recipe steps, one to deactivate it, one to activate it, and one to rebuild/reset an index.

We could also think about a way to detect that a recipe is running, such that features that don't want to run during the recipe can detect it. In the Recipe service, have a flag to detect that one is running. This way the background tasks would not be triggered, and we could also skip automatic indexing of content items, and detect when the recipe is successful to trigger an update of the index.

agriffard:
Does the Lucene-index recipe step fix this issue?

Skrypt:
[...]

PR #6021, opened 2020-04-23
Use SessionHelper for saving

### LINKED ISSUE #6008: Workflow: Content not available when create event is fired

jtkech:
Fixes #6008, see my comments in the related issue

- So in the content manager we may need to call a session save before an handler so that it can retrieve the related item with a session query. But we also may need to call the session save after if an handler does a session query and also another mutation on the related item.

- So here the idea is to use the `ISessionHelper` to save items and commit the session, an item being saved to the session immediately and also cached so that it can be saved again before committing. This way we can always do `_sessionHelper.Save()` before the handlers.

- But part of the issue may be only related to Workflows (see below). So, as i assume that we will never have `ContentXXXingEvent` activities, only `ContentXXXedEvent` ones, here right now we call `_sessionHelper.Save()` between the "XXXing" and "XXXed" handlers (as mostly done before).

---

- **That said part of the issue may be only related to Workflow Content events / tasks**. I think that handlers are not intended to use the `ContentManager` that calls them, but for a Content event, this is a content handler that triggers the event, here meaning that it executes the related WF(s) activities that may use the `ContentManager` that may call handlers that may trigger WF(s).

- Also normally handlers use a context e.g `CreateContentContext` to retrieve the related item(s). When a WF Content event is triggered this context is not passed but the related content item is passed, so maybe some Content activities would have to retrieve this content item from the worflow context, this in place of using the `ContentManager` to retrieve it and also to upate it.

- Hmm, ATM my feeling is that when a WF is not triggered by a Content event, e.g a Request event, it can use Content activities **to behave e.g as a controller**. But when a WF is triggered by a Content event **it would have to behave as a content handler**.

Need to sleep, i will think about it tomorrow ;)

### DIFF
--- a/src/OrchardCore/OrchardCore.ContentManagement/DefaultContentManager.cs
+++ b/src/OrchardCore/OrchardCore.ContentManagement/DefaultContentManager.cs
@@ -11,6 +11,7 @@ using OrchardCore.ContentManagement.Metadata;
 using OrchardCore.ContentManagement.Metadata.Builders;
 using OrchardCore.ContentManagement.Metadata.Settings;
 using OrchardCore.ContentManagement.Records;
+using OrchardCore.Data;
 using OrchardCore.Modules;
 using YesSql;
 using YesSql.Services;
@@ -21,6 +22,7 @@ namespace OrchardCore.ContentManagement
     {
         private readonly IContentDefinitionManager _contentDefinitionManager;
         private readonly ISession _session;
+        private readonly ISessionHelper _sessionHelper;
         private readonly ILogger _logger;
         private readonly IContentManagerSession _contentManagerSession;
         private readonly IContentItemIdGenerator _idGenerator;
@@ -31,6 +33,7 @@ namespace OrchardCore.ContentManagement
             IContentManagerSession contentManagerSession,
             IEnumerable<IContentHandler> handlers,
             ISession session,
+            ISessionHelper sessionHelper,
             IContentItemIdGenerator idGenerator,
             ILogger<DefaultContentManager> logger,
             IClock clock)
@@ -39,6 +42,7 @@ namespace OrchardCore.ContentManagement
             Handlers = handlers;
             ReversedHandlers = handlers.Reverse().ToArray();
             _session = session;
+            _sessionHelper = sessionHelper;
             _idGenerator = idGenerator;
             _contentManagerSession = contentManagerSession;
             _logger = logger;
@@ -190,14 +194,14 @@ namespace OrchardCore.ContentManagement
                     else
                     {
                         // Save the previous version
-                        _session.Save(contentItem);
+                        _sessionHelper.Save(contentItem);
 
                         contentItem = await BuildNewVersionAsync(contentItem);
                     }
                 }
 
                 // Save the new version
-                _session.Save(contentItem);
+                _sessionHelper.Save(contentItem);
             }
 
             return contentItem;
@@ -264,13 +268,13 @@ namespace OrchardCore.ContentManagement
 
             if (previous != null)
             {
-                _session.Save(previous);
+                _sessionHelper.Save(previous);
                 previous.Published = false;
             }
 
             contentItem.Published = true;
 
-            _session.Save(contentItem);
+            _sessionHelper.Save(contentItem);
 
             await ReversedHandlers.InvokeAsync((handler, context) => handler.PublishedAsync(context), context, _logger);
         }
@@ -314,7 +318,7 @@ namespace OrchardCore.ContentManagement
             publishedItem.Published = false;
             publishedItem.ModifiedUtc = _clock.UtcNow;
 
-            _session.Save(publishedItem);
+            _sessionHelper.Save(publishedItem);
 
             await ReversedHandlers.InvokeAsync((handler, context) => handler.UnpublishedAsync(context), context, _logger);
         }
@@ -337,7 +341,7 @@ namespace OrchardCore.ContentManagement
 
                 if (latestVersion != null)
                 {
-                    _session.Save(latestVersion);
+                    _sessionHelper.Save(latestVersion);
                 }
             }
 
@@ -386,10 +390,10 @@ namespace OrchardCore.ContentManagement
             // invoke handlers to add information to persistent stores
             await Handlers.InvokeAsync((handler, context) => handler.CreatingAsync(context), context, _logger);
 
-            await ReversedHandlers.InvokeAsync((handler, context) => handler.CreatedAsync(context), context, _logger);
-
-            _session.Save(contentItem);
+            _sessionHelper.Save(contentItem);
             _contentManagerSession.Store(contentItem);
+
+            await ReversedHandlers.InvokeAsync((handler, context) => handler.CreatedAsync(context), context, _logger);
 
             if (options.IsPublished)
             {
@@ -408,9 +412,10 @@ namespace OrchardCore.ContentManagement
             var context = new UpdateContentContext(contentItem);
 
             await Handlers.InvokeAsync((handler, context) => handler.UpdatingAsync(context), context, _logger);
-            await ReversedHandlers.InvokeAsync((handler, context) => handler.UpdatedAsync(context), context, _logger);
 
-            _session.Save(contentItem);
+            _sessionHelper.Save(contentItem);
+
+            await ReversedHandlers.InvokeAsync((handler, context) => handler.UpdatedAsync(context), context, _logger);
         }
 
         public async Task<TAspect> PopulateAspectAsync<TAspect>(IContent content, TAspect aspect)
@@ -446,7 +451,7 @@ namespace OrchardCore.ContentManagement
             {
                 version.Published = false;
                 version.Latest = false;
-                _session.Save(version);
+                _sessionHelper.Save(version);
             }
 
             await ReversedHandlers.InvokeAsync((handler, context) => handler.RemovedAsync(context), context, _logger);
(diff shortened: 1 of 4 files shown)


Skrypt:
I think we still need to have also the ability to get a content item without passing it in the WF activities as these could be triggered also later on and mutated in between. So when for example you call a GetContentItemAsync on a WF activity this is mainly to intentionally make sure that you get the proper last modified content item instead of using the dated stored data of the WF itself.

jtkech:
> I think we still need to have also the ability to get a content item without passing it in the WF activities

**Yes i agree**, e.g when a WF is triggered / executed by an HttpRequestEvent. Here i was thinking about the special case when triggered by a`ContentXXXedEvent`, meaning it is executed through a content handler, so i thought it would have to behave as a content handler. **But here also you're right**, e.g if in a WF, after a `ContentCreatedEvent`, and before using an `UpdateContentTask`, you wait by using a `TimerEvent` activity, then you're right the last task needs to retrieve the item from the db.

E.g I thought about a `ContentCreatedEvent` => `CreateContentTask` that may crash the app, but nothing prevents to do the same in a regular handler.

> this is mainly to intentionally make sure that you get the proper last modified content item instead of using the dated stored data of the WF itself which could be inapropriate in certain use cases.

Because of the above i agree then ;)

jtkech:
@duncanhoggan in #6009 you said recently

> If a User Task is acted on this triggers a content update through the admin controller, if this content is updated as a follow on of the user action a new version of the item is created and the state update is not saved.

Could you give me more infos on your WF when you saw this other issue? Do you mean that you still used a `ContentCreatedEvent` for starting the WF or a `ContentUpdatedEvent`? Then an user task, if so a `RegisterUserTask` or an `AssignUserRoleTask`? Then an `UpdateContentTask` as in your 1st WF?

And did you tried it with the last dev branch, or with your changes in your PR #6009? If so with session save only before the handlers or also after the handlers?

jtkech:
@duncanhoggan maybe i will keep this PR as it is still useful for some scenarios, but i don't think this is the right fix. I think we rather need a little change in `UpdateContentTask` as suggested below. If you could do another PR and try it, i will help you, otherwise i will do if you can't ;)

- So, what i did here is useful if after a session save, an handler does a query and then a mutation without another session save (explicitly or through a content manager method).

    But this is not the right fix. Okay it solves the original issue when subscribing to the `ContentCreatedEvent`, but, after creating an item through the admin, if you have subscribed to `ContentUpdatedEvent`, only the content manager `NewAsync()` would have been called and in which there is no session save to move, here we would have to move the save in the controller.

    **So finally i think that `UpdateContentTask` needs a little change because the problem is how we use it. When it is triggered by a `ContentEvent`, we should use the content item being mutated that has been passed to the WF context input, without doing another query**.

- So, in `UpdateContentTask`, in place of `GetContentItemIdAsync()`, i suggest `GetContentAsync()` so that if the expression is an id string it does `GetAsync(id, Latest)` for you, and if no expression it retrieves the content item from the WF context input or properties. So something like this

      var contentItem = (await GetContentAsync(workflowContext)).ContentItem;
      //var contentItemId = await GetContentItemIdAsync(workflowContext);
      //var contentItem = await ContentManager.GetAsync(contentItemId, VersionOptions.Latest);

    So, if triggered e.g by a `ContentCreatedEvent` you would have to not provide any expression or, with the above change and regarding the code, an expression providing a content item (not only an id) e.g from the WF context input / properties / lastResult is already also possible.

   We would have to change the UI to say that we can also provide a content item, and not display `ContentItemId` for this expression, as done for `PublishContentTask`. And for the other tasks that already say the following, we could add that the expression can also evaluate a contentItemId.

      Enter a JavaScript expression that evaluates to the content item to be published.
      For example: input("Content"). If no expression is specified, the workflow's
      "Content" input variable or "Content" property will be used.

    Note: Hmm, maybe not necessary to do the same for the `RetrieveContentTask`.

deanmarcussen:
**So finally i think that `UpdateContentTask` needs a little change because the problem is how we use it. When it is triggered by a `ContentEvent`, we should use the content item being mutated that has been passed to the WF context input, without doing another query**.

>   Note: Hmm, maybe not necessary to do the same for the `RetrieveContentTask`.

I think I agree with this, when I first saw the `UpdateContentTask` workflow PR, I assumed that it was intended for use with the `RetrieveContentTask`, where it is making a query, to retrieve the content item first.

On a side note, just because of a comment on the other pr/issue, which maybe/maynot related.
When passing the ContentItem object around (to workflows etc), it is important to make sure it is the same object reference being mutated, and saved, or the `Yes.Sql` `IdentityMap` will not know it is the same object to be updated.

jtkech:
@deanmarcussen 

Yes there are many scenarios and any task can be used after any other one, so i think that the right fix is to change how a WF task retrieve an item. If there is no item being mutated, that's okay as it is currently done. But when triggered by a content event (through the content handler), a WF task should retrieve the item being mutated (if acting on the same item id) and be able to do that.

Seems to be ok to retrieve the same entity if the item has already been queried, even with `DraftRequired` and then with `Latest`, as the 1st one build a new version with latest = true and does a session save, so the 2nd retrieve the same cached instance, but yes still potential issues.

Here the issue was when **creating an item**, when subscribing to `CreatedContentEvent` the item was not yet saved and it has been solved by moving up the session save in the content manager `CrateAsync()`, but if you subscribe to the `UpdatedContentEvent` you still have the same issue as the update handlers are called just after the content manager `NewAsync()`, so no session save

---

So the idea is

- I will harmonize the UI for WF task javascript expressions that can use the WF context input / properties / lastResult, for some we are saying that they can provide a content item, but regarding the current code, we can also say that the expression can also be used to provide an item id.

    For others we are just saying that they can provide a content item id, so here i will change the code (very simple) to do the same as above that still allows to just provide an id.

- Then, knowing that when a WF content event is triggered, the item being mutated is passed to the WF context input, when there is no expression we 1st try to get the item from the WF context input, as currently done, but when you provide an item id (or an item), if this id (or the id of the item) is the same as the id of the item in the WF context input, we use the WF context input instance.

Skrypt:
> When passing the ContentItem object around (to workflows etc), it is important to make sure it is the same object reference being mutated, and saved, or the Yes.Sql IdentityMap will not know it is the same object to be updated.

This is why you generally get the content item with RetrieveContentTask and then you mutate it with the WF input data... But we're probably all saying the same thing. We all agree that there is 2 different use cases here. One that will require to retrieve the data from the DB because it might have mutated over time and we don't want it to rely on old data from the WF context. And a second one where it will actually not care if the data is old or not. It will update the content item with the WF context data simply. Last updates wins.

jtkech:
- @duncanhoggan sorry i didn't remember the `UserTak` that is waiting for an user action through a new button when editing the related item, so i could start some tests with your last use case.

- @Skrypt yes and nothing better than real tests ;)

    You're right, e.g if a content handler triggers a WF it passes the item to the WF input, but then if you have an blocking task, when the WF is resumed later on, the WF context input is the one that has been persisted and that may be stale.

    So i reversed my strategy by doing quite the same as before, by getting the item through the content manager, which is ok when the WF activity is not triggered directly by a content event, otherwise most of the time the item has already been queried so it returns a cached instance, this is only on a Created Event that it returns nothing, and this is here that i fallback to the WF context input that has been passed by the content handler.

    I'll do more tests tomorrow and maybe open another PR.

- For infos, if you have a WF that starts by a `ContentUpdatedEvent` followed by an `UpdateContentTask` that calls `UpdateAsync()` and then triggers again a `ContentUpdatedEvent`, you crash the app ;)

- @deanmarcussen about importing batched items, maybe good to say in the docs that before importing a bunch of items, you may want to disable existing worflows that are triggered by a Content Event ;)

Skrypt:
> For infos, if you have a WF that starts by a ContentUpdatedEvent followed by an UpdateContentTask that calls UpdateAsync() and then triggers again a ContentUpdatedEvent, you crash the app

lol

deanmarcussen:
> * @deanmarcussen about importing batched items, maybe good to say in the docs that before importing a bunch of items, you may want to disable existing worflows that are triggered by a Content Event ;)

Noted: for docs (much to document there)

For info when we talked about lucene in this context https://github.com/OrchardCMS/OrchardCore/issues/2609 an event handler was considered, to halt indexing during import.

But with your deferred task for lucene, the indexing is already paused, during the session, and the queries in the deferred task, for the most part at least, should be retrieving from the session, as they have already been queried and are in the session.

There are also events that get fired when importing a recipe (not specifically content, just any recipe), which could also be used to halt things, but some people will still want their workflows to fire when importing, and some will not (depends on the workflow, I suspect, probably mostly not, but for example a workflow that updates a calculated field value, would still want to run, probably).

What we could do is implement a recipe running workflow event / state, at some point, driven by those events, so workflows could choose to no-op

Regarding this pr? 
Will you keep it was well as another workflow related pr's?
It seems to me it has a usefulness that extends past workflows?

Get some sleep ;)

jtkech:
@deanmarcussen 

> What we could do is implement a recipe running workflow event / state, at some point, driven by those events, so workflows could choose to no-op

Good idea, or some additional contextual data, i was talking about this just to be sure we are aware of this kind of things, and it seems to be the case ;)

> Regarding this pr?
Will you keep it was well as another workflow related pr's?
It seems to me it has a usefulness that extends past workflows?

I will keep this PR on which I need to focus more, but first i will open a new PR to fix the WF issue as it can be fixed independently.

jtkech:
Related and experimented in #6187

PR #6031, opened 2020-04-24
Set max width on activity/task span

duncanhoggan:
Just some style updates to the activity block to prevent them from becoming super long (as seen below)

image

Now it looks like this 

image

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Workflows/Assets/Styles/workflow-canvas.scss
+++ b/src/OrchardCore.Modules/OrchardCore.Workflows/Assets/Styles/workflow-canvas.scss
@@ -68,6 +68,14 @@
                 }
             }
         }
+        
+        span {
+            display: inline-block;
+            max-width: 300px;
+            white-space: nowrap;
+            overflow: hidden;
+            text-overflow: ellipsis;
+        }
     }
 
     .jtk-connector {


deanmarcussen:
Does this need a rebuild of the gulp file for the wwwroot/css file, as the only update here is to the sass?

duncanhoggan:
Happy to do the build, I assumed the CI/CD would handle that. Will run the tasks and get the css and minified css included👍

src/OrchardCore.Modules/OrchardCore.Workflows/Assets/Styles/workflow-canvas.scss
```diff
@@ -68,6 +68,14 @@
                 }
             }
         }
+        
+        span {
+            display: inline-block;
+            max-width: 300px;
```
Skrypt, line 74: It doesn't prevent the box itself to not become too wide by doing this. I can find other type of activities/events that are not using span at all.

Skrypt: Example : 

image

deanmarcussen:
> I assumed the CI/CD would handle that. 

Just for info, when you change the sass, you have to build the css. The CI can't do that, because we need the generated css in the git repository, for everyone using it :)

duncanhoggan:
@Skrypt hmm, yeah I was solving a very specific issue there, I'll update the PR to be more generic.
@deanmarcussen Thanks for the heads up, I'll make sure to get all the styles pre-compiled 👍.

agriffard:
Please merge form dev and rebuild the assets with gulp

src/OrchardCore.Modules/OrchardCore.Workflows/Assets/Styles/workflow-canvas.scss
```diff
@@ -68,6 +68,14 @@
                 }
             }
         }
+        
+        span {
+            display: inline-block;
```
Skrypt, line 73: It doesn't prevent the box itself to not become too wide by doing this. I can find other type of activities/events that are not using span at all.

80280428-faaab600-86d1-11ea-9198-05e18af70e31

Skrypt (changes requested):
Use a more global class selector.

agriffard:
Closing the dontmerge labels.

PR #6034, opened 2020-04-25
Upgrade TheAgencyTheme

agriffard:
https://www.npmjs.com/package/startbootstrap-agency v6.0.0

Changes:
Use package.json at the root of TheAgencyTheme instead of copying the repo in wwwroot.
Changed ResourceManifest to use Resources and added jquery.easing.
No more `vendors` folder.
img moved to assets/img (but copied but the media recipe step)

Changes in Css classes:
Head:

```
<header class="masthead">
            <div class="container">
                <div class="masthead-subheading">Welcome To Our Studio!</div>
                <div class="masthead-heading text-uppercase">It's Nice To Meet You</div>
                <a class="btn btn-primary btn-xl text-uppercase js-scroll-trigger" href="#services">Tell Me More</a>
            </div>
</header>
```
Footer:

```
<footer class="footer py-4">
            <div class="container">
                <div class="row align-items-center">
                    <div class="col-lg-4 text-lg-left">Copyright © Your Website 2020</div>
                    <div class="col-lg-4 my-3 my-lg-0">
                        <a class="btn btn-dark btn-social mx-2" href="#!"><i class="fab fa-twitter"></i></a><a class="btn btn-dark btn-social mx-2" href="#!"><i class="fab fa-facebook-f"></i></a><a class="btn btn-dark btn-social mx-2" href="#!"><i class="fab fa-linkedin-in"></i></a>
                    </div>
                    <div class="col-lg-4 text-lg-right"><a class="mr-3" href="#!">Privacy Policy</a><a href="#!">Terms of Use</a></div>
                </div>
            </div>
</footer>
```


agriffard:
Changed the primary color variable in scss to use green instead of yellow and removed the header background:

image

hishamco:
Seem it's Orchard Core website ;) but I'm not sure if it's good idea to do a modification on current themes because the user may except it is the exact open source template without modification

agriffard:
I could do worse if you don't stop me ;-)

image

hishamco:
This is may cool for Orchard Core website

agriffard:
image

JoshTango:
the color needs to be a variable that you can edit in the admin UI.

Skrypt:
At this point I think we could also bundle everything in a single .css file instead of using the bootstrap-oc.css file to override the base theme. We are building the entire theme now ...

agriffard:
> we could also bundle everything in a single .css file

Ok, done @Skrypt.

src/OrchardCore.Themes/TheAgencyTheme/TheAgencyTheme.csproj
```diff
@@ -11,18 +11,22 @@
...
     <ProjectReference Include="..\..\OrchardCore\OrchardCore.ResourceManagement.Abstractions\OrchardCore.ResourceManagement.Abstractions.csproj" />
   </ItemGroup>
 
   <ItemGroup>
-    <EmbeddedResource Remove="**\*.less;**\*.scss;" />
-    <EmbeddedResource Remove="wwwroot\gulpfile.js" />
-    <EmbeddedResource Remove="wwwroot\index.html" />
-    <EmbeddedResource Remove="wwwroot\package-lock.json" />
-    <EmbeddedResource Remove="wwwroot\package.json" />
+    <Folder Include="wwwroot\scss\" />
```
Skrypt, line 29: Needed?

agriffard: Ok, removed.

Skrypt:
I've not tried it but LGTM

deanmarcussen:
So breaking, somewhat, on existing installs, due to lots of changes and a whole new version (v6) from StartBootstrap.

Also LGTM

Suggest is another candidate to explain the changes during todays meeting

src/OrchardCore.Themes/TheAgencyTheme/TheAgencyTheme.csproj
jtkech: @agriffard 

We exclude from embedded resources all files under the `Assets` folder **but nothing is excluded under `wwwroot`**, unless some well known things as `node_modules` that we exclude from everywhere.

So if a theme has its own asset building logic under its wwwroot folder, there can be more or less files that need to be explicitly excluded, as here `package-lock.json`. 

So i think we need to keep these specific exclusions

You can try it and check the size of the generated assembly. Good to test it with all the possible files under wwwroot, e.g after having installed all the needed things to build the asset pipeline.

ns8482e:
I would still prefer to use build script provided in agency theme source

src/OrchardCore.Themes/TheAgencyTheme/TheAgencyTheme.csproj
agriffard: Those files are not here anymore in wwwroot.
The package.json is now at the root of the theme and restores the files in node_modules.

jtkech: Ah okay but we also embed files that are at the root of the module / theme, unless well known things as `node_module` or e.g csharp files, only files under the `Assets` folder are not embedded.

So we have to also explicitly exclude root files if they are quite big as e.g `package-lock.json` if it is now at the root but still there.

jtkech: @agriffard so here would be (maybe idem for other themes / modules)

    <EmbeddedResource Remove="package-lock.json" />

That said currently the size of this file is 267 ko while the size of the agency theme assembly (debug one so bigger) is 9680 ko

agriffard: Well, I did it but package-lock.json is just 1k because it just needs bootstrap and startbootstrap-agency, while the previous package.json in wwwroot needed all the gulp dependencies.

jtkech: Ah okay, indeed not needed for low size files

deanmarcussen:
> I would still prefer to use build script provided in agency theme source

So @agriffard there is still set of build scripts provided by the Start Agency Theme source?

Would it be better to use these, rather than our gulpfile (which cannot be directly copied into the project and used, if someone is taking a copy of this theme, and using it as a base to get started themselves)
Or maybe we could provide an adapted gulpfile in the theme itself, which actually works for the project?

It would be good to get this in before RC2 as it represents a breaking change to the theme

CSurieux:
> Would it be better to use these, rather than our gulpfile (which cannot be directly copied into the project and used, if someone is taking a copy of this theme, and using it as a base to get started themselves)
> Or maybe we could provide an adapted gulpfile in the theme itself, which actually works for the project?

Yes please, it should be a very nice to have (and may be a real form with captcha rather than the thing where any automate will post.
Thanks for all great work anyway.

PR #6038, opened 2020-04-26
Fixes rendering draft on front end

jtkech:
Currently we render the draft version on the front end, not the published one


jtkech:
Hi all, are you aware that currently, with the last dev branch

We are rendering draft versions on the front end ;)

sebastienros:
This PR is not bringing AngleSharp, just fixing the package reference

Issue #6013, opened 2020-04-21
HtmlPart with standard editor error in FlowPart


agriffard:
See https://github.com/OrchardCMS/OrchardCore/pull/6006#issuecomment-617426850

js error with code mirror.

/cc @ns8482e

ns8482e:
@agriffard However I am not using Standard editor

agriffard:
This is only occurring whne you add a widget with HtmlPart in a FlowPart on the fly?
After saving it and reloaded, there is no error, right?

ns8482e:
No, when I try to create new Page content type or edit existing Page content typ it gives this error

PR #6040, opened 2020-04-26
Fix null error when rendering html part in flow editor

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6013

Because the Flow editor renders shapes for items to get the scripts injected, when the script is intended to work with an elements id from the shape, that isn't rendered to html, there is a javascript error in the browser console.

@ns8482e I think I got all the places where this is done, if I've missed any, and you see the error again refer back to this


ns8482e:
I remember now, it’s due to mock rendering of widget scripts without having DOM html related to PR #5507

PR #5422, opened 2020-01-31
Check state validity of configured tenants.

jtkech:
Fixes #5419

PR #6045, opened 2020-04-28
Update configuration docs

deanmarcussen:
@jtkech before I forget again ;)

From your notes on the original pr https://github.com/OrchardCMS/OrchardCore/pull/5422

> So now, for the documentation

> To provide preconfig values for a given tenant you need a tenant valid state as Uninitialized. This will automatically create a tenant at runtime using these values.

> To only provide config values for a given tenant don't specify a tenant state. These values will be only used when the tenant is manually created.

I have updated to support the second scenario (configure values for a tenant after it has been created, i.e. specific blob storage keys, or similar)

For the first scenario is it only applicable for use with `.WithTenants()` in an mvc app ?
I can see the values coming into the system when used with `"State": "Unititialized"` but from the admin there is no way to go ahead and set the tenant up (that I see)


jtkech:
@deanmarcussen no problem i will answer you tomorrow, really need to sleep now ;)

jtkech:
> For the first scenario is it only applicable for use with .WithTenants() in an mvc app ?

No, it's now the same in both cases

> I can see the values coming into the system when used with "State": "Unititialized" but from the admin there is no way to go ahead and set the tenant up (that I see)

just tried

    "OrchardCore": {
      "Tenant1": {
        "State": "Uninitialized"
        "RequestUrlPrefix": "tenant1"
      }

And then i can see it in the tenants admin index page and i can do a setup on it

deanmarcussen:
> And then i can see it in the tenants admin index page and i can do a setup on it

Got it @jtkech . Must have had a bad build from shifting branches yesterday, as I tried it then, and couldn't run the setup.

Could now, so docs updated, hopefully correct ;)

jtkech:
Correct ;)

Just to say that when the tenant has been created, you can remove `"State": "Uninitialized"`, but this is not necessary, you can also **keep it** as it is now overridden by the value in `tenants.json`.

Issue #4255, opened 2019-09-12
Creating content in a nested workflow fails (`database table is locked`)


jean:
We have a workflow where a convenor can create a meeting slot and invite multiple people to meet. 

- An item is created to represent the meeting.
- We iterate invitees and kick off a confirm/decline workflow for each of them. 

This blows up due to the _create item_ task. If we remove that task, it works. 
Is this supposed to work, are we doing something wrong? 

Here's a zip to reproduce the issue: 
[Exportworkflows.zip](https://github.com/OrchardCMS/OrchardCore/files/3604790/Exportworkflows.zip)

## Create item and start a workflow instance per invitee

Here we're creating a placeholder `Test2` item, but in reality that will be a meeting-slot item which lists the meeting details and the invitees.

![Invitation Workflow - Dev 1 Booking Invitation Meeting](https://user-images.githubusercontent.com/84800/64768681-182c7480-d574-11e9-9b8b-f39122fba3a4.png)

## Send email and await response for an invitee

![Invitation Workflow - Dev 2 Send Email Invitation Meeting](https://user-images.githubusercontent.com/84800/64768679-182c7480-d574-11e9-86bf-14d2237e6ec2.png)

-----

With the attached recipe, we can reproduce this at https://try.orchardproject.net/ , but we can't see a traceback.

Here's a local traceback:

```
2019-09-11 15:44:03.2032|Default|80000021-0001-ff00-b63f-84710c7967bb||OrchardCore.Workflows.Services.WorkflowStore|ERROR|IWorkflowHandler thrown from OrchardCore.Workflows.Http.Handlers.WorkflowRoutesHandler by SqliteException Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 6: 'database table is locked'.
   at Microsoft.Data.Sqlite.SqliteException.ThrowExceptionForRC(Int32 rc, sqlite3 db)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader(CommandBehavior behavior)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteScalar()
   at System.Data.Common.DbCommand.ExecuteScalarAsync(CancellationToken cancellationToken)
--- End of stack trace from previous location where exception was thrown ---
   at Dapper.SqlMapper.ExecuteScalarImplAsync[T](IDbConnection cnn, CommandDefinition command) in C:\projects\dapper\Dapper\SqlMapper.Async.cs:line 1217
[...]

PR #6054, opened 2020-04-29
Fixes #6008 Update Content Activities

### LINKED ISSUE #6008: Workflow: Content not available when create event is fired

jtkech:
Fixes #6008 

@duncanhoggan could you re-do some tests by forking this branch or just copy `UpdateContentTask` and the little change in `DefaultContentManagerSession` needed in some scenarios.

- **You could retry it when triggered by a `ContentCreatedEvent`**. We use `DraftRequired` so that the content manager returns a new draft version or not by checking if the type is `Versionable`. Most of the time after a content event we retrieved a cached instance, but not here.

   So here, after having mutated this version this is not our responsability to publish it, it will be done or not by the "caller" that triggered the WF, e.g the admin controller => content handler => `ContentCreatedEvent` => the item is published or not by the controller.

- Then if related to an existing item and not triggered by a content event, e.g an `HttpRequestEvent`, as in your last WF sample in #6009, after the `UpdateContentTask` it's up to you to use or not a `PublishContentTask`.

- Then when using an `UserTaskEvent` edit action that adds a button when editing the related content item, when clicking on it this is always a `submit.save` that is posted, not a `submit.publish`, so the draft version is just saved, so here also it is up to you to use or not a `PublishContentTask`.

    So in your last WF sample in #6009 you would have to not use `PublishContentTask` after the `UpdateContentTask` but after the `UserTaskEvent`. Otherwise, after clicking the edit user button, it will always use the existing draft or create new one but without publishing it, as you saw.

    Then, if you didn't see any update, i think that there is another problem, maybe with your `ContentProperties` expression as i tried quite the same WF on my side and it was working


duncanhoggan:
@jtkech I got round to doing some testing, below are my findings.

- `ContentCreated` -> `UpdateContent` this results in an updated item in drat mode if i call `PublishContent` after it updates correctly ⭐.
- `HttpRequestEvent` -> `UpdateContent` All works as expected in isolation ⭐. Although if i call a `HttpRequestEvent` -> `UpdateContent`  from another workflow (nested) i get database lock exceptions ❌.
- `UserTaskEvent` ->  `UpdateContent` -> `PublishContent` works great ⭐.

Your changes look great, we just need to address that nested case and i think it could be good to add a `Publish` option to the `UpdateContent` task to ensure that the updated item is published in those cases where it isn't published (the first and last case that you mentioned).

jtkech:
- For your 2nd test we already have this issue, e.g. sqlite doesn't support concurrent writes so they wait for each other, here both worflows are not executing in the same scope, so that would be okay, but both are still waiting for each other because WF1 will only commit its transaction at the end of its scope, but it is waiting for WF2 that is also waiting for WF1 to commit its transaction.

    See #4530 and my last comments in #4255.

    **So here you need to use a `CommitTransactionTask` activity after the `UpdateContentTask` and before triggering your nested workflow.**

- I added the `Publish` option

jtkech:
@duncanhoggan and @deanmarcussen as you will have to review my changes related to the new `ValidateAsync()`.

Sorry but because of some recent changes that have been merged, now an update content activity may do a **session cancel** through a new `ValidateAsync()`, so it is important to add the validation errors to the underlying model state. Otherwise the controller, not aware of the failing validation, will do another save and still persist the item.

We already saw this case: 1. session save - 2. query + mutation - 3. we need a save to take into account the last mutation. Here, **because the cancel can be done earlier**, the new one is: 1. **session cancel** - 2. query - 3. session save => the item is persisted even the session has been canceled in 1.

**Todo: This use cases need to be supported transparently, but in a separate PR.**

So we really need to check if the `UpdateContentTask` activity is executing in a "regular" mode, or as an inline handler from the `ContentsHandler`, or as an inline driver from the `UserTaskEventContentDriver`.

**So i did some updates and re-do all the tests** that was working on my side.

- If used in a "regular" mode e.g. after an `HttpRequestEvent`, the `UpdateContentTask` works as before and i added a `Publish` property to publish or not the item after updating.

- If used e.g. after a `ContentCreatedEvent`, the `UpdateContentTask` will update the item and, because here it acts as an handler, it will neither call the content manager `UpdateAsync()` nor the `PublishAsync()`, this is the responsability of the underlying controller. And if the validation fails, **it reports the errors to the underlying model state that will be displayed in the admin UI**.

- Idem if used after an `UserTaskEvent` but here because the `UpdateContentTask` acts as a driver. Here also this is the underlying controller that has the responsability to save the item and then publish it or not. Anyway here we are executing before the regular display drivers that will overwrite your updates if related to regular parts / fields. But still useful to add some specific data as we do through some custom drivers. Then about publishing, the problem was that before an user action button was always sending a `submit.Save`, now when you define an edit user action e.g. `Go`, you will have 2 buttons `Go and Save` and `Go and Publish`.

- Finally, **we now support new scenarios**, you now can use an `ContentUpdatedEvent` followed by an `UpdateContentTask`, this because here it will act as an inline handler and then not call the content manager `UpdateAsync()`, **so there is no more an infinite loop in this case**.

Maybe i will need to also update other content task activities, but not sure, will see tomorrow.

deanmarcussen:
> * If used e.g. after a `ContentCreatedEvent`, the `UpdateContentTask` will update the item and, because here it acts as an handler, it will neither call the content manager `UpdateAsync()` nor the `PublishAsync()`, this is the responsability of the underlying controller. And if the validation fails, **it reports the errors to the underlying model state that will be displayed in the admin UI**.

Ok, makes sense @jtkech 

I wanted the `ValidateAsync` to be disconnected from a `ModelState` style of errors, because in some cases there is no `ModelState` that we want to work with.
 
But where there is (and there is a need for it), it makes good sense to join these errors into the `ModelState` as you have done.

(The other reason was because it was about validation failures, not successes, so didn't want to list all the succeeded validations, just the errors, or we could have just used a `ModelState` dictionary).

> call the content manager `UpdateAsync()`, **so there is no more an infinite loop in this case**.
👍 

> Maybe i will need to also update other content task activities, but not sure, will see tomorrow.

jtkech:
@deanmarcussen when you will have time

@agriffard requested a review from you

src/OrchardCore.Modules/OrchardCore.Contents/Workflows/Activities/ContentActivity.cs
```diff
@@ -27,6 +28,18 @@ protected ContentActivity(IContentManager contentManager, IWorkflowScriptEvaluat
...
+
+        protected bool FromContentHandler { get; private set; }
+
+        protected string OriginalCorrelationId { get; private set; }
+
+        protected bool AsCorrelatedContentDriver { get; private set; }
+
+        protected bool AsCorrelatedContentHandler { get; private set; }
+
+        protected bool AsCorrelatedContentDriverOrHandler { get; private set; }
```
deanmarcussen, line 41: Ok @jtkech I had some time to have a look.

It feels complicated (but it is a complicated scenario), and I can see it all works :)

Why do these properties need to be protected? 
From what I can see they are all set internally in the `ContentActivity` by various evaluations where you calculate how the inputs were recieved?

I don't mind that they are protected, it just confused me when I was trying to understand what they did, because I thought maybe one of the implementations had to set them, but then it didn't seem that is the case.

**Edit** Is it just so the workflows can return the `Noop` outcome? That makes sense then. (maybe a comment?)

src/OrchardCore.Modules/OrchardCore.Contents/Views/Items/UpdateContentTask.Fields.Edit.cshtml
```diff
@@ -12,7 +12,7 @@
     <label asp-for="ContentItemIdExpression">@T["Content Item ID"]</label>
     <input type="text" asp-for="ContentItemIdExpression" class="form-control code" asp-validation-class-for="ContentItemIdExpression" />
     <span asp-validation-for="ContentItemIdExpression"></span>
-    <span class="hint">@T["Enter a JavaScript expression that evaluates to the Content Item's ID(ContentItemId) to be updated."]</span>
+    <span class="hint">@T["Enter a JavaScript expression that evaluates to the Content Item's ID (ContentItemId) to be updated. For example: correlationId()."]</span>
```
deanmarcussen, line 15: What is a `correlationId()` is this a method we have somewhere?

jtkech: Yes, we can use it in a javascript expression, see in `WorkflowMethodsProvider.cs`

src/OrchardCore.Modules/OrchardCore.Contents/Workflows/Activities/ContentActivity.cs
jtkech: @deanmarcussen yes it is complicated because a Content Task may be executed inline e.g. as a content handler. This when inline after a Content Event triggered by the `OC.Contents.Workflows.Handlers ContentsHandler` that calls `TriggerWorkflowEventAsync` which sets the WF context input and correlationId by using the related content item. Inline meaning that between the Content Event and Content Task there is no waiting activity, meaning that the Content Task is executed in the same scope and acts as a content handler and should not call content manager methods.

Note: I found a solution by triggering the WF in a deferred task, so in another scope, and it was much more simple, but then e.g. a Notify Task after a Content Event was not working because the controller redirects before the WF execution. So i didn't use this and let execute inline a WF that is triggered by a Content Event.

> Why do these properties need to be protected?

Yes, these properties are mainly managed here, particularly though `OnInputReceivedAsync()` to see if the WF has been triggered from e.g. a content handler and to hold the `OriginalCorrelationId`. Then i made them protected so that they can be used by derived classes, see in `UpdateContentTask.cs` where `AsCorrelatedContentHandler` and `AsCorrelatedContentDriverOrHandler` are used and more frequently.

jtkech:
Maybe here i will do some simplifications because the session concerns in #6223 have been fixed on YesSql side, not yet available through nuget.

src/OrchardCore.Modules/OrchardCore.Contents/Workflows/Activities/ContentActivity.cs
deanmarcussen: Cool, yes, this is what I thought you were doing.

If there is an opportunity for simplification that would be great :)

`Inline` or `Inscope` make a bit more sense to me when reading this, rather than `AsCorrelateContentDriverOrHandler`

Is there a way the naming of these methods could be more generic, i.e. `Inline` or `Inscope`

This `AsCorrelatedContentDriverOrHandler` feels so tied to a `Driver` or `Handler` (and one day I guess there will be GraphQL mutations to take into account here too, for example).

Not trying to make more work for you here, and maybe my suggestions make no sense, so only if you agree.

jtkech: @deanmarcussen thanks for the feedback

> If there is an opportunity for simplification that would be great :)

I'll try, maybe i can remove the distinction between driver / handler, maybe some simplifications after the recent changes on yesSql side. I will rework on it.

> Is there a way the naming of these methods could be more generic, i.e. Inline or Inscope

Thanks for the suggestions, i like these names, i will think about it

> This AsCorrelatedContentDriverOrHandler feels so tied to a Driver or Handler (and one day I guess there will be GraphQL mutations to take into account here too, for example).

It is already related to admin controller, api controller, and yes one day maybe to graphQL, but normally always through content manager methods **and then content handlers**. About drivers it's only related to the WF `UserTaskEvent` that let you add user buttons when editing the related item, and that trigger an `UserTaskEvent` through a specific driver.

jtkech:
@deanmarcussen regarding your last review i made some changes

In fact in place of saying `executed inline from a content driver / handler`, we should say `executed inline from a content event` whatever the source that triggered the `ContentEvent`.

I removed the distinction between `from handler` and `from driver`, with the last one (related to the `UserTaskEvent`) the `ContentItem` was not passed to the WF input, so i was not able to check it when a WF is started / resumed. Now all content event sources pass the related content item.

These content event sources also pass a marker that we check, not passing this marker allows the scenario where we would want to trigger a `ContentEvent` but by still allowing an inline and correlated `ContentTask` to execute a content manager method.

**So now, a WF `ContentTask` checks only one property that is `InlineFromContentEvent`**.

deanmarcussen:
@jtkech I like it.

My only comment (and it is minor) is that i had to look to see the difference between `FromContentEvent` and `InlineFromContentEvent`

So one is a property that is set `FromContentEvent` and the other is used to evaluate if it was `InlineFromContentEvent` ?

But I think it looks good, so when you are finished with it mark it ready :)

jtkech:
@deanmarcussen 

> My only comment (and it is minor) is that i had to look to see the difference between FromContentEvent and InlineFromContentEvent

When `FromContentEvent` is set we don't know yet if the `ContentEvent` and the `ContentTask` are related to the same `ContentItemId`, we know this only when the task is executed and get a content item (according to its `Content` expression). So `InlineFromContentEvent` also means that the task and the event are related to the same `ContentItemId` (there are correlated). In that case the content task may act on the content item but without using content manager methods.

But i saw other cases, e.g. if a `CreateContentTask` is executed inline from a `ContentCreatedEvent`, even there are not correlated, if the event is a starting ativity this will end up in an infinite work. So, if a task is executed inline from an event and if the event is a starting activity, we have to also check if the task may trigger the same event, this even if the task and the event are not related to the same `ContentItemId`.

I will add this checking asap

Would be good to finish this PR before rc2 as it fixes many use cases where the app can crash. There are other tasks / events that would need the same but i limit this PR to the `Content` category.

jtkech:
@sebastienros 

Maybe too late for rc2 and not so important for me

But just to say that it fixes many scenarios where the application crashes, e.g. when a starting `ContentEvent` is followed inline by a `ContentTask` that calls an handler that triggers the same starting `ContentEvent`, which ends up in an infinitive loop

jtkech:
@sebastienros and others

So we discussed about preventing e.g. a WF infinite loop (#6455) but at another level, but it can't supersede all what is done in this PR specifically for content events / tasks. So here, as a reminder that may be used in another triage meeting, some uses cases to show what i mean.

- You are saving a draft through the admin UI => drivers update is successful => handlers => a WF content event => a WF content task that does a publish => **You are still notified that a draft has been saved even the content item has been published**.

    Here, when executed inline from a content handler related to the same content item id we don't call content manager methods, so the content item is not published.

- You are saving a draft through the admin UI => drivers update is successful => handlers => a WF content event => WF update content task where `ValidateAsync()` fails => **But you will still see that the draft has been saved successfully even the session has been canceled**.

    Here, when executed inline from a content handler related to the same content item id, if the validation fails we add errors to the underlying model state, so you will be notified that the related WF task failed to update the content item, and with error messages.

- **Idem with the `ApiController` where e.g. you may wrongly return `Ok(contentItem)`**.

- Note: Here when triggering a content event we not only pass the content item to the WF input but also a `ContentEventContext` with more infos, this in a normalized way that should be used for any future content event, as i did for the below `UserTaskEvent` that is a content event.

- The `UserTaskEvent`also belongs to the WF content events,it allows to add a button(s) e.g. `Go` when you edit the related item, the event is then triggered when you submit by using this button, but the submit was always a `submit.save`, here we have 2 options `Go and save` and `Go and publish`.

- Here i also added a `DraftSavedAsync` event, and its related WF `ContentDraftSaved` event, that is triggered when you save an item as a draft **but only if the content item has been validated**.

jtkech:
Fixes #6596, fixes #6597, fixes #6598, fixes #6599

jtkech:
@deanmarcussen

You are still a pending reviewer for this one ;) but only if you have time. I could merge it myself but because we talked about it in a recent triage meeting i prefer that someone else approves it (or not) before merging it (or just closing it).

In the meeting it was related to infinite loops but this PR supports many other scenarios. If needed i can help you to understand what i did, and to do some simple repros see my previous comment or the recent related issues i opened.

deanmarcussen:
@jtkech sorry, took some time to think about this one.

For the detecting and cancelling infinite loops all seems to make sense.

For the Inline (from driver, and external) it looks like you have changed the shape of this pr for the better.

Can you explain though?

Previously when I looked at it, you were detecting (with some cleverness) when a workflow event came from a driver, or a controller, from memory to decide whether to save, or not? 

And for knowing whether to add model state errors etc?

From the triage conversation this was the area of concern that was blocking it, as Seb was concerned that workflows should have to know about these things.

It looks however you have changed it so that the workflow does not figure it out itself, but is passed some information informing it?

Can you just explain how it is operating in those cases? I'm looking but there are so many other scenarios you have fixed with this one, that I'm not quite sure I'm following the logic there

jtkech:
@deanmarcussen no problem, thanks to look at it again

I didn't have any time as #6356 was not so easy to fix ;) but i will give you soon more infos step by step

jtkech:
@deanmarcussen 

If i'm not clear / incomplete as i didn't have enough time we could have a chat / talk on gitter / skipe. So here some more infos as i can in a first step that may indirectly answer to your questions

- The main place where content events are triggered is in [this content handler](https://github.com/OrchardCMS/OrchardCore/blob/d08bf26e0fbb8f39dfdfa6be3203e292c72bd97a/src/OrchardCore.Modules/OrchardCore.Contents/Workflows/Handlers/ContentsHandler.cs#L70), there is also one triggered [in this driver](https://github.com/OrchardCMS/OrchardCore/blob/d08bf26e0fbb8f39dfdfa6be3203e292c72bd97a/src/OrchardCore.Modules/OrchardCore.Workflows/UserTasks/Drivers/UserTaskEventContentDriver.cs#L90) that is specific to the `UserTaskEvent`. That's why before i was talking about inline from an handler or inline from a driver, **but i harmonized the way a content event is triggered, so here let's only talk about inline from a content handler**, but we could just talk about inline from a content event.

- In fact we don't care from where e.g. the above content handler has been called, can be a controller, an api controller or another WF, the main concern is to know if this event is followed by a content activity without any intermediate waiting activity, **meaning that this activity is executed inline from the content handler that has triggered the content event**.

- When we trigger a content event with WF input data, before starting / resuming a related WF instance, the WF manager sends an `OnInputReceivedAsync()` event to all its activities. This is in the `ContentActivity` base class that we capture this WF input containing content event data. Then, because the WF manager rebuild all activities of a given WF instance when it resumes it, if there is an intermediate waiting activity we will loose these content event data, otherwise in the content activity we can use them, and this is how we detect that we are executing inline from a content event, if it is a starting event, if we are acting on the same correlated content item id and so on, and we can decide what we do.

- So, if a content activity is executed inline from a correlated (acting on the same content item id) content event, we don't allow to call content manager methods, for the same reason that e.g. a content handler is not allowed to do so. If it fails it populates the underlying model state in case the original initiator is e.g. a controller / api controller, if the initiator is another WF it will only populate the model state of a `NullUpdater`, for nothing but without failing (we could check if it is a null updater), and also pass the errors to the WF lastResult.

- Otherwise it means that the content activity acts as a mini controller, e.g. by retrieveing a given content item id by itself, and is allowed to call content manager methods, if it fails it doesn't populate any model state, it only return errors through the WF context lastResult

deanmarcussen:
@jtkech I am reading and thinking ;)

deanmarcussen:
> If not executed inline from a correlated content event, it means that the content activity is the original initiator by acting e.g. as a mini controller, e.g. by retrieveing a given content item id by itself using its expression property, and is allowed to call content manager methods, if it fails it doesn't populate any model state, it only returns errors through the WF context lastResult.

@jtketch I had to leave triage early last night, so we did not get to talk about this.

We do need to talk about it before merging, my concern is that we are fixing something, on top of fixing something else, with regard to the inline event, and the workflow acting as a mini controller. All else is fine.

It seems to me we have a fatal flaw in the idea of a workflow being a mini controller, and calling these content manager methods.

I went to use these a couple of days ago, and they were not adding up to me.

Let's talk about a `CreatedEvent`, or an `UpdatedEvent`, then followed by an `UpdateContentTask`. To me it makes sense, that the two would follow each other, and that the `UpdateContentTask` would not be calling the content manager to retrieve an item.

It seems a little weird that a workflow would be it's own initiator, or mini controller.

I realise we have recently added some of these tasks, `RetrieveContent` and possibly the `UpdateContent`, but that the `PublishContent` has existed for some time. But it concerns me that maybe the design of these new tasks is the issue here, requiring us to apply more fixes, on top of them.

However I did not write these tasks, nor design workflows, so maybe I am wrong here.

I would have to look back at O1 to see how the event -> task flow worked better.

jtkech:
@deanmarcussen 

> @jtketch I had to leave triage early last night, so we did not get to talk about this.

No problem, nothing urgent for me

For the other concerns i understand and i have no closed opinion, it only depends on what we allow to do.

Here some oher infos / examples

The `UpdateContentTask` has a script expression property to provide a contenItemId, if we are just after a content event the script can just retrieves it from the WF context input, and here in fact, as we are executing inline from the content event, we will use the content item of the WF input if its contentItemId it is the same as the one provided by the script. And if so we are not allowed to call content manager methods, not only `GetAsync()` but also `UpdateAsync()` and so on.

If there is an intermediate activity before, e.g. a timer event or any other waiting event, because when a WF instance is persisted we don't save the whole content item but only the contentItemId as a correlationId, the script can just be `correlationId()` meaning that it is retrieved from the WF context correlationId. But here we are not executing inline from the content event but in another scope, so we are allowed to call content manager methods including `GetAsync()` to retrieve the content item.

**In place of saying as a mini controller / initiator maybe better to say as a standalone activity as it has to retrieve the content item by itself**. And in the above case this is already the case but here the real initiator is still the content event but  the `UpdateContentTask` is executed in another scope, the same of the last intermediate event.

Another exemple is an `HttpRequestEvent` providing a correlationId and followed by an `UpdateContentTask` using it to retrieve the content item by itself. Here we are not executing inline from a content event, so as a standalone activity but the real initiator is the request event, there is always a WF event (of any type) before an `UpdateContentTask`.

Finally, the script expression can just be an hard coded string `4eczzkt0yb4znw8hwmpv01bccw`, maybe this is what we don't want but right now nothing prevents it. It's an higher level of autonomy but as said there is always a WF event before.

> It seems a little weird that a workflow would be it's own initiator, or mini controller.

See the above

What i said about the script expression, providing a contenItemId, is the same for the `RetrieveContentTask`. Maybe this task can be removed as, at least here, all other content tasks can retrieve a content item by themselves.

src/OrchardCore/OrchardCore.ContentManagement/DefaultContentManager.cs
```diff
@@ -242,6 +242,22 @@ public async Task<ContentItem> GetVersionAsync(string contentItemVersionId)
             return await LoadAsync(contentItem);
         }
 
+        public async Task SaveDraftAsync(ContentItem contentItem)
+        {
+            if (!contentItem.Latest || contentItem.Published)
```
sturatcliffe, line 247: Hi @jtkech - this line doesn't seem right as it prevents handlers being fired when adding a draft to an already published content item.

jtkech: @sturatcliffe Thanks for the feedback

- Normally only one content item version can be `Published` and only version can be the `Latest`.

- An active version is `Published` OR the `Latest` (so can be both). So a version is not active if it is neither `Published` nor the `Latest`, there is a pending audit trail module to manage these archived versions, otherwise we don't take them into account because they are considered as soft deleted versions.

- So let's only talk about active versions, if a version is `Published` and `Latest` it means that there is no active draft version for the related content item. When we publish it again, a new version is created that is marked as `Published` and `Latest`, and we set to false the `Published` and `Latest` of the previous version that becomes an archived version.

- If we save as a draft this `Published` and `Latest`, a new version is created that is only marked as `Latest`, the published version is still the same but no more the `Latest` and we now have an active draft version. When we edit a content item we always edit the `Latest` version, so here it will be the new active draft version. If you save it again as a draft version, we don't create a new version, we just update this version that is still the active draft version.

Sorry for the long story and there are different ways to describe content item versions ;)

**So what i can say here is that an active draft version is always the `Latest` and of course can't be `Published`**.

And `SaveDraftAsync()` is intended to call the related handlers only if we are saving a new active draft version.

I don't know in which context you are using `SaveDraftAsync()` but at this point your content item version needs to be the `Latest` and you have to take care that you have only one `Latest`. For this, e.g. when updating a content item, we first get this content item through the content manager by using the option `DraftRequired` this will create a non published version but that becomes the latest.

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Handlers/IContentHandler.cs
```diff
@@ -20,6 +20,8 @@ public interface IContentHandler
         Task ValidatedAsync(ValidateContentContext context);
         Task VersioningAsync(VersionContentContext context);
         Task VersionedAsync(VersionContentContext context);
+        Task DraftSavingAsync(SaveDraftContentContext context);
```
lazcool, line 23: Hi @jtkech,

Should these 2 be added to `IContentPartHandler` too?

Btw, we've been using this branch for a while now and we've very happy with it and had no issues.

jtkech: @lazcool Good catch, forgot it ;)

jtkech: Done ;)

> Btw, we've been using this branch for a while now and we're very happy with it and had no issues.

Cool thanks. If you have time can you add some comments on your own usage of this branch in the main thread? It may help us to decide if it is worth to review this PR again before merging, or just close it.

lazcool: Sorry, it's took a week, work's been manic 😅.

We use the new draft saved event, to sync the content item to a preview graph database (we basically replicate content items as nodes and relationships to published and preview graphs).

We'll also be using the new draft saved event to publish external events to Azure Event Grid when a new draft version is saved. (We have micro-service UI apps that listen for the events and then update their local stores with the content from the graphs.)

We currently publish draft events using the existing Orchard Core events, such as updated, but without this PR, we've found it impossible not to publish false-positive events, because the context just isn't there (and we have to do delayed processing of the events, which is complicated, messy and racey).

I have another question for you:
Would it make sense to fire the new draft save events when the user adds or deletes a taxonomy term?

At the moment, as soon as they, for example, add a new term, a new draft version of the taxonomy is saved, but no draft saved event gets fired. That means we don't get an opportunity to sync the new item to our graph or publish an event off the back of it.

Issue #5715, opened 2020-03-13
Preserve authentication tokens stored in external authentication cookies


ArieGato:
Hi,

I'm trying to get access to the tokens issued by an Open ID Connect server. I configured a Hybrid Flow client with code id_token token.
The following code is always returning null for the tokens.
```
    var token = await _httpContextAccessor.HttpContext.GetTokenAsync("id_token");
    var rt = await _httpContextAccessor.HttpContext.GetTokenAsync("refresh_token");
```

I tested the code in a test aspnet core web app and then it works fine. I already tested by setting SaveTokens to true., but no success. Any suggestions?

Thnx,
Arjan

kevinchalet:
Setting `SaveTokens` to true should definitely work. If it doesn't, we have a bug somewhere.

@ArieGato do you have a repro we can try to test it locally?

ArieGato:
I don't think it's a problem with OpenIdDict. And it's quite difficult to create a repro with OrchardCore with an external OpenId Connect server. What do you need to see the behavior? Can it be a clone of the OrchardCore repo?

kevinchalet:
@ArieGato yeah, a clone of OC using the OpenID client feature is fine.

ArieGato:
I forked OC here: [Repro](https://github.com/ArieGato/OrchardCore)
A readme file is added to the OrchardCore.Cms.Web project to reproduce. I'm assuming you have an OpenID Connect Server lying around somewhere ;-)

ArieGato:
@kevinchalet forgot to mention you for the repro

kevinchalet:
@ArieGato I'll take a look when I have a moment.

ArieGato:
> @ArieGato thanks, I'll take a look when I have a moment.

@kevinchalet did you see any chance to look into the issue?

kevinchalet:
I did: the tokens are correctly stored by the OIDC handler in the "external authentication cookie", but they are not flowed by Identity when creating the "application cookie" (which is a quite common issue and not specific to Orchard: https://github.com/aspnet/Identity/issues/628).

If we wanted to flow the tokens as authentication properties from the external to the application cookies, we'd have to override the default `SignInManager` used by Identity. It's doable, but it also means there's a high risk the final application cookie will be quite big.

[...]

PR #6057, opened 2020-04-29
Preserve external authentication tokens

ArieGato:
Add UserToken and UserTokenIndex
Add IUserAuthenticationTokenStore<IUser> to UserStore
Add boolean StoreExternalTokens to OpenIdClientSettings


sebastienros:
Just for my personal knowledge could you point me to some article explaining why someone would use this option?

Is it better to store these token as a separate document or as a property of the User document?
Is all the lifetime management of these tokens handled by ASP.NET Core Identity?
If the tokens were stored in the User document, I assume the index would not be necessary. How many token can be created for a single User (avg and worst case)?

ArieGato:
this PR is a result of issue https://github.com/OrchardCMS/OrchardCore/issues/5715.

I don't know of any article, but I can explain why someone would like to use this option.
I was trying to use this method to get the token: https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.authentication.authenticationtokenextensions.gettokenasync?view=aspnetcore-3.1

With this option you can use one of the tokens as bearer token for an external api's call that uses the same OpenId connect server.

I guess that there can be a maximum of 5 entries per user. There are three different types of tokens.  id_token, access_token and refresh_token. And the asp.net core identity also sets token_type and expires_at. I was thinking about skipping all other than the actual three possible token types. Maybe @kevinchalet can comment on that.

Concerning adding these to the User document I wasn't sure what this would do with performance. Each login will insert/update the tokens. If this is preferred I can rework it.

ns8482e:
Why would someone should store tokens in database? To access API when user is not around and not using an application?
 
I can understand to store it as claims in cookie while user is logged-in and using the application.

kevinchalet:
@ns8482e storing tokens in the main authentication cookie is not a viable option, as tokens tend to be big, which makes the resulting cookie super big (and browsers like Safari have very strict limits).

This PR looks good, but I'll let @sebastienros comment on the DB/YesSql aspect of it 😄

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
```diff
@@ -524,5 +531,111 @@ public async Task<IList<IUser>> GetUsersForClaimAsync(Claim claim, CancellationT
...
+
+        #region IUserAuthenticationTokenStore
+        public async Task<string> GetTokenAsync(IUser user, string loginProvider, string name, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            if (user == null)
+            {
+                throw new ArgumentNullException(nameof(user));
+            }
+
+            if (string.IsNullOrWhiteSpace(loginProvider))
```
kevinchalet, line 543: Technically, you should throw an `ArgumentException` instead of a `ArgumentNullException` here. Also, using `IsNullOrEmpty` is generally enough.

ns8482e:
> @ns8482e storing tokens in the main authentication cookie is not a viable option, as tokens tend to be big, which makes the resulting cookie super big (and browsers like Safari have very strict limits).
> 
> This PR looks good, but I'll let @sebastienros comment on the DB/YesSql aspect of it 😄

Storing tokens in database means, any custom module can query user database and get the access_token to access external APIs, for the given user regardless of user is currently logged in or not. Is that what this PR trying to achieve?  

Storing in cookie might not be viable option but it  guaranties that, any module can't access the access_token without user being logged in to the application.

@kevinchalet clarify if I misunderstood

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
agriffard: @ArieGato Please take the changes into account.

kevinchalet:
> Storing in cookie might not be viable option but it guaranties that, any module can't access the access_token without user being logged in to the application.

I'm not sure how this can be seen as a negative thing, as it enables new scenarios for people needing them.

If you don't want your OC app to be able to do that, then don't grant it the ability to use the refresh token grant, so that it will be unable to get new access tokens to perform actions on behalf of the user after the initial access tokens expired.

ns8482e:
@kevinchalet I am not seeing it as negative to enable new scenarios people are seeking.  

It's just seems to me that storing token in database is not insecure enough without proper token access management for modular application where one module can get and save the token and another module ( may be developed by different vendor from module gallery? ) has full access to it .  It should not allow all modules to access to the token by-default. 

There should be some privacy in-place and token access management should be controlled by site owners.

kevinchalet:
@ns8482e and how would keeping the tokens in the authentication cookies - instead of storing them in the DB - solve that? A module has full access to the ASP.NET context and could extract the tokens from the `HttpContext` using `HttpContext.AuthenticateAsync()` just as easily as from the DB using the `UserManager` APIs.

OrchardCore has never been designed to prevent such attacks: once a malicious module is enabled, it's not just your entire website that can be compromised, but also any process running on the host under the same identity as your OC website. It's certainly not specific to OrchardCore (nor to its predecessor that worked roughly the same way): the same exact thing would happen if you used a malicious NuGet package in any non-Orchard app.

ArieGato:
Can we store an id and encryption key in a session cookie, or add claims for encryption via claims transformation? That way the tokens can only be decrypted by the users process.

ns8482e:
> @ns8482e and how would keeping the tokens in the authentication cookies - instead of storing them in the DB - solve that? A module has full access to the ASP.NET context and could extract the tokens from the `HttpContext` using `HttpContext.AuthenticateAsync()` just as easily as from the DB using the `UserManager` APIs.

Correct, But its different - to access token from cookie, request has to initiated by logged in user to read token from cookie. Where as once token is stored in database - request can be initiated by any module without user's request.

kevinchalet:
@ns8482e as I said, protecting against malicious modules is not part of OC's threat model. It's your responsibility to check whether the modules you're using are safe or not.

ns8482e:
> protecting against malicious modules is not part of OC's threat model

Then don't include code that could allow to build malicious modules. Let this be separate as extension outside orchard core

kevinchalet:
The simple fact OC modules are .NET assemblies executed in the same process - and under the same identity - as the rest of your app allows them to do nasty things. This PR doesn't make that better or worse.

sebastienros:
@ns8482e this is not malicious code, or our ContentManager is malicious code, our User module is malicious code, ... As Kevin said, the security boundary is to only install or enabled modules that only do what they are meant to do. If a module can format the hard disk, or export the db to an external server, then it's not even a security issue as long as this is the public intent of this module.

ns8482e:
May be I'm not getting the point across well enough. It’s not about malicious module, it’s about protecting access_token, storing token in database is not good idea in multi module application where module can be written by different vendors. 

If we store in db then there should be token access management and only trusted modules can get the access token.

Or orchard core can provide interface to Token store, let site owners implement token storage outside orchard content db

ns8482e:
For example, may be I don’t want SQL query or lucene query modules to read access token

kevinchalet:
> If we store in db then there should be token access management and only trusted modules can get the access token.

What a brilliant idea! Go ahead and send a PR so you can show us how you'd implement that 👍

ns8482e:
IMO access_token related to login session should not be stored in content db when db can be accessible by multiple vendor modules that can be misused or compromised intentionally or unintentionally. 

Instead it should be deferred to site owners to implement their own storage mechanism for token store that can be tied to the session and not the the user in db.

From documentation https://docs.microsoft.com/en-us/aspnet/core/security/authentication/social/additional-claims?view=aspnetcore-3.1

`Use a custom ITicketStore for the Cookie Authentication Middleware's SessionStore to store identity across requests. Preserve large quantities of identity information on the server while only sending a small session identifier key to the client`

SessionStore can be implemented to use database other than content db, caching service or vault.

kevinchalet:
And what would prevent a "misused or compromised" (= malicious) module from replacing `ITicketStore` by an instance intercepting the tokens resolved from the external storage when they are retrieved (or added)? Strictly nothing.

@ArieGato can you please change the two things we mentioned so we can merge this PR?

ns8482e:
> And what would prevent a "misused or compromised" (= malicious) module from replacing `ITicketStore` by an instance intercepting the tokens resolved from the external storage

Only the session identifier mapped to token that stored in external  storage and the session identifier comes from cookie stored user's computer, so modules can't get to access_token until user requests for it. and `ITicketStore`  doesn't provides a  way to read access_token for all sessions (unlike content manager allows to query all user's and their tokens)

kevinchalet:
> Only the session identifier mapped to token that stored in external storage and the session identifier comes from cookie stored in user's computer, so modules can't get to access_token until user requests for it.

"user requests for it?"... that's not how things work. Once created and as long as it's still valid, a cookie is automatically sent to the website every time the user visits it, which gives a potential malicious OC module a chance to extract its session identifier and resolve the actual cookie payload from whatever backing database/vault you use. The user has **no influence** on this mechanism.

Access tokens are typically very short-lived (access tokens issued by Azure AD last for 1 hour, which is a very common value in the industry): if you managed to steal an access token that is still valid from the database, this means the user visited the website less than an hour ago, and we're back to my initial point. The same point applies to refresh tokens, that just generally live longer.

> and ITicketStore doesn't provides a way to read access_token for all sessions (unlike content manager allows to query all user's and their tokens)

Heh, the fact `ITicketStore` doesn't expose a method to list all the existing sessions doesn't mean you're protected. Let's say you store your authentication cookies as secrets in Azure Key Vault: there's absolutely nothing preventing a malicious module from stealing the access token or the credentials used to interact with AKV (when using Managed Service Identity, the client secret is exposed as an environment variable named `MSI_SECRET`). Once obtained, these credentials can be used to perform the same actions as the legit app, including _getting_ or _enumerating_ the secrets if these permissions have been granted. Using or not using `ITicketStore` doesn't make any difference.

I'm sorry, but what you suggest doesn't offer a substantial security improvement: once a malicious module is added to an OC app, all hope is lost.

Here's **my final word**: if you don't like this opt-in feature, don't use it.

If you're still convinced you understand all this stuff better than the ASP.NET team, go tell them their token storage mechanism is stupid.

ns8482e:
> "user requests for it?"... that's not how things work. Once created and as long as it's still valid, a cookie is automatically sent to the website every time the user visits it,

That's the whole point I am arguing, Back to my original comment,  Is this feature intend to use access_token when user is not around ( not visiting the site)? Any module should have access to access_token only when user is vising the site, it should not be as easy as easy to just query the db to get the token.

kevinchalet:
@ns8482e [I already answered this question](https://github.com/OrchardCMS/OrchardCore/pull/6057#issuecomment-623137961): yes.

ArieGato:
I made the changes @kevinchalet mentioned. I'm still waiting on the verdict concerning the YesSql part. @sebastienros should I add it to the user document?

sebastienros:
This PR is about IUserAuthenticationTokenStore, not ITokenStore, right? And ITokenStore is used to override where tokens are stored for a user session, for instance in ASP.NET Session instead of a cookie?

kevinchalet:
@sebastienros exact 😎

src/OrchardCore.Modules/OrchardCore.OpenId/Configuration/OpenIdClientConfiguration.cs
```diff
@@ -68,6 +68,7 @@ public void Configure(string name, OpenIdConnectOptions options)
             options.GetClaimsFromUserInfoEndpoint = true;
             options.ResponseMode = settings.ResponseMode;
             options.ResponseType = settings.ResponseType;
+            options.SaveTokens = settings.StoreExternalTokens;
```
sebastienros, line 71: Where is it used?

ArieGato: Setting SaveTokens to true, the authentication tokens are preserved in the ExternalLoginInfo in the AccountController.

ArieGato:
> Based on the public API, it seems that tokens can't be read/stores without an `IUser` instance so I think it makes sense to just use the current User object. For instance adding a `Tokens` property to the User class. Then the new tables can be removed, and it makes the code much simpler.

So no problem that the user is updated multiple times per login?

sebastienros:
@kevinchalet your concern about storing everything in the cookie is mitigated with custom implementations of ITokenStore, I assume there is one somewhere on the Internet that stores then in IDistributedCache, or indirectly in the session (which can be backed by for instance).

kevinchalet:
@sebastienros I'm well aware of how it works: eventually, it's all about defining what your priorities are: are you willing to have increased latency just for the sake of storing more claims (or bigger claims) in your authentication cookies?

Not to mention it's another thing to configure, as ASP.NET's default "distributed cache" implementation is in-memory.

sebastienros:
I am well aware you are ;) But it was a question because I am not very familiar with this :/ I want to know the options if someone is asking. Also if one is based on IDistributedCache then on one node everything is still in memory, so no latency, but actually faster as not in a cookie.

kevinchalet:
> Also if one is based on IDistributedCache then on one node everything is still in memory, so no latency, but actually faster as not in a cookie.

It can't be faster: even with an `ITicketStore` backed by an in-memory `IDistributedCache`, the authentication ticket is still encrypted using ASP.NET Core Data Protection before being stored in the cookie. The only difference is that you'll store less claims in the cookie (basically, just the special "session ID" claim used by the cookies middleware), but it will still involve crypto operations for both creating and validating the cookies (+ the cache lookup needed to resolve the rest of the cookie payload).

> But it was a question because I am not very familiar with this :/ I want to know the options if someone is asking.

It's definitely something that could be implemented - but likely not in this PR, which is kinda unrelated.

Actually, it's pretty much like cookie validation: by default, ASP.NET Core Identity only re-validates authentication cookies after 30mn. Of course, if you need immediate revocation, you can opt for a shorter duration, but then, you have to pay for it 😄 
That's also the reason why access tokens tend to be non-revocable in perf-sensitive applications.

ArieGato:
I think storing the sessions in memory is not very convenient. For instance in a kubernetes cluster, pods can be killed for no reason at all. Just for being a pod ;-) All session info will be lost. So some sort of persistence would essential for this.

sebastienros:
@ArieGato it's not convenient if you don't enable the Distributed storage of these tokens, or if you don't have session afinity if you keep them in memory.

@kevinchalet you are the first to tell us that big cookies are bad for perf. Here I was saying that it's faster as you don't have to send them back and forth, and the cost to decrypt them is the same from memory or from cookie, so I still think it can be faster is the claims are in memory instead of the cookies, as long as we are using a single node. Opt-in, and not part of this PR, just triggered some related questions.

kevinchalet:
> @kevinchalet you are the first to tell us that big cookies are bad for perf.

I never said that (it's **your** point, actually 😅).

The main problem with cookies is that [browsers have very strict size limits](http://browsercookielimits.squawky.net/) (per-cookie or per-domain). For instance, authentication will stop working on Safari if you exceed 4KB per domain, which is ridiculously small and super easy to hit (there are tons of tickets on dotnet/aspnetcore or aspnet/security about that).

> so I still think it can be faster is the claims are in memory instead of the cookies, as long as we are using a single node

I guess you'll have to create a benchmark to confirm that :trollface: 

More seriously, is there a realistic case where you'd use a non-persistant/in-memory cache for your cookies? As @ArieGato said, it would offer a terrible UX, as yours users would have to re-authenticate every time your host decides to stop or recycle your app.

kevinchalet:
Fun fact: in ASP.NET Core 1.0/1.1/2.0, the default cookie manager used by the cookies authentication middleware (doing chunking under the hood) was configured to throw an exception if it was unable to re-assemble the authentication cookie. If the total size of your cookies exceeded the browsers limits, some of the chunked cookies were simply ignored by your browser, which caused an exception for every subsequent request, making the entire website unusable for impacted users (typically, iOS users) 🤣

 This was only fixed in 2.1: https://github.com/aspnet/Security/pull/1482

sebastienros:
Sorry if I misserderstood then.

Yeah I was involved in the cookie split issue.

Didn't think about the scenario when the host is recycled ... or a tenant restarted in OC's case too.

ArieGato:
I reworked the UserTokens into the User document. @sebastienros Although I understand the arguments, I think it's not very clean that the user token collection is now always included with user. The tokens in the collection are encrypted and thus unusable.

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
```diff
@@ -524,5 +531,106 @@ public async Task<IList<IUser>> GetUsersForClaimAsync(Claim claim, CancellationT
...
+        public Task<string> GetTokenAsync(IUser user, string loginProvider, string name, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            if (user == null)
+            {
+                throw new ArgumentNullException(nameof(user));
+            }
+
+            if (string.IsNullOrEmpty(loginProvider))
+            {
+                throw new ArgumentException(nameof(loginProvider));
```
kevinchalet, line 545: @ArieGato FYI, this overload takes an exception message, not just a parameter name like `ArgumentException`. You'll probably want something like this:

```suggestion
                throw new ArgumentException("The login provider cannot be null or empty.", nameof(loginProvider));
```

Same for the other occurrences.

kevinchalet:
@sebastienros haha, no problem. Let me know if you'd like to discuss more about this scenario 😄

PBMikeW:
Thanks for this PR @ArieGato solved my problem, Some documentation on how to decrypt / use the token would have saved me a couple of days trying to access Power Bi APi!

PR #6058, opened 2020-04-29
Fixing the file naming convention - Edit.cshtml not Editor.cshtml

grexican:
Looking at the source, and having just run through an example, looks like editor templates should end it `.Edit.cshtml` not `.Editor.cshtml`


grexican:
Hmm I also see that the naming convention is wrong. It's not HtmlBody it's HtmlBody**Part**.

I'll update the PR

grexican:
ok, PR is ready.

PR #6067, opened 2020-04-30
Patch 1

grexican:
Found two more markdown docs that mislabeled as `.Editor.` I think this is all of them. I searched the repo source this time and found only three files that contained _Editor__. This PR contains two of them, and the other was already fixed in a previously merged PR.


hishamco:
Thanks for contribution @grexican

PR #6072, opened 2020-04-30
Fix login redirect for external providers

### LINKED ISSUE #6049: Bug with External (Azure AD) authentication redirect
When using an external login provider to login to a site, on the first request (after the login callback) the Page does not reflect the fact that the user is authenticated. Refreshing the page manually works as expected. 

I tried stepping through the code and everything looks good.
https://github.com/OrchardCMS/OrchardCore/blob/acee23ef121a09568198ebda164d591ae9f084ed/src/OrchardCore.Modules/OrchardCore.Users/Controllers/AccountController.cs#L387

This line get's called with `returnUrl= null` which triggers this line https://github.com/OrchardCMS/OrchardCore/blob/acee23ef121a09568198ebda164d591ae9f084ed/src/OrchardCore.Modules/OrchardCore.Users/Controllers/AccountController.cs#L276

But, on the frontend, I still see the Log In button as you can see in the gif below. However, clicking on the URL bar and hitting enter loads the user just fine. The gif below is using Azure AD authentication but we saw the same behavior with Github authentication.

oc_login_bug

jptissot:
fix #6049

I tested Azure AAD and it works as expected. I assume the other providers will work as well.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.OpenId/Manifest.cs
+++ b/src/OrchardCore.Modules/OrchardCore.OpenId/Manifest.cs
@@ -20,7 +20,7 @@ using OrchardCore.OpenId;
     Name = "OpenID Client",
     Category = "OpenID Connect",
     Description = "Authenticates users from an external OpenID Connect identity provider.",
-    Dependencies = new[] { OpenIdConstants.Features.Core }
+    Dependencies = new[] { OpenIdConstants.Features.Core, "OrchardCore.Users" }
 )]
 
 [assembly: Feature(
--- a/src/OrchardCore.Modules/OrchardCore.OpenId/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.OpenId/Startup.cs
@@ -73,19 +73,14 @@ namespace OrchardCore.OpenId
                 ServiceDescriptor.Scoped<INavigationProvider, AdminMenu>(),
             });
 
-            services.Configure<CookiePolicyOptions>(options =>
+            services.ConfigureApplicationCookie(options =>
             {
-                options.OnAppendCookie = cookieContext =>
-                {
-                    // Disabling same-site is required for OpenID's module prompt=none support to work correctly.
-                    // Note: it has no practical impact on the security of the site since all endpoints are always
-                    // protected by antiforgery checks, that are enforced with or without this setting being changed.
-                    // 2020-03-23; Moved the SameSiteNode.None here, this will require that the site runs on HTTPS;
-                    if (cookieContext.CookieName.StartsWith("orchauth_"))
-                    {
-                        cookieContext.CookieOptions.SameSite = SameSiteMode.None;
-                    }
-                };
+                // Disabling same-site is required for OpenID's module prompt=none support to work correctly.
+                // Note: it has no practical impact on the security of the site since all endpoints are always
+                // protected by antiforgery checks, that are enforced with or without this setting being changed.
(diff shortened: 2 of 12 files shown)


src/OrchardCore.Modules/OrchardCore.OpenId/Startup.cs
```diff
@@ -73,19 +73,14 @@ public override void ConfigureServices(IServiceCollection services)
...
-            services.Configure<CookiePolicyOptions>(options =>
+            services.ConfigureApplicationCookie(options =>
             {
-                options.OnAppendCookie = cookieContext =>
-                {
-                    // Disabling same-site is required for OpenID's module prompt=none support to work correctly.
-                    // Note: it has no practical impact on the security of the site since all endpoints are always
-                    // protected by antiforgery checks, that are enforced with or without this setting being changed.
-                    // 2020-03-23; Moved the SameSiteNode.None here, this will require that the site runs on HTTPS;
-                    if (cookieContext.CookieName.StartsWith("orchauth_"))
```
jptissot, line 84: @deanmarcussen I'm curious to know if this has any other implications ? This one was disabling the SameSiteMode only for the orcauth_ cookies

deanmarcussen: For me I don't think this changes anything. It is still configuring the application cookie, it's just done at the time the cookie is generated, rather than doing it later, in the `CookieMiddleware`

/cc @MatthijsKrempel

src/OrchardCore.Modules/OrchardCore.OpenId/Manifest.cs
```diff
@@ -20,7 +20,7 @@
     Name = "OpenID Client",
     Category = "OpenID Connect",
     Description = "Authenticates users from an external OpenID Connect identity provider.",
-    Dependencies = new[] { OpenIdConstants.Features.Core }
+    Dependencies = new[] { OpenIdConstants.Features.Core, "OrchardCore.Users" }
```
kevinchalet, line 23: Erf, why this change? Decoupling the OpenID module from Users was deliberate and reintroducing the coupling is not something we should be doing.

jtkech: @jptissot @deanmarcussen just for infos, if the added dependency is just to register / configure things after, could be adjusted with the startup `Order` / `ConfigureOrder` properties.

jptissot:
I will double check if removing the line in the user's module fixes the issue. What are the security implications of using SameSiteMode.None?

deanmarcussen:
Happy for this to be reverted (at least temporarily)

@jptissot for info @kevinchalet suggestion for `SameSiteMode.Lax` in the `Users` `Startup` resolves any chrome warnings related to same site cookies over http. 

So if this is also enough to solve your external login issue as well, then it is indeed the better fix.

@jtkech Yes, I assume the dependency added was just for ordering, so if we still need to have the `SameSiteMode.None` for the external providers, the `Order` over a dependency should be enough.

@jptissot will you revert it? and are able test against the external provider

agriffard:
Reverted.

src/OrchardCore.Modules/OrchardCore.OpenId/Manifest.cs
jptissot: Yes that was the only reason, I was not aware of the Order / ConfigureOrder properties, thanks.

jptissot:
I will test with the removed xookie policy

PR #6077, opened 2020-05-01
Set isNew to true when building editor

### LINKED ISSUE #6076: Adding widget to flowpart does not consider the IsNew context used from driver.
Adding a widget to a flow part with various fields that allow setting default values when new instance is created  (like numeric field, boolean field) is not respected.
Example:
Add a widget with a nemric field and set its default value to 100\. Then add the widget to the flow part .
The default value in editor is ignored.
This is because of 
 
https://github.com/OrchardCMS/OrchardCore/blob/acee23ef121a09568198ebda164d591ae9f084ed/src/OrchardCore.Modules/OrchardCore.Widgets/Views/ContentCard.cshtml#L22
 
setting the Isnew context variable to false.

giannik:
fixes  #6076

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Widgets/Views/ContentCard.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Widgets/Views/ContentCard.cshtml
@@ -19,7 +19,7 @@
         }
 
         //Build Editor for Content Item
-        dynamic contentItemEditor = await ContentItemDisplayManager.BuildEditorAsync(contentItem, updater, false, "", Model.PrefixValue);
+        dynamic contentItemEditor = await ContentItemDisplayManager.BuildEditorAsync(contentItem, updater, true, "", Model.PrefixValue);
 
         //We don't show Actions and Side bar the parent editor has its own buttons.
         contentItemEditor.Actions = null;


hishamco:
@agriffard please merge if it's looks good to you

src/OrchardCore.Modules/OrchardCore.Widgets/Views/ContentCard.cshtml
```diff
@@ -19,7 +19,7 @@
         }
 
         //Build Editor for Content Item
-        dynamic contentItemEditor = await ContentItemDisplayManager.BuildEditorAsync(contentItem, updater, false, "", Model.PrefixValue);
+        dynamic contentItemEditor = await ContentItemDisplayManager.BuildEditorAsync(contentItem, updater, true, "", Model.PrefixValue);
```
ns8482e, line 22: This is for existing content and isnew should be false

deanmarcussen: Could you look into what the correct fix is Niraj, if this is not it

ns8482e:
```suggestion

        //AJAX request is new request and will not have CollectionShape.
        var isNew = Model.CollectionShape == null ? true : false ;
        dynamic contentItemEditor = await ContentItemDisplayManager.BuildEditorAsync(contentItem, updater, isNew, "", Model.PrefixValue);
```

ns8482e: All new widgets are added using ajax call. In ajax call it doesn't have CollectionShape

ns8482e:
@agriffard  @deanmarcussen little problem! suggestion is not applied yet, now its broken for existing widgets

agriffard:
Reverted it on dev.

deanmarcussen:
Thank you @ns8482e for the quick eyes (my bad for approving it to quickly)

Can you do another pr with the correct fix from your suggestion?

ns8482e:
@deanmarcussen  PR created #6084

PR #6084, opened 2020-05-02
Set isNew when creating new Flow Part Widgets

### LINKED ISSUE #6076: Adding widget to flowpart does not consider the IsNew context used from driver.

ns8482e:
Fixes #6076

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Widgets/Views/ContentCard.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Widgets/Views/ContentCard.cshtml
@@ -19,7 +19,9 @@
         }
 
         //Build Editor for Content Item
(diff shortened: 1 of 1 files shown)


deanmarcussen:
Thank you Niraj

PR #6106, opened 2020-05-05
Fixes bulk actions filtering

jtkech:
In the admin content items index page script, bulk actions are selected by their `".dropdown-menu .dropdown-item"` classes, the problem is that there are other items with the same classes, e.g. the filter actions, the actions of a given item and so on.

So as soon as you have selected multiple items, even if you are not cliking on a bulk action, but on another action, this action is posted **but also the bulk one, with no action but posted**, the first one is done and redirect first, but because of the 2nd one you may have some disconnected errors, a flashing confirm dialog and so on.

**So here i filtered the items to only retrieve those that have a `data-action` attribute.**

Todo: **There are many other index pages to update**

**Update**: Done

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Views/Admin/List.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Views/Admin/List.cshtml
@@ -149,16 +149,16 @@
                 return $(this).data("action");
             })
             .one("click", function () {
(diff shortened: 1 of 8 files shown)


deanmarcussen:
Nice one @jtkech I saw the same issue on https://github.com/OrchardCMS/OrchardCore/pull/5862 and fixed it in a different (but similar way), but only for the `AdminList`.

I will check to see that this way, which is nicer, will work over there (I'm sure it does)

Issue #6033, opened 2020-04-25
Consider using the default same-site mode in the OpenID module


kevinchalet:
To allow SPAs to use silent authorization requests (i.e `prompt=none` requests sent in an iframe), the OpenID module currently disables the same-site feature, that would prevent the Orchard authentication cookie from being sent as part of the iframe request.

This works well and was until now the recommended way for SPAs to get new access tokens without using refresh tokens. Unfortunately, this approach will soon be impossible, due to major browsers now banning third-party cookies:
  - https://webkit.org/blog/10218/full-third-party-cookie-blocking-and-more/
  - https://techcrunch.com/2020/01/14/google-wants-to-phase-out-support-for-third-party-cookies-in-chrome-within-two-years/

We should consider removing the same-site logic and encourage SPAs to use refresh tokens, which is consistent with what other actors in the industry are heading to: https://auth0.com/blog/securing-single-page-applications-with-refresh-token-rotation/

rneto:
For those who cannot currently deal with a secure token refresh flow in SPAs, another alternative might be to host Orchard Core CMS and the SPA in a _same-site_ enviroment like _spa.web.dev_ and _orchard.web.dev_.

PR #6109, opened 2020-05-05
Fix login redirect for external providers

### LINKED ISSUE #6049: Bug with External (Azure AD) authentication redirect

jptissot:
fixes #6049

I tested this with AAD on port 5001 and everything worked as expected. I removed the SameSiteMode.None from the OpenId Module but I did not test that. I assume it would work.

/cc @kevinchalet @deanmarcussen

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Users/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Startup.cs
@@ -160,7 +160,8 @@ namespace OrchardCore.Users
                 // 2019-12-10; Removed, since https://github.com/aspnet/Announcements/issues/390
                 // 2020-02-17; Reenabled since we have compensation logic for backwardscompatibility
                 // 2020-03-23; Moved the SameSiteNode.None to the Startup of the OIDC Server
-                options.Cookie.SameSite = SameSiteMode.Strict;
+                // 2020-05-05; Removed to fix external login providers
+                // options.Cookie.SameSite = SameSiteMode.Strict;
             });
 
             services.AddSingleton<IIndexProvider, UserIndexProvider>();
--- a/src/OrchardCore.Modules/OrchardCore.Users/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Startup.cs
@@ -153,15 +153,6 @@ namespace OrchardCore.Users
 
                 options.LoginPath = "/" + userOptions.Value.LoginPath;
                 options.AccessDeniedPath = "/Error/403";
-
-                // Disabling same-site is required for OpenID's module prompt=none support to work correctly.
-                // Note: it has no practical impact on the security of the site since all endpoints are always
-                // protected by antiforgery checks, that are enforced with or without this setting being changed.
-                // 2019-12-10; Removed, since https://github.com/aspnet/Announcements/issues/390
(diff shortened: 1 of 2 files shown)


kevinchalet:
> I removed the SameSiteMode.None from the OpenId Module but I did not test that.

It will kill the silent authentication feature for SPAs, but it's unavoidable anyway:
https://github.com/OrchardCMS/OrchardCore/issues/6033.

src/OrchardCore.Modules/OrchardCore.Users/Startup.cs
```diff
@@ -160,7 +160,8 @@ public override void ConfigureServices(IServiceCollection services)
                 // 2019-12-10; Removed, since https://github.com/aspnet/Announcements/issues/390
                 // 2020-02-17; Reenabled since we have compensation logic for backwardscompatibility
                 // 2020-03-23; Moved the SameSiteNode.None to the Startup of the OIDC Server
-                options.Cookie.SameSite = SameSiteMode.Strict;
+                // 2020-05-05; Removed to fix external login providers
```
kevinchalet, line 163: Personally, I'd remove the entire code comment and forget this part of our history :trollface:

deanmarcussen:
Works w/both http and https on Chrome 81.0.4044.129

src/OrchardCore.Modules/OrchardCore.Users/Startup.cs
jptissot: Done

jptissot:
> > I removed the SameSiteMode.None from the OpenId Module but I did not test that.
> 
> It will kill the silent authentication feature for SPAs, but it's unavoidable anyway:
> #6033.

Should I undo the change to the OpenId module ?

kevinchalet:
@jptissot no no, keep it, thanks.

rneto:
> > I removed the SameSiteMode.None from the OpenId Module but I did not test that.
> 
> It will kill the silent authentication feature for SPAs, but it's unavoidable anyway: #6033.

@kevinchalet, any suggestions on how to deal with that silent authentication issue in SPAs after this regression change? I'm using a basic Orchard Core CMS (builder.Services.AddOrchardCms()) as my Identity Server using OpenID Connect.

kevinchalet:
@rneto the recommended approach is described in the link you quoted.

rneto:
Thanks @kevinchalet, I also think it's time to token refresh in SPAs.

Issue #6121, opened 2020-05-06
Workflow Singleton Discussion


jtkech:
- We already saw that if a WF instance is already started but not yet stored, an incoming event may start another instance, even if the related WF type is marked as a Singleton. Maybe we will fix it in a dedicated PR.

- Then working on workflows, i also saw that if a WF type, marked as a Singleton, has 2 starting event activities, each event create a new instance even if there is already an existing one but started by the other event. I expected that the previous instance would have been used.

    So, beyond the Singleton concern, we can't have 2 starting event activities followed by a `Join` activity, it will never complete because each event will start a separate WF instance. So, we have to find a single main starting event, followed by a `Fork` activity, and this is only at this point that we can wait for multiple waiting events (not used as starting events), and then a `Join` activity.

    I'm okay with this but just to know if it is intended to work like this, otherwise i could try to fix it.

- **Update**: But allowing different incoming events, that may happen concurrently, acting on the same WF instance will be hard to implement safely. So i'm fine with how it is implemented, meaning that multiple starting events define somewhere multiple workflows in only one definition.

    Maybe add to the doc that `Singleton` means only one instance but only per starting event activity. And then that we can't `Join` starting events, only waiting events after a `Fork`.

jtkech:
related to #5997

PR #6124, opened 2020-05-06
Allow a Timer waiting activity in any Workflow

### LINKED ISSUE #5996: Workflow Correlation
If a starting event activity is triggered by `TriggerEventAsync()` with a non null `correlationId `, as done for a `Content` event where the `correlationId ` is set to the `ContentItemId`.

Then a subsequent Timer activity (so not used as a starting activity) will never be resumed, this because the Timer BgTask uses `TriggerEventAsync()` with a null `correlationId `.

In `TriggerEventAsync()` we already use `isExclusive` (false by default), maybe we need another option `isCorrelable` (true by default). So that any timer activity in a blocking state will be resumed even if the tirmer event is always triggered with a null `correlationId `.

Notice that both options seem to be closely tied to the `TimerEvent`.

jtkech:
Fixes #5996 

A timer event activity is always started (if used as a starting activity) / resumed (if used as a waiting event) by a timer event triggered by a background task that is not aware of any correlationId. So currently we can't have e.g. a Content Event that is correlated to a content item, followed by a waiting timer event.

~~So here when the previous activity triggers the timer activity we memorize the current correlationId and we set the WF context correlationId to null, so that the blocking activity is persisted with no correlationId and then can be resumed each minute. Then when expired we restore the original correlationId.~~

~~This is one way to fix it, but there are other ways and it may depend on what we think about #6121~~

**Update**: Use instead a new `isAlwaysCorrelated` parameter passed to `TriggerEventAsync()`


agriffard:
Conflicts to solve

Issue #6133, opened 2020-05-08
RazorPages with IContentItemDisplayManager.UpdateEditorAsync  ModelState is always valid in Admin


ns8482e:
When used RazorPages in Admin to Edit content item using IContentItemDisplayManager - `BuildEditorAsync` and `UpdateEditorAsync` ModelState is not reflecting the actual errors

It always returns valid.

Issue #5667, opened 2020-03-04
Error using _Layout.cshtml in a custom module


dodyg:
My custom module uses a page and when I try to access, I got

```
InvalidOperationException: The following sections have been defined but have not been rendered by the page at '/Areas/My.Theme.Bootstrap/Pages/Shared/_Layout.cshtml': 'Header'. To ignore an unrendered section call IgnoreSection("sectionName").
```
Although I have this statement already at my theme

![section](https://user-images.githubusercontent.com/1761249/75856457-10e87400-5dfd-11ea-9015-4ea37053f6fa.png)

![theme](https://user-images.githubusercontent.com/1761249/75856557-42f9d600-5dfd-11ea-8dd8-971abc6cc6e9.png)

This is where my view is located within the theme.

dodyg:
My `_ViewImports.cshtml` has

```

@inherits OrchardCore.DisplayManagement.Razor.RazorPage<TModel>
@addTagHelper *, Microsoft.AspNetCore.Mvc.TagHelpers
@addTagHelper *, OrchardCore.DisplayManagement
@addTagHelper *, OrchardCore.ResourceManagement
@addTagHelper *, OrchardCore.Menu


```

dodyg:
Full stack is 

```
System.InvalidOperationException: The following sections have been defined but have not been rendered by the page at '/Pages/Shared/_Layout.cshtml': 'Header'. To ignore an unrendered section call IgnoreSection("sectionName").
   at Microsoft.AspNetCore.Mvc.Razor.RazorPage.EnsureRenderedBodyOrSections()
   at Microsoft.AspNetCore.Mvc.Razor.RazorView.RenderLayoutAsync(ViewContext context, ViewBufferTextWriter bodyWriter)
   at Microsoft.AspNetCore.Mvc.Razor.RazorView.RenderAsync(ViewContext context)
   at Microsoft.AspNetCore.Mvc.ViewFeatures.ViewExecutor.ExecuteAsync(ViewContext viewContext, String contentType, Nullable`1 statusCode)
   at Microsoft.AspNetCore.Mvc.ViewFeatures.ViewExecutor.ExecuteAsync(ViewContext viewContext, String contentType, Nullable`1 statusCode)
[...]

PR #6135, opened 2020-05-08
RazorPage enhancements

ns8482e:
Fixes: #6133 
Fixes: #5667 
 
1. Sync Zone enhancements made in RazorPage.cs to Page.cs
2. Added PageModelUpdater for PageModel.

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement/ModelBinding/ModelBinderAccessorFilter.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/ModelBinding/ModelBinderAccessorFilter.cs
@@ -5,7 +5,7 @@ using Microsoft.Extensions.DependencyInjection;
 
 namespace OrchardCore.DisplayManagement.ModelBinding
 {
-    public class ModelBinderAccessorFilter : IActionFilter , IPageFilter
+    public class ModelBinderAccessorFilter : IActionFilter, IPageFilter
     {
         public void OnActionExecuted(ActionExecutedContext context)
         {
@@ -20,7 +20,12 @@ namespace OrchardCore.DisplayManagement.ModelBinding
                 modelBinderAccessor.ModelUpdater = new ControllerModelUpdater(controller);
             }
         }
+
         public void OnPageHandlerSelected(PageHandlerSelectedContext context)
+        {
+        }
+
+        public void OnPageHandlerExecuting(PageHandlerExecutingContext context)
         {
             var page = context.HandlerInstance as PageModel;
 
@@ -29,10 +34,6 @@ namespace OrchardCore.DisplayManagement.ModelBinding
                 var modelBinderAccessor = context.HttpContext.RequestServices.GetRequiredService<IUpdateModelAccessor>();
                 modelBinderAccessor.ModelUpdater = new PageModelUpdater(page);
             }
-        }
-
-        public void OnPageHandlerExecuting(PageHandlerExecutingContext context)
-        {
         }
 
         public void OnPageHandlerExecuted(PageHandlerExecutedContext context)
--- a/src/OrchardCore/OrchardCore.DisplayManagement/ModelBinding/ModelBinderAccessorFilter.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/ModelBinding/ModelBinderAccessorFilter.cs
@@ -1,10 +1,11 @@
-﻿using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.AspNetCore.Mvc.RazorPages;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace OrchardCore.DisplayManagement.ModelBinding
 {
-    public class ModelBinderAccessorFilter : IActionFilter
+    public class ModelBinderAccessorFilter : IActionFilter , IPageFilter
     {
         public void OnActionExecuted(ActionExecutedContext context)
         {
@@ -18,6 +19,24 @@ namespace OrchardCore.DisplayManagement.ModelBinding
                 var modelBinderAccessor = context.HttpContext.RequestServices.GetRequiredService<IUpdateModelAccessor>();
                 modelBinderAccessor.ModelUpdater = new ControllerModelUpdater(controller);
             }
+        }
+        public void OnPageHandlerSelected(PageHandlerSelectedContext context)
+        {
+            var page = context.HandlerInstance as PageModel;
+
+            if (page != null)
+            {
+                var modelBinderAccessor = context.HttpContext.RequestServices.GetRequiredService<IUpdateModelAccessor>();
+                modelBinderAccessor.ModelUpdater = new PageModelUpdater(page);
+            }
+        }
+
+        public void OnPageHandlerExecuting(PageHandlerExecutingContext context)
+        {
+        }
+
+        public void OnPageHandlerExecuted(PageHandlerExecutedContext context)
+        {
         }
     }
 }
--- a/src/OrchardCore/OrchardCore.DisplayManagement/ModelBinding/PageModelUpdater.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/ModelBinding/PageModelUpdater.cs
@@ -1,46 +1,21 @@
 using System;
 using System.Linq.Expressions;
 using System.Threading.Tasks;
-using Microsoft.AspNetCore.Mvc.ModelBinding;
 using Microsoft.AspNetCore.Mvc.RazorPages;
 
 namespace OrchardCore.DisplayManagement.ModelBinding
 {
     public class PageModelUpdater : PageModel, IUpdateModel
     {
-        private readonly PageModel _page;
-
-        public PageModelUpdater(PageModel page)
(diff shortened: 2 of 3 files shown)


src/OrchardCore/OrchardCore.DisplayManagement/RazorPages/Page.cs
```diff
@@ -226,6 +226,98 @@ public TagBuilder Tag(dynamic shape, string tag)
...
+
+            return RenderSection(name, required: true);
+        }
+
+        /// <summary>
+        /// Renders a zone from the layout.
+        /// </summary>
+        /// <param name="name">The name of the zone to render.</param>
+        /// <param name="required">Whether the zone is required or not.</param>
+        public IHtmlContent RenderSection(string name, bool required)
```
sebastienros, line 269: No. just don't create this method and let the users take the responsibility to call GetAwaiter(). We should not encourage that, even more at this level that can be highly concurrent.

ns8482e:
Sorry I don’t understand what you mean. I am just filling the gap between RazorPage.cs and Page.cs. 

This method allows user an ability to render zone in RazorPages similar to MVC views

src/OrchardCore/OrchardCore.DisplayManagement/RazorPages/Page.cs
ns8482e: I can remove this method from Page.cs but then the behavior will be inconstant between MVC views and RazorPages. 

As we can't remove this method from RazorPage.cs because it's required there to hide base method.

deanmarcussen:
So all the page model changes make sense.

In the linked issue https://github.com/OrchardCMS/OrchardCore/issues/5667 there is a razor pages layout at `Pages/Shared/_Layout` 

@jtkech will know, will a `Pages/Shared/_Layout` override the `/Views/Layout`, i.e. theme layout when specified in the @page 

I don't use razor pages enough to know for sure myself.

If that is the way to override the theme layout for a specific page, then the other changes in this pr make sense to me, just want to confirm that with @jtkech 

And of course we should make the same changes to `IsSectionDefined` and `Render` as we already did in your other pr @ns8482e

src/OrchardCore/OrchardCore.DisplayManagement/RazorPages/Page.cs
```diff
@@ -226,6 +226,61 @@ public TagBuilder Tag(dynamic shape, string tag)
...
+            // We can replace the base implementation as it can't be called on a view that is not an actual MVC Layout.
+
+            if (name == null)
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+
+            var zone = ThemeLayout[name];
+
+            return zone != null && zone.Items.Count > 0;
```
jtkech, line 245: So now we can just return `zone != null` ;)

src/OrchardCore/OrchardCore.DisplayManagement/ModelBinding/PageModelUpdater.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            return base.TryUpdateModelAsync<TModel>(model, prefix);
+        }
+
+        public new Task<bool> TryUpdateModelAsync<TModel>(TModel model, string prefix, params Expression<Func<TModel, object>>[] includeExpressions) where TModel : class
+        {
+            return base.TryUpdateModelAsync<TModel>(model, prefix, includeExpressions);
+        }
+
+        public new bool TryValidateModel(object model)
```
jtkech, line 36: We don't need this one as it is public in the base class

src/OrchardCore/OrchardCore.DisplayManagement/ModelBinding/PageModelUpdater.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            return base.TryValidateModel(model);
+        }
+
+        public new bool TryValidateModel(object model, string prefix)
```
jtkech, line 41: Idem

src/OrchardCore/OrchardCore.DisplayManagement/ModelBinding/PageModelUpdater.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        private readonly PageModel _page;
+
+        public PageModelUpdater(PageModel page)
+        {
+            _page = page;
+            PageContext = page.PageContext;
+        }
+
+        public new ModelStateDictionary ModelState => _page.ModelState;
```
jtkech, line 19: We don't need this line as the base class retrieve the same through the PageContext

src/OrchardCore/OrchardCore.DisplayManagement/ModelBinding/PageModelUpdater.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Linq.Expressions;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.AspNetCore.Mvc.RazorPages;
+
+namespace OrchardCore.DisplayManagement.ModelBinding
+{
+    public class PageModelUpdater : PageModel, IUpdateModel
+    {
+        private readonly PageModel _page;
```
jtkech, line 11: We don't need it, see below

src/OrchardCore/OrchardCore.DisplayManagement/ModelBinding/PageModelUpdater.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        public PageModelUpdater(PageModel page)
+        {
+            _page = page;
```
jtkech, line 15: We don't need it, see below

src/OrchardCore/OrchardCore.DisplayManagement/ModelBinding/PageModelUpdater.cs
```diff
@@ -0,0 +1,46 @@
...
+            PageContext = page.PageContext;
+        }
+
+        public new ModelStateDictionary ModelState => _page.ModelState;
+
+        public new Task<bool> TryUpdateModelAsync<TModel>(TModel model) where TModel : class
+        {
+            return base.TryUpdateModelAsync<TModel>(model);
```
jtkech, line 23: No luck that the base class one are protected ;)
Idem for the 2 followings

src/OrchardCore/OrchardCore.DisplayManagement/ModelBinding/ModelBinderAccessorFilter.cs
```diff
@@ -19,5 +20,23 @@ public void OnActionExecuting(ActionExecutingContext context)
...
             }
         }
+        public void OnPageHandlerSelected(PageHandlerSelectedContext context)
+        {
+            var page = context.HandlerInstance as PageModel;
+
+            if (page != null)
+            {
+                var modelBinderAccessor = context.HttpContext.RequestServices.GetRequiredService<IUpdateModelAccessor>();
+                modelBinderAccessor.ModelUpdater = new PageModelUpdater(page);
```
jtkech, line 30: Is it better to do it here? Just tried in `OnPageHandlerExecuting()`, it also works

I suggest it because the `ControllerModelUpdater` is set in `OnActionExecuting()`

src/OrchardCore/OrchardCore.DisplayManagement/RazorPages/Page.cs
```diff
@@ -226,6 +226,61 @@ public TagBuilder Tag(dynamic shape, string tag)
...
+        public Task<IHtmlContent> RenderSectionAsync(string name, bool required)
+        {            
+            if (name == null)
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+
+            var zone = ThemeLayout[name];
+
+            if (required && zone != null && zone is Shape && zone.Items.Count == 0)
```
jtkech, line 276: This one is okay because finally we didn't change it in `RazorPage` ;)

jtkech:
@ns8482e 

While doing a review i took more time to fix another issue, while trying the `PageModelUpdater` through the demo module starting with `/demo/foo/list`, some errors were logged. The `DisplayDriverBase.Copy()` doesn't work, seems to be only used in the demo module, it fails in runtime binding errors, i fixed it and will do another PR tomorrow

@deanmarcussen 

As i remember, so quicly done and needs to be confirmed

If it doesn't inherit from our custom `Page` (in the _ViewImports) you can define a layout in a regular way

    @{
        Layout = "_FooLayout"; // Notice without file extension
    }

`_FooLayout.cshtml` can be e.g. in `Views/Shared` or `Pages/Views/Shared`

If using the theming engine by inheriting from our custom `Page`, you can change the theme layout template, as we can with our `RazorPage`, by setting `ViewLayout` that add an alternate at the end


    @{
        ThemeLayout = "YourLayout";
    }

But here `YourLayout.cshtml` needs to be under the `Views` folder as a regular shape

ns8482e:
@deanmarcussen @jtkech It's been a while - I need to catch up on this. - Thanks

jtkech:
@ns8482e simpler to have committed my suggested changes ;) If you want to take a look

jtkech (approved):

ns8482e:
@jtkech Thank you.

PR #6144, opened 2020-05-09
Revisit Workflow Queries

jtkech:


src/OrchardCore.Modules/OrchardCore.Workflows/Http/Services/HttpRequestRouteActivator.cs
```diff
@@ -50,40 +49,37 @@ private async Task RegisterRoutesAsync()
...
-                    index.ActivityName == HttpRequestFilterEvent.EventName)
+                var pendingWorkflows = await _session
+                    .Query<Workflow, WorkflowBlockingActivitiesIndex>(index =>
+                        index.ActivityName == HttpRequestFilterEvent.EventName)
                     .Skip(skip)
                     .Take(pageSize)
                     .ListAsync();
 
-                if (!query.Any())
+                if (!pendingWorkflows.Any())
```
sfmskywalker, line 59: Just wondering whether `.ListAsync()` returns a deferred or evaluated enumerable? If deferred, then you might consider doing a `ToList()` to prevent another evaluation (when doing `.Count()` on line 71).

jtkech: I checked that `ListAsync()` may return `_query._session.Get<T>()` that create a [new List here](https://github.com/sebastienros/yessql/blob/e4dcb31da47a22a8554a8f7f84a5207c4c2b6abe/src/YesSql.Core/Session.cs#L436)

But maybe i checked it too quickly, just saw that it may return `localTransaction.Connection.QueryAsync<Document>()` so i have to check this one, thanks.

PR #6175, opened 2020-05-12
Highlight invalid fields

### LINKED ISSUE #6174: Invalid fields are not highlighted while editing content item.
When editing content item in admin - if fields/parts has validation error validation summary is displayed but invalid fields/parts are not highlighted.

### Fields

- [x] TextField
- [x] ContentPickerField
- [x] LocalizationSetContentPickerField
- [x] HtmlField
- [x] LinkField
- [x] NumericField
- [x] YoutubeField
- [x] MarkdownField
- [x] MediaField
- [x] DateField
- [x] TimeField
- [x] DateTimeField

### Parts

- [x] TitlePart
- [x] HtmlBodyPart  
- [x] LiquidPart
- [x] MarkdownPart

### LINKED ISSUE #6177: Saves Invalid Date fields.
Server side validation are not performed for required Date, Time and DateTimeFields. When disabled the browser default form validation via developer tools.

invalid-dates

ns8482e:
Fixes #6174 
Fixes #6177 

Following Fields are fixed

TextField
 ContentPickerField
 LocalizationSetContentPickerField
 HtmlField
 LinkField
 NumericField
 YoutubeField
 DateField
 DateTimeField
 TimeField

Following Parts are fixed
 
TitlePart
HtmlBodyPart
MarkdownBodyPart


jptissot:
I will try this when I get a chance

sebastienros:
Am I correct to say we were just missing the `Prefix` value in the error key?
You added a expression evaluator where we could just keep `nameof(property)`. I am afraid that it will just add a risk for exception and perf issue (maybe useless allocations). And I couldn't find a single example of `vm.A.B` (multiple member accesses).

ns8482e:
@sebastienros yes you are right - It's missing Prefix values for all these fields/parts. I repeated same mistake when I copied paste code from one of such field.

While fixing my code - I realized the need for expression traverse  where I use `vm.A.B` to avoid such mistake as `nameof(vm.A.B)` only returns `B`

So, I thought including an extension method would help others avoid such mistakes.

agriffard:
Please fix the conflicts @ns8482e

ns8482e:
@agriffard fixed conflicts.

hishamco:
@agriffard is this ready to go?

deanmarcussen:
I think we should do it without the extension method.

Yes, it opens up more room for error, but the initial error was not applying any extra field values to the prefix, so to fix that would remove the need for the expression tree.

ns8482e:
@davidhayden @agriffard  I'll update as requested - Thanks
Btw - here, expression tree is not compiled or executed - just converted to string

ns8482e:
@deanmarcussen @agriffard Updated to use nameof

PR #6182, opened 2020-05-13
Solution filter

agriffard:
Add a OrchardCore.Filter.slnf file to load only 140 projects (Unload docs, templates and tests projects).

https://docs.microsoft.com/en-us/visualstudio/ide/filtered-solutions?view=vs-2019

Not sure it is really faster to load.
Please try and tell me.


hishamco:
I have a better idea for this, coz I'm suffering from such large solution ;) we could create multiple solutions for different purpose

1. `OrchardCore.sln`

2. `OrchardCore.Tests.sln`

3. `OrchardCore.docs.sln`

.. etc

For instance anyone interested in documentation will choose `OrchardCore.docs.sln` which load the minimal and required projects

4.

agriffard:
Working with large code bases in Visual Studio:
https://www.youtube.com/watch?v=hWk9TT5osL0&list=PLReL099Y5nRem6UA4w849hTfd0LzBIsAk&index=3 

Using solution filters.

hishamco:
I think there's a value to create some filters or document this at least. This is very useful for laptops that don't have very enough RAM like mine ;)

PR #6186, opened 2020-05-14
Update PublishLaterPart.Edit.cshtml

jtkech:
After watching the last meeting

As used in other places, the tilde is used to be tenant aware and also virtual folder aware


sebastienros:
I am not sure about that. Static files should be tenant independent, or at least that's what I thought.

jtkech:
@sebastienros yes i thought about that as static files of a given module are shared accross tenants

So using the `~` (as we use everywhere) generates tenant prefixed url

And this is what we need when using e.g. the `ITenantFileProvider` that is specific per tenant.

And anyway, without the `~` it doesn't work under a virtual folder

PublishLater

PR #6187, opened 2020-05-14
YesSql Session Wrapper

jtkech:
Just to describe 2 "issues" related to the session that i saw, if there are real issues maybe better to do it on YesSql side, at least this wrapper, quickly done and registered in place of the `ISession`, shows what i mean. I will need to focus on it again but as i remember the 2 concerns were

- A save followed by a query, and then a mutation, needs another save to persist the last mutation.

- A cancel followed by a query, a mutation and then a save, needs another cancel to not persist the last mutation.

- TODO: See what to do when `CommitAsync()` is called explicitly as in `UserStore`.

### DIFF
--- a/src/OrchardCore/OrchardCore.Data/SessionWrapper.cs
+++ b/src/OrchardCore/OrchardCore.Data/SessionWrapper.cs
@@ -32,7 +32,7 @@ namespace OrchardCore.Data
 
             if (!_canceled)
             {
-                // A save followed by a query and then a mutation, needs another save to persist the last mutation.
+                // A save followed by a query, and then a mutation, needs another save to persist the last mutation.
                 foreach (var item in _saved)
                 {
                     _session.Save(item);
--- a/src/OrchardCore/OrchardCore.Data/SessionWrapper.cs
+++ b/src/OrchardCore/OrchardCore.Data/SessionWrapper.cs
@@ -1,0 +1,88 @@
+using System;
+using System.Collections.Generic;
+using System.Data.Common;
+using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


src/OrchardCore/OrchardCore.Data/SessionWrapper.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            if (_session == null)
+            {
+                return Task.CompletedTask;
+            }
+
+            if (!_canceled)
+            {
+                // A save followed by a query, and then a mutation, needs another save to persist the last mutation.
+                foreach (var item in _saved)
```
sebastienros, line 36: This is doing an extra save no matter what. Does it actually issues an extra Update in the db, or simply a check that the object hasn't changed (yessql does that I think).

jtkech: As i remember, if no query was done it is a no-op, otherwise i think it does an extra update

> or simply a check that the object hasn't changed (yessql does that I think).

I didn't see that.

Then i think it is the same if we save an item1, then save an item2 and do an explicit `commtAsync()` (as in UserStore), here there will be extra updates of item1 and item2, only updates not creations as `_identityMap` is only cleared on `ReleaseSession()` when disposing the session.

jtkech:
@sebastienros 

As i have time i will re-do some tests with or without the wrapper. And / or maybe i will fork yesSql and try to update the code on yesSql side and do some unit tests

jtkech:
I did some tests and opened a new issue #6223, maybe followed by a new PR

PR #6195, opened 2020-05-15
use validating handlers from Content Display Manager

### LINKED ISSUE #6194: Validation logic is duplicated for AutoroutePart and AliasPart
We are duplicating the validating login in the display driver and in the handler. For example AutoroutePart and AliasPart have the same set of validation rules, in two different places.

Ideally, we should define the logic in a handler and use this when importing, saving, cloning etc.

MichaelPetrinolis:
remove validation logic from autoroute and alias display driver

closes #6194


ns8482e:
@MichaelPetrinolis IMO you still need to keep `ModelState.AddModelError` in DisplayDriver As field names are not always match the ModelStateKey. Also some field/part use EditViewModel that does not update underlying Field/Part in case of error.

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Handlers/ValidateContentContext.cs
```diff
@@ -7,13 +10,31 @@ public ValidateContentContext(ContentItem contentItem) : base(contentItem)
         }
 
         public ContentValidateResult ContentValidateResult { get; } = new ContentValidateResult();
+
+        public string BuildPrefix<TPart>(string memberName)
```
ns8482e, line 14: See PR #6175 nameof doesn't work always - I have created ModelState extension method there, Probably we could use that logic here

MichaelPetrinolis: In the context of ContentParts I think that using nameof is enough. The only issue  would be with named parts, which is solved by setting the Prefix in validate context when running for each part.

src/OrchardCore/OrchardCore.ContentManagement.Display/ContentDisplayManager.cs
```diff
@@ -170,6 +171,26 @@ public async Task<IShape> UpdateEditorAsync(ContentItem contentItem, IUpdateMode
...
+            await _contentHandlers.InvokeAsync((handler, validateContentContext) => handler.ValidatedAsync(validateContentContext), validateContentContext, _logger);
+            if (!validateContentContext.ContentValidateResult.Succeeded)
+            {
+                foreach (var error in validateContentContext.ContentValidateResult.Errors)
+                {
+                    if (error.MemberNames.Any())
+                    {
+                        foreach (var member in error.MemberNames)
+                        {
+                            context.Updater.ModelState.AddModelError(context.HtmlFieldPrefix, member, error.ErrorMessage);
```
ns8482e, line 185: However this line is not just for `AliasPart `or `AutoroutePart` this could be for every Parts and Fields - Orchard provided or Custom module provided and having just `nameof()` may show errors in ValidationSummary but would enough to represent accurate field in UI editor.

For example if the part is consist of complex properties and not just primitive types, e.g `AddressPart` as following

   ```
 public class AddressPart : ContentPart
    {
        public string Street {get;set;}
        public State State {get;set;}
        public Country Country {get;set;}
    }
    public class State 
    {
        public string Abbr {get;set;}
        public string Name {get;set;}
    }
    public class Country 
    {
        public string Abbr {get;set;}
        public string Name {get;set;}
    }
```

using member name with `nameof(State.Name)`  and `nameof(Country.Name)`would generate same key `Name` the correct name should be `State.Name` and `Country.Name`

Comment (MichaelPetrinolis) [reply src/OrchardCore/OrchardCore.ContentManagement.Display/ContentDisplayManager.cs:185]: @ns8482e thanx for the clarification, I see the point. But IMO we should handle how to add the name of the part as a prefix among with the html Prefix used in the driver in the validation context. In the example, State and Country should be able to validate themselves and in the validating handler the developer should add State.ValidationErrors and Country.ValidationErrors prefixed by partName. If we prefer the penalty of creating an expression and traversing the tree in order to find the name for each validation for giving an API to resolve nested object names, we could change it.

**Comment by deanmarcussen** [comment]
Yes and No? 

So for some info.

The original intention behind the `ValidateAsync` method on the content manager was to provide an alternative validation point for `ContentApi` `Workflow` and the `Import` methods, and initially to only validate certain important features like Autoroute and Alias, when unicity was crucial to the system being able to operate correctly.

The design request (from @sebastienros) was to not tie it to UI related validation, so that API related concerns could remain independent of UI related concerns.

I imagine it's for this reason we don't call the UI Drivers from an API call (which would mean this code duplication was not necessary)

Definitely something to think about if it got brought together like this.

**Comment by ns8482e** [comment]
> The original intention behind the `ValidateAsync` method on the content manager was to provide an alternative validation point for `ContentApi` `Workflow` and the `Import` methods,

IMO  `ValidateAsync` should take care of data integrity for Content Item as per content type definition regardless of mode of access. However it should still relate the validation error to field/part accurately as per content type definition ( As in my example `State.Name` not just `Name`)

DisplayDriver should also have some it's own UI validation. For example, `DateTimePart` - where property validated in `ValidateAsync` is `Value` where as for UI the property validated is `LocalDateTime`  which updates underlying `Value`. Here duplication is required.

**Comment by sebastienros** [comment]
I have concerns with moving some of the UI validation to these validate handlers. We use these for APIs and recipes, to ensure that the model is valid, not that the form used to edit a part is valid.
And we would need to update the data model before validating. And the UI might split one datum into multiple field, so we wouldn't be able to define where the issue is coming from (validation messages). For instance some Update methods use the view model which might contain more data than the actual part/field.

Or maybe we should do both, but there could be duplication of messages, or checks.
What about explicitly calling the Validate method from the driver Update if that makes sense.

**Comment by MichaelPetrinolis** [comment]
Exposed a ValidateAsync method through UpdatePartEditorContext to the UpdateAsync method of ContentPartDisplayDriver, and an extension method to ModelState to add validationresults

**Comment by MichaelPetrinolis** [comment]
although travis ci seems passed, it is shown as in progress

**Comment by deanmarcussen** [comment]
@MichaelPetrinolis Were you able to resolve the prefixing issue?

i.e. making sure the correct prefix was applied to the `ModelState` so that fields remain highlighted when in an invalid state?

**Comment by MichaelPetrinolis** [comment]
@deanmarcussen just got my hands on this PR after some time :) 
Checking to see if it works as expected and try to remember what I have done.

**Comment by MichaelPetrinolis** [comment]
Only AliasPartHandler and AutoroutePartHandler use the Validating callbacks and they work as expected.
I don't know if there is a better way to provide the validation context to display drivers, I changed the ContentDisplayManager to have the minimum impact on code changes so that the display driver can call the validation logic and set the model errors.

We could/should add a validating logic on PartSettings as well. For example, we perform the following check
```
if (!string.IsNullOrEmpty(model.Pattern) && !_templateManager.Validate(model.Pattern, out var errors))
{
       context.Updater.ModelState.AddModelError(nameof(model.Pattern), S["Pattern doesn't contain a valid Liquid expression. Details: {0}", string.Join(" ", errors)]);
}
``` 
on every Settings Display Driver that has a Liquid pattern. Not on this PR but we can check if it is doable.

**Comment by deanmarcussen** [comment]
> Only AliasPartHandler and AutoroutePartHandler use the Validating callbacks and they work as expected.

Great. Will have a proper look this week.

**Comment by agriffard** [comment]
@deanmarcussen Can you please review this PR?

**Review by deanmarcussen — APPROVED** [review]
> What about explicitly calling the Validate method from the driver Update if that makes sense.

I like the approach you have here of calling `Validate` from the driver `Update`.
In terms of the other raised issues I think you've resolved them.

Fields are not relevant here, as they do not have handlers, so any `Handler` related to them would be custom by the user, and specific to the `Part` that was holding them, so not easily used between them.

Named Parts also would be handled by the `Prefix`, but still be custom to the user I suspect

Impact / changes is minimal.

LGTM

@agriffard this removes some translations which are now redundant, but I hope the new(ish) translations are there (the value was kept the same between them)

File: src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Handlers/ValidateContentContext.cs [hunk 3]
```diff
@@ -7,13 +13,27 @@ public ValidateContentContext(ContentItem contentItem) : base(contentItem)
...
-        public static void Fail(this ValidateContentContext context, params string[] errors)
+
+        public static void Fail(this ValidateContentContext context, params ValidationResult[] errors)
         {
             context.ContentValidateResult.Fail(errors);
         }
+
+        public static void Fail(this ValidateContentContext context, string errorMessage, params string[] memberNames)
+        {
+            if (memberNames.Any())
```
sebastienros, line 29: NullReferenceException if the params is not set

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Handlers/ValidateContentContext.cs
```diff
@@ -7,13 +13,27 @@ public ValidateContentContext(ContentItem contentItem) : base(contentItem)
...
 
         public ContentValidateResult ContentValidateResult { get; } = new ContentValidateResult();
+
     }
 
     public static class ValidateContentContextExtensions
     {
-        public static void Fail(this ValidateContentContext context, params string[] errors)
+
+        public static void Fail(this ValidateContentContext context, params ValidationResult[] errors)
```
sebastienros, line 22: NullReferenceException if the params is not set

sebastienros:
Can it be an issue that the Display Driver now depends on the handler?

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Handlers/ValidateContentContext.cs
jtkech: According to the doc

> If you send no arguments, the length of the params list is zero.

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Handlers/ValidateContentContext.cs
jtkech: Idem i think

PR #6200, opened 2020-05-16
Add constructor overload to CultureDictionaryRecord

hishamco:
Add additional overload to `CultureDictionaryRecord` that accept `messageId` and `translations` parameter could be useful for some cases:

- When there's no context is required, like in my case in `JsonStringLocalizer` [here](https://github.com/OrchardCoreContrib/OrchardCoreContrib/blob/dev/src/OrchardCoreContrib.Localization.Abstractions/CultureDictionaryRecordWrapper.cs), so the workaround that I did is to subclass the `CultureDictionaryRecord`

- Adding `params` also useful to
    - Avoid creating new array in the caller
    - Make `CultureDictionaryRecord` works fine for localizers that doesn't support pluralization


hishamco:
I need to remove the extra `params` in the second constructor to remove the ambiguous, because the the second constructor will be called all the time

PR #6202, opened 2020-05-16
Make CultureDictionary IEnumerable

hishamco:


test/OrchardCore.Tests/Localization/CultureDictionaryTests.cs
```diff
@@ -50,5 +53,28 @@ public void IntexerThrowsPluralFormNotFoundExceptionIfSpecifiedPluralFormDoesntE
...
+
+        [Fact]
+        public void EnumerateCultureDictionary()
+        {
+            // Arrange
+            var dictionary = new CultureDictionary("ar", _arPluralRule);
+            dictionary.MergeTranslations(new List<CultureDictionaryRecord>
+            {
+                new CultureDictionaryRecord("Hello", null, new[] { "مرحبا" }),
+                new CultureDictionaryRecord("Bye", null, new[] { "مع السلامة" })
```
sebastienros, line 65:
```suggestion
                new CultureDictionaryRecord("Hello", "مرحبا"),
                new CultureDictionaryRecord("Bye", "مع السلامة")
```

hishamco:
@sebastienros is it fine for you to merge this?

agriffard:
@jtkech Can you please review?

jtkech:
Okay will do soon ;)

sebastienros:
Not sure why you need it but it doesn't hurt

sebastienros (approved):

jtkech:
As i remember @hishamco was talking about perf but not sure ;)

hishamco:
Not about Perf @jtkech ;) that's another issue, the idea to make `CultureDictionary` foreachable which is useful in some cases

jtkech:
Okay cool then

Yes because i didn't see a current usage in the code

hishamco:
Perhaps there's no use case I'm the source at that time but I need to check now. If there's not at least you can iterate over this collection like the other once

sebastienros:
> Perhaps there's no use case

I asked because this library can be used outside of OC, so I assumed there was some use case. But if you don't even have one then don't do anything. The only thing you can do is introduce bugs, or create more public surface to maintain.

hishamco:
Seem I made a typo in my last reply, what I mean is perhaps there's no use case in the source code at that time but there are some use cases like:

- Po Parser could return a `CultureDictionary`

- Filter the translations in `CultureDictionary` , this could be useful if I want to provide a UI for translation per component or view

there could be more use cases outside OC

sebastienros:
Ok, and this change doesn't seem risky. But it was good to talk about when to not do anything.

hishamco:
You are right .. at the time of writing the PR (7 months ago) there's a case where I need to traverse the `CultureDictionary` records I find there's no proper way to do that ;)

I think making `PoParser` returns `CultureDictionary` will be a proof / use case for this PR if you agreed with

Thanks

PR #6226, opened 2020-05-20
Sanitize link field

deanmarcussen:
Last todo from the sanitization pr.

This returns an error on malformed input on the link field.

I was going to try and parse the Url and or Text back out and save it without an error, but the way the sanitizer works with anglesharp there is not a guarantee of indexes to find the url and text values afterwards.

So easier to return an error on an attempted click jack / malformed input

example (would probably not work as a click jack, but you get the picture)

error on clickjack

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Drivers/LinkFieldDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Drivers/LinkFieldDisplayDriver.cs
@@ -9,15 +9,18 @@ using OrchardCore.ContentManagement.Display.Models;
 using OrchardCore.ContentManagement.Metadata.Models;
 using OrchardCore.DisplayManagement.ModelBinding;
 using OrchardCore.DisplayManagement.Views;
+using OrchardCore.Infrastructure.Html;
 
 namespace OrchardCore.ContentFields.Drivers
 {
     public class LinkFieldDisplayDriver : ContentFieldDisplayDriver<LinkField>
     {
+        private readonly IHtmlSanitizerService _htmlSanitizerService;
         private readonly IStringLocalizer S;
 
-        public LinkFieldDisplayDriver(IStringLocalizer<LinkFieldDisplayDriver> localizer)
+        public LinkFieldDisplayDriver(IHtmlSanitizerService htmlSanitizerService, IStringLocalizer<LinkFieldDisplayDriver> localizer)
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.ContentFields/Drivers/LinkFieldDisplayDriver.cs
```diff
@@ -81,8 +84,16 @@ public override async Task<IDisplayResult> UpdateAsync(LinkField field, IUpdateM
                     updater.ModelState.AddModelError(Prefix, S["The text default value is required for {0}.", context.PartFieldDefinition.DisplayName()]);
                 }
 
-                // Run this through a sanitizer in case someone puts html in it.
-                // No settings.
+                if (!string.IsNullOrEmpty(field.Url) || !string.IsNullOrEmpty(field.Text))
+                {
+                    // Test the input against the sanitizer.
+                    var aTag = @"<a href=""" + (field.Url ?? "") + @""">" + (field.Text ?? "") + "</a>";
```
hishamco, line 90: Use `string.Empty` if it's possible

sebastienros, line 90: I am afraid that any special will break this.

For instance Url and Text are not html encoded here, while the view I think they will, and a sanitizer would happen after html encoding. Like have a text with `<script>` would be fine in theory, because it's a text that will be encoded, but I think that this code right now will prevent it.

sebastienros: In this case it would make the code less readable. And the compiler handles that for you.

deanmarcussen: I suspect you are correct. 
Quickly done, so I need to go back and look for a better example of an actual attack point here (if there is one, I'm not totally sure)

I'll mark this not ready and finish up some research first

sebastienros: Maybe this field can't be protected. Or the validation should only be based on white list. You can also check what WP does.

agriffard:
What is the status of this PR @deanmarcussen?

PR #6239, opened 2020-05-22
NodaTime 3

agriffard:
https://github.com/nodatime/nodatime/releases/tag/3.0.0
https://nodatime.org/versions


Skrypt:
Should be fine to merge

sebastienros:
Risky ... no? Major version, just released, it doesn't fix any known issue for us, or is there?
Maybe wait after rc2?

agriffard:
Ok, it can wait.

Skrypt:
Doesn't fix anything. Just adds new functions. One is a function to match Windows time id's to IANA timezone id's ... could be usefull. Also, if I understand after reading the release notes it makes the Nodatime 3.0 .net standard 2.0 compliant. It also updates the IANA Timezone database to most recent ones which is really handfull for those needing these timezone updates.

Risky still if we don't have any unit tests to make sure timezones will still work but I'd be surprised of that.

Issue #6234, opened 2020-05-21
Correct way to generate url for a bagged item


jwu-au:
Hi Orchard Team,

Suppose I have below code for generating a content item link for a bagged item `Service`
```liquid
  {% for service in services.Content.BagPart.ContentItems %}
        {% assign serviceShape = service | shape_build_display: "Summary" %}
        {% a display_for: serviceShape.ContentItem, class: "btn btn-primary" %}
              {{ service.TitlePart.Title }}
        {% enda %}
  {% endfor %}
```

So I have to generate a shape from the service `JObject` , and then pass its `ContentItem` to the `a tag` helper. Is there a directly/another way to generate a link for a contained item?

ns8482e:
> ```
> {% assign serviceShape = service | shape_build_display: "Summary" %}
> ```

May be you don't need to build display.  you can directly use `service` for `display_for`.

jwu-au:
Hi @ns8482e ,

Because the `service` inside a bag is a `JObject`, by passing service directly into `display_for`, it throws below:
```
An unhandled exception occurred while processing the request.
InvalidCastException: Unable to cast object of type 'Newtonsoft.Json.Linq.JObject' to type 'OrchardCore.ContentManagement.ContentItem'.
OrchardCore.DisplayManagement.Liquid.TagHelpers.LiquidTagHelperActivator.CallPropertySetter<TDeclaringType, TValue>(Action<TDeclaringType, TValue> setter, object target, object value) in LiquidTagHelperActivator.cs, line 242
```

jtkech:
In the `BuildDisplayFilter` if the input value is a `JObject` we do `jObject.ToObject<ContentItem>()`. We would need to do the same in the `ContentLinkTagHelper`

jwu-au:
Hi @jtkech ,

I created a PR https://github.com/OrchardCMS/OrchardCore/pull/6242 for fixing the issue.
I'm not sure it is the best way to do it.
From my testing, it should fix all properties assignments from jObject in the tag helper.
Have a review.

ns8482e:
> InvalidCastException: Unable to cast object of type 'Newtonsoft.Json.Linq.JObject' to type 'OrchardCore.ContentManagement.ContentItem'.

@jtkech May be ContentItem should implement implicit operator for `JObject`?

jtkech:
@jwu-au good idea

@ns8482e good idea also

Need to think about it

jtkech:
[...]

PR #6242, opened 2020-05-22
fixed jObject bind for properties

jwu-au:


jwu-au:
@ns8482e You're right that we cannot tell what user passed in. From my understanding that the responsibility in LiquidTagHelperActivator is to convert what user passed in to target type (PropertyType), if any error happens in this point, it is good time for user to get the error, and correct input object.

And I firstly looked at if I can fix this in the ContentLinkTagHelper, I found that I need to actually change the PropertyType for `DisplayFor` to (maybe) `object` type, as you said, do the checking there, convert it if it is a JObject. And also need to do it the same for other properties in the same tag helper such as "EditFor" etc. But eventually may ended up with the same line `jObject.ToObject(property.PropertyType)` . I'm happy to see an example to do the conversion in a tag helper.

Thank you.

ns8482e:
> From my understanding that the responsibility in LiquidTagHelperActivator is to convert what user passed in to target type (PropertyType),

IMO  `LiquidTagHelperActivator` should not do any conversion as it just a helper between View and `TagHelper` has no knowledge of underlying functionality of requested `TagHelper`

Unless you find a way to be `100%` sure that input JObject is convertible to `PropertyType` without any dataloss.

jtkech:
As commented in #6234, here a copy

Hmm, when working from `ContentItem.Content` we get a `JObject` from which we can get parts that are also `JObject` and so on. So the problem is that we can't get e.g. a real `BagPart` whose `ContentItems` property is really a `ContentItem` collection.

i think the problem is only for liquid because in razor we can use `ContentItemExtensions` as this one `As<TPart>(this ContentItem contentItem)`

So maybe one solution would be to create a liquid filter that allows to retrieve real parts / fields from a `ContentElement` (so a `ContentItem`, or a `ContentPart`, or a `ContentField`), not its json representation, e.g `contentItem | get_element "Services"`.

agriffard:
Can you provide a PR with the changes you suggested @jtkech?

jtkech:
@agriffard need to focus on it again when i will have time, but added on my list ;)

jtkech:
So, as said by @sebastienros in the related issue #6234, the right fix would be

> I would either create filter that convert a JObject to a ContentItem, or have an implicit conversion from JObject to ContentItem.

But, in place of using `Model.ContentItem.Content` that is a `JObject`, just tried the following by using `Model.Content` that is a `Shape` and then `Model.Content.Services` that is a `BagPart` holding a real `List<ContentItem>`.

    {% for service in Model.Content.Services.ContentItems %}
          {% a display_for: service, class: "btn btn-primary" %}
                {{ service.DisplayText }}
          {% enda %}
    {% endfor %}

So i close this PR, feel free to re-open it if it doesn't work for you

PR #6243, opened 2020-05-22
Add a note to localization module docs

### LINKED ISSUE #6240: Incomplete information in Localization documentation?
Here https://docs.orchardcore.net/en/dev/docs/reference/modules/Localize/#publishing-localization-files docs say 

> Add the following configurations to your [Web Project].csproj file to include them as Content

However, when I'm creating a new module that I want to provide localizations within the module, that doesn't work like that. The .po file has to be included in the project as an embbeded resource if not it is not taken into account as localization file

hishamco:
Fixes #6240

### DIFF
--- a/src/docs/reference/modules/Localize/README.md
+++ b/src/docs/reference/modules/Localize/README.md
@@ -42,6 +42,9 @@ Add the following configurations to your `[Web Project].csproj` file to include 
   </ItemGroup>
 ```
 
+!!! note
+    Translation files can be provided by a module, in that case they are embedded automatically in the module assembly unless Visual Studio added some bad item groups in the project file.
+
 ## Recipe Step
 
 Cultures can be added during recipes using the settings step. Here is a sample step:
--- a/src/docs/reference/modules/Localize/README.md
+++ b/src/docs/reference/modules/Localize/README.md
@@ -32,11 +32,13 @@ Especially if mounting a volume at `/App_Data/` as mounting hides pre-existing f
 ### Publishing Localization files
 
 The PO files need to be included in the publish output directory. 
-Add the following configurations to your `[Web Project].csproj` file to include them as EmbeddedResource.
+Add the following configurations to your `[Web Project].csproj` file to include them as Content.
 
 ```xml
   <ItemGroup>
-    <EmbeddedResource Include="Localization\**" />
+    <Content Include="Localization\**" >
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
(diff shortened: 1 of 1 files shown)


sebastienros:
I don't think so. We even have a step to copy the files. Needs more thoughts.

hishamco:
If I'm not wrong the MSBuild step is for recursively copy the localization folder for the `OrchardCore.Translations` NuGet packages @jtkech correct me if I'm wrong

sebastienros:
@jtkech is that the solution to the issue?

jtkech:
No, i don't think so

- Translation files can be provided by a module, in that case there are embedded automatically in the module assembly unless VStudio added some bad item groups in the project file.

- Some can be tenant specific if under the tenant folder where there are served as physical files.

- Finally they can be under the app root Localization folder where there are served through the `ContentRootFileProvider`, so normally as physical files.

    We have a custom targets file to copy them from the packages locations, and to mark them to be published, this automatically on building the app.

    So no need to mark them as `EmbeddedResource`, and no need to mark them either as `Content` unless you want to see them under the VStudio solution explorer.

hishamco:
> So no need to mark them as EmbeddedResource, and no need to mark them either as Content unless you want to see them under the VStudio solution explorer

Normally if a module developer started to localize some string he will add a `.po` files from within VS, @jersiovic could please let us know more details about your localization strings location or a minimal repo to reproduce the issue

jersiovic:
The comment in docs is correct but incomplete, I don't think you need to change it, so it was ok as it was, so I would undo the change in this PR.
I think you should add an extra explanation for this case "Translation files can be provided by a module, in that case there are embedded automatically in the module assembly unless VStudio added some bad item groups in the project file."

Why? Because when you start to have problems with localization, you take a look to this page in the docs. And there you read .po files has to be included as Content. Then you look your .po file in your module included as Embeded Resouce and you change it to be included as Content. At this point your problem is more dificult to solve, because no where no one says in that scenario it has to be included as Embeded Resource.
An example of scenario when you start to have problems with localization is the one that happened me. I was trying to localize default messages provided by validation annotations like for example Required Annotation (I mean when you don't provide a custom ErrorMessage to [Required] annotation). This is sth that simply doesn't work. So, in that case trying to solve sth that is not posible to solve, to get wrong how has to be included a .po file in your module is sth that doesn't help. Because the docs don't say it explicitely that it has to be Embeded and in fact seems to say the opposite, cause they don't talk explicitly of the scenario for .po files in modules.

Issue #6218, opened 2020-05-19
ShapeCreatedContext does not have service provider


deanmarcussen:
The `ShapeCreatedContext` does not have a serviceProvider like the `ShapeDisplayContext` do.

This means services resolved through an `IShapeTableProvider` with `OnCreated` either need to be resolved through the constructor, which causes disposed exceptions, and scope capturing, or by directly accessing the `ShellScope.Services`.

Is there any reason not to have a `IServiceProvider` available on the events?

The `DefaultShapeFactory` is scoped, so should not cause any capturing issues I think

PR #6250, opened 2020-05-22
Add IServiceProvider to shape Creating Created contexts

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6218

/cc @jtkech Good to do this?


jtkech:
@deanmarcussen looks good as the `DefaultShapeFactory` is a scoped service

PR #6255, opened 2020-05-23
Import Recipe from Json

agriffard:
Allow to import a json recipe from a text area in admin:

image


src/OrchardCore.Modules/OrchardCore.Deployment/Controllers/ImportController.cs
```diff
@@ -102,7 +102,54 @@ public async Task<IActionResult> Import(IFormFile importedPackage)
...
+
+        [HttpPost]
+        public async Task<IActionResult> Json(string json)
+        {
+            if (!await _authorizationService.AuthorizeAsync(User, Permissions.Import))
+            {
+                return Forbid();
+            }
+
+            if (!string.IsNullOrWhiteSpace(json))
```
deanmarcussen, line 126: you could use that `IsJson()` extension I added here, and rather than a notifier add an error to the model state.

Then check if the model state is valid, and `View(json)` to return the bad value to the form.
It should probably be a class
`ImportJsonViewModel`

Then your binding in the view can use `asp-for` instead of the manual id and name you have.

Should end up looking something like this
```
            if (ModelState.IsValid)
            {
                ...run the recipe...

                return RedirectToAction(nameof(Json));
            }

            // If we got this far, something failed, redisplay form
            return View(model);
```

src/OrchardCore.Modules/OrchardCore.Deployment/Views/Import/Json.cshtml
```diff
@@ -0,0 +1,27 @@
+<h1>@RenderTitleSegments(T["Import JSON"])</h1>
+
+<style asp-name="codemirror"></style>
+<script asp-name="codemirror" depends-on="admin" at="Foot"></script>
+<script asp-name="codemirror-addon-display-autorefresh" at="Foot"></script>
+<script asp-name="codemirror-mode-javascript" at="Foot"></script>
+
+<form asp-action="Json">
+    <div class="form-group">
+        <textarea class="form-control mb-2" rows="20" id="json" name="json"></textarea>
```
deanmarcussen, line 10: needs to be `asp-for` not the manual id and name

src/OrchardCore.Modules/OrchardCore.Deployment/Views/Import/Json.cshtml
```diff
@@ -0,0 +1,27 @@
+<h1>@RenderTitleSegments(T["Import JSON"])</h1>
```
deanmarcussen, line 1: Suggest JSON Import to keep the translations the same as the menu item?

src/OrchardCore.Modules/OrchardCore.Deployment/Controllers/ImportController.cs
agriffard: ok

src/OrchardCore.Modules/OrchardCore.Deployment/Views/Import/Json.cshtml
agriffard: ok

src/OrchardCore.Modules/OrchardCore.Deployment/Views/Import/Json.cshtml
agriffard: ok

src/OrchardCore.Modules/OrchardCore.Deployment/Controllers/ImportController.cs
```diff
@@ -102,7 +109,55 @@ public async Task<IActionResult> Import(IFormFile importedPackage)
...
+                var tempArchiveFolder = PathExtensions.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+
+                try
+                {
+                    Directory.CreateDirectory(tempArchiveFolder);
+                    System.IO.File.WriteAllText(Path.Combine(tempArchiveFolder, "Recipe.json"), model.Json);
+
+                    await _deploymentManager.ImportDeploymentPackageAsync(new PhysicalFileProvider(tempArchiveFolder));
+
+                    _notifier.Success(H["Recipe imported."]);
```
deanmarcussen, line 149: return RedirectToAction(nameof(Json));

deanmarcussen: Just to do a return to here on success.
We could have a catch here, to add a model state error, but equally we don't on the import package controller.
One day we will want better error messages from the recipe executor, but that's another pr.
(When you run this on a prod server, where the environment is not development with errors, you just get a 500 response back because development exception handling is turned of)

agriffard: I didn't use `return RedirectToAction(nameof(Json));` because I want the json to be kept in the textarea.

deanmarcussen: So stays even after success? cool, sounds fine to me if you prefer that

Issue #6260, opened 2020-05-24
Sanitize Html removes css classes


agriffard:
1) When you use an HtmlField in a content type with a Widget Stereotype, the css classes are removed.

2) With ou without 'Sanitize Html' checked, the line breaks are also removed.

deanmarcussen:
1. Yes, classes are considered susceptible to class jacking (https://html5sec.org/#123) so removed from the sanitizer by default

There's a unit test showing how to configure the sanitizer to accept classes, and other attributes.
https://github.com/OrchardCMS/OrchardCore/blob/736cdd5977196baf7ca2d5f9a77ec28ad3af4e22/test/OrchardCore.Tests/Html/HtmlSanitizerTests.cs#L27

Maybe we could accept classes by default, I admin I'm not sure how class jacking could be used in our scenario?

2. I couldn't repro this?

jptissot:
Do we have some docs related to the sanitizer? I was searching how to configure it recently,thanks for the tip Dean.

agriffard:
Yes, please accept classes by default.

deanmarcussen:
Haven't got docs yet :(

Can you do a bit of research @agriffard and see if classes are going to be safe for us?

I think they will because there's nowhere for the user to enter a malicious script that could be used to classjack.

It is easy to configure this by default, so pending some research I'll put a pr in :)

agriffard:
It might be safe, you can not inject javascript with the class attribute, only with scripts, `on` events, img src or a href.

PR #6262, opened 2020-05-24
Replace input by textarea in HtmlField.Edit

agriffard:
Related to #6260 to fix the line breaks removed after saving.


agriffard:
The problem comes when you use an input instead of a textarea.

Code mirror is useful because of the aut-expand of the lines and the syntax colorization.

PR #6268, opened 2020-05-25
Allow classes by default with the html sanitizer

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6260

We should just give this a couple of days to make sure it's a good idea

Anyone can still alter the options, by using `Configure` to register a different `Action` to be invoked against the sanitizer.

A new registration would cancel the old `Action`. We could have a list of `Actions` and add to them, so it is progressive, rather than replacing.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Markdown/Services/MarkdownPipelineOptions.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Markdown/Services/MarkdownPipelineOptions.cs
@@ -6,7 +6,6 @@ namespace OrchardCore.Markdown.Services
 {
     public class MarkdownPipelineOptions
     {
-        public Action<MarkdownPipelineBuilder> Configure { get; set; }
-        public IList<Action<MarkdownPipelineBuilder>> Configure2 { get; set; } = new List<Action<MarkdownPipelineBuilder>>();
+        public IList<Action<MarkdownPipelineBuilder>> Configure { get; } = new List<Action<MarkdownPipelineBuilder>>();
     }
 }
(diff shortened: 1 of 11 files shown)


deanmarcussen:
I will make some changes to this, to make the configuration seamless and more additive

deanmarcussen:
Changes made, extension method to configure a list of actions, so it does not override the defaults of adding classes, if you add some more allow attributes etc

src/OrchardCore/OrchardCore.Infrastructure/Html/HtmlSanitizerService.cs
```diff
@@ -9,9 +9,9 @@ public class HtmlSanitizerService : IHtmlSanitizerService
 
         public HtmlSanitizerService(IOptions<HtmlSanitizerOptions> options)
         {
-            if (options.Value.Configure != null)
+            foreach(var action in options.Value.Configure)
             {
-                options.Value.Configure.Invoke(_sanitizer);
+                action.Invoke(_sanitizer);
```
sebastienros, line 14:
```suggestion
                action?.Invoke(_sanitizer);
```

src/OrchardCore.Modules/OrchardCore.Markdown/Services/MarkdownPipelineOptions.cs
```diff
@@ -1,10 +1,12 @@
 using System;
+using System.Collections.Generic;
 using Markdig;
 
 namespace OrchardCore.Markdown.Services
 {
     public class MarkdownPipelineOptions
     {
         public Action<MarkdownPipelineBuilder> Configure { get; set; }
+        public IList<Action<MarkdownPipelineBuilder>> Configure2 { get; set; } = new List<Action<MarkdownPipelineBuilder>>();
```
sebastienros, line 10: Remove `set`

deanmarcussen: Done (and extra left over acton remove)

PR #6313, opened 2020-05-31
Use of Assets.json in Preview and Template Modules

ns8482e:
- Moved js.cookie.js to resource.
- Moved contentpreview.edit.js and templatepreview.edit.js to Assets\js folder


src/OrchardCore.Modules/OrchardCore.ContentPreview/ResourceManifest.cs
```diff
@@ -0,0 +1,17 @@
...
+    public class ResourceManifest : IResourceManifestProvider
+    {
+        public void BuildManifests(IResourceManifestBuilder builder)
+        {
+            var manifest = builder.Add();
+
+            manifest
+                .DefineScript("oc-contentpreview-edit")
+                .SetUrl("~/OrchardCore.ContentPreview/Scripts/contentpreview.edit.min.js", "~/OrchardCore.ContentPreview/Scripts/contentpreview.edit.js")
+                .SetVersion("1.0");
```
deanmarcussen, line 14: Can you put this to version `1.0.0` I think we have trouble sometimes if it doesn't have a minor version.

And call it `contentpreview-edit` rather than `oc-` - I don't think we're prefixing any other scripts with oc?

src/OrchardCore.Modules/OrchardCore.ContentPreview/Views/ContentPreview.Button.cshtml
```diff
@@ -10,4 +10,4 @@
 <div id="previewContentItemId" style="display:none" data-value="@Model.ContentItem.ContentItemId"></div>
 <div id="previewContentItemVersionId" style="display:none" data-value="@Model.ContentItem.ContentItemVersionId"></div>
 
-<script asp-src="~/OrchardCore.ContentPreview/Scripts/contentpreview.edit.js" depends-on="admin" at="Foot"></script>
+<script asp-name="oc-contentpreview-edit" depends-on="admin" at="Foot"></script>
```
deanmarcussen, line 13: the dependency should go in the resource manifest I imagine?

deanmarcussen (commented):
Looks generally fine, just a couple of tweaks

src/OrchardCore.Modules/OrchardCore.ContentPreview/ResourceManifest.cs
ns8482e: updated

src/OrchardCore.Modules/OrchardCore.ContentPreview/Views/ContentPreview.Button.cshtml
ns8482e: done

src/OrchardCore.Modules/OrchardCore.ContentPreview/ResourceManifest.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    public class ResourceManifest : IResourceManifestProvider
+    {
+        public void BuildManifests(IResourceManifestBuilder builder)
+        {
+            var manifest = builder.Add();
+
+            manifest
+                .DefineScript("contentpreview-edit")
+                .SetDependencies("admin")
```
deanmarcussen, line 13: I see now @ns8482e when I look closer.

So for front end preview the dependency is jQuery (because we wouldn't want to include the admin js on the front end), and for the template module it is admin.

I think it could just be jquery for both, but you would need to test it (or I maybe wrong) - and check the loading order, to make sure.

Or we just go back to how you had it `<script asp-name='prev' depends-on="admin/jquery">

I'm ok either way

ns8482e: Either way is fine for me, as it's replacing following line - only applicable in `/admin`

`<script asp-src="~/OrchardCore.ContentPreview/Scripts/contentpreview.edit.js" depends-on="admin" at="Foot"></script>`

to named script as below

`<script asp-name="contentpreview-edit" at="Foot"></script>`

Originally i kept `depends-on` in `cshtml` thinking the possibility that admin theme can be changed, and/or when template is overridden.

deanmarcussen: Cool, let me check again tomorrow. I think my comment was wrong.

For infos. 
We always use `admin` as the script and style name (because we do everywhere), and a custom admin theme should still register its own script and style named `admin`. Then there's no problem, or conflict, and things just keep working.

deanmarcussen:
In the `ContentPreview/Index` view there is still a reference to `<script asp-src="~/OrchardCore.Templates/js.cookie.js" depends-on="jquery" at="Foot"></script>`

Unless I missed a change to that in your pr?

Other comments from yesterday all good.

ns8482e:
> In the `ContentPreview/Index` view there is still a reference

It's included in PR - here https://github.com/OrchardCMS/OrchardCore/pull/6313/files#diff-b9f64a7ecaba0b505e133c87af69ea6e

deanmarcussen (approved):

agriffard:
Please fix the conflict.

ns8482e:
> Please fix the conflict.

Done

PR #6314, opened 2020-05-31
Remove form tag from ComingSoon content template

### LINKED ISSUE #6305: Coming soon: Form issue
The form is generated two times causing this problem:

image

ns8482e:
Fixes #6305

### DIFF
--- a/src/OrchardCore.Themes/TheComingSoonTheme/Recipes/Snippets/Content__ComingSoon.liquid
+++ b/src/OrchardCore.Themes/TheComingSoonTheme/Recipes/Snippets/Content__ComingSoon.liquid
@@ -4,13 +4,10 @@
         <div class="row h-100">
             <div class="col-12 my-auto">
                 <div class="masthead-content text-white py-5 py-md-0">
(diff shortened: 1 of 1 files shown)


agriffard:
This does not solve the display problem.
The form needs to be in the `<div class="masthead-content ...` as it was before.

ns8482e:
@agriffard - Fixed using 100% form height.  

Also added style for validation-errors

comingsoon

agriffard:
Ok, thank you.

The workflow is not redirected to the succes page though.
There is an error: `Value cannot be null. (Parameter 'value')`

@sfmskywalker or @MichaelPetrinolis Can you please have a look?

sebastienros:
Can we have the styles in the css file? There is one file already that can be used (bootstrap-oc)

ns8482e:
I added style to ComingSoon as specific style for form is required on this content type Only.

Adding style to css file will make theme not usable by other content types

sebastienros:
I am sure we can find a way in css to apply this style only to the form on this page, or move the form in this template if it shouldn't be used for all types.

sebastienros:
@ns8482e thank you, dean fixed it on a different PR, now the wrapper has a custom css

ns8482e:
Thank you  @deanmarcussen @sebastienros

Issue #6312, opened 2020-05-31
Content Api should use merge ArrayHandling.Replace


deanmarcussen:
The Content ApiController needs to use `ArrayHandling.Replace` when updating existing content, or array values are duplicated.

The same applies to the `ImportAsync()` and potentially other places.

PR #6315, opened 2020-06-01
Prevent array duplication when merging existing content

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6312

This fixes the issue, but I wonder if this behaviour should not be the default when merging content items?

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Controllers/ApiController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Controllers/ApiController.cs
@@ -16,6 +16,8 @@ namespace OrchardCore.Content.Controllers
     [Authorize(AuthenticationSchemes = "Api"), IgnoreAntiforgeryToken, AllowAnonymous]
     public class ApiController : Controller
     {
+        private static readonly JsonMergeSettings UpdateJsonMergeSettings = new JsonMergeSettings { MergeArrayHandling = MergeArrayHandling.Replace };
+
         private readonly IContentManager _contentManager;
         private readonly IAuthorizationService _authorizationService;
         private readonly IStringLocalizer S;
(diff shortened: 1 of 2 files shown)


src/OrchardCore.Modules/OrchardCore.Contents/Controllers/ApiController.cs
```diff
@@ -121,7 +122,7 @@ public async Task<IActionResult> Post(ContentItem model, bool draft = false)
                     return this.ChallengeOrForbid();
                 }
 
-                contentItem.Merge(model);
+                contentItem.Merge(model, new JsonMergeSettings { MergeArrayHandling = MergeArrayHandling.Replace });
```
hishamco, line 125: Store it in static filed instead of new up on every call

```csharp
private static readonly JsonMergeSettings _jsonMergeSettings = new JsonMergeSettings { MergeArrayHandling = MergeArrayHandling.Replace };
```

src/OrchardCore/OrchardCore.ContentManagement/DefaultContentManager.cs
```diff
@@ -783,7 +783,7 @@ private async Task<ContentValidateResult> UpdateContentItemVersionAsync(ContentI
                 await RemovePublishedVersionAsync(updatingVersion, evictionVersions);
             }
 
-            updatingVersion.Merge(updatedVersion);
+            updatingVersion.Merge(updatedVersion, new JsonMergeSettings { MergeArrayHandling = MergeArrayHandling.Replace });
```
hishamco, line 786: Same as above

deanmarcussen: done

src/OrchardCore.Modules/OrchardCore.Contents/Controllers/ApiController.cs
```diff
@@ -16,6 +16,8 @@ namespace OrchardCore.Content.Controllers
     [Authorize(AuthenticationSchemes = "Api"), IgnoreAntiforgeryToken, AllowAnonymous]
     public class ApiController : Controller
     {
+        private static readonly JsonMergeSettings UpdateJsonMergeSettings = new JsonMergeSettings { MergeArrayHandling = MergeArrayHandling.Replace };
```
hishamco, line 19: Use `_updateJsonMergeSettings` because this is the convention we 're using for `private` fields

deanmarcussen: not the convention we seem to be using so much for static privates.

Issue #4845, opened 2019-11-19
Taxonomies Term Localization


urbanit:
Create a ContentType: `CategoryTerm` with Title, Localization Part, AutoRoute
Create a Taxonomy: `Categories` with terms of `CategoryTerm`
Add `CategoryTerm` to taxonomy --> a term is created
Edit term & add localized version --> **Error!**
`No webpage was found for the web address: http://{url}/OrchardCore.ContentLocalization/Admin/Localize`

sebastienros:
We can't localize terms, only taxonomies

urbanit:
How we 'sync' terms? Meaning Term_1_Gr -> Term_1_En and vice versa?

sebastienros:
I talked about it with @jptissot and we had found a way, he will remember I am sure

urbanit:
Cool, I will get an answer from @jptissot and then I will close the issue...

dodyg:
How do you resolve situation where you have a ContentItem that are multi-lingual that has a `Taxonomy Field`? 

![taxonomy](https://user-images.githubusercontent.com/1761249/71308870-2d822c00-240a-11ea-84b1-7b2841f9ad17.png)

dodyg:
@urbanit did you ever found a workaround for this issue?

dodyg:
My workaround is to put a `Language field` in the term and filter the term using `where` filter based on this when showing the taxonomy term list. The downside is that in the admin you have double the length for the terms (English and Arabic).

urbanit:
@dodyg until now, the guides are that we add localization part to taxonomies and you add all localized taxonomies to your content type. Each time you create a content item you check the term from localized tazonomy... What we miss here is that termX(langA) and termX(langB) are kinda of "translated" terms...

dodyg:
The problem with localized taxonomy is that it fails at "Taxonomy Field". Taxonomy field only accepts one version. 

So if you have a taxonomy in three languages, you will need three "Taxonomy Fields", one for each language -- which  is silly.

jptissot:
I encountered the same issue. We either need a LocalizedTaxonomyPicker Field or change the options in the current picker to allow selecting a Localized Taxonomy

sebastienros:
Or can we select the localized version automatically if it exists? From the culture of the currently edited content item?

urbanit:
- [x] Localized Version of Taxonomy with #6316 
- [ ] Add Alias to sync terms of localized taxonomies
[...]

PR #6316, opened 2020-06-01
Taxonomy Localized Field

urbanit:
- [x] Allow Taxonomies Localized Version
- [x] Partially solves #4845

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Taxonomies/Liquid/TaxonomyTermFilter.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Taxonomies/Liquid/TaxonomyTermFilter.cs
@@ -6,14 +6,26 @@ using Fluid;
 using Fluid.Values;
 using Microsoft.Extensions.DependencyInjection;
 using Newtonsoft.Json.Linq;
+using OrchardCore.ContentLocalization.Services;
 using OrchardCore.ContentManagement;
 using OrchardCore.Liquid;
+using OrchardCore.Localization;
 using OrchardCore.Taxonomies.Fields;
 
 namespace OrchardCore.Taxonomies.Liquid
 {
     public class TaxonomyTermsFilter : ILiquidFilter
     {
+        private readonly ILocalizationEntries _localizationEntries;
+        private readonly ILocalizationService _localizationService;
+
+        public TaxonomyTermsFilter(
+            ILocalizationEntries localizationEntries,
+            ILocalizationService localizationService)
+        {
+            _localizationEntries = localizationEntries;
+            _localizationService = localizationService;
+        }        
         public async ValueTask<FluidValue> ProcessAsync(FluidValue input, FilterArguments arguments, TemplateContext ctx)
         {
             if (!ctx.AmbientValues.TryGetValue("Services", out var services))
@@ -25,6 +37,7 @@ namespace OrchardCore.Taxonomies.Liquid
 
             string taxonomyContentItemId = null;
             string[] termContentItemIds = null;
+            string localizedCulture = null;
 
             if (input.Type == FluidValues.Object && input.ToObjectValue() is TaxonomyField field)
             {
@@ -38,6 +51,7 @@ namespace OrchardCore.Taxonomies.Liquid
             {
                 taxonomyContentItemId = jobj["TaxonomyContentItemId"].Value<string>();
                 termContentItemIds = ((JArray)jobj["TermContentItemIds"]).Values<string>().ToArray();
+                localizedCulture = jobj.Root["LocalizationPart"]["Culture"].Value<string>();
             }
             else if (input.Type == FluidValues.Array)
             {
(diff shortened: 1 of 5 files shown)


src/OrchardCore.Modules/OrchardCore.Taxonomies/OrchardCore.Taxonomies.csproj
```diff
@@ -23,6 +23,7 @@
     <ProjectReference Include="..\..\OrchardCore\OrchardCore.Liquid.Abstractions\OrchardCore.Liquid.Abstractions.csproj" />
     <ProjectReference Include="..\..\OrchardCore\OrchardCore.Navigation.Core\OrchardCore.Navigation.Core.csproj" />
     <ProjectReference Include="..\..\OrchardCore\OrchardCore.ResourceManagement\OrchardCore.ResourceManagement.csproj" />
+    <ProjectReference Include="..\OrchardCore.ContentLocalization\OrchardCore.ContentLocalization.csproj" />
```
hishamco, line 26: Why you didn't use `ContentLocalization.Abstraction`?

urbanit: I need ContentLocalization for _localizationEntries... No? Btw, what's the gain of Abstractions?

src/OrchardCore.Modules/OrchardCore.Taxonomies/Liquid/TaxonomyTermFilter.cs
```diff
@@ -38,6 +51,7 @@ public async ValueTask<FluidValue> ProcessAsync(FluidValue input, FilterArgument
             {
                 taxonomyContentItemId = jobj["TaxonomyContentItemId"].Value<string>();
                 termContentItemIds = ((JArray)jobj["TermContentItemIds"]).Values<string>().ToArray();
+                localizedCulture = jobj.Root["LocalizationPart"]["Culture"].Value<string>();
```
sebastienros, line 54: There should be a LocalizedAspect for that. This way you don't depend on a specific part, and it's strongly typed.

urbanit: Not sure if there is, I couldn't get it. Any hint, appreciated..

hishamco: > There should be a LocalizedAspect for that.

I think Seb means `CultureAspect`

deanmarcussen:
Nice one @urbanit this looks like a good start.

Sebastien suggests we should have a `LocalizationAspect` but right now I don't think we have one, so you would have to implement it.

For an example of how aspects work look at the https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.Html/Handlers/HtmlBodyPartHandler.cs for the `BodyAspect`

The idea would be that the `LocalizationAspect` is in the `ContentLocalization.Abstractions` project and the handler implementation in the `ContentLocalization` module.

This way you don't have to reference the module (which is bad!)

I would imagine that you need the same thing for the `LocalizationEntries`. A `LocalizationEntryAspect` which will provide the entries, if the `ContentLocalization` module is activated.

Right now, with this code, unless the `ContentLocalization` module is activiated, you will get a null reference exception when using taxonomies, because the service is not registered.

If we didn't use an aspect for this, we would need a `NullLocalizationEntries` service, registered by default.

But I think another aspect will work well.

This begs the question then if you need the `Localized` setting on the field.
If you removed it, then any taxonomy that was localized, would automatically localize it's fields. (if the aspect provided the entries), and if it wasn't localized the aspect would provide no entries, because the handler would not run.

So that's a question, rather than a request, because you know more about the requirements, but I think it makes sense, that it happens automatically, if you localize the taxonomy?

You might also want to consider the same feature for the `Tags` driver.

hishamco:
@urbanit if you need help, I can handle the localization parts, `LocalizationAspect`, `NullLocalizationEntries` .. etc and leave the remaining things to you ;)

urbanit:
@deanmarcussen thanks for the feedback. You are right that ContentLocalization should not be enabled in order Taxonomies to run smoothly. So, Aspect (now it;s clear how it works) is best approach. 

@hishamco any help is appreciated! If you have time and mood, please go ahead.

About localized and tags: it could be cases that we do not want to localize taxonomies. Tags is one of them... For a tag, maybe it could be user decision to show all related content items or current language only.

src/OrchardCore.Modules/OrchardCore.Taxonomies/OrchardCore.Taxonomies.csproj
hishamco: We don't reference module implementation from another module, I think @deanmarcussen point to that below

hishamco:
> Sebastien suggests we should have a LocalizationAspect but right now I don't think we have one, so you would have to implement it.

@deanmarcussen why we didn't use `CultureAspect`?

sebastienros:
Yeah, CultureAspect is that is the name that we have, sorry

Issue #6323, opened 2020-06-02
ShortCodes Improvements


hishamco:
Yesterday I started playing with shortcode, there's already work done by @deanmarcussen in this area, this is just a continuation of what we already have

The aim is to improve many areas including:

- [x] Create Rich APIs, that allow us to manipulate with each piece in shortcode such as tags, attributes .. etc

- [ ] Enhance shortcode by adding `Attributes` collection coz some advance scenario will need to have them

- [x] Make shortcode development easier, so we should focus on the functionality and leave the heavy lifting to the abstraction APIs to do it the repetitive stuff such as shortcodes parsing

- [x] Allow to target multiple shortcodes

This list may increase according to our needs

sebastienros:
Duplicate of #6199

PR #6324, opened 2020-06-02
Add ShortCodes Rich APIs

hishamco:
Address part of #6323


hishamco:
FYI, the unit tests was passed before merge the change from `[media]` to `[image]` may be it is a good feature to target more than one shortcode with the same logic

This will allow us to write a single shortcode for both `[audio]` & `[video]` for instance

sebastienros:
https://github.com/sebastienros/shortcodes

hishamco:
I saw you are started shortcodes projects several days ago but I don't know what was the path for I

I think the APIs that you introduced is cool for parsing which is the thing that I need to use

If you don't mind I can do a few changes to built on top of the project that you already started, or if you have another plan please let me know

hishamco:
Closing this while the Shortcode implemented based on Seb Shortcodes library, I may add few features on top of it in OCC repo

Issue #6122, opened 2020-05-06
Workflow Liquid evaluator: JavaScriptEncoder vs HtmlEncoder


netwavebe:
I'm using workflows to send a bunch of emails when an order is placed.

Liquid is used to construct the body of the emails. The input data has `"Country": "België"` that I'm rendering with `{{order.UserProfile.Country}}<br />`.

However, workflow liquid evaluator is using JavaScriptEncoder rather than HtmlEncoder, see https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.Workflows/Expressions/LiquidWorkflowExpressionEvaluator.cs#L39

This results in an email with `Belgi\u00EB` in the body.

In some case using JavaScriptEncoder could make sense (as workflow uses Javascript) but in this context HtmlEncoder would be the preferred option.

I think developers should be able to choose which encoder to use in a workflow.

jptissot:
I had the same issue. I believe there was a way to fix it with either `| raw` or `| html | raw`, I forget. 

I agree that some Task / Activity fields should use the HtmlEncoder by default.

netwavebe:
@sebastienros are you sure this issue needs to be closed?

`| raw` does not produce valid html; it outputs `België` but it needs to be `Belgi&euml;`

netwavebe:
In my case, it's the confirmation email of an order, using 292 lines of liquid with data from various sources (product, customer, ...).

There are a lot of places where the JavaScriptEncoder might produce invalid html (product description, client name or address, ...).

I know it's possible to write a custom liquid filter, but I would need to use it almost everywhere.

In the case of an email activity, using HtmlEncoder would make more sense I think.

sebastienros:
| html_encode | raw

sebastienros:
So you are using an Email task? And the body of the email contains liquid?
So yes it should not have been evaluated by the workflow expression.

Reopening to understand why it has been.

netwavebe:
Correct, I'm using an email task and the body contains liquid. This is evaluated by LiquidWorkflowExpressionEvaluator that uses JavaScriptEncoder.

deanmarcussen:
I think the workflow evaluator needs an overload for the encoder (defaults to JavaScript) and the email task can provide the html encoder.
Also the javascript encoder should be loaded from di.

sebastienros:
[...]

PR #6330, opened 2020-06-03
Reviewing encoders usages

sebastienros:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6122

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Workflows/Activities/CreateContentTask.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Workflows/Activities/CreateContentTask.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Text.Encodings.Web;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Localization;
 using Newtonsoft.Json;
@@ -16,11 +17,18 @@ namespace OrchardCore.Contents.Workflows.Activities
     public class CreateContentTask : ContentTask
     {
         private readonly IWorkflowExpressionEvaluator _expressionEvaluator;
+        private readonly JavaScriptEncoder _javaScriptEncoder;
 
-        public CreateContentTask(IContentManager contentManager, IWorkflowExpressionEvaluator expressionEvaluator, IWorkflowScriptEvaluator scriptEvaluator, IStringLocalizer<CreateContentTask> localizer)
+        public CreateContentTask(
+            IContentManager contentManager,
+            IWorkflowExpressionEvaluator expressionEvaluator, 
+            IWorkflowScriptEvaluator scriptEvaluator, 
+            IStringLocalizer<CreateContentTask> localizer,
+            JavaScriptEncoder javaScriptEncoder)
             : base(contentManager, scriptEvaluator, localizer)
         {
             _expressionEvaluator = expressionEvaluator;
+            _javaScriptEncoder = javaScriptEncoder;
         }
 
         public override string Name => nameof(CreateContentTask);
@@ -63,7 +71,7 @@ namespace OrchardCore.Contents.Workflows.Activities
 
             if (!String.IsNullOrWhiteSpace(ContentProperties.Expression))
             {
-                var contentProperties = await _expressionEvaluator.EvaluateAsync(ContentProperties, workflowContext);
+                var contentProperties = await _expressionEvaluator.EvaluateAsync(ContentProperties, workflowContext, _javaScriptEncoder);
                 contentItem.Merge(JObject.Parse(contentProperties));
             }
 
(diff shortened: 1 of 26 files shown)


sebastienros:
/cc @netwavebe @adelemam

Would you mind testing this branch and see if that fixes your issues?

sebastienros:
I made sure that an encoder was only used on the targeted content, to prevent the default Unicode ranges from altering the result. For instance a Username should not be encoded with JavaScriptEncoder like it was, or something containing Greek letter would result in `\u039B` inside the username.

I also used the encoder from the DI.

jtkech:
@sebastienros 

Didn't tried everything but LGTM so i approved it

#6054 is also related to WF content activities and fixes many potential issues, see my comment here https://github.com/OrchardCMS/OrchardCore/pull/6054#issuecomment-637068321. So maybe good to also merge it before, or after but i will have to solve conflicts ;)

netwavebe:
@sebastienros

I confirm this PR fixes my problem: Belgi&#xEB; is rendered in the email!

src/OrchardCore.Modules/OrchardCore.Contents/Workflows/Activities/CreateContentTask.cs
```diff
@@ -16,11 +17,18 @@ namespace OrchardCore.Contents.Workflows.Activities
...
         private readonly IWorkflowExpressionEvaluator _expressionEvaluator;
+        private readonly JavaScriptEncoder _javaScriptEncoder;
 
-        public CreateContentTask(IContentManager contentManager, IWorkflowExpressionEvaluator expressionEvaluator, IWorkflowScriptEvaluator scriptEvaluator, IStringLocalizer<CreateContentTask> localizer)
+        public CreateContentTask(
+            IContentManager contentManager,
+            IWorkflowExpressionEvaluator expressionEvaluator, 
+            IWorkflowScriptEvaluator scriptEvaluator, 
+            IStringLocalizer<CreateContentTask> localizer,
+            JavaScriptEncoder javaScriptEncoder)
```
deanmarcussen, line 27: I wondered if this should be a `JavaScriptEncoder` or `HtmlEncoder`, or even a `NullEncoder`, as it would affect liquid values going into the database (I thought) but I tried it as it is now, and it didn't seem to be an issue.

For the rest LGTM

Skrypt: We pass JSON here so the data will be javascript encoded. 
Unless we pass a HTML field which has some encoded HTML chars in it. Though nothing prevents to decode it later on. I think it's fine like it is. Maybe double check for a HTML field outcome.

jtkech: I also thought about not using any encoder in `CreateContentTask` and `UpdateContentTask` but was not sure.

Maybe what we need here is to do some `SanitizeHtml` as we do in some driver `UpdateAsync()`. If so, as we did handler `ValidateAsync()` to do part of the driver updating work in WF activities and API controllers, we would need to do the same for sanitization.

deanmarcussen: Ok I checked what happens here.

A liquid expression like this
```
{% assign text = "België2" %}
{
  "DisplayText": "{{ text }}"
}
```

gets javascript encoded to this
```
{
  "DisplayText": "Belgi\u00EB2"
}
```

but `JSON` unecodes it (presumably because it respects javascript encoding), so after a `JObject.Parse` it becomes
```
{
  "DisplayText": "België2"
}
```

So what gets saved in the database is ultimately correct.

In which case why bother encoding it in the first place maybe?

For info @jtkech `SanitizeHtml` with the sanitizer is difficult to do on individual properties (possible on a property as a whole), but it was felt that `Workflows` was out of scope - as someone with an edit `Workflow` permission can do so many things already

sebastienros: Why bother? Try with `België2"` and without encoding it would break the json.

sebastienros: I am merging, but it doesn't mean we can't change it if we think it's wrong.

src/OrchardCore.Modules/OrchardCore.Twitter/Services/TwitterClient.cs
```diff
@@ -9,12 +9,11 @@ namespace OrchardCore.Twitter.Services
 {
     public class TwitterClient
     {
-        private readonly HttpClient _client;
+        internal static HttpClient _client = new HttpClient();
```
deanmarcussen, line 12: Regression issue for https://github.com/OrchardCMS/OrchardCore/issues/6373 is here.

Because over here https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.Twitter/Startup.cs

it's registered with the `HttpClientFactory`

So either needs to go back to using the factory, or be registered correctly. Preference @sebastienros ?

PR #6333, opened 2020-06-03
Upgrade packages in Apis.GraphQl

agriffard:
Update https://www.npmjs.com/package/graphiql to 1.0.3


agriffard:
I had to change the file node_modules\graphql-language-service-utils\esm\validateWithCustomRules.js to point to the correct rules.

agriffard:
There is an error using graphiql in admin:
`Uncaught Error: Must provide Source. Received: undefined.`

src/OrchardCore.Modules/OrchardCore.Apis.GraphQL/Assets/js/App.tsx
```diff
@@ -123,7 +123,6 @@ function App() {
                 onToggleExplorer={handleToggleExplorer}
             />
             <GraphiQL
-                ref={ref => (_graphiql = ref)}
```
Jetski5822, line 126: Are you sure this needs to be removed?

agriffard: It is not working anymore with it.

No more `ref` property.

PR #6335, opened 2020-06-03
Upgrade npm dependencies (babel)

agriffard:
Use latest version of babel.

Can be a bit problematic because the generated .js file (not the .min.js), has some extra code added:

`function _typeof(obj) { "@babel/helpers - typeof"; `

so the local .js file will be different than the cdn one.

@jptissot Can you please have look at it?
May be something to check in the gulp file.


jptissot:
I am pretty sure it's ok, the code itself is generated in the minified file, but the string is not in the output.

PR #6339, opened 2020-06-04
Add default edit display option to publish later

deanmarcussen:
The default (standard) edit mode option needs to be added as a shape as well, or anytime someone creates a custom edit mode they are no longer able to select the `Standard` option as it has no shape.

/cc @Piedone I would have pushed this to your pr, but can't edit your prs

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.PublishLater/Views/PublishLaterPart.Option.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.PublishLater/Views/PublishLaterPart.Option.cshtml
@@ -1,0 +1,4 @@
+﻿@{
+    string currentEditor = Model.Editor;
+}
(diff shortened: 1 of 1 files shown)


Piedone:
I swear there was an option under PRs to clone and change it even if it was in a fork somewhere here where there is now only GH Desktop: 
image Something along the lines of a special head in the repository but can't find it.

Piedone:
OK, so I don't know where did the particular feature I'm talking about go but why maintainer modifications are not available for my PRs is because they're made from a fork under an org: https://github.community/t/how-can-we-enable-allow-edits-from-maintainers-by-default/2847/10 (it would be this checkbox: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request-from-a-fork "Allow edits from maintainers"). It's a pity.

PR #6355, opened 2020-06-06
Fixes a 404 for Media drag-thumbnail.png under a virtual folder

jtkech:
The following line in `app.js` doesn't work under a virtual folder, so the media drag / drop image was returning a 404

    self.dragDropThumbnail.src = '/OrchardCore.Media/Images/drag-thumbnail.png';

Easy to check under a virtual folder by using this in the starup

            app.Map("/virtual", innerBuilder =>
            {
                if (env.IsDevelopment())
                {
                    innerBuilder.UseDeveloperExceptionPage();
                }

                innerBuilder.UseStaticFiles();

                innerBuilder.UseOrchardCore();
            });

---

~~I installed the latest version of node and npm and after running the gulp pipeline i had many changed files, so i reverted all the non related changes and only kept the updated `media.js`.~~

~~See in `media.js`, the gulp pipeline added 4 times this kind of changes~~

~~Before:  confirmDialog(_objectSpread({}, $("#deleteFolder").data(), {~~
~~After:   confirmDialog(_objectSpread(_objectSpread({}, $("#deleteFolder").data()), {}, {~~

~~That said i tried it and it works~~


deanmarcussen:
> See in `media.js`, the gulp pipeline added 4 times this kind of changes
> 
> ```
> Before:  confirmDialog(_objectSpread({}, $("#deleteFolder").data(), {
> After:   confirmDialog(_objectSpread(_objectSpread({}, $("#deleteFolder").data()), {}, {
> ```

I think this is just some changes recently from the newer version of babel @jtkech 

Not sure if a full rebuild of the js was done after upgrading babel @agriffard ?

jptissot:
I saw the same thing with the JQuery 3.5.1 branch

jtkech:
Okay, so i updated my first comment to be just related to the issue: the media drag / drop image returns a 404 under a virtual folder. And added some infos to check it under a virtual folder.

jtkech:
I will retry and merge this one to not have to solve other conflicts and having to re-run the gulp pipeline

PR #6361, opened 2020-06-07
Fixes #6102 media field values initialization on editing

### LINKED ISSUE #6102: Media Field loses value when publishing before Edit Page is fully loaded (RC1)
In the AdminTheme, adding some media to a Media Field, publishing, then immediately publishing again will cause the media field to be empty if the second publish happens early enough in the page load process. 

This can happen any time a publish happens fast enough during page load, even with content that's been there for months. I discovered this because I recently updated some Content Types, went through and republished every page on a website to add the new Parts to GraphQL, then had no media left on the site 😢.

jtkech:
Fixes #6102 

- Before having loaded the related media items, we were setting the vuejs model with empty entries, now, even if the related items doesn't exist anymore, we just use the current paths.

- Also, if multiple items were loaded, each item was added one by one in the vuejs model resulting in incomplete updates, now we use another array so that we can update the model in one pass.


jtkech:
I will retry and merge this one before having other conflicts and then have to re-install all packages and re-run the gulp pipeline. @agriffard after you updated `bootstrap-slider` to 11.0.2, it seems that
`OrchardCore.Resources/package-lock.json` was not updated, see the last of my changed files

johnrom:
Confirmed, images are no longer disappearing!

PR #5357, opened 2020-01-24
Reduce length of indexes for Content Fields Indexing

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/5196 and fixes https://github.com/OrchardCMS/OrchardCore/issues/4753

Reduce length to maximum MySql can support (768) and add BigText columns that were missing to LinkField

Also fixes a null ref that can occur when a `Create()` migration fails

Still to test on MySql just to make sure the install completes ok

PR #6362, opened 2020-06-07
Added db indexes to Users and OpenId index tables

jersiovic:


src/OrchardCore.Modules/OrchardCore.Users/Migrations.cs
```diff
@@ -42,10 +42,33 @@ public int UpdateFrom3()
...
             );
 
             return 4;
         }
+        public int UpdateFrom4()
+        {
+            SchemaBuilder.AlterTable(nameof(UserIndex), table =>
+            {
+                table.CreateIndex($"IDX_{nameof(UserIndex)}_{nameof(UserIndex.NormalizedUserName)}", nameof(UserIndex.NormalizedUserName));
+                table.CreateIndex($"IDX_{nameof(UserIndex)}_{nameof(UserIndex.NormalizedEmail)}", nameof(UserIndex.NormalizedEmail));
```
Skrypt, line 55: I'm not sure but shouldn't UserName and Email be unique? Maybe it's just not enforced in the DB but it is in the code.

Skrypt: So, as I see NormalizedUserName could be null for multiple users. And NormalizedEmail could also be null for multiple users. I think these 2 should be non-null and Unique.

deanmarcussen: I think I agree @Skrypt

src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Migrations/OpenIdMigrations.cs
```diff
@@ -80,5 +80,67 @@ public int UpdateFrom2()
...
+            {
+                table.CreateIndex($"IX_{nameof(OpenIdAppByRedirectUriIndex)}_{nameof(OpenIdAppByRedirectUriIndex.RedirectUri)}", nameof(OpenIdAppByRedirectUriIndex.RedirectUri));
+            });
+            SchemaBuilder.AlterTable(nameof(OpenIdAppByRoleNameIndex), table =>
+            {
+                table.CreateIndex($"IX_{nameof(OpenIdAppByRoleNameIndex)}_{nameof(OpenIdAppByRoleNameIndex.RoleName)}", nameof(OpenIdAppByRoleNameIndex.RoleName));
+            });
+            SchemaBuilder.AlterTable(nameof(OpenIdAuthorizationIndex), table =>
+            {
+                table.CreateIndex($"IX_{nameof(OpenIdAuthorizationIndex)}_{nameof(OpenIdAuthorizationIndex.Subject)}", nameof(OpenIdAuthorizationIndex.Subject));
```
deanmarcussen, line 105: Someone needs to check the lengths of these names against postgresql, as we have had issues there before.

@jptissot can you comment?

src/OrchardCore.Modules/OrchardCore.Users/Migrations.cs
jersiovic: Sorry but I don't get how your comment is related with the purpose of the PR. I'm adding indexes for those fields used in any query for making queries faster. But you are asking me to add constraints to the db? 
That should be matter of another PR,isn't it? 
In fact, IMO is better to add only those constraints you really need cause each extra constraint you add to the db hits a bit the performance (I've seen projects where no contraints where used because they hit performance a lot in comparison with not havind them).
But as I mentioned I think it should be another discursion in another PR. 
What do you think guys?

jersiovic: Maybe I didn't unterstood sth

src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Migrations/OpenIdMigrations.cs
jersiovic: I've checked migrations runs properly at sqlite, sql server and postgresql.
Also I've created openid applications and reguested access and refresh token without problem with postgresql and sqlite.
It is enough, or need any extra check?

jersiovic: PostgreSQL 12

deanmarcussen: MySql is the one that is most likely to fail, so we do need to check that. (sorry)

It's generally an exception on migration that occurs with mysql.

Great to have checked postgresql thanks.

jersiovic: ok, I will install it

jersiovic: You were right 🥇 
There was two indexes that where too much long for MySQL.
I've commited a change that doesn't apply those indexes when dialect is MySql

Great job

deanmarcussen: I only know because I broke MySql for 3 months ;)

So chances are if the indexes are too long for MySql they are also too long for MS SQL.
The only difference is MS SQL gives you a warning, instead of an exception, and then throws an exception when you try to update the field with a value that is too long.

(See here for how we had to manage it for the sql fields index https://github.com/OrchardCMS/OrchardCore/pull/5357/files)

The way to know if MS SQL is going to have the same potential issue, is to use SQL Management Studio to drop and create the index, you might get a warning, like this 

`Warning! The maximum key length for a nonclustered index is 1700 bytes. The index 'IDX_AliasPartIndex_Alias' has maximum length of 3001 bytes. For some combination of large values, the insert/update operation will fail.`

src/OrchardCore.Modules/OrchardCore.Users/Migrations.cs
Skrypt: The indexes added in this PR are fine if they are added for performance reason but While we are at it why not fixing the indexes constraints and else... I made a review of the code and found some issues in it and reported it. You can fix it with this PR or not as you wish.

deanmarcussen:
@jersiovic you read my mind. The ideal would be to limit the column sizes so we can get under the MySql limits as well. 

If @kevinchalet agrees that it is ok to reduce the size of some of the columns, which are at 256 characters.

The MySql limit is 768 characters w/4 for utf-8 collation = 3072.

So with each column at 256 * 4 cols * 4 chars = 4096 so if we could get them down to 128 (2 of them?) we have (128 * 2 * 4 = 1024) + (256 * 2 *4 = 2048) = 1024 + 2048 = 3072

jersiovic:
The lack off those indexes is causing Azure SQL server with 50 DTU is consuming 100% of cpu  during a few minutes each half hour approx. This db has only 50 users registered, 1569 tokens, and 711 authorizations
image

I 've commited a change that limit the string columns on those long indexes to a legth of128 (thank you @deanmarcusen for the numbers). I finally limited all three columns instead of two cause it seems it is not expected they grow to that sizes. Unless @kevinchalet  has sth to say

kevinchalet:
> @kevinchalet no, CreateReduceIndexTable/CreateMapIndexTable creates the mapindex table or reduceindex table, primary key and foreign key but it leaves to us the responsibility of defining the other db indexes.

Interesting. For some reasons, I assumed that the Orchard migrations automatically created a SQL index for the columns defined via these methods 😃 

> @kevinchalet by the way do you think we can limit the size of the columns in order to be able to define also those indexes for MySql ?

> If @kevinchalet agrees that it is ok to reduce the size of some of the columns, which are at 256 characters.

Sure. You can take a look at the entity configs we use in the EF Core stores:
  - https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.EntityFrameworkCore/Configurations/OpenIddictEntityFrameworkCoreApplicationConfiguration.cs
  - https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.EntityFrameworkCore/Configurations/OpenIddictEntityFrameworkCoreAuthorizationConfiguration.cs
  - https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.EntityFrameworkCore/Configurations/OpenIddictEntityFrameworkCoreScopeConfiguration.cs
  - https://github.com/openiddict/openiddict-core/blob/dev/src/OpenIddict.EntityFrameworkCore/Configurations/OpenIddictEntityFrameworkCoreTokenConfiguration.cs

jersiovic:
@kevinchalet  thank you 
I see on EF store you limit Subject to 450. No problem if I limit it to only 330 character here to don't reach limits pointed by @deanmarcussen ?

jersiovic:
Ups now I get following exception on SqlLite. You knw when you start a PR but not when it will be finished. It  looks like sth fails at Yessql with ater column comand ... I will see

´´´
Microsoft.Data.Sqlite.SqliteException
  HResult=0x80004005
  Message=SQLite Error 1: 'near "alter": syntax error'.
  Source=Microsoft.Data.Sqlite
  StackTrace:
   at Microsoft.Data.Sqlite.SqliteException.ThrowExceptionForRC(Int32 rc, sqlite3 db)
   at Microsoft.Data.Sqlite.SqliteCommand.<PrepareAndEnumerateStatements>d__64.MoveNext()
   at Microsoft.Data.Sqlite.SqliteCommand.<GetStatements>d__54.MoveNext()
   at Microsoft.Data.Sqlite.SqliteDataReader.NextResult()
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader(CommandBehavior behavior)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader()
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteNonQuery()
   at Dapper.SqlMapper.ExecuteCommand(IDbConnection cnn, CommandDefinition& command, Action`2 paramReader) in /_/Dapper/SqlMapper.cs:line 2797
   at Dapper.SqlMapper.ExecuteImpl(IDbConnection cnn, CommandDefinition& command) in /_/Dapper/SqlMapper.cs:line 568
   at YesSql.Sql.SchemaBuilder.Execute(IEnumerable`1 statements)
   at YesSql.Sql.SchemaBuilder.AlterTable(String name, Action`1 table)
   at OrchardCore.OpenId.YesSql.Migrations.OpenIdMigrations.UpdateFrom3() in D:\PROYECTOS\OrchardCoreLatest\src\OrchardCore\OrchardCore.OpenId.Core\YesSql\Migrations\OpenIdMigrations.cs:line 87

  This exception was originally thrown at this call stack:
    [External Code]
    OrchardCore.OpenId.YesSql.Migrations.OpenIdMigrations.UpdateFrom3() in OpenIdMigrations.cs
```

jersiovic:
I've filled an issue at YesSql repo https://github.com/sebastienros/yessql/issues/244

jersiovic:
I read this "You can not use the ALTER TABLE statement to modify a column in SQLite. Instead you will need to rename the table, create a new table, and copy the data into the new table.".
@sebastienros this is sth we should expect YesSql does for us? Or is it our job at Orchard migrations to do that job of renaming the table, creating a new one and copy its content?

kevinchalet:
> @kevinchalet thank you
> I see on EF store you limit Subject to 450. No problem if I limit it to only 330 character here to don't reach limits pointed by @deanmarcussen ?

Yeah, feel free to opt for a lower limit. I chose a very large value since it's a value that OpenIddict doesn't control (it's typically bound to the user identifier), but in Orchard, a lower value would be fine.

deanmarcussen:
@jersiovic Yes, I feel your pain.

The lack of `ALTER TABLE` in SqLite is dificult.

Here is how we decided to handle it for the `AliasPart` index changes.

https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.Alias/Migrations.cs

- Don't use an alter table, or a second migration to add the index.
- Change the `Create` migration to use the new max column lengths
- Put the index migrations in the `Create` migration

This way the new values will apply to new databases created, but not to old schemas.

It becomes the responsiblity of the user to alter any existing schema if they want to.
It's not perfect, but it doesn't break anything, and is enough for me ;)

If you use a const for the `MaxLength`, then in the `IndexProvider` you can `substring` any values that are greater than the max, and prevent exceptions.

jersiovic:
> @jersiovic Yes, I feel your pain.
> 
> The lack of `ALTER TABLE` in SqLite is dificult.
> 
> Here is how we decided to handle it for the `AliasPart` index changes.
> 
> https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.Alias/Migrations.cs
> 
> * Don't use an alter table, or a second migration to add the index.
> * Change the `Create` migration to use the new max column lengths
> * Put the index migrations in the `Create` migration
> 
> This way the new values will apply to new databases created, but not to old schemas.
> 
> It becomes the responsiblity of the user to alter any existing schema if they want to.
> It's not perfect, but it doesn't break anything, and is enough for me ;)
> 
> If you use a const for the `MaxLength`, then in the `IndexProvider` you can `substring` any values that are greater than the max, and prevent exceptions.

jersiovic:
@deanmarcussen ok, thank you, that's what I did in my Orchard fork, but was not sure is it was  good for you. 
Ok I will truncate data in te index, thank you for the info ;)

agriffard:
Please fix the conflicts in Migrations @jersiovic

agriffard:
Please solve the conflicts.

jtkech:
Just for info, we worked recently with @deanmarcussen on indexes

> As i saw, if we have an unclustered index on A1, A2, An, the only way to get is used is when al the required columns for the produced SELECT OR for a produced JOIN are included in A1, A2, An

> Then, and only if the above condition is true, in the WHERE / OR / AND parts, an unclustered index may be used if it includes some of the colums that are checked, and / or a clustered one for the other columns

> So, when we do a QueryIndex(...) as the produced select is on all columns, unclustered indexes are never used and seem to be useless, unless we define an index including all columns

> Then, when we do a query on Document with an inner join on SomeIndexTable, so on **DocumentId**, an unclustered index may be used but only if it includes **DocumentId**

So, here seems to be a good candidate to define an index including `DocumentId` followed by the columns that may be used in a predicate, unless the index size becomes too large

src/OrchardCore.Modules/OrchardCore.Users/Migrations.cs
Skrypt: @jersiovic Normally your database should use constraints because changes could be coming from other applications using the same database tables. In fact here, the UserIndex should be unique and non-null to avoid having any scripts that could try to spoof a user or email by adding a record in these tables. This is an extra layer of security. In fact this UserIndex is more sensible than any other tables in OC so it makes sense to use database constraints on them to be sure that the data is compliant with our business logic.

Someone could say that if someone hacks this table that's because he got access to the database in the first place. But it is quite easy to run SQL Queries in OC from the Admin UI. It's worth nothing to test if someone could do a simple INSERT from these SQL Queries to see what I mean.

Piedone:
@jersiovic Is this something you'd like to revisit any time soon or should we close?

kevinchalet:
@Piedone I'm not a YesSQL expert, but it seems we now have indexes in place in both the OpenID and Users modules:

https://github.com/OrchardCMS/OrchardCore/blob/d783019615bf873cf2b5cab99b370ac934439d14/src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Migrations/OpenIdMigrations.cs#L27-L163

https://github.com/OrchardCMS/OrchardCore/blob/d783019615bf873cf2b5cab99b370ac934439d14/src/OrchardCore.Modules/OrchardCore.Users/Migrations.cs#L20-L83

Piedone:
Hmm, we do have more indexes today (both in the SQL and YesSql sense) with some overlap with the ones in this PR, but not completely.

I'd argue that SQL indexes are usage-dependent, since they only provide a benefit if your app actually issues a lot of queries for the indexed columns. So, we should only add them if the use case is clearly a very frequent one (like you'll benefit from it if you use the login, not matter how you use it).

So, I'd pass on those changes without some clear arguments for them, but I also see some other schema changes that might be generally relevant.

jersiovic:
> @jersiovic Is this something you'd like to revisit any time soon or should we close?

@Piedone This was a 2020 PR, I'm a bit out of this problem. I only remember that it fixed our performance problems with openid cleaning background task, for an app with 300 users requesting tokens per day. 

It's going to be hard I revisit this soon, so no problem for me if you close this pr

PR #6249, opened 2020-05-22
Extensible filters for contents list

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6012
Fixes https://github.com/OrchardCMS/OrchardCore/issues/4921

Uses zones to provide an extensible `Header` to the contents admin list

~~Need to sort out why the shape `Creating` events don't have a service provider first (https://github.com/OrchardCMS/OrchardCore/issues/6218)~~

PR #6366, opened 2020-06-07
Exclude non listable content types from contents

### LINKED ISSUE #6363: Content item type UI Listing
Hello, If I make the content type unlistable i don't see it the content items list when the user is admin, but if the user is in another role like contributor or author i still see the content type listable

another question? is it possible for the user in role to manage his own content in the dashboard ,I can see a permission for media to manage his own , but nothing similar for content items. of course apart from writing your own module. thank you

hishamco:
Fixes #6363


src/OrchardCore.Modules/OrchardCore.Contents/Controllers/AdminController.cs
```diff
@@ -149,6 +149,13 @@ public async Task<IActionResult> List(ListContentsViewModel model, PagerParamete
                 {
                     query = query.With<ContentItemIndex>(x => x.ContentType.IsIn(listableTypes));
                 }
+
+                // Exclude non listable content types
```
deanmarcussen, line 153: This will break other features (you won't be able to view menus anymore), so you can't do this.

I don't entirely understand the issue (better repro steps might help), but is it related to there being no types available to list when changing a user permissions.

If so, it might be that the `if.Any()` three lines up may need to be changed to be an empty enumerable. Assuming that works.

So try removing the `if.Any()` line and see if that fixes the issue

hishamco: The issue is when you create a non listable content type it's disappear only with administrator users, which is totally wrong

I will try what you suggested before

hishamco: > So try removing the if.Any() line and see if that fixes the issue

Removing the `if` that you point out clear all the results, because the `listableTypes` is empty

deanmarcussen: Maybe you could post some screenshots demonstrating the issue as this just doesn't feel like the right fix, and I can't repro the issue

hishamco: To reproduce the issue:

- Create a content type with `Listable unchecked

- Add new content item for the newly added content type, the item will not display in the content items page, which is the correct behavior

- Create a new user with a `Contributor` role

- Log  out and sign in with the newly added user

- Visit the content items page, you will see the new content item, which is wrong

deanmarcussen: > Removing the if that you point out clear all the results, because the listableTypes is empty

This is correct, and is probably (without trying) the correct fix.

The listable in the ui is based on wether the role can edit the type.

So contributor should list nothing, as they can edit nothing.

Editor can, so it should list correctly.

hishamco: I will try it for `Editor` role, but according to the original issue, the content item(s) show up for non admins roles

hishamco: > So contributor should list nothing, as they can edit nothing. (maybe, maybe not... but there is a bigger issue if they're supposed to be all listed)

Contributors don't view or edit their own content, removing the `if` seems works, but I'm still don't know why `Contributors` doesn't view their content, unless it is by design

hishamco:
@deanmarcussen I did some tests, seems removing the `if` you mentioned fixes the issue nothing but we need a way after this PR to view the content(s) that created by `Contributor`, otherwise the content items page will always be empty

deanmarcussen:
This needs to be reviewed, to understand what content items a `Contributor` should be able to view (I believe it should be only their own content items, none of the content items owned by anyone else).

This pr will fix the issue, but we will also need to resolve how `Contributors` can view only their own content.

The changes made in https://github.com/OrchardCMS/OrchardCore/pull/6249 should allow for us adding that flexibility to the permissions check / filter checks

agriffard:
Please resolve the conflicts and merge the dev branch.

hishamco:
Seems many changes happens in the `AdminController`

Skrypt:
Ok, I found out while working on https://github.com/OrchardCMS/OrchardCore/pull/7104 that if we remove 
`&& HasOwnership(context.User, contentItem)` in the ContentTypeAuthorizationHandler it will remove all the content items that are not owned by the current user from that ContentItems list.

Though, in that PR I kept it and decided to make the content items non-editable by removing the edit button and removing the url set on the content item "title - contentype name".

PR #6368, opened 2020-06-07
Add new field when creating user for generating a random password

JoshLefebvre:
So the system I am currently working on will not have users register there own accounts. Instead an admin will go and create accounts for users within their organization (enterprise). Current default behavior when creating a user from the portal is to create a user with with a null password.

I have added a new field to generate a random password whenever a user is created as this is standard enterprise behavior. 
image

**A few questions regarding this behavior:**
1. Is this something that the OC project would be interested in ? Otherwise skip all other questions and I will keep it in my local fork.
2. I am using a password generator I found from a github snippet that uses the set Identity options, therefore this could easily be configured by the user in the startup.cs or from registration settings etc. However is the display driver the best place for this functionality. If I didn't have to inject  userManager.Options.Password I would use a static helper class. I was thinking perhaps a PasswordService instead ?
3. I would like to send the user an email upon account creation (standard enterprise behavior), I see that there is a User CreatedAsync Handler, however I would like to customize my email message so perhaps the OrchardCore.User module would not be the best place. 

Obviously the easiest solution here would be to use a 3rd party SSO provider, however not all my tenants will be using SSO and some may require local OC users.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=6368) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=6368) before we can accept your contribution.<br/><br/>:x: JoshLefebvre [sign now](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=6368)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OrchardCMS/OrchardCore?pullRequest=6368) it.</sub>

sebastienros:
I think there that ASP.NET Identify services already create a random password if none is provided. This is what should already happen when we use create a local account for oauth. /cc @MichaelPetrinolis 

Then about sending an email, you should use the Workflow module for that.

And at some point you will have to show us what you are building. I like what I saw in your GitHub repositories.

JoshLefebvre:
Awesome! Didn't even think to use workflows! I will give this a try. 
Haven't had much of a chance to play around with the OC oauth modules in some time, I will read through the documentation again to see if this fits my use case.

Yes I am currently heavily invested in Orchard Core for a business I will be launching in the next few months. As a result I have developed some custom modules I will require. If they align with any future roadmaps I would be happy to integrate them in!

MichaelPetrinolis:
@sebastienros when we create a user without password, the passwordhash is set to null.

sebastienros:
> the passwordhash is set to null

So I assume it's safer than a password actually, meaning that no password could result in a null password has with brute force?

MichaelPetrinolis:
indeed, if passwordhash is null then no hash checks take place and fails immediately. https://github.com/dotnet/aspnetcore/blob/605c522fa3e875fd6d3aefa783a71d1745b7e4c7/src/Identity/Extensions.Core/src/UserManager.cs#L845
So instead of generating a password it would be better to send a reset password link to the enterprise users or use an azure ad instead

sebastienros:
@JoshLefebvre so you see, no need for using any custom random password generator. Just ensure that the field is null (not that the password is an empty string) and you are good.

JoshLefebvre:
Okay yeah actually this approach works much better and seems more secure as it allows the user to set their own password upon getting the reset url via email. 

I still think there should be an option to trigger a "Send Reset Password Email" upon creation. I took a look at workflows and there doesn't appear to be a "Send Reset Password Email" action. For the time being I will create a custom UserCreatedEventHandler in one of my modules based on the following logic:
image

Thanks for the help guys !

sebastienros:
Good idea, we should have such a task. You could just create one in the OrchardCore.Users.ResetPassword feature and submit a PR. Would be useful.

PR #6369, opened 2020-06-08
Fixes #6344 shape items ordering

### LINKED ISSUE #6344: WidgetsList Part shuffles in UI
Added a WidgetList Part in Custom Content Type and tried to add a widgets more than 15, when previewing the order shuffles. 

Kindly do the needful.

Note : When the number of widgets < 16, the UI works perfectly.

jtkech:
Fixes #6344 

- **Widget items are ordered when editing** through a layer or a widget list part, then for rendering, when they are added to the zone shape as `IPositioned` items, they don't provide any `Position` as **we would just need to preserve the original order**.

- But shape items are ordered by using `_items.Sort(FlatPositionComparer.Instance)` **which doesn't preserve the original order** when positions are equal and there are more than 16 items.

- One solution would be to use `OrderBy()` but that may have some perf impact, so here when adding a widget item to the zone shape, **we just pass its original position**.


deanmarcussen:
> _items.Sort(FlatPositionComparer.Instance) which doesn't preserve the original order when positions are equal and there are more than 16 items.

Reads like an issue in the `FlatPositionComparer.Instance` if it fails to behave correctly when there are more than 16?

thomasebsmith:
> Reads like an issue in the `FlatPositionComparer.Instance` if it fails to behave correctly when there are more than 16?

It's actually due to the fact that `List<T>.sort` is a stable insertion sort when there are 16 or fewer elements in a list, but turns into an unstable Quicksort if there are more (see the "Remarks" section [here](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.sort?view=netcore-3.1)).

RajasekarPR:
@jtkech @deanmarcussen Any update on this ?

jtkech:
@thomasebsmith you're right and here we don't have any `IPositioned.Position`, we just need to preserve the original order of items as it has been don while editing.

@RajasekarPR i think it will come soon in the dev branch now that rc2 has been shipped

RajasekarPR:
@jtkech  Thanks 👍

deanmarcussen:
@thomasebsmith good find on the Sort that makes sense.

@jtkech my concern is that this isn't just an issue with widget ordering. I can produce the same problem with menus, and therefore probably with taxonomies.

So without changing to `OrderBy` we would have to fix it in many places. But equally I understand  the perf concern.

jtkech:
@deanmarcussen 

> So without changing to `OrderBy` we would have to fix it in many places.

Yes ;) Or use `OrderBy`

> But equally I understand the perf concern.

Maybe no more a problem with the libraries that have been rewritten, we would need to do benchmarks but too lazy ;) I will do some research

jtkech:
@deanmarcussen 

I think that's okay [since this PR](https://github.com/dotnet/corefx/pull/24116) `Use Array.Sort instead of custom QuickSort implementation in LINQ's OrderBy`. So i will suggest to use `OrderBy()`.

deanmarcussen:
@jtkech good research :)

So looks like the way to go

jtkech:
@deanmarcussen 

Indeed it also affects e.g. the admin menu if you have more than 16 root menu items

Without applying this PR

Menu

If applying this PR

Menu2

jtkech:
Maybe fixes #6484

PR #6370, opened 2020-06-08
Make import integration tests behave with dotnet test

deanmarcussen:
Fixes an issue where the test ran correctly under VS, but not under `dotnet test`

Note: We will still see an error log in the test run, as the controller logs an error


jtkech:
It also run correctly with `dotnet test` on my machine.

That said what you did also works in both cases, so i will approve it if it fixes an issue on your side

Is it because we throw a `ValidationException` in `ImportAsync()`? If so maybe good in the related controllers to do a catch, at least on the `ValidationException`, and then return e.g. a `Problem` result.

Issue #1868, opened 2018-05-30
Ability to disable cache during development


d-miles:
Would like to see the ability to disable caching during development, either with a setting or even a variable on the Layout or something like that.

Issue is that I have a landing page with lots of cached pieces, each having a different context as well as different expirations. If I want to modify the markup of these pieces, I have to find a way to invalidate them all, or change the name and the expiration, then change back later.

Also one thing I've noticed is that if I have a named cached block with a expiration:
```
{% cache "my-cache", expires_after: "06:00:00" %}
Hello world
{% endcache %}
```

I can't invalidate this block by setting `expires_after` to a lower value, like `00:00:01` (by the way `00:00:00` is invalid, but is valid for `{% shape %}` caching, may want to make it consistent), unless I change the name of the cached block. Basically, I think that maybe the cached block's name should include the expiration or depend on the expiration so that if I change it, it resets it or at least adheres to the new expiration?

Would like to have some kind of discussion regarding this!

sebastienros:
/cc @paynecrl97 

I think it's a very good idea. Not sure how it should be designed, maybe like we handle CDN in O1, with a setting that would have these:

- Enabled (cache is enabled)
- Disabled (cache is disabled)
- Production only (enabled in production, disabled in development)

Also should it apply to all cached data (fragments, output cache), I think so, but other cached business data, just html cache.

This should be relatively easy to implement. As a setting on the `ISite` object. Any html caching module should take the value into account.

paynecrl97:
Agree, and yeah- it should be easy enough to do.

Maybe the three options should be

- Enabled
- Disabled
- From Config (the config file determines whether or not it's enabled)

paynecrl97:
FYI as a workaround- the `{% cache ... %}` tags are designed to work with or without the dynamic cache feature enabled.

In theory, you should be able to just disable that feature locally and the tags will just fall through as if they weren't there.

PR #6390, opened 2020-06-10
Add site settings to disable dynamic cache

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6331
Fixes https://github.com/OrchardCMS/OrchardCore/issues/1868

Options are the same as resources, plus one extra setting for `Enabled and debug` to get the debug shapes.

Screenshot 2020-06-10 at 13 53 18

Also allows configuration of shape debug mode
Screenshot 2020-06-10 at 13 56 19

Tried to keep this as simple as possible, so pre caching operations are still done (create cache context etc) and the `DynamicCache` just becomes a noop when caching is disabled.

Maybe there are other places / features that need to use this that I have missed?


deanmarcussen:
Removed note on purging caches as realised this would only work in development, not across an actual distributed cache instance.

src/OrchardCore.Modules/OrchardCore.Settings/Views/Settings.Edit.cshtml
```diff
@@ -97,4 +100,19 @@
             </div>
         </div>
     </div>
+    <div class="tab-pane fade" id="cache" role="tabpanel" aria-labelledby="cache-tab">
+        <h5>@T["Settings for dynamic caches"]</h5>
+        <div class="row form-group">
+            <div class="col-xl-6" asp-validation-class-for="CacheMode">
+                <label asp-for="CacheMode">@T["Cache Mode"]</label>
+                <select asp-for="CacheMode" class="form-control">
+                    <option value="@CacheMode.FromConfiguration.ToString()">@T["From environment — enabled in <em>Production</em>, disabled otherwise"]</option>
```
jtkech, line 109: Maybe `FromEnvironment`

agriffard:
Please fix conflict and take @jtkech 's remark into account if needed.

src/OrchardCore.Modules/OrchardCore.Settings/Views/Settings.Edit.cshtml
deanmarcussen: I think I agree @jtkech but copied from the `Resources` where it is `FromConfiguration`, so didn't want to change it

i.e.
```
                case ResourceDebugMode.FromConfiguration:
                    options.DebugMode = !_env.IsProduction();
                    break;
```

The only change I made was to call it `CacheMode` not `CacheDebugMode` because there it was more about enabling / disabling the cache (which isn't a debug exactly, or putting it into debug).

so ended up with 
```
        FromConfiguration,
        Enabled,
        DebugEnabled,
        Disabled
```

jtkech: Okay cool then

jtkech:
> Please fix conflict and take @jtkech 's remark into account if needed.

That's okay

PR #6406, opened 2020-06-12
Preserves shape name if type only differs by display mode

### LINKED ISSUE #6206: Display mode for parts and named shapes issue.
After creating display mode for various custom parts (added by this pr  #5329) and then tried to alter alter shapes position in my Content.cshtml using with the new part display mode  : 
```
   @await DisplayAsync(Model.Content.TitlePart)
   @{Model.Content.Remove("TitlePart")};
``` 
This does not work. Instead you have to use the name of the custom display mode eg : 

```
 @await DisplayAsync(Model.Content.Named("TitlePart-TitlePart_Display__MyCustomDisplayMode"))

  @{Model.Content.Remove("TitlePart-TitlePart_Display__MyCustomDisplayMode")};

```
This is a bit cumbersome , having to know the display mode applied to work with the shape.

Would it be possible to preserve the original part name , so you could work with shape part name , regardless of the display mode. ?

So shape.Metadata.Name = "TitlePart" will not change based on the display mode ? 
@deanmarcussen 
Thanks.

jtkech:
Fixes #6206 

- **Allows to use e.g. `DisplayAsync(Model.Content.TitlePart)` even if using a custom display mode**, this in place of having to use e.g.

      DisplayAsync(Model.Content.Named("TitlePart-TitlePart_Display__MyCustomDisplayMode"))

    But maybe we don't want that `DisplayAsync(Model.Content.TitlePart)` renders the title in any display mode, let me know. Otherwise, we can still control the rendering through placement as the `Differentiator` is still different for 2 display modes.

    Finally here we assume that we are not combining / rendering the same part / field in 2 display modes at the same time, otherwise we would have a name conflict.

- **Updates for parts the alternates related to a given display mode**, as already done for fields.


deanmarcussen:
Could we potentially maintain the `Differentiator` name somewhere on the Shape as well.

Maybe it would be useful as a property on `ShapeMetadata` or maybe dynamically on the shape.

Super useful for debugging to know what the `Differentiator` is, and here we would lose it.

But I like the idea of keeping the Name consistent with that of the `TitlePart`. 

Does that work with fields as well, as they would be `PartName-FieldName` then? I think so, just checking.

jtkech:
@deanmarcussen 

> Super useful for debugging to know what the Differentiator is, and here we would lose it.

We don't lose it, only the shape name is preserved regardless its differentiator if the type only differs by the display mode, the `Differentiator ` is kept as before

> Does that work with fields as well, as they would be PartName-FieldName then?

Yes ;)

src/OrchardCore/OrchardCore.DisplayManagement/Views/ShapeResult.cs
```diff
@@ -113,7 +114,7 @@ private async Task ApplyImplementationAsync(BuildShapeContext context, string di
 
             ShapeMetadata newShapeMetadata = newShape.Metadata;
             newShapeMetadata.Prefix = _prefix;
-            newShapeMetadata.Name = _differentiator ?? _shapeType;
+            newShapeMetadata.Name = _name ?? _differentiator ?? _shapeType;
```
deanmarcussen, line 117: @jtkech what I meant by the differentiator no longer being available to the debugger, is because it is not part of the `ShapeMetadata`, so when you change it to a shortened `Name` (which is cool, and useful) it becomes harder to debug, or use the `ConsoleLog` to find meaningful information about how to use placement on that shape.

I wouldn't mind seeing it as a property on `ShapeMetadata` if @sebastienros will let us get a way with it :), or even just `newShape.Differentiator = _differentiator` would be enough for me.

We set the `menuShape.Differentiator` the second way, even though there it is not placeable as it is not from driver.

jtkech: @deanmarcussen okay i got it ;)

I will take a look this night

deanmarcussen: @jtkech I like it :)

I think there are templates in the blog recipe using the old names `Content-BlogPost.liquid`

PR #6411, opened 2020-06-12
Allows ZoneShapes to be overridden

### LINKED ISSUE #6407: Can't override ZoneShapes
I was trying to override the `ContentZone` implementation in `ZoneShapes.cs` to get some different behaviour for some tabbed shapes, but could not, because it is not attributed with 
`[Feature(Application.DefaultFeatureId)]`

I've just removed the service descriptors for it instead, but was there a reason it didn't get attributed along with the other core shapes @jtkech ?

jtkech:
Fixes #6407


deanmarcussen:
Awesome thanks @jtkech

Issue #6359, opened 2020-06-06
Unable to Store OrchardCore.Media.Azure Connection Strings in Azure KeyVault


JoshLefebvre:
I am trying to store some of my appsettings value in my Azure KeyVault for security purposes, however due to the current naming conventions used in the appsettings.json files I am unable to due to illegal dot characters in some of the setting names:
![image](https://user-images.githubusercontent.com/22512098/83955830-4f8b3180-a825-11ea-88ad-5c0902895918.png)

From the Azure Documentation:

> Colons, which are normally used to delimit a section from a subkey in ASP.NET Core configuration, aren't allowed in key vault secret names. Therefore, two dashes are used and swapped for a colon when the secrets are loaded into the app's configuration.

As far as I know there aren't any similar configurations to allow . (periods) in azure keyvault names. Has anyone figured out a workaround for this without having to overload IKeyVaultSecretManager ?

jtkech:
We also allow underscores (which is the new rule) in place of dots, this to be compatible with linux env variables, so try the following

    OrchardCore_Media_Azure--ConnectionString

JoshLefebvre:
Yea Az Key Vault is still complaining:
![image](https://user-images.githubusercontent.com/22512098/83957310-34282280-a835-11ea-9352-127dbcd1a710.png)
Seems only dashes and alpha numeric.

JoshLefebvre:
Do you think it would be worth filing a change request ?

jtkech:
> Do you think it would be worth filing a change request ?

Yes i think it is worth to have an override of [KeyVaultSecretManager](https://github.com/Azure/azure-sdk-for-net/blob/779ddcd153436ce9105c6d3765463e3da26d52d3/sdk/extensions/Azure.Extensions.AspNetCore.Configuration.Secrets/src/KeyVaultSecretManager.cs#L24), and maybe an extension helper using [AddAzureKeyVault](https://github.com/Azure/azure-sdk-for-net/blob/779ddcd153436ce9105c6d3765463e3da26d52d3/sdk/extensions/Azure.Extensions.AspNetCore.Configuration.Secrets/src/AzureKeyVaultConfigurationExtensions.cs#L25), e.g. `AddOrchardCoreAzureKeyVault()` that we could use at the app level.

This through a separate package e.g. `OrchardCore.Azure.KeyVault` as we don't want to reference azure packages from the app by default, as we do for `OrchardCore.Shells.Azure`

[...]

PR #6422, opened 2020-06-14
Add new module for OrchardCore.KeyVault.Azure

JoshLefebvre:
Based on conversations had in https://github.com/OrchardCMS/OrchardCore/issues/6359


JoshLefebvre:
If this looks good, I would be happy to add some further documentation on how to use

deanmarcussen:
Looks great, will take a closer look tomorrow

+ 1 for some docs  :)

src/OrchardCore/OrchardCore.KeyVault.Azure/Extensions/AzureKeyVaultWebHostBuilderExtension.cs
```diff
@@ -0,0 +1,40 @@
...
+                    clientId, 
+                    clientSecret,  
+                    new CustomKeyVaultSecretManager()
+                );
+            });
+
+            return builder;
+        }
+
+
```
jtkech, line 38: remove 2 extra lines

src/OrchardCore/OrchardCore.KeyVault.Azure/OrchardCore.KeyVault.Azure.csproj
```diff
@@ -0,0 +1,20 @@
...
+
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\OrchardCore.Abstractions\OrchardCore.Abstractions.csproj" />
+  </ItemGroup>
+
+
```
jtkech, line 15: remove one extra line

src/OrchardCore/OrchardCore.KeyVault.Azure/Services/CustomKeyVaultSecretManager.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// See https://github.com/OrchardCMS/OrchardCore/issues/6359.
+    /// </summary>
+    public class CustomKeyVaultSecretManager : DefaultKeyVaultSecretManager
+    {
+        public override string GetKey(SecretBundle secret)
+        {
+            var key = secret.SecretIdentifier.Name.Replace("---", "_").Replace("--", ":");
+            return key;        
+        }
+
```
jtkech, line 21: remove extra line

src/OrchardCore.Build/Dependencies.props
```diff
@@ -22,6 +22,7 @@
     <PackageManagement Include="MailKit" Version="2.6.0" />
     <PackageManagement Include="Markdig" Version="0.20.0" />
     <PackageManagement Include="MessagePack" Version="2.1.115" />
+    <PackageManagement Include="Microsoft.Extensions.Configuration.AzureKeyVault" Version="3.1.3" />
```
jtkech, line 25: Move it to the `Dependencies.AspNetCore.props` file

    <PackageManagement Include="µ.Extensions.Configuration.AzureKeyVault" Version="$(AspNetCoreVersion)" />

For `AspNetCoreVersion` we currently use `3.1.5`, so we will have to check that it still builds.

src/OrchardCore/OrchardCore.KeyVault.Azure/Services/CustomKeyVaultSecretManager.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// Key Vault Input: "OrchardCore--OrchardCore---Shells---Database--ConnectionString".
+    /// Output: "OrchardCore:OrchardCore_Shells_Database:ConnectionString".
+    /// See https://github.com/OrchardCMS/OrchardCore/issues/6359.
+    /// </summary>
+    public class CustomKeyVaultSecretManager : DefaultKeyVaultSecretManager
+    {
+        public override string GetKey(SecretBundle secret)
+        {
+            var key = secret.SecretIdentifier.Name.Replace("---", "_").Replace("--", ":");
+            return key;
```
jtkech, line 19: Maybe just return without using the key variable, but as you want as it may be useful for debugging ;)

src/OrchardCore/OrchardCore.KeyVault.Azure/Services/CustomKeyVaultSecretManager.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace OrchardCore.Azure.KeyVault.Services
+{
+    /// <summary>
+    /// A custom override of the DefaultKeyVaultManger class that retrieves secrets from azure keyvault and translates ---
+    /// to the OC format using single underscore (illegal character in Azure KeyVault) and -- to : to define a section
+    /// Examples:
+    /// Key Vault Input: "OrchardCore--OrchardCore---Shells---Database--ConnectionString".
+    /// Output: "OrchardCore:OrchardCore_Shells_Database:ConnectionString".
+    /// See https://github.com/OrchardCMS/OrchardCore/issues/6359.
```
jtkech, line 12: Maybe this would have a better place in a `README.md` file under our `docs` folder

But that should not block this PR ;)

src/OrchardCore/OrchardCore.KeyVault.Azure/Services/CustomKeyVaultSecretManager.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// </summary>
+    public class CustomKeyVaultSecretManager : DefaultKeyVaultSecretManager
```
jtkech, line 14: Why not just naming it `AzureKeyVaultSecretManager`

JoshLefebvre:
Thanks @jtkech. Made all above requested changes and added the documentation.

jtkech:
@JoshLefebvre perfect i approved it

Just one last thing, maybe not the right place for the readme as `OrchardCore.KeyVault.Azure` is not a module but a core library. It does not bother me but i don't often work on documentation, i'm too lazy ;) so we need a confirmation, @agriffard can you confirm?

deanmarcussen:
Looks good @JoshLefebvre 

I fixed a couple of typos in the docs.

Can you move it to the core section as @jtkech suggested, and also it needs an entry in the `mkdocs.yml` or it won't display on readthedocs

xperiandri:
Is it ready to merge?

src/docs/reference/core/KeyVault.Azure/README.md
```diff
@@ -0,0 +1,40 @@
+# Azure Key Vault (`OrchardCore.Azure.KeyVault`)
+The Azure Key Vault configuration provider adds app configuration values from the Azure Key Vault in order to safeguard your cryptographic keys and secrets used by your app. It also contains custom override of the DefaultKeyVaultManger class that retrieves secrets from Azure Key Vault and translates ---
```
agriffard, line 2: DefaultKeyVaultManger => DefaultKeyVaultManager

src/docs/reference/core/KeyVault.Azure/README.md
```diff
@@ -0,0 +1,40 @@
...
+to an underscore (_)  and -- to a colon (:). Both underscores and colons are illegal characters in Azure KeyVault.
+
+Example:
+Key Vault Input: "OrchardCore--OrchardCore---Shells---Database--ConnectionString".
+Output: "OrchardCore:OrchardCore_Shells_Database:ConnectionString".
+See https://github.com/OrchardCMS/OrchardCore/issues/6359.
+
+## Configuration
+You'll need to specify the name of your Azure Key Vault and [register a service principle](https://docs.microsoft.com/en-us/azure/key-vault/general/group-permissions-for-apps) in Active Directory for accessing your key vault using an access control policy.
```
agriffard, line 11: principle => principal

src/docs/reference/core/KeyVault.Azure/README.md
```diff
@@ -0,0 +1,40 @@
...
+```json
+"OrchardCore_Azure_KeyVault": {
+    "KeyVaultName": "", // Set the name of your Azure Key Vault.
+    "AzureADApplicationId": "", // Set the Azure AD Application Id
+    "AzureADApplicationSecret": "" //Set the Azure AD Application Secret
+}
+```
+
+!!! note
+    You should **never check in your client secret into source control** as this defeats the purpose of using a key vault in the first place. Instead set your client secret as an environment variable on your machine, create a seperate azurekeyvault.json file and add it to your `.gitignore`, or use user secrets.
```
agriffard, line 22: Instead set => Instead, set

agriffard, line 22: seperate => separate

src/docs/reference/core/KeyVault.Azure/README.md
agriffard, line 2: newline after title.

src/docs/reference/core/KeyVault.Azure/README.md
agriffard, line 11: You'll need => You will need

src/OrchardCore/OrchardCore.KeyVault.Azure/Extensions/AzureKeyVaultWebHostBuilderExtension.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace OrchardCore.Azure.KeyVault.Extensions
+{
+    public static class AzureKeyVaultWebHostBuilderExtension
+    {
+        /// <summary>
+        /// Adds Azure Key Vault as a Configuration Source.
+        /// </summary>
+        /// <param name="builder">The web host builder to configure.</param>
+        /// <returns>The web host builder.</returns>
+        public static IHostBuilder UseOrchardCoreAzureKeyVault(this IHostBuilder builder)
```
sebastienros, line 15: `AddOrchardCoreAzureKeyVault`

src/docs/reference/core/KeyVault.Azure/README.md
```diff
@@ -0,0 +1,40 @@
...
+```csharp
+using OrchardCore.KeyVault.Azure;
+public class Program
+{
+    public static Task Main(string[] args)
+        => BuildHost(args).RunAsync();
+
+    public static IHost BuildHost(string[] args) =>
+        Host.CreateDefaultBuilder(args)
+            .UseOrchardCoreAzureKeyVault()
```
sebastienros, line 35: `AddOrchardCoreAzureKeyVault`

src/docs/reference/core/KeyVault.Azure/README.md
```diff
@@ -0,0 +1,40 @@
+# Azure Key Vault (`OrchardCore.Azure.KeyVault`)
+The Azure Key Vault configuration provider adds app configuration values from the Azure Key Vault in order to safeguard your cryptographic keys and secrets used by your app. It also contains custom override of the DefaultKeyVaultManger class that retrieves secrets from Azure Key Vault and translates ---
+to an underscore (_)  and -- to a colon (:). Both underscores and colons are illegal characters in Azure KeyVault.
+
+Example:
+Key Vault Input: "OrchardCore--OrchardCore---Shells---Database--ConnectionString".
+Output: "OrchardCore:OrchardCore_Shells_Database:ConnectionString".
+See https://github.com/OrchardCMS/OrchardCore/issues/6359.
+
+## Configuration
```
sebastienros, line 10: Can we have a small example that uses directly `AzureKeyVaultSecretManager` such that standard ASP.NET tutorials can be followrd?

JoshLefebvre: Do you mean an example of IConfiguration? I have been using it to retrieve my media and database connection strings. For example in DatabaseShellConfigurationSources.cs:
image

Using the following Keys:
OrchardCore--OrchardCore---Shells---Database--ConnectionString
OrchardCore--OrchardCore---Shells---Database--DatabaseProvider (Not actually a secret, but required to get the section)

deanmarcussen: What Seb was asking for is something that matches the getting started guides from the ASP.NET docs here https://docs.microsoft.com/en-us/aspnet/core/security/key-vault-configuration?view=aspnetcore-3.1

Where they are using the `AzureServiceTokenProvider` callbacks and such.

Which makes me also wonder if we should not be using the callback technique, rather than having our own configuration secrets for it?

jtkech: @deanmarcussen 

Here @JoshLefebvre just override the `DefaultKeyVaultSecretManager` with a new `AzureKeyVaultSecretManager` and provide an `AddOrchardCoreAzureKeyVault()` extension that uses it and other config values.

But people can just follow the aspnet documentation and then use `new AzureKeyVaultSecretManager()` in place of `new DefaultKeyVaultSecretManager()`

So maybe just remove the `AddOrchardCoreAzureKeyVault()` that is just an example on how to use the new `AzureKeyVaultSecretManager` and let people follow the doc and do their own extension if they want.

xperiandri:
Cool, so is only getting started guide left?

src/docs/reference/core/KeyVault.Azure/README.md
xperiandri: Do I understand your implementation right that you have added replacement of `---` to `_` along with default `--` to `:` behavior?

agriffard:
Please resolve conflict.

MarcBruins:
The preferred way of communicating with the Azure Keyvault for a web app is through Managed identity rather then a service principal with a object id and password. Can we make sure that that scenario is also supported? I can work on this if you like

Skrypt:
@MarcBruins I think that ultimately we should have an external OC Management tool that could be deployed as an Azure Function that would allow to configure all things related with the Azure Portal services. Here, what I don't like is the fact that OC configures it's own Azure Services. This can easily lead to security issues.

Also, I think that it could be a good way to start experiment things with the new Azure SDK.

agriffard:
Can you please fix the conflicts @JoshLefebvre?

agriffard:
Conflicts solved.

src/docs/reference/core/KeyVault.Azure/README.md
jtkech: @xperiandri  Exactly, this is what @JoshLefebvre did ;)

sebastienros: Is that something that is common? Either in OC or ASP.NET with Keyvault?

src/docs/reference/core/KeyVault.Azure/README.md
```diff
@@ -0,0 +1,41 @@
...
+You will need to specify the name of your Azure Key Vault and [register a service principal](https://docs.microsoft.com/en-us/azure/key-vault/general/group-permissions-for-apps) in Active Directory for accessing your key vault using an access control policy.
+
+```json
+"OrchardCore_Azure_KeyVault": {
+    "KeyVaultName": "", // Set the name of your Azure Key Vault.
+    "AzureADApplicationId": "", // Set the Azure AD Application Id
```
sebastienros, line 17: Is that required when using managed identity in Azure? I believe we can also associated an Azure Web App for instance to be allowed to access a KV account without anything to setup.

mariojsnunes: Yes, it's the AzureServiceTokenProvider. I got it working on my OrchardCore project.
Some notes:
I was only able to get in working in a .net standard project (which is fine).
I think there is a new API for .net core, but I couldn't get it to work at the time (6 months ago)

deanmarcussen:
Just to note this will need to move to the new azure library package https://docs.microsoft.com/en-us/dotnet/core/compatibility/3.1-5.0#azure-microsoft-prefixed-azure-integration-packages-removed

agriffard:
@JoshLefebvre Can you please apply the changes sugested by @deanmarcussen ?

JoshLefebvre:
@agriffard yes sorry for the hiatus. I have a lot more free time now and hope to actively start contributing to this project again. I will try to get this done today

JoshLefebvre:
@deanmarcussen  @agriffard so I've updated the Key Vault package to make use of the new Azure SDK package Azure.Extensions.AspNetCore.Configuration.Secrets. I have also added the [Azure.Identity](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/identity/Azure.Identity/README.md) package which is now the recommended way to authenticate with Azure resources via Active Directory token authentication and should resolve the issues people were mentioning above .

Azure Identity supports:
- Service principal authentication
- Managed identity authentication
- User principal authentication

agriffard:
Can you please review it @deanmarcussen ?

deanmarcussen:
Great, thanks for updating this @JoshLefebvre - they've definitely improved the libraries a lot since moving to the new Azure packages.

Couple of thoughts, and I will bring @jtkech in here as well, as he may agree / disagree

Because this is a program level configuration, I don't think it should be using the `OrchardCore` appsettings section. That would suggest it is configurable at a tenant level, but as it is wholly for the host, it isn't tenant configurable. (please correct me if I'm wrong)

I'm also thinking it's not a module. Because we can't load it modularly, for the same reasons.

Which makes it more like the Shells.Azure configuration, which we did as a seperate nuget package, as part of the `OrchardCore` folder, rather than as part of modules.

Lastly we need to pick a name. Because it's KeyVault Configuration - rather than KeyVault itself, which would have a read/write option, and we might be using it for read / write (i.e. it'd be good to do a `Shells.KeyVault` version at some point with read/write, and I'm hoping to do a Secrets KeyVault provider for the Secrets module https://github.com/OrchardCMS/OrchardCore/pull/7891, again for read/write). Both of those might connection to a different KeyVault - in the case of secrets, it might be a different vault for each tenant (as an option).

Azure have called it `Azure.Extensions.AspNetCore.Configuration.Secrets` - which is not helpful ;)

Would `OrchardCore.Configuration.KeyVault` be a good name?
or
`OrchardCore.Configuration.Azure` (I prefer KeyVault as it seems explicit and more obvious, but naming is hard)

jtkech:
@deanmarcussen okay i will take a look on it this night

jtkech:
@deanmarcussen 

Yes i agree, didn't see it was under the `OC.Modules` folder, need to be under the `OrchardCore` folder.

Then yes, good to have custom config / settings sources for all tenants and tenant specific, as we did from blob and database, but yes in a separate PR. So we already have `OrchardCore_Shells_Azure`, maybe rename it to `OrchardCore_Shells_Azure_Blob` so that we will be able to use `OrchardCore_Shells_Azure_KeyVault`. But that's okay to keep `OrchardCore_Shells_Azure` and then use `OrchardCore_Shells_KeyVault` ;)

Meanwhile for this PR, idem here i would suggest `OrchardCore.Configuration.Azure.KeyVault`, but maybe too long, in our solution there is no project with more than 3 terms, **so i also suggest `OrchardCore.Configuration.KeyVault`**.

agriffard:
Please merge dev and solve appsettings.json conflict.

deanmarcussen:
@JoshLefebvre sounds like @jtkech and I agree on the name, and to move it out of modules.

> I don't think it should be using the OrchardCore appsettings section

I withdraw my comment here, we already hosted the shells settings under `OrchardCore` so for consistency lets keep that the same. Even if it is only host only, not tenant based.

So just to move the folder, and rename thank you

jtkech:
I agree

JoshLefebvre:
Thanks @jtkech and @deanmarcussen. Yes agreed, makes much more sense in the OrchardCore module now that you explain it. I've also updated the name to OrchardCore.Configuration.KeyVault as suggested :)

agriffard:
Build retry

jtkech (approved):

jtkech:
LGTM

agriffard:
Thank you. 163rd project in the solution.

PR #6423, opened 2020-06-14
Display more media in library

### LINKED ISSUE #6418: Display more media in Media library
We should improve the layout of the media library (admin and picker) in order to display more than just 5 media per page. 

Hide the Assets title, hide the breadcrumb, move grid buttons to top bar.

agriffard:
Fixes #6418


src/OrchardCore.Modules/OrchardCore.Media/wwwroot/Scripts/media.js
```diff
@@ -162,7 +162,7 @@ function initializeMediaApplication(displayMediaApplication, mediaApplicationUrl
             return result;
           },
           thumbSize: function thumbSize() {
-            return this.smallThumbs ? 160 : 240;
+            return this.smallThumbs ? 110 : 195;
```
deanmarcussen, line 165: You can't change this size here I think because it is not a supported size for the media resizing. So will provide unresized thumbnails

agriffard: ok, then:
`return this.smallThumbs ? 100 : 240;`

src/OrchardCore.Modules/OrchardCore.Media/wwwroot/Scripts/media.js
```diff
@@ -936,6 +936,165 @@ Vue.component('sortIndicator', {
     }
   }
 });
+// <upload> component
+Vue.component('upload', {
+  template: '\
```
deanmarcussen, line 941: @agriffard I might be missing something, but it seems like the `.media.min.js` file didn't get a rebuild here ?

PR #6454, opened 2020-06-18
CodeMirror improvements

### LINKED ISSUE #6452: CodeMirror css
I noticed that the css (scss) of codemirror in TheAdmin theme is outdated.

We should copy the latest codemirror css file into it. Or figure out a way to use the css from the node_modules folder

agriffard:
Fixes #6452

Also add new features:
- Fullscreen support (used in Templates Create/Edit)
- Autoclose tags (in HtmlBody part and field and text with codemirror editor)
- Autocomplete in Sql (Query Create/Edit)
- LineWrapping (ex: Templates) 
- StyleActiveLine


jptissot:
I will test this. 

Is there a way to reduce the amount of js includes ? or would that mean getting rid of the cdn's ?

jptissot:
@agriffard looks good. I tested it with a type that had a Markdown field and a TextField (codemirror) and the styles looked a bit weird. I think it's because simplemde uses CodeMirror and they override the styles. Maybe we should wrap the simpleMDE styles into a parent .simplemde-codemirror class and use sass to generate the wrapper. This way, the overrides would only be applied to the simplemde codemirror and not all instances on the same page.

simplemde.scss
```
.simplemde-editor {
  .CodeMirror {
    ...
  }
}
.....
```

```
<div class="form-group simplemde-editor">
    <label asp-for="Markdown">@Model.TypePartDefinition.DisplayName()</label>
    <span class="hint">@T["The markdown text of the content item."]</span>
    <textarea asp-for="Markdown" rows="5" class="form-control content-preview-text"></textarea>
</div>
```

agriffard:
ok, @jptissot : `.simplemde-editor` added around codemirror in simplemde.scss.

src/OrchardCore.Modules/OrchardCore.Markdown/Views/MarkdownField.Edit.cshtml
```diff
@@ -8,7 +8,7 @@
     var culture = await Orchard.GetContentCultureAsync(Model.Field.ContentItem);
 }
 
-<div class="form-group">
+<div class="form-group simplemde-editor">
```
jptissot, line 11: I think you have to do the same for the MarkdownBodyPart

agriffard: ok, done.

PR #6488, opened 2020-06-23
Add menu display text (differentiator) classes to menu shapes

deanmarcussen:
When using multiple menus on a site it is possible to do a lot of the styling with css, which can avoid having many, many templates.

This adds the menu differentiator (which is calculated from the display text) as an additional class on the menu.

So out of the box you get `class="menu menu-main-menu"`

I already did this to taxonomies when I did the term shapes.


agriffard:
Can you please fix the conflict?

deanmarcussen:
Conflicts resolved

PR #6494, opened 2020-06-23
add SaveTokens support to all authentication providers

MichaelPetrinolis:
adds option to save tokens on all authentication providers
Also fixes twitter admin menu


sebastienros:
Is it related to the recent PR that implemented the ASP.NET idendity provider to store tokens? Such that now these middleware will store the token using this provider?

MichaelPetrinolis:
Exactly.

Issue #6446, opened 2020-06-17
Implement Cards for admin placement


deanmarcussen:
We should carry this recent improvement to O1 over to OC
https://github.com/OrchardCMS/Orchard/pull/8310

I think we can make a Column modifier in here work as well

hishamco:
This is cool, but I think tabs can organize the fields too, unless the fields per tab are too much

sebastienros:
Are there any other things we mentioned last time we talked about it? Like maybe other properties that would make sense to be able to set on these?

deanmarcussen:
I think I can make a columns property useful.

There were concerns about columns not being useful because of height mismatches between cards but I think I can make it work with flex

deanmarcussen:
It needs a modifier for a position for each tab/card as well.

ns8482e:
@sebastienros @deanmarcussen I have made some improvements on ContentCard in PR #6298. Preferably we could enhance that PR and use same shape by creating a method that Harvest ContentCard from placement similar to tab harvest method

PR #6509, opened 2020-06-25
Tabs, Cards, and Columns for the Admin

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6446

Also includes classes that we can apply to the `widget-editor` for related issue https://github.com/OrchardCMS/OrchardCore/issues/6501

Demo was at the meeting, so not a long description here.

Some examples of placement however

Modifier for tabs remains `#`, but now supports `;before` as a position modifier for the tab grouping
Modifier for Cards is `%`, also supports `;` as a position modifier
Modifier for Columns is `|`, supports `;`, and `_` as a column modifier, examples include `_9` and will be applied as `col-md-9` automatically, or `_lg-9` which will be applied as `col-lg-9`

`_9` should be sufficient for most, as it will by default break at `md` 

`;` must be immediately after, whereas `:` is the shape placement, and applies anywhere in the placement string.
 
```
  "HtmlBodyPart_Edit": [
    {
      "place": "Parts:3%Primary",
      "contentType": [
        "SimpleColumn"
      ]
    },
    {
      "place": "Parts:3%Primary|Content_9",
      "contentType": [
        "CardColumn"
      ]
    },
    {
      "place": "Parts:0#Right;3", 
      "contentType":[
        "TabLeftMediaRightBlock"
      ]
    },
    {
      "place": "Parts:0%Second Aside;5",
      "contentType":[
        "CardFirstAsideMediaSecondAside"
      ]
    },
    {
      "place": "Parts:0|Second Aside;5",
      "contentType":[
        "ColumnFirstAsideMediaSecondAside"
      ]
    },
    {
      "place": "Parts:0|Aside_9;0",
      "contentType":[
        "ColumnAsideMediaBlock"
      ]
    }
```


src/OrchardCore.Themes/TheAdmin/Assets/scss/TheAdmin.scss
```diff
@@ -1139,28 +1139,46 @@ ul.pager li a {
 .related {
     @extend .float-right;
 }
-// Responsive sizes
-// Creates mixins to target all breakpoints
```
deanmarcussen, line 1143: These mixins have been removed, as they are now in bootstrap, and the bootstrap versions are better. These where overriding the bootstrap ones, and making auto column arranging fail, i.e. 
```
<div class="col"></div>
<div class="col"></div>

deanmarcussen:
@ns8482e I thought about whether or not the `ColumnCard` could be a part of this, and ended up thinking that they are probably two similar but different things.

The `ColumnCard` focuses on moveable content items, (that are outside each other), and the focus here is on unmovable parts of content items, was my eventual thinking.

However it does work inside of a `ColumnCard` so these can still be applied quite successfully to items that are inside a `Bag` (ie. inside a `ColumnCard`)

I removed the need for the call to `HarvestAndSortTabs` and did it as part of the shape

agriffard:
Can you please add some documentation about placement when you have the time?

Issue #6517, opened 2020-06-26
Dynamic cache dependency tag helper


deanmarcussen:
There is a liquid tag to add dependencies to a cache block, but no corresponding Razor one.

I needed one today, so have implemented it

hishamco:
I think there's a built-in one in ASP.NET or you need to add something special?

jtkech:
@hishamco

We have a custom `DynamicCacheTagHelper` using cache contexts taylored for OrchadCore, look at the code it uses an `IDynamicCacheService` using an `ICacheContextManager` and an `IDynamicCache` using an `IDistributedCache` and so on.

@deanmarcussen

Just remembered that `DynamicCacheTagHelper` has a `dependencies` attribute and can be nested, isn't it sufficient?

deanmarcussen:
@hishamco ours is better ;)

@jtkech yes, I tried to see if nesting was relevant here, but it creates different cache blocks (cache-ids), so the inner nested cache block gets cached by the parent cache-id, and I need to be able to set the cache dependencies, after the cache block has started, rather than nesting another block.

So it needed something that could access the `ICacheScopeManager` and add dependencies to the current cache context. 

I maybe missing something, but I just implemented what we have for Liquid, and seemed to be missing for Razor

hishamco:
Thanks for clarification Jean & Dean .. go for it

jtkech:
@deanmarcussen okay

PR #6518, opened 2020-06-26
Add cache-dependency tag helper

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6517

Corresponding tag helper to work with dynamic cache entries to add dependencies after the cache block has started
i.e.

```
<cache cache-id="foo">
   var query = myquery;
   foreach(var item in query)
  {
      <cache-dependency dependency="contentitemid:@item.ContentItemId"/>
  }
<cache>
```

Works fine, but did discover an issue with the dynamic cache tag helper, where by default it conflicts with the ASP.NET Core Tag Helper, and both are activated, meaning the inner content is built by the ASP.NET Core tag helper, stored in it's memory cache, and then passed of to the dynamic cache tag helper, which then stores it, in it's cache.

This is because they are both targeting the `cache` element, but it means the `cache-dependency` doesn't have a scope available in the scope manager, as it is being rendered by the ASP.NET Core one.

Removing the ASP.Net Core one fixes the issue
`@removeTagHelper Microsoft.AspNetCore.Mvc.TagHelpers.CacheTagHelper, Microsoft.AspNetCore.Mvc.TagHelpers`

@jtkech wondered if you had any ideas on targeting the dynamic cache tag helper better, to block to ASP.NET one out of the box, or if we just update the docs?

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.DynamicCache/TagHelpers/CacheDependencyTagHelper.cs
+++ b/src/OrchardCore.Modules/OrchardCore.DynamicCache/TagHelpers/CacheDependencyTagHelper.cs
@@ -1,0 +1,50 @@
+using System;
+using Microsoft.AspNetCore.Razor.TagHelpers;
+using OrchardCore.Environment.Cache;
+
+namespace OrchardCore.DynamicCache.TagHelpers
+{
+    [HtmlTargetElement("cache-dependency", Attributes = DependencyAttributeName)]
+    public class CacheDependencyTagHelper : TagHelper
+    {
+        private const string DependencyAttributeName = "dependency";
+
+        /// <summary>
+        /// Gets or sets a <see cref="string" /> with the dependency to invalidate the cache with.
+        /// </summary>
+        [HtmlAttributeName(DependencyAttributeName)]
+        public string Dependency { get; set; }
+
+        private readonly ICacheScopeManager _cacheScopeManager;
+
+        public CacheDependencyTagHelper(
+            ICacheScopeManager cacheScopeManager)
+
+        {
+            _cacheScopeManager = cacheScopeManager;
+        }
+
+
+        /// <inheritdoc />
+        public override void Process(TagHelperContext context, TagHelperOutput output)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (output == null)
+            {
+                throw new ArgumentNullException(nameof(output));
+            }
+
+            if (!String.IsNullOrEmpty(Dependency))
+            {
+                _cacheScopeManager.AddDependencies(Dependency);
+            }
+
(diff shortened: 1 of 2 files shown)


jtkech:
@deanmarcussen yes, just tried, you're right

Yes because the mvc one is resolved just by using its `<cache>` tag name so **maybe we should have used a different tag name** as `<dynamic-cache>`.

Just tried the following that works, so, if don't want to change the tag name, **looks like the best solution to prevent the mvc one from being executed**, and worth to add it to the documentation as it seems good for perf / allocation

    @removeTagHelper Microsoft.AspNetCore.Mvc.TagHelpers.CacheTagHelper, Microsoft.AspNetCore.Mvc.TagHelpers

Then in our `DynamicCacheTagHelper` in place of just using

            // Clear the contents of the "cache" element since we don't want to render it.
            output.SuppressOutput();

Maybe good to use

            // Clear the contents of the "cache" element since we don't want to render it.
            if (output.IsContentModified)
            {
                output.SuppressOutput();
            }

src/OrchardCore.Modules/OrchardCore.DynamicCache/TagHelpers/CacheDependencyTagHelper.cs
```diff
@@ -0,0 +1,50 @@
...
+    [HtmlTargetElement("cache-dependency", Attributes = DependencyAttributeName)]
+    public class CacheDependencyTagHelper : TagHelper
+    {
+        private const string DependencyAttributeName = "dependency";
+
+        /// <summary>
+        /// Gets or sets a <see cref="string" /> with the dependency to invalidate the cache with.
+        /// </summary>
+        [HtmlAttributeName(DependencyAttributeName)]
+        public string Dependency { get; set; }
```
jtkech, line 16: Why not `Dependencies` allowing to add one or multiple dependencies
By using a `char[] SplitChars = new[] { ',', ' ' }` as in `DynamicCacheTagHelper`

deanmarcussen: @jtkech yes, I thought about that as well, but didn't because that was not what the liquid filter was doing.
And the usage is likely to be similar, as part of a loop, (it was for me anyway) e.g.
```
<cache cache-id="foo">
   var query = myquery;
   foreach(var item in query)
  {
      <cache-dependency dependency="contentitemid:@item.ContentItemId"/>
  }
<cache>
```

But if you think it should be, then I am happy to change it?

Also wondered about making it a selfclosing tag, as it would never have content, but again didn't because it would force it to be used self closing, where many of our other ones like `script` are not forced self closing, even though they don't render content

jtkech: @deanmarcussen okay cool

I just wanted to make sure we didn't miss anything

PR #6528, opened 2020-06-28
Add openId  client request sample

hyzx86:


src/docs/reference/modules/OpenId/README.md
```diff
@@ -49,9 +49,31 @@ Available settings are:
 - Enable Authorization Endpoint.
 - Enable Logout Endpoint.
 - Enable User Info Endpoint.
-- Allow Password Flow: It requires that the Token Endpoint is enabled. More info at <https://tools.ietf.org/html/rfc6749#section-1.3.3>
+- Allow Password Flow: It requires that the Token Endpoint is enabled. More info at <https://tools.ietf.org/html/rfc6749#section-1.3.3> \
+   Rquest sample:
```
deanmarcussen, line 53:
```suggestion
   Request sample:
```

src/docs/reference/modules/OpenId/README.md
```diff
@@ -49,9 +49,31 @@ Available settings are:
...
+      "client_secret": "client_secret",
+      "username":"username",
+      "password":"password",
+      "scope": "openid profile roles"
+    }
+    ```
 - Allow Client Credentials Flow: It requires that the Token Endpoint is enabled. More info at <https://tools.ietf.org/html/rfc6749#section-1.3.4>
-- Allow Authorization Code Flow: It requires that the Authorization and Token Endpoints are enabled. More info at <http://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth>
+- Allow Authorization Code Flow: It requires that the Authorization and Token Endpoints are enabled. More info at <http://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth> \
+ Rquest sample:
```
deanmarcussen, line 67:
```suggestion
 Request sample:
```

src/docs/reference/modules/OpenId/README.md
```diff
@@ -49,9 +49,31 @@ Available settings are:
 - Enable Authorization Endpoint.
 - Enable Logout Endpoint.
 - Enable User Info Endpoint.
-- Allow Password Flow: It requires that the Token Endpoint is enabled. More info at <https://tools.ietf.org/html/rfc6749#section-1.3.3>
+- Allow Password Flow: It requires that the Token Endpoint is enabled. More info at <https://tools.ietf.org/html/rfc6749#section-1.3.3> \
+   Request sample:
+    ``` json 
+    //EndPoint="/connect/token" ,Method=POST,Content-Type=application/x-www-form-urlencoded 
+    {
```
kevinchalet, line 56: The token request MUST be formatted using `x-www-form-urlencoded` so it's kinda confusing to use JSON here (the token endpoint returns JSON but doesn't accept JSON inputs).

hyzx86: Yes, I understand what you mean, I should use a JavaScript script as an example

src/docs/reference/modules/OpenId/README.md
hyzx86: I have tested that it is feasible, a complete example: https://github.com/JiZhouSoft/Jizhousoft.OrchardCore.Sample/blob/cd1fd0c42893c6d0607f8642451d1a98649b9769/vue-ts-admin/src/api/users.ts#L10

src/docs/reference/modules/OpenId/README.md
hyzx86: @kevinchalet  I have updated the sample codes

src/docs/reference/modules/OpenId/README.md
```diff
@@ -49,9 +49,72 @@ Available settings are:
...
 - Enable Logout Endpoint.
 - Enable User Info Endpoint.
-- Allow Password Flow: It requires that the Token Endpoint is enabled. More info at <https://tools.ietf.org/html/rfc6749#section-1.3.3>
+- Allow Password Flow: It requires that the Token Endpoint is enabled. More info at <https://tools.ietf.org/html/rfc6749#section-1.3.3> 
+
+Typescript request example using axios:
+
+``` typescript 
+    
+  const data: any =
```
hishamco, line 58: Remove the extra line above

src/docs/reference/modules/OpenId/README.md
```diff
@@ -49,9 +49,72 @@ Available settings are:
...
+      formDatas.push(encodeURIComponent(it) + '=' + encodeURIComponent(data[it]))
+    }
+  } 
+  var res = axios.request({
+        url: ApiUrl + '/connect/token',
+        method: 'post',
+        data: formData.join('&'),
+        headers: { 'Content-Type': 'application/x-www-form-urlencoded' }
+  })
+  return res
```
hishamco, line 82: Add new line before `return`

src/docs/reference/modules/OpenId/README.md
```diff
@@ -49,9 +49,72 @@ Available settings are:
...
+```
+Complete example at <https://github.com/JiZhouSoft/Jizhousoft.OrchardCore.Sample/blob/cd1fd0c42893c6d0607f8642451d1a98649b9769/vue-ts-admin/src/api/users.ts#L10>
+
+
```
hishamco, line 86: Remove the extra line

src/docs/reference/modules/OpenId/README.md
```diff
@@ -49,9 +49,72 @@ Available settings are:
...
+      formData.push(encodeURIComponent(it) + '=' + encodeURIComponent(data[it]))
+    }
+  } 
+  var res = axios.request({
+        url: ApiUrl + '/connect/token',
+        method: 'post',
+        data: formData.join('&'),
+        headers: { 'Content-Type': 'application/x-www-form-urlencoded' }
+  })
+   return res
```
hishamco, line 115: Same here

hishamco:
@kevinchalet it is look good to you after fixing the format issues?

src/docs/reference/modules/OpenId/README.md
```diff
@@ -49,9 +49,72 @@ Available settings are:
...
+
+``` typescript 
+    
+  const data: any =
+  {
+        username: username,
+        password: password,
+        // must read from config
+        client_id: 'client_id',
+        client_secret: 'client_secret',
```
kevinchalet, line 64: I know axios can be used in both browser and server apps, but I think most people will try to use this snippet in a browser app. In this case, we shouldn't encourage them to use a `client_secret` here, just in case.

src/docs/reference/modules/OpenId/README.md
```diff
@@ -49,9 +49,72 @@ Available settings are:
...
+
+Typescript request example using axios :
+
+``` Typescript     
+
+  const data: any =
+    {
+      "grant_type": "code",
+      "client_id": "client_id",
+      "client_secret": "client_secret",
```
kevinchalet, line 98: Same remark.

src/docs/reference/modules/OpenId/README.md
```diff
@@ -49,9 +49,72 @@ Available settings are:
...
-- Allow Authorization Code Flow: It requires that the Authorization and Token Endpoints are enabled. More info at <http://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth>
+- Allow Authorization Code Flow: It requires that the Authorization and Token Endpoints are enabled. More info at <http://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth> 
+
+Typescript request example using axios :
+
+``` Typescript     
+
+  const data: any =
+    {
+      "grant_type": "code",
```
kevinchalet, line 96: It's not a valid `grant_type`: the standard value is `authorization_code` for the authorization code flow.

src/docs/reference/modules/OpenId/README.md
```diff
@@ -49,9 +49,72 @@ Available settings are:
...
+      "client_id": "client_id",
+      "client_secret": "client_secret",
+      "scope": "openid profile roles"
+    }
```
kevinchalet, line 100: The mandatory `code` property is missing. You should also add the `redirect_uri`, that is required when a `redirect_uri` parameter was sent as part of the authorization request (with OpenID Connect, it's a mandatory parameter).

src/docs/reference/modules/OpenId/README.md
hyzx86: Thank you. It's all fixed

src/docs/reference/modules/OpenId/README.md
hyzx86: I added a note message as a reminder

src/docs/reference/modules/OpenId/README.md
hyzx86: I have removed the second example code but I organize the request information according to the prompt message on the interface
image

agriffard:
@kevinchalet Can you please confirm if it is ok to be merged?

src/docs/reference/modules/OpenId/README.md
```diff
@@ -49,7 +49,45 @@ Available settings are:
...
+        headers: { 'Content-Type': 'application/x-www-form-urlencoded' }
+  })
+  
+  return res
+```
+Complete example at <https://github.com/JiZhouSoft/Jizhousoft.OrchardCore.Sample/blob/cd1fd0c42893c6d0607f8642451d1a98649b9769/vue-ts-admin/src/api/users.ts#L10>
+
+!!! note
+    For `client_secret` and `client_id` as confidential information, do not send it directly through the browser application. The sample code here is only used as a demonstration to send an authentication request through the client.
+Generally speaking, we should rewrite an `action` which only accept the username and password parameters, and then returns the verification result.
```
kevinchalet, line 89: Well, I don't think we should recommend that. In most cases, the best option is to use an interactive flow like code and declare the client as a public application.

kevinchalet (changes requested):

src/docs/reference/modules/OpenId/README.md
hyzx86: Sorry, I don't quite understand. Could you give me some more hints? About `declaring the client as a public application`

Piedone: @kevinchalet could you follow up here, please?

Piedone:

src/docs/reference/modules/OpenId/README.md
kevinchalet: I'm still super convinced having JS/TS code samples is the right approach. If links to the OAuth 2.0/OIDC specifications are not enough, maybe HTTP request samples would be better?

hyzx86: I also thought about not using code, but I feel it will complicate the login process. Now I use oidc-client, which is very easy to use.

Piedone: So, what should we do here? Let's get to the finish line with this PR.

kevinchalet: I'd recommend closing it: we shouldn't encourage users to implement the protocol themselves by adding low-level JS/TS snippets. They are excellent client libraries out there 😃

hyzx86: Maybe we should create a sample SAP project with front and back end separation
just like  `ts-oidc-client`

Piedone: Let's do that, then.

Piedone:
Closing as per Kévin's remarks: https://github.com/OrchardCMS/OrchardCore/pull/6528#discussion_r1523643888 So, thank you for your contribution, but we'd rather not add this to the docs.

PR #6530, opened 2020-06-29
Tab, card, and column placement docs

deanmarcussen:
Adds docs for creating placement rules for tabs, cards, and columns


Rushabhmaster:
When this feature is going to be in the release?

jptissot:
@Rushabhmaster This is already available on the Cloudsmith dev feed. This will be released to nuget with the next version. I assume 1.0 at this point.

PR #6532, opened 2020-06-29
Use ShapeFactory as binding for Content Zone

deanmarcussen:
Uses the `ShapeFactory` binding added in https://github.com/OrchardCMS/OrchardCore/pull/6514 for the `ContentZone` rather than resolving from the service provider in the `DisplayContext`


deanmarcussen:
It's the convention for code shapes here @hishamco, it is case insensitive so does support both, but often shapes are created with `New.Shape(Property: value)` so get properties get bound with that naming style.

hishamco:
> It's the convention for code shapes here

I expect that, but it breaks a rule for method params naming in general ;)

Issue #6534, opened 2020-06-29
Menu differentiator not working


deanmarcussen:
The menu differentiator has stopped working since https://github.com/OrchardCMS/OrchardCore/pull/6406

I see the problem, when items are added to the menu, the incorrect property for the new differentiator is being used.

Will pr shortly, it should be a quick fix

PR #6535, opened 2020-06-29
Fix menu and taxonomy differentiator

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6534

@jtkech should be this way I think ?


jtkech:
@deanmarcussen yes thanks for fixing my fixes ;)

PR #6536, opened 2020-06-29
Improve GraphQL Query Schema validation and logging

hyzx86:


src/OrchardCore.Modules/OrchardCore.OpenId/Configuration/OpenIdClientConfiguration.cs
```diff
@@ -6,9 +6,11 @@
 using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Authentication.OpenIdConnect;
 using Microsoft.AspNetCore.DataProtection;
+using Microsoft.AspNetCore.Mvc.Localization;
```
hishamco, line 9: I don't we need the added using?!!

src/OrchardCore.Modules/OrchardCore.OpenId/Configuration/OpenIdClientConfiguration.cs
```diff
@@ -100,10 +100,14 @@ public void Configure(string name, OpenIdConnectOptions options)
...
-            {
-                _logger.LogWarning("The OpenID Connect module is not correctly configured.");
+            var errorResult = (await _clientService.ValidateSettingsAsync(settings)).Where(r => r != ValidationResult.Success);
 
+            if (errorResult.Any())
+            {
+                foreach (var item in errorResult)
+                {
+                    _logger.LogWarning(item.ErrorMessage);
+                }
```
hishamco, line 110: Just add a line after and it will looks good to me

hyzx86: sorry，“ add a line after ”  after  where ? 
I don't quite understand your mean.

hishamco: Before the `return`, there's always should be line there in the code base

src/OrchardCore.Modules/OrchardCore.OpenId/Configuration/OpenIdClientConfiguration.cs
```diff
@@ -100,10 +100,14 @@ public void Configure(string name, OpenIdConnectOptions options)
         private async Task<OpenIdClientSettings> GetClientSettingsAsync()
         {
             var settings = await _clientService.GetSettingsAsync();
-            if ((await _clientService.ValidateSettingsAsync(settings)).Any(result => result != ValidationResult.Success))
-            {
-                _logger.LogWarning("The OpenID Connect module is not correctly configured.");
```
deanmarcussen, line 105: Sorry guys, this line shouldn't go, as now you no longer know where the error is coming from.

If you want to log the validation errors as well, they should be concated together here

Here's a quick example from elsewhere
```
                            if (_logger.IsEnabled(LogLevel.Warning))
                            {
                                _logger.LogWarning("The OpenId Connect module is not correctly configure : '{Errors}'", string.Join(", ", result.Errors));
                            }
```

hishamco: @deanmarcussen, hmm I didn't see much changes from the previous code, except he logs the inner error message instead of "The OpenID Connect module is not correctly configured."

deanmarcussen: @hishamco multiple inner error message**S** and without any context of where those error messages are from

hyzx86: Hi ， I have commit the  new codes , please help me verify.

src/OrchardCore.Modules/OrchardCore.OpenId/Configuration/OpenIdClientConfiguration.cs
```diff
@@ -100,9 +102,11 @@ public void Configure(string name, OpenIdConnectOptions options)
         private async Task<OpenIdClientSettings> GetClientSettingsAsync()
         {
             var settings = await _clientService.GetSettingsAsync();
-            if ((await _clientService.ValidateSettingsAsync(settings)).Any(result => result != ValidationResult.Success))
+            var errorResult = (await _clientService.ValidateSettingsAsync(settings)).Where(r => r != ValidationResult.Success);
```
deanmarcussen, line 105:
```suggestion
            var errors = (await _clientService.ValidateSettingsAsync(settings)).Where(r => r != ValidationResult.Success);
```

src/OrchardCore.Modules/OrchardCore.OpenId/Configuration/OpenIdClientConfiguration.cs
```diff
@@ -100,9 +102,11 @@ public void Configure(string name, OpenIdConnectOptions options)
...
+
+            if (errorResult.Any())
```
deanmarcussen, line 107:
```suggestion
            if (errors.Any() && _logger.IsEnabled(LogLevel.Warning))
```

src/OrchardCore.Modules/OrchardCore.OpenId/Configuration/OpenIdClientConfiguration.cs
```diff
@@ -100,9 +102,11 @@ public void Configure(string name, OpenIdConnectOptions options)
...
             {
-                _logger.LogWarning("The OpenID Connect module is not correctly configured.");
+                _logger.LogError($"The OpenId Connect module is not correctly configure : { string.Join(", ", errorResult.Select(r => r.ErrorMessage))}" );
```
deanmarcussen, line 109:
```suggestion
                _logger.LogWarning("The OpenID Connect module is not correctly configured : {Errors}", string.Join(", ", errors.Select(e => e.ErrorMessage)));
```

src/OrchardCore.Modules/OrchardCore.OpenId/Configuration/OpenIdClientConfiguration.cs
hyzx86: if the `LogLevel.Warning` is not enabled , It will return an error setting information

src/OrchardCore.Modules/OrchardCore.OpenId/Configuration/OpenIdClientConfiguration.cs
hyzx86: If you write like this, will the result of the following calculation be filled into the {Errors} expression?

src/OrchardCore.Modules/OrchardCore.Lucene/GraphQL/LuceneQueryFieldTypeProvider.cs
```diff
@@ -39,23 +42,35 @@ public async Task<IChangeToken> BuildAsync(ISchema schema)
...
-
-                if (query.ReturnContentItems &&
-                    type.StartsWith("ContentItem/", StringComparison.OrdinalIgnoreCase))
+                try
                 {
-                    var contentType = type.Remove(0, 12);
-                    var queryField = BuildContentTypeFieldType(schema, contentType, query);
-                    if (queryField != null)
+                    var querySchema = JObject.Parse(query.Schema);
+                    if (!querySchema.ContainsKey("type"))
```
hyzx86, line 48: We have to add a sample Schema on the page ,but I don't know the right schema for here.

hyzx86: https://github.com/OrchardCMS/OrchardCore/issues/6554

src/OrchardCore.Modules/OrchardCore.Queries/Controllers/AdminController.cs
```diff
@@ -155,13 +156,17 @@ public async Task<IActionResult> CreatePost(QueriesCreateViewModel model)
             {
                 return NotFound();
             }
-
+            if (!string.IsNullOrWhiteSpace(query.Schema) && !CheckSchema(query.Schema))
+            {
+                _notifier.Error(H["Invalid schema JSON supplied."]);
+                model.Editor = await _displayManager.BuildEditorAsync(query, updater: _updateModelAccessor.ModelUpdater, isNew: true);
```
deanmarcussen, line 162: This should add just add a `ModelState` error, and see what the result is when it is handled later in the code


```
            // If we got this far, something failed, redisplay form --> All errors should be handled by this section
            model.Editor = editor;
```

src/OrchardCore.Modules/OrchardCore.Lucene/GraphQL/LuceneQueryFieldTypeProvider.cs
deanmarcussen: Please leave that for a seperate pr

src/OrchardCore.Modules/OrchardCore.Queries/Controllers/AdminController.cs
hyzx86: Hi @deanmarcussen  , thanks for your help ,  `ModelState` added

hyzx86: image

src/OrchardCore.Modules/OrchardCore.Queries/Controllers/AdminController.cs
```diff
@@ -155,13 +160,15 @@ public async Task<IActionResult> CreatePost(QueriesCreateViewModel model)
             {
                 return NotFound();
             }
-
+            if (!string.IsNullOrWhiteSpace(query.Schema) && !CheckSchema(query.Schema))
+            {
+                ModelState.AddModelError("Schema", S["Invalid schema JSON supplied."]);
```
deanmarcussen, line 165: Ok great, this works better.

Sorry, what I should have said though, is to implement `IValidatableObject ` on the `QueriesCreateViewModel`.

Apply the test for valid JSON on that View Model by implementing `IValidatableObject`.

Here is an example from where it is used elsewhere in the code https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.Users/ViewModels/LinkExternalLoginViewModel.cs

I think you will need to apply the same to the edit controller `QueriesEditViewModel` View Model

Then in the `Queries.Fields.Edit.cshtml` View you should be able to apply 
`<span asp-validation-for="Schema"></span>` (like is done for `Name` in the same file) and you will get a validation error on the schema field itself

hyzx86: But the value of Schema comes from query variable, not model, model.Schema is always empty

hyzx86: image
And the query object cannot be filled if it is verified during creation

deanmarcussen: Ok, I just added a string extension for testing json here https://github.com/OrchardCMS/OrchardCore/pull/6614

But I have had a look at the Queries `AdminController` and understand what is driving that viewmodel now.

I was wrong to suggest `IValidateableObject`.

The query `Schema`, is actually managed and updated by the `QueryDisplayDriver.cs`

It already performs some checks for validity so the test you are doing for valid Json should move into the `QueryDisplayDriver.cs` `UpdateAsync()` method, along with the other checks for valid Names etc.

PR #6542, opened 2020-06-30
Export Site Settings

### LINKED ISSUE #5325: There is no way to export custom site settings like Login settings

agriffard:
Fixes #5325, #5558

Very basic implementation.
Still need to find a good way to add the site settings (currently with hardcoded strings in the Driver).

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Settings/Deployment/AllSiteSettingsDeploymentSource.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Settings/Deployment/AllSiteSettingsDeploymentSource.cs
@@ -1,0 +1,43 @@
+using System.Threading.Tasks;
+using Newtonsoft.Json.Linq;
+using OrchardCore.Deployment;
+using OrchardCore.Entities;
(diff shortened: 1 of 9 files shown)


agriffard:
image

hishamco:
If each module can provide it's own via `IDeployementPlan`, why we didn't introduce an aggregator that scans all the registered settings and list the as shown above

agriffard:
Replaced by #6570

PR #6546, opened 2020-06-30
fix simplemde css issue

### LINKED ISSUE #6529: Multiple markdown editors in a content type stop working
I have several markdown fields with wyswyg editors. They used to work and now each of these fields stop working. They become unresponsive. Also an "x" appears if you try multiple clicks on the non responsive editor.

standard-editor

jptissot:
fix #6529


jptissot:
I tried to add 
`<style name="simplemdecss" asp-src="~/OrchardCore.Markdown/Styles/simplemde.min.css" debug-src="~/OrchardCore.Markdown/Styles/simplemde.css" depends-on="simplemde"></style>`

but that did not work, I had to add a seperate style tag.

agriffard:
That is because you cannot make a dependency between a style and a script.

jptissot:
But there is a SimpleMde css declaration no ?

Issue #6199, opened 2020-05-16
[Feature] Implement wordpress shortcodes


sebastienros:
These code snippets allow content editors to inject specially constructed html like media, twitter embeds, youtube videos, ... 

https://wordpress.com/support/shortcodes/

Right now a content editor would need to use liquid tags. This can also be done with pre-existing shortcodes like `[media]` (https://github.com/OrchardCMS/OrchardCore/pull/6016).

The suggestion is 
- To enhance the `IShortCodeService` with a service that lists existing  shortcodes and their usage. This would allow features that allow for them to provide a window with the list of available ones.
- Create an admin view where "custom" shortcodes can be created or modified. This would allow to create custom ones, using Liquid.
- Create a Shortcode widget that allows to reuse these shortcodes in Flow Part or layers.

**Orchard 1 Tokens**

This could be seen as very similar to token in O1.

One advantage is that the shortcodes don't have the same scope, for instance autoroute will still use Liquid. Shortcodes are targetting content editors.

Another difference is that they should be editable, such that complex ones can be created easily by an admin. This means that if a shortcode is missing a useful argument, or needs to be lower cased, a custom one can be added, or an existing one edited. 

They can be imported/exported.

TODO:
Evaluate all issues that tokens have in O1 and see how shortcodes reflect with it

sebastienros:
From https://developer.wordpress.org/reference/functions/add_shortcode/#more-information

> The shortcode callback will be passed three arguments: the shortcode attributes, the shortcode content (if any), and the name of the shortcode.
Shortcode attribute names are always converted to lowercase before they are passed into the handler function. Values are untouched.

A shortcode parser should allow for inner shortcodes, building a tree that is processed in post-order (https://en.wikipedia.org/wiki/Tree_traversal#Post-order_(LRN)).

Escaping: https://diveinwp.com/escape-shortcodes-in-wordpress/

Sample UI:
![image](https://user-images.githubusercontent.com/1165805/82126167-c3bf4180-975f-11ea-8194-75d0ceaee2f3.png)

A custom implementation could use Razor templates from a specific folder. Could also include liquid files. Or be based on shapes and a special prefix. But that might allow for security issues, and it would be hard to define metadata (description).

jptissot:
Would it be possible to have access to a context object / pass multiple arguments to shortcodes. I was thinking of using shortcodes to add inline translation support to a ContentItem. 

[...]

Issue #6183, opened 2020-05-13
Can't overwrite media


devlife:
I have a client who wants to be able to overwrite media. The tenant is configured with Azure Storage. If he deletes an image, Image.png and then uploads a different image with the same name then we still see the original image thumbnail.

I've tried purging media cache as well but that didn't have any effect.

deanmarcussen:
Have you tried clearing your browser cache?

ImageSharp uses a default browser cache of 14 days (from memory), so even when evicted from the Media Cache (which would happen automatically if the item is deleted from the media library), the browser will still cache the image for a period of time.

ImageSharp also evaluates based on last modified dates, but when using blob storage this translates into the date it was uploaded into blob storage (when not using blob storage, this can be an issue, if you are just renaming files to test with, as the last modified date doesn't change during a rename)

We also have cache busting support for images (you can apply this to the front end)

But we don't yet have that the admin area. 

Here's the original issue for this https://github.com/OrchardCMS/OrchardCore/issues/3925

I did open a pr but it never got merged, so I closed it.

Perhaps it should be a setting for the media app?

Rather than on by default (which the pr did)

devlife:
No I haven't tried clearing browser cache. That def seems like a pain. Out of curiosity is there a reason not to reopen the PR?

deanmarcussen:
It's a pretty simple PR, probably just need to get the design requirements correct, though i know Sebastien was concerned about the performance implications. 
(Note they are memory cached after the first request)

Requirements request:

Do we want default media cache busting as a setting?

This would be then applied to the admin thumbnails, and the `asset_url` helpers

devlife:
Personally I think that would be a good addition assuming I'm not missing some other reason it hasn't been implemented yet. And the memory caching makes sense. If we could clear that specific cached item I think that would be fantastic.

deanmarcussen:
@devlife have you verified this was browser cache related?

devlife:
@deanmarcussen sorry I didn't realize you were waiting for me to verify. I will do so this week.

sebastienros:
Can we try to implement the cache with the FileInfo modified time? This doesn't require FileWatcher

deanmarcussen:
If it is possible to make it work without a file watcher we could then have a site setting to control this behaviour by default

deanmarcussen:
> Hmm, as i understand we would have to grab the file info, at least periodically, to see maybe first if the file still 

From your other comment @jtkech 

[...]

PR #6561, opened 2020-07-02
Shortcodes refactor

deanmarcussen:
Related to https://github.com/OrchardCMS/OrchardCore/issues/6199

- Implements `Shortcodes` parser by @sebastienros 
- Renames `ShortCode` to `Shortcodes` as I got it wrong :(
The rename includes files, folders, and csproj, so will create a new nuget package name. We did this a while ago with the `Github` to `GitHub` change, and it worked ok, so I think is ok to do it.
- Renames interfaces, and removes unecesary interfaces
- Implements the `[image]` shortcode using the new parser

Has one issue with the `MarkDig` processor which replaces `"` which and html encoded version, so arguments have to be used with a single quote, not a double quote. 
Need to look further into that issue.

Does not include
- The shape Shortcode I demoed a week ago, still deciding if that's useful
- Admin editable Shortcodes. That's for another pr

Wanted to get some progress going however, with the changes that are necesary, and was wanting to implement a `Shortcode` yesterday, but didn't want to because it will be so trivial with the new parser


src/OrchardCore.Modules/OrchardCore.Media/Shortcodes/ImageShortcodeProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+                {
+                    content = args.NamedOrDefault("src");
+                    if (string.IsNullOrEmpty(content))
+                    {
+                        // Do not handle the deprecated media shortcode in this edge case.
+                        return new ValueTask<string>("[image]");
+                    }
+                }
+
+                // TODO handle virtual path, i.e. ~/
```
deanmarcussen, line 43: @jtkech if you have a minute, I wanted to handle virtual paths here `~/` to be able to use this for static files as well. Can you point me in the right direction for how to get the virtual path, if running under tenant, or virtual folder? Thanks :)

src/OrchardCore.Modules/OrchardCore.Media/Shortcodes/ImageShortcodeProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+using Microsoft.AspNetCore.WebUtilities;
+using Shortcodes;
+using OrchardCore.Infrastructure.Html;
+
+namespace OrchardCore.Media.Shortcodes
+{
+    public class ImageShortcodeProvider : NamedShortcodeProvider
+    {
+        private readonly IHtmlSanitizerService _htmlSanitizerService;
+        private readonly IMediaFileStore _mediaFileStore;
```
jtkech, line 13: Looks like these 2 fields are unused

src/OrchardCore.Modules/OrchardCore.Media/Shortcodes/ImageShortcodeProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+            : base(
+            new Dictionary<string, ShortcodeDelegate>
+            {
+                ["image"] = EvaluateImageAsync(mediaFileStore, htmlSanitizerService),
+                ["media"] = EvaluateImageAsync(mediaFileStore, htmlSanitizerService)
+            }
+        )
+        {
+            _mediaFileStore = mediaFileStore;
+            _htmlSanitizerService = htmlSanitizerService;
```
jtkech, line 25: Idem

src/OrchardCore.Modules/OrchardCore.Media/Shortcodes/ImageShortcodeProvider.cs
```diff
@@ -0,0 +1,82 @@
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.WebUtilities;
+using Shortcodes;
+using OrchardCore.Infrastructure.Html;
```
jtkech, line 6: Sort usings

src/OrchardCore.Modules/OrchardCore.Media/Shortcodes/ImageShortcodeProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+            _mediaFileStore = mediaFileStore;
+            _htmlSanitizerService = htmlSanitizerService;
+        }
+
+        private static ShortcodeDelegate EvaluateImageAsync(IMediaFileStore mediaFileStore, IHtmlSanitizerService htmlSanitizerService)
+        {
+            return (args, content) =>
+            {
+                // Handle edge case of self closing shortcodes.
+                if (string.IsNullOrEmpty(content))
```
jtkech, line 33: Maybe use `String` framework type

src/OrchardCore.Modules/OrchardCore.Media/Shortcodes/ImageShortcodeProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+                {
+                    content = args.NamedOrDefault("src");
+                    if (string.IsNullOrEmpty(content))
```
jtkech, line 36: idem

src/OrchardCore.Modules/OrchardCore.Media/Shortcodes/ImageShortcodeProvider.cs
jtkech: In the media startup we build the `DefaultMediaFileStore` with a `mediaUrlBase` combined with the `originalPathBase` and used to init a `_requestBasePath` used by `MapPathToPublicUrl()` to take into account e.g. a virtual folder

So here i would say to just skip these 2 chars `~/` and then use `MapPathToPublicUrl()`, or just skip the `~` as the path will be normalized by `MapPathToPublicUrl()`


    if (content[0] == '~')
    {
        content = mediaFileStore.MapPathToPublicUrl(content.Substring(1));
    }
    else if (!content.StartsWith("//", StringComparison.Ordinal) && !content.StartsWith("http", StringComparison.OrdinalIgnoreCase))
    {
        content = mediaFileStore.MapPathToPublicUrl(content);
    }

deanmarcussen: @jtkech sorry, I didn't explain clearly enough.

I was hoping to support `~/` as a link to static files, that are not stored in the media file store, e.g. something like
`~/MyTheme/myembeddedbrandicon.jpg`, and also (possibly?) the files from the tenant static file provider.

src/OrchardCore.Modules/OrchardCore.Media/Shortcodes/ImageShortcodeProvider.cs
deanmarcussen: Done

src/OrchardCore.Modules/OrchardCore.Media/Shortcodes/ImageShortcodeProvider.cs
jtkech: @deanmarcussen oops sorry, do you mean in the context of e.g. a `StaticFileShortcodeProvider`? Or maybe a `ResourceUrlShortcodeProvider`. As i understand it would be a scoped service and used to generate an url e.g. from `~/MyTheme/brandicon.jpg` to `{virtualFolder}/{tenantPrefix}/MyTheme/brandicon.jpg`

First, we can use the request base path including an eventual virtual folder and the tenant url prefix, but because in a given request scope in the context of a given tenant, we can create a scope and render things in the context of another tenant, normally we would need to use the original base path (before it includes any tenant url prefix) and append to it the tenant url prefix related to the current shell scope (not the request scope). That's why i made an httpContext feature holding the original base path, but maybe it's too much here ;) and this is not what we are doing most of the time.

So here i would say to just do what we do in the related liquid filters, e.g. the `ResourceUrlFilter` that uses the request pathBase through an `_httpContextAccessor` to generate an url. Then, if you want to support versioning, what we do in the `AppendVersionFilter` by using an `IFileVersionProvider`. Finally, if i understand correctly, because here you just have to generate an url, you don't have to care about any satic file provider, e.g. the tenant static file provider, they will just work as when we generate an url through liquid filters.

deanmarcussen: awesome @jtkech I understand how that side of the system works now, and why we use the httpContent feature in the media file store, but yes I agree it is too much here. Cheers

This will be a little used feature, I suspect, but wanted to support it, so will do as per the liquid filters.

I'll hold of on the versioning feature for it, as Seb has some suggestions for refactoring that a little to avoid the file watchers, re https://github.com/OrchardCMS/OrchardCore/issues/6183

jtkech: @deanmarcussen Okay cool

Yes, i understand about file watchers, would need more time but i took a quick look in #6183 where seb said

> Can we try to implement the cache with the FileInfo modified time? This doesn't require FileWatcher

Hmm, as i understand we would have to grab the file info, at least periodically, to see maybe first if the file still exists and then if its last modified time has changed, a kind of a light file watcher. Or if it is fast enough (which is the case for embedded files) maybe grab this file info each time we generate an url, if so maybe we could just append something re-computed from this modified time (if it has changed) in place of doing a hash on the whole file. Hmm, maybe worth to do at least for embedded files so without having to compute a hash for all of them.

deanmarcussen: cool, I've implemented for the `~/` virtual path.

If you could take a look, I'll probably merge it after you approve, before moving onto other parts of the feature for other prs.

Also I refactored to use the `IShortcodeProcessor` interface directly, to improve perf on the old `[media]` tag.

src/OrchardCore.Modules/OrchardCore.Media/Shortcodes/ImageShortcodeProvider.cs
```diff
@@ -0,0 +1,107 @@
...
+using OrchardCore.ResourceManagement;
+using Shortcodes;
+
+namespace OrchardCore.Media.Shortcodes
+{
+    public class ImageShortcodeProvider : IShortcodeProvider
+    {
+        private static readonly ValueTask<string> Null = new ValueTask<string>((string)null);
+        private static readonly ValueTask<string> ImageShortcode = new ValueTask<string>("[image]");
+        private static HashSet<string> Shortcodes = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
```
jtkech, line 17:
```suggestion
        private static readonly HashSet<string> Shortcodes = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
        {
            "image",
            "media" // [media] is a deprecated shortcode, and can be removed in a future release.
        };
```

src/OrchardCore.Modules/OrchardCore.Shortcodes/Services/ShortcodeService.cs
```diff
@@ -0,0 +1,24 @@
...
+    public class ShortcodeService : IShortcodeService
+    {
+        private readonly ShortcodesProcessor _shortcodesProcessor;
+        private readonly IEnumerable<IShortcodeProvider> _shortcodeProviders;
+
+        public ShortcodeService(IEnumerable<IShortcodeProvider> shortcodeProviders)
+        {
+            _shortcodeProviders = shortcodeProviders;
+            _shortcodesProcessor = new ShortcodesProcessor(shortcodeProviders);
+
```
jtkech, line 16: Remove extra line

src/OrchardCore.Modules/OrchardCore.Media/Shortcodes/ImageShortcodeProvider.cs
jtkech: @deanmarcussen 

Okay just did a little review otherwise LGTM

src/OrchardCore.Modules/OrchardCore.Shortcodes/Services/ShortcodeService.cs
deanmarcussen: And I removed need to store the providers locally, as they go straight to the processor.

jtkech: Cool, didn't see it ;)

Issue #6567, opened 2020-07-02
AliasPartContentAliasProvider doesn't return a content item ID unless alias is prefixed with "alias:"


kalizec:
The parameters of the AliasPartContentAliasProvider seem to indicate that this provider can be provided with any alias, with or without "alias:" in front of it. To me AliasPartContentAliasProvider.cs lines 25-26 look like their place should be outside of the if statement checking whether the alias starts with "alias:" or not.

kalizec:
Not sure whether I did this correct. First time for me doing an issue/pr on OrchardCore (or GitHub for that matter).

deanmarcussen:
This is correct, as it is used by the `IOrchardHelper.GetContentItemByHandleAsync` method, and requires prefixing with an alias.

We renamed the extension helper method prior to rc2, from `GetContentItemByAliasAsync` to `GetContentItemByHandleAsync` to remove some of the confusion.

Now that RC2 has shipped out, in my opinion, it would be ok to remove the obsolete methods, and rename the underlying manager and providers to (for example) `AliasPartContentHandleProvider` to clear anymore confusion.

We could also potentially then have an extension for `GetContentItemByAliasAsync` which automatically applies the required `alias:`, and the same for `GetContentIemBySlugAsync`, e.g applies `slug:`

//cc @agriffard just some ideas see what you think too?

sebastienros:
> We could also potentially then have an extension for GetContentItemByAliasAsync which automatically applies the required alias:, and the same for GetContentIemBySlugAsync, e.g applies slug:

This

sebastienros:
Just to explain, it was made when I implemented dynamic cache, and we needed a way to identify dependencies on content items from different ways. So I made the notion of  "alias" (actually a handle) such that the cache doesn't rely on anything directly (alias, autoroute, ...) and content items can be logically linked by anything.

But I understand that it's bad to expose it as a way to load content items in helpers. So creating custom methods to load them is better. We can still keep using the "alias" methodology in the cache.

PR #6568, opened 2020-07-02
#6567 AliasPartContentAliasProvider should work without prefix

kalizec:
AliasPartContentAliasProvider should work for alias parameter regardless of whether it is prefixed with "alias:".

Moved contentItemId retrieval outside of statement.


deanmarcussen:
Thanks for this @kalizec 

The code is actually working as it is supposed to.

This is a provider that is ultimately called by the helper extension `IOrchardHelper.GetContentItemByHandleAsync` and the requirement when using the helper is to provide a handle prefix, i.e. `alias:` or `slug:` which are the two default oob providers.

I will close this and comment further on the issue you opened

Issue #6583, opened 2020-07-04
Admin menu item returns all content types instead of the selected one in admin view


osolvi:
Version `1.0.0-rc2-13700`

![image](https://user-images.githubusercontent.com/39261036/86507411-aa8e5500-bde0-11ea-91f8-f76127b9d133.png)

![image](https://user-images.githubusercontent.com/39261036/86507536-9434c900-bde1-11ea-813c-155e4e470a08.png)

deanmarcussen:
Oops, yes I can repro this, and know where the issue is.

PR #6589, opened 2020-07-05
Rework query filter to include list options model

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6583

Sorry my bad on this one, I needed to review it better after I refactored the original pr.

I've returned the design of the filters, to before my refactor, where the filter has access to the `ContentOptionsViewModel` allowing it to have values set by the `AdminController`, and (potentially mutate) those values, to alter the query.

This fixes the admin menus content types issue, and includes more flexibility for anyone writing extensible filters. (I couldn't quite decide writing this, whether it should have that functionality or not, but have realised it does need it.)

I've also renamed a couple of the new services, to better meet our standard naming conventions, using `updater`, and query service, rather than provider.

Well tested, and this should be the final refactor.

@jtkech if you have time can you look over the names used here as well, and offer any suggestions?


jtkech:
> @jtkech if you have time can you look over the names used here as well, and offer any suggestions?

After a quick look i saw names as `ContentsOrder` and `ContentsStatus` with the term `Contents` with a `s`, and other names with the term `Content` without a `s` including `ContentOptionsViewModel`, but not so important ;)

Otherwise LGTM

> I've also renamed a couple of the new services, to better meet our standard naming conventions, using updater, and query service, rather than provider.

LGTM

deanmarcussen:
> After a quick look i saw names as `ContentsOrder` and `ContentsStatus` with the term `Contents` with a `s`, and other names with the term `Content` without a `s` including `ContentOptionsViewModel`, but not so important ;)

Thanks @jtkech ... Naming is Hard.

I went with `Contents` because they are from the `Contents` module, not `ContentManagement`, and the original `ContentOptionsViewModel` has `ContentsStatus` inside it., but it already has the `...Options` pluralization.

PR #6609, opened 2020-07-08
Fix regression - tags shape missing in blog template

deanmarcussen:
Fixes a regression from https://github.com/OrchardCMS/OrchardCore/pull/6406 where we changed how the name of a shape was maintained.

@jtkech that change to handling shape names with display types, and putting the differentiator on the shape metadata has been really useful this past week. Cheers :)


jtkech:
@deanmarcussen

Cool happy to know it is useful ;)

Yes i know i may have missed some updates, i thought about backwards compatibility, maybe fallback to the differentiator if not found by the shape name, and then maybe to the shape type, but i think it'okay as it is.

PR #6610, opened 2020-07-09
Don't fail when creating a scope on a dependent shell that has been disposed.

### LINKED ISSUE #6605: System.ArgumentNullException: Can't resolve a scope on tenant
```
System.ArgumentNullException: Can't resolve a scope on tenant: Contoso (Parameter 'ServiceProvider')
   at OrchardCore.Environment.Shell.Scope.ShellScope..ctor(ShellContext shellContext)
   at OrchardCore.Environment.Shell.Builders.ShellContext.CreateScope()
   at OrchardCore.Environment.Shell.ShellHost.GetScopeAsync(ShellSettings settings)
   at OrchardCore.Modules.ModularTenantContainerMiddleware.Invoke(HttpContext httpContext)
   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware.Invoke(HttpContext context)
```

Steps to reproduce this exception:
  - Create a new OrchardCore site.
  - Enable the OpenID server feature.
  - Create a new public application the code flow enabled and with `client_id`: `postman` and `redirect_uri`: `https://oauth.pstmn.io/v1/callback`.
  - Create a new OrchardCore tenant named `Contoso` and configure it.
  - Create a new `demo_api` scope with the  `Contoso` tenant checked.
  - Enable the OpenID validation feature in the `Contoso` tenant and choose `Default` as the server tenant.
  - Add a test controller using API authentication:

```csharp
[Feature(OpenIdConstants.Features.Validation)]
public class MyController : Controller
{
    [Authorize(AuthenticationSchemes = "Api")]
    [HttpGet("~/test")]
    public IActionResult Test() => Content(User.GetUserIdentifier());
}
```
  - Open Postman and configure it to retrieve an access token using the authorization code flow:

image

  - Use the access token to query the demo controller in the `Contoso` tenant:

image

  - Go to the OpenID server configuration page in the main tenant. Save the changes.

  - Make a new HTTP call to the demo controller. If you get a non-error response, go back to the OpenID server configuration page and save the changes a second time and try again.

**Marking as P0 as this bug puts all the impacted tenants in a completely unusable state until they are manually restarted.**

@jtkech has anything changed recently that could explain this issue?

jtkech:
Fixes #6605 

    // We can't create a scope with a null 'ServiceProvider' meaning that the shell has been disposed. Normally, the
    // 'ShellHost' removes the shell from its collection as soon as it is released so that a new one will be created.
    // But this may happen when a shell releases itself its dependent shells and disposes those that are not in use.
    if (_placeHolder || ServiceProvider == null)
    {
        return null;
    }

As i remember at some point we were checking the `ServiceProvider` before creating a scope, but we removed this check and added a `throw new ArgumentNullException()` in the `ShellScope` constructor.

### DIFF
--- a/src/OrchardCore/OrchardCore.Abstractions/Shell/Builders/ShellContext.cs
+++ b/src/OrchardCore/OrchardCore.Abstractions/Shell/Builders/ShellContext.cs
@@ -48,7 +48,10 @@ namespace OrchardCore.Environment.Shell.Builders
 
         public ShellScope CreateScope()
         {
-            if (_placeHolder)
+            // We can't create a scope with a null 'ServiceProvider' meaning that the shell has been disposed. Normally, the
+            // 'ShellHost' removes the shell from its collection as soon as it is released so that a new one will be created.
+            // But this may happen when a shell releases itself its dependent shells and disposes those that are not in use.
+            if (_placeHolder || ServiceProvider == null)
             {
                 return null;
             }
--- a/src/OrchardCore/OrchardCore.Abstractions/Shell/Builders/ShellContext.cs
+++ b/src/OrchardCore/OrchardCore.Abstractions/Shell/Builders/ShellContext.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using OrchardCore.Environment.Shell.Builders.Models;
+using OrchardCore.Environment.Shell.Models;
 using OrchardCore.Environment.Shell.Scope;
 
 namespace OrchardCore.Environment.Shell.Builders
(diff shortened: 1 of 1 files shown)


src/OrchardCore/OrchardCore.Abstractions/Shell/Builders/ShellContext.cs
```diff
@@ -48,7 +48,10 @@ public PlaceHolder()
 
         public ShellScope CreateScope()
         {
-            if (_placeHolder)
+            // We can't create a scope with a null 'ServiceProvider' meaning that the shell has been disposed. Normally, the
+            // 'ShellHost' removes the shell from its collection as soon as it is released so that a new one will be created.
+            // But this may happen when a shell releases itself its dependent shells and disposes those that are not in use.
```
sebastienros, line 53: Please fix this line, do you mean:

```suggestion
            // But this may happen when a shell releases its dependent shells and disposes those that are not in use.
```

jtkech:
@sebastienros 

Don't merge it immediately, i think i remember why we removed this check, i think this is when we would want to create a scope on a disabled shell, which is another case with a null service provider, and with this check it would end up in an infinite loop. So i think that we also need to check if the shell is disabled, and if so let it throw an exception as before.

jtkech:
@sebastienros Okay, i think it is ready now

PR #6619, opened 2020-07-09
Upgrade YesSql

agriffard:
Version https://www.nuget.org/packages/YesSql.Core/2.0.0-beta-1514

Change methods calls to `CreateMapIndexTable`, ...:
https://github.com/sebastienros/yessql/blob/eedac41a8f631398f3f0a8b881dc68473757c6e7/src/YesSql.Abstractions/ISchemaBuilder.cs#L19

Do we pass `null` everywhere for the `collection` parameter?

Problem to replace the code here because the types don't exist anymore:
https://github.com/OrchardCMS/OrchardCore/blob/0d957fba1be08a197b087df724a1d09634108477/src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Migrations/OpenIdMigrations.cs#L65

Do we remove the Migration 2?


hishamco:
> Do we pass null everywhere for the collection parameter?

Why we did use optional parameter? or create another overload decorated with `ObsoleteAttribute`?

src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Migrations/OpenIdMigrations.cs
```diff
@@ -62,21 +70,24 @@ public int UpdateFrom1()
 
         public int UpdateFrom2()
         {
-            SchemaBuilder.DropReduceIndexTable("OpenIdApplicationByPostLogoutRedirectUriIndex");
-            SchemaBuilder.DropReduceIndexTable("OpenIdApplicationByRedirectUriIndex");
-            SchemaBuilder.DropReduceIndexTable("OpenIdApplicationByRoleNameIndex");
+            SchemaBuilder.DropReduceIndexTable("OpenIdApplicationByPostLogoutRedirectUriIndex", null);
```
deanmarcussen, line 73: @agriffard I would not drop this migration.

You can create a class in the migration with the name, and that should be sufficent 
```
internal class OpenIdApplicationByPostLogoutRedirectUriIndex {}
```

src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Migrations/OpenIdMigrations.cs
```diff
@@ -62,21 +70,24 @@ public int UpdateFrom1()
...
+            SchemaBuilder.DropReduceIndexTable("OpenIdApplicationByRedirectUriIndex", null);
+            SchemaBuilder.DropReduceIndexTable("OpenIdApplicationByRoleNameIndex", null);
 
-            SchemaBuilder.CreateReduceIndexTable(nameof(OpenIdAppByLogoutUriIndex), table => table
+            SchemaBuilder.CreateReduceIndexTable(typeof(OpenIdAppByLogoutUriIndex), table => table
```
deanmarcussen, line 77:
```suggestion
            SchemaBuilder.CreateReduceIndexTable<OpenIdAppByLogoutUriIndex>(), table => table
```

src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Migrations/OpenIdMigrations.cs
```diff
@@ -62,21 +70,24 @@ public int UpdateFrom1()
...
-                .Column<int>(nameof(OpenIdAppByLogoutUriIndex.Count)));
+                .Column<int>(nameof(OpenIdAppByLogoutUriIndex.Count)),
+                null);
 
-            SchemaBuilder.CreateReduceIndexTable(nameof(OpenIdAppByRedirectUriIndex), table => table
+            SchemaBuilder.CreateReduceIndexTable(typeof(OpenIdAppByRedirectUriIndex), table => table
                 .Column<string>(nameof(OpenIdAppByRedirectUriIndex.RedirectUri))
-                .Column<int>(nameof(OpenIdAppByRedirectUriIndex.Count)));
+                .Column<int>(nameof(OpenIdAppByRedirectUriIndex.Count)),
+                null);
```
deanmarcussen, line 85: the collection name parameter is optional, so I would not pass null. The intent was for it to be non breaking, so we shouldn't need to change this. Just the types.

Can you use the generics everywhere?

src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Migrations/OpenIdMigrations.cs
agriffard: ok

src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Migrations/OpenIdMigrations.cs
agriffard: ok

kevinchalet:
IMHO, it's a bit ugly to avoid empty classes in the OpenID stuff just to make YesSql happy. Do we know why a CLR `Type` is now required by YesSql?

deanmarcussen:
> IMHO, it's a bit ugly to avoid empty classes in the OpenID stuff just to make YesSql happy. Do we know why a CLR `Type` is now required by YesSql?

Yeah, it's not super pretty. We have to do a similar thing for migrations where we are affecting settings where the settings class is in another project. Might actually be able to make them private, not internal, @agriffard , which keeps them in the migration class (my bad I think we do that elsewhere)

The change in YesSql was related to adding support for collections, and separating indexes across different collections, so @sebastienros chose to use a type (though I am not sure if that was purely for convenience or technical reason). There is a naming convention that can impact that, but I think in that case the `Type` is still required.

deanmarcussen:
@agriffard I forgot about this. Can you merge dev on top of it, and if all passes LGTM

Issue #6615, opened 2020-07-09
Field and Body editor alignments / sizing


deanmarcussen:
There are some inconsistencies with sizing of some field elements settings and other views.

Some hints are full-width, some are restricted to col-md-6, other col-sm-6

Hint boxes, should also align together with the rest of the fields

![Screenshot 2020-07-09 at 08 24 53](https://user-images.githubusercontent.com/13782679/87010075-f4e74b80-c1bd-11ea-8477-e4b873fe097b.png)

hishamco:
Should we agree to use `col-md-6` and move the first hint to the next line to save some spacing?

deanmarcussen:
Hint texts should all be full width, not columned, and the hint box just needs it's sizing to match the checkboxes etc.

The sanitize html check box should line up with the beginning of the other elements.

hishamco:
Let me check this all the content fields

deanmarcussen:
Update on this.

There are two / three issues to resolve.

One is a regression from https://github.com/OrchardCMS/OrchardCore/pull/6509
The second is a semantic regression / change from a long time ago when we migrated from bootstrap 3 to bootstrap 4
The third is the actual sizing of fields editors in the admin (`col-md-6`)

Let me explain (and sorry for the novel)

Bootstrap 3 had no auto sizing column classes, such as, `col`, `col-md` . Bootstrap 4 has auto sizing classes.  https://getbootstrap.com/docs/4.1/layout/grid/#auto-layout-columns

We had our own autosizing bootstrap extensions which are used in _some_ places in the admin. But not in others (probably because no one knew about them, and they are not obvious because of the semantic changes in bs4.)

So in our admin css `col-md` kind of means to be a _mediumish_ column depending on the breakpoint, rather than be a column at the `md` breakpoint. This is superconfusing!

Here's the sass that makes it so
```
.col-md {
        @extend .col-12, .col-sm-12, .col-md-10, .col-lg-8, .col-xl-6;
}
```
So at `md` it's actually 10 units, but at `xl` it's actually 6 units. So it's not an `md` column. It's conceptually a medium sized column.

This made sense in bootstrap 3, because there was no concept of autosizing columns. So in bs3 `col-md` meant nothing. However in bs4 `col-md` now has significant meaning.

If you search the cshtml files there are roughly 30/40 usages of `col-sm` `col-md` `col-lg` currently not doing what they were doing. 

[...]

PR #6624, opened 2020-07-10
Admin css / bootstrap regressions

deanmarcussen:
Related to https://github.com/OrchardCMS/OrchardCore/issues/6615

To demonstrate a little what is going on


deanmarcussen:
This fixes regressions from #6509 and returns the admin css sizing to as before.

It also fixes a regression from moving to bootstrap 4 (from v3) where col-md had no meaning in bs3, but does in bs4.
Our css was using a custom `col-md` to indicate a medium size column, which then was affecting bs4 `col-md` which has a meaning in bs4.

To resolve this, I have refactored the editor views using `col-md` and `col-sm`, expecting a medium or small size column to use a better named class `col-medium` and `col-small` where they previously used `col-md`.

This returns all admin editor views to the previous sizing conventions.

It does not, however, totally fix the issue that in some places we use a `col-medium` and others we use things like `col-md-6` but I figure we should address that separately, as @hishamco has started doing in his pr.

@hishamco this should not interfere with your pr, but will cause some conflicts. I suspect you didn't see it because maybe you need to refresh your browser cache.

hishamco:
@deanmarcussen shall I wait for your PR to merge, or just refresh my browser cache?

src/OrchardCore.Modules/OrchardCore.Deployment/Views/Items/RecipeFileDeploymentStep.Fields.Edit.cshtml
```diff
@@ -4,7 +4,7 @@
 
 <div class="form-group">
     <div class="row">
-        <div class="col-sm">
+        <div class="all">
```
hishamco, line 7: Just I have a doubt about this one, or the intent to take the full width? If is that, why I didn't see `all` in SCSS file?

Also we could have a better meaningful name

deanmarcussen: yeah good catch, typo

hishamco: It happens ;)

Issue #6594, opened 2020-07-06
Media Azure inconsistency


ruisdias:
Hi,

I discovered a little bug that I would like to report to you guys.

My Type structure is:

Type
{
title
BagPart<Type2>  {
            title,
            BagPart<Type3> {
                      title,
                      image
          }
    }
}

So my Type3 is the one that has a media field and it's an item inside of a bag part in the Type2.
Type2 is an item inside of a bag part in the Type parent.

In this case, every change that I make to the item (Type) it will result in a new id to the items from Type3.

So if I save it as a draft, it will create a new Id for every Type3, so the version which is still published lost its media field because in the blob storage already as new id which is not the one here.

Please note that ContentItemVersionId is always null for Type3, and ContentItemId is almost used as a versionId because it's always generated a new one.

How to reproduce:
- Create an item with a  media file( image, pdf, wtv.);
- Publish it;
- Check-in your blob storage that the file was uploaded with the content item id;
- Change anything in the item;
- Save it as a draft;
- Check-in your blob storage for the folder with the old content item id;

deanmarcussen:
I can repro this @ruisdias 

So when working with local media (not azure) it is resetting the `ContentItemId` on the nested items, and then moving the old media into the new medias folder / name.

(This is using the attached editor, which I presume you are using)

So nothing is getting lost, but some extra images seem to be piling up, from the old `ContentItemId` name.

What is happening with blob storage? Your issue suggests you are losing your media? Which suggests maybe another issue with blob storage and the attached field, which maybe different.

@Skrypt has raised a few issues with blob storage and the attached field also. But I'm still need some clear understanding of what the issue there is and how to repro it.

ruisdias:
Yes, I can lose data, images in this case.

Imagine that I want to do a preview feature, when I save it as draft the image of the published one disappears.

With blob storage, the reset of the id is always renaming the folders that leads to not consistent data.

ruisdias:
And answering to you I'm using the attached edit because I wanted to prevent other possible errors, like cache on my frontend application.

deanmarcussen:
Thanks for the extra info @ruisdias without it I would have thought I fixed the issue.

But actually there are two issues here.
- Content Item Ids not being maintained across nested bags/widgets (easy fix)
- The Attached Media editor has an issue with it's handling of temp files. (hard to fix)

[...]

PR #6626, opened 2020-07-10
Maintain nested contained ids across bags and flows

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6594

wip, have identified another issue with the media attached driver and the content preview functionality


deanmarcussen:
This fixes two issues.

- When nesting attached media fields in bags (attached fields relies on the content item id being constant) we need to maintain the internals of this data when updating.

- When using Preview on an attached media field media items are lost (not just in bags, but any media field), because the preview request intercepts a driver call to move items from a temp folder, to the correct folder.

This implements a preview action attribute on the preview controller to apply a value to the `HttpContext`, so the `AttachedMediaService` can check if we are in preview mode, and not apply the file move.

Done with an attribute, to provide a consistent way of knowing if we are in preview, if that becomes necesary.
We have been using `-1` on the content item id as well, (which maybe slightly related to lucene) but I have not changed that - we could at some point rationalise the two techniques if appropriate (I leave that up to you @jtkech as I think you know what it does!)

Also only done with an action attribute, a razor page content preview renderer doesn't seem necessary.

jtkech:
@deanmarcussen 

Hmm, finally i would have just use `httpContext.Items` (without an attribute) to be more global, but as you want.

That's okay for the razor page content preview, in that case the same controller is first used so here also we just need that the preview mode is applied before calling `UpdateEditorAsync()`, then if there is a `PreviewPart` this is the `PreviewStartupFilter` that does a redirection.

Yes, in this PR you can just do the same check in `LuceneIndexingContentHandler` in place of

            // A previewed content item is transient, and is marked as such with a negative id.
            if (context.ContentItem.Id == -1)
            {
                return Task.CompletedTask;
            }

deanmarcussen:
@jtkech I did wonder about the need for an extra project, just for an attribute.

I will drop it, as you have questioned it as well, and just use Items.

> LuceneIndexingContentHandler

Thank you ! I had forgotten where we were using the -1

deanmarcussen:
@jtkech changes made to just use `Items`.

Can you check. I have also added a check into the `IndexingTaskManager` which I felt was missing the same check as `LuceneIndexingContentHandler`

Lastly I see `IndexingTaskManager` is checking for `ContentItemId=0` but `LuceneIndexingContentHandler` is not. Feels like maybe it should too?

We also have a `PreviewController` as part of the `Templates` feature, but it is not calling `UpdateEditor` so I don't think it has the same issue

jtkech:
@deanmarcussen 

> Can you check. I have also added a check into the IndexingTaskManager which I felt was missing the same check

Good catch you're right

> Lastly I see IndexingTaskManager is checking for ContentItemId=0 but LuceneIndexingContentHandler is not. Feels like maybe it should too?

Just tried, if we create, validate and save an item there are subsequent events / contexts with a `ContentItem.Id` not equal to 0, and in the deferred task we only act on the last one. But if the validation fails there is only an updated event with an Id equal to 0, it has no impact as we will do a `luceneIndexManager.DeleteDocumentsAsync()` on a `ContentItemId` that it will just not found, but you're right better to check it here also.

> We also have a PreviewController as part of the Templates feature, but it is not calling UpdateEditor so I don't think it has the same issue

That's okay as we preview the changes that have been done to the template without having changed any content item. Hmm, but why not still setting `Items["Preview"]` if we will need it for another reason in another place.

src/OrchardCore.Modules/OrchardCore.ContentPreview/Controllers/PreviewController.cs
```diff
@@ -63,10 +66,6 @@ public async Task<IActionResult> Render()
             var contentItemId = Request.Form["PreviewContentItemId"];
             var contentItemVersionId = Request.Form["PreviewContentItemVersionId"];
 
-            // Unique contentItem.Id that only Preview is using such that another
-            // stored document can't have the same one in the IContentManagerSession index
-
-            contentItem.Id = -1;
```
jtkech, line 69: We still need this because of the above comments, it is necessary when we use a preview part / aspect to redirect (through the `PreviewStartupFilter`) to the preview path e.g. a razor page that will query and need to retrieve the previewed content item through the `IContentManagerSession`.

deanmarcussen: Returned ;)

deanmarcussen:
> Update: Seems very simple but TBH would need more thinkings as there are many ways to think about it. Similarly in some drivers we were checking if Id == 0 to say that the content item is new, but this was no more true with items contained in a bag. That's why we introduced the IsNew property in BuildEditorContext. But we didn't do that in the UpdateContentContext that's why we still check Id == 0 in IndexingTaskManager. So first here i think it would be better to also have an IsNew property in the UpdateContentContext and then check it in the UpdatedAsync() of CreateIndexingTaskContentHandler to call or not _indexingTaskManager.CreateTaskAsync().

@jtkech have updated to include your initial suggestions.

Nothing is ever simple. Need to think more about this, `isNew`, `isPreview`.

It makes sense, but equally `isPreview` only makes sense if the `Preview` module is enabled (for example), so I kind of prefer to have it less known, on the content manager context?

jtkech:
@deanmarcussen Okay cool, i'm fine with it

So my last suggestion is to make these checks in the `UpdateAsync` only in drivers / handlers, as it is already done in `LuceneIndexingContentHandler`, so  in `MediaFieldDisplayDriver` in place of `AttachedMediaFieldFileService`, and in `CreateIndexingTaskContentHandler` in place of `IndexingTaskManager`.

And then add `handlers` in the comment that you use in 2 places, maybe like this

    Add a marker item to the request so that drivers / handlers can be aware of an active preview mode.

deanmarcussen:
> Nothing is ever simple. Need to think more about this, isNew, isPreview.

@jtkech I looked at the changes that would be required to add `isNew` and `isPreview` to both the `BuildEditorAsync` methods and the `ContentUpdatedContext`.

After I got to changes in 30 files, I looked at the changes and decided it was **not the thing to do**. Even though it **maybe the right thing to do**, if you know what I mean. 

However if you do think it is the right think to do, and worth doing, then I will make the change

> But i agree, sometimes we want to fix it at another level, e.g. thinking here that this is the reponsability of the service, e.g. to support more use cases, so as you want ;)

I originally did it in the `Handler` and `DisplayDriver` then felt it was more the responsibility of the service, so moved it to the service. Then I wondered about how that would impact on something like a background task using this service, but I think we have a null http context (or similar) even for that case, so it would still be ok to do it in the service. But if you think it is better not to be in the services, I'm happy to move it back up.

I'm updating the comment to reflect comment around services / handlers / drivers

Let me know when you have time, I'm uncertain in my opinion on both these issues, so it helps to have decisions confirmed :)

jtkech:
@deanmarcussen 

No problem i'm fine with what you did

Yes, my last suggestion was to only check it in drivers / handlers but by still using `httpContext.Items`, not the update contexts ;) Yes, for each background task there is an isolated fake httpContext.

sebastienros:
Preview.Abstractions project with a "Feature" class in it (PreviewFeature) with a 'Previewing' boolean property that is set/unset by the preview feature, and read from other modules.

deanmarcussen:
> Preview.Abstractions project with a "Feature" class in it (PreviewFeature) with a 'Previewing' boolean property that is set/unset by the preview feature, and read from other modules.

Done

jtkech:
@deanmarcussen see my comment in #6887, because in a WF we may want to prevent (or not) some work after a ContentUpdatedEvent in preview mode, we may need a a script method and / or a liquid filter checking the new incoming ContentPreviewFeature, and that could be used by a WF If / Else task. But in another PR ;)

Issue #6620, opened 2020-07-09
Containers liquid filter


arkadiuszwojcik:
Currently existing `container` liquid filter is very useful but it is limited to return just one, direct parent element for list item. However there are scenarious where we would like to return whole list of parents up to the root. In my particular case such list is required to render breadcrumbs (hierarchy of elements have few levels). I have created my own filter `containers` and now I wonder if this should be out of box Orchard Core filter?

jtkech:
Looks like a good idea, feel free to open a new PR to suggest it

PR #6631, opened 2020-07-10
Containers liquid filter to return list of ancestors 

arkadiuszwojcik:
This PR address feature proposal mentioned in #6620


jtkech:
Will review it asap

src/OrchardCore.Modules/OrchardCore.Liquid/Filters/ContainersFilter.cs
```diff
@@ -0,0 +1,50 @@
...
+                throw new ArgumentException("A Content Item was expected");
+            }
+
+            var containers = new List<ObjectValue>();
+
+            string containerId = contentItem.Content?.ContainedPart?.ListContentItemId;
+
+            while (containerId != null)
+            {
+                contentItem = await _contentManager.GetAsync(containerId);
```
jtkech, line 34: @arkadiuszwojcik 

I suggest to define your liquid filter in the `OC.Lists` module, there are already other filters in this module under a `Liquid` folder, and then register it in the startup of the same module, this through the `LiquidStartup` that will be executed only if the `Lists` feature is enabled and also only if the `Liquid` feature is enabled, because of the related `[RequireFeatures("OrchardCore.Liquid")]` attribute.

    [RequireFeatures("OrchardCore.Liquid")]
    public class LiquidStartup : StartupBase
    {
        {
            services.AddLiquidFilter<ListCountFilter>("list_count");
            services.AddLiquidFilter<ListItemsFilter>("list_items");
            // Register here your liquid filter
        }
    }

arkadiuszwojcik: Good idea. In such case "list_ancestors" or "list_item_ancestors" will be better filter name.

deanmarcussen:
I'm a little concerned about the recursive nature of this.

For a item nested 4 or 5 deep, it'll make 4 or 5 sql calls.

So it's no unbounded, but still seems a lot.

It might be better to store an array on the parent container, and as they're added check that, and add it to the item being added.

jtkech:
@deanmarcussen

Yes i thought about that but was thinking that if there are not so much ancestors and if people know what they do

Maybe grab these infos when editing saving a child but will be complex when e.g. we move an element from one list to another (not sure we can do this), need more thinkings ...

arkadiuszwojcik:
This is how List is designed at the moment, so some operations are much more complex than others. My particular case is related to breadcrumbs rendering. I have such list hierarchy: `Authors -> Books -> Chapters`. So when I render chapter I would like to render  breadcrumbs: `Home -> Author_name (as link) ->Book_name (as link) -> Chapter_name` , at this moment getting all containers up to the root is the only solution. Maybe it would be worth to include information about complexity of each list operation in doc?

jtkech:
@deanmarcussen @arkadiuszwojcik 

For your use case why not just using the existing container filter in a loop

    {% for counter in (0..9) %}
        {% assign container = item | container %}

        {% if container.ContentItemId == item.ContentItemId %}
        
            {% break %}
        {% endif %}

        
        {{ container | display_text }}

        {% assign item = container %}
    {% endfor %}

deanmarcussen:
@jtkech @arkadiuszwojcik yes agree, I have used the loop when building autoroute slugs previously.

@arkadiuszwojcik for breadcrumbs (i.e. things on the front end that need to be faster than the admin) the solution I am using is to recursively split the path segments backwards, look each path part up in `IAutorouteEntries.TryGetEntryByPath`.

Then I have the four or five content item ids that make up the path, and I can do a single query against the `ContentItemIndex` index to obtain the display text, which is all I need for a breadcrumb.
Equally if more data is needed you could also return the document.

Then optionally I might wrap the breadcrumb shape in a cache block with cache dependencies varied by route.

jtkech:
The little problem is the recursive call of `_contentManager.GetAsync()`, maybe a better way, and there is a workaround with a liquid loop to achieve the initial goal. Also it doesn't bluid, i think because some liquid filters have been moved recently.

So right now i close this one, @arkadiuszwojcik feel free to open a new PR if you want to suggest again this liquid filter, and by adding it in the `OC.Lists` module and with the right name.

PR #6637, opened 2020-07-11
Avoid AUTH by OpenId if the account is disabled

### LINKED ISSUE #6621: The user has been disabled, but can still authenticated by OpenId.
The OpenId module does not seem to handle the User.IsEnabled field.

hishamco:
Fixes #6621

### DIFF
--- a/src/OrchardCore/OrchardCore.Users.Core/Services/UserService.cs
+++ b/src/OrchardCore/OrchardCore.Users.Core/Services/UserService.cs
@@ -71,7 +71,7 @@ namespace OrchardCore.Users.Services
 
             if (!(user as User).IsEnabled)
             {
-                reportError(string.Empty, S["The specified account is disabled, please contact an administrator."]);
+                reportError(string.Empty, S["The specified user is not allowed to sign in."]);
                 
                 return null;
             }
--- a/src/OrchardCore/OrchardCore.Users.Core/Services/UserService.cs
+++ b/src/OrchardCore/OrchardCore.Users.Core/Services/UserService.cs
@@ -69,6 +69,13 @@ namespace OrchardCore.Users.Services
                 return null;
             }
 
+            if (!(user as User).IsEnabled)
+            {
+                reportError(string.Empty, S["The specified account is disabled, please contact an administrator."]);
+                
+                return null;
+            }
+
             return user;
         }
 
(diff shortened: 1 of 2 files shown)


src/OrchardCore/OrchardCore.Users.Core/Services/UserService.cs
```diff
@@ -69,6 +69,13 @@ public async Task<IUser> AuthenticateAsync(string userName, string password, Act
                 return null;
             }
 
+            if (!(user as User).IsEnabled)
+            {
+                reportError(string.Empty, S["The specified account is disabled, please contact an administrator."]);
```
sebastienros, line 74: Could it be the same error message as line 56, where and account is locker from SignInManager?

`reportError(string.Empty, S["The specified user is not allowed to sign in."]);`

Or change SignIngManager if we have a custom implementation.

sebastienros: Otherwise LGTM.

hishamco:
Seb I edit the message directly from GitHub, coz my local repo was corrupted ;( If it's looks good :shipit:

src/OrchardCore/OrchardCore.Users.Core/Services/UserService.cs
```diff
@@ -69,6 +69,13 @@ public async Task<IUser> AuthenticateAsync(string userName, string password, Act
                 return null;
             }
 
+            if (!(user as User).IsEnabled)
+            {
+                reportError(string.Empty, S["The specified user is not allowed to sign in."]);
```
deanmarcussen, line 74: Can you check the `AccountController` @hishamco 

There are two checks there, one for email should be confirmed, the other for account disabled. And the strings are different for both.

Ideally the behaviour should be the same

hishamco: Do you mean this one https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.Users/Controllers/AccountController.cs#L166 if it is that the previous message before last commit looks similar, but Seb suggested that in https://github.com/OrchardCMS/OrchardCore/pull/6637#discussion_r453959838

PR #6644, opened 2020-07-12
Use readonly properties in file store entries

hishamco:


deanmarcussen:
Why? was there a bug and it wasn't working?

Has it been tested?

hishamco:
Actually there's no bug, but there's no need for the baking fields here nothing more

deanmarcussen:
> Has it been tested?

So not tested then?

hishamco:
> Has it been tested?

I can make unit tests to test them, but I find that there's no need while the change is not too much, I will write the unit tests then ...

deanmarcussen:
> I can make unit tests to test them, but I find that there's no need while the change is not too much, I will write the unit tests then ...

I wasn't asking for unit tests, just to know whether you connected to blob storage (or blob development storage) and it still works?

There has been so many issues with prefixing and semantics with blob storage that I'm reluctant to approve a change if it hasn't been tried. (Knowing that if it breaks something I'll have to come along and figure out why)

hishamco:
Got your point Dean, I will try to use blob development storage if there's a change, coz I didn't work with with yet

Thanks

agriffard:
Can you please finalize it or close it if it is not meant to be merged?

hishamco:
I'm not able to connect to blob storage since then, that's why I didn't update the status, but the changes that I made doesn't affect nything it's just refactoring

agriffard:
Closing while it is not tested. Reopen if needed.

PR #6648, opened 2020-07-13
Don't Detach() a document already loaded for update

### LINKED ISSUE #6356: Content step in a migration recipe fails
Hi,

A taxonomy field is not added to a custom part when I use a TaxonomyContentItemId that was added by a recipe migration in the same data migration.

I created a repo for this: https://github.com/ArieGato/OrchardCore/tree/ariegato/recipe-migration

To reproduce follow these steps:

1. Create a new web site by using the Coming Soon theme.
2. Activate the 'Recipe Migration Issue' module.
The migration checks for the existence of a taxonomy content item by alias. When it's not present it creates it by using a recipe migration.
3. Check the content types. The 'Custom' type has been created with a Custom part, but the part cannot be found at the Content Parts page. When editing the Custom Part from the Custom type details a 404 page not found error is shown.

Now to see how it should work:

When the Taxonomy content item with the alias has already been created, the migration works fine.
1. Create a new web site by using the Blog recipe (which creates the taxonomy categories)
2. Activate the module Recipe Migration Issue.
3. Check the Custom Part in the Content Parts page. The part is created correctly with the Taxonomy field.

Thnx,
Arjan

jtkech:
Fixes #6356 

- First, when we get a **single document** for caching, if it has already been loaded for mutation we do a **session detach** but the loaded object may have been already flushed and **if we save it again a new record is persisted, that's why here we just throw in that case**. The opposite is okay, if we first get a single document for caching, we know that it will not be mutated and saved, so we can still detach it allowing to load after an isolated object for mutation.

- **This is what happens in a migration recipe** where all steps are executed in the same scope and that also uses a `Content` step. **The definition record is first loaded for mutation** by a previous migration and then the `Content` step **does a get for caching**. This ends up to **a duplicate definition record, the first one being incomplete**.

- **So here the solution is** to let as before all regular automatic migrations executed in the same activating scope, which make sense, and that **the `RecipeMigrator` uses a `DeferredTask`** so that the migration recipe will be executed after disposing the scope used to activate the shell and where all other regular migations has been done and committed.

    And **each step of a migration recipe is now executed in a new scope** (with a new parameter saying that the shell should not be activated as it is in progress) **allowing to mix any kind of recipe steps**, as with other regular recipes.


jtkech:
I found a better and simpler way to fix it

So, when we load a document for caching that has already been loaded, currently we detach it before, the problem is that it may have been flushed so we can get and cache uncommitted data, and if we save it again a new record is persisted.

The simpler solution in that case is to just return the loaded document but indicating that it should not be cached. It is still thread safe as the loaded document is not put in a shared cache and then only used in a given scope.

Beautiful it's mainly 2 line of code ;)

The other changes are just because of the signature change of `GetForCaching()` that not only returns a document but also a bool indicating if it can be cached or not.

jtkech:
For infos it fixes the related issue from @ArieGato 

> Very nice!! I tested a more complex scenario and all works well. I'll try to come up with some more edge cases, but I'm confident it works nice!

src/OrchardCore/OrchardCore.ContentManagement/DatabaseContentDefinitionStore.cs
```diff
@@ -25,7 +25,7 @@ public Task<ContentDefinitionRecord> LoadContentDefinitionAsync()
         /// <summary>
         /// Gets a single document (or create a new one) for caching and that should not be updated.
         /// </summary>
-        public Task<ContentDefinitionRecord> GetContentDefinitionAsync()
+        public Task<(bool, ContentDefinitionRecord)> GetContentDefinitionAsync()
```
sebastienros, line 28: I think that the "caching" information should not leak at this level. It should always be either cacheable or not. Another solution is to change this method to two:
- CreateContentDefinitionRecord
- GetContentDefinitionRecord

And the user would check if Get returned null before calling Create. Right now it's more like a GetOrCreate() and the cachability of the result will vary.

src/OrchardCore/OrchardCore.Data/SessionHelper.cs
```diff
@@ -39,22 +39,23 @@ public async Task<T> LoadForUpdateAsync<T>(Func<T> factory = null) where T : cla
         }
 
         /// <inheritdocs/>
-        public async Task<T> GetForCachingAsync<T>(Func<T> factory = null) where T : class, new()
+        public async Task<(bool, T)> GetForCachingAsync<T>(Func<T> factory = null) where T : class, new()
```
sebastienros, line 42: More like `GetOrCreate` than `Get`. That would make the factory usage obvious.

src/OrchardCore/OrchardCore.ContentManagement/DatabaseContentDefinitionStore.cs
jtkech: Okay, i'm open to any suggestions

That said, this is no more the case in the `distributedCache` branch, it is done at the `IDocumentStore` level and only checked at the `IDocumentManager<>` level, then any service using an `IDocumentManager<>` don't have to care about this.

    /// <inheritdoc />
    public Task<ContentDefinitionRecord> LoadContentDefinitionAsync() => _documentManager.GetMutableAsync();

    /// <inheritdoc />
    public Task<ContentDefinitionRecord> GetContentDefinitionAsync() => _documentManager.GetImmutableAsync();

> It should always be either cacheable or not

Yes i agree but this is the singularity i described in my comments. In some rare cases the caller should not create a new instance but should not cache the returned one. But this only happens in a given scope where we can't refresh the cache for a document that has already been loaded for update and that may have been flushed.

> it's more like a GetOrCreate() and the cachability of the result will vary.

Yes exactly, before it was already a `GetOrCreate()` with a factory, but, because of the above, even the result is not null it doesn't mean that it can be cached

I will think about it

src/OrchardCore/OrchardCore.Data/SessionHelper.cs
jtkech: So here, in the `distributedCache` branch, it is done at the `IDocumentStore` (can be an `IFileDocumentStore`) level, same problem but cacheable is only checked at the `IDocumentManager` level.

> More like GetOrCreate than Get. That would make the factory usage obvious.

Yes i agree, before the change it was already using a factory, would you mean `GetOrCreateForCachingAsync`? In the `distributedCache` branch the related method is `GetImmutableAsync()` that would become `GetOrCreateImmutableAsync()`

I will think about it, but again i'm open to any suggestions, let me know

PR #6656, opened 2020-07-13
Format CommonPart DateTime

### LINKED ISSUE #6655: Edit CommonPart Created date
If you want to edit the CreatedDate of a content by adding a CommonPart and checking the 'Display date editor' setting, then the input is displayed with seconds and milliseconds.

Ex if the value is 2020-07-14T00:00:12.345:

image

Unfortunately, a datetime-local input doesn't let you edit the seconds or milliseconds:
https://stackoverflow.com/questions/58023438/cant-set-seconds-value-of-input-type-datetime-local-to-be-0
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local

I suggest to display only the hours and minutes or else, you won't be able to set a "rounded" date.

agriffard:
Fixes #6655

Use `asp-format="{0:yyyy-MM-ddTHH:mm}"` with `<input type="datetime-local" />` to edit the CommonPart CreatedDate with only hours and seconds.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Views/CommonPart-Date.Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Views/CommonPart-Date.Edit.cshtml
@@ -4,7 +4,7 @@
 <div class="form-group" asp-validation-class-for="LocalDateTime">
     <div class="row col-md-6 col-lg-3">
         <label asp-for="LocalDateTime">@T["Created"]</label>
-        <input asp-for="LocalDateTime" class="form-control" type="datetime-local" />
+        <input asp-for="LocalDateTime" asp-format="{0:yyyy-MM-ddTHH:mm}" class="form-control" type="datetime-local" />
         <span asp-validation-for="LocalDateTime"></span>
     </div>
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.Contents/Views/CommonPart-Date.Edit.cshtml
```diff
@@ -4,7 +4,7 @@
 <div class="form-group" asp-validation-class-for="LocalDateTime">
     <div class="row col-md-6 col-lg-3">
         <label asp-for="LocalDateTime">@T["Created"]</label>
-        <input asp-for="LocalDateTime" class="form-control" type="datetime-local" />
+        <input asp-for="LocalDateTime" asp-format="{0:yyyy-MM-ddTHH:mm}" class="form-control" type="datetime-local" />
```
hishamco, line 7: Is there any space should be added between the date and time?

agriffard: The format has an impact on the value which needs a json format, so with the `T`, not with a space between date and time.

Issue #6664, opened 2020-07-14
Order Categorized Items for each Taxonomy Term


serverside-is:
Just like has been done with Lists and Contained items, we need to be able to sort/order the items categorized with a taxonomy term.

As an example, please look at the Orchard1 site [https://www.spgl.pt](https://www.spgl.pt), where the first page has widgets, each of them showing/projecting content items categorized with a particular taxonomy term (belonging to a taxonomy called Featured, with terms like Carousel, FeatureTop, AsideSquares, and so on). Of course, this is just an example.

I implemented this, with the following principles:

- No breaking changes.
- Possibility of enabling or disabling ordering of content items for each Taxonomy Item (disabled by default).
- When ordering is turned on for a Taxonomy, a “Order Items” button appears after each Taxonomy Term of that Taxonomy. This gives access to the page where the items categorized with that term can be sorted.
- When a content item is categorized with a term, it goes into the first (higher order) position.
- On the sorting page, only the latest version of each content item is shown. If the item is a draft, and the published version is also categorized with the term, the order of the published version is also updated.

Technically, the order each categorized content item inside each term is stored in a new dictionary on the TaxonomyField and appears on the TaxonomyIndex. The TermContentItemIds was maintained as is, to prevent breaking changes. The sorting interface is heavily based on what was done in the Lists module.

![orderCategorizedItems](https://user-images.githubusercontent.com/3681658/87438011-eeeed180-c5e6-11ea-8527-d093ec6bba99.gif)

PR #6665, opened 2020-07-14
Order Categorized Items for each Taxonomy Term

serverside-is:
Solves #6664


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=6665) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=6665) before we can accept your contribution.<br/><br/>:x: serverside-is [sign now](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=6665)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OrchardCMS/OrchardCore?pullRequest=6665) it.</sub>

deanmarcussen:
This looks interesting. 

It would be great to see a demo if you might be able to show it at the meeting?

https://www.orchardcore.net/ or https://www.orchardcore.net/meeting

src/OrchardCore.Modules/OrchardCore.Taxonomies/Controllers/AdminController.cs
```diff
@@ -284,6 +291,64 @@ public async Task<IActionResult> Delete(string taxonomyContentItemId, string tax
             return RedirectToAction("Edit", "Admin", new { area = "OrchardCore.Contents", contentItemId = taxonomyContentItemId });
         }
 
+        public async Task<IActionResult> OrderCategorizedContentItems(string taxonomyContentItemId, string taxonomyItemId)
```
sebastienros, line 294: POST

src/OrchardCore.Modules/OrchardCore.Taxonomies/Drivers/TaxonomyPartDisplayDriver.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using AngleSharp.Html.Dom;
```
sebastienros, line 5: Left over?

serverside-is:
> This looks interesting.
> 
> It would be great to see a demo if you might be able to show it at the meeting?
> 
> https://www.orchardcore.net/ or https://www.orchardcore.net/meeting

Sorry for the late reply. Had to switch to another project the last few days. Hope to resume development of my first Orchard Core site next week. Will try to be at the meeting.

> This looks interesting.
> 
> It would be great to see a demo if you might be able to show it at the meeting?
> 
> https://www.orchardcore.net/ or https://www.orchardcore.net/meeting

src/OrchardCore.Modules/OrchardCore.Taxonomies/Controllers/AdminController.cs
serverside-is: ok

src/OrchardCore.Modules/OrchardCore.Taxonomies/Drivers/TaxonomyPartDisplayDriver.cs
serverside-is: To be honest I'm not sure where that came from. Will delete, also on AdminController.cs.

serverside-is:
> How do you assign the order value when an item is moved. Do you change the value for all the content items?

The order of every item (categorized with that specific term) is revised. At least the items between the old and the new position will be updated. If there was a “hole” left by a removed item, then more items will be updated. In the end, on the Order field of the TaxonomyIndex we will have a sequential list of integers starting with 1 (the last content item). Because the order is descending, a new item will have order MAX+1 and no other item will be touched.

This can probably be improved in performance, if we purposely leave gaps between order values (like starting/refreshing with multiples of 10 instead of sequential integers). Those gaps could be used for the order value of moved items, without touching other items. And maybe updating/refreshing all categorized content items only when we run out of “gaps” in the order value of two consecutive items. I will give this some thought.

deanmarcussen:
> The order of every item (categorized with that specific term) is revised. 

For the list ordering, for some limited optimization, I only updated the current page of items orders. 

So moving an update form top to bottom would limit the updates to 10 items (if using a 10 item pager).

It does mean you have to increase the page size, to get from one page, to another.

Not sure if you have paged this, or if that is a possibility?

serverside-is:
> > The order of every item (categorized with that specific term) is revised.
> 
> For the list ordering, for some limited optimization, I only updated the current page of items orders.
> 
> So moving an update form top to bottom would limit the updates to 10 items (if using a 10 item pager).
> 
> It does mean you have to increase the page size, to get from one page, to another.
> 
> Not sure if you have paged this, or if that is a possibility?

Implemented paging. But the “number of items per page” is a new option, for each taxonomy (right after the “Enable ordering” one). This way we don’t modify the number of items per page visible on the front end, when we need to change it on the “ordering page” of the dashboard.

Maintained the logic used for storing the order of the content items but, as was done on the Lists module, limited the number of content items that need to be updated (now, only those sitting between the old and new position of the moved item will be updated).

Still need to test this better.

And there is a problem with the use of a Dictionary on the TaxonomyField: entries added to it appear on the document json, but entries removed from it are not removed from the json. Tried to understand why, with no success. I might need to replace that dictionary with something else.

Also need to update the documentation, before considering this ready.

src/OrchardCore.Modules/OrchardCore.Taxonomies/Controllers/AdminController.cs
serverside-is: This is a GET, called when opening the page or when the pager is clicked. The POST is OrderCategorizedContentItemsPost, called when a categorized content item is moved to a new position.

deanmarcussen:
> Implemented paging. But the “number of items per page” is a new option, for each taxonomy (right after the “Enable ordering” one). This way we don’t modify the number of items per page visible on the front end, when we need to change it on the “ordering page” of the dashboard.

I would probably just have used the `ISite.PageSize` property for this, just to keep it simpler.
 
> And there is a problem with the use of a Dictionary on the TaxonomyField: entries added to it appear on the document json, but entries removed from it are not removed from the json. Tried to understand why, with no success. I might need to replace that dictionary with something else.

Can you point me to where in the code, i.e. which events this is not working on?

I suspect when you're editing it from the field driver it will work fine, but when it's being altered, presumably by the `Post` it will just need to use the `.Alter()` / `.Apply()` content item extensions, because it is not updating the `Elements` dictionary.

`.Alter()` and `.Apply()` are extensions allowing you to update field properties, otherwise they are basically readonly, but when it is called from a driver we update the `Elements` at the end of the driver call.

You'll see I used them in the `ContainerService` for when it is called by ajax, but I haven't quite seen the flow of calls here.

serverside-is:
> 
> I would probably just have used the `ISite.PageSize` property for this, just to keep it simpler.
> 

I really didn't want to have the paging changed on the frontend, just because someone is modifying the order of the content items on the backend and needed bigger pages for that. But I agree, it is more complex like this. The driver is the same for both scenarios, though, witch I think is nice.

> 
> Can you point me to where in the code, i.e. which events this is not working on?
> 

The `Dictionary<string, int> TermContentItemOrder` that was added to the TaxonomyField, to store the Order value of the content item for each taxonomy item, is being modified in two situations:

The first one happens when someone is reordering categorized items (the "post"). That doesn’t add nor remove items from the Dictionary, it just changes the order (Value) of the item for a specific taxonomy term (Key), and it works as expected, on `RegisterCategorizedItemOrder()`, where Content is written directly. No problem here.

The second one happens when we edit a categorized content item (normally!), and use a TaxonomyField to add or remove taxonomy terms. Those taxonomy terms must be added or removed from both properties of the TaxonomyField: TermContentItemIds and TermContentItemOrder. They must always represent the same terms (the first property, untouched by me, stores the list of taxonomy terms the content item is categorized with, and the second property, added by me, stores the order of the item for each of those taxonomy terms). This is done on the driver (`_taxonomyFieldService.SyncTaxonomyFieldProperties(field)` is called from there to update the second property, based on changes made to the first property). As you explained, on `ContentFieldDisplayDriver.cs`, `.Apply()` is called. But although it works perfectly for the first property (`string[]`), it doesn't work for the second (`Dictionary<string, int>`) when we are _removing_ items. Here are prints of whats going on before and after the `.Apply()`, when we edit a content item, and remove one taxonomy term on the taxonomy field:
BeforeApply
AfterApply
As you can see, although the Dictionary has only 2 elements, 3 elements persist on the json. The third one was not removed.

deanmarcussen:
> As you can see, although the Dictionary has only 2 elements, 3 elements persist on the json. The third one was not removed.

Try adding this `field.Content.TermContentItemOrder.RemoveAll();` with a comment to explain why.

JsonMerge.ArrayHandling.Replace doesn't handle dictionaries, so it will need to be cleared first.

We do that in the `WidgetsListPartDisplay` driver for the same reason, e.g.

```
            // Remove any content or the zones would be merged and not be cleared
            part.Content.Widgets.RemoveAll();
```

serverside-is:
> 
> Try adding this `field.Content.TermContentItemOrder.RemoveAll();` with a comment to explain why.
> 

That's it! Works perfectly. Thank you!!

serverside-is:
I think this is ready to use.

A note about Drafts and Published versions:

If the categorized content item has a Published and a Draft version, using the same content item summary to represent both of them (and keep their order in sync), doesn’t work - even if we didn't care about the lower flexibility, both versions might have a different set of taxonomy terms, potentially needing to appear on different "term" ordering pages. To solve this, both versions will appear on the ordering interface. To distinguish the “content item summary object” that represents a Draft version of the item, I added a pencil with a blue background, right after the drag handle. This solves the problem while keeping things simple: All the user needs to know is that the items that have the “draft icon” are not visible on the front end (they are a Draft), so the position he puts them on is where he wants the item to appear (or move to) when he publishes the draft.

There are two improvements that I’m leaving for another time:

- Prevent problems arising from optimistic concurrency (the list was changed by someone else while the user is working on it)
- Prevent problems arising from older versions of categorized content items being brought back to life via the Audit Trail Module (they will come with order values that might collide with existing ones)

src/OrchardCore.Modules/OrchardCore.Taxonomies/Controllers/AdminController.cs
sebastienros: I saw an object being modified. But there is no Save so it's ok. However altering this object seems weird.

src/OrchardCore.Modules/OrchardCore.Taxonomies/Indexing/TaxonomyIndex.cs
```diff
@@ -37,7 +38,7 @@ public override void Describe(DescribeContext<ContentItem> context)
             context.For<TaxonomyIndex>()
                 .Map(contentItem =>
                 {
-                    if (!contentItem.IsPublished())
+                    if (!(contentItem.IsPublished() || contentItem.Latest))
```
sebastienros, line 41: Need to think about it. If the index was meant to only contain the published version, that would be wrong.

serverside-is: The index only contained the Published version. I changed that, because we need to get the Draft version for the ordering page. Every preexisting use of the index was reviewed, to make sure that it is taking the Published version only.

src/OrchardCore.Modules/OrchardCore.Taxonomies/Controllers/AdminController.cs
serverside-is: On the last version of this method, I'm using the `TermPart` to pass some information to the `TermPartContentDriver` (Ordering and OrderingPageSize ). `TermPart` is never saved. I'm reusing the Driver for the ordering page, that's why I extract the TermPartViewModel from the result. This way, the Driver's `DisplayAsync` method is used for both the front end and the ordering page, instead of creating separate code for the ordering page.
Contrary to what happens on the Lists module, here we are dealing with content items (terms) that are stored inside a parent content item (taxonomy), witch forces less orthodox solutions.

src/OrchardCore.Modules/OrchardCore.Taxonomies/Indexing/TaxonomyIndex.cs
deanmarcussen: I think this is problematic.

The index was definitely designed to only take published.

This breaks around 50% of my queries, but that I can live with.

For example
Published : TermA
Draft: TermC & TermB

Would produce three index records
FooId, TermA
FooId, TermB
FooId, TermC

So FooId is now categorized with TermC, even though it's only in draft, and a query on TermC will return FooId

I think the index would have to also contain values for published and latest, which is what makes it more breaking

serverside-is: Yes, I understand. Looks like I disregarded the code that already exists outside the “official” Orchard Core solution.

It's sad that the index was made for Published version only, from the start. Doesn't seem to be the case with the ContainedPartIndex. As you know, with Lists we can order contained Items that are not published yet, and the index has both the Latest and the Published versions (even without those two extra columns).

Should I go back to the original index (plus the “Order” column), and give up on the ability to order Draft versions of categorized content items (only keeping eventual drafts in sync with its published versions)?

Occurred to me creating a second index with both the published and latest versions, and the extra “Published”, “Latest” and “Order” columns, to be used only when Ordering is enabled. But didn't sound too good. Maybe there is a better (non-breaking) way...

(My insistence on being able to order Draft versions comes from my user base. They will want to create a draft and move it to the intended position, before publishing it - instead of having the item appear at the top for a while, just because it’s new. But I can always build an “in-house” solution for my particular use case. No problem.)

Another thing that could change is the pagination. I saw that demo a few days ago where, instead of an option for the “page size”, there was a dropdown at the bottom of the page with several options for the page size. I think that is a better solution for this “ordering page” too.

serverside-is: I went a step back and gave up on the ability to order content items that don’t have a published version. This way, there’s no need to make breaking changes to the index (we stick to published items only, on the index).

Even so, if a content item also has a draft (categorized with the same term), the draft order is kept in sync with the moved published version.

There was another problem that needed a solution. When an item is cloned, or unpublished and later published again (etc.), it’s Order value regarding a Term can already be in use by another content item. That would result in duplicate Order values. Now this is detected and solved (see `EnsureUniqueOrderValues`).

(Among other minor improvements, I simplified the `OrderCategorizedContentItems` on the `AdminController`. It doesn’t do `BuildDisplayAsync` anymore, which was confusing and uncalled for.)

deanmarcussen:
I would like to revisit this @serverside-is 

Couple of things have changed.
- The taxonomies have been moved to `Latest` and `Published` in another pr, which is good for this because you wanted that functionality I recall?

Suggestions on the UI
- If you want to keep the ordering count setting, there should also be a matching setting for listing, rather than using the default site settings (just to keep it making sense for all)
- The `Order Items` button is actually really useful just as a `List Items` button, that exists whether you have ordering enabled or not. So I'm wondering if you agree about just turning it into `List Items` all the time, and when ordering is enabled, then they can also order.
- That would make the count setting more an admin count vs a front end count I imagine.

What do you think?

serverside-is:
Yes, a few days ago I noticed this change in the other pr, and changed the state of this pr to "draft" - until I find a bit of time to review and update this.

I think your suggestions make sense.

serverside-is:
I did a new implementation of this, simpler that the first one (that one was a little over-engineered).

There is:
- A setting to turn manual ordering on and off for each taxonomy item (same as previous implementations).
- No setting for the page size. I believe we should wait for #7069 to be ready, and use that, both on Taxonomies and Lists.
- A "List Items" button, for each term (instead of the previous "Order Items" button). If ordering is enabled, the categorized items list can be ordered.

On the Lists module, cloned items receive the next order number. The Taxonomy module does that with cloned items and imported items (guess it makes sense?). When the Audit Trail module comes out, items that are reverted to older versions will also need to be taken care of (right?).

Issue #6672, opened 2020-07-15
Modules should not reference other modules


deanmarcussen:
While working on another pr I noticed just how many modules are referencing other modules for no need.

Mostly this is just because they are doing so to get transitive references.

The lists project is an exception to this (dynamic permissions) and a seperate issue https://github.com/OrchardCMS/OrchardCore/issues/6247

PR #6673, opened 2020-07-15
Fix modules referencing other modules

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6672

Takes direct references on the required abstractions/core projects instead of transitively depending on other modules to provide those references.

This should be every module, except lists


deanmarcussen:
Ok, while I was there, I reviewed other references, removed superfluous references, and ordered them all.

Sorry many changes, but you go to do one thing, and see all the other bits...

Issue #6172, opened 2020-05-12
MediaField: multiple fields in one contentitem cause DOM multiple Id error


PiemP:
Hi guys, 

I have a contenttype with multiple MediaField and I see, in my browser console, that it cause an error on DOM structure because every media field have some hidden input used for validation messages (I suppose) with the same value in the Id attribute.

If you want, I can do a PR to solve this issue: if doesn't exist any other activity that would fix on this problem.

The issue come from MediaFieldEditLocalization.cshtml file in the relative module.

PS. The best solution is to put MediaFieldEditLocalization.cshtml in the "Edit template" for all the MediaField in the contentitem but now I don't know how to do that

PiemP:
Hi guys,

I have found another issue with Media in Attached mode that disable the possibilities to move selected text or a images inside a textarea or into Html editor. Now I'm working on a PR #6167, I have to create another PR to solve this issue or I can merge the solution for this problem into the PR #6167?

Thank you

deanmarcussen:
Not sure of the issue you have found here @PiemP but one way to avoid the multiple id warnings (though these are generally just warnings, not errors), so I'm not sure if it is your issue) on the `MediaFieldEditLocalization` is to add it via a shape provider just once, like is done for the modals

https://github.com/OrchardCMS/OrchardCore/blob/6cddd9f9904ff117ba82587f39fc76f49fac3cc3/src/OrchardCore.Modules/OrchardCore.Html/Media/MediaShapes.cs

But again, not sure of the details of the actual issue you are experiencing, so this may not be relevant

PiemP:
@deanmarcussen The problem with the DOM, at the moment, doesn't give other problems than the remarks on the browser console, but It could create issue in usage because the `MediaFieldEditLocalization` issue is all over MediaField code...
At this moment I don't have repro because I'm working on the second issue I have found in  `MediaField-Attached.Edit.cshtml` file. In the `initializeAttachedMediaField` function are bind and prevented the `drop` and the `dragover` events over all the DOM document and this cause the impossibilities of move selected image and selected text inside html editor in the same contentype. 

I'm looking a the `MediaShapes` to repro this good practice. 

Hope all I have write is understandable.

Thank you.

PS. I'm preparing a new PR for fix all this issue in the MediaField.

agriffard:
Reopened because @sebastienros argued about the related PR during the meeting that the use of the shapes was a complex solution and might be hard to understand why we did it like this in the future.            
[...]

PR #6676, opened 2020-07-15
Refactor media shapes and correct formatting in views

deanmarcussen:
Fixes issues with https://github.com/OrchardCMS/OrchardCore/issues/6172

/cc @PiemP


src/OrchardCore.Modules/OrchardCore.Media/Startup.cs
```diff
@@ -38,7 +37,6 @@
 using OrchardCore.Media.Shortcodes;
 using OrchardCore.Media.Services;
 using OrchardCore.Media.Settings;
-using OrchardCore.Media.Shape;
```
deanmarcussen, line 41: removed extra usings added for no need

src/OrchardCore.Modules/OrchardCore.Media/Views/MediaField-Attached.Edit.cshtml
```diff
@@ -3,54 +3,52 @@
...
 
 @{
-var settings = Model.PartFieldDefinition.GetSettings<MediaFieldSettings>();
+    var settings = Model.PartFieldDefinition.GetSettings<MediaFieldSettings>();
     var mediaFieldId = Html.IdFor(m => m);
-    }
+}
 
+<script asp-src="~/OrchardCore.Media/Scripts/media.js" asp-name="media" at="Foot" depends-on="admin, vuejs, sortable, vuedraggable"></script>
+<style asp-src="~/OrchardCore.Media/Styles/media.min.css" debug-src="~/OrchardCore.Media/Styles/media.css"></style>
```
deanmarcussen, line 11: returned this to the view it belongs to.
the resource manager handles multiple script inclusions by design, so you don't need to add a wrapper to get this on the page only once.

src/OrchardCore.Modules/OrchardCore.Media/Views/MediaField-Attached.Edit.cshtml
```diff
@@ -3,54 +3,52 @@
 @using OrchardCore.ContentManagement.Metadata.Models
 
 @{
-var settings = Model.PartFieldDefinition.GetSettings<MediaFieldSettings>();
+    var settings = Model.PartFieldDefinition.GetSettings<MediaFieldSettings>();
```
deanmarcussen, line 6: returned all the formatting in this view to good styling

src/OrchardCore.Modules/OrchardCore.Media/Views/Shared/MediaFieldEditLocalization.cshtml
```diff
@@ -0,0 +1,12 @@
+@if (!Context.Items.ContainsKey("MediaFieldEditLocalization"))
+{
+    @* Insert only once per request *@
```
deanmarcussen, line 3: this was all you needed to do to resolve the issue with multiple ids in the dom.

We use the shapes technique in the other modules, so that the media feature can be disabled, and the shapes will not load a modal for the media app (which is disabled)

What you did here was add a shape wrapper in the media module, which required that the media module was active (so was unecesary)

PiemP:
understand... thanks @deanmarcussen... more easy to do than to explain

sebastienros:
Thanks. So we didn't need custom shapes, my assumptions were right?

deanmarcussen:
> Thanks. So we didn't need custom shapes, my assumptions were right?

Yes, the assumptions are correct.

We still need all the shape wrappers for the `HtmlField` as they are used to prevent the `HtmlField` / `HtmlBody` adding a wrapper if the `Media` feature is disabled. So as no to depend on the `Media` feature, and require it, in order to have an `HtmlField` (at least this is my assumption).

That being said, it would be good to rationalize some of these models / trumbowyg scripts at some point, when I have time, as there is much duplicate code between them, creating confusion, and extra checks in scripts, for no particular need.

Using the wrappers there also defeats the purpose of being able to easily add an editor more, because you also need to add a shape wrapper to make it work. There is probably a simpler way, but for another pr

PR #6298, opened 2020-05-28
ContentCard and Forms Enhancements

ns8482e:
1. ContentCard refactoring and enhancements - code moved from `ContentCard.cshtml` to `OnDisplay` event.
2. Showing inline visuals for Form elements - Label, Input, Button, TextArea, Validation.
3. ContentCard toolbar is now extensible to inject `tool button` from any module without needing to change alternates of ContentCard.Edit.cshtml
4. Unified `Add Item` and `Insert Item` into single shape `ToolInsertWidget.cshtml`
5. Using `placement.json` to place shapes on to Tool - sidebar and support of Tabs
6. Created shapes for Widget Attribute editor and Workflow selectord.( TODOs in separate PRs) 
7. Rewrite of Title binding to support other properties. e.g `For`, `DefaultValue`, `ErrorMessage` etc
8. Validation for `ContainedContentType` for `BagPart `and `FlowPart`

Also,

Fixes: #4275 
Fixes: #6276 
Fixes: #3161 

contentcard-form

### Adding Tool button in ContentCard toolbar from from a module.

1. Define `IShapeTableProvider` and add it to Startup.
2. Define  `OnCreated` and `OnDisplaying` events and add tool button shape. e.g. `Views\ToolToggleWidgets.cshtml`   

```
 builder.Describe("ContentCard")
            .OnCreated(async context =>
            {
                //Create toolbutton shape and add it to footer
                dynamic cardShape = context.Shape;
                IShape toggleTool = await context.ShapeFactory.New.ToolToggleWidgets();
                toggleTool.Metadata.Name = "ToolToggleWidgets";
                cardShape.Footer.Add(toggleTool, "after");

            })
            .OnDisplaying(context =>
           {
               dynamic cardShape = context.Shape;
               // Configure tool button based on current widget content item or card shape
               cardShape.Footer.ToolInsertWidget.XXXX= cardShape.XXXX;

               //If tool button is not related to current content item or card shape that it can be removed from footer
              cardShape.Footer.Remove("ToolToggleWidgets")
           }
```

PR #6689, opened 2020-07-15
ValidationPart fix prefix

### LINKED ISSUE #4275: Forms: Broken Validation widget
Right now when using validation widget let say for field 'UserName' generated output is:

```
<div class="widget-body">
<span class="field-validation-valid" data-valmsg-for="ValidationPart.UserName" data-valmsg-replace="true"></span>
</div> 
```
This doesn't work if there is field validation message. I played a bit with Orchard source code and applied dirty hack in `DefaultHtmlDisplay `class:

```
var localContext = new DisplayContext(context);
localContext.HtmlFieldPrefix = shapeMetadata.Prefix ?? "";

// HACK
if (localContext.HtmlFieldPrefix == "ValidationPart")
    localContext.HtmlFieldPrefix = "";
```

So basically I ignore `HtmlFieldPrefix ` for ValidationWidget. With this hack validation message is displayed:

```
<div class="widget-body">
<span class="field-validation-error" data-valmsg-for="UserName" data-valmsg-replace="true">Field must be non empty</span>
</div>
```

I don't know overall Orchard Core architecture so I have no idea what should be proper soultion for this issue.

ns8482e:
~~- Added Error Message property~~
- Removed Prefix on Display

Fixes #4275

Split into smaller PR from #6298

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Forms/Drivers/ValidationPartDisplay.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Forms/Drivers/ValidationPartDisplay.cs
@@ -18,8 +18,7 @@ namespace OrchardCore.Forms.Drivers
         {
             return Initialize<ValidationPartEditViewModel>("ValidationPart_Fields_Edit", m =>
             {
-                m.For = part.For;
-                m.ErrorMessage = part.ErrorMessage;
+                m.For = part.For;                
             });
         }
 
@@ -29,8 +28,7 @@ namespace OrchardCore.Forms.Drivers
 
             if (await updater.TryUpdateModelAsync(viewModel, Prefix))
             {
-                part.For = viewModel.For?.Trim();
-                part.ErrorMessage = viewModel.ErrorMessage?.Trim();
+                part.For = viewModel.For?.Trim();                
             }
 
             return Edit(part);
--- a/src/OrchardCore.Modules/OrchardCore.Forms/Drivers/ValidationPartDisplay.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Forms/Drivers/ValidationPartDisplay.cs
@@ -11,8 +11,6 @@ namespace OrchardCore.Forms.Drivers
     {
         public override IDisplayResult Display(ValidationPart part)
         {
-            // Empty Prefix on Display shapes
-            Prefix = "";
             return View("ValidationPart", part).Location("Detail", "Content");
         }
 
--- a/src/OrchardCore.Modules/OrchardCore.Forms/Drivers/ValidationPartDisplay.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Forms/Drivers/ValidationPartDisplay.cs
@@ -18,7 +18,7 @@ namespace OrchardCore.Forms.Drivers
         {
             return Initialize<ValidationPartEditViewModel>("ValidationPart_Fields_Edit", m =>
             {
-                m.For = part.For;                
+                m.For = part.For;
             });
         }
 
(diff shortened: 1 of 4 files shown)


src/OrchardCore.Modules/OrchardCore.Forms/Drivers/ValidationPartDisplay.cs
```diff
@@ -11,6 +11,8 @@ public class ValidationPartDisplay : ContentPartDisplayDriver<ValidationPart>
     {
         public override IDisplayResult Display(ValidationPart part)
         {
+            // Empty Prefix on Display shapes
+            Prefix = "";
```
deanmarcussen, line 15: What happens if the field that has the error is contained inside another widget, like the container widget? 
Can that happen? Does it work ok then?
Or equally if the `ValidationPart` is contained in a container does that then cause a problem because it's prefix would have included the container prefix?

Or am I finding problems where there are none?

ns8482e:
Prefix is used to bind model. Here, for Display it’s irrelevant. Form designer is proving `For` property

src/OrchardCore.Modules/OrchardCore.Forms/Drivers/ValidationPartDisplay.cs
ns8482e: @deanmarcussen I guess it could be only an issue if  same`ValidationPartDisplay`  driver is used for `Display `and `Edit` in same request.

deanmarcussen: Yes, that was my concern. I think actually it is ok, because `BuildPrefix` is called multiple times each request.
I need to look a bit further

ns8482e:
@deanmarcussen moved code from DisplayDriver to view

src/OrchardCore.Modules/OrchardCore.Forms/Drivers/ValidationPartDisplay.cs
ns8482e: Ok - No change i driver - Removed from View

src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/ValidationPart.cshtml
```diff
@@ -1,4 +1,11 @@
 @using OrchardCore.Forms.Models
 @model ShapeViewModel<ValidationPart>
+@{
+    var htmlFieldPrefix = this.ViewContext.ViewData.TemplateInfo.HtmlFieldPrefix;
+    this.ViewContext.ViewData.TemplateInfo.HtmlFieldPrefix = "";
```
deanmarcussen, line 5:
```suggestion
   ViewContext.ViewData.TemplateInfo.HtmlFieldPrefix = "";
```

deanmarcussen: Yeah, that feels better. Maybe a comment as to why as well?

src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/ValidationPart.cshtml
```diff
@@ -1,4 +1,11 @@
...
+}
 
 @Html.ValidationMessage(Model.Value.For)
+@{
+    this.ViewContext.ViewData.TemplateInfo.HtmlFieldPrefix = htmlFieldPrefix;
```
deanmarcussen, line 10:
```suggestion
   ViewContext.ViewData.TemplateInfo.HtmlFieldPrefix = htmlFieldPrefix;
```

ns8482e:
> Where does the new `ErrorMessage` property get used / displayed?

It's taken from my other PR ( as it's related to ValidationPart) - where we could use to show custom error message ErrorMessage.
But I guess it's not useful in this PR until I have all together - so I will remove it for now  from here.

Issue #6585, opened 2020-07-04
Setup Error


yoalley:
System.IO.DirectoryNotFoundException
  HResult=0x80070003
  Message=Could not find a part of the path 'F:\Git\net core\cms\OrchardCore\src\OrchardCore.Cms.Web\App_Data\Sites\Default\ContentDefinition.json'.
  Source=System.Private.CoreLib
  StackTrace:
   在 System.IO.FileStream.ValidateFileHandle(SafeFileHandle fileHandle)
   在 System.IO.FileStream.CreateFileOpenHandle(FileMode mode, FileShare share, FileOptions options)
   在 System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
   在 System.IO.StreamWriter.ValidateArgsAndOpenPath(String path, Boolean append, Encoding encoding, Int32 bufferSize)
   在 System.IO.StreamWriter..ctor(String path, Boolean append)
   在 System.IO.File.CreateText(String path)
   在 OrchardCore.ContentManagement.FileContentDefinitionStore.SaveContentDefinitionAsync(ContentDefinitionRecord contentDefinitionRecord) 在 F:\Git\net core\cms\OrchardCore\src\OrchardCore\OrchardCore.ContentManagement\FileContentDefinitionStore.cs 中: 第 68 行
   在 OrchardCore.ContentManagement.ContentDefinitionManager.UpdateContentDefinitionRecord() 在 F:\Git\net core\cms\OrchardCore\src\OrchardCore\OrchardCore.ContentManagement\ContentDefinitionManager.cs 中: 第 348 行
   在 OrchardCore.ContentManagement.ContentDefinitionManager.StorePartDefinition(ContentPartDefinition contentPartDefinition) 在 F:\Git\net core\cms\OrchardCore\src\OrchardCore\OrchardCore.ContentManagement\ContentDefinitionManager.cs 中: 第 132 行
   在 OrchardCore.ContentManagement.Metadata.ContentDefinitionManagerExtensions.AlterPartDefinition(IContentDefinitionManager manager, String name, Action`1 alteration) 在 F:\Git\net core\cms\OrchardCore\src\OrchardCore\OrchardCore.ContentManagement.Abstractions\IContentDefinitionManager.cs 中: 第 60 行
[...]

PR #6690, opened 2020-07-15
#6585 Fixing Linux setup by making sure a directory created before creating a file

rserj:
image
Setup fails on Installation from Docker-Linux container. The directory needs to be created before file creation. On the image above Setup failed because there wasn't "Default" folder.

### DIFF
--- a/src/OrchardCore/OrchardCore.ContentManagement/FileContentDefinitionStore.cs
+++ b/src/OrchardCore/OrchardCore.ContentManagement/FileContentDefinitionStore.cs
@@ -65,6 +65,12 @@ namespace OrchardCore.ContentManagement
         {
             lock (this)
             {
+                var directoryPath = Path.GetDirectoryName(Filename);
+                if (!Directory.Exists(directoryPath))
+                {
+                    Directory.CreateDirectory(directoryPath);
+                }
+
                 using (var file = File.CreateText(Filename))
                 {
                     var serializer = new JsonSerializer();
--- a/src/OrchardCore/OrchardCore.ContentManagement/FileContentDefinitionStore.cs
+++ b/src/OrchardCore/OrchardCore.ContentManagement/FileContentDefinitionStore.cs
@@ -1,3 +1,4 @@
+using System;
 using System.IO;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


src/OrchardCore/OrchardCore.ContentManagement/FileContentDefinitionStore.cs
```diff
@@ -65,6 +65,12 @@ public Task SaveContentDefinitionAsync(ContentDefinitionRecord contentDefinition
         {
             lock (this)
             {
+                var directoryPath = Path.GetDirectoryName(Filename);
```
hishamco, line 68: Nice catch!! but I'm not sure how this work in Windows or MacOS? @rserj could please test in on other OS

Thanks

rserj: @hishamco thank you for your review. 
In the image above I reproduced issue on Linux, I commented out the fix and was able to reproduce issue on Windows, please see image below
image
I tested fix on Windows and Linux as well, it works fine, unfortunately, I can not test on MasOS, but following to API documentation
https://docs.microsoft.com/en-us/dotnet/api/system.io.file.createtext?view=netcore-3.1
DirectoryNotFoundException is expected behavior, 
>DirectoryNotFoundException
The specified path is invalid (for example, it is on an unmapped drive).

It only fails if you try to install Agency recipe with some of the external Database such as MySql

src/OrchardCore/OrchardCore.ContentManagement/FileContentDefinitionStore.cs
```diff
@@ -65,6 +65,12 @@ public Task SaveContentDefinitionAsync(ContentDefinitionRecord contentDefinition
...
+                if (!Directory.Exists(directoryPath))
```
sebastienros, line 69:
```suggestion
                if (!String.IsNullOrEmpty(directoryPath) && !Directory.Exists(directoryPath))
```

sebastienros: Otherwise you can get a NRE. I had the same issue last week on another repo.

PR #6708, opened 2020-07-17
Users with "Manage Users" permission can't delete themselves or change their roles

### LINKED ISSUE #6718: Disable some actions for logged users who have "Manage Users" permission
Users with "Manage Users" permission can't:

    Enable / Disable their accounts

    Change their roles

    Delete themselves

hishamco:
Fixes #6718

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Users/Views/UserButtons.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Views/UserButtons.cshtml
@@ -5,7 +5,14 @@
     var user = Model.User as User;
 }
 <a asp-action="Edit" asp-route-id="@user.Id" class="btn btn-primary btn-sm">@T["Edit"]</a>
-<a asp-action="Delete" asp-route-id="@user.Id" class="btn btn-danger btn-sm" itemprop="RemoveUrl UnsafeUrl">@T["Delete"]</a>
+@if (user.UserName == User.Identity.Name)
+{
+    <a class="btn btn-danger btn-sm disabled" itemprop="RemoveUrl UnsafeUrl">@T["Delete"]</a>
+}
+else
+{
+    <a asp-action="Delete" asp-route-id="@user.Id" class="btn btn-danger btn-sm" itemprop="RemoveUrl UnsafeUrl">@T["Delete"]</a>
+}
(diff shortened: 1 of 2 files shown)


hishamco:
Delete

Edit

agriffard:
The way you did it is actually 'A connected user cannot do some specific actions related to its own user in the admin (change roles, delete, disable)'.

src/OrchardCore.Modules/OrchardCore.Users/Views/UserButtons.cshtml
```diff
@@ -5,7 +5,14 @@
     var user = Model.User as User;
 }
 <a asp-action="Edit" asp-route-id="@user.Id" class="btn btn-primary btn-sm">@T["Edit"]</a>
-<a asp-action="Delete" asp-route-id="@user.Id" class="btn btn-danger btn-sm" itemprop="RemoveUrl UnsafeUrl">@T["Delete"]</a>
+@if (user.UserName == User.Identity.Name)
+{
+    <a class="btn btn-danger btn-sm disabled" itemprop="RemoveUrl UnsafeUrl">@T["Delete"]</a>
```
agriffard, line 10: Something like 
`<a class="btn btn-danger btn-sm @((user.UserName == User.Identity.Name) ? "disabled" : "")" itemprop="RemoveUrl UnsafeUrl">@T["Delete"]</a>` 
could avoid duplicating the line.

But it could be better to do it in the controller and return an error message if it occurs.

Same for the other changes.

agriffard:
Please create also an issue (not just a PR) when you try to fix a problem that is not trivial like this one.

src/OrchardCore.Modules/OrchardCore.Users/Views/UserButtons.cshtml
hishamco: I already did something similar to what you did but I faced an issues with the TagHelper, anyhow I will try to have a look again

hishamco:
> The way you did it is actually 'Prevent a connected user from doing some specific actions related to its own user in the admin (change roles, delete, disable)'.

If I'm not wrong the current logged admin **SHOULD NOT** delete his account or change his roles, otherwise unexpected behavior may occurs

agriffard:
I don't disagree with the problem you are trying to fix. 
I just found that the title was a bit misleading because it is related to any user that has the manage users permission.

Issue #6520, opened 2020-06-26
Lucene Query Api the the endpoints work with get only i can't get to work with post requests


hamdiRuwiss:
what type of body do we use when querying the endpoint because i tried most and nothing work with post
here is the get request :
```
https://localhost:5001/api/lucene/content?query={
  "query": {
    "bool":
    {
      "must" :[
        {"term": { "Content.ContentItem.ContentType": "Meal" }},
        { "term": { "Meal.AdType": "{{type}}" }},
           ]
    }
  },
  "sort": {
    "Content.ContentItem.CreatedUtc": {
      "order": "desc",
      "type": "double"
    }
  },
  "size": 3
}&indexName=Meal&parameters={"type": "Rent" }
```
This works fine in postman but when i try it using post items null

I also noticed the endpoints result is not consistent the content return the items with count and the document doesn't

deanmarcussen:
The post request and the get request should be the same

(i.e. don't put anything in the post body, put it all in the query string)

hamdiRuwiss:
ok that was intentional  , it woks if i try post with querystring , becuase i am building a search form with multiple options i thaught i could send the query in the post .

sebastienros:
We might want to accept the query in the body if it's a POST. @hamdiRuwiss Do you want to provide a PR for that?
That would probably mean split the current method in two.

hamdiRuwiss:
Yes i would like to do it

PR #6712, opened 2020-07-18
Modifying the Lucene api To accept Post Form Data

hamdiRuwiss:
Modifying the Lucene Api To accept Post Form Data When Content Type is Plain/Text
Also in the documents endpoint return the same result schema as content 
I have no idea if it was intentional to return items only but i believe the count is useful when paging through the result

Fixes https://github.com/OrchardCMS/OrchardCore/issues/6520


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=6712) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=6712) before we can accept your contribution.<br/><br/>:x: hamdiRuwiss [sign now](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=6712)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OrchardCMS/OrchardCore?pullRequest=6712) it.</sub>

src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/ApiController.cs
```diff
@@ -35,23 +36,29 @@ public class ApiController : Controller
...
-                ReturnContentItems = true
-            };
+            var result = await LuceneQueryApi(indexName, query, parameters, returnContentItems: true);
 
-            var queryParameters = parameters != null ?
-                JsonConvert.DeserializeObject<Dictionary<string, object>>(parameters)
-                : new Dictionary<string, object>();
+            return new ObjectResult(result);
+        }
+        [HttpPost]
```
jtkech, line 43: Add a new line before

src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/ApiController.cs
```diff
@@ -35,23 +36,29 @@ public class ApiController : Controller
...
 
-            var result = await _luceneQuerySource.ExecuteQueryAsync(luceneQuery, queryParameters);
+            var result = await LuceneQueryApi(indexName, query, parameters, returnContentItems: true);
 
             return new ObjectResult(result);
         }
 
-        [HttpPost, HttpGet]
+
+        [HttpGet]
```
jtkech, line 61: Remove extra line before

src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/ApiController.cs
```diff
@@ -63,19 +70,43 @@ public class ApiController : Controller
...
+            {
+                return this.ChallengeOrForbid();
+            }
+
+            var result = await LuceneQueryApi(indexName, query, parameters);
+
+            return new ObjectResult(result);
+        }
+
+        private async Task<Queries.IQueryResults> LuceneQueryApi(string indexName, string query, string parameters, bool returnContentItems = false)
```
jtkech, line 95: Here the convention is to name it `LuceneQueryApiAsync`

src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/ApiController.cs
```diff
@@ -63,19 +70,43 @@ public class ApiController : Controller
...
-                Template = query
+                Template = query,
+                ReturnContentItems = returnContentItems
             };
 
             var queryParameters = parameters != null ?
                 JsonConvert.DeserializeObject<Dictionary<string, object>>(parameters)
                 : new Dictionary<string, object>();
 
             var result = await _luceneQuerySource.ExecuteQueryAsync(luceneQuery, queryParameters);
```
jtkech, line 108: Because there is no ther await before, here you can just return

    return _luceneQuerySource.ExecuteQueryAsync(luceneQuery, queryParameters);

And then remove the async modifier above line 95

src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/ApiController.cs
```diff
@@ -35,23 +36,29 @@ public class ApiController : Controller
...
 
-            var queryParameters = parameters != null ?
-                JsonConvert.DeserializeObject<Dictionary<string, object>>(parameters)
-                : new Dictionary<string, object>();
+            return new ObjectResult(result);
+        }
+        [HttpPost]
+        [Route("content")]
+        public async Task<IActionResult> ContentPost(
+            [FromForm] string indexName,
```
deanmarcussen, line 46: What is we want to post as json?

Is it possible (am not sure, but I think this would work) to make this a `ViewModel` without the `[FromForm]` such that it would bind from both `Form` and `Json` posts?

src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/ApiController.cs
hamdiRuwiss: done

src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/ApiController.cs
hamdiRuwiss: Yes ,  Thank you

src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/ApiController.cs
hamdiRuwiss: I thought about it,but then there are differences when using [ApiController] attribute the default source binder for complex type inference is [FormBody] and since the default media formatter is Json it will through an exception not supported formatter unless you use json body with application/json content type .Besides the query parameter value has liqued syntax which i think will require the client to escape or prefill the values of the parameters in the query
I am not sure as well ,maybe if we need custom processing.
I was thinking about it when i thought could we have a form submission request inside the workflow that read these form values and then  run a lucene query task and inspect the json result after.

deanmarcussen: Hmm, I would still think this makes more sense as a json body than a form, for an api controller.

The parameters object must be supplied as json already in the get request, but it doesn't accept liquid syntax. Maybe you have misunderstood how the parameters object works here. It supplies the parameters, which are then read into the queries template context, and the template is then parsed as liquid using those parameter values.

It seems as unexpected as the post request only accepting query strings, that an api request would take `Form` values, (that must contain json in the parameters value).

Do you have a requirement to send form values to a lucene controller, rather than sending a json body? 

I'm just trying to understand the use case here?

> I was thinking about it when i thought could we have a form submission request inside the workflow that read these form values and then run a lucene query task and inspect the json result after.

I'm not sure how workflows connects with this api controller? Or maybe I have misunderstood the comment

hamdiRuwiss: You are right, json is more suitable for the api controller especially for multi-client dev.
problem is , i don't know how would i deal with the query template parameter as json string since it has these {{ }} the controller binder will through an exception when trying to deserialize the value from the body.
Do you mean to send the query template parameter as json object becuase this would mean dealing with DSL of searching

hamdiRuwiss: Workflow is a different thought sorry i mixed it here.

deanmarcussen: Ok, cool, lets figure it out.

I'm not sure exactly, I would imagine trying to send the template parameter as a string value on a json object (not trying to deal with the DSL as a model).

You're right it might need some escaping (`JSON.stringify()`), or possibly a `JsonConvertor` that could handle the property as a string, rather than trying to treat it as a jobject. Hopefully it is simpler than that though.

Shall we get something that doesn't work, and see what that looks like, and see if I can help you from there? 

We should have a look and see what elastic search is doing here as well

jtkech:
@agriffard will need more time to review this one

Before working for nothing, @Skrypt @deanmarcussen what do you think?

Skrypt:
We need to look at how ElasticSearch works with creating indexes and insert data on those.

https://www.elastic.co/guide/en/elasticsearch/reference/current/getting-started-index.html

deanmarcussen:
> Before working for nothing, @Skrypt @deanmarcussen what do you think?

I like the concept of taking a POST (although elastic takes a PUT) with a JSON body - not so much taking it as a body with Form fields via an API

@hamdiRuwiss was concerned that there would be issues with liquid in a JSON body, but I think we should be able to handle that - client side would have to escape any liquid fields, but that seems reasonable, and is probably already required to send it as part of the query string anyway I imagine.

jtkech:
So here, after a quick look, my feeling is to use a json body with `application/json`, and a model on the controller action side, the model could be of `LuceneQuery` type itself if the `parameters` are still provided through the query string, otherwise a dedicated one including a `parameters` field.

This is what we are doing in our `HttpRequestExtensions.PostAsJsonAsync()` used in `BlogPostApiControllerTests` that targets the `ApiController` of the `OC.Contents` module where the model is a `ContentItem` and where the `draft` parameter may be passed through the query string. Note: A custom `ContentItemConverter : JsonConverter` is defined.

jtkech (commented):
See my last comment in the main conversation

agriffard:
@hamdiRuwiss Can you please take the changes suggested by @jtkech into account?

agriffard:
@jtkech Can you please review it?

jtkech:
Sorry i worked on other things

Need to focus on it again, i will do it asap ;)

src/OrchardCore.Modules/OrchardCore.Lucene/LuceneQueryDTO.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace OrchardCore.Lucene
+{
+
+   [JsonConverter(typeof(LuceneQueryDTOConverter))]
+   public class LuceneQueryDTO
+   {
+      public string IndexName { set; get; }
+      public string Query { set; get; }
+      public string Parameters { set; get; }
```
deanmarcussen, line 11: why does this need a custom json convertor?

Can the model not just be `Dictionary<string,object>` here instead?

hamdiRuwiss: Yes , it will work this will get rid of the json converter

> :    public async Task<IActionResult> ContentPost(Dictionary<string,object> query)
        {

            LuceneQueryDTO queryDto=new LuceneQueryDTO(){
                IndexName = query["indexName"].ToString(),
                Query = query["query"].ToString(),
                Parameters = query["parameters"].ToString()
            };

sebastienros: Even with a DTO, I don't understand the need of a custom converter.

hamdiRuwiss: ok, The dto is used by the get and the post. For the get it is simply mapping from the query as strings
for the post the json load is three parts the indexName is string the query and the parameters are dynamic json that's is why we can not have the default binding to dto object works out of the box. we needed to load the json as whole objects then parse different properties like what @deanmarcussen suggested . or using custom converter
@jtkech pointed me to the contentItem api and i saw how it worked with half static half dynamic content of content items
if I did the same this would look like :
 public class LuceneQueryDTO
   {
      public string IndexName { set; get; }
      public dynamic Query { set; get; }
      public dynamic Parameters { set; get; }
   }
for the sake of get endpoint i avoided doing this because we would need different inputs
I am not sure which way to go . I prefer what @jtkech pointed because it keeps the code clean and explain the intent

deanmarcussen: > for the sake of get endpoint i avoided doing this because we would need different inputs

I would suggest making the endpoints different (that is for me the point of what you are doing here, is seperating out the differences between a `Get` and a `Post` )

```
public class LuceneQueryModel
{
  public string IndexName { set; get; }
  public string Query { set; get; }
  public Dictionary<string, object> Parameters { set; get; }
}

Other option you could also do is leave the `Get` and `Post` exactly as they are, and make a new endpoint `Put` with the new binding. Non breaking that way. and Elastic Search uses `Put` rather than `Post`, so is in keeping to have a `Put` endpoint for search (`Post` suggests that it's creating something here, when really it's just a query, that is idempotent)

Skrypt: POST means "create new" as in "Here is the input for creating a user, create it for me". PUT means "insert, replace if already exists" as in "Here is the data for user 5".

So neither are good for retrieving the data. This is why I'm not sure why we want it to work with a POST when this should work only with a GET if it retrieves only data.

Skrypt: Okay, I need to see the usage of POST and PUT on ElasticSearch. Because, here, this ApiController is meant to retrieve the data from a Lucene Query alias ElasticSearch query. Which means we should be able to pass a JSON Query in either a form or a querystring input/param. Though, there is no difference here in this PR about doing a GET or a POST.

Here is the difference clearly defined on this page for ElasticSearch : 

https://www.elastic.co/guide/en/elasticsearch/guide/current/index-doc.html#_using_our_own_id
https://www.elastic.co/guide/en/elasticsearch/guide/current/index-doc.html#_autogenerating_ids

And of course a GET is only a GET...

Skrypt (approved):
I approve these changes. And this ApiController might require more work to be compliant with how ElasticSearch API works. But the changes here are fine since they are not even compliant with what ElasticSearch does but should be backward compatible with what we currently have.

This ApiControler requires to be eventually refactored (possible breaking change) to be able to index data from a POST or PUT request.

PR #6731, opened 2020-07-21
Inner recipe overwrite parent variables

### LINKED ISSUE #6717: Recipes within recipes overwrite variables configuration
When running a recipe inside a recipe with the `Recipes` step the `variables` section / `_variablesMethodProvider` is overwritten with the values from the inner recipe.

Example
```
{
  "name": "CKEditorsample",
  "displayName": "CKEditor Sample",
  "description": "CKEditor Sample based on TheBlogTheme",
  "author": "ThisNetWorks",
  "website": "https://github.com/ThisNetWorks",
  "version": "1",
  "issetuprecipe": true,
  "categories": [ "default" ],
  "tags": [ "blog" ],  
  "variables": {
    "ckeditorSampleContentItemId": "[js:uuid()]"      <----- this variable is overwritten by the variables section in the blog recipe
  },
  "steps": [
     {
      "name": "Recipes",
      "Values": [
        {
          "ExecutionId" : "",
          "Name" : "Blog"                <----- this recipe has a variables section and overwrites the variablesmethodprovider
        }
      ]
     },
    {
      "name": "Feature",
      "enable": [
        "ThisNetWorks.OrchardCore.CKEditor"
      ],
      "disable": [
      ]
    }
  ]
}
```

This happens because the `IRecipeExecutor` is a Singleton, and assigns the `variableMethodsProvider` to a private variable as it reads the `recipe.json`, so on the second recipe run it overwrites the value.

This should be easy to fix, probably with a dictionary wrapping the `variablesMethodsProviders` for each recipe, keyed to the execution id, which is always a new guid.

Hmm, does it need to be concurrent? 

@jtkech you are fixing some recipe issues at the moment, but I think this is unrelated / a different issue.

jtkech:
Fixes #6717 and refactoring

- Fixes the related issue by holding in a dictionary additional method providers per recipe executionId. @deanmarcussen i made the recipe executor transient, not to fix the issue but to be thread safe even if we use here a simple dictionary.

- Here the singularity is that the recipe executor creates a new scope on each step that may be based on a new shell / container, e.g. after updating the features, so resolving a service in a given step scope, even it is a tenant singleton, may return another instance (including the recipe executor itself if you try it).

    That's why we need to resolve the script manager in each step scope, and we need that the recipe executor only injects e.g. app level transients / singletons as the shell host, that's okay for the shell settings that stays the same as a recipe is not intended to reload it, only externally after its execution e.g. at the end of a setup.

- But **i'm not sure that the injection and the usage of the `IRecipeEventHandler` are good**, they would need to be app level transients / singletons. Maybe at least tenant level transients / singletons, and then okay for the event at the beginning but they would need to be part of the setup features. But at the end of a recipe, maybe better to be resolved in a new scope that may be based on a new shell, as we do at the end of a setup for the `ISetupEventHandler`.

    They also have step events called for each step, here maybe better to resolve them in each step scope. Notice that here they were called each time we call a scoped `IRecipeStepHandler`, fixed here by calling them outside the related loop. **So i'm tempted to remove them, we have no concrete implementations and i think there are not used or very rarely**.

- In the recipes admin controller, we now release the tenant after a recipe execution, e.g if they have changed some site settings that needs to release the tenant so that it will be rebuilt. If features have changed it is automatically done through the shell descriptor manager, at the end of a setup it is done somewhere when updating shell settings.

Need more thinkings about the need / usage of `IRecipeEventHandler` (not `IRecipeStepHandler`), but @deanmarcussen as it is already done, when you will have time can you try it to check if it fixes the usage of your inner recipe.

**Update**: That said there is still an unit test that is failing, will see tomorrow

### DIFF
--- a/test/OrchardCore.Tests/Apis/ContentManagement/DeploymentPlans/BlogPostValidateDeploymentPlanTests.cs
+++ b/test/OrchardCore.Tests/Apis/ContentManagement/DeploymentPlans/BlogPostValidateDeploymentPlanTests.cs
@@ -46,17 +46,15 @@ namespace OrchardCore.Tests.Apis.ContentManagement.DeploymentPlans
                 Assert.Throws<HttpRequestException>(() => response.EnsureSuccessStatusCode());
 
                 // Confirm creation of both content items was cancelled.
-                using (var shellScope = await BlogPostDeploymentContext.ShellHost.GetScopeAsync(context.TenantName))
+                var shellScope = await BlogPostDeploymentContext.ShellHost.GetScopeAsync(context.TenantName);
+                await shellScope.UsingAsync(async scope =>
                 {
-                    await shellScope.UsingAsync(async scope =>
-                    {
-                        var session = scope.ServiceProvider.GetRequiredService<ISession>();
-                        var blogPosts = await session.Query<ContentItem, ContentItemIndex>(x =>
-                            x.ContentType == "BlogPost").ListAsync();
+                    var session = scope.ServiceProvider.GetRequiredService<ISession>();
+                    var blogPosts = await session.Query<ContentItem, ContentItemIndex>(x =>
+                        x.ContentType == "BlogPost").ListAsync();
 
-                        Assert.Single(blogPosts);
-                    });
-                }
+                    Assert.Single(blogPosts);
+                });
             }
         }
     }
--- a/src/OrchardCore.Modules/OrchardCore.Recipes/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Recipes/Controllers/AdminController.cs
@@ -19,6 +19,7 @@ namespace OrchardCore.Recipes.Controllers
 {
     public class AdminController : Controller
     {
+        private readonly IShellHost _shellHost;
         private readonly ShellSettings _shellSettings;
         private readonly IExtensionManager _extensionManager;
         private readonly IAuthorizationService _authorizationService;
@@ -29,6 +30,7 @@ namespace OrchardCore.Recipes.Controllers
         private readonly IHtmlLocalizer H;
 
         public AdminController(
+            IShellHost shellHost,
             ShellSettings shellSettings,
             ISiteService siteService,
             IExtensionManager extensionManager,
@@ -38,6 +40,7 @@ namespace OrchardCore.Recipes.Controllers
             IRecipeExecutor recipeExecutor,
             INotifier notifier)
         {
+            _shellHost = shellHost;
             _shellSettings = shellSettings;
             _siteService = siteService;
             _recipeExecutor = recipeExecutor;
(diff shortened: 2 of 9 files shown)


deanmarcussen:
@jtkech tried, and it fixes the inner recipe issue.

Was thinking about thread safety as well when I suggested a transient here, but I see the dilemma with the `IRecipeEventHandler` 

Need more thinking :)

jtkech:
> @jtkech tried, and it fixes the inner recipe issue.

Okay cool

I fixed an unit test that was expecting an exception that was no more thrown because i was using our `InvokeAsync()` extension in place of a foreach loop.

I kept the `IRecipeEventHandler` as is to not break anything, but knowing that, as they are used, they would need to be e.g. app level singletons, not tenant level singletons, otherwise they may be used in a shell container they don't belong.

deanmarcussen:
> I kept the `IRecipeEventHandler` as is to not break anything, but knowing that, as they are used, they would need to be e.g. app level singletons, not tenant level singletons, otherwise they may be used in a shell container they don't belong.

Cool, probably for the best, someone will be using them.

How about a comment on their interface that they must be registered as app level singletons?

(if not for someone else, just for us, to remember / know next time)

test/OrchardCore.Tests/Apis/ContentManagement/DeploymentPlans/BlogPostValidateDeploymentPlanTests.cs
```diff
@@ -46,17 +46,15 @@ public async Task ShouldFailWhenAutoroutePathIsNotUnique()
                 Assert.Throws<HttpRequestException>(() => response.EnsureSuccessStatusCode());
 
                 // Confirm creation of both content items was cancelled.
-                using (var shellScope = await BlogPostDeploymentContext.ShellHost.GetScopeAsync(context.TenantName))
+                var shellScope = await BlogPostDeploymentContext.ShellHost.GetScopeAsync(context.TenantName);
```
deanmarcussen, line 49: Interesting @jtkech was this my bad usage of `shellScope` by wrapping it in a using block (it shouldn't have one?)

jtkech: @deanmarcussen You don't need it as `UsingAsync()` will do more and at the end will also dispose the scope asynchronously, but it would have no impact to use an using block, it would just try to dispose an already disposed scope.

        /// <summary>
        /// Execute a delegate using this shell scope.
        /// </summary>
        public async Task UsingAsync(Func<ShellScope, Task> execute)
        {
            if (Current == this)
            {
                await execute(Current);
                return;
            }

            using (this)
            {
                StartAsyncFlow();

                await ActivateShellAsync();

                await execute(this);

                await BeforeDisposeAsync();

                await DisposeAsync();
            }
        }

Issue #6548, opened 2020-06-30
Dynamic Cache Tag Helper conflicts with ASP.NET Core cache tag helper


deanmarcussen:
There is an issue with the dynamic cache tag helper, where by default it conflicts with the ASP.NET Core Tag Helper, and both are activated, meaning the inner content is built by the ASP.NET Core tag helper, stored in it's memory cache, and then passed of to the dynamic cache tag helper, which then stores it, in it's cache.

This is because they are both targeting the `cache` element, but it means the `cache-dependency` doesn't have a scope available in the scope manager, as it is being rendered by the ASP.NET Core cache.

Removing the ASP.Net Core one fixes the issue
@removeTagHelper Microsoft.AspNetCore.Mvc.TagHelpers.CacheTagHelper, Microsoft.AspNetCore.Mvc.TagHelpers

So we could either just update the docs to say you have to remove the ASP.NET Core tag helper, or consider renaming the tag helper `dynamic-cache` perhaps?

I think clarifying the name would be a bit breaking, but a better solution long term.

@jtkech was able to repro as well, opening this issue so we can triage it

Related pr which identified the issue https://github.com/OrchardCMS/OrchardCore/pull/6518

sebastienros:
New name

Two hard things, caching and naming

hishamco:
What if we prefix all Orchard Core TagHelper using `oc` to avoid such collision?

Skrypt:
I experienced this issue. Seems like the CacheTagHelper needs to be removed still in the _ViewImport file so that it doesn't collide with the default cache tag.

Also, a bit of documentation would help.

```C#
<dynamic-cache cache-id="dealers" expires-after="@TimeSpan.FromHours(1)">
    @foreach (var item in dealers)
    {
        <div>@item</div>
        <cache-dependency dependency="contentitemid:@item.ContentItemId"/>
    }
    <cache-dependency dependency="contentitemid:@item.ContentItemId"/>
<dynamic-cache>
```

PR #6737, opened 2020-07-21
Rename dynamic cache tag helper

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6548

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.DynamicCache/TagHelpers/DynamicCacheTagHelper.cs
+++ b/src/OrchardCore.Modules/OrchardCore.DynamicCache/TagHelpers/DynamicCacheTagHelper.cs
@@ -12,7 +12,7 @@ using OrchardCore.Environment.Cache;
 
 namespace OrchardCore.DynamicCache.TagHelpers
 {
-    [HtmlTargetElement("cache", Attributes = CacheIdAttributeName)]
+    [HtmlTargetElement("dynamic-cache", Attributes = CacheIdAttributeName)]
     public class DynamicCacheTagHelper : TagHelper
     {
         private const string CacheIdAttributeName = "cache-id";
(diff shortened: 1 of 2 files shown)


hishamco:
Still I see a value of using `oc` as TagHelper prefix that can be configured into `_ViewImports.cshtml` not as a part of TagHelper name, this will avoid the name conflict from Microsoft and other 3rd parties TagHelpers. But I will leave this open to the community to decide

jtkech:
@deanmarcussen 

Here the singularity is that the mvc `cache` tag helper can be selected only by its tag name without any attribute.

**So i suggest `distributed-cache`** because the mvc one needs a `name` attribute, and our one a `cache-id` attribute.

And also because our tag helper uses `IDistributedCache`.

deanmarcussen:
> **So i suggest `distributed-cache`** because the mvc one needs a `name` attribute, and our one a `cache-id` attribute.
> 
> And also because our tag helper uses `IDistributedCache`.

It's kinda wordy? But so is `<dynamic-cache>` But I'm not hearing any better offers from anyone...

@hishamco The idea is to not have to configure anything special in `_ViewImports` (because how will people know they have to - it's been three years of having this tag before we realised it conflicted). So I don't see how having to use a prefix fixes anything.

hishamco:
Three years!! that's too long ;) the tag prefix will not change the name, we can keep the tag name as it was, so no breaking changes. It just a token that can differentiate our tag helpers from other 3rd-party once and avoid us from thinking about new names whenever create a tag helper and AspNetCore has similar one. It is something like old WebForms controls: `asp:Button`, `asp:Label`, IMHO the prefix will add a value nothing else

deanmarcussen:
Maybe I am misunderstanding the prefix suggestion. 

It is something the user has to add to their `_ViewImports` file? so that taghelpers from a particular  assembly are prefixed?

So doesn't prevent the collision, just provides another way of managing them.

If I have misunderstood, show an example?

hishamco:
> It is something the user has to add to their _ViewImports file?

Exactly!!

> So doesn't prevent the collision

Why?!! if we prefix our tag helpers in all assemblies, no collision will happen, because we should write `oc:` first, then the VS intellisense will recognize them

jtkech (approved):

PR #6739, opened 2020-07-22
Document link tag helper and improve Resources docs

### LINKED ISSUE #6740: Script helper no longer registering script at Foot by default
The documentation says that scripts using the tag helper will render in the FootScript by default if no “at” is declared however I noticed instead scripts are being rendered in the head section by default.

Is the documentation wrong or the code wrong?

scleaver:
Have been using the link tag helper lately and noticed there is no documentation for it. Add samples and made a few small improvements to the resources docs.

Noticed a few issues with the link tag functionality and script tag but will register these as issues.

Also fixes #6740 as discussed as this is related.


src/docs/reference/modules/Resources/README.md
```diff
@@ -365,6 +369,23 @@ The style block will only be injected once based on its name and can optionally
     }
 </style>
 ```
+#### Link tag
+
+A link tag is used to define the relationship between the current document and an external resource such as a a favicon or stylesheet. For a stylesheet, however, use the [style helper](#register-a-named-script). 
+
+``` liquid tab="Liquid"
+{% link rel:"icon", type:"image/png", sizes:"16x16", src:"~/media/favicon/favicon-16x16.png" %}
```
deanmarcussen, line 377: Cool, but don't use this `~/media/favicon/favicon-16x16.png` to get to the media library folders, as it doesn't take into account any tenant prefix.

`~/TheTheme` the `~` prefix will get you to a static file hosted in a `wwwroot` folder, but not a media folder, which is stored in the `App_Data` (so would work, but not across tenants)

For accessing the media library with any of the script / link tag helpers you would have to use the `@Html.AssetUrl` helper inside the `src` attribute. And I'm not sure how we would do this with liquid, it maybe something we are missing.

We might need to target `asset-src` on the `LinkTagHelper` and map the public url to the source url.

scleaver: @deanmarcussen no problem I can change these urls.

I didn't want to use static assets in my theme for favicon as you would want to change this if the theme is used by more than one tenant. Ideally you would have this in a custom setting but in my case I was under the pump to get it done so I was taking shortcuts.

I am going to raise an issue actually as I found a couple of problems in the link tag helper:

- The liquid asset_url helper which I wanted to use in my case to refer to a media file doesn't work inside the link tag helper
- the link tag helper works from within liquid field, eg. from a layer, but if the tag is in a static liquid template it doesn't work - no idea why
- The razor version of the helper doesn't work at all actually if my code example is right. The example in the documentation I added just renders the link exactly as is in the exact spot it is placed within a template. Now that I am writing this i am thinking it might be that I was missing the using statement - will check
- link helper doesn't seem to support named resources either which would be handy.

I will just check whether I had the using and will log these issues.

scleaver: I have tested the razor helper again and the issue was due to the helper reference missing.  I also re-tested the liquid helper and now the only outstanding issue is that you can't use the asset_url helper.

scleaver:
Ready to merge now I believe.

scleaver:
@deanmarcussen in regards to your comment about the asset URL in liquid I worked out you can do this. The reason you have to do it like this makes sense to me... it could be addressed at a later time so it is simplier but it's ok for now I think. I might add it to the documentation to help others.

```
{% assign image_url = 'favicon/favicon-16x16.png' | asset_url %}
{% link rel:"icon", type:"image/png", sizes:"16x16", src:image_url %}
```

src/docs/reference/modules/Resources/README.md
```diff
@@ -365,6 +369,35 @@ The style block will only be injected once based on its name and can optionally
     }
 </style>
 ```
+#### Link tag
+
+A link tag is used to define the relationship between the current document and an external resource such as a a favicon or stylesheet. For a stylesheet, however, use the [style helper](#register-a-named-script).
```
deanmarcussen, line 374: a a favicon 

double a a

I can't edit this, or would have, but other than that looks good

src/docs/reference/modules/Resources/README.md
```diff
@@ -232,14 +236,14 @@ You can append a version hash that will be calculated, and calculation cached, a
 
 ##### Specify location
 
-By default all scripts are rendered in the footer. You can override it like this:
+By default all scripts are rendered whereever the `FootScript` helper is located [See Foot Resources](#foot-resources). You can override it like this:
```
agriffard, line 239: whereever => wherever

scleaver: Thanks @agriffard - I changed the wording here based on the outcome of #6740

deanmarcussen:
@agriffard if you are happy with the changes LGTM

PR #6748, opened 2020-07-22
initialize TheAgencyTheme npm packages

kevinsmgov:
The root package.json did not have an install for TheAgencyTheme project like it did for the other bootstrap themes.


agriffard:
It has been removed. 
It is not needed anymore.
The styles and scripts are taken from the dist folder.

kevinsmgov:
So, just for my understanding, we're no longer using _task runner_ for less, scss, and typescript builds when modifying themes?

The reason I ask is that it crashed when I tried a build because the theme "The Agency" hadn't had its npm packages installed (I had to manually install them in the theme in order for the tool to work). 

If not, what's the technique for rebuilding bootstrap with different variable values?

agriffard:
Ok, you had an error because bootstrap-oc.scss has a reference to node_modules/bootstrap.
We may need a package.json below the project, similar to the one in TheTheme.

agriffard:
Replaced by #6803

PR #6516, opened 2020-06-26
Fix inline add widget control not using contained content types setting

peterkeating:
Flows editor can be configured to only display a defined collection of widgets (https://github.com/OrchardCMS/OrchardCore/pull/5970). While the add widget control at the bottom of the flow editor respects the contained content type setting, the inline one was displaying all content types with stereo type of widgets.

PR #6750, opened 2020-07-22
Showing Content Types as per ContainedContentType settings for FlowParts. 

### LINKED ISSUE #6276: Content Types with restricted FlowParts are only enforced the first time
We have created a custom ContentType with the following parts:
* Title
* Autoroute
* Flow
* Location

We have also restricted the FlowPart so that it can only contain *Html* ContentType instances.

When we create a new content of this new type, the available types to use are in fact restricted to just *Html*.

Howerver, after having inserted the first (restricted) one, if we try to further add more widgets to that very same content, they are not restricted: all of them are available.

We are using OC 1.0.0-rc1-13168. Can someone confirm or are we doing something wrong?

ns8482e:
Fixes #6276 

- Refactored code to have single Insert shape for FlowPart, BagPart, WidgetsListPart.
- Added Validation for restricted contained types for BagPart, FlowPart.

Split into smaller PR from #6298 

/cc @deanmarcussen @peterkeating


deanmarcussen:
I'm surprised to see another shape being made with `New()` when the behaviour is driven by 3 different class types and it could have been produced by a driver in a zone

ns8482e:
> I'm surprised to see another shape being made with `New()` when the behaviour is driven by 3 different class types and it could have been produced by a driver in a zone

I hear ya.  - I was being lazy and moved code from other PR as is - as the shape properties variables used for `New()` are only available  inside `Parts` template for all three parts.

agriffard:
Please fix the conflict.

agriffard:
@deanmarcussen Can you please review this PR?

deanmarcussen:
Same comment as before, I'm not really keen on introducing a new shape type like this.

I prefer @peterkeating solution not introducing a shared shape type, until the flow part edit code and content card shapes would be refactored to use a driver based solution

deanmarcussen:
I also don't like the idea of introducing a forced check in the drivers when editing an item, to ensure it only contains the allowed content types.

This prevents an administrator allowing a special content type, when setting something up for a user/editor, and then disabling entry of new items of this type.

ns8482e:
> Same comment as before, I'm not really keen on introducing a new shape type like this.
> 
> I prefer @peterkeating solution not introducing a shared shape type, until the flow part edit code and content card shapes would be refactored to use a driver based solution

The new shape I am adding is in Part's shape - and Driver based solution adds shape to `Content` not `ContentPart`. It's same reason as  `ShortCode` shape in other edit parts e.g. below.

https://github.com/OrchardCMS/OrchardCore/blob/140c5485f995c083b20fb83db7427d7fa30db6f0/src/OrchardCore.Modules/OrchardCore.ContentFields/Views/HtmlField-Multiline.Edit.cshtml#L8

ns8482e:
> I also don't like the idea of introducing a forced check in the drivers when editing an item, to ensure it only contains the allowed content types.
> 
> This prevents an administrator allowing a special content type, when setting something up for a user/editor, and then disabling entry of new items of this type.

It's validation based on settings selected by `Admin`, I guess when admin set's restriction on content types to be selected then it better to have validation available out-of-the-box.

ns8482e:
Closing this as other pr  #6516 is merged

PR #6753, opened 2020-07-22
 Add new property settings to OrchardCore.Mvc.Web project to address Microsoft.AspNetCore.Antiforgery error

### LINKED ISSUE #6752: Cannot find reference assembly 'Microsoft.AspNetCore.Antiforgery.dll'
When running the OrchardCore.Mvc.Web sample project there is an error:

InvalidOperationException: Cannot find reference assembly 'Microsoft.AspNetCore.Antiforgery.dll' file for package Microsoft.AspNetCore.Antiforgery

scleaver:
Fixes #6752 - solution attained from [this post](https://github.com/toddams/RazorLight/issues/294)

### DIFF
--- a/src/OrchardCore.Mvc.Web/OrchardCore.Mvc.Web.csproj
+++ b/src/OrchardCore.Mvc.Web/OrchardCore.Mvc.Web.csproj
@@ -5,11 +5,7 @@
   <PropertyGroup>
     <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework>
     <AspNetCoreHostingModel>InProcess</AspNetCoreHostingModel>
-  </PropertyGroup>
-
-  <PropertyGroup>
     <PreserveCompilationReferences>true</PreserveCompilationReferences>
-    <PreserveCompilationContext>true</PreserveCompilationContext>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/OrchardCore.Mvc.Web/OrchardCore.Mvc.Web.csproj
+++ b/src/OrchardCore.Mvc.Web/OrchardCore.Mvc.Web.csproj
@@ -1,10 +1,15 @@
 ﻿<Project ToolsVersion="15.0" Sdk="Microsoft.NET.Sdk.Web">
 
   <Import Project="..\OrchardCore.Build\Dependencies.props" />
-  
+
   <PropertyGroup>
     <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework>
     <AspNetCoreHostingModel>InProcess</AspNetCoreHostingModel>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <PreserveCompilationReferences>true</PreserveCompilationReferences>
+    <PreserveCompilationContext>true</PreserveCompilationContext>
   </PropertyGroup>
 
   <ItemGroup>


scleaver:
Quick consolidation of propertygroup - should be ready to merge.

jtkech:
@scleaver 

I can't repro, normally here it's up to the user to use `PreserveCompilationReferences`

We set it to true in `OrchardCore.Cms.Web` because we don't support precompiled application level **shape** views, so it allows to compile these app level **shape** views at runtime (and they need to be published).

But that's okay for precompiled app level mvc views

scleaver:
@jtkech

Maybe it is just an error on Mac OS? As soon as I put in the `PreserveCompilationReferences` this fixed the problem. Perhaps this is just masking the real issue. I will spin up my Windows VM and see if I can reproduce it there.

jtkech:
Okay cool, better to check if there is not another issue, otherwise i would not be closed to have it by default as in Cms.Web  ;)

deanmarcussen:
For info @jtkech I could see it on mac as well (not tried pc)

Screenshot 2020-07-24 at 11 20 46

But someone @scleaver should try the https://github.com/OrchardCMS/OrchardCore.Samples module as well, as we don't seem to need it there, and maybe there is another issue

jtkech:
Okay then to have it by default, maybe at the end of the existing property group as in `Cms.Web`.

The trace confirms that a runtime compilation is done, it's like it doesn't find the precompiled views in this context, can you check that on building `OrchardCore.Mvc.HelloWorld.Views.dll` and `OrchardCore.Mvc.Web.Views.dll` are outputed.

Do you see it in dev mode and / or from the command line and / or in a published context?

Maybe a problem with the `Microsoft.NET.Sdk.Web` sdk under mac OS, or if you are not using Visual Studio maybe because of another tool, e.g. `OmniSharp` under VS Code

pranavkm:
> Do you see it in dev mode and / or from the command line and / or in a published context?

That would help. If you're seeing this in when running from a CLI, 

a) Does a `refs` folder exist in the `bin/Debug/net5.0` directory?
b) If it does not, are you able to share msbuild binary logs from building the project? If you run `dotnet build /bl` a `msbuild.binlog` should be generated in the directory.

sebastienros:
it's netcoreapp3.1

deanmarcussen:
> Do you see it in dev mode and / or from the command line and / or in a published context?

Was just building from command line `Debug`, `Development` to test and see if I could repro

The `Refs` folder exists, but is empty

SDK Version `3.1.302`

ASP.NET Core Version `3.1.6`

The `OrchardCore.Mvc.HelloWorld.Views.dll` exist, in the root of the `bin` folder, not `refs`.

Build logs attached thanks for taking a look @pranavkm 

On Windows, same `dotnet run` works fine, and all the framework `refs` are in the `refs` folder.

We need to get you a Mac @jtkech !

[msbuild.binlog.zip](https://github.com/OrchardCMS/OrchardCore/files/5005669/msbuild.binlog.zip)

jtkech:
> We need to get you a Mac @jtkech !

And too lazy to emulate it on my old machine ;)

> The Refs folder exists, but is empty

Yes we check it in some places e.g. before registering runtime compilation services, regarding our code, one possibility was to be in `IsDevelopment()` and having a `refs` folder but with missing compilation assemblies. So it may explain the issue, if so we would have to check if the `refs` folder exists and if it is not empty.

@deanmarcussen i will do a PR so that you can re-try it on your mac.

deanmarcussen:
> No, if we don't specify `PreserveCompilationReferences` (as it is currently in Mvc.Web.csproj) that is false by default with the `Sdk.Web`, on my Windows it doesn't output any refs folder, and it works as it doesn't trigger a runtime compilation.

I'll fire up bot the mac and windows vm and send you some gifs tomorrow as you are too much the night owl to skype :)

deanmarcussen:
Mac and Windows builds @jtkech with `bin` and `obj` folders deleted before building

using `dotnet build /bl` on both.

SDK 3.1.302 on both

[mac.zip](https://github.com/OrchardCMS/OrchardCore/files/5015345/mac.zip)
[windows.zip](https://github.com/OrchardCMS/OrchardCore/files/5015348/windows.zip)

Windows has ` refs` folder, mac doesn't

jtkech:
@deanmarcussen thanks for the zip files

First, we always reference the `RuntimeCompilation` package transitively as we need it e.g. to update the relative paths of precompiled views at runtime. But we exclude the execution of its targets file by using `ExcludeAssets="build"` so that people can do what they want at the app level e.g. by referencing it directly or by just setting some properties as we do.

At some point AspnetCore also added this targets file in the `buildTransitive` folder, it was still working for us but the `refs` was always outputed even if we don't want it by specifying `PreserveCompilationReferences = false` in the app project file.  That's why i did a recent change to also excludes this folder by using `ExcludeAssets="build;buildtransitive"`.

---

**windows.zip**

- Okay it works but indeed in the log file we can see that `RuntimeCompilation.targets` is still executed and that the `refs` folder is created and populated as the related property has been set to true, which is not the case on my machine.

    **Property reassignment**: $(PreserveCompilationReferences)="true" (previous value: "false") at C:\Users\dean\.nuget\packages\microsoft.aspnetcore.mvc.razor.runtimecompilation\3.1.4\buildTransitive\netcoreapp3.1\Microsoft.AspNetCore.Mvc.Razor.**RuntimeCompilation.targets** (4,5)1

    Here maybe a case sensitive issue, in `ExcludeAssets="build;buildtransitive"` we may need to use `buildTransitive"` in place of `buildtransitive`. On your windows VM maybe the underlying file system is still case sensitive. Anyway better to use `buildTransitive` that also works on my machine. **I will push this change so that you can re-try it**.

---

**mac.zip**

- Okay, nothing is outputed in a `refs` folder which is not created. You were talking about an empty `Refs` folder that could explain the issue but in your `after.txt` there is no `refs` folder, so here the question is **After this new build, did it still fail when running?** I can repro if, before running the app (already built) in dev mode, i create an empty `refs` folder. So, maybe here you just needed to remove this empty folder

jtkech:
@scleaver @deanmarcussen sorry for the too long comments ;)

The  macOS log doesn't create and populate the `refs` folder, that's okay because by default (without the PCRefs property) we don't want runtime compilation. What was confusing is that the windows log creates and populates the `refs` folder, but here also that's okay because you were using another OC version before a recent change fixing it.

**So for me both logs are okay.**

That said @deanmarcussen you were talking about an empty `refs` folder, and i can repro if before running the app in dev mode, i create an empty refs folder. So my question is, after your cleaning and rebluid did it work on macOS?

@scleaver when it fails, can you check if an empty `refs` folder exists? If so can you delete it and see if it works, or if an empty refs folder is created again?

deanmarcussen:
@jtkech sorry busy day yesterday and this slipped under my radar to recheck.

Yes, old version on Windows (my bad didn't pull dev when I started it)

Probably also an old version on the mac had created the `refs` folder (so many branches, not sure which one I was on when I first tested this)

Can confirm that with no `refs` folder it works fine on the mac, and that builds of current dev are not creating an empty one

So I'm guessing the change that "fixed" it was the addition of `buildtransitive` a few weeks ago?

jtkech:
> @jtkech sorry busy day yesterday and this slipped under my radar to recheck.

@deanmarcussen no problem ;)

> Can confirm that with no refs folder it works fine on the mac, and that builds of current dev are not creating an empty one

Cool, so for me at this point **i would say that there is no issue**

> So I'm guessing the change that "fixed" it was the addition of buildtransitive a few weeks ago?

Yes, it prevents from having a populated `refs` folder if we don't want runtime compilation. But it doesn't explain the presence of an **empty** `refs` folder. Maybe it was there and then after the fix it was emptied but the folder itself was not successfully deleted. Note: Maybe i will do a PR to not only check that the `refs` folder exists but also that it is not empty

deanmarcussen:
> Cool, so for me at this point **i would say that there is no issue**

I think you're right, **there is no issue**

But now I can't make it fail, even with an empty `refs` folder.

too many branches involved on my repo, so I can't be sure which one I was on when it was failing, or that I had the change to `buildtransitive` on the branch I was testing with.

I will revert that change on my current branch today, just to see. 

But it seems like we are chasing ghosts. Maybe @scleaver could retry with current dev as well

jtkech:
> But now I can't make it fail,

Okay

> even with an empty refs folder.

I can repro with an empty `refs` folder AND `IsDevelopment = true`, maybe not the case here.

So i close the issue, @scleaver re-open it if you still have an issue with the `OC.Mvc.Web` project by using the last dev and on a clean solution

PR #6754, opened 2020-07-23
Default location of scripts should be at Foot

### LINKED ISSUE #6740: Script helper no longer registering script at Foot by default

scleaver:
Fixes #6740


scleaver:
Discussions in #6740 decided to update the documentation rather than apply this change. Will fix this in another commit.

PR #6755, opened 2020-07-23
Removed unused variables from few files

paragpkulkarni:


src/OrchardCore.Modules/OrchardCore.AdminMenu/Controllers/NodeController.cs
```diff
@@ -39,8 +39,7 @@ public class NodeController : Controller
             _displayManager = displayManager;
             _factories = factories;
             _adminMenuService = adminMenuService;
-            _authorizationService = authorizationService;
-            New = shapeFactory;
+            _authorizationService = authorizationService;
```
deanmarcussen, line 42: Thanks @paragpkulkarni clean up is good.
Here, and for the others, I think you could remove them from the constructor as well, as if they are not being used, they do not need to be resolved.

src/OrchardCore.Modules/OrchardCore.CustomSettings/Recipes/CustomSettingsStep.cs
```diff
@@ -13,15 +12,12 @@ namespace OrchardCore.CustomSettings.Recipes
...
     public class CustomSettingsStep : IRecipeStepHandler
     {
-        private readonly ISiteService _siteService;
-        private readonly CustomSettingsService _customSettingsService;
+        private readonly ISiteService _siteService;        
 
         public CustomSettingsStep(
-            ISiteService siteService,
-            CustomSettingsService customSettingsService)
+            ISiteService siteService)
```
hishamco, line 18: I think we can move the param inline instead of two lines

src/OrchardCore.Modules/OrchardCore.AdminMenu/Controllers/NodeController.cs
paragpkulkarni: Removed IShapeFactory shapeFactory from constructor as per the comment

src/OrchardCore.Modules/OrchardCore.CustomSettings/Recipes/CustomSettingsStep.cs
paragpkulkarni: Constructor parameters are inline now.

src/OrchardCore.Modules/OrchardCore.BackgroundTasks/Controllers/BackgroundTaskController.cs
```diff
@@ -43,8 +42,7 @@ public class BackgroundTaskController : Controller
             _backgroundTasks = backgroundTasks;
             _backgroundTaskManager = backgroundTaskManager;
             New = shapeFactory;
-            _siteService = siteService;
-            _notifier = notifier;
+            _siteService = siteService;
```
deanmarcussen, line 45: You could also remove from the constructor in this controller

paragpkulkarni: Resolved notifier object from constructor

Issue #6756, opened 2020-07-23
ContentManager should return null instead of throwing exception


deanmarcussen:
A change was made a while back to the `ContentManager.GetAsync()` method to throw an exception if the supplied `contentItemId` was null or empty.

I like the principle, but it's causing too much boilerplate code in my apps, and it should noop returning null, if no id is supplied.

Consider :
```
if(!String.IsNullOrEmpty(contentItemId))
{
  var contentItem = await _contentManager.GetAsync(contentItemId);
  if (contentItem != null)
  {
     ... do stuff
  }
}
```

Versus :
```
  var contentItem = await _contentManager.GetAsync(contentItemId);
  if (contentItem != null)
  {
     ... do stuff
  }
```

even our `ItemController` is having to do two checks to avoid a null reference exception
```
            if (contentItemId == null)
            {
                return NotFound();
            }

            var contentItem = await _contentManager.GetAsync(contentItemId, jsonPath);

            if (contentItem == null)
            {
                return NotFound();
            }
```

hishamco:
Agree for this change

PR #6757, opened 2020-07-23
Return null instead of exception from ContentManager.GetAsync

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6756


deanmarcussen:
> LGTM, just one comment is this [line](https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore/OrchardCore.ContentManagement.GraphQL/Queries/ContentItemQuery.cs#L52) is safe?

No safer or less by this change. I imagine GraphQL would just return a null result, which you would expect, but you could check that if you like

Issue #5409, opened 2020-01-30
User accounts improvements


sebastienros:
I'd like to revive this one:

Right now, is it possible to create a new account with an email that already exists in the database?
1. I think we should support being able to log with a username or an email, and it might just be the only option
2. I don't think we should support being able to log with only the email or only the username (i.e. prevent the other one) as long as we don't expose if the entry exists in the db when the login fails, to prevent lookups
3. Unicity:
usernames should be unique (and case insensitive to prevent case spoofing)
emails should be unique (and case insensitive to prevent case spoofing)
4. Usernames and Emails should be unique, as a whole, such that an existing email can't be used as a new username
5. Like this PR we should be able to enforce the email to be valid before enabling an account, as a setting (in the admin)
6. Usernames should not be changed if they are used to represent a user reference (they are). Or we should introduce a new identity for users and use it, and also store usernames for historical data, not reference integrity (like we need to for tenants)
7. Emails can be changed as they are not used as user identity (and they shouldn't, because they can be changed)
8. I don't see why the username and the email should not have the same value, even if they can be changed individually, as long as the value doesn't collide with another existing user. There should then be no validation on the username, but validate the email.

sebastienros:
c.f. https://github.com/OrchardCMS/OrchardCore/pull/2744

agriffard:
If username can be changed, we need to apply the changes to the content items that have it as owner.

Skrypt:
I think what @sebastienros means is that the Owner should be kept for history but we should also have an immutable ID to identify the real Owner instead. That way, no need to change the Owner when the username is changed.

sebastienros:
The identifier of a user account should not be editable

1. For security reasons, as if an account is deleted, another account could use a previous identifier to impersonate technical permissions.
2. For data integrity as an identifier is used for referencing to an account, and that would break any reference (e.g. link to user, owner property in content item)
3. For idempotency of import/export such that two systems share a logical identifier that is used to migrate data. Hence an auto-increment integer like the document id should not be used as a unique identifier.

The identifier of a user account should be random, and not entered by a user

To prevent accounts from being impersonated in case an identifier was leaked. So even at the time of the creation of an account, the identifier should be computer generated, or have a part that is computer generated.
The identifier should be normalized such that its usage is case insensitive, and any attempt to use a different case will match an existing account. For instance all values should be stored and looked up with a defined char case.

Implications

- We can't use the UserName as an identifier, as it is provided by user input
[...]

PR #6759, opened 2020-07-23
Implement UserId

deanmarcussen:
Related to https://github.com/OrchardCMS/OrchardCore/issues/5409

This is a start towards having a `UserId` throughout the OC ecosystem.

It implements a `UserId` property for `Users` 

From the issue
> The identifier of a user account should not be editable

Done in this pr

> Implications

> We can't use the UserName as an identifier, as it is provided by user input

Done in this pr
> We need to create a new property to store the identifier

Done in this pr
> We can use the existing usernames as identifiers for existing accounts provided we can't edit identifier anymore. That will prevent a data migration.

Done in this pr (see notes on where this maybe an issue)
> We need to understand what User.Identity.Name represents and how it's used relative to the username property, which was used an an identifier as of now.

In the context of permissions, this line from the `ContentTypeAuthorizationHandler`
`user.Identity.Name == content.Owner;`

This should related to the `Name` `Claim` property on claims.
Nothing in this pr changes that.
The `UserId` (and previously `DocumentId`) are mapped to the `NameIdentifier` claim, ref: https://docs.microsoft.com/en-us/dotnet/api/system.security.claims.claimtypes.nameidentifier

User.Identity.Name should map to the `Name` claim, ref: https://docs.microsoft.com/en-us/dotnet/api/system.security.claims.claimtypes.nameidentifier

So for this pr nothing changes in relation to content ownership (next pr).

> We should not use the document id anymore but the new identifier.

Done in this pr

The idea here is to introduce the `UserId` property on the `IUser` as a guid / generated id, and later consider this in the context of content item ownership.

I suspect we will want to consider storing both the `Owner` property as a historical value / friendly name for the ui, and an `OwnerId` which would map to the `Claim.NameIdentifier` -> `UserId`


agriffard:
Please fix the conflicts.

Skrypt:
What does it breaks ? 🥰

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
```diff
@@ -110,13 +120,8 @@ public async Task<IdentityResult> DeleteAsync(IUser user, CancellationToken canc
...
         {
-            int id;
-            if (!int.TryParse(userId, out id))
-            {
-                return null;
-            }
-
-            return await _session.GetAsync<User>(id);
+            // For backwards compatability we check also check the UserId against the NormalizedUserName.
+            return await _session.Query<User, UserIndex>(u => u.UserId == userId || u.NormalizedUserName == userId).FirstOrDefaultAsync();
```
deanmarcussen, line 124: For when we review this agin, this is one bit which concerns me, but I think it is ok

Skrypt:
Can't we create simply a SQL migration that would generate UID's for all the current existing user accounts?

src/OrchardCore.Modules/OrchardCore.Users/Migrations.cs
```diff
@@ -6,11 +6,22 @@ namespace OrchardCore.Users
...
     public class Migrations : DataMigration
     {
+        // This is a sequenced migration which on new schemas is complete after UpdateFrom2.
         public int Create()
         {
             SchemaBuilder.CreateMapIndexTable<UserIndex>(table => table
                 .Column<string>("NormalizedUserName")
                 .Column<string>("NormalizedEmail")
+                .Column<bool>("IsEnabled", c => c.NotNull().WithDefault(true))
+                .Column<string>("UserId")
```
Skrypt, line 16: Length = 26 chars

src/OrchardCore.Modules/OrchardCore.Users/Migrations.cs
```diff
@@ -49,5 +62,18 @@ public int UpdateFrom3()
 
             return 4;
         }
+
+        public int UpdateFrom4()
+        {
+            // The length is the default as for backwards compatability this value will be the NormalizedUserName
+            SchemaBuilder.AlterTable(nameof(UserIndex), table => table
+                .AddColumn<string>(nameof(UserIndex.UserId)));
```
Skrypt, line 70: Length = 26 chars

Skrypt:
In OrchardCore.ContentManagement migrations.cs file.

```C# 
       public int UpdateFrom5()
        {
            SchemaBuilder.AlterTable(nameof(ContentItemIndex), table => table
                .AddColumn<string>("UserId", column => column.Nullable().WithLength(26))
            );
            return 6;
        }
```

deanmarcussen:
So I did this to be migration free, because @sebastienros didn't want to run a migration.

Otherwise when we do the OwnerId on ContentItems we will also have to run a migration against all the content items to move the owner over as well. 

Lot of risk of that failing due to timeouts on sites with large quantities of content items (or large quantities of users).

Because of that I can't set the `UserId` column to 26 because the current `UserName` property is not limited in length. It defaults to 256 on SQL Server, but will be different on other databases (probably only sqlite as it ignores max lengths).

(we should probably fix that in code as well as currently it will just throw an exception depending on the database)

If we did a migration we'd probably have to do it with YesSql, rather than SQL directly, to get support across all the databases (your sql script is great, but would it work on anything that isn't MS SQL?, i.e. postgres / mysql etc)

Skrypt:
I personally think it will require a SQL migration. I for exemple have a UserProfileIndex table which will require me to sync a new UserId column. Of course, the sql I wrote is a mock-up and works only in SQL Server but nothing prevents to compose the equivalent queries for other rdb's. Here I just tried to show that it can be done without requiring YesSQL or adding a complex fixed code in OC. We're still in RC1 we can still do breaking changes, I'd rather compose a SQL migration with some comments rather than trying to make something that migrates the data correctly only 90% with YesSQL... or that would require some changes to make it work. Also using the email as a userid doesn't make it harder to guess like a uid is, I'd prefer generating a uid for all of them and be done with it at least.

agriffard:
Among every pending PRs, please focus on this one.
Explaining if there are breaking changes and what they are is of course an important part.

sebastienros:
Upgrade step, from the admin.

deanmarcussen:
Ok @Skrypt I made an `OwnerId` property on `ContentItem` and an upgrade step.

Not as automatic as you might have liked. 

Still much todo, and merge conflicts to resolve, but some progress...

Skrypt:
Ok will pick up from there and make changes in a separate branch to test.

deanmarcussen:
@kevinchalet and @MichaelPetrinolis 

Here I am trying to do big risky things (without breaking anything) for good reasons...

Would appreciate any input as to whether this will have any flow on effect on `OpenId` by changing all the security to use `Claims.NameIdentifier` over `User.Identity.Name` or any other thoughts?

agriffard:
Is the upgrade step completely transparent?

deanmarcussen:
It is, as there is no longer an upgrade step, but equally this pr is not finished / ready

Still to be discussed. 
Should we still store the `OwnerName` property somewhere?

The blog recipe uses it for displaying the owner of a content item, and this is now being translated into a Guid.

Should we have something like we do for tags, where `OwnerName` becomes a less well known property (not updated on change).

Or should we be doing a lookup on the user table to get an owner name? (extra sql call)

Skrypt:
It is quite impossible to make the upgrade step completely transparent. Some people will have custom index providers that will need to be updated.

sebastienros:
About `OwnerName`, why not use the `Author` property that should contain the displayed value?

Skrypt:
@sebastienros I'm not sure what you are proposing. An author is quite different than a owner?

sebastienros:
I understand that, still one rock two birds.
First bird is that we don't need to do a query to get the owner's name, we can use the author's name for the current version.
Second bird is that what we are displaying (on the front-end at least) should probably not be the owner of the content item, but what we think is the Author?

Then from what I see in the code, the author is "kept" across edits, which means the field is only changed if the users decides to do it, and we can even put names of people that don't exist. Audit trail with still record who (user id, TBD) changed the content.

Skrypt:
I think we need to define what is Author, Owner and what they should be used for. I see here in Github they use something similar but they also allow to link a user to a shortcode @skrypt 

What would happen for example if we would change the Owner name now that we would have a UserId. Does that @skrypt-also shortcode will still render the proper UserId page? Same thing with the ContentItemIndex. We could keep the Owner name as a reference "in time" and also the "author" for that same matter. Though, the UserId should also be resolved in any case. So, if someone changes it's UserName, we don't require to update every content items he is related to. `@skrypt` here should be able resolve to 2 different UserId profile page if anyone took it over time. Of course this seems awkward and maybe this is not what we want but then we always have the second option of migrating the database on every UserName change to update the Owner name on every ContentItem.

deanmarcussen:
> About OwnerName, why not use the Author property that should contain the displayed value?
> I understand that, still one rock two birds.

I think this is the way to go - we are only talking about the sample blog recipe.

For the future I think `Owner` and `Author` will never be satisfactory for all purposes.
`Owner` is an `Authorization` concern, and `Author` a friendly helper.

One of the issues that bought this pr about was a `UserPickerField` - which we can't do, if we don't have the `UserId` to fix it against.

It won't go into this pr, but it would follow shortly after, and I think it addresses the need for people to be able to define their own identification properties, for the variety of reasons / use cases, that they may have.

So my proposal is keep it simple oob, and allow extension via a `UserPickerField` for those who want it / need it.

src/OrchardCore.Modules/OrchardCore.Settings/Deployment/SiteSettingsDeploymentSource.cs
```diff
@@ -69,6 +69,10 @@ public async Task ProcessDeploymentStepAsync(DeploymentStep step, DeploymentPlan
                         data.Add(new JProperty(nameof(ISite.SuperUser), site.SuperUser));
                         break;
 
+                    case "SuperUserId":
```
deanmarcussen, line 72: this is not needed now

src/OrchardCore.Modules/OrchardCore.Users/Migrations.cs
```diff
@@ -1,16 +1,46 @@
...
+                .CreateIndex("IDX_UserIndex_IsEnabled", "DocumentId", "IsEnabled")
+            );
+
+            SchemaBuilder.AlterTable(nameof(UserIndex), table => table
+                .CreateIndex("IDX_UserIndex_UserId", "DocumentId", "UserId")
+            );
+
+            SchemaBuilder.AlterTable(nameof(UserIndex), table => table
+                // This index will be used for lookups when logging in.
+                .CreateIndex("IDX_UserIndex_UserName", "DocumentId", "NormalizedUserName")
```
deanmarcussen, line 43: Should have index on email

src/OrchardCore/OrchardCore.Infrastructure/Security/AuthorizationHandlers/SuperUserHandler.cs
```diff
@@ -20,19 +21,31 @@ public SuperUserHandler(ISiteService siteService)
...
             {
-                foreach (var requirement in context.Requirements.OfType<PermissionRequirement>())
-                {
-                    context.Succeed(requirement);
-                }
+                SucceedAllRequirements(context);
+            }
+            // This method is maintained for backwards compatability during an upgrade migration where a user id has not been set yet.
+            // It can be removed in a later release.
+            else if (String.IsNullOrEmpty(site.SuperUserId) && String.Equals(context.User.Identity.Name, site.SuperUser, StringComparison.OrdinalIgnoreCase))
```
deanmarcussen, line 38: not required now

src/OrchardCore/OrchardCore.Setup.Abstractions/ISetupUserIdGenerator.cs
```diff
@@ -0,0 +1,7 @@
+namespace OrchardCore.Setup.Services
+{
+    public interface ISetupUserIdGenerator
```
deanmarcussen, line 3: To avoid a reference between `Setup` and `Users` we have two id generators, one for during setup (because the user is not created until after the recipe is run - i.e. the user creation is performed through setup event handlers) and one for user creation, after setup

src/docs/reference/modules/Liquid/README.md
```diff
@@ -338,7 +338,7 @@ Gives access to the current site settings, e.g `Site.SiteName`.
 | `MaxPageSize` | `100` | The maximum page size that can be set by a user. | 
 | `PageSize` | `10` | The default page size of lists. | 
 | `SiteName` | `My Site` | The friendly name of the site. | 
-| `SuperUser` | `admin` | The user name of the site's super user. | 
+| `SuperUserId` | `4kxfgfrxqmdpnt5n508cqvpvca` | The user id of the site's super user. |
```
deanmarcussen, line 341: revert

deanmarcussen:
Updated after review with @sebastienros yesterday

- It now runs a small migration on the user to set the new `UserId` property to that of the existing `UserName`, and update the YesSql Index. This is the current value of the `Owner` of any existing Content Items. Ensuring that existing content is still owned, by the same user, and ensuring that the `UserId` property is unique and set in the indexes.
- New users are assigned a unique `UserId`
- New content made by any user assigns their `UserId` value, whether guid, or unique name, to the `ContentItem.Owner` property

Reviews welcome.

@Skrypt you have done much with content permissions recently, so a test drive to ensure all continues correctly would be helpful.

src/OrchardCore/OrchardCore.ContentManagement/Handlers/ContentsHandler.cs
```diff
@@ -30,7 +30,7 @@ public override Task CreatingAsync(CreateContentContext context)
             var httpContext = _httpContextAccessor.HttpContext;
             if (context.ContentItem.Owner == null && (httpContext?.User?.Identity?.IsAuthenticated ?? false))
             {
-                context.ContentItem.Owner = context.ContentItem.Author = httpContext.User.FindFirstValue(ClaimTypes.NameIdentifier);
+                context.ContentItem.Owner = httpContext.User.FindFirstValue(ClaimTypes.NameIdentifier);
```
Skrypt, line 33: Reason?

deanmarcussen: The Owner is no longer the same as the Author.

The Author is `User.Name` (set two lines below)
The Owner is ClaimType of NameIdentifier (so UserId)

Skrypt: Okay and are we using also ContentItemIndex.Owner the same way?

deanmarcussen: Yes, no change to `ContentItemIndex.Owner` it keeps the same value.

So for an old user / old content item the `Owner` = "admin" (which is now by migration the User.UserId value - for old users)
For a new user -> new content Item the `Owner` = "userguid"

Skrypt (approved):
LGTM

sebastienros:
@agriffard google for "trigger-happy"

Skrypt:
lol, it's fine really. I think the structure is there. If there's anything we will surely know it in the few next days.
I've applied this on my current website and permissions we're working fine and all.
I think it is the more "transparent" that we can have for now.
Remaining is custom related and should be done with SQL.

PR #5436, opened 2020-02-01
Rendering zones before rendering layout

ns8482e:
Fixes #3947, #1994, #993

Rendering zones before Layout allow zones to register resources to ResourceManager.

PR #6768, opened 2020-07-24
Fix: IsSectionDefined  exception in Layout

### LINKED ISSUE #6606: IsSectionDefined throws exception
using `IsSectionDefined` in a `Layout` page throws an exception of `HtmlContentBuilder` does not contain a definition for `Items`

ns8482e:
Fixing exception in Layout when IsSectionDefined  is used.

Fixes #6606

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
@@ -269,7 +269,7 @@ namespace OrchardCore.DisplayManagement.Razor
 
             var zone = ThemeLayout[name];
 
-            return !(zone is ZoneOnDemand) && zone != null;
+            return zone != null;
         }
 
         /// <summary>
@@ -336,9 +336,8 @@ namespace OrchardCore.DisplayManagement.Razor
             }
 
             var zone = ThemeLayout[name];
-            
-            // We don't throw on a 'ZoneOnDemand' shape that has no items.
-            if (required && !(zone is ZoneOnDemand) && zone is Shape && zone.Items.Count == 0)
+
+            if (required && zone != null && zone is Shape && zone.Items.Count == 0)
             {
                 throw new InvalidOperationException("Zone not found: " + name);
             }
--- a/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
@@ -7,6 +7,7 @@ using Microsoft.AspNetCore.Mvc.ViewFeatures;
 using Microsoft.Extensions.DependencyInjection;
 using OrchardCore.DisplayManagement.Shapes;
 using OrchardCore.DisplayManagement.Title;
+using OrchardCore.DisplayManagement.Zones;
 using OrchardCore.Settings;
 
 namespace OrchardCore.DisplayManagement.Razor
@@ -268,7 +269,7 @@ namespace OrchardCore.DisplayManagement.Razor
 
             var zone = ThemeLayout[name];
 
-            return zone != null;
+            return !(zone is ZoneOnDemand) && zone != null;
         }
 
         /// <summary>
@@ -335,8 +336,9 @@ namespace OrchardCore.DisplayManagement.Razor
             }
 
             var zone = ThemeLayout[name];
-
-            if (required && zone != null && zone is Shape && zone.Items.Count == 0)
+            
+            // We don't throw on a 'ZoneOnDemand' shape that has no items.
+            if (required && !(zone is ZoneOnDemand) && zone is Shape && zone.Items.Count == 0)
             {
                 throw new InvalidOperationException("Zone not found: " + name);
             }
--- a/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
@@ -268,7 +268,7 @@ namespace OrchardCore.DisplayManagement.Razor
 
             var zone = ThemeLayout[name];
 
-            return zone != null && zone.Items.Count > 0;
+            return zone != null && ((zone is IHtmlContent) || (zone is Shape && zone.Items.Count > 0));
         }
 
         /// <summary>
--- a/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
@@ -268,7 +268,7 @@ namespace OrchardCore.DisplayManagement.Razor
 
             var zone = ThemeLayout[name];
 
-            return zone != null && ((zone is IHtmlContent) || (zone is Shape && zone.Items.Count > 0));
+            return zone != null;
         }
 
         /// <summary>
--- a/src/OrchardCore/OrchardCore.DisplayManagement/Theming/ThemeLayout.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/Theming/ThemeLayout.cs
@@ -37,7 +37,7 @@ namespace OrchardCore.DisplayManagement.Theming
                 {
                     foreach (var zone in layout.Properties.ToArray())
                     {
-                        if (zone.Value is IShape shape)
+                        if ( zone.Value != null && zone.Value is IShape shape)
                         {
                             // Render each layout zone
                             var htmlZone = await DisplayAsync(shape);
--- a/src/OrchardCore/OrchardCore.DisplayManagement/Theming/ThemeLayout.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/Theming/ThemeLayout.cs
@@ -37,7 +37,7 @@ namespace OrchardCore.DisplayManagement.Theming
                 {
                     foreach (var zone in layout.Properties.ToArray())
                     {
-                        if ( zone.Value != null && zone.Value is IShape shape)
+                        if ( !(zone.Value is ZoneOnDemand) && zone.Value is IShape shape)
                         {
                             // Render each layout zone
                             var htmlZone = await DisplayAsync(shape);


src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
```diff
@@ -268,7 +268,7 @@ public new bool IsSectionDefined(string name)
 
             var zone = ThemeLayout[name];
 
-            return zone != null && zone.Items.Count > 0;
+            return zone != null && ((zone is IHtmlContent) || (zone is Shape && zone.Items.Count > 0));
```
deanmarcussen, line 271: Why not just `zone != null`

It becomes fragile when you check all these types etc - why it broke in the first place :)

ns8482e: > Why not just `zone != null`

I agree but only `zone != null` is not enough, As on-demand zone returned  are always not null and has no `Items`.

> why it broke in the first place :)

It broke because of PR #5436, where we added logic to pre-render zones before layout.

jtkech: Yes i understand, the pre-rendered zones (that are `IHtmlContent`) are no more `Shape` so without an `Items` property. But that means that this check is still not complete as the following one in `RenderSectionAsync()`.

            if (required && zone != null && zone is Shape && zone.Items.Count == 0)

So maybe in `ThemeLayout` where we pre-render zones we could check if it is a `Shape` and then if it has `Items` (or the opposite) wrap the `htmlZone` result in a custom `IHtmlContent` that we could check in these both places.

ns8482e: > But that means that this check is still not complete as the following one in `RenderSectionAsync()`.

This line is not a problem as - if zone is required and not defined anywhere - then there won't be any pre-rendering, hence here when called `RenderSectionAsync()` it will always be on-demand zone `Shape`  with `Items.Count == 0`

jtkech: Here my **very little concern** is not to prevent it to throw but to behave as before and then to throw if it would have done it before. E.g. if a zone has no items when we pre-render it, with the previous code it would have thrown in `RenderSectionAsync()` if required, and in `IsSectionDefined()` it would have return false.

I tried it with the blog recipe, i removed the only one widget in the footer zone, indeed in that case the layer filter doesn't create a footer zone on demand, but i think it is still possible to have a zone without any items when we pre-render it, and we would need to know it to behave as before.

Note: Seems strange to not throw in `RenderSectionAsync()` if the zone is required and the zone is null, only if not null and doesn't have any items, but i think there is a good reason ;)

ns8482e: I guess the original code condition (`Items.Count == 0`) is defined to differentiate between defined zone vs zone on-demand.  So, IMO there no change in behavior between before and after for `RenderSectionAsync` 



> Note: Seems strange to not throw in `RenderSectionAsync()` if the zone is required and the zone is null, only if not null and doesn't have any items, but i think there is a good reason ;)

May be it's because of confusing behavior of `ZoneOnDemand` 

Calling `var zone = ThemeLayout[name];` is null but `var zone = ThemeLayout.Zones[name];` is not null, also calling `ThemeLayout.Header` is not null

jtkech: @ns8482e 

Yes this is what the layout filter does `layout.Zones[widget.Zone]`

In the same filter, just tried `var test = layout.Zones["Test"]` that returns a `ZoneOnDemand` but then if you don't add any items we don't retrieve it in `ThemeLayout.cs`. To retrieve it, before pre-rendering, as a shape without any items i needed to also do `layout["Test"] = test` (which happens somewhere if you add an item).

So, it is still possible to pre-render a zone shape without any item, but i agree that normally it should not happen, so finally i'm fine with what you did ;)

deanmarcussen: ok, so I am ok with it, but I think someone has to work very hard to create a zone with no items.

The bit I want to clarity is any confusion about the `ZoneOnDemand`
> calling ThemeLayout.Header is not null.

I think it is null when you call `ThemeLayout.ZoneThatDoesNotExist` 
You need to put something in it as @jtkech says with `layout["Test"] = test` or `ThemeLayout.ZoneThatDoesNotExist.Add`.

The reason I want to clarify @ns8482e is because I see in your other prs often you are doing `if(Model.Content != null && Model.Content is Shape etc)` and we should not have to be doing this.

If you are having to do this then something is wrong somewhere, either with the way you are creating zones or the way the `ZoneOnDemand` is working.

ns8482e: @jtkech @deanmarcussen  It's valid concern,  `var zone = ThemeLayout[name];`  Here `zone` could be null `required && zone == null` is broken now.

All I am saying is -   `IsSectionDefined` is broken by pre-rendering zones, PR #5436 - that this PR is providing fix for. 
However `RenderSectionAsync` and `RenderSectionTag` are unaffected by change in PR #5436 as it only pre-renders the zones that exists(by keys of properties) . 

It might have broken by some other change as `ThemeLayout[name]` is returning null now, where as IMHO it should return not null `ZoneOnDemand`

Regarding `ZoneHolding`

@deanmarcussen calling `Model.PropertyThatDoesNotExist` on `Shape` instance returns null, where as `Layout.PropertyThatDoesNotExist` on `ZoneHolding` instance returns not null ZoneOnDemand object. In both cases neither `_properties` contain key `PropertyThatDoesNotExist` nor `_items` contains `Named` shape. So, checking `Model.Content != null` is valid check in needed code on `dynamic` to avoid exception.

ns8482e: @deanmarcussen I guess you are correct! for `IsSectionDefined`  comparing only `zone != null` should be enough as `operator !=` evaluates to `false` for `ZoneOnDemand`  and Ideally  empty zone  defined without any items like `<zone name="Footer"></zone>` in Views,  still should be considered as defined zone.

jtkech: @deanmarcussen @ns8482e 

Some findings by doing more tests on the `Layout` that is a `ZoneHolding`.

> It might have broken by some other change as ThemeLayout[name] is returning null now, where as IMHO it should return not null ZoneOnDemand

Yes `Layout["SomeZone"]` may return null but not `Layout.SomeZone` because `ZoneHolding` overrides `TryGetMember()` of the `Shape` but not `TryGetIndex()` of the `Composite`, but i don't think it's due to a recent change.

Note: I noticed that getting `Layout.SomeZone` also set `Layout["SomeZone"]` that is no more null, which is not the case after getting `Layout.Zones["SomeZone"]` that only returns a `ZoneOnDemand` without setting the related layout property, here maybe a missing line of code but i don't think so (see below in the next section).

> Side Question: if everything is prerendered as IHtmlContent how can you get a Shape in there?

When a zone is pre-rendered a zone tag helper may be called and add a zone that will not be pre-rendered but rendered by the layout as a shape. I tried it by using `{% zone "Test" %} Hello world {% endzone %}` in `Widget-RawHtml.liquid` that is pre-rendered in the `Footer` and then using `{% render_section "Test", required: true %}` in `Layout.liquid`. @ns8482e even i just used `{% zone "Test" %}{% endzone %}` i got a shape with one item.

---

Yes, it's not easy to chek everything as it depends on how things are overridden (or not) in classes and their derived. So let's only talk of the accessor that are used in the related part of code we are talking about.

So for me `Layout.Zones["SomeZone"]` is like an helper that returns a `ZoneOnDemand` that is not yet set in the layout properties, and this is when you `AddAsync()` an item that it does it with `parent[_potentialZoneName] = zone`. And the same happens when we use the `zone` tag helper (razor or liquid) even with no content.

So yes if a zone exists in a layout property "normally" it has at least one item.

---

Before we were checking that a zone shape has items, **but only if this zone shape was not null**, if null it doesn't fail even the zone is required, seems weird but it is like this since a long time.

- So if we really want to make it behaves as before we would need to check it before pre-rendering and e.g. wrap the result in a wrapper to be able to check it afterwards, as i was first suggesting. Note: @ns8482e when you added `IsSectionDefined()` you didn't do the exact same check, here **the zone can't be null** and it makes sense.

---

Otherwise, if we all agree that if a zone exists in a layout property normally it has items, i'm tempted to remove the check of `zone.Items.Count`.

- If so in `IsSectionDefined()` we could just return `zone != null`.

- Then in `RenderSectionAsync()` just check `if (required && zone == null)` before throwing. We would also need to update the `{% render_section %}` liquid tag.

    But notice that before it was not failing if the zone was null, so it changes the original behavior and may be a little breaking change. But here if we don't check `zone == null` the `required` param would become useless.

    But for me it makes sense that when a zone is required it should not be null.

    As i can see in razor we always use `RenderSectionAsync()` with `required: false`, and in liquid we only use the default value for the `Content` zone, but the default value is still false. Note: in razor the default value is true.

---

The last possibility is to make the `required` param obsolete ;)

ns8482e: >  * Or in `IsSectionDefined()` just return `zone != null`

Done! removed all other conditions. Also added `zone != null` in `ThemeLayout` to exclude `ZoneOnDemand` shapes that might have added by calling `Layout.SomeZone`  which is as good as `null`


> * Then in `RenderSectionAsync()` just check `if (required && zone == null)`

To throw an exception for required zone when it's null - I think it's an existing bug and we should create separate issue for that to track separately and create separate PR if we decide to fix the behavior.


> * that we agree that an existing zone always has items,

As verified, once zone is defined `<zone name="SomeZone"></zone>` in  Views, creates at-least on item (with empty string) , In case If developer removes all items manually from zone - IMHO it still considered zone as defined  zone and so in that case it should not throw exception, is its fine for me to check only `if (required && zone == null)` in `RenderSectionAsync`. ( But in separate issue and PR if we decide to fix it.)

jtkech: @ns8482e Okay cool, that's a good compromise

> Also added zone != null in ThemeLayout to exclude ZoneOnDemand shapes that might have added by calling Layout.SomeZone

Just tried, the added zone is not null so you don't need to add a null check in `ThemeLayout`. Anyway the `is` operator can be applied to a null value. So you don't need `zone.Value != null && zone.Value is IShape shape`, anyway `zone.Value is IShape shape` is enough.

ns8482e: > Note: I noticed that getting `Layout.SomeZone` also set `Layout["SomeZone"]` that is no more null

@jtkech added `zone.Value != null` in `ThemeLayout` because of above, Just safety check in case if user calls `var x = Layout.SomeZone` but don't add any shape to it then, `Layout["SomeZone"]` will have `ZoneHolding`  which is as good as `null` and we don't want to pre-render it.

jtkech: @ns8482e 

Okay but in that case `zone.Value` is not null so the null check is useless, and the rendering result is just an empty string. If you really want to not render it in that case you would have to check it is a `Shape` and has no items.

But that' okay to not check it at all.

ns8482e: > Okay but in that case `zone.Value` is not null so the null check is useless

Am I missing something? `zone.Value`  is an object of `ZoneHolding` but `== null` evaluates to true. - as ZoneHolding overrides the `==` and `!=` 

However, I am fine with removing the check

jtkech: I don't see these operator overrides but as i tried in that case `zone.Value` is a `ZoneOnDemand` with no items.

EmptyZone

ns8482e: @jtkech https://github.com/OrchardCMS/OrchardCore/blob/1e7fc40ec74bb1b0b2c05f88acfdd05d9945006a/src/OrchardCore/OrchardCore.DisplayManagement/Zones/ZoneHoldingBehavior.cs#L171-L181

If we exclude `ZoneOnDemand` from pre-rendering then if we decide to throw exception in `RenderSectionAsync` for  `if (required && zone == null)`  will be true and throws exception.

If don't include the null check in pre-render then `ZoneOnDemand` will be pre-rendered with empty string then (required && zone == null)`  will be false - won't throw exception.

jtkech: @ns8482e 

Oops didn't see these overrides, thanks, it never ends ;)

Oh yes i understand the goal, if we add an item in a `ZoneOnDemand`, it is enlisted in a property of the parent layout but it is no more a `ZoneOnDemand`, so a `ZoneOnDemand` is considered as a null zone.

But through the debugger i can see that they are not called, need to see why.

jtkech: @ns8482e 

Okay, this is because the layout properties is a dictionary of string keys and `object` values. The operator overrides get called if you use `if (zone.Value is ZoneOnDemand myZone)` an then do the null check on `myZone`.

ns8482e: yes through debugger you could see an object but when it executes`zone.Value != null` evaluates to `false` if it's `ZoneOnDemand`

jtkech: No `zone.Value != null` evaluates to `true` and `zone.Value == null` evaluates to `false` if it has been get from the layout properties `dictionary<string, object>` because the operator overload is not called

But if the object is strongly typed to a `ZoneOnDemand`, e.g. by doing `Zone.Value is ZoneOnDemand typed`, and then do a null check on `typed`, it is the opposite as the operator overload is called in that case.

jtkech: So i think there is an original issue when we do in `RenderSectionAsync()`

            var zone = ThemeLayout[name];

            if (required && zone != null && zone is Shape && zone.Items.Count == 0)
            {
                throw new InvalidOperationException("Zone not found: " + name);
            }

I think we do `zone != null` to not fail on a `ZoneOnDemand` that is a shape but that doesn't have any items. But because we get a non typed object from `ThemeLayout[name]`, the custom operator is not called and `zone != null` is true.

jtkech: @ns8482e 

So one suggestion would be in `RenderSectionAsync()`


            // We don't throw on a 'ZoneOnDemand' shape that has no items.
            if (required && !(zone is ZoneOnDemand) && zone is Shape && zone.Items.Count == 0)

In `IsSectionDefined()`

            // An empty 'ZoneOnDemand' is not a defined section.
            return zone != null && !(zone is ZoneOnDemand);

In `ThemeLayout` (just saw that you did it)

            // We don't pre-render a 'ZoneOnDemand' shape that has no items.
            if (!(zone.Value is ZoneOnDemand) && zone.Value is IShape shape)

So that the zone will not be recognized as a defined section, but will still not throw on rendering (here an empty string)

deanmarcussen: @ns8482e @jtkech Awesome! This is a great discussion.

> So one suggestion would be in RenderSectionAsync()

These suggestions make sense to me too.

I have one last suggestion. 
Maybe we need a unit test around `ZoneOnDemand` and how it is behaving, so that we know what you have done here is provable in all the variations you have been testing?

We have one for the `NilInstance` shape, but not yet `ZoneOnDemand` 

It might help us, or someone else, in the future when we have to wonder why we did this

ns8482e: @jtkech Updated as you sugggested! 

One last question, I still not understand, why runtime is not calling instance's operator and calling boxed - object's operator ? Even though it's boxed as object - it's still an instance of `ZoneOnDemand`  inside `_properties` dictionary.

As `ZoneOnDemand` boxed as object and it's calling object's operator instead of it's own operator then its seems operator definition seems useless here

jtkech: @ns8482e 

> @jtkech Updated as you sugggested!

Cool thanks

> One last question, I still not understand, why runtime is not calling instance's operator

Yes i saw that when i tried it, it is evaluated at compile time. Also tried when used as a `Shape` (its base type), the following test doesn't pass either, so no polymorphism, not like virtual methods where it is evaluated at runtime.

        [Fact]
        public void ZoneOnDemandShapeShouldEqualToNull()
        {
            var zoneHolding = new ZoneHolding(() => new ValueTask<IShape>(new Shape()));
            Shape zoneOnDemand = new ZoneOnDemand(() => new ValueTask<IShape>(new Shape()),
                zoneHolding, "ZoneName");

            Assert.True(zoneOnDemand == null);
        }

> ... then its seems operator definition seems useless here

Yes and every time it is not used as a `zoneOnDemand` (or a derived type) but as an `object` or as a less specialized type e.g. as a `Shape` that is its base type.

@deanmarcussen 

> Maybe we need a unit test around ZoneOnDemand and how it is behaving

Maybe, but do we need an unit test to show something that is not behaving as it was originally intended. About the `Nil.Instance` it's the same, i created the following test that doesn't pass either.

        [Fact]
        public void NilObjectShouldEqualToNull()
        {
            object nil = Nil.Instance;
            Assert.True(nil == null);
        }

I think that's okay for these operator overloads defined for `Nil` and also for `Composite`. For `ZoneOnDemand` maybe better to just remove them (if not used elsewhere) if now we are using the `is` operator.

deanmarcussen: @jtkech 
> Maybe, but do we need an unit test to show something that is not behaving as it was originally intended. About the `Nil.Instance` it's the same, i created the following test that doesn't pass either.
> 
> ```
>     [Fact]
>     public void NilObjectShouldEqualToNull()
>     {
>         object nil = Nil.Instance;
>         Assert.True(nil == null);
>     }
> ```
> 
> I think that's okay for these operator overloads defined for `Nil` and also for `Composite`. For `ZoneOnDemand` maybe better to just remove them (if not used elsewhere) if now we are using the `is` operator.

Well if it isn't behaving as intended, then a test is a perfect way to prove that. 
Or to demonstrate that it should be used in a specific manner, i.e. with `is` 

Saves us all getting the debugger out...

But nothing to do for this pr. Maybe another :)

jtkech: @deanmarcussen Okay make sense, i will do a PR asap

jtkech: @ns8482e @deanmarcussen 

Sorry guys, while working on more unit tests (always good to do ;)) i saw that the operator overloads are called when a `ZoneOnDemand` is used as a `dynamic` (so runtime evaluation is done), as done in `IsSectionDefined()` and `RenderSectionAsync()`, but still not in `ThemeLayout` because we use the `ZoneHolding.Properties` that is a `Dictionary<string, object>`.

So what i suggested for `ThemeLayout` is still needed

But `IsSectionDefined()` could just return `zone != null`

And `RenderSectionAsync()` could be let as it was

But not against to let the 2 last ones as suggested, with the `is ZoneOnDemand` check for clarity.

I will complete the unit tests tomorrow

ns8482e: @jtkech It seems you are getting confused. IMO we made right decision to use `!(zone is ZoneOnDemand)` because of following reasons we already discussed earlier.

1. `ThemeLayout[name]` returns `null` for zone that doesn't exists - as `ZoneHolding` doesn't override `TryGetIndex()`
2. `ThemeLayout.Zones[name]` returns `ZoneOnDemand` for zone that doesn't exists. as `ZoneOnDemand` overrides `TryGetIndex()` but doesn't sets `_properties`
3. `ThemeLayout.ZoneName` returns `ZoneOnDemand` for zone that doesn't exists and sets `_properties`
4. `==` and `!=` operators only works on `ZoneOnDemand` types and doesn't work on `object`

IMO Following line is behind all this confusion.
https://github.com/OrchardCMS/OrchardCore/blob/6cddd9f9904ff117ba82587f39fc76f49fac3cc3/src/OrchardCore/OrchardCore.DisplayManagement/Zones/ZoneHoldingBehavior.cs#L54
 

> i saw that the operator overloads are called when a `ZoneOnDemand` is used as a `dynamic`

Yes it only works when we call `ThemeLayout.Zones[name]` or  `ThemeLayout.ZoneName`  because the returned values is of type `ZoneOnDemand` and it works when `ThemeLayout[name]` is really `null`

But it doesn't work if `ThemeLayout.ZoneName` is called already and `ZoneOnDemand` has set in `_properties` and then after when accessed it returns as `object` and  `==` and `!=` operators are failing.

If we remove above line - then we can safely use `zone != null` every where.

jtkech: @ns8482e 

I agree with all what you said above unless that if `ThemeLayout` is **dynamic** which is the case in `IsSectionDefined()` and `RenderSectionAsync()`, `ThemeLayout[name]` uses the `Composite.TryGetIndex()` that uses its `TryGetMemberImpl()`, **and return a dynamic object** whose runtime type is evaluated. So, if an empty `ZoneOnDemand` has been added, the operator works on `ThemeLayout[name]`. I should have tested it in these 2 other places, not only in `ThemeLayout.cs` where we need to use `ZoneHolding.Properties` but that returns objects.


I did all the related unit tests and more, i will commit a PR, and i tried the following

- So in `ThemeLayout` i did `_ = ThemeLayout.SomeZone` that adds an empty `ZoneOnDemand` in the layout

- Then i kept the suggestion in `ThemeLayout` to not pre-render this empty `ZoneOnDemand`

- Then `RenderSectionAsync("SomeZone", required: true)` doesn't fail with the current code

- And `IsSectionDefined()` that only returns `zone != null` says that the section is not defined

    And through the debugger i can see the instance of the above `ZoneOnDemand` that has not been pre-rendered

ns8482e: @jtkech - Updated

jtkech: Cool, sorry for the long story ;)

ns8482e: @jtkech  So, What happen if we change 
`foreach (var zone in layout.Properties.ToArray())` 
to 
`foreach (dynamic zone in layout.Properties.ToArray())` in `ThemeLayout`. Will it then evaluate `zone.Value == null` to `true` for `ZoneOnDemand`

jtkech: @ns8482e 

Exactly, i thought about this but then it will have to create dynamic objects, then evaluate the runtime types to call the operator overloads, for perf here better to still not working with dynamic objects

ns8482e: @jtkech I don't want drag further - But - I still believe that s line 54 is useless - in below code.

Based on your finding - I guess line 49 will always be true. So actually it's never using `ZoneOnDemand `that set in `_properties` instead it always creating new one. So we just remove the line 54.

https://github.com/OrchardCMS/OrchardCore/blob/6cddd9f9904ff117ba82587f39fc76f49fac3cc3/src/OrchardCore/OrchardCore.DisplayManagement/Zones/ZoneHoldingBehavior.cs#L49-L54

jtkech: @ns8482e 

> I don't want drag further - But - I still believe that s line 54 is useless - in below code.

No problem i like complicated things ;) I saw that, yes we could remove this line so that `zoneHolding.SomeZone` would not add an empty `ZoneOnDemand` that we would have to check, but if we do multiple accesses to a dynamic property before adding any item, it prevents to re-create a new `ZoneOnDemand` (as it does see below) each time. But on another side this is not what we do in the related code for `Zones` line 86 that doesn't call `TrySetMemberImpl()`.

> Based on your finding - I guess line 49 will always be true. So actually it's never using ZoneOnDemand that set in _properties instead it always creating new one. So we just remove the line 54.

No because if `result` is already a `ZoneOnDemand`, `(null == result)` is false because here result is an `object`. Notice that in the related code for `Zones`, `parentMember` is a `dynamic` so if it is a `ZoneOnDemand`, `(parentMember == null)` is true, that's why here it always return a new `ZoneOnDemand`.

Notice that the above `TryGetMember()` is called when using `zoneHolding.SomeZone`, when using `zoneHolding["SomeZone"]` this is the `TryGetIndex()` of the `Composite` that is called. And for `Zones` it is different as it also overrides `TryGetMemberImpl()` that is called by `TryGetIndex()` of `Composite` ;)

**So for me that's okay as it is**, but if you want to remove the above line 54 we could dig into it.

PR #6798, opened 2020-07-31
Return search results in the correct order

### LINKED ISSUE #6791: Search Results Not Displayed in Correct Order
When search results are displayed, they are often in the wrong order.

I have reproduced this issue on version 1.0.0-rc2-13968 and on the dev branch at edc370705709a7ee94c2940fb17859701afb1b75.

### Reproducing
1. Create a new Orchard Core CMS site.
2. Start the site, and initialize it with the blog recipe.
3. Go to the admin backend, and create two new articles: one with the title "Matches Worse" and one with the title "Matches Better". Make sure you create "Matches Worse" before creating "Matches Better".
4. Go to /Search, and search for "Matches Better".
5. The article called "Matches Worse" will be listed first, even though the "Matches Better" article is a better match.

### Screenshots
bad-search-results

### Possible Causes
I think this is caused by the lack of ordering in OrchardCore.Lucene's [SearchController](https://github.com/OrchardCMS/OrchardCore/blob/acee23ef121a09568198ebda164d591ae9f084ed/src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/SearchController.cs#L140). Since there is no OrderBy clause in the `queryDb` query, `containedItems` is likely to be in a different order than the correctly ordered `contentItemIds`.

thomasebsmith:
Fixes #6791.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/SearchController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/SearchController.cs
@@ -137,7 +137,10 @@ namespace OrchardCore.Lucene.Controllers
                 end = Convert.ToInt32(pagerParameters.After) + pager.PageSize + 1;
             }
 
-            var contentItemIds = await _searchQueryService.ExecuteQueryAsync(query, searchSettings.SearchIndex, start, end);
+            // Take the next (pager.PageSize + 1) results.
+            var contentItemIds = (await _searchQueryService.ExecuteQueryAsync(query, searchSettings.SearchIndex, start, end))
+                .Take(pager.PageSize + 1)
+                .ToList();
 
             // We Query database to retrieve content items.
             IQuery<ContentItem> queryDb;
@@ -155,7 +158,8 @@ namespace OrchardCore.Lucene.Controllers
                     .Take(pager.PageSize + 1);
             }
 
-            var containedItems = await queryDb.ListAsync();
+            // Sort the content items by their rank in the search results returned by Lucene.
+            var containedItems = (await queryDb.ListAsync()).OrderBy(x => contentItemIds.IndexOf(x.ContentItemId));
 
             // We set the PagerSlim before and after links
             if (pagerParameters.After != null || pagerParameters.Before != null)
--- a/src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/SearchController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/SearchController.cs
@@ -137,9 +137,7 @@ namespace OrchardCore.Lucene.Controllers
                 end = Convert.ToInt32(pagerParameters.After) + pager.PageSize + 1;
             }
 
-            // Take the next (pager.PageSize + 1) results.
             var contentItemIds = (await _searchQueryService.ExecuteQueryAsync(query, searchSettings.SearchIndex, start, end))
-                .Take(pager.PageSize + 1)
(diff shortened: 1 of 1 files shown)


hishamco:
Could you please fix the build?

deanmarcussen:
Closed to restart CI

src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/SearchController.cs
```diff
@@ -155,7 +158,8 @@ public async Task<IActionResult> Search(SearchIndexViewModel viewModel, PagerSli
                     .Take(pager.PageSize + 1);
             }
 
-            var containedItems = await queryDb.ListAsync();
+            // Sort the content items by their rank in the search results returned by Lucene.
+            var containedItems = (await queryDb.ListAsync()).OrderBy(x => contentItemIds.IndexOf(x.ContentItemId));
```
deanmarcussen, line 162: I'd be interested to know if this could be part of the sql query above, i.e. sorted at source, rather than being done by linq afterwards.

src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/SearchController.cs
```diff
@@ -137,7 +137,10 @@ public async Task<IActionResult> Search(SearchIndexViewModel viewModel, PagerSli
                 end = Convert.ToInt32(pagerParameters.After) + pager.PageSize + 1;
             }
 
-            var contentItemIds = await _searchQueryService.ExecuteQueryAsync(query, searchSettings.SearchIndex, start, end);
+            // Take the next (pager.PageSize + 1) results.
```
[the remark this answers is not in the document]
thomasebsmith: > I think the pager values are taken into account in the previous 5-6 lines?

You're right. This change is not needed.

src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/SearchController.cs
thomasebsmith: > I'd be interested to know if this could be part of the sql query above, i.e. sorted at source, rather than being done by linq afterwards.

I don't think it is possible to sort the search results in the SQL query using YesSql. Since the order of the search results is not predictable, a custom sort would need to be used. My understanding is that this is usually accomplished with a `CASE` statement:
```sql
ORDER BY CASE WHEN ContentItemId = '<ID of 1st result>' THEN '1'
              WHEN ContentItemId = '<ID of 2nd result>' THEN '2'
              WHEN ContentItemId = '<ID of 3rd result>' THEN '3'
              ...
              END ASC
```

I don't know of a way to generate custom orderings like that in YesSql.

I also noticed that [`GetAsync(IEnumerable<string> contentItemIds, bool latest = false)`](https://github.com/OrchardCMS/OrchardCore/blob/a747017be08d5839223ee4debfa028c7ac426f6e/src/OrchardCore/OrchardCore.ContentManagement/DefaultContentManager.cs#L94) orders the results of its SQL query using Linq.

jtkech: I think that's okay to not do ordering on YesSql side as it would not reduce the number of returned items.

deanmarcussen: Yeah, I agree, I was just curious if you could. Then I googled it and looked at the sql required...

PR #6809, opened 2020-08-02
Implement the new localized display names/descriptions store APIs introduced in OpenIddict 3.0 beta3

kevinchalet:
OpenIddict 3.0 beta3 - that will be released next week - introduces new APIs allowing to store localized versions of user facing strings:
  - Application display names.
  - Scope display names.
  - Scope descriptions.

Note: this PR implements the OpenIddict stores methods but doesn't update the GUI to support adding multiple display names/descriptions. I'll open a separate ticket to track that for a future Orchard version.

### DIFF
--- a/src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Models/OpenIdApplication.cs
+++ b/src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Models/OpenIdApplication.cs
@@ -1,4 +1,5 @@
 using System.Collections.Immutable;
+using System.Globalization;
 using Newtonsoft.Json.Linq;
 
 namespace OrchardCore.OpenId.YesSql.Models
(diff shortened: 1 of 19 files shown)


src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Models/OpenIdApplication.cs
```diff
@@ -25,20 +24,23 @@ public class OpenIdApplication
...
+        /// Gets or sets the display name associated with the current application.
         /// </summary>
         public string DisplayName { get; set; }
 
         /// <summary>
-        /// Gets or sets the physical identifier
-        /// associated with the current application.
+        /// Gets or sets the localized display names associated with the application.
+        /// </summary>
+        public ImmutableDictionary<CultureInfo, string> DisplayNames { get; set; }
```
kevinchalet, line 39: FYI, I tested manually and it works just fine (most likely because JSON.NET supports `TypeConverter`, for which there's a dedicated `CultureInfoConverter` to convert `CultureInfo` instances from/to strings):

```json
{
    "DisplayNames":{"fr-FR":"Ma super app"}
}
```

PR #6813, opened 2020-08-03
Blocking Detection module

infofromca:
Blocking calls can lead to ThreadPool starvation. Ouputs a warning to the log when blocking calls are made on the ThreadPool.

ref:
https://github.com/benaadams/Ben.BlockingDetector


sebastienros:
Not sure about making it a module, but would be nice at least to have a snapshot of what it finds when we run it on the default recipes. Can you dump the results you find locally? I have done that in the past, but it was a few years ago.

agriffard:
Closing. No activity for a long time.
Please fix the conflicts and reopen if needed.

PR #6823, opened 2020-08-04
Add UI permissions based on the user role

### LINKED ISSUE #6820: Revise Contributor role
As we all knew the `Contributor` role not be able to add a new content items, so how the user in this role should be called contributor?!!

I think we need to discuss and agree about the permission for this role before 1.0.0 release

/cc @agriffard @deanmarcussen @sebastienros

### LINKED ISSUE #6847: Disable action buttons based on role permession(s)
There was a notice behavior when we the user's role changed in the admin site, especially when you trying to click action buttons on content items

Many of us during the demo of Impersonation by @agriffard last two weeks notice a 404 when he published a content with a contributor user, I think Seb point to that too

So we need to avoid such errors before it happen, a proper way is to add a UI permissions to those action buttons, so the user will able to click the action buttons that has a permission to them

hishamco:
Fixes #6847

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.PublishButton.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.PublishButton.cshtml
@@ -1,22 +1,28 @@
-@*@if (Authorizer.Authorize(Permissions.PublishContent, (IContent)Model.ContentItem)) {*@
+@using Microsoft.AspNetCore.Authorization
+@inject IAuthorizationService AuthorizationService
+
 @{
+    ContentItem contentItem = Model.ContentItem;
     var returnUrl = Context.Request.Query["returnUrl"];
-}
-@if (String.IsNullOrWhiteSpace(returnUrl))
-{
-    <button type="submit" name="submit.Publish" class="publish-button btn btn-success" value="submit.Publish">@T["Publish"]</button>
-}
-else
-{
-    <div class="btn-group">
-        <button class="publish-button btn btn-success" type="submit" name="submit.Publish" value="submit.Publish">@T["Publish"]</button>
-        <button type="button" class="btn btn-success dropdown-toggle dropdown-toggle-split" data-reference="parent" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-            <span class="sr-only">@T["Toggle Dropdown"]</span>
-        </button>
-        <div class="dropdown-menu">
-            <button class="dropdown-item" type="submit" name="submit.Publish" value="submit.PublishAndContinue">@T["and continue"]</button>
-        </div>
-    </div>
+    var hasPublishContentPermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.PublishContent, contentItem);
 }
 
-@* } *@
+@if(hasPublishContentPermission)
+{
+    if (String.IsNullOrWhiteSpace(returnUrl))
+    {
+        <button type="submit" name="submit.Publish" class="publish-button btn btn-success" value="submit.Publish">@T["Publish"]</button>
+    }
+    else
+    {
+        <div class="btn-group">
+            <button class="publish-button btn btn-success" type="submit" name="submit.Publish" value="submit.Publish">@T["Publish"]</button>
+            <button type="button" class="btn btn-success dropdown-toggle dropdown-toggle-split" data-reference="parent" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                <span class="sr-only">@T["Toggle Dropdown"]</span>
+            </button>
+            <div class="dropdown-menu">
+                <button class="dropdown-item" type="submit" name="submit.Publish" value="submit.PublishAndContinue">@T["and continue"]</button>
+            </div>
+        </div>
+    }
+}
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.SaveDraftButton.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.SaveDraftButton.cshtml
@@ -1,19 +1,28 @@
+@using Microsoft.AspNetCore.Authorization
+@inject IAuthorizationService AuthorizationService
+
 @{
+    ContentItem contentItem = Model.ContentItem;
     var returnUrl = Context.Request.Query["returnUrl"];
+    var hasEditContentPermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.EditContent, contentItem);
 }
-@if (String.IsNullOrWhiteSpace(returnUrl))
+
+@if(hasEditContentPermission)
 {
-    <button class="primaryAction btn btn-primary" type="submit" name="submit.Save" value="submit.Save">@T["Save Draft"]</button>
-}
-else
-{
-    <div class="btn-group">
-        <button class="btn btn-primary" type="submit" name="submit.Save" value="submit.Save">@T["Save Draft"]</button>
-        <button type="button" class="btn btn-primary dropdown-toggle dropdown-toggle-split" data-reference="parent" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-            <span class="sr-only">@T["Toggle Dropdown"]</span>
-        </button>
-        <div class="dropdown-menu">
-            <button class="dropdown-item" type="submit" name="submit.Save" value="submit.SaveAndContinue">@T["and continue"]</button>
+    if (String.IsNullOrWhiteSpace(returnUrl))
+    {
+        <button class="primaryAction btn btn-primary" type="submit" name="submit.Save" value="submit.Save">@T["Save Draft"]</button>
+    }
+    else
+    {
+        <div class="btn-group">
+            <button class="btn btn-primary" type="submit" name="submit.Save" value="submit.Save">@T["Save Draft"]</button>
+            <button type="button" class="btn btn-primary dropdown-toggle dropdown-toggle-split" data-reference="parent" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                <span class="sr-only">@T["Toggle Dropdown"]</span>
+            </button>
+            <div class="dropdown-menu">
+                <button class="dropdown-item" type="submit" name="submit.Save" value="submit.SaveAndContinue">@T["and continue"]</button>
+            </div>
         </div>
-    </div>
+    }
 }
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.SummaryAdmin.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.SummaryAdmin.cshtml
@@ -43,7 +43,7 @@
                         <button type="button" class="btn btn-secondary btn-sm dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
                             @T["Actions"]
                         </button>
-                        <div class="dropdown-menu dropdown-menu-right">
+                        <div id="actionsMenu" class="dropdown-menu dropdown-menu-right">
                             @await DisplayAsync(Model.ActionsMenu)
                         </div>
                     </div>
@@ -57,3 +57,15 @@
 {
     <div class="col primary">@await DisplayAsync(Model.Content)</div>
 }
+
+<script at="Foot" type="text/javascript">
+    $(function(){
+        
+        $("#actionsMenu.dropdown-menu").each(function(i, e){
+            var count = $(e).children().length;
+            if(count == 0){
+                $(this).parent().hide();
+            }
+        });
+    });
+</script>
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Views/Contents-Button-Actions.SummaryAdmin.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Views/Contents-Button-Actions.SummaryAdmin.cshtml
@@ -6,12 +6,23 @@
     var hasPublished = await ContentManager.HasPublishedVersionAsync(contentItem);
     var hasDraft = contentItem.HasDraft();
     var hasClonePermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.CloneContent, contentItem);
+    var hasDeletePermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.DeleteContent, contentItem);
+    var hasEditPermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.EditContent, contentItem);
+    var hasPreviewPermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.PreviewContent, contentItem);
+    var hasPublishPermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.PublishContent, contentItem);
 }
 
 @if (hasDraft)
 {
-    <a asp-action="Publish" asp-controller="Admin" asp-route-area="OrchardCore.Contents" asp-route-contentItemId="@contentItem.ContentItemId" asp-route-returnUrl="@FullRequestPath" itemprop="UnsafeUrl" class="dropdown-item btn-sm">@T["Publish Draft"]</a>
-    <a asp-action="Preview" asp-controller="Item" asp-route-area="OrchardCore.Contents" asp-route-contentItemId="@contentItem.ContentItemId" class="dropdown-item btn-sm">@T["Preview Draft"]</a>
+    if(hasPublishPermission)
+    {
+        <a asp-action="Publish" asp-controller="Admin" asp-route-area="OrchardCore.Contents" asp-route-contentItemId="@contentItem.ContentItemId" asp-route-returnUrl="@FullRequestPath" itemprop="UnsafeUrl" class="dropdown-item btn-sm">@T["Publish Draft"]</a>
+    }
+    
+    if(hasPreviewPermission)
+    {
+        <a asp-action="Preview" asp-controller="Item" asp-route-area="OrchardCore.Contents" asp-route-contentItemId="@contentItem.ContentItemId" class="dropdown-item btn-sm">@T["Preview Draft"]</a>
+    }
 
     <div class="dropdown-divider"></div>
 }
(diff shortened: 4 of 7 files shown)


src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.SaveDraftButton.cshtml
```diff
@@ -1,19 +1,26 @@
...
 @if (String.IsNullOrWhiteSpace(returnUrl))
 {
     <button class="primaryAction btn btn-primary" type="submit" name="submit.Save" value="submit.Save">@T["Save Draft"]</button>
 }
 else
 {
     <div class="btn-group">
-        <button class="btn btn-primary" type="submit" name="submit.Save" value="submit.Save">@T["Save Draft"]</button>
-        <button type="button" class="btn btn-primary dropdown-toggle dropdown-toggle-split" data-reference="parent" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+        <button class="btn btn-primary @((hasEditOwnContentPermission && contentItem?.Owner == User.Identity.Name) || hasEditContentPermission ? String.Empty : "disabled")" type="submit" name="submit.Save" value="submit.Save">@T["Save Draft"]</button>
```
hishamco, line 18: I'm not quite sure why the `Model.ContentItem` is `null`

deanmarcussen:
@hishamco I'm afraid this is not the correct solution.

The issue, as I explained to you on your last pr related to this, starts as a problem with the content type filter, which is designed to only list content types for which the user has the `Edit` permission.

The bug, or issue, is that when the filter is finds no editable content types, it is currently falling back to listing all content types.

That is the issue to resolve, and following that the issue would be how could a contributor list their own content.

One of the answers to that, is that they can't, and the role maybe more intended as a permission for contributors contributing via live writer, or other entry points to the system.

Note: This is how it was in O1 as well

hishamco:
> The bug, or issue, is that when the filter is finds no editable content types, it is currently falling back to listing all content types.

This is another issue, I started with the issue that raise during the demo of @agriffard last two weeks when he publish a content and got 404, then after checking the roles permissions I realize there is no issue with that, except the UI permissions so that's why I point to that in the original issue

I think we need to change the issue's title to avoid misunderstanding

hishamco:
We need to open another PR for the issue that you mentioned

deanmarcussen:
Ok maybe they are seperate issues. 

Could you read the comments in the code on how to check for content permissions. This is not the correct way

hishamco:
> Ok maybe they are seperate issues.

Yep, this is another one, I remembered the one that you meant, there is an open PR that needs a revise

> Could you read the comments in the code on how to check for content permissions. This is not the correct way

Which comments do you mean? If I'm not wrong `AuthorizeAsync()` is the way to check the permissions

src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.PublishButton.cshtml
```diff
@@ -1,22 +1,27 @@
-@*@if (Authorizer.Authorize(Permissions.PublishContent, (IContent)Model.ContentItem)) {*@
+@using Microsoft.AspNetCore.Authorization
+@inject IAuthorizationService AuthorizationService
+
 @{
+    ContentItem contentItem = Model.ContentItem;
     var returnUrl = Context.Request.Query["returnUrl"];
+    var hasPublishOwnContentPermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.PublishOwnContent, contentItem);
```
deanmarcussen, line 7: Please read the code comments @hishamco 

```
    public class CommonPermissions
    {
        // Note - in code you should demand PublishContent, EditContent, or DeleteContent
        // Do not demand the "Own" variations - those are applied automatically when you demand the main ones

        // EditOwn is the permission that is ultimately required to create new content. See how the Create() method is implemented in the AdminController
```

hishamco: Aha, I think I read a similar comment, but m question is how the non `xxxOwn` permission verify if the the current user is owner of the content, just to understand and satisfy myself

Thanks

sebastienros:
It's not clear what this PR is trying to fix, it seems unrelated to the issue that is linked. Can you open an issue and describe it?

hishamco:
Sure

src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.PublishButton.cshtml
deanmarcussen: Can you update this pr to remove all checks for Own variants, so we can review what it is doing. The PermissionHandler will morph permission checks as required to validate that the user is allowed access to the requested resource.

hishamco: I will remove the own checks if it's handled behind the scene

src/OrchardCore.Modules/OrchardCore.Contents/Views/Contents-Button-Actions.SummaryAdmin.cshtml
```diff
@@ -6,32 +6,17 @@
...
 
-    <div class="dropdown-divider"></div>
-}
-
-@if (hasDraft && hasPublished)
-{
-    <a asp-action="DiscardDraft" asp-controller="Admin" asp-route-area="OrchardCore.Contents" asp-route-contentItemId="@contentItem.ContentItemId" asp-route-returnUrl="@FullRequestPath" itemprop="UnsafeUrl RemoveUrl" data-title=@T["Discard Draft"] data-message=@T["Are you sure you want to discard this draft?"] data-ok-text="@T["Yes"]" data-cancel-text="@T["No"]" data-ok-class="btn-primary" class="dropdown-item btn-sm">@T["Discard Draft"]</a>
-}
-
-@if (hasPublished)
```
deanmarcussen, line 24: thank you @hishamco the pr becomes more understandable now.

So you are changing all of these to display as disabled, rather than not displaying at all.

I don't think that's the way to go here (but if it is there are many more permission checks you would have to change)

I would rather see it as it was, the button not shown.

i.e. here you are showing discard draft, where there is no draft to discard, which makes no sense.

src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsAdminList-Create.cshtml
```diff
@@ -1,8 +1,14 @@
...
 @model ContentOptionsViewModel
+@inject IContentManager ContentManager
+@inject IAuthorizationService AuthorizationService
+
 <div class="form-group">
     <div class="btn-group">
         @if (Model.CreatableTypes.Any())
         {
+            ContentItem contentItem = await ContentManager.NewAsync(Model.CreatableTypes?.First().Value);
+            var hasEditContentPermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.EditContent, contentItem);
```
deanmarcussen, line 11: this check should have already happened in the controller, and it shouldn't be in the `CreatableTypes` list.

If it isn't happening in the controller it should.

It would also have to happen on every type, not just the first one

src/OrchardCore.Modules/OrchardCore.Contents/Views/Contents-Button-Actions.SummaryAdmin.cshtml
hishamco: > So you are changing all of these to display as disabled, rather than not displaying at all.

I was think about that before, but hide them will make the buttons look ugly, so I prefer to disable them rather than hide them

> I would rather see it as it was, the button not shown.

Let me check them again, they was working well before I remove the check that you asked for, so let me check everything again

Thanks Dean

src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsAdminList-Create.cshtml
hishamco: Seems the check was there before I added my changes, I will remove it then

sebastienros: The list of CreatableTypes should just not contain the ones you can't create. And even that is a minor issue IMO if that complexifies the code too much (or make it too slow).

src/OrchardCore.Modules/OrchardCore.Contents/Views/Contents-Button-Actions.SummaryAdmin.cshtml
```diff
@@ -6,32 +6,17 @@
...
-    <a asp-action="Publish" asp-controller="Admin" asp-route-area="OrchardCore.Contents" asp-route-contentItemId="@contentItem.ContentItemId" asp-route-returnUrl="@FullRequestPath" itemprop="UnsafeUrl" class="dropdown-item btn-sm">@T["Publish Draft"]</a>
-    <a asp-action="Preview" asp-controller="Item" asp-route-area="OrchardCore.Contents" asp-route-contentItemId="@contentItem.ContentItemId" class="dropdown-item btn-sm">@T["Preview Draft"]</a>
+<a asp-action="Publish" asp-controller="Admin" asp-route-area="OrchardCore.Contents" asp-route-contentItemId="@contentItem.ContentItemId" asp-route-returnUrl="@FullRequestPath" itemprop="UnsafeUrl" class="dropdown-item btn-sm @(hasDraft && hasPublishPermission)? string.Empty : "disabled")">@T["Publish Draft"]</a>
+<a asp-action="Preview" asp-controller="Item" asp-route-area="OrchardCore.Contents" asp-route-contentItemId="@contentItem.ContentItemId" class="dropdown-item btn-sm @(hasDraft && hasPreviewPermission? string.Empty : "disabled")">@T["Preview Draft"]</a>
+<div class="dropdown-divider"></div>
 
-    <div class="dropdown-divider"></div>
-}
-
-@if (hasDraft && hasPublished)
```
sebastienros, line 19: Please don't change this behavior, I assume it was decided to not show the actions if they don't apply

hishamco: Ok

hishamco: > I assume it was decided to not show the actions if they don't apply

Dean prefers to hide the actions instead of disable them too, but when I hide the other actions before the commit the UI becomes ugly :( I can provide a screenshot

hishamco: UI Pemissions

deanmarcussen: I don't see the problem with hiding the `Edit` button here if the user does not have permission.

It is consistent to hide with all the other permission related ui, i.e. the menu looks pretty bare to, but that is what you would expect when you don't have permissions to execute an operation.

If you try to change this behaviour it will have to many ripple effects in other areas @hishamco

hishamco: I don't have any problems at all, but I find the UI/UX is not consistent, so I will hide the actions then

src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsAdminList-Create.cshtml
deanmarcussen: This code you've added needs to be in the controller, or not at all

src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.SummaryAdmin.cshtml
```diff
@@ -57,3 +57,15 @@
 {
     <div class="col primary">@await DisplayAsync(Model.Content)</div>
 }
+
+<script at="Foot" type="text/javascript">
+    $(function(){
+        
+        $(".dropdown-menu").each(function(i, e){
```
deanmarcussen, line 64: pop an id on this one, so it doesn't affect other `.dropdown-menu` on the page

src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsAdminList-Create.cshtml
hishamco: @deanmarcussen AFAIK when the permission is not applied in the controller we returns `Forbid`, so in this case how can I check the permission If I move it to the controller?

deanmarcussen: See where the list is being created

https://github.com/OrchardCMS/OrchardCore/blob/c8e7ea5cdf74042568f61465a21474338640b6fc/src/OrchardCore.Modules/OrchardCore.Contents/Controllers/AdminController.cs#L126

https://github.com/OrchardCMS/OrchardCore/blob/c8e7ea5cdf74042568f61465a21474338640b6fc/src/OrchardCore.Modules/OrchardCore.Contents/Controllers/AdminController.cs#L163

hishamco: I didn't dig into the code before, it's seems the check is already there, so I remove the one in the view

src/OrchardCore.Modules/OrchardCore.Contents/Views/Contents-Button-Edit.SummaryAdmin.cshtml
```diff
@@ -1,12 +1,23 @@
...
-<a edit-for="@contentItem" asp-route-returnUrl="@FullRequestPath" class="btn btn-primary btn-sm">@T["Edit"]</a>
+@if(hasEditContentPermission)
+{
+    <a edit-for="@contentItem" asp-route-returnUrl="@FullRequestPath" class="btn btn-primary btn-sm">@T["Edit"]</a>
+}
+
+@if (hasPublished)
+{
+    <a display-for="@contentItem" class="btn btn-success btn-sm" itemprop="">@T["View"]</a>
+}
```
deanmarcussen, line 20: Screenshot 2020-08-20 at 09 51 14

???

hishamco: Strange!! let me check again ...

hishamco: Nah!! the condition get copied twice during merge conflict

deanmarcussen (approved):

src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.PublishButton.cshtml
```diff
@@ -1,22 +1,28 @@
-@*@if (Authorizer.Authorize(Permissions.PublishContent, (IContent)Model.ContentItem)) {*@
+@using Microsoft.AspNetCore.Authorization
+@inject IAuthorizationService AuthorizationService
+
 @{
+    ContentItem contentItem = Model.ContentItem;
     var returnUrl = Context.Request.Query["returnUrl"];
+    var hasPublishContentPermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.PublishContent, contentItem);
```
Skrypt, line 7: PublishOwnContent. These permissions are not taking in consideration when we want to use PublishOwn or EditOwn they need to be changed.

hishamco: > These permissions are not taking in consideration when we want to use PublishOwn or EditOwn

Why?

src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.SummaryAdmin.cshtml
```diff
@@ -57,3 +57,15 @@
 {
     <div class="col primary">@await DisplayAsync(Model.Content)</div>
 }
+
+<script at="Foot" type="text/javascript">
```
jptissot, line 61: Why not use a named script with a class selector ? Instead of having tons of instances of this code on the page ?

deanmarcussen: @netwavebe already fixed it @jptissot exactly as you suggested.

src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.PublishButton.cshtml
deanmarcussen: @Skrypt code is good here I think. a PublishContent permission request is morphed to PublishOwn. You shouldn't request PublishOwn directly (see comments in CommonPermissions)

src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.SummaryAdmin.cshtml
jptissot: Ah cool, I did not realize this was an old pr :)

netwavebe: It got merged to dev a few days ago. See https://github.com/OrchardCMS/OrchardCore/pull/7070

jptissot: Thanks @netwavebe :)

src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.PublishButton.cshtml
Skrypt: It doesn't work if you have set a PublishOwn_ContentTypeName permission which is dynamic. So the issue is that the PR makes it impossible to have a role that has Publish/Edit/Delete own permission on a single custom content type. I have a PR coming up.

Skrypt: See https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.Contents/Security/ContentTypeAuthorizationHandler.cs#L58

chinasqzl:
ContentsDriver.Edit
Dynamic("Content_PublishButton").Location("Actions:10")
results.Add(Dynamic("Content_SaveDraftButton").Location("Actions:20"));

ContentItem contentItem = Model.ContentItem;
var hasPublishContentPermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.PublishContent, contentItem);



Model.ContentItem is null

hishamco:
@Skrypt can you please check this in your PR?

Skrypt:
The ContentItem passed to the ContentTypeAuthorizationHandler can sometimes be null but in this case here it's not supposed to.

In the ContentsDriver it assumes that it renders a ContentItem already. If the ContentItem == null then the ContentTypeAuthorizationHandler will simply return the previous Auth result without processing anything further. I looked last night and ContentTypeAuthorizationHandler is fine. Though here if the ContentItem == null then the ContentDriver should throw an exception and it's a different issue I think.

chinasqzl:
image
image

@Skrypt

Skrypt:
Yes please PR has moved to https://github.com/OrchardCMS/OrchardCore/pull/7104 now.
Here the issue comes from the ContentTypeAuthorizationHandler if you are not seeing the proper buttons.
This commit should fix your issue : https://github.com/OrchardCMS/OrchardCore/commit/79082938825cd0a40bcbc4756d497c738807ba81#diff-f32a3f59eaeef643f8774e25b903ff08R48
Though, I tested everything in https://github.com/OrchardCMS/OrchardCore/pull/7104

hishamco:
Is it regression bug or something else related?

Skrypt:
This PR fixes one part of the issues. I think we fixed both the same issues in 2 different PR's. Mine is going one step further by introducing 2 new permissions for the admin UI. I'm not sure it's a regression. I think the issue was there since a while and adding the permissions on the buttons just revealed the underlying issue.

PR #6826, opened 2020-08-04
Use admin url prefix instead of magic string

hishamco:


deanmarcussen:
hisham, hisham, hisham, what are you doing?

That property has no relationship to the url, it's related to the name of the `Pages/Admin` folder

You are trying to fix something that isn't broken...

hishamco:
> That property has no relationship to the url, it's related to the name of the Pages/Admin folder

Noooooo!! I though it's related to the URL not on the folder, I think we need some comments to avoid misunderstanding, furthermore it would be nice to use `const` or `readonly` for casing like this and admin menu name .. etc

PR #6832, opened 2020-08-04
Added Material Icon themes

### LINKED ISSUE #6831: Material Icon themes in Admin
Include material icons for all themes

ns8482e:
Fixes #6831  to support following

```html
<i class="material-icons">donut_small</i>
<i class="material-icons-outlined">donut_small</i>
<i class="material-icons-two-tone">donut_small</i>
<i class="material-icons-round">donut_small</i>
<i class="material-icons-sharp">donut_small</i>
```

### DIFF
--- a/src/OrchardCore.Themes/TheAdmin/Views/Layout.cshtml
+++ b/src/OrchardCore.Themes/TheAdmin/Views/Layout.cshtml
@@ -9,7 +9,7 @@
 
     <!-- Bootstrap CSS -->
     <style asp-name="admin" asp-src="~/TheAdmin/Styles/TheAdmin.min.css" debug-src="~/TheAdmin/Styles/TheAdmin.css" depends-on="jQuery-ui"></style>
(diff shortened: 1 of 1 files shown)


sebastienros:
Where do we use these icons in the admin?

ns8482e:
@sebastienros material icon's were already referenced in `Layout.cshtml` before this PR. I added support to include other icon themes.

sebastienros:
Couldn't find where we were using them

sebastienros:
Still can't btw ;) just to be clear

ns8482e:
@sebastienros I don't know if material icons are being used in OC source or not- But as It was defined in `Layout.cshtml` already, I was using in my admin templates, without defining new admin theme. 
 
But when I needed to change icon-theme, I proposed this PR.

Skrypt:
Yes but it's worth nothing if we are not using these icons at all in the admin theme.
If we plan using them then specify where and let's have a PR that includes these changes.
Else, here we add extra resources to the admin theme for nothing.

PR #6834, opened 2020-08-04
Bootstrap 4.5.2

agriffard:
https://github.com/twbs/bootstrap/releases/tag/v4.5.2


deanmarcussen:
few more days until we see if there is 4.5.3 ?

deanmarcussen:
Can you merge dev and update shortcodes as well (I had to add bootstrap to it as well), then we should merge it

PR #6846, opened 2020-08-06
Close button should clear filter by content type

### LINKED ISSUE #6844: Content Items - Manage Items Filter not cleared when filter text is cleared out with the 'x'
In the content items ->Manage content screen , if a text filter is applied and then cleared , the results without the filter are not shown. There is an 'x' to clear out the filter. It should reset and clear the filter to display all results.

hishamco:


hishamco:
ContentTypeFilter

@tabzone you could try my changes ...

tabzone:
@hishamco - tested and the content type filter did not work on SQLLite install . It does not filter on content type
artcicle
 now. 

Also the original bug I mentioned was for the filters textbox . i have attached the screenshot

https://recordit.co/XBvfkKFLPY

hishamco:
> Also the original bug I mentioned was for the filters textbox 

I check it before but there's not "x" button that is way I checked the drop down ;) anyhow I will remove the issue from my PR and I will check your recording and come back to you

hishamco:
As I saw from your screen cast the issue is not related to my PR, but my PR still valid to solve another issue ;) I will have a look again to the search textbox

hishamco:
ContentTypeFilter

tabzone:
The 'x' on the filter maybe coming from theme based on inspect css. I have enabled TheAdmin theme

hishamco:
> I have enabled TheAdmin theme

I'm with the same theme too ;) please check it in FireFox which is the once that I made all my tests on

hishamco:
@agriffard please your final review before merge

src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsAdminList-Filters.cshtml
```diff
@@ -24,5 +24,32 @@
             var newAction = "@formActionWithoutTypeName" + "/" + value;
             $("#items-form").attr("action", newAction);
         });
+
+        $('button.close').click(function(){
+            var url = window.location.href;
+            var newUrl = removeUrlParameter(url, 'Options.SelectedContentType');
```
deanmarcussen, line 30: No. `window.location.href` This would never be an acceptable solution.

The close button only closes the window, it doesn't clear the filter.
If it was to clear the filter (I'd be alright with that) it would activate the `All content types` link

hishamco: > No. window.location.href This would never be an acceptable solution.

Why?!! my intend to make all the content items visible, this will done after apply the filter at server-side, how this could be done on client-side?!!

agriffard:
The x button is to close the filter.
To reset the filter, click on 'All content types' .

PR #6849, opened 2020-08-07
Update trumbowyg svgPath

khurram4u2:
removed extra semicolon and wrapped the path in inverted comas.


jtkech:
Oops looks like your previous PR was merged too quickly, no problem ;)

PR #6857, opened 2020-08-08
Add scroll to top link into the admin site

### LINKED ISSUE #6724: Recommendation UX Improvement in Features sections / Admin
Once user going thru the Feature list, select required components he need to goes up till the beginning to enable / disable them.
Recommendation to add btn at the end.

image

hishamco:
Fixes #6724


hishamco:
Just struggling with gulp issue in my laptop, hope to fix it and the PR will be ready

hishamco:
@agriffard could you please do a `gulp run` coz I update the NodeJs but still have some issues when I run npm commands

deanmarcussen:
It looks like you have added this to every page in the admin where scrolling occurs.

Many pages, i.e. the edit content page, locate their buttons at the bottom of the screen, so there is no use for this extra scroll to top.

There's only a couple of pages where it's relevant, so for most pages, it's just a useless button

Perhaps you could add it just to the page indicated by the issue

hishamco:
@deanmarcussen the common scenario I saw in almost the CMSs that the go to the top button show up whenever the page has a scroll, so adding it on the admin layout is the suitable place if I'm not wrong, coz I don't wanna to add this on every page that we expect that a long content, coz the maintainability will be hard. I don't think we have an issue to add to the layout

src/OrchardCore.Modules/OrchardCore.Resources/wwwroot/Styles/bootstrap.css
```diff
@@ -4,10 +4,10 @@
 */
 
 /*!
- * Bootstrap v4.5.2 (https://getbootstrap.com/)
+ * Bootstrap v4.5.0 (https://getbootstrap.com/)
```
Skrypt, line 7: This looks like a regression here

sebastienros: Ping!

sebastienros: You might need to merge and rebuild

agriffard: This comment is on an outdated file.

The PR file has the correct version v4.5.2 merged.

Skrypt (approved):
LGTM

hishamco:
Thanks a lot @agriffard, but still I'm not sure why I'm facing gulp issue all the time ;)

deanmarcussen:
I don't know if your ui has changed from the screenshot you supplied but in that screenshot the scroll up obscures the buttons on the right of the screen.

Many pages will not benefit from this feature and in fact it will be worse on them rather than better.

This is why I don't want to see this added to the layout.

To unblock this pr please add it to the one screen where the issue was reported.

hishamco:
> To unblock this pr please add it to the one screen where the issue was reported.

Dean adding it to the features screen is not enough IMHO, because its useful in many other screen too such as content items, content types, content parts .. etc, I don't wanna repeat the code every time when we need it. One more thing the button will show up once the content don't fit into the page and browser scroll is appear

@Skrypt @sebastienros  your thoughts

Skrypt:
I would have added it to the general Layout.
It needs to be integrated without blocking existing UI elements.
I will take a look at the PR.

Skrypt:
I just saw the result and we can't merge this. The icon is too big and it should be integrated better. I don't see how we can use this kind of control in the admin without having it displaying on top of other content. The right margin of the content doesn't leave us a lot of space to get it displaying in that space either.

image

hishamco:
Thanks Jasmin for your feedback, ya the icon probably little bit big, I will minimize it, regarding the content it should be on top and that's how it looks in almost websites

Skrypt:
So here the issue is that we want to be able to quickly press on those Bulk Action buttons. We could instead just add a position: sticky to these and always display them as soon as they disapear from the ViewPort. That would remove the need of a scroll to top button. Don't waste too much time on this one. We made a mistake by thinking we should add this I think.

hishamco:
> We could instead just add a position: sticky to these and always display them as soon as they disapear from the ViewPort

I didn't understand how can sticking them will solve the issue if there are many items in the list, could you please elaborate?

Skrypt:
Make the buttons to stick on the header of the website like Github does here.

image

hishamco:
I got what you mean now ;) this could be a better option, let me update the PR and share a screenshot then

Skrypt:
I'd say, let's close this PR and start a fresh one.

hishamco:
@Skrypt I have a try to make this happen, here is how it look, your feedback

Sticky

Skrypt:
Beautiful 🥰

hishamco:
I just change the `Bulk Action` to be aligned with the search, so both can be sticky

hishamco:
I will push a PR soon, may I need you help in some bootstrap css classes ;)

PR #6860, opened 2020-08-09
Fixes manifest circular dependency

### LINKED ISSUE #6691: Manifest circular dependency
Just saw in the manifests this circular dependency

    OC.Contents => OC.Liquid => OC.Contents

Also the `ContainerFilter` defined in `OC.Liquid` may have to be defined in `OC.Lists`, didn't check other ones.

jtkech:
Fixes #6691 

- **I kept the dependency `Contents => Liquid`** because e.g. the `FullTextAspectSettingsDisplayDriver` injects `ILiquidTemplateManager`. So features having a direct / transitive dependency on `Contents` don't need a direct dependency on `Liquid`, this was already the case for most of them, and they don't need a custom startup with a `[RequireFeatures("Liquid")]` on a feature they depend on, so i did some cleanups.

    I also removed `OC.Liquid` from the recipes as they also contains the `OC.Contents` feature. Note: Because of transitive dependencies we could remove many other features, but maybe clearer to still show in recipes a more complete list.

- Then i moved some liquid filters as the `container` one from `OC.Liquid` to `OC.Lists`

- **I fixed the circular dependency by removing `Liquid => Contents`**, i moved the member access strategies related to content management in the static ctor of the `Contents` startup. I also removed the dependency `Liquid => Shortcodes` by using instead `[RequireFeatures("Shortcodes")]` to register the `shortcode` liquid filter.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Startup.cs
@@ -224,6 +224,8 @@ namespace OrchardCore.Contents
 
             services.AddLiquidFilter<BuildDisplayFilter>("shape_build_display");
             services.AddLiquidFilter<ContentItemFilter>("content_item_id");
+            services.AddLiquidFilter<DisplayTextFilter>("display_text");
+            services.AddLiquidFilter<DisplayUrlFilter>("display_url");
         }
     }
 
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Startup.cs
@@ -1,4 +1,5 @@
 using System;
+using Fluid;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Routing;
@@ -10,6 +11,7 @@ using OrchardCore.ContentManagement;
 using OrchardCore.ContentManagement.Display;
 using OrchardCore.ContentManagement.Display.ContentDisplay;
 using OrchardCore.ContentManagement.Handlers;
+using OrchardCore.ContentManagement.Metadata.Models;
 using OrchardCore.ContentManagement.Routing;
 using OrchardCore.Contents.AdminNodes;
 using OrchardCore.Contents.Controllers;
@@ -33,6 +35,7 @@ using OrchardCore.Deployment;
 using OrchardCore.DisplayManagement;
 using OrchardCore.DisplayManagement.Descriptors;
 using OrchardCore.DisplayManagement.Handlers;
+using OrchardCore.DisplayManagement.Views;
 using OrchardCore.Entities;
 using OrchardCore.Feeds;
 using OrchardCore.Indexing;
@@ -53,6 +56,17 @@ namespace OrchardCore.Contents
     public class Startup : StartupBase
     {
         private readonly AdminOptions _adminOptions;
+
+        static Startup()
+        {
+            TemplateContext.GlobalMemberAccessStrategy.Register<ContentItem>();
+            TemplateContext.GlobalMemberAccessStrategy.Register<ContentElement>();
+            TemplateContext.GlobalMemberAccessStrategy.Register<ShapeViewModel<ContentItem>>();
+            TemplateContext.GlobalMemberAccessStrategy.Register<ContentTypePartDefinition>();
+            TemplateContext.GlobalMemberAccessStrategy.Register<ContentPartFieldDefinition>();
+            TemplateContext.GlobalMemberAccessStrategy.Register<ContentFieldDefinition>();
+            TemplateContext.GlobalMemberAccessStrategy.Register<ContentPartDefinition>();
+        }
 
         public Startup(IOptions<AdminOptions> adminOptions)
         {
@@ -120,6 +134,14 @@ namespace OrchardCore.Contents
 
             services.AddScoped<IDisplayManager<ContentOptionsViewModel>, DisplayManager<ContentOptionsViewModel>>();
             services.AddScoped<IDisplayDriver<ContentOptionsViewModel>, ContentOptionsDisplayDriver>();
+
+            // Liquid
+            services.AddScoped<ILiquidTemplateEventHandler, ContentLiquidTemplateEventHandler>();
+
+            services.AddLiquidFilter<BuildDisplayFilter>("shape_build_display");
+            services.AddLiquidFilter<ContentItemFilter>("content_item_id");
+            services.AddLiquidFilter<DisplayTextFilter>("display_text");
+            services.AddLiquidFilter<DisplayUrlFilter>("display_url");
         }
 
         public override void Configure(IApplicationBuilder builder, IEndpointRouteBuilder routes, IServiceProvider serviceProvider)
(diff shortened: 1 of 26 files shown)


src/OrchardCore.Modules/OrchardCore.Taxonomies/Manifest.cs
```diff
@@ -19,5 +19,6 @@
     Id = "OrchardCore.Taxonomies.ContentsAdminList",
     Name = "Taxonomies Contents List Filters",
     Description = "Provides taxonomy filters in the contents list.",
+    Dependencies = new[] { "OrchardCore.Taxonomies" },
```
hishamco, line 22: Why we need to self dependency?!!

jtkech: It's not a self dependency, this is the `OrchardCore.Taxonomies.ContentsAdminList` feature that depends on the `OrchardCore.Taxonomies` feature

src/OrchardCore.Modules/OrchardCore.Contents/Startup.cs
```diff
@@ -1,4 +1,5 @@
 using System;
+using Fluid;
```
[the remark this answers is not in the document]
jtkech: No, System ones always at the top

src/OrchardCore.Modules/OrchardCore.Taxonomies/Manifest.cs
hishamco: My thought while the feature in the same module no need to reference another feature in the same module

src/OrchardCore.Modules/OrchardCore.Contents/Startup.cs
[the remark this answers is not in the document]
hishamco: I'm `Fluid` should come after `OrchardCore.*`

src/OrchardCore.Modules/OrchardCore.Taxonomies/Manifest.cs
jtkech: System ones at the top, others by alphabetic order, so Fluid before OC ones

I'm using under VStudio the `Remove and Sort Usings` cmd when you right click on the usings

Or `ctrl + RG` through the keyboard.

src/OrchardCore.Modules/OrchardCore.Contents/Startup.cs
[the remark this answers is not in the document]
jtkech: System ones at the top, others by alphabetic order, so Fluid before OC ones

I'm using under VStudio the `Remove and Sort Usings` cmd when you right click on the usings

Or `ctrl + RG` through the keyboard.

src/OrchardCore.Modules/OrchardCore.Taxonomies/Manifest.cs
jtkech: A feature may depend (or not) on another feature of the same module

We do that in many other modules

src/OrchardCore/OrchardCore.DisplayManagement/Shapes/PageTitleShapes.cs
```diff
@@ -32,7 +31,7 @@ public IPageTitleBuilder Title
         }
 
         [Shape]
-        public async Task<IHtmlContent> PageTitle(IHtmlHelper Html)
+        public async Task<IHtmlContent> PageTitle()
```
deanmarcussen, line 34: this isn't your change @jtkech but is shouldn't need to be using `ShellScope` as the service provider is available on the `Context` when injected

src/OrchardCore.Modules/OrchardCore.Contents/Startup.cs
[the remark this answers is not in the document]
hishamco: I knew ;) `Remove and Sort Usings` Order them alphabetically, AFAIK `System.*` come first, then `Microsoft.*` after that `OrchardCore.*`

src/OrchardCore.Modules/OrchardCore.Taxonomies/Manifest.cs
hishamco: Ok, thanks

src/OrchardCore.Modules/OrchardCore.Contents/Startup.cs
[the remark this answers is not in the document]
jtkech: Okay, why not and i'm open to any ordering rules ;)

But currently our `.editor.config` file only contains

    # Sort using and Import directives with System.* appearing first
    dotnet_sort_system_directives_first                        = true

src/OrchardCore/OrchardCore.DisplayManagement/Shapes/PageTitleShapes.cs
jtkech: @deanmarcussen 

Yes i agree by using `PageTitle(DisplayContext DisplayContext)`.

Anyway, `PageTitleShapes : IShapeAttributeProvider` is a scoped service, why not just injecting the needed services in its ctor. I tried it and it works, but i noticed that the ctor is called twice. I found why, it is resolved through an `IEnumerable<>` in `ShapeAttributeBindingStrategy` and also on demand through the shape binding delegate.

This is because

    public static IServiceCollection AddShapeAttributes<T>(this IServiceCollection services) ...
    {
        services.AddScoped<T>();
        services.AddScoped<IShapeAttributeProvider, T>();

        return services;
    }

In a given scope not good to have multiple instances of the same scoped service, so i think i will do another PR to fix it by using the following

    public static IServiceCollection AddShapeAttributes<T>(this IServiceCollection services) ...
    {
        services.AddScoped<T>();
        services.AddScoped<IShapeAttributeProvider>(sp => sp.GetRequiredService<T>());

        return services;
    }

src/OrchardCore.Modules/OrchardCore.Contents/Startup.cs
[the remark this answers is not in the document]
hishamco: If is that go ahead, seems I'm influenced by what I saw and deal with in ASP.NET Core repo ;)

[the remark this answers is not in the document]
jtkech: Okay cool ;)

[See here](https://github.com/dotnet/aspnetcore/blob/0ace9d91b63967de686414509c17c1850ea5e500/src/SignalR/server/StackExchangeRedis/src/Internal/RedisProtocol.cs#L10)

jtkech:
@deanmarcussen 

> So essentially Contents now requires Liquid

**Yes but it was already the case, i didn't change the `OC.Contents` manifest**

> there is no two ways about it, due to the full text driver etc?

If we really want i think we could remove this dependency but not with an immediate solution, e.g. by splitting the full text settings driver with one requiring Liquid. Also right now many features needing Liquid only have a transitive dependency on it through Contents. So i opted to keep this dependency and removed the few remaining direct dependency on Liquid that right now are useless.

Note: I noticed that the `PageTitle` core shape also uses `ILIquidTemplateManager`

> It would be nice to log a warning when there is a circular dependency detected. It is already hard to know you are not creating one when you add a dependency to the graph. (only if it is easy and trivial to do)

Yes good idea, we would have to focus again on the `OrderByDependenciesAndPriorities()` utility

> I also wondered about debug logging startup orders, only because I had to do it the other day, to figure out where I needed to put some middleware to get in between a couple of modules (auth and media)

Good idea, personally in `ModularTenantRouterMiddleware` in debug mode i put a break point line 110 when we build the tenant pipeline, then i lookup the non public `_components` property of the `appBuilder`.

Otherwise you can lookup to the overriden `Order` or `ConfigureOrder` properties, being 0 by default, then alphabetical order, knowing that those added from the app with our `Configure()` helper are enlisted first for a same order value.

PR #6899, opened 2020-08-13
Useless OrderBy() and IsDefaultShell() in OC.Tenants

jtkech:


Jetski5822:
TryGetSettings doesnt do case insensitive matching, do we care?

jtkech:
Hi @Jetski5822 

Where i used it normally it should match exactly, but in `ValidateViewModel()` we still do case insensitive checkings

I also thought about making the `_shellContexts` itself a case insensitive concurrent dictionary, but didn't do it for perf

jtkech:
@Jetski5822 because you did the `ApiController` ;)

In the `ApiController` when creating a tenant we don't do all the checks, e.g. if the `same host and prefix already exists`, if the `prefix contains more than one segment`. Do you think that we should do the same checks as in the `AdminController` in `ValidateViewModel()`? Or has this been done for perf reasons?

As said above, in the `AdminController` in this PR we still do a case insensitive check for the tenant name unicity, but in the `ApiController` when creating a tenant we use the case sensitive `_shellHost.TryGetSettings()`. Was it again for perf reasons? Or do you think we should be case insensitive as in the `AdminController`?

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs
```diff
@@ -95,12 +95,12 @@ public async Task<IActionResult> Create(CreateApiViewModel model)
             shellSettings["Secret"] = Guid.NewGuid().ToString();
             shellSettings["RecipeName"] = model.RecipeName;
 
-            if (!IsDefaultShell() && string.IsNullOrWhiteSpace(shellSettings.RequestUrlHost) && string.IsNullOrWhiteSpace(shellSettings.RequestUrlPrefix))
+            if (String.IsNullOrWhiteSpace(shellSettings.RequestUrlHost) && String.IsNullOrWhiteSpace(shellSettings.RequestUrlPrefix))
```
jtkech, line 98: I removed `IsDefaultShell()` here because it is already checked before line 69

            if (!IsDefaultShell())
            {
                return Forbid();
            }

jtkech:
Finally, in `OC.Tenants.AdminController` i reused case insensitive lookups for tenant names, as it was before, so that both `OrchardCore.Tenants/Edit/Tenant1` and `orchardcore.tenants/edit/tenant1` are working.

So here

- In `OC.Tenants.AdminController` I just removed an useless `.OrderBy()`.

- In `OC.Tenants.ApiController` I just removed an useless `!IsDefaultShell()` that is already done before.

- And just replaced some `string` with the framework type `String`

PR #6900, opened 2020-08-13
Remove the .svg from the allowed extensions

hishamco:


deanmarcussen:
In principle an easy fix to the solution, but also just offloads the issue onto users

Can we give this one a few days, I would like the time to do some research and see how others handle the issue.

hishamco:
I will look into that too ...

hishamco:
@deanmarcussen what if I sanitize the svg content before the uploading the file, so this way we can prevent such attack, I can modify the PR if sound looks good to you

ns8482e:
May be we could sanitize similar to html field

deanmarcussen:
@hishamco It might be worth another look (as until IE became an issue, the CSP was looking to be the better solution), but it concerns me (I already had a look, and feel it's problematic)

The sanitizer is opt-in on tags/attributes. It doesn't handle svg's by default, so you have to start adding all the svg related tags, and it has a tendency to reorganise markup.

The problem is there are a lot of tags/attributes for svg's re here https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute and here https://developer.mozilla.org/en-US/docs/Web/SVG/Element

My concern is that even adding all these tags / attributes and then sanitizing it, we are essentially becoming an svg editor, and I don't really want to have to maintain that complication, as we're a cms, not an svg editor, nor do I want to have the scenario where we miss one, and get more bugs for it.

hishamco:
> The sanitizer is opt-in on tags/attributes. It doesn't handle svg's by default, so you have to start adding all the svg related tags, and it has a tendency to reorganise markup.

My intend to read the svg content as string in the client-side and apply the sanitizing, so we don't need to maintain all the svg related tags

Skrypt:
We can't just remove .svg's from the allowed extensions.

PR #6903, opened 2020-08-14
Add option to allow Lucene query syntax in search

### LINKED ISSUE #6829: Add option to not escape search queries
Right now, all search queries are [escaped](https://github.com/OrchardCMS/OrchardCore/blob/acee23ef121a09568198ebda164d591ae9f084ed/src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/SearchController.cs#L123) before being sent to Lucene. This makes sense in most cases, but in some cases, it would be useful to let the user write Lucene query parser syntax directly.

An option should be added to Orchard Core to turn off search query escaping. This could allow easy implementation of advanced searches (e.g. the user could search for "title:search AND tag:lucene").

I am willing to work on a pull request for this feature.

Here's an example of what this could look like:

search-escape-setting
---
advanced-search

thomasebsmith:
Closes #6829.

This PR adds a setting to allow Lucene query syntax in search forms. The setting defaults to false.

I'm not sure what the behavior should be when a user enters invalid Lucene syntax. Right now, the error is caught and no results are returned to the user.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/SearchController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/SearchController.cs
@@ -5,6 +5,7 @@ using System.Threading.Tasks;
 using Lucene.Net.QueryParsers.Classic;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Localization;
 using Microsoft.Extensions.Logging;
 using OrchardCore.ContentManagement;
 using OrchardCore.ContentManagement.Records;
@@ -32,6 +33,7 @@ namespace OrchardCore.Lucene.Controllers
         private readonly LuceneAnalyzerManager _luceneAnalyzerManager;
         private readonly ISearchQueryService _searchQueryService;
         private readonly ISession _session;
+        private readonly IStringLocalizer S;
         private readonly IEnumerable<IPermissionProvider> _permissionProviders;
         private readonly dynamic New;
         private readonly ILogger _logger;
@@ -45,6 +47,7 @@ namespace OrchardCore.Lucene.Controllers
             LuceneAnalyzerManager luceneAnalyzerManager,
             ISearchQueryService searchQueryService,
             ISession session,
+            IStringLocalizer<SearchController> stringLocalizer,
             IEnumerable<IPermissionProvider> permissionProviders,
             IShapeFactory shapeFactory,
             ILogger<SearchController> logger
@@ -58,6 +61,7 @@ namespace OrchardCore.Lucene.Controllers
             _luceneAnalyzerManager = luceneAnalyzerManager;
             _searchQueryService = searchQueryService;
             _session = session;
+            S = stringLocalizer;
             _permissionProviders = permissionProviders;
             New = shapeFactory;
             _logger = logger;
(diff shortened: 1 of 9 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=6903) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=6903) before we can accept your contribution.<br/><br/>:x: thomasebsmith [sign now](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=6903)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OrchardCMS/OrchardCore?pullRequest=6903) it.</sub>

src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/SearchController.cs
```diff
@@ -137,8 +136,24 @@ public async Task<IActionResult> Search(SearchIndexViewModel viewModel, PagerSli
...
 
-            var contentItemIds = (await _searchQueryService.ExecuteQueryAsync(query, searchSettings.SearchIndex, start, end))
-                .ToList();
+            var terms = viewModel.Terms;
+            if (!searchSettings.AllowLuceneQueriesInSearch)
+            {
+                terms = QueryParser.Escape(terms);
+            }
+
+            IList<string> contentItemIds;
```
deanmarcussen, line 145: Why are we using List here not `IEnumerable and Enumerable.Empty<string>()`?

I know it's not your change @thomasebsmith but is there something about lucene that requires us to call `.ToList()` (@Skrypt ?) 

I think for errors, we should probably log them at least, and I _think_ it would be ok to add a generic `Incorrect query syntax` error to the `ModelState` dictionary for the `Terms` property

Skrypt: I don't like this code either. We should never use a Try / Catch to silently avoid a potential error without logging it or returning an error message to the user so that he understands that his query failed to return some result(s). 

@deanmarcussen The ToList() is necessary here if we look later we are doing contentItemIds.IndexOf() with it which would require an Extension of the IEnumerable so that it works which would not make any sense since ...

thomasebsmith: @deanmarcussen Actually, calling `.ToList()` was part of one of my previous changes (#6798). It is needed for `contentItemIds.IndexOf`, which is called later in the controller.

src/OrchardCore.Themes/TheAgencyTheme/Views/Shared/Search.liquid
```diff
@@ -9,6 +9,12 @@
 
 {{ Model.SearchForm | shape_render }}
 
+{% unless Model.SearchResults %}
+  {% if Model.Terms %}
+    {% helper "span", validation_for: "Terms", class: "alert alert-danger" %}
+  {% endif %}
+{% endunless %}
+
```
thomasebsmith, line 17: This liquid block and the similar one in the Blog theme template are needed since I can't find a Liquid equivalent of `ModelState.IsValid`. If there is no search results shape but there are search terms, there must be a validation error.

deanmarcussen: @Skrypt can you review? I am happy with the update from @thomasebsmith 

Sorry @thomasebsmith I think I already hassled you on the last pr about the `ToList()` :)

agriffard:
@Skrypt Can you please review this PR?

src/OrchardCore.Themes/TheAgencyTheme/Views/Shared/Search.liquid
Skrypt: I need to look at this to approve PR first. It's probably working but maybe there's a better way to do this.

src/OrchardCore.Modules/OrchardCore.Lucene/Views/_ViewImports.cshtml
```diff
@@ -3,6 +3,7 @@
 
 *@
 
+
```
Skrypt, line 6: Remove extra lines, keep only one.

Skrypt (approved):
LGTM

PR #6921, opened 2020-08-16
Fixes async / await usage

### LINKED ISSUE #6919: Async / MARS issues editing part definitions with queries on SQL Server
Only occurs on SQL Server

Repro 
- Create a ListPart
- Edit settings to `EnableOrder`

4 times out of 5 the ADO connection will throw an exception `The connection does not support MultipleActiveResultSets`
Screenshot 2020-08-16 at 17 28 11

Enable MARS (not that I want to) and the exception changes to 

"The transaction operation cannot be performed because there are pending requests working on this transaction"

The query being executed that causes this is from the `IContainerService.SetInitialOrder()` 

```
var contanerContentItemsQuery = _session.QueryIndex<ContentItemIndex>(x => x.ContentType == contentType && (x.Published || x.Latest));
            var containerContentItems = await contanerContentItemsQuery.ListAsync();
```

Of note, this is not using the Database Content Definition Store. It's using the File Definition Store.
So there is no other queries going to the database, nor is there any `_session.Save()` occuring.

I think the issue is happening because of async void calls in the `DefaultContentDefinitionDisplayManager`

This is the method causing issues

```
        public async Task<dynamic> UpdateTypePartEditorAsync(ContentTypePartDefinition contentTypePartDefinition, IUpdateModel updater, string groupId = "")
        {
            if (contentTypePartDefinition == null)
            {
                throw new ArgumentNullException(nameof(contentTypePartDefinition));
            }

            dynamic typePartDefinitionShape = await CreateContentShapeAsync("ContentTypePartDefinition_Edit");
            var layout = await _layoutAccessor.GetLayoutAsync();

            _contentDefinitionManager.AlterTypeDefinition(contentTypePartDefinition.ContentTypeDefinition.Name, typeBuilder =>
            {
                typeBuilder.WithPart(contentTypePartDefinition.Name, async typePartBuilder =>
                {
                    typePartDefinitionShape.ContentPart = contentTypePartDefinition;

                    var partContext = new UpdateTypePartEditorContext(
                        typePartBuilder,
                        typePartDefinitionShape,
                        groupId,
                        false,
                        _shapeFactory,
                        layout,
                        updater
                    );

                    //  BindPlacementAsync(partContext).GetAwaiter().GetResult();

                    //  _handlers.InvokeAsync((handler, contentTypePartDefinition, partContext) => handler.UpdateTypePartEditorAsync(contentTypePartDefinition, partContext), contentTypePartDefinition, partContext, _logger).GetAwaiter().GetResult();

                    await  BindPlacementAsync(partContext);

                    await _handlers.InvokeAsync((handler, contentTypePartDefinition, partContext) => handler.UpdateTypePartEditorAsync(contentTypePartDefinition, partContext), contentTypePartDefinition, partContext, _logger);
                });
            });

            return typePartDefinitionShape;
        }
```

Swap the async / await on the `typeBuilder.WithPart` for a `GetAwaiter().GetResult()` and the issue is resolved.

Problem is I think the async voids are just heading off into never never land, and not completing before the session is trying to Commit.

That's why I say 4 times out of 5.

If you put it in the debugger (i.e. give it a chance to complete the async database queries), it never fails.

The `ContentDefinitionDisplayManager` is actually peppered already with `.GetAwaiter().GetResult()` calls on some of the other methods, and async await void on some of the others.

I guess it's not a heavy usage api, so `.GetAwaiter().GetResult()` is probably better than breaking the api by turning it async, but there's probably a nicer way to fix it.

Any ideas @jtkech ?

jtkech:
Fixes #6919

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentTypes/Editors/DefaultContentDefinitionDisplayManager.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentTypes/Editors/DefaultContentDefinitionDisplayManager.cs
@@ -82,7 +82,7 @@ namespace OrchardCore.ContentTypes.Editors
 
             var layout = await _layoutAccessor.GetLayoutAsync();
 
-            _contentDefinitionManager.AlterTypeDefinition(contentTypeDefinition.Name, typeBuilder =>
+            await _contentDefinitionManager.AlterTypeDefinitionAsync(contentTypeDefinition.Name, async typeBuilder =>
             {
                 var typeContext = new UpdateTypeEditorContext(
                     typeBuilder,
@@ -94,9 +94,9 @@ namespace OrchardCore.ContentTypes.Editors
                     updater
                 );
 
-                BindPlacementAsync(typeContext).GetAwaiter().GetResult();
+                await BindPlacementAsync(typeContext);
 
-                _handlers.InvokeAsync((handler, contentTypeDefinition, typeContext) => handler.UpdateTypeEditorAsync(contentTypeDefinition, typeContext), contentTypeDefinition, typeContext, _logger).GetAwaiter().GetResult();
+                await _handlers.InvokeAsync((handler, contentTypeDefinition, typeContext) => handler.UpdateTypeEditorAsync(contentTypeDefinition, typeContext), contentTypeDefinition, typeContext, _logger);
             });
 
             return contentTypeDefinitionShape;
@@ -140,7 +140,7 @@ namespace OrchardCore.ContentTypes.Editors
             UpdatePartEditorContext partContext = null;
             var layout = await _layoutAccessor.GetLayoutAsync();
 
-            _contentDefinitionManager.AlterPartDefinition(contentPartDefinition.Name, partBuilder =>
+            await _contentDefinitionManager.AlterPartDefinitionAsync(contentPartDefinition.Name, async partBuilder =>
             {
                 partContext = new UpdatePartEditorContext(
                     partBuilder,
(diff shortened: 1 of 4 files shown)


sebastienros:
Not the good fix. We should make the methods sync instead. Breaking change... Or at least another overload and we can change our own usages.

jtkech:
@sebastienros Sorry i don't understand

E.g. `UpdateTypePartEditorAsync()` is async, it calls e.g. `await CreateContentShapeAsync()`

Then it calls `_contentDefinitionManager.AlterTypeDefinition()` that has an `Action<>` parameter (the `alteration` action parameter) that can't be async, **and this is for this delegate that we were using async / await** which was resulting in non awaited tasks

In 2 other places in the same file, the delegate of an `.AlterSomething()` already uses `.GetAwaiter().GetResult()`, but not in 2 other places that i updated in this PR.

deanmarcussen:
@jtkech maybe we should add extra extension methods to support a `Func<T,Task>` instead.

They are only extension methods so should be able to work as an overload without breaking anyone.

jtkech:
@deanmarcussen

Yes, in an old PR i already suggested to make all async around content definition, including `UpdateContentDefinitionRecord()` that currently uses a `.GetAwaiter().GetResult()`

In fact, the 2 async / await usage we are fixing are not in the action delegate of `.AlterTypeDefinition()`, but in the action delegate of `.WithPart()` or `WithField()`, same issue but there are not extensions, but yes we could also create `.WithPartAsync()` and `WithFieldAsync()` builder extensions.

Maybe in another PR, here it just fixes an async / await usage in a delegate related to an `Action<>`, i already fixed the same kind of [issue here](https://github.com/OrchardCMS/OrchardCore/commit/4c0071e42df9f4751c3854047d688c3c2583ada1#diff-9074dde53100bfc5a0d91bbb75dde7b6) a while ago, but we missed these two ones. Anyway, if nobody does it, i will merge it myself ;)

Hmm, @sebastienros when you said

> Not the good fix. We should make the methods **sync** instead

Did you mean?

> Not the good fix. We should make the methods **async** instead

Only one character changes the whole meaning, if so i now understand and i will go this way, sorry ;)

I didn't do that immediately as before you were not always a fan of making async things around content definition.

jtkech:
Okay i made the change, so more things are async

But a little more changes, so @deanmarcussen if you can check that it still fixes your issue, if you have time ;)

As said, there are many other things that are not async around content definition, but at some point it was intentionally kept like this because the content definition record is cached, and to have a less overhead of awaited `Task`

deanmarcussen:
@jtkech works for me, and I put a Task.Delay(1000) in on one of the queries, and could see it was still properly awaited before the session tried to dispose, so all good.

I'm assuming Seb meant `async` rather than sync as well, and while there's a little more code for us to maintain with both a normal version, and an async version, this is not code that needs to be maintained much, so I think it makes sense.

So all looks good to me, unless @sebastienros has meant something different :)

Yes, I understand about not making the content definitions themselves not async

jtkech:
I merged it as it fixes an annoying issue

PR #6923, opened 2020-08-17
Check for groupings existance in ContentZone

### LINKED ISSUE #6922: Index out of range from ContentZoneShape
After  upgrading to latest dev branch Im getting an error from the contentzone Shape , when evaluating for Cards is called :

image

Its coming from my Content.cshtm and specifically from the call to : 

```
@await DisplayAsync(Model.Header)
```

My header zone is empty and in the ContentZone shape there is no code logic if groupings.Count =0 so it fails at the line with groupings[0].

Ti fix, I simply added  :

```
  if ( groupings.FirstOrDefault() != null){
        // Evaluate for cards.

  }
```
 
and it worked.

@deanmarcussen is this an appropriate fix ?

giannik:
fixes #6922 
Check for groupings existance in ContentZone shape before evaluating for Cards

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement/Zones/ZoneShapes.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/Zones/ZoneShapes.cs
@@ -103,20 +103,12 @@ namespace OrchardCore.DisplayManagement.Zones
 
                 htmlContentBuilder.AppendHtml(await DisplayAsync(container));
             }
-            else
+            else if (groupings.Count == 1)
             {
-               if ( groupings.Count == 1) {
-                   // Evaluate for cards.
-                    var cardGrouping = await ShapeFactory.CreateAsync("CardGrouping", Arguments.From(
-                        new
-                        {
-                            Grouping = groupings[0],
-                            ContentItem = Shape.ContentItem
-                        }));
-
-                    htmlContentBuilder.AppendHtml(await DisplayAsync(cardGrouping));
-               }
-                
+                // Evaluate for cards.
+                var cardGrouping = await ShapeFactory.CreateAsync("CardGrouping", Arguments.From(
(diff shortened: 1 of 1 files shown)


src/OrchardCore/OrchardCore.DisplayManagement/Zones/ZoneShapes.cs
```diff
@@ -105,15 +105,18 @@ public async Task<IHtmlContent> ContentZone(dynamic DisplayAsync, dynamic Shape,
...
             else
             {
-                // Evaluate for cards.
-                var cardGrouping = await ShapeFactory.CreateAsync("CardGrouping", Arguments.From(
-                    new
-                    {
-                        Grouping = groupings[0],
-                        ContentItem = Shape.ContentItem
-                    }));
+               if ( groupings.FirstOrDefault() != null){
```
deanmarcussen, line 108: Yeah I can see how it's possible to still end up with an empty zone, if you've removed the `TitlePart` manually for example.

The better change here (I think I originally had it in, and then removed it, would be

```
else if (groupings.Count == 1)
{
  var cardGrouping ... etc
}
```

src/OrchardCore/OrchardCore.DisplayManagement/Zones/ZoneShapes.cs
```diff
@@ -105,7 +105,7 @@ public async Task<IHtmlContent> ContentZone(dynamic DisplayAsync, dynamic Shape,
             }
             else
```
deanmarcussen, line 106: can you make it `else if (groupings.Count == 1)` here, rather than another if block inside it

giannik: Oh, of course.Done.

Issue #6954, opened 2020-08-24
Display Configuration options in Admin pages


agriffard:
Create admin pages to display the different IOptions<> that can be configured in the application.

These options can be set in appsettings.json, in ShellSettings by tenant, in Database.

Ex: MediaOptions
Will help to check what are the current properties values.

Simple form page with for each property:
- Name
- Value
- Hint

Add link to documentation (and create if it does not exists).

Provide help to know how to define/override configuration (Name of section, json example).

Use a specific Permission (only for Administrator) if sensitive data.

Find a way to display a menu or a link somewhere in the admin.

Todo: List all options that can be configured.

Please comment to say if you find this useful or not or if you want to add a suggestion.

agriffard:
See note in https://docs.orchardcore.net/en/dev/docs/reference/core/Configuration/#ioptions-configuration for SMTP Settings

> On the admin there will be no indication that this override happened, and the value displayed there will still be the one configured in site settings, so if you choose to do this you'll need to let your users know.

Find a way to display the UI accordingly if it is configured in appsettings.

PR #6955, opened 2020-08-24
Options Admin pages

agriffard:
Related to #6954


agriffard:
/Admin/Media/Options

image

deanmarcussen:
This is useful, I am quite often looking to see what sizes I allowed when writing templates etc.

We need to find somewhere to put it in the menu.

I would like to see the Media Cache menu move to `Configuration`, it doesn't add up to me for it to be in `Content`.

What about `Configuration -> Media -> Options` and `Configuration -> Media -> Media Cache`

You could put the same alert notification I did on Media Cache at the top to say it's configured with `appsettings.json` maybe, so it's clear.

If we did it for some of the other ones, like `Azure Media` we'd need to have a`ViewModel` for the Options, so that the passwords weren't passed, but we could fill the value with a placeholder, so that you know a value has been supplied.

agriffard:
image

agriffard:
I don't find another good example of Options to display in an Admin page.

deanmarcussen:
How about the Azure blob options?

They have passwords, which we would want to not send to the view, but it would be good to show they have a value (getting it wrong is the most common reason it doesn't work)

agriffard:
I tried to add the Azure Blob options but I don't know why the page /Admin/MediaAzureBlob/Options is not found.

agriffard:
I don't understand why the admin menu is not displayed, I can't access the page and the Permission is not in the Administrator Role.

deanmarcussen:
> I don't understand why the admin menu is not displayed, I can't access the page and the Permission is not in the Administrator Role.

Fixed, and changed the `Options` menu to `Media Options` (as when they both show it was a bit odd.)

Nice idea with the password, but doesn't work. It could be `AccountKey` or it could be a SAS signature which is tricker to do with, ie `https://xxx.blob.core.windows.net/dev-media?sv=2019-02-02&st=2020-09-19T16%3A38%3A05Z&se=2020-09-20T16%3A38%3A05Z&sr=c&sp=rl&sig=kOyAC7DzL04naC7dOY0dudVN10t0jp%2FGp0%2BJffVAXck%3D

Also the booleans don't show.

Screenshot 2020-09-19 at 17 47 29

Maybe I was wrong and we shouldn't show azure blob options at all?

Or we should be just saying that it has a value or doesn't have a value. That's the biggest need I think. (Saves debugging to see if you've managed to configure appsettings correctly.)

What do you think?

agriffard:
I can display the page but I still don't see the permission or the admin menu.

I don't know if we need to display the connection string or the fact that it has been configured.

Should we create Options pages for `OrchardCore.Redis`, `OrchardCore_Shells_Azure`, `OrchardCore_Shells_Database` and  `OrchardCore_DataProtection_Azure`?

deanmarcussen:
Is the blob feature enabled?

agriffard:
> Is the blob feature enabled?

Yes.

agriffard:
ok, I got it now.
image

deanmarcussen:
> Should we create Options pages for OrchardCore.Redis, OrchardCore_Shells_Azure, OrchardCore_Shells_Database and OrchardCore_DataProtection_Azure

- OrchardCore_Shells_Azure, OrchardCore_Shells_Database NO - they're host level only.
- OrchardCore_DataProtection_Azure Yes, Later is fine. As per the blob
- OrchardCore.Redis- not sure need to look (might be tenant level, might not). could be later if useful.

For the Azure.Blob how about the UI is

`ConnectionString` : `Configured` or `Not Configured` 
if `String.Empty(options.ConnectionString)` or not, so not showing password (or part of password at all)?

The Create Container setting needs to be shown as a bool too ;)

Sorry I forgot about this one.

agriffard:
@deanmarcussen Changes applied but there is a little inconvenience:
I don't see the admin menu link anymore and I don't see the ViewAzureMediaOptions permission in the Administrator role.

src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
```diff
@@ -440,5 +437,15 @@ public object CreateFileResult(IFileStoreEntry mediaFile)
...
         }
+
+        public IActionResult MediaApplication()
+        {
+            return View();
+        }
+
+        public IActionResult Options()
+        {
+            return View(_mediaOptions);
```
deanmarcussen, line 448: This one just needs a permission check

src/OrchardCore.Modules/OrchardCore.Media.Azure/Controllers/AdminController.cs
```diff
@@ -0,0 +1,31 @@
...
+        private readonly MediaBlobStorageOptions _options;
+
+        public AdminController(IOptions<MediaBlobStorageOptions> options)
+        {
+            _options = options.Value;
+        }
+
+        public IActionResult Options()
+        {
+            var model = new OptionsViewModel
```
deanmarcussen, line 20: also needs a permission check
`_authorizationService.AuthorizeAsync(User, Permissions.ViewAzureMediaOptions);`

agriffard: Ok, permissions added.

deanmarcussen (approved):

PR #4063, opened 2019-08-19
Handle media max file limit at client-side

hishamco:
Fixes #3962

PR #6973, opened 2020-08-27
Use package.json for Media File Upload

agriffard:
See https://github.com/OrchardCMS/OrchardCore/pull/4063#issuecomment-612583250

Reference npm package: https://www.npmjs.com/package/blueimp-file-upload for jquery.fileupload.js in OrchardCore.Media


deanmarcussen:
Looks good, does drag and drop still work? I remember looking at this ages ago, and I thought that the blueimp version we used was a bit older and they _might_ have changed some of the event names

agriffard:
Drag and drop is working to upload a media.

deanmarcussen:
@agriffard I tweaked this a bit - it worked, but didn't when the media app was rebuilt, because of script ordering from the validate plugin.

I've moved the files to `node_modules` and removed the validate plugin, as it would still be unused at this time.

For future reference to use it, we need these two plugins.
https://github.com/blueimp/jQuery-File-Upload/blob/master/js/jquery.fileupload-process.js
https://github.com/blueimp/jQuery-File-Upload/blob/master/js/jquery.fileupload-validate.js

and then we need to configure it to use the values from our `MediaOptions` class

Issue #6645, opened 2020-07-12
Manage Own Media issue


staymo:
Just found that the permission "Manage Own Media" is not acting as what it means -- user with this permission can view/edit/delete all the media files. Is this a bug?
To reproduce,  just create a new user, don't assign it to any role, then give the role authenticated following permissions:
View all content;
Access admin panel;
Manage Own Media

version:     <PackageReference Include="OrchardCore.Application.Cms.Core.Targets" Version="1.0.0-rc2-13799" />

any idea? thanks.

hishamco:
I think I saw similar issue before, let me try to find it ...

staymo:
@hishamco Thanks.
Just noticed that I failed to put the version. my version is:
 "OrchardCore.Application.Cms.Core.Targets" Version="1.0.0-rc2-13799"

hishamco:
Thanks @staymo I will try to reproduce that too, @deanmarcussen is `Manage Own Media` works as expected

sebastienros:
I don't remember if we implemented the "username" folder /cc @matiasmolleja

matiasmolleja:
Hello,
Sorry for being so late!
I never did. 
I found the permission created and never removed it. 
Maybe it could be better if we simply remove it until we add the concept of ownership to media items?

hishamco:
Ok @matiasmolleja let me have a look if I can implement it or remove until we implement it

sebastienros:
Yes we can remove it, and keep an issue open to track the feature.

hishamco:
Closed with the respect of #9369

PR #6980, opened 2020-08-28
Remove 'ManageOwnMedia' permission

hishamco:
Address #6645

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Media/Permissions.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Permissions.cs
@@ -36,10 +36,12 @@ namespace OrchardCore.Media
                 new PermissionStereotype
                 {
                     Name = "Author",
+                    Permissions = new[] { ManageMedia } // Replace this by ManageOwnMedia when it's implemented
                 },
                 new PermissionStereotype
                 {
                     Name = "Contributor",
+                    Permissions = new[] { ManageMedia } // Replace this by ManageOwnMedia when it's implemented
                 },
             };
         }
--- a/src/OrchardCore.Modules/OrchardCore.Media/Permissions.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Permissions.cs
@@ -8,12 +8,11 @@ namespace OrchardCore.Media
     public class Permissions : IPermissionProvider
     {
         public static readonly Permission ManageMedia = new Permission("ManageMediaContent", "Manage Media");
-        public static readonly Permission ManageOwnMedia = new Permission("ManageOwnMedia", "Manage Own Media", new[] { ManageMedia });
         public static readonly Permission ManageAttachedMediaFieldsFolder = new Permission("ManageAttachedMediaFieldsFolder", "Manage Attached Media Fields Folder");
 
         public Task<IEnumerable<Permission>> GetPermissionsAsync()
         {
-            return Task.FromResult(new[] { ManageMedia, ManageOwnMedia, ManageAttachedMediaFieldsFolder }.AsEnumerable());
+            return Task.FromResult(new[] { ManageMedia, ManageAttachedMediaFieldsFolder }.AsEnumerable());
         }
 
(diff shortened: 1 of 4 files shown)


deanmarcussen (changes requested):
I have to block this. 
Here you have found where the `ManageOwnMedia` permission is being used - in the `AttachedMediaFieldsAuthhandler`.

But you have then removed it, making the permission handler useless as it no longer requires any permissions granted to the user to pass.

hishamco:
I was thinking to replace it by `ManageMedia` permission, but I forgot to cc you

sebastienros:
The issue is to remove the permission.  You should probably also do that in this PR, or comment it. It will be reintroduced when the specific handler to check ownership is implemented.

hishamco:
>  You should probably also do that in this PR

If it's removed from all the places in the controller some places will be unsecured

@deanmarcussen is the last change in `AttachedMediaFieldsAuthhandler` fine for you or you suggesting something else?

deanmarcussen:
Yes have reviewed last change in `AttachedMediaFieldsAuthHandler` and the change you have made keeps it secured.

As has already been said the issue is to remove the permission, so that needs to be done (or commented preferably), otherwise this pr does not fix the issue.

It would be replaced by the ManageMedia permission in any other controller actions.

hishamco:
> It would be replaced by the ManageMedia permission in any other controller actions.

That means all the actions use it will be unsecured

deanmarcussen:
> > It would be replaced by the ManageMedia permission in any other controller actions.
> 
> That means all the actions use it will be unsecured

I don;'t think I understand this comment.

Can you just remove the `ManageOwnMedia` permission and update the pr, then comment if there is somewhere you think there is an issue please

src/OrchardCore.Modules/OrchardCore.Media/Permissions.cs
```diff
@@ -8,12 +8,11 @@ namespace OrchardCore.Media
     public class Permissions : IPermissionProvider
     {
         public static readonly Permission ManageMedia = new Permission("ManageMediaContent", "Manage Media");
-        public static readonly Permission ManageOwnMedia = new Permission("ManageOwnMedia", "Manage Own Media", new[] { ManageMedia });
```
hishamco, line 11: Oops, seems I forgot to remove it ;)

src/OrchardCore.Modules/OrchardCore.Media/Permissions.cs
```diff
@@ -37,12 +36,10 @@ public IEnumerable<PermissionStereotype> GetDefaultStereotypes()
                 new PermissionStereotype
                 {
                     Name = "Author",
-                    Permissions = new[] { ManageOwnMedia }
```
deanmarcussen, line 40: I think this will prevent these two roles from being able to select media from the media library when editing content.

Can you test and verify?

If that is the case, then they'll need the `ManageMedia` permission by default

hishamco: > If that is the case, then they'll need the ManageMedia permission by default

I think that too, but in this case all non Administrator will be equivalent, let me test and verify ...

hishamco: @deanmarcussen this will not work either because of this check

https://github.com/OrchardCMS/OrchardCore/blob/e6aa8e4f3a42d79c7bd2627b622cd792163cb37b/src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs#L77-L80

I'm wondering how it works before for non Administrator permission?!!

hishamco:
I need to do some tests to ensure that everything as expected

hishamco:
Seems the role permissions was not set in the Contributor permission in my laptop, now everything works as expected, @deanmarcussen what's the difference between `ManageMedia` & `ManageAttachedMediaFieldsFolder`? coz the contributor is able to create media folder(s) and upload media file(s)

deanmarcussen (approved):

Issue #6913, opened 2020-08-15
Attached media files are not included in the deployment plan


lukaskabrt:
I have a ContentItemType with a Media field with Attached editor type. 

If I try to execute a deployment plan definition with the following steps:
* Media
* ContentItem

I got a ZIP file, that contains files from media library, but none of the attached images. I would expect attached images to be also included. Example: [Import.zip](https://github.com/OrchardCMS/OrchardCore/files/5079550/Import.4.zip)

Version: 1.0.0-rc2-14115

deanmarcussen:
Are you using blob storage, or local storage?

Seems to work ok for me with local storage, but blob storage might be listing the folders differently.

lukaskabrt:
I am using blob storage.

deanmarcussen:
> I am using blob storage.

Ok, I thought it might be blob storage related.

The problem is that blobstorage does not correctly implement the `includeSubDirectories = true` option, it never has.

This is because it's not like a normal file system where you can make one request and list everything, so it would have to make recursive requests for every directory it finds.

The newer api we use may have changed that, but I suspect it hasn't. Will have a look.

Ironically the media step driver does this recursion already itself, but the  recipe step does not, when using `Include All Media`.

So the short term workaround should be to specify the attached fields folder `mediafields`, rather than `Include All Media` when creating the step.

This issue with the `includeSubDirectories = true` affects this step, and the GraphQL media query - it's the only place we use it.

Will have a bit of a think and decide whether recursively fetching directory contents inside the media service itself with repeated blob storage requests is really a good idea or if we can handle it some other way.

Issue #6840, opened 2020-08-06
Remote Deploy throws exception when hosted in Azure App Service


devlife:
I have an instance of OC RC2 running in an Azure App Service (Windows .NET 4.7). I enabled Remote Deploy, added a remote instance and tried to push a deployment plan to the remote instance. However, I'm getting an exception 

```
System.UnauthorizedAccessException:
   at System.IO.FileSystem.CreateDirectory (System.IO.FileSystem, Version=4.1.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a)
   at System.IO.DirectoryInfo.Create (System.IO.FileSystem, Version=4.1.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a)
   at OrchardCore.Deployment.Core.Services.TemporaryFileBuilder+<SetFileAsync>d__6.MoveNext (OrchardCore.Deployment.Core, Version=1.0.0.0, Culture=neutral, PublicKeyToken=nullOrchardCore.Deployment.Core, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null: C:\projects\orchardcore\src\OrchardCore\OrchardCore.Deployment.Core\Services\TemporaryFileBuilder.csOrchardCore.Deployment.Core, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null: 38)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw (System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e)
   at OrchardCore.Media.Deployment.MediaDeploymentSource+<ProcessDeploymentStepAsync>d__2.MoveNext (OrchardCore.Media, Version=1.0.0.0, Culture=neutral, PublicKeyToken=nullOrchardCore.Media, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null: C:\projects\orchardcore\src\OrchardCore.Modules\OrchardCore.Media\Deployment\MediaDeploymentSource.csOrchardCore.Media, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null: 60)
```

[...]

PR #7004, opened 2020-09-03
Update blob storage to search folder hierarchies

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6913

Probably fixes https://github.com/OrchardCMS/OrchardCore/issues/6840

Blob storage never handled seaching subfolders correctly, so this should now work for the media step, and better in graphql queries. (behaves identical to normal file system now)


deanmarcussen:
@agriffard was this agreed to merge at triage?

agriffard:
Doesn't ring a bell.

PR #7008, opened 2020-09-03
Toggles password visibility on login form

Skrypt:
orchard


src/OrchardCore.Modules/OrchardCore.Users/Views/Account/Login.cshtml
```diff
@@ -37,8 +37,12 @@
                             <a asp-controller="ResetPassword" asp-action="ForgotPassword" class="label-link">@T["Forgot your password?"]</a>
                         }
                     </label>
-
-                    <input asp-for="Password" class="form-control" tabindex="2" />
+                    <div class="input-group">
+                        <input asp-for="Password" class="form-control" tabindex="2" />
+                        <div class="input-group-append">
+                            <button class="btn btn-outline-secondary" type="button" id="togglePassword"><i class="icon far fa-eye"></i></button>
```
agriffard, line 43: Requires font-awesome.

Skrypt: Yes

Skrypt:
Don't merge yet I found an issue. For some reason it doesn't switch the icon all the time.

src/OrchardCore.Modules/OrchardCore.Users/Views/Account/Login.cshtml
agriffard: Then require the Resource or else if the theme doesn't have it, it won't work.

Skrypt: done

Skrypt:
Ok so I wanted to use the less dependencies possible on this by using javascript directly.
Did not want either to mess around with adding a function to wait to document to be ready in plain javascript as it can vary between browser versions. I'm handling both cases simply.

src/OrchardCore.Modules/OrchardCore.Users/Views/Account/Login.cshtml
```diff
@@ -81,3 +86,23 @@
         </div>
     }
 </div>
+<script type="text/javascript" at="Foot">
```
hishamco, line 89: While we have a dependency on jQuery, it will be much simpler if we use it

src/OrchardCore.Modules/OrchardCore.Users/Views/Account/Login.cshtml
```diff
@@ -81,3 +86,23 @@
...
+        type = password.getAttribute('type') === 'password' ? 'text' : 'password';
+        password.setAttribute('type', type);
+
+        // toggle the eye slash icon
+        icon = this.getElementsByClassName('icon')[0];
+        if(icon.getAttribute('data-icon')){ // if the icon is rendered as a svg
+            type === 'password' ? icon.setAttribute('data-icon', 'eye') : icon.setAttribute('data-icon', 'eye-slash');
+        }
+        else{ // if the icon is still a <i> element
+            type === 'password' ? icon.classList.remove('fa-eye-slash') : icon.classList.remove('fa-eye');
```
hishamco, line 104: It could be done with single condition instead of two

Skrypt: I could have chained the conditional operator but I don't do that normally for the sake of simpler readability mainly... not an issue.

src/OrchardCore.Modules/OrchardCore.Users/Views/Account/Login.cshtml
Skrypt: Well I explained why I did not use it. Now this is just a question of taste at this point. I generally prefer not having to rely on JQuery for everything. If we do this then we should also move this javascript code in the OrchardCore.Users module as a script resource for future extensibility purposes. Though, I'm not sure this is needed unless someone convince me of the opposite. It's just a visual component ...

hishamco: No problem, but while there are some UI components are using jQuery heavily seems it needs difficult to rid off the dependency or we may need a time

src/OrchardCore.Modules/OrchardCore.Users/Views/Account/Login.cshtml
```diff
@@ -81,3 +86,23 @@
         </div>
     }
 </div>
+<script type="text/javascript" at="Foot">
+    togglePassword = document.querySelector('#togglePassword');
+    password = document.querySelector('#Password');
+
+    togglePassword.addEventListener('click', function (e) {
+        // toggle the type attribute
```
deanmarcussen, line 94: cool, lgtm to me.

Only comment is when inlining javascript comments like this they render in the page.

I changed views recently to be razor comments instead, because I felt like we shouldn't be showing our comments to the world, i.e. in the output html.

You might agree / disagree?


Makes sense to me to do this with javascript rather than jquery as this is a view that is often themed, and shouldn't require the theme to use jquery.

hishamco (approved):
LGTM

src/OrchardCore.Modules/OrchardCore.Users/Views/Account/Login.cshtml
Skrypt: Makes sense I changed it to Razor comments

deanmarcussen:
LGTM if everyone is happy with the UI change

ns8482e:
Nice! Can we apply same in Tenant Setup screen?

hishamco:
@ns8482e the icon is fine, the eye should show up with type="password" and eye slash with type="text"

ns8482e:
> @ns8482e the icon is fine, the eye should show up with type="password" and eye slash with type="text"

OK

hishamco:
Perhaps the GIF is confusing little bit but it's right ;)

Skrypt:
> Nice! Can we apply same in Tenant Setup screen?

Done

PR #7009, opened 2020-09-04
ShellScope usages

jtkech:
In some rare places `UsingAsync()` was not used, but it is important that each scope executes in an isolated async flow.


netwavebe:
Is it possible that because of this change all IndexProviders are called twice? First during CreatedAsync(), then again during PublishedAsync().

The same happens when cloning an item, but in the first call (CreatedAsync) the contentItem is empty. My code crashed because I'm trying to get field values to use in the index.

It's easy to fix my code, but just to be sure: is this intended behaviour?

Skrypt:
Would need to be tested with a clean install and a breakpoint in the ContentItemIndexProvider.

Skrypt:
Tested it and I can confirm that it is the intended behaviour. I believe that there is a reason we are saving a draft first and then publishing the draft. That's how the DefaultContentManager works to call the proper chain of events on a content item.

Open an issue to explain what you are trying to achieve with more details so we can help.

netwavebe:
@Skrypt : I'm getting different results when cloning. During the first call it's almost empty (no title, no fields). As a result, I could not get the data I need to create the index. Second pass contains everything so problem fixes itself.

No need to open an issue, I fixed my code. I just wanted to make sure this was indended behaviour.

Skrypt:
You might want to check your Controller to see what it does or maybe a custom Handler that would trigger the ContentManager on each modification made on a content item. I just tried and in the source code when we clone it passes only once in the ContentItemIndexProvider.Describe method so when it clones it only needs to pass once for creating a draft of the copied content item.

netwavebe:
Strange, I tried this with the most basic IndexProvider I have (no custom handler). Fetched dev this morning.

netwavebe:
Callstack of the first call (with empty content item):
image

Callstack of the second call:
image

Skrypt:
Ok so this is coming from a Workflow Activity call.

netwavebe:
Not mine. I only have one unrelated workflow in this solution.

Skrypt:
Could you add a breakpoint on the GetByStartActivityAsync method so that we know which activity got triggered there?

netwavebe:
"ContentCreatedEvent"

PR #7010, opened 2020-09-04
Allows GetSettings() to be safer

jtkech:
Prevents from failing if it happens when the related shell, being released, is removed temporarily from the list

### DIFF
--- a/src/OrchardCore/OrchardCore/Shell/ShellHost.cs
+++ b/src/OrchardCore/OrchardCore/Shell/ShellHost.cs
@@ -29,8 +29,8 @@ namespace OrchardCore.Environment.Shell
         private readonly ILogger _logger;
 
         private bool _initialized;
-        private ConcurrentDictionary<string, ShellContext> _shellContexts = new ConcurrentDictionary<string, ShellContext>();
-        private ConcurrentDictionary<string, ShellSettings> _shellSettings = new ConcurrentDictionary<string, ShellSettings>();
+        private readonly ConcurrentDictionary<string, ShellContext> _shellContexts = new ConcurrentDictionary<string, ShellContext>();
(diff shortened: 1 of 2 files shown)


src/OrchardCore/OrchardCore/Shell/ShellHost.cs
```diff
@@ -29,7 +29,8 @@ public class ShellHost : IShellHost, IDisposable
         private readonly ILogger _logger;
 
         private bool _initialized;
-        private ConcurrentDictionary<string, ShellContext> _shellContexts = new ConcurrentDictionary<string, ShellContext>();
+        private readonly ConcurrentDictionary<string, ShellContext> _shellContexts = new ConcurrentDictionary<string, ShellContext>();
+        private readonly ConcurrentDictionary<string, ShellSettings> _shellSettings = new ConcurrentDictionary<string, ShellSettings>();
```
deanmarcussen, line 33: I don't see anything removing shellsettings from this dictionary?

Did I miss something or are they accumulating / being overwritten on line 187

jtkech: @deanmarcussen 

When a given shell is created, it is added to the list and **never removed permanently** as we can't remove a shell at runtime. **It can be removed temporarily** when released / reloaded, but just before a new shell is re-added.

The new shell settings collection is updated at the 3 places where we call `_shellContexts.TryAdd()` and if it returns true, **but a given shell settings is never removed, only updated**, so that we can safely use `GetSettings()`.

deanmarcussen: Cool, thanks for clarifiying @jtkech

deanmarcussen (approved):

Issue #6996, opened 2020-09-02
Admin menus Deployment plan should update not recreate


deanmarcussen:
Related to similar issue https://github.com/OrchardCMS/OrchardCore/issues/6933

The admin menus deployment plan / recipe creates new admin menus every time the deployment is run.

It should (probably) check for an existing menu of the same name and either update or replace it.

We may need options for both update (ie merge) and replace.

PR #7014, opened 2020-09-04
Replace or create an existing admin menu tree when deploying

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6996

Changes the behaviour, and that behaviour had a comment on it (so @Skrypt if you happen to remember why it was done that way)

Have made it replace an existing tree, or create if it doesn't exist. (service already handles that)

Merging an existing tree with a new one would make no sense as there is not enough unique ids to merge the arrays

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.AdminMenu/Recipes/AdminMenuStep.cs
+++ b/src/OrchardCore.Modules/OrchardCore.AdminMenu/Recipes/AdminMenuStep.cs
@@ -33,8 +33,8 @@ namespace OrchardCore.AdminMenu.Recipes
             foreach (JObject token in model.Data)
             {
                 var adminMenu = token.ToObject<Models.AdminMenu>(serializer);
-                
-                // When the id is not supplied create a new tree, otherwise replace or create the existing tree.
+
+                // When the id is not supplied create a new menu, otherwise replace the existing menu, or create a new menu.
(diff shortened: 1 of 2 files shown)


src/OrchardCore.Modules/OrchardCore.AdminMenu/Recipes/AdminMenuStep.cs
```diff
@@ -33,8 +33,14 @@ public async Task ExecuteAsync(RecipeExecutionContext context)
             foreach (JObject token in model.Data)
             {
                 var adminMenu = token.ToObject<Models.AdminMenu>(serializer);
-                adminMenu.Id = Guid.NewGuid().ToString("n");// we always add it as a new tree.
-                await _AdminMenuService.SaveAsync(adminMenu);
+
+                // When the id is not supplied create a new menu, otherwise replace the existing menu, or create a new menu.
```
agriffard, line 37: Rephrase comment/add comma: 2 times 'create a new menu' at the beginning and at the end.

deanmarcussen: Better?

Skrypt:
Actually this is @matiasmolleja who did this PR and we lost history because I renamed the module entirely.

Issue #7013, opened 2020-09-04
Regression - Admin Site - Specific content type list, allow only new content item of the selected type


PiemP:
Hi all, 

I have see that the issue #4985 is already here. I have looked at the code and I have see that the code to render the button is right but the code in the controller doesn't filter the contentitem type when a contentitem type is "selected". 

Would you like that i  create a PR about it?

Thank you. 

PS. when you have filtered the contentitem list with the "Filters" text field, could be useful to have a reset button to clear the filter text field and submit the form to have the list unfiltered?

PiemP:
Sorry, 

the code in the controller to manage the issue exists but this variable is false and the user that I use can create this kind of content item
![backend_contentitem_list](https://user-images.githubusercontent.com/2736009/92223713-253b2580-eea1-11ea-8654-7274447e4cd6.png)

deanmarcussen:
`CanCreateSelectedContentType` is a special property that a menu item can provide to allow non createable content to become creatable.

It's used for Menus. _Content -> Menus_

So I don't think it's related to your issue, i.e. it's not intended for that use.

If I understand the issue, you want the new button to be restricted to just the content type provided by an admin menu content type link?

You could probably do that in the same area of code, just not checking against `CanCreateSelectedContentType`

PiemP:
> So I don't think it's related to your issue, i.e. it's not intended for that use.

Thank you for your explaination. I believe to have understand the meaning of `CanCreateSelectedContentType`.

There something wrong in this code because if the option `CanCreateSelectedContentType` is set to **false**, `model.Options.CreatableTypes` is null and the action fill `CreatableTypes` with all the types that are creatable.  

I can do a PR about it? I can add a button to clear the text field used to search content items in the admin interface where content items are listed or is it possible to overwrite the view `ContentsAdminList-Search.cshtml` of the module `OrchardCore.Contents`? 

Thank you

deanmarcussen:
> There something wrong in this code because if the option `CanCreateSelectedContentType` is set to **false**, `model.Options.CreatableTypes` is null and the action fill `CreatableTypes` with all the types that are creatable.

There's nothing wrong in this code (for it's purpose, which is related to menus) but for your purpose it probably needs to do something else as well. 

So yes, please feel free to make a pr.

If you want to do a pr for a filter clearing button please do that as a different pr (seperate issue, easier to review if it's one change at a time)

PiemP:
[...]

PR #7018, opened 2020-09-04
Content items: Reset button in search

PiemP:
Related #7013 

clearfilter_button


agriffard:
Which browser are you using?

On a Chrome (or Microsoft Edge), an `input type="search"` has an X button to clear the content when you start typing:

image

PiemP:
> Which browser are you using?
Firefox

hishamco (changes requested):
I'm stopping this because we already have similar issue, as @agriffard said the clear button is show up in the modern browsers

hishamco:
It's addressed while I made this PR #6846

PiemP:
I respect your decision, indeed before do this PR I have asked about the possibility to rewrite `ContentsAdminList-Search.cshtml` in my custom admin theme. If you believe is not necessary is good for me. 

For my opinion the "X" approach is good for a search engine where a user do a search and then want to reset the first search and do a new search, like google, but is not same good for a search field like this one where a user could do a search and then  want to see this field empty to list all items. In google you can't do an empty search, here could be have sense to allow user to have a button that clear the filter and submit it in one action.

Thank you and sorry, I have lost to do a search over closed issue before do this PR.

Skrypt:
I can definitely see an issue that could be fixed. Here, there is no issue with the X button. It should only clear the text in the textfield. But having a button to reset the search would be more intuitive.

agriffard:
Please merge dev.

Maybe with an icon and a toggle tooltip?
Here are some suggestions (Guess what is my preference):

image
image
image

PiemP:
I have opted for the **times** icon and I have changed the title of the button from "Clear" to "Reset". 

For my opinion the **trash** icon and the **paper** icon doesn't fit the element and could give some misunderstanding. If you don't like the **times** icon (because too similar to the browser one) I have another option: a **broom** icon.

PiemP:
@Skrypt In dark mode the search form buttons border "shine". Only to be sure, is it a desired behavior?

SharedScreenshot

agriffard:
Please merge dev and fix the conflict.

PiemP:
> Please merge dev and fix the conflict.

done.

Piedone:

PiemP:
@Piedone probably is better to close it... when I have time I try to check this thing and if is necessary I open a new issue and a new PR. 

Thank you.

Piedone:
I see, thank you for your quick reply.

PR #7022, opened 2020-09-05
Unused shell services

jtkech:


jtkech:
I tried to remove these services to make it simpler to keep in sync tenants in a distributed environment. But it breaks `IFeatureEventHandler` whose the only one implementation is `RoleUpdater`, i didn't see it when using the blog recipe but here a functional test fails for the agency setup that renders the login page in place of the landing page.

Will see tomorrow if there is a simpler way and / or how to keep in sync tenants when enabling a feature that registers an `IFeatureEventHandler` as the roles module does.

So, right now i close this PR

PR #7033, opened 2020-09-07
Unused ShellSettings direct property

jtkech:
We use shell settings direct properties for the properties coming from the settings sources as `tenants.json`, and that we need to start a tenant. For the other properties coming from the regular config sources, we use the shell settings indexer.

Unless for the `Description` that is a regular config value but for which we created a direct property

So, because we never use this `Description` direct property in the code, only the indexer, i removed it so that it is clearer that we only use shell settings direct properties for values coming from settings sources as `tenants.json`.

### DIFF
--- a/src/OrchardCore/OrchardCore.Abstractions/Shell/ShellSettings.cs
+++ b/src/OrchardCore/OrchardCore.Abstractions/Shell/ShellSettings.cs
@@ -36,12 +36,6 @@ namespace OrchardCore.Environment.Shell
             Name = settings.Name;
         }
 
-        public string Description
-        {
-            get => _configuration["Description"];
-            set => _configuration["Description"] = value;
-        }
-
         public string Name { get; set; }
 
         public string Identifier


hishamco:
I remembered that I used here https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Index.cshtml#L164 but let me check if I used the indexer or not

hishamco:
I used the indexer https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs#L103 for a reason after you convinced me if I'm not wrong

But I'm still I see it's useful to keep the `Description` and use it directly here, unless you have another though

jtkech:
You use `entry.Description` where entry is not a `ShellSettings` but a `ShellSettingsEntry` initialized in the controller

hishamco:
Ya, which is using an indexer behind the scene, ok it's fine to remove it because `ShellSettings` is using the indexer too

hishamco (approved):
LGTM

hishamco:
One more thing can we create a new class to have a default properties instead of using message strings

jtkech:
`ShellSettings` uses 2 inner `ShellConfiguration`

One for **settings values** coming from `tenants.json` by default, they are well known values that we nedd to pre-create a tenant, that' why we created direct properties

One for **config values** coming from any regular config sources, and we don't know them, that's why we access them by the `ShellSettings` indexer, or directly through this inner `ShellConfiguration` that we retrieve when injecting `IShellConfiguration`.

But yes we could also have well known **config values** direct properties, e.g. `ConfigurationString`, `DatabaseProvider` and so on, but maybe it would be confusing.

So because right now the only one direct property for a **config value** was `Description`, and because it is never used in the current code, i think it is better to just remove it.

hishamco:
> But yes we could also have well known config values direct properties, e.g. ConfigurationString, DatabaseProvider and so on, but maybe it would be confusing.

I didn't mean direct properties, at least a class that hold the default properties names to avoid magic string

hishamco:
Please :shipit:

jtkech:
Okay but this kind of other class can't be defined in `OC.Abstractions` which is not aware of these specific properties, so for a specific property in another specific core library or module, and this is what the `OC.Tenants` module does with its view models initialized through its controller, so that you can use e.g. `entry.Description` in `Index.cshtml`.

deanmarcussen (approved):

PR #7046, opened 2020-09-09
add unique filename and safe delete function

### LINKED ISSUE #7025: Export on published Windows Container failing
I've been using the windows container to get a poc up quickly to prove out some concepts. I built out a few content definitions and everything has been working fine for a couple week. I finally created a deployment plan so I could move it to another server. When I attempted to execute the plan, the site threw an exception and didn't seem to write anything to log. I was able grab the source and debug it in a local docker instance to repro. The site is failing with:

**_UnauthorizedAccessException: Access to the path 'C:\Windows\TEMP\rraza435.4c2' is denied._**
System.IO.FileSystem.GetFindData(string fullPath, ref WIN32_FIND_DATA findData)
System.IO.FileSystem.RemoveDirectory(string fullPath, bool recursive)
System.IO.Directory.Delete(string path, bool recursive) 
OrchardCore.Deployment.Core.Services.TemporaryFileBuilder.Dispose() in TemporaryFileBuilder.cs **_line 25_**
OrchardCore.Deployment.Controllers.ExportFileController.Execute(int id) in ExportFileController.cs

Apparently the process has perms to generate the temp files,  but can't delete them. 
I am able to repro this by:
- spinning up a new windows container instance
- run headless recipe
- create a single content definition
- create a deployment plan
- add a single step to export content definitions
- Execute the deployment plan
- Click Select on the File Download screen.

VintageMo:
Add a unique filename for the physical export file and add a safe delete function to prevent UnauthorizedAccessException in a windows container. 
Fixes #7025
The windows container was failing with an UnauthorizedAccessException when the TemporaryFileBuilder is disposed. The Directory.Delete(folder, recursive) call was throwing the exception. To work around, I walk the directory, set the file attributes to Normal, delete the file, then delete the empty folder. This safe delete no longer throws the exception. One side effect of this exception was the code was orphaning a zip file in the Windows/Temp folder. This filename matched the deployment plan, so once it was orphaned, it stayed broken. This also would blow up if two admins export a development plan at the same time. To fix, I save the archive file with a random filename and then export result with the correct deployment plan name for the zip. 

Aside... The export file clean up was a bit confusing at first. I finally found it happening in a ResultFilterAttribute. It would be a bit cleaner if a new TemporaryPhysicalFileResult class is created that inherits from PhysicalFileResult. Then the File.Delete() could occur in the overridden ExecuteResultAsync() method. It'd keep all of the logic for the FileResult in one place.

### DIFF
--- a/src/OrchardCore/OrchardCore.Deployment.Core/Services/TemporaryFileBuilder.cs
+++ b/src/OrchardCore/OrchardCore.Deployment.Core/Services/TemporaryFileBuilder.cs
@@ -31,7 +31,6 @@ namespace OrchardCore.Deployment.Core.Services
         {
             foreach (string file in Directory.GetFiles(targetDirectory))
             {
-                File.SetAttributes(file, FileAttributes.Normal);
                 File.Delete(file);
             }
 
--- a/src/OrchardCore/OrchardCore.Deployment.Core/Services/TemporaryFileBuilder.cs
+++ b/src/OrchardCore/OrchardCore.Deployment.Core/Services/TemporaryFileBuilder.cs
@@ -22,9 +22,25 @@ namespace OrchardCore.Deployment.Core.Services
             {
                 if (Directory.Exists(Folder))
                 {
-                    Directory.Delete(Folder, true);
+                     SafeDeleteDirectory(Folder);
                 }
             }
+        }
+
+        private static void SafeDeleteDirectory(string targetDirectory)
+        {
+            foreach (string file in Directory.GetFiles(targetDirectory))
+            {
+                File.SetAttributes(file, FileAttributes.Normal);
+                File.Delete(file);
+            }
+
+            foreach (string dir in Directory.GetDirectories(targetDirectory))
+            {
+                SafeDeleteDirectory(dir);
+            }
+
+            Directory.Delete(targetDirectory, false);
         }
 
         public async Task SetFileAsync(string subpath, Stream stream)
--- a/src/OrchardCore.Modules/OrchardCore.Deployment.Remote/Controllers/ExportRemoteInstanceController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Deployment.Remote/Controllers/ExportRemoteInstanceController.cs
@@ -72,30 +72,25 @@ namespace OrchardCore.Deployment.Remote.Controllers
 
             using (var fileBuilder = new TemporaryFileBuilder())
             {
-                archiveFileName = PathExtensions.Combine(Path.GetTempPath(), filename);
+                archiveFileName = PathExtensions.Combine(Path.GetTempPath(), Path.GetRandomFileName());
 
                 var deploymentPlanResult = new DeploymentPlanResult(fileBuilder, new RecipeDescriptor());
                 await _deploymentManager.ExecuteDeploymentPlanAsync(deploymentPlan, deploymentPlanResult);
 
-                if (System.IO.File.Exists(archiveFileName))
-                {
-                    System.IO.File.Delete(archiveFileName);
-                }
-
                 ZipFile.CreateFromDirectory(fileBuilder.Folder, archiveFileName);
             }
 
-            HttpResponseMessage response;
-
             try
             {
+                HttpResponseMessage response;
+
                 using (var requestContent = new MultipartFormDataContent())
                 {
                     requestContent.Add(new StreamContent(
                         new FileStream(archiveFileName,
                         FileMode.Open, FileAccess.Read, FileShare.ReadWrite, 1, FileOptions.Asynchronous | FileOptions.SequentialScan)
                     ),
-                        nameof(ImportViewModel.Content), Path.GetFileName(archiveFileName));
+                        nameof(ImportViewModel.Content), filename);
                     requestContent.Add(new StringContent(remoteInstance.ClientName), nameof(ImportViewModel.ClientName));
                     requestContent.Add(new StringContent(remoteInstance.ApiKey), nameof(ImportViewModel.ApiKey));
 
--- a/src/OrchardCore.Modules/OrchardCore.Deployment/Controllers/ExportFileController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Deployment/Controllers/ExportFileController.cs
@@ -54,8 +54,9 @@ namespace OrchardCore.Deployment.Controllers
 
             using (var fileBuilder = new TemporaryFileBuilder())
             {
-                archiveFileName = PathExtensions.Combine(Path.GetTempPath(), filename);
-
+                // use random name for archive to prevent collision
+                archiveFileName = PathExtensions.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+                
                 var recipeDescriptor = new RecipeDescriptor();
                 var recipeFileDeploymentStep = deploymentPlan.DeploymentSteps.FirstOrDefault(ds => ds.Name == nameof(RecipeFileDeploymentStep)) as RecipeFileDeploymentStep;
 
@@ -72,9 +73,23 @@ namespace OrchardCore.Deployment.Controllers
                     recipeDescriptor.Tags = (recipeFileDeploymentStep.Tags ?? "").Split(',', StringSplitOptions.RemoveEmptyEntries);
                 }
 
-                var deploymentPlanResult = new DeploymentPlanResult(fileBuilder, recipeDescriptor);
-                await _deploymentManager.ExecuteDeploymentPlanAsync(deploymentPlan, deploymentPlanResult);
-                ZipFile.CreateFromDirectory(fileBuilder.Folder, archiveFileName);
+                try
+                {
+                    var deploymentPlanResult = new DeploymentPlanResult(fileBuilder, recipeDescriptor);
+                    await _deploymentManager.ExecuteDeploymentPlanAsync(deploymentPlan, deploymentPlanResult);
+                    ZipFile.CreateFromDirectory(fileBuilder.Folder, archiveFileName);
+                }
+                catch
+                {
+                    var file = new FileInfo(archiveFileName);
+
+                    if (file.Exists)
+                    {
+                        file.Delete();
+                    }
+
+                    throw;
+                }
             }
 
             return new PhysicalFileResult(archiveFileName, "application/zip") { FileDownloadName = filename };


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=7046) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=7046) before we can accept your contribution.<br/><br/>:x: VintageMo [sign now](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=7046)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OrchardCMS/OrchardCore?pullRequest=7046) it.</sub>

deanmarcussen:
The change to the unique filename looks good.

I'm not sure about the safe delete method.
It feels like there is something specific about the windows container / permissions that are causing the issue.

I can't repro the issue on a docker for linux container.

Could you provide some more details as to what permissions these temp files are being created with on windows, as it seems strange that they are being created with the incorrect permission, presumably readonly.

If they are being created readonly, I'd like to understand why that's happening

VintageMo:
It's definitely specific to windows containers. I tried to make that clear in my issue and PR. If you just replace the line to call to my method with Directory.Delete(), it will blow up 100% of the time in windows containers. That's why I'm shocked no one else reported it. I couldn't get a comment on my problem in the last week.  
I just ran it again and it looks like the inner recipe file has the attribute, **_Archive_**. I can only assume that this was because they were processed and compressed by the ZipFile class. The Directory.Delete(folder, recursive) fails when the inner file has the attribute **_Archive_**. When set to Normal, delete proceeds as expected.

sebastienros:
> That's why I'm shocked no one else reported it.

I don't know anyone (but you) that uses OC in Docker for windows. Most use Docker on linux.

Could you explain how you found this solution? Is the issue documented somewhere? Isn't there a way to grant better permission sets on the account for the docker container?

VintageMo:
I originally set up the linux container and built out some definitions, but I later had to switch to windows container to match the projects I was testing in docker desktop. 
I _found_ my solution by confirming that within the container, I could actually delete the single file recipe.json file with Normal attribute set. Next step, I verified that i could delete the folder. I was only dealing with a folder with a single file so it was pretty easy to narrow down. After creation, the only thing that touches it is the ZipFile call. This is not a new workaround. The recursive delete method has been suspect at other times in the last 20 years of c#. After a while, I got used to avoiding the recursive flag and being explicit with my deletes. I couldn't find a .net core issue related to this. I will do some digging. Does this really sound like something that should need extra permissions? I would think the vanilla container config should allow you to create and delete your files in the temp folder. None of the orchard code appears to be holding reference to the files after they are created. From what I've read, when no other locks are held on the files, a recursive delete should only throw UnauthorizedAccessException when a readonly file is found in the subtree. The Archive attribute shouldn't cause this exception. I will see if I can create a console app repro with a few lines.

Skrypt:
Generally when you can't delete a file that's because there is a lock process on it. Maybe, the issue comes from when we create the ZipFile itself. Maybe the file is still locked by that compression process which is not disposing the file from it's streamwriter or something like it. I would be surprised that the issue comes from the folder permissions.

VintageMo:
I stripped it down to the basics. Creating an empty folder in the temp path can repro the behavior. The call succeeds in IIS Express and local console app, but fails when a container is deployed to docker desktop. 
At its simplest, this **_passes_**:
```
            Directory.CreateDirectory("foldername");
            Directory.Delete(folder, false);

            Directory.CreateDirectory("foldername");
            Directory.Delete(folder, true);

            Directory.CreateDirectory("c:\windows\temp\foldername");
            Directory.Delete(folder, false);

```
and this **_fails_**:
```
            Directory.CreateDirectory("c:\windows\temp\foldername");
            Directory.Delete(folder, true);
```
In a windows container, the recursive directory delete on a newly created folder in **_Windows\Temp_** throws UnauthorizedAccessException. It will create/delete the directory in the /app folder with no exceptions thrown. I get what you are saying about something holding the lock, but both recursive and non-recursive delete should fail. The early framework had this same sort of bug in like 2002. I saw a tag on my related issue that this will be handled post 1.0. Unfortunately, I don't have my choice of containers. Due to time constraints, waiting for a post-1.0 fix will probably force us to go with a different cms :/ I'm rooting for OrchardCore, but not being able to export the content definitions and items is a deal breaker for mgmt and our cms option needs to be sealed by the end of this month.

Skrypt:
Which version of Windows you are using in your Docker image?

VintageMo:
I was trying to be brief. Next time I'll put it all in. 
I first used the dockerfile as created by VS2019:
FROM mcr.microsoft.com/dotnet/core/runtime:3.1-nanoserver-1809 AS base

Then I changed it to use the dockerfile in the root of the orchard solution:
FROM mcr.microsoft.com/dotnet/core/aspnet:3.1

Skrypt:
Here is the result that I get from my tests : 

``` CMD
2020-09-11 19:21:12.721 -04:00 [INF] 1 - Created folder : C:\app\foldername
2020-09-11 19:21:12.745 -04:00 [INF] 1 - Deleted folder recursive : C:\app\foldername
2020-09-11 19:21:12.746 -04:00 [INF] 2 - Created folder : C:\Windows\TEMP\foldername
2020-09-11 19:21:12.759 -04:00 [ERR] Access to the path 'C:\Windows\TEMP\foldername' is denied.
2020-09-11 19:21:12.721 -04:00 [INF] 1 - Created folder : C:\app\foldername
2020-09-11 19:21:12.745 -04:00 [INF] 1 - Deleted folder recursive : C:\app\foldername
2020-09-11 19:21:12.746 -04:00 [INF] 2 - Created folder : C:\Windows\TEMP\foldername
2020-09-11 19:21:12.759 -04:00 [ERR] Access to the path 'C:\Windows\TEMP\foldername' is denied.
[2020-09-11T23:21:12.000Z][LOGMONITOR] INFO: Entrypoint processs exit code: 0
[2020-09-11T23:21:12.000Z][LOGMONITOR] INFO: Entrypoint processs exit code: 0
```

So the issue is that when we are using `Path.GetTempPath()` on a Windows container in Docker. It will use the adminstrator account to run the process and by default this account has the C:\Windows\temp\ folder as it's default temporary folder instead of C:\app\foldername. So, I need to look where the recipe files gets created. If they use the same `Path.GetTempPath()` then we could simply fix it by changing these.

hishamco:
>  but add a condition to never use the C:\Windows\Temp folder

Can we create a `Temp` folder inside `OrchardCore.Cms.Web` that can be used as a temporary storage? This can be used for deployment, recipes, media, installation or anything we can think about

Skrypt:
Yeah I was thinking about /App_Data/Temp folder. Just wrote this : 

```
using Serilog;
using System;
using System.IO;

namespace RemoveTempFolder
{
    class Program
    {
        static void Main(string[] args)
        {
            Log.Logger = new LoggerConfiguration()
            .WriteTo.File("consoleapp.log")
            .CreateLogger();

            try
            {
                var folder = Directory.CreateDirectory("foldername");
                Log.Information("1 - Created folder : " + folder.FullName);
                Directory.Delete(folder.FullName, true);
                Log.Information("1 - Deleted folder recursive : " + folder.FullName);

                if (Path.GetTempPath().StartsWith("C:\\Windows\\TEMP"))
                {

                    folder = Directory.CreateDirectory(Environment.CurrentDirectory + "\\App_Data\\Temp\\foldername");
                }
                else
                {
                    folder = Directory.CreateDirectory(Path.GetTempPath() + "foldername");
                }

                Log.Information("2 - Created folder : " + folder.FullName);
                Directory.Delete(folder.FullName, true);
                Log.Information("2 - Deleted folder recursive : " + folder.FullName);

                folder = Directory.CreateDirectory("C:/windows/temp/foldername");
                Log.Information("3 - Created folder : " + folder.FullName);
                Directory.Delete(folder.FullName, true);
                Log.Information("3 - Deleted folder recursive : " + folder.FullName);
            }
            catch (Exception e)
            {
                Log.Error(e.Message);
            }
        }
    }
}

```

VintageMo:
App_Data is definitely better than the windows folder. I was a bit surprised it picked a temp folder under windows in the first place. Where are your non-recursive calls?
The main thing I was pointing out is that:
Directory.Delete(folder, **true**)
Directory.Delete(folder, **false**)
both calls should return the same result on an empty folder, whether it is successful delete or an exception. There's either a bug there... or they built in a safeguard to prevent you from doing recursive delete explicitly in the sensitive windows os folder. The fix is to delete the files and then the folder and not try a recursive delete. If you weren't allowed to delete it, Directory.Delete(folder, false) wouldn't work. If the TemporaryFileBuilder isn't building symbolic links, it's the safest, most transparent, way to do the delete.

Skrypt:
I removed the non-recursive calls after finding they we're working fine. I wanted to have tests that could potentially fail here so I took the worst case scenario path.

VintageMo:
Did you confirm that my PR fixes the problem and allows the export to work? In SafeDeleteDirectory() I can drop the SetAttribute call now that I've eliminated the Archive attribute as the culprit. It's the fix to the phantom exception. 
Now that I've narrowed it, I will submit the issue in corefx, so they can either fix the Directory.Delete() bug or document why we can't recursive delete in the windows\temp folder.

Skrypt:
Well I think that the proper fix would be to use the App_Data/Temp folder simply as we should never use this Windows/Temp folder with OC. The reason is that I feel concerned about other places where we could need to recursively delete files like this.

I did not test your PR yet but I'm pretty sure it fixes the issue differently.

Here 2 things to consider : 

Is deleting recursively in the Windows/Temp folder should be allowed? (I don't think so)
Should we be using the Windows/Temp folder? (I don't think so)
Should the administrator account be resolving `C:/Windows/Temp` as it's temp folder? 

https://github.com/dotnet/runtime/issues/19389

Based on the fact that if someone uses a Active Directory user account with his Docker container ; it should work fine by using simply `Path.GetTempPath()`. So, as @deanmarcussen said in the last meeting. It should be using a user folder for that matter because it requires to have the right permissions. Here, the only workaround that I can see is that the web application has r/w/d access to the App_Data folder so I think we should use it as the fallback folder instead of "C:/Windows/Temp".

Now if we think about Linux I don't think we should be concerned since it will always be set into the proper /usr/username/temp folder.

At this point, yes, I think that this is "maybe" an issue to report but at the same time, we should never use the administrator/root accounts to run a Docker container.

Skrypt:
Ok I think one workaround for this would be to set the TMP and TEMP environment variable on your Windows Docker containers simply.

Here is what I did with my test project : 

```dockerfile
FROM mcr.microsoft.com/dotnet/core/runtime:3.1

COPY LogMonitor.exe LogMonitorConfig.json C:/LogMonitor/
SHELL ["C:\\LogMonitor\\LogMonitor.exe", "cmd", "/S", "/C"]

WORKDIR /app
COPY build/release /app

ENV TMP C:/app/App_Data/temp/
ENV TEMP C:/app/App_Data/temp/

ENTRYPOINT C:\LogMonitor\LogMonitor.exe C:\app\RemoveTempFolder.exe
```

And here is the result : 

```C#
2020-09-11 21:30:28.494 -04:00 [INF] 1 - Created folder : C:\app\foldername
2020-09-11 21:30:28.516 -04:00 [INF] 1 - Deleted folder recursive : C:\app\foldername
2020-09-11 21:30:28.517 -04:00 [INF] 2 - Created folder : C:\app\App_Data\temp\foldername
2020-09-11 21:30:28.517 -04:00 [INF] 2 - Deleted folder recursive : C:\app\App_Data\temp\foldername
2020-09-11 21:30:28.518 -04:00 [INF] 3 - Created folder : C:\windows\temp\foldername
2020-09-11 21:30:28.530 -04:00 [ERR] Access to the path 'C:\windows\temp\foldername' is denied.
2020-09-11 21:30:28.494 -04:00 [INF] 1 - Created folder : C:\app\foldername
2020-09-11 21:30:28.516 -04:00 [INF] 1 - Deleted folder recursive : C:\app\foldername
2020-09-11 21:30:28.517 -04:00 [INF] 2 - Created folder : C:\app\App_Data\temp\foldername
2020-09-11 21:30:28.517 -04:00 [INF] 2 - Deleted folder recursive : C:\app\App_Data\temp\foldername
2020-09-11 21:30:28.518 -04:00 [INF] 3 - Created folder : C:\windows\temp\foldername
2020-09-11 21:30:28.530 -04:00 [ERR] Access to the path 'C:\windows\temp\foldername' is denied.
```

I think, in the light of this last test that we should simply add in the documentation that when running a Windows Docker container, the TMP and/or TEMP folder should be set to /App_Data/temp of your website folder.

Skrypt (approved):
Ok I think this PR is LGTM as it fixes the issues. We can add more later on.

src/OrchardCore/OrchardCore.Deployment.Core/Services/TemporaryFileBuilder.cs
```diff
@@ -22,11 +22,27 @@ public void Dispose()
...
+                     SafeDeleteDirectory(Folder);
                 }
             }
         }
 
+        private static void SafeDeleteDirectory(string targetDirectory)
+        {
+            foreach (string file in Directory.GetFiles(targetDirectory))
+            {
+                File.SetAttributes(file, FileAttributes.Normal);
```
Skrypt, line 34: Remove this as you said.

VintageMo:
I tried to sync and it said I was behind. Thanks for removing.

deanmarcussen (changes requested):
Thanks for doing further testing @Skrypt and @VintageMo 

From what I'm reading it still sounds like the issue is specific to windows docker and would be resolved by setting a different `TEMP` environment variable - to a location that was suitable.

Can someone confirm this is the case? Or am I wrong here

I remain reluctant to have the `SafeDelete` method in as a fix to what is still sounding like an environment specific issue.

We (as in our codebase) shouldn't be using an AppData/Temp folder directly. 

But if we said (docs) to specify it through an Environment variable that makes sense.

The `TEMP` environment variable changes location in Azure and other hosting platforms

Skrypt:
Yes, I've started making changes locally to make use of a ORCHARD_APP_DATA and ORCHARD_APP_DATA_TEMP env var in the TemporaryFileBuilder. A good workaround right now is to simply set the TEMP and/or TMP env var in the dockerfile so that Path.GetTempPath() resolves to something different than C:\Windows\Temp. So my suggestion is to make the App_Data/Temp folder location configurable. Of course, if you don't set these env var then the possibility of ending with C:\Windows\Temp will still be possible, so this is why I think this PR is a good addition for that matter. I don't want to force people into using a folder or an other. If Path.GetTempPath() can resolve to C:\Windows\Temp then this PR fixes that use case.

Skrypt:
I'm closing this PR since we found that we can resolve the issue by setting the TEMP env var of the Docker container.

PR #7049, opened 2020-09-09
Upgrade Test package dependencies

agriffard:
```
    "jest": "^26.4.2",
    "jest-cli": "^26.4.2",
    "puppeteer": "^5.2.1" 
```

puppeteer was 2.1.1 before this.


agriffard:
@jptissot Can you please have a look to determine why the test-run is not working anymore.

jptissot:
@agriffard The issue was an old node version. I also moved the deletion of App_Data before running  the test instead of after. Less risk of it not working.

jptissot (approved):

PR #7055, opened 2020-09-10
ChallengeOrForbid in Api controllers

### LINKED ISSUE #7039: Fix all the API controllers decorated with [Authorize(AuthenticationSchemes = "Api")] to return ChallengeOrForbid("Api")
Most of the OC API controllers are configured to use the "Api" scheme for authentication, but return `ChallengeOrForbid()`, that ultimately uses the default challenge scheme (which is typically the cookie handler when the Users module is enabled).

To fix that, the API controllers must be updated to do `ChallengeOrForbid("Api")` instead of `ChallengeOrForbid()`, so that challenges are properly handled by OpenIddict when the OpenID token validation feature is enabled.

/cc @deanmarcussen @OrchardSkills

agriffard:
Fixes #7039

Added an optional `authenticationScheme` to the `ChallengeOrForbid` method.

Replaced `this.ChallengeOrForbid()` by `this.ChallengeOrForbid("Api")` in Api controllers.

`api/queries` doesn't seem to have any ChallengeOrForbid.

### DIFF
--- a/src/OrchardCore/OrchardCore.Mvc.Core/Utilities/ControllerExtensions.cs
+++ b/src/OrchardCore/OrchardCore.Mvc.Core/Utilities/ControllerExtensions.cs
@@ -8,19 +8,23 @@ namespace OrchardCore.Mvc.Utilities
         /// Returns the proper actionresult for unauthorized or unauthenticated users.
         /// Will return a forbid when the user is authenticated.
         /// Will return a challenge when the user is not authenticated.
-        /// If a authentication scheme is specified, will return a challenge to it.
         /// </summary>
         /// <param name="controller"></param>
-        /// <param name="authenticationScheme"></param>
         /// <returns>The proper actionresult based upon if the user is authenticated</returns>
-        public static ActionResult ChallengeOrForbid(this Controller controller, string authenticationScheme = null)
-        {
-            if (!string.IsNullOrEmpty(authenticationScheme))
-            {
-                return controller.User.Identity.IsAuthenticated ? (ActionResult)controller.Forbid(new string[] { authenticationScheme }) : (ActionResult)controller.Challenge(new string[] { authenticationScheme });
-            }
+        public static ActionResult ChallengeOrForbid(this Controller controller)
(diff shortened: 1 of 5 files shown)


src/OrchardCore/OrchardCore.Mvc.Core/Utilities/ControllerExtensions.cs
```diff
@@ -8,11 +8,18 @@ public static class ControllerExtensions
...
         /// </summary>
         /// <param name="controller"></param>
+        /// <param name="authenticationScheme"></param>
         /// <returns>The proper actionresult based upon if the user is authenticated</returns>
-        public static ActionResult ChallengeOrForbid(this Controller controller)
+        public static ActionResult ChallengeOrForbid(this Controller controller, string authenticationScheme = null)
         {
+            if (!string.IsNullOrEmpty(authenticationScheme))
+            {
+                return controller.User.Identity.IsAuthenticated ? (ActionResult)controller.Forbid() : (ActionResult)controller.Challenge(new string[] { authenticationScheme });
```
kevinchalet, line 20: We should also call `Forbid(new string[] { authenticationScheme })` instead of just `Forbid()`.

deanmarcussen: I think there's an overload that just takes a string value to avoid creating `new string[]` unless the api has changed.

agriffard: There is no method with only a string: https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.controllerbase.challenge?view=aspnetcore-3.1

agriffard: `Forbid(new string[] { authenticationScheme })` added.

deanmarcussen: Just updated the extensions to match what is used in `ControllerBase`

kevinchalet (approved):

PR #7070, opened 2020-09-13
Prevent rendering same script multiple times by naming it

netwavebe:
There is a script in `Content.SummaryAdmin.cshtml` that is rendered for each content item. In the 'Manage Content' page this results in the same script being rendered multiple times.

You can verify this behaviour by going to /Admin/Contents/ContentItems and view the source:

```
<script type="text/javascript">
    $(function(){
        
        $("#actionsMenu.dropdown-menu").each(function(i, e){
            var count = $(e).children().length;
            if(count == 0){
                $(this).parent().hide();
            }
        });
    });
</script>
<script type="text/javascript">
    $(function(){
        
        $("#actionsMenu.dropdown-menu").each(function(i, e){
            var count = $(e).children().length;
            if(count == 0){
                $(this).parent().hide();
            }
        });
    });
</script>
<script type="text/javascript">
    $(function(){
        
        $("#actionsMenu.dropdown-menu").each(function(i, e){
            var count = $(e).children().length;
            if(count == 0){
                $(this).parent().hide();
            }
        });
    });
</script>
...
```

Solved by adding a name to the script.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.SummaryAdmin.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.SummaryAdmin.cshtml
@@ -43,7 +43,7 @@
                         <button type="button" class="btn btn-secondary btn-sm dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
                             @T["Actions"]
                         </button>
-                        <div id="actionsMenu" class="dropdown-menu dropdown-menu-right">
+                        <div class="actions-menu dropdown-menu dropdown-menu-right">
                             @await DisplayAsync(Model.ActionsMenu)
                         </div>
                     </div>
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.SummaryAdmin.cshtml
```diff
@@ -58,7 +58,7 @@
     <div class="col primary">@await DisplayAsync(Model.Content)</div>
 }
 
-<script at="Foot" type="text/javascript">
+<script asp-name="removeEmptyActionsMenu" at="Foot" type="text/javascript">
     $(function(){
         
         $("#actionsMenu.dropdown-menu").each(function(i, e){
```
deanmarcussen, line 64: Looks good. 

Do you mind changing the script where it uses an id of `actionsMenu` to use a class of `.actions-menu` as well?

I doubt it'll work properly rendering multiple id's in the dom, and is never good practise

netwavebe: You're absolutely right, Dean.

I added a second commit.

deanmarcussen (approved):
Nice one thanks @netwavebe

Issue #4489, opened 2019-10-07
[GraphQL] Subscriptions support


carlwoodhouse:
Supporting subscriptions would allow us to notify consumers of the api that content has changed; which they might use for invalidating caching, rebuilding gatsby sites, etc.

more info:
* https://graphql-dotnet.github.io/docs/getting-started/subscriptions/
* https://github.com/graphql-dotnet/server/blob/develop/samples/Samples.Server/Startup.cs* 
https://graphql.org/blog/subscriptions-in-graphql-and-relay/

MatthijsKrempel:
We have a simular request from the organisation, any pointers to how we should proceed?

sebastienros:
You might be able to add a subscription already

Start with something like this `services.AddSingleton<ISchemaBuilder, ContentItemSubscription>();` which will give you access to the `ISchema` that contains access to queries, mutations and subscriptions. From that just follow the documentation from GraphQL.NET

carlwoodhouse:
Yeah just follow what’s in the links; I think the only thing is you have to
enable web sockets .. plenty of examples about

MatthijsKrempel:
Ok, not quite familiar with it, but I will have a look.
I don't think it is as simple as adding a schema, we also have to have the middleware handling the request.

jrestall:
The sample link @carlwoodhouse provided is really good https://github.com/graphql-dotnet/server/blob/develop/samples/Samples.Server/Startup.cs

I would suggest implementing this under its own feature such as "OrchardCore.Apis.GraphQL.Subscriptions" so that WebSockets is only enabled when subscriptions are used. So you would call AddWebSockets/UseWebSockets from a SubscriptionsStartup class like how the Users module does it [here](https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.Users/Startup.cs).

As for what the schema would look like for subscriptions, we need to discuss. Here's some initial options:

```
type Subscription {
  contentItem(where: SubscriptionWhereInput) : ContentItemSubscriptionPayload

  # OPTION 1
  blogPost(where: BlogPostSubscriptionWhereInput): BlogPostSubscriptionPayload

  # OPTION 2
  blogPost(where: SubscriptionWhereInput): BlogPostSubscriptionPayload

  # OPTION 3
  blogPostCreated : BlogPost
  blogPostUpdated: BlogPost
  blogPostRemoved: BlogPost
[...]

PR #7073, opened 2020-09-14
GraphQL 3.0

agriffard:
Upgrade between 2.4.0 and 3.0.0.
https://github.com/graphql-dotnet/graphql-dotnet/releases

There are new features and many breaking changes.

Migration steps:
https://graphql-dotnet.github.io/docs/guides/migration3/

Many errors still need to be fixed before it can build.

Here are few notes on the changes applied in OC:

IFieldNameConverter => INameConverter 
  Type parentType => IComplexGraphType
  NameFor => NameForField 

ResolveFieldContext => IResolveFieldContext 

INodeVisitor Validate => Task<INodeVisitor> IValidationRule.ValidateAsync

Data loaders now return an IDataLoaderResult<T> rather than a Task<T>

/cc @Jetski5822


agriffard:
@carlwoodhouse Can you please also have a look at this PR?

carlwoodhouse:
@agriffard i can; need some time to go through and understand all the changes - especially around di and the execution contexts for dataloaders.

carlwoodhouse:
A quick glance tells me we could and should refactor some of the di stuff now it supports it properly. Probably could remove the asynclock resolver too if serialexecution policies support dataloaders. I'll try and dig into it more this week.

carlwoodhouse:
Only just realised you said it's not finished ;)

Quick list from a skim:

* We need to change DI to not use the UserContext but the first class support in RequestServices - this will be a breaking change for anyone who has custom objects already
* We should look at the changes to dataloaders and the execution strategy, we do use a custom exec strat at the moment so we coujld use parallel exec in a thread-safe way - but some of the changes might mean that could change and we could possibly remove the asynclock resolver.

There might be a question if this is A LOT to make it work as if its worth it vs a posisble shift to hotchocolate discussed here: https://github.com/OrchardCMS/OrchardCore/issues/4489#issuecomment-673614269 . This is probably less work then that still to be fair. Im happy to help out with it if you like @agriffard

agriffard:
Please have a look @carlwoodhouse.

agriffard:
@carlwoodhouse, @Jetski5822 Can you please have a look at it: There are only 4 errors left about dependency resolver and dataloader.

src/OrchardCore/OrchardCore.Apis.GraphQL.Abstractions/Extensions/FieldBuilderResolverExtensions.cs
```diff
@@ -1,14 +1,14 @@
...
 using GraphQL.Builders;
-using GraphQL.Types;
 using OrchardCore.Apis.GraphQL.Resolvers;
 
 namespace OrchardCore.Apis.GraphQL
 {
     public static class FieldBuilderResolverExtensions
     {
-        public static FieldBuilder<TSourceType, TReturnType> ResolveLockedAsync<TSourceType, TReturnType>(this FieldBuilder<TSourceType, TReturnType> builder, Func<ResolveFieldContext<TSourceType>, Task<TReturnType>> resolve)
+        public static FieldBuilder<TSourceType, TReturnType> ResolveLockedAsync<TSourceType, TReturnType>(this FieldBuilder<TSourceType, TReturnType> builder, Func<IResolveFieldContext<TSourceType>, Task<TReturnType>> resolve)
```
carlwoodhouse, line 11: we need to look if we can remove `ResolveLockedAsync` now dataloaders support serial execution strategies (we switched to a custom parallel strategy that uses locks to allow dataloader support; but a serial strategy with no locks is probably more performant and less error prone (ie, someone forgot to use ResolveLockedAsync instead of ResolveAsync which is the normal graphqldotnetresolver) - could be saved for another PR though imo.

src/OrchardCore.Modules/OrchardCore.Media/GraphQL/MediaFieldQueryObjectType.cs
```diff
@@ -26,7 +26,7 @@ public MediaFieldQueryObjectType()
                 .Resolve(x =>
                 {
                     var paths = x.Page(x.Source.Paths);
-                    var context = (GraphQLContext)x.UserContext;
+                    var context = (GraphQLUserContext)x.UserContext;
```
carlwoodhouse, line 29: use `.RequestServices.` on `x` You can just resolve the provider from there.

src/OrchardCore/OrchardCore.Apis.GraphQL.Abstractions/Extensions/ResolveFieldContextExtensions.cs
```diff
@@ -59,9 +60,9 @@ public static bool HasPopulatedArgument<TSource>(this ResolveFieldContext<TSourc
             return source;
         }
 
-        public static IServiceProvider ResolveServiceProvider<T>(this ResolveFieldContext<T> context)
+        public static IServiceProvider ResolveServiceProvider<T>(this IResolveFieldContext<T> context)
```
carlwoodhouse, line 63: we should be able to remove this extension.`IResolveFieldContext` now has `RequestServices` as a property

src/OrchardCore/OrchardCore.ContentManagement.GraphQL/Queries/ContentItemQuery.cs
```diff
@@ -47,7 +48,7 @@ public Task BuildAsync(ISchema schema)
             return Task.CompletedTask;
         }
 
-        private Task<ContentItem> ResolveAsync(ResolveFieldContext context)
+        private Task<ContentItem> ResolveAsync(IResolveFieldContext context)
         {
             var contentItemId = context.GetArgument<string>("contentItemId");
             var contentManager = _httpContextAccessor.HttpContext.RequestServices.GetService<IContentManager>();
```
carlwoodhouse, line 54: not really sure why this is using _httpContextAccessor; it should probably just use `context.RequestServices`

src/OrchardCore/OrchardCore.Apis.GraphQL.Abstractions/GraphQLUserContext.cs
```diff
@@ -1,11 +1,12 @@
 using System;
+using System.Collections.Generic;
 using System.Security.Claims;
 using System.Threading;
 using Microsoft.AspNetCore.Http;
 
 namespace OrchardCore.Apis.GraphQL
 {
-    public class GraphQLContext
+    public class GraphQLUserContext : Dictionary<string, object>
```
carlwoodhouse, line 9: maybe we can remove `ServiceProvider` from this file; the service provider should always be resolved through .RequestServices`; unless some other things then fieldresolvers use this - not entirely sure why this stores the HttpContext either; that should def be removed.

src/OrchardCore.Modules/OrchardCore.Media/GraphQL/MediaAssetObjectType.cs
```diff
@@ -19,7 +19,7 @@ public MediaAssetObjectType()
                 .Resolve(x =>
                 {
                     var path = x.Source.Path;
-                    var context = (GraphQLContext)x.UserContext;
+                    var context = (GraphQLUserContext)x.UserContext;
```
carlwoodhouse, line 22: use .RequestServices. on x You can just resolve the provider from there.

src/OrchardCore/OrchardCore.ContentManagement.GraphQL/Queries/ContentItemsFieldType.cs
```diff
@@ -64,9 +65,9 @@ public ContentItemsFieldType(string contentItemName, ISchema schema, IOptions<Gr
             _defaultNumberOfItems = settingsAccessor.Value.DefaultNumberOfResults;
         }
 
-        private async Task<IEnumerable<ContentItem>> Resolve(ResolveFieldContext context)
+        private async Task<IEnumerable<ContentItem>> Resolve(IResolveFieldContext context)
         {
-            var graphContext = (GraphQLContext)context.UserContext;
+            var graphContext = (GraphQLUserContext)context.UserContext;
```
carlwoodhouse, line 70: hink this doesnt need to be created and passed down the stack, the context is passed anyway and has the RequestServices on it

src/OrchardCore/OrchardCore.ContentManagement.GraphQL/Queries/Types/ContentItemType.cs
```diff
@@ -37,7 +37,7 @@ public ContentItemType(IOptions<GraphQLContentOptions> optionsAccessor)
                 .Name("render")
                 .ResolveLockedAsync(async context =>
                 {
-                    var userContext = (GraphQLContext)context.UserContext;
+                    var userContext = (GraphQLUserContext)context.UserContext;
                     var serviceProvider = userContext.ServiceProvider;
```
carlwoodhouse, line 41: just use `context.RequestServices.`

carlwoodhouse (changes requested):

PR #7077, opened 2020-09-15
Fix VS Code tasks.

Skrypt:
Fix Visual Studio Code tasks to allow to publish for Linux or Windows as self-contained or not.
Update also tasks.json to version 2.0.0.
Make Publish tasks to use /build/release for using same folder than CI.
Also, now cleans the /build/release folder before publishing to prevent mixing self-contained and non self-contained builds.


.dockerignore
```diff
@@ -1,5 +1,5 @@
 #ignore all files
-*/*
+**
 
 #except release
-!/.build/release/*
\ No newline at end of file
+!/build/release
```
Skrypt, line 5: Could someone try this on his CI to see if it's equivalent.

Skrypt:
I'm merging this one as it has no big incidence and it simply works better.

Issue #5126, opened 2019-12-25
Uniform button shape names for Admin


ns8482e:
For Admin, Different button shapes are used, It would be nice to have standard naming convention for all these button shapes, so that developer can create predictable alternates for Admin Theme.

e.g.
- ContentPreview_Button
- Content_PublishButton
- Content_SaveDraftButton

For example, I would like to show "Actions" zone as drop-down similar to https://github.com/OrchardCMS/OrchardCore/issues/3688#issuecomment-541206623 and overriding above shapes with alternates.  As following

`Content-Dropdown.PublishButton.cshtml` ( Content_PublishButton__Dropdown)
`Content-Dropdown.SaveDraftButton.cshtml` (Content_SaveDraftButton__Dropdown)

However for preview naming is different than above,

`ContentPreview-Dropdown.Button.cshtml `( ContentPreview_Button__Dropdown )

Notice the position of `-` in filenames above, its confusing.
 
I suggest to have all button shape uniform naming.

agriffard:
Can you provide a PR?

ns8482e:
Fix in PR #7086 is also related, due to not having uniform naming and shape type contains `__` and `_`
The original shape type should not contain `__` or `_`

I could wok on PR to change shape type  to not contain `_` or `__`.  
Also note that here prefix `Content` is not used to represent  `stereotype` of the content item

So change would be as as following, 

ContentPreview_Button --> ContentPreviewButton
Content_PublishButton --> ContentPublishButton
Content_SaveDraftButton --> ContentSaveDraftButton

agriffard:
Can this be closed?

PR #7086, opened 2020-09-16
Renamed  SummaryAdmin shapes of ContentsDriver

ns8482e:
Following `ContentsDriver` creates four shapes with different shape types. 

https://github.com/OrchardCMS/OrchardCore/blob/85b4f99618ba263b55cb72e366d973e1967b2298/src/OrchardCore.Modules/OrchardCore.Contents/Drivers/ContentsDriver.cs#L51-L54

However the actual shape type is considered before `__`  i.e. all above shapes are resolved to the same shape type `Contents_SummaryAdmin` 

Renamed these shapes as following to apply unique placement record for each shape

`Contents_SummaryAdmin__Tags`  renamed to  `ContentsTags_SummaryAdmin`
`Contents_SummaryAdmin__Meta` renamed to `ContentsMeta_SummaryAdmin`
`Contents_SummaryAdmin__Button__Edit` renamed to `ContentsButtonEdit_SummaryAdmin`
`Contents_SummaryAdmin__Button__Actions` renamed to `ContentsButtonActions_SummaryAdmin`

so that placement will target single shape type.

```
{
    "ContentsButtonActions_SummaryAdmin": [
      {
        "shape":"ContentsButtonEditNoView_SummaryAdmin"
      }
    ]    
}
```

Fixes #7002


deanmarcussen:
So the issue is that using a seperator in the shape name stops placement from hitting it?

My gut feel is that we might be better off renaming the shapes, than having a differentiator that noone will know about?

A differentiator would normally have something regarding a content item / part / field name, so it's a bit weird to be differentiating between zones.

Related https://github.com/OrchardCMS/OrchardCore/issues/5151 (feel free to add a comment about how placement is impacted by seperators @ns8482e

ns8482e:
> So the issue is that using a seperator in the shape name stops placement from hitting it?

Yes,  It's because original shape type is identified before `__`  in shape templates. Here! The original shape type for Contents_SummaryAdmin__Tags , Contents_SummaryAdmin__Meta, Contents_SummaryAdmin__Button__Edit, and Contents_SummaryAdmin__Button__Actions is Contents_SummaryAdmin hence placement is applied to all shapes


> Related #5151 (feel free to add a comment about how placement is impacted by seperators @ns8482e

Yes, I can update the PR to change ShapeType to to not have `_` or `__` and change the template name without using  differentiator.

e.g.

Contents_SummaryAdmin__Tags ----> SummaryAdminTags ---->SummaryAdminTags.cshtml
Contents_SummaryAdmin__Meta ----> SummaryAdminMeta ----> SummaryAdminMeta.cshtml
Contents_SummaryAdmin__Button__Edit ----> SummaryAdminButtonEdit ----> SummaryAdminButtonEdit.cshtml
Contents_SummaryAdmin__Button__Actions ----> SummaryAdminButtonActions ----> SummaryAdminButtonActions.cshtml

deanmarcussen:
> Contents_SummaryAdmin__Tags ----> SummaryAdminTags ---->SummaryAdminTags.cshtml

Wouldn't it be `Contents_SummaryAdmin__Tags` ----> `ContentsTagsSummaryAdmin` ?

ns8482e:
> Wouldn't it be `Contents_SummaryAdmin__Tags` ----> `ContentsTagsSummaryAdmin` ?

Sure It will work just thought to remove prefix `Contents` to avoid confusion with other shapes as Here it's not representing stereo type. These shapes will be use for all content items regardless of their stereo type.

deanmarcussen:
> > Wouldn't it be `Contents_SummaryAdmin__Tags` ----> `ContentsTagsSummaryAdmin` ?
> 
> Sure It will work just thought to remove prefix `Contents` to avoid confusion with other shapes as Here it's not representing stereo type. These shapes will be use for all content items regardless of their stereo type.

I think the `Contents` was more an indicator which driver it was produced by, not related to stereotype.

I would keep it, otherwise there is no indicator that the shape is even content related. 

It would be good to see the change, and I want to discuss this at triage 

- This could be achieved by a shape provider
- Should placement work for everything?
    - Placement has been historically focused on front end display, but there are now more requests / expectations that it will work with all the admin shapes
    - If so, we have a larger shape renaming job to come, and a better shape name convention that we need to follow, that would need to reduce seperators when they have no specific meaning (like in this case, the seperator has no meaning, it is just getting in they way)

ns8482e:
> * we have a larger shape renaming job to come, and a better shape name convention that we need to follow

Yes It's related #5151 and #5126 

> I think the `Contents` was more an indicator which driver it was produced by, not related to stereotype.

Ok, I agree for `Contents` prefix, if it's indicator of the driver  - Then I feel `SummaryAdmin` suffix is redundant, and the shape type should be without `displayType` suffix i.e.. `ContentsTags` 

And it's also align with the other shape `ContentsMetadata` produce by same driver but for `Detail` display type. 

So following placement would more appropriate

```
"ContentsTags": [
      {
        "displayType" : "SummaryAdmin",
        "place":"-"
      }
    ]
```

compared to


```
"ContentsTagsSummaryAdmin": [
      {
        "displayType" : "SummaryAdmin",
        "place":"-"
      }
    ]
```

deanmarcussen:
`ContentsTags_SummaryAdmin`

ns8482e:
> `ContentsTags_SummaryAdmin`

Read display type from context or hardcode in shape type? e.g.

`Shape("ContentsTags_SummaryAdmin", new ContentItemViewModel(model)).Location("SummaryAdmin", "Tags:10"),`
or
`Shape($"ContentsTags_{context.DisplayType}", new ContentItemViewModel(model)).Location("SummaryAdmin", "Tags:10"),`

sebastienros (approved):

PR #7104, opened 2020-09-19
Admin UI Permissions adjustments for EditOwnContent on a specific content type

Skrypt:
As a user I want to be able to create my own drafts and list them from the admin.
As a user I want to be able to see restricted content items from the admin content list.
As a user I want to be able to see only allowed content items from the admin content list.

## Repro 
Repro steps of the issue are : 

- Create a Content Type named "Acme".
- Create a new user Role named "SomeRole".
- Set "Edit own content" permission on that "Acme" content type in that "SomeRole" role.
- Create a new user named "NewUser" and set it's role to "SomeRole".
- Access the admin UI with that "NewUser" user and try to create a "Acme" content.

Now, you should be able to create a new content item by using the admin menu but never to see those content items in the Content elements admin page because you don't have access to it.

## PR Comments

Allows to see the admin content list if a user has a role that allows to do EditOwnContent on a specific content type.
Actually, when we do this the user cannot see the admin content list because the permission is set globally to EditContent which is to allow editing any content type. I added a Permission to allow to see the admin content list instead of evaluating if a user can edit at least one content type which is less demanding.

Also adding a new permission to allow to see restricted content items or not for a specific role. This defaults to **not showing** restricted content items which I think is the most common usage.

Removing the OrchardCore.ContentPreview Permissions.cs file in favor of using the OrchardCore.Contents Permissions.cs file permissions which we're not used at all.

Using CommonPermissions.*PermissionName* everywhere we can. Tested and working against older permissions since it is evaluating the Permission name set when we do `new Permission("PermissioName", "Description");` and not the assembly name.


hishamco:
Thanks Jasmin for starting this, I will review it once its ready, but first I need to understand the issue, coz during the previous the old PR I already did some checks for "Own" permissions, but @deanmarcussen pointed me to the comment where there's no need to check them, perhaps this another check, I hope @deanmarcussen review this too - if he has a time -

Thanks

Skrypt:
After doing a second check what you did was fine. Though the issue I'm trying to fix is the fact that when we create a new Role that has only some few permissions on a specific content type it's not working properly. So, for example, set access to the admin in a test role and add publish own content on a specific content type. If you want to see the content list you need to set Orchard Content "edit own content" permission at least.

So, the issue I found is that we never evaluate for specific content type permissions.

hishamco:
Is it ready for review?

Skrypt:
Yes go on, I think it needs a review now. Maybe some small adjustments left to do mainly on checking if everything works properly with permissions in different context.

src/OrchardCore.Modules/OrchardCore.Contents/Security/ContentTypeAuthorizationHandler.cs
```diff
@@ -33,89 +32,10 @@ protected override async Task HandleRequirementAsync(AuthorizationHandlerContext
...
-            if (permission == null)
-            {
-                return;
-            }
-
-            // Lazy load to prevent circular dependencies
-            var authorizationService = _serviceProvider.GetService<IAuthorizationService>();
-
-            if (await authorizationService.AuthorizeAsync(context.User, permission))
+            if (await ContentTypeAuthorizationHelper.AuthorizeDynamicPermissionAsync(_httpContextAccessor.HttpContext, requirement.Permission, contentItem))
```
deanmarcussen, line 35: This suggests that there was or is some issue with how the `ContentTypeAuthorizationHandler` was handling dynamic permissions.

I don't think we should have to be calling a helper in other places to check authorization, we should only ever need to call `AuthorizeAsync`. Otherwise we are fighting against the ASP.NET Authorization system, and lose extensibility.

But I don't understand the repro for the issue, so maybe some more info / repro

Skrypt: Let me try something and I'll get back to you on that. I believe that the issue is that the ContentTypeAuthorizationHandler never gets called. Or if it does, the ContentItem passed = null and therefore doesn't handle the current ContentType we should be authorizing.

Skrypt: Repro steps of the issue is : 

- Create a Content Type named "Acme".
- Create a new user Role named "SomeRole".
- Set "Edit own content" permission on that "Acme" content type in that "SomeRole" role.
- Create a new user named "NewUser" and set it's role to "SomeRole".
- Access the admin UI with that "NewUser" user and try to create a "Acme" content.

You will see that the button to save a draft is not displayed at all even though you have the permission to.
You can also try adding "Publish or unpublish Acme" permission on that "NewRole" and it won't show the publish button.

Skrypt: Ok I confirm that the ContentTypeAuthorizationHandler gets called. So I'm making a change now to this PR to revert it back on calling the AuthorizationService.

Skrypt:
The only thing remaining is making the Clone work. It fails retrieving the Clone Template.

src/OrchardCore.Modules/OrchardCore.Contents/Services/DefaultContentsAdminListFilter.cs
```diff
@@ -92,6 +92,11 @@ public async Task FilterAsync(ContentOptionsViewModel model, IQuery<ContentItem>
                 {
                     query.With<ContentItemIndex>(x => x.ContentType.IsIn(listableTypes.Select(t => t.Name).ToArray()));
                 }
+                else
+                {
+                    //We never list HtmlWidget in the Content Items List
+                    query.With<ContentItemIndex>(x => x.ContentType != "HtmlWidget");
```
Skrypt, line 98: I think we should never list any ContentType items which have a Stereotype. I need to do a double check on that one.

src/OrchardCore.Modules/OrchardCore.Contents/Drivers/ContentsDriver.cs
```diff
@@ -44,34 +54,59 @@ public override IDisplayResult Display(ContentItem model, IUpdateModel updater)
                         ctx.Shape.Metadata.Alternates.Add($"{stereotype}__ContentsMetadata");
                     }
                 });
+
+                results.Add(contentsMetadataShape);
+
+                var hasViewPermission = await _authorizationService.AuthorizeAsync(context.User, CommonPermissions.ViewContent, contentItem);
```
deanmarcussen, line 60: So I guess the advantage of doing all this in the view was that the shapes are built quickly, because this is the front end `DisplayAsync` call and the shapes are not placed.

By moving it here it's more work to render the front end.
(not much)

So is there any advantage to moving all of these out of the views?

For the most part though it looks like you found the issue, and a good solution

Skrypt: It's more for the `ActionsMenu:10` shape that sometimes was rendering without having any allowed action(s) to append to it's dropdown.

Skrypt: Reverted that code as it was redundant in the view.

src/OrchardCore.Modules/OrchardCore.Contents/Services/DefaultContentsAdminListFilter.cs
```diff
@@ -89,6 +89,11 @@ public async Task FilterAsync(ContentOptionsViewModel model, IQuery<ContentItem>
                 query.With<ContentItemIndex>(x => x.ContentType.IsIn(listableTypes.Select(t => t.Name).ToArray()));
             }
 
+            if(!await _authorizationService.AuthorizeAsync(_httpContextAccessor.HttpContext.User, Permissions.ViewAllAdminContentList))
+            {
+                query.With<ContentItemIndex>(x => x.Owner == _httpContextAccessor.HttpContext.User.Identity.Name);
```
deanmarcussen, line 94: I like this solution to listing only owned items, but I'm not sure about the name / description of it.

src/OrchardCore.Modules/OrchardCore.Contents/Permissions.cs
```diff
@@ -44,7 +46,9 @@ public Task<IEnumerable<Permission>> GetPermissionsAsync()
                 PreviewContent,
                 CloneContent,
                 CloneOwnContent,
-                AccessContentApi
+                AccessContentApi,
+                AccessAdminContentList,
```
deanmarcussen, line 50: Less sure if we need this extra permission. Does it add more confusion, or less?

src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.SummaryAdmin.cshtml
```diff
@@ -59,11 +72,11 @@
 }
 
 <script asp-name="removeEmptyActionsMenu" at="Foot" type="text/javascript">
-    $(function(){
-        
-        $(".actions-menu.dropdown-menu").each(function(i, e){
+    $(function () {
+
+        $(".actions-menu.dropdown-menu").each(function (i, e) {
```
deanmarcussen, line 77: Half these checks are now done in the driver? 
And the other half are still done in the views.

You said that was to not have an empty button when there were no items in the list?

So do we need the script / not need the script as well?

src/OrchardCore.Modules/OrchardCore.Contents/Views/Contents-Button-Actions.SummaryAdmin.cshtml
```diff
@@ -5,34 +5,37 @@
     ContentItem contentItem = Model.ContentItem;
     var hasPublished = await ContentManager.HasPublishedVersionAsync(contentItem);
     var hasDraft = contentItem.HasDraft();
-    var hasClonePermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.CloneContent, contentItem);
-    var hasDeletePermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.DeleteContent, contentItem);
-    var hasEditPermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.EditContent, contentItem);
-    var hasPreviewPermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.PreviewContent, contentItem);
-    var hasPublishPermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.PublishContent, contentItem);
+    var hasDeletePermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.CommonPermissions.DeleteContent, contentItem);
```
deanmarcussen, line 8: The common permissions are assigned = to the modules permission statically, so it doesn't matter which one we check.

It's possible we could (maybe another pr) just remove the modules permissions, to get rid of any confusion there.

src/OrchardCore.Modules/OrchardCore.ContentPreview/Views/ContentPreview.Button.cshtml
```diff
@@ -1,20 +1,18 @@
-@using Microsoft.AspNetCore.Authorization
 @using OrchardCore.ContentManagement
-@using OrchardCore.ContentPreview
 @inject IAuthorizationService AuthorizationService
 
 @{
     ContentItem contentItem = Model.ContentItem;
     var previewId = Guid.NewGuid();
-    var hasPreviewPermission = await AuthorizationService.AuthorizeAsync(User, Permissions.ContentPreview, contentItem);
+    var hasPreviewPermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.CommonPermissions.PreviewContent, contentItem);
```
deanmarcussen, line 7: The comparison done here is against `Permission.Name`, but changing this makes the permission Permissions.ContentPreview uselessish.

src/OrchardCore.Modules/OrchardCore.Contents/Permissions.cs
Skrypt: The other option would be to evaluate if a user has access to any content. This would be a little more intense than simply allowing or not access to that list. Before we we're allowing users with the Permission.EditContent but this is a permission that allows editing every content type (too global). I made sure that this permission is added by default for the roles that used to have access to that list. So, if you never had to restrict per content type you will maybe be confused but this is adding more fine tuned access level for people.

src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.SummaryAdmin.cshtml
Skrypt: That might be an optimization but not necessary.

src/OrchardCore.Modules/OrchardCore.Contents/Views/Contents-Button-Actions.SummaryAdmin.cshtml
Skrypt: Yes they are the same. But we should be using the CommonPermissions as they are shared globally. Here in the OrchardCore.Contents module it doesn't really matter but I think that we should still be using CommonPermissions for keeping "consistency" throughout the app.

src/OrchardCore.Modules/OrchardCore.ContentPreview/Views/ContentPreview.Button.cshtml
Skrypt: Permissions.ContentPreview and CommonPermissions.PreviewContent should be the same?

src/OrchardCore.Modules/OrchardCore.Contents/Services/DefaultContentsAdminListFilter.cs
Skrypt: Find one 🙈 😛

src/OrchardCore.Modules/OrchardCore.ContentPreview/Views/ContentPreview.Button.cshtml
Skrypt: Maybe not I'll take a look

src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.SummaryAdmin.cshtml
Skrypt: Ok I removed this code entirely as now it's shown/hidden appropriately based on permissions the user has.

src/OrchardCore.Modules/OrchardCore.Contents/Services/DefaultContentsAdminListFilter.cs
Skrypt: public static readonly Permission ViewRestrictedItemsAdminContentList = new Permission("ViewAllAdminContentList", "View restricted/unrestricted content item(s) in admin content list");

src/OrchardCore.Modules/OrchardCore.ContentPreview/Views/ContentPreview.Button.cshtml
Skrypt: Ok they are not the same. I'm looking if I did a mistake. If I understand Permissions.ContentPreview is a global permission. Then for OrchardCore.Contents module we have CommonPermissions.PreviewContent which is also a Global permission for all content types. So, basically it tells that we can disable the Preview globally in 2 places.

Permissions.ContentPreview is used in only 2 places : 
Here and in the PreviewController. So I believe that it should be replaced by OrchardCore.Contents.CommonPermissions.PreviewContent in both places.

Skrypt: Ok removed the OrchardCore.ContentPreview Permission.cs file.

src/OrchardCore.Modules/OrchardCore.Contents/Views/Contents-Button-Actions.SummaryAdmin.cshtml
Skrypt: Done

src/OrchardCore.Modules/OrchardCore.Contents/Views/_ViewImports.cshtml
```diff
@@ -12,7 +12,6 @@
 @using OrchardCore.ContentManagement.Metadata.Settings
 @using OrchardCore.Contents.Workflows.Activities
 @using OrchardCore.Contents.Workflows.ViewModels
-@using OrchardCore.Deployment
```
deanmarcussen, line 15: Why remove this and add it to the other templates?

src/OrchardCore.Modules/OrchardCore.Contents/Views/Contents-Button-Actions.SummaryAdmin.cshtml
```diff
@@ -1,38 +1,42 @@
...
 @{
     ContentItem contentItem = Model.ContentItem;
     var hasPublished = await ContentManager.HasPublishedVersionAsync(contentItem);
     var hasDraft = contentItem.HasDraft();
-    var hasClonePermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.CloneContent, contentItem);
-    var hasDeletePermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.DeleteContent, contentItem);
-    var hasEditPermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.EditContent, contentItem);
-    var hasPreviewPermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.PreviewContent, contentItem);
-    var hasPublishPermission = await AuthorizationService.AuthorizeAsync(User, OrchardCore.Contents.Permissions.PublishContent, contentItem);
+    var hasDeletePermission = await AuthorizationService.AuthorizeAsync(User, CommonPermissions.DeleteContent, contentItem);
```
deanmarcussen, line 9: So you changed everything to use `CommonPermissions` everywhere?

Did you keep the permissions file in `Contents` but not in `Preview`?

src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.SummaryAdmin.cshtml
```diff
@@ -57,15 +71,3 @@
 {
     <div class="col primary">@await DisplayAsync(Model.Content)</div>
 }
-
-<script asp-name="removeEmptyActionsMenu" at="Foot" type="text/javascript">
```
deanmarcussen, line 61: This was not working anyway. I can see the console error. (loads after jQuery)

src/OrchardCore.Modules/OrchardCore.Contents/Controllers/AdminController.cs
```diff
@@ -164,7 +172,7 @@ public async Task<IActionResult> List(ListContentsViewModel model, PagerParamete
                 {
                     if (ctd.GetSettings<ContentTypeSettings>().Listable)
                     {
-                        var authorized = await _authorizationService.AuthorizeAsync(User, Permissions.EditContent, await _contentManager.NewAsync(ctd.Name));
+                    var authorized = await _authorizationService.AuthorizeAsync(User, CommonPermissions.EditContent, await _contentManager.NewAsync(ctd.Name));
```
deanmarcussen, line 175: formatting?

deanmarcussen, line 175: and are you using `await _contentManager.NewAsync()` or are you using `new ContentItem()`?

src/OrchardCore.Modules/OrchardCore.Widgets/Views/_ViewImports.cshtml
```diff
@@ -2,12 +2,13 @@
     For more information on enabling MVC for empty projects, visit http://go.microsoft.com/fwlink/?LinkID=397860
 
 *@
+@using OrchardCore.ContentManagement;
+@using OrchardCore.DisplayManagement.Views;
+@using Microsoft.AspNetCore.Authorization;
```
deanmarcussen, line 7: why moving above taghelpers? most other places it is below?

src/OrchardCore.Modules/OrchardCore.Contents/Drivers/ContentsDriver.cs
```diff
@@ -44,34 +54,59 @@ public override IDisplayResult Display(ContentItem model, IUpdateModel updater)
...
+                var hasEditPermission = await _authorizationService.AuthorizeAsync(context.User, CommonPermissions.EditContent, contentItem);
+                var hasPublishPermission = await _authorizationService.AuthorizeAsync(context.User, CommonPermissions.PublishContent, contentItem);
+                var hasDeletePermission = await _authorizationService.AuthorizeAsync(context.User, CommonPermissions.DeleteContent, contentItem);
+                var hasPreviewPermission = await _authorizationService.AuthorizeAsync(context.User, CommonPermissions.PreviewContent, contentItem);
+                var hasClonePermission = await _authorizationService.AuthorizeAsync(context.User, CommonPermissions.CloneContent, contentItem);
```
deanmarcussen, line 65: I wish we weren't having to do this in a `Display` method, but I don't see a better way :(

deanmarcussen (commented):
Just what I said on Skype about the two other features Add Content To Deployment and Export To Deployment Target which are also adding to the `Actions` zone.

You might have missed that demo ;)

src/OrchardCore.Modules/OrchardCore.Contents/Controllers/AdminController.cs
sebastienros, line 175:
```suggestion
  // We want to list the content item if the user can edit their own items at least. It might display content items the user won't be able to edit though.
                    var authorized = await _authorizationService.AuthorizeAsync(User, CommonPermissions.EditContent, await _contentManager.NewAsync(ctd.Name) { Owner =  User.Username });
```

src/OrchardCore.Modules/OrchardCore.Widgets/Views/_ViewImports.cshtml
Skrypt: To be honest it's never the same everywhere but it feels more natural to add the using first to me.

Skrypt: See https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.Deployment/Views/_ViewImports.cshtml
It doesn't really matter.
What matters is that I'm adding `@using Microsoft.AspNetCore.Authorization;`

src/OrchardCore.Modules/OrchardCore.Contents/Views/Contents-Button-Actions.SummaryAdmin.cshtml
Skrypt: I removed the one in the Preview module because the other permission in the Contents module or the CommonPermissions.PreviewContent was not used anywhere.

hishamco:
Please resolve the conflicts

src/OrchardCore.Modules/OrchardCore.Contents/Views/_ViewImports.cshtml
Skrypt: Because else I get this error when trying to build these templates : 

```
'CommonPermissions' is an ambiguous reference between 'OrchardCore.Contents.CommonPermissions' and 'OrchardCore.Deployment.CommonPermissions' [E: \ Repositories \ OrchardCore \ src \ OrchardCore.Modules \ OrchardCore.Contents \ OrchardCore.Contents.csproj]
```

src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.SummaryAdmin.cshtml
Skrypt: Good I find this too hacky.

src/OrchardCore.Modules/OrchardCore.Contents/Drivers/ContentsDriver.cs
Skrypt: I'll think about this one. Maybe there is a better way.

src/OrchardCore.Modules/OrchardCore.Contents/Controllers/AdminController.cs
Skrypt: solved

Skrypt: Doesn't make any difference with the changes I made in the ContentTypeAuthorizationHandler.

Skrypt: Doesn't make any difference with the changes I made in the ContentTypeAuthorizationHandler. Unless you want me to revert the changes in the handler. Here passing the owner or not doesn't change the result.

Skrypt: Doesn't make any difference with the changes I made in the ContentTypeAuthorizationHandler. Unless you want me to revert the changes in the handler. Here passing the owner or not doesn't change the result.

> and are you using await _contentManager.NewAsync() or are you using new ContentItem()?

Only difference is that we are calling the events of the DefaultContentManager this way. Which I don't like, but gives the same result for us.

src/OrchardCore.Modules/OrchardCore.Contents/Views/Contents-Button-Actions.SummaryAdmin.cshtml
Skrypt: Here we need to add the check for permissions on wether a user can or cannot Export as JSON from the deployment module.

src/OrchardCore.Modules/OrchardCore.Contents/Controllers/AdminController.cs
```diff
@@ -114,17 +119,20 @@ public async Task<IActionResult> List(ListContentsViewModel model, PagerParamete
...
 
                 var creatableList = new List<SelectListItem>();
-                if (contentTypes.Any())
+                if (contentTypeDefinitions.Any())
                 {
-                    foreach (var contentTypeDefinition in contentTypes)
+                    foreach (var contentTypeDefinition in contentTypeDefinitions)
                     {
-                        creatableList.Add(new SelectListItem(new LocalizedString(contentTypeDefinition.DisplayName, contentTypeDefinition.DisplayName).Value, contentTypeDefinition.Name));
+                        if (await _authorizationService.AuthorizeAsync(context.User, CommonPermissions.EditContent, new ContentItem { ContentType = contentTypeDefinition.Name }))
```
Skrypt, line 132: Not to myself : Use content manager or not.

src/OrchardCore.Modules/OrchardCore.Contents/Views/Contents-Button-Actions.SummaryAdmin.cshtml
Skrypt: Else the other way around would be to specify each time `OrchardCore.Contents.CommonPermissions.SomePermission`

src/OrchardCore.Modules/OrchardCore.Contents/Services/DefaultContentsAdminListFilter.cs
```diff
@@ -78,7 +78,7 @@ public async Task FilterAsync(ContentOptionsViewModel model, IQuery<ContentItem>
                 {
                     if (ctd.GetSettings<ContentTypeSettings>().Listable)
                     {
-                        var authorized = await _authorizationService.AuthorizeAsync(_httpContextAccessor.HttpContext.User, Permissions.EditContent, new ContentItem { ContentType = ctd.Name });
+                        var authorized = await _authorizationService.AuthorizeAsync(_httpContextAccessor.HttpContext.User, CommonPermissions.EditContent, new ContentItem { ContentType = ctd.Name });
```
deanmarcussen, line 81: I'm still uncomfortable with the change to both the ContentTypeAuthorizationHandler and not using the contentManager.NewAsync() and setting the owner property if it's required.

i.e. being explicit.

Like as is done here https://github.com/OrchardCMS/OrchardCore/blob/5e3f2c37c93265ecf37a361ae3128d3a1afa7bfe/src/OrchardCore.Modules/OrchardCore.Contents/Controllers/AdminController.cs#L316

and many other places.

I can live with the change to `CommonPermissions` (I don't think it matters much either way to be honest)

Skrypt: Well Sebastien doesn't like that I change it to CommonPermissions and last week when I explained him why I was making the change in the ContentTypeAuthorizationHandler he said to keep it because he now understand that it's required for making "Own" permissions variations to return proper results. Now, for using the ContentManager or not, it's a question of taste at this point since it doesn't make any big difference.

I can maybe explain again today in the meeting why I made those changes and why I made them and show you what it fixes. I just merged latest dev yesterday on top of my website and needed to apply this PR on top of it so that I can see the publish/edit buttons for users that only have EditOwnContent on a specific content type.

Skrypt: Also if you look closely at the code in the ContentTypeAuthorizationHandler it doesn't make sense to always require to pass a Owner to that method since what we would pass is what we already are looking against in this Handler.

``` C#
        private static bool HasOwnership(ClaimsPrincipal user, ContentItem content)
        {
            if (user == null || content == null)
            {
                return false;
            }

            return user.Identity.Name == content.Owner;
        }
```

Here we pass a User in this method which we would get from the _HttpContextAccessor.HttpContext.User.Identity.Name. So we would basically do : `return _HttpContextAccessor.HttpContext.User.Identity.Name == _HttpContextAccessor.HttpContext.User.Identity.Name;` which is simply passing in that HasOwnership method for nothing.

Also, it means that I need to inject the IHttpContextAccessor everywhere I want to do fake content items and assign those the current Owner... which is an extra step done for nothing too.

Skrypt: I've re-read your comment and looked at the code here : 

https://github.com/OrchardCMS/OrchardCore/blob/5e3f2c37c93265ecf37a361ae3128d3a1afa7bfe/src/OrchardCore.Modules/OrchardCore.Contents/Controllers/AdminController.cs#L316-L319

It is true that in this case we are evaluating the same values against each others. Which is because we pass a "fake" Owner. But what we need to evaluate here is if the current user has right to create a content item of that content type only.

So we could easily do `if (!await _authorizationService.AuthorizeAsync(User, Permissions.EditContent, new ContentItem { ContentType = id }))`

So, we are assigning a Owner to the ContentItem before evaluating the permission but we could easily use a Fake ContentItem here again to evaluate without using the ContentManager.NewAsync() returned one.

The only reason we are assigning a Owner here on the ContentItem created is because we want to have it stored on the ContentItem for a later Auth eval.

Skrypt: Here what I propose is that we assign a Owner by default in the DefaultContentManager.NewAsync() method like @sebastienros suggested. And we remove the extra `contentItem.Owner = User.Identity.Name;` done everywhere we do so. I suggest we also keep the change I made in the ContentTypeAuthorizationHandler and use it with or without creating a ContentItem with the DefaultContentManager.

Here, I want to point on the fact that when we do a NewAsync() we should evaluate if the current user has the right to create this ContentItem before even calling this NewAsync() method logically too.

Skrypt:
I reverted the changes in the ContentTypeAuthorizationHandler, used the ContentManager everywhere as requested and I,m also passing the Owner everywhere. Doesn't make any difference to me so far. If I find an issue later on I'll advise.

src/OrchardCore/OrchardCore.Contents.Core/AuthorizationServiceExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// </summary>
+        public static async Task<bool> IsAuthorizedToAContentTypeAsync(this IAuthorizationService service, ClaimsPrincipal user, Permission requiredPermission, IEnumerable<ContentTypeDefinition> contentTypeDefinitions)
+        {
+            foreach(var contentTypeDefinition in contentTypeDefinitions)
+            {
+                var permission = GetOwnerVariation(requiredPermission);
+                var contentTypePermission = ContentTypePermissionsHelper.ConvertToDynamicPermission(permission);
+                permission = ContentTypePermissionsHelper.CreateDynamicPermission(contentTypePermission, contentTypeDefinition);
+                
+                if(await service.AuthorizeAsync(user, permission, new ContentItem{ ContentType = contentTypeDefinition.Name, Owner = user.Identity.Name }))
```
Skrypt, line 25: @jtkech @deanmarcussen The only thing remaining is to define if I should be injecting the ContentManager to create the fake ContentItem in this method. I personnally think that using the ContentManager everywhere is not a good solution everytime but I'd like to understand better your opinions about this.

Open to everyone comments.

Skrypt:
> Add Content To Deployment
> Just what I said on Skype about the two other features Add Content To Deployment and Export To Deployment Target which are also adding to the `Actions` zone.
> You might have missed that demo ;)

About this I looked and the permissions seems to be applied. These are validated in the Driver directly through a method named DoesUserHavePermissionsAsync() in the DeploymentPlanService.

Skrypt:
About "Add Content To Deployment and Export To Deployment Target" I looked and the permissions seems to be applied. These are validated in the Driver directly through a method named DoesUserHavePermissionsAsync() in the DeploymentPlanService.

src/OrchardCore/OrchardCore.Contents.Core/AuthorizationServiceExtensions.cs
jtkech: @Skrypt @deanmarcussen 

I'm fine here to use `new ContentItem` in place of `NewAsync()`

Hmm, maybe unless if an auth handler check on the contentItem resource other things than the type or the owner, things that would be updated / added by `NewAsync()` e.g. by an activating / activated handler.

Just now i'm thinking about an auth handler that would check if a content item has a given part / field, but the handler could inject the definition manager, and i don't think that we will support this kind of permission ;)

So because these other things would be coupled to the content type, the only parameter that we pass to `NewAsync()`, when `NewAsync()` is only used for permissions purposes, I'm fine to rather use `new ContentItem()`. And if all agree we could do the same in the other places where `NewAsync()` is only used for the same reason.

Skrypt: Yes, and without saying it too loud I think this might cause eventual security issues. Anyone could take the contentItem passed in the activating / activated handlers and modify intentionally the Owner == "Admin" so that he can bypass the actual security. Here is why I wanted also to use a strong identifier for user accounts ... see UserId PR comments. 😉

deanmarcussen: Ok so for me, it's actually what @jtkech said about Parts and extensibility.

I looked back at some old O1 security code I had for managing User Groups

The code is not exactly the same, O1 had `IAuthorizationServiceEventHandler` but it's got good similarities to an `AuthorizationHandler` and the code should transplant over if/when I need to.

The first thing the custom  `IAuthorizationServiceEventHandler` does is 
```
var part = context.Content.As<UserGroupPart>();
```

so by not calling `.NewAsync()` everywhere the `ContentItem` parts are not loaded onto the kernel, and there is a loss of extensibility.

Now to be fair, the O1 code I looked at wasn't really interested in whether or not to show the `New` content item menu. And technically it could look at the content definition as @jtkech says. But that's ...

- more complicated
- only necesary because here you are making a breaking change to the concept of never creating a `ContentItem` directly, and bypassing it the content manager. Which just feels like a bad / unnecessary modification.

Regards your security concerns @Skrypt I would keep it for exactly that reason. It allows a developer to write extensible code, that can attach different permissions. Think about User Grouping, and flexing ownership of content through the use of groups.

A developer will always be able to read content items that aren't owned by them. A simple query in a view through `ISession` would do it, and we've always said that security is about providing permissions to secure what you can do through the ui. Not preventing malicious developers.

Skrypt: Yeah, I see your point. Though we can create specific AuthorizationHandlers for each use cases now too. And we can call them by order of priority. Here, when we create a ContentTypeAuthorizationHandler, I believe it is for handling permissions related with ContentTypes. There is also nothing preventing creating a ContentPartAuthorizationHandler specific for the need of when using a UserGroupPart.

So, here if I would use the contentManager.NewAsync() method it would weld the ContentParts to the content item. But, it is not here that I would require to weld them but in the Controller or View that requires to render it. So, in this AuthorizationHandler, the only thing I should be authorizing is related with "ContentType" only.

Skrypt: I agree that bypassing the contentManager here in that case is not a necessity. Though, doing so, prevents from adding extra processing that would happen when processing those event handlers. So, over time it would prevent from adding extra processing time every time we need to evaluate a permission. If we can do so then I think it makes sense. This is why mainly I'm asking everyone to think about if it's a necessity here to use the ContentManager.NewAsync() in that specific place. Else, easier for me to just use it and merge the PR... 😉

deanmarcussen: > Else, it would be easier for me to just use ContentManager.NewAsync() and merge the PR..

Please just do this :) 

Don't kill extensibility... even if you can't see a use for it someone will need it one day.
This is not on the front end, so the processing cost of running a handler is not something to be super concerned about (IMHO)

Skrypt: I agree to disagree with your opinion on that one. Because it doesn't kill the extensibility when we Authorize a content item on a admin controller action we use the ContentManager. We use it when it's needed. But I will just do this because we're all tired of this PR 😄

Skrypt: Done, now IsAuthorizedToAContentTypeAsync is dependend on the IContentManager. Which means it will create a new contentItem for each ContentType present in the ContentDefinitions and it will trigger the handlers for each of those content types. So basically it will exponentially become slower if someone does something funky in it's ActivatedAsync/ActivatingAsync,  InitializingAsync/InitializedAsync events.

Skrypt:
Some fixes. I removed the permission checking for the RenderWhen for the Actions shape as we always display at least the "View" or edit button in all cases. Also ViewContent is always true unless it is removed from the Anonymous user. But in that case you won't be able to see anything.

PR #7106, opened 2020-09-19
Vue draggable Resource version

agriffard:
Set the correct url and version in ResourceManifest for vuedraggable.
https://www.npmjs.com/package/vuedraggable


Skrypt (approved):
LGTM

PR #7115, opened 2020-09-22
Admin UI : Return URL

Skrypt:
Return to admin content list after saving draft or publishing the first time.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/AdminMenu.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/AdminMenu.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Linq;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Localization;
 using OrchardCore.ContentManagement;
 using OrchardCore.ContentManagement.Metadata;
@@ -16,15 +17,18 @@ namespace OrchardCore.Contents
     {
         private readonly IContentDefinitionManager _contentDefinitionManager;
         private readonly IContentManager _contentManager;
+        private readonly IHttpContextAccessor _httpContextAccessor;
         private readonly IStringLocalizer S;
 
         public AdminMenu(
-            IStringLocalizer<AdminMenu> localizer,
             IContentDefinitionManager contentDefinitionManager,
-            IContentManager contentManager)
+            IContentManager contentManager,
+            IHttpContextAccessor httpContextAccessor,
+            IStringLocalizer<AdminMenu> localizer)
         {
             _contentDefinitionManager = contentDefinitionManager;
             _contentManager = contentManager;
+            _httpContextAccessor = httpContextAccessor;
             S = localizer;
         }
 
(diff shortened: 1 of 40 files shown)


src/OrchardCore.Modules/OrchardCore.Contents/AdminMenu.cs
```diff
@@ -55,6 +59,7 @@ public async Task BuildNavigationAsync(string name, NavigationBuilder builder)
                         var ci = await _contentManager.NewAsync(contentTypeDefinition.Name);
                         var cim = await _contentManager.PopulateAspectAsync<ContentItemMetadata>(ci);
                         var createRouteValues = cim.CreateRouteValues;
+                        createRouteValues.Add("returnurl", _httpContextAccessor.HttpContext.Request.PathBase + "/Admin/Contents/ContentItems");
```
jtkech, line 62: Here the singularity is to build a route value that is a route itself, we would need to build an `IUrlHelper` with an `IUrlHelperFactory` and a `ViewContext / ActionContext` and then use `.RouteUrl()` with some route values, as it is done through `NavigationManager.BuildMenuAsync()`. This is menu shapes that call `.BuildMenuAsync()` with a `ViewContext / ActionContext` param, but it is not passed when calling all `BuildNavigationAsync()`.

But there is an `IActionContextAccessor` and we have a `ViewContextAccessor` service, so by injecting an `IUrlHelperFactory` we could build an `IUrlHelper`.

OR use the `LinkGenerator` service, there are some methods that need additional params, with some methods you just need to also pass the current `HtpContext` from which it retrieves what it needs, see how it is used in `ContentTypesAdminNodeNavigationBuilder`, e.g. with `_linkGenerator.GetPathByRouteValues()`.

src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsAdminList-Create.cshtml
```diff
@@ -1,19 +1,22 @@
 @model ContentOptionsViewModel
+@{
+    var returnUrl = @Context.Request.Path;
```
jtkech, line 3: We have the following helper

    var returnUrl = FullRequestPath;

src/OrchardCore.Modules/OrchardCore.Contents/AdminMenu.cs
Skrypt: done

src/OrchardCore.Modules/OrchardCore.Contents/AdminMenu.cs
```diff
@@ -55,6 +63,12 @@ public async Task BuildNavigationAsync(string name, NavigationBuilder builder)
                         var ci = await _contentManager.NewAsync(contentTypeDefinition.Name);
                         var cim = await _contentManager.PopulateAspectAsync<ContentItemMetadata>(ci);
                         var createRouteValues = cim.CreateRouteValues;
+                        createRouteValues.Add("returnurl", _linkGenerator.GetPathByRouteValues(_httpContextAccessor.HttpContext, "", new
```
deanmarcussen, line 66: It probably matches case insensitive, but shouldn't it be `returnUrl` ?

hishamco: Why we introduce a static field for it, because it has been used in many places and to avoid case comparison

agriffard:
Why doesn't it return to the Edit page of the content item?

Skrypt:
> Why doesn't it return to the Edit page of the content item?

We have buttons for that matter "Publish and continue" and "Save draft and continue". So the normal behavior should be to return to the Content List after publishing or saving a new content item.

src/OrchardCore.Modules/OrchardCore.Contents/AdminMenu.cs
Skrypt: In fact if we look at the login page it's `ReturnUrl` but here I kept it lowercase. Maybe a task would be to make them all consistent throughout the app. Though, I'm pretty sure it will work any case.

jtkech: RouteValueDictionary is case insensitive so anyway it works

Personally, for lisibility i prefer camel case `returnUrl` as it is done in some places as `action`, `controller`, `contentItemId`, we also have pascal case in some other places as `ContentItemId`.

Skrypt: Yeah I made it consistent for the url as lowercase for cross-plat compatibility. The vars can remain camel cased.

PR #7116, opened 2020-09-22
Admin UI content list : Always remove non-listable content types

### LINKED ISSUE #6363: Content item type UI Listing

Skrypt:
Always remove non-listable content types from the content list.

Fixes #6363, Closes #6366

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Security/ContentTypeAuthorizationHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Security/ContentTypeAuthorizationHandler.cs
@@ -38,14 +38,14 @@ namespace OrchardCore.Contents.Security
 
             if (contentItem != null)
             {
-                if(contentItem.Owner != null)
+                if (contentItem.Owner != null) // If we want to evaluate a specific content item authorization
                 {
                     if (OwnerVariationExists(requirement.Permission) && HasOwnership(context.User, contentItem))
                     {
                         permission = GetOwnerVariation(requirement.Permission);
                     }
                 }
-                else
+                else // If we want to evaluate if a user has authorization on a ContentType globally
                 {
                     if (OwnerVariationExists(requirement.Permission))
                     {
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Security/ContentTypeAuthorizationHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Security/ContentTypeAuthorizationHandler.cs
@@ -38,7 +38,8 @@ namespace OrchardCore.Contents.Security
 
             if (contentItem != null)
             {
-                if(contentItem.Owner != null){
+                if(contentItem.Owner != null)
+                {
                     if (OwnerVariationExists(requirement.Permission) && HasOwnership(context.User, contentItem))
                     {
                         permission = GetOwnerVariation(requirement.Permission);
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Security/ContentTypeAuthorizationHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Security/ContentTypeAuthorizationHandler.cs
@@ -26,6 +26,7 @@ namespace OrchardCore.Contents.Security
                 return;
             }
 
+            // If we are not evaluating a ContentItem then return.
             if (context.Resource == null)
             {
                 return;
@@ -37,9 +38,18 @@ namespace OrchardCore.Contents.Security
 
             if (contentItem != null)
             {
-                if (OwnerVariationExists(requirement.Permission) && HasOwnership(context.User, contentItem))
+                if(contentItem.Owner != null){
+                    if (OwnerVariationExists(requirement.Permission) && HasOwnership(context.User, contentItem))
+                    {
+                        permission = GetOwnerVariation(requirement.Permission);
+                    }
+                }
+                else
                 {
-                    permission = GetOwnerVariation(requirement.Permission);
+                    if (OwnerVariationExists(requirement.Permission))
+                    {
+                        permission = GetOwnerVariation(requirement.Permission);
+                    }
                 }
             }
 
@@ -98,6 +108,11 @@ namespace OrchardCore.Contents.Security
             if (permission.Name == Permissions.PreviewContent.Name)
             {
                 return Permissions.PreviewOwnContent;
+            }
+
+            if (permission.Name == Permissions.CloneContent.Name)
+            {
+                return Permissions.CloneOwnContent;
             }
 
             return null;
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Services/DefaultContentsAdminListFilter.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Services/DefaultContentsAdminListFilter.cs
@@ -77,7 +77,8 @@ namespace OrchardCore.Contents.Services
             else
             {
                 var listTypeDefinitions = _contentDefinitionManager.ListTypeDefinitions();
-                var listableTypes = new List<ContentTypeDefinition>();
+                var userListableTypes = new List<ContentTypeDefinition>();
+                var nonListableTypes = new List<ContentTypeDefinition>();
                 foreach (var ctd in listTypeDefinitions)
                 {
                     if (ctd.GetSettings<ContentTypeSettings>().Listable)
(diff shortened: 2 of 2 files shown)


hishamco:
I think I already did a PR for that, let me check ..

deanmarcussen:
@hishamco did do a pr for this over here https://github.com/OrchardCMS/OrchardCore/pull/6366
and I felt it was better if we removed the `.Any()` check.

It reduces the sql rather than using `IsNotIn`.

hishamco:
I will review it again or we will keep this if it's better

Skrypt:
The issue is that ListableTypes can be --= null and therefore we never specify which ContentType A,B,C in the SQL Query. So. when it find a ListableType the Query does : "WHERE ContentType IS IN (A, B, C)" but when a user has none ListableTypes unrestricted then the Query returns every content items no matter which content type. So I believe we should in this case get all the non-listable content types and specify we don't want to list them simply. This is not supposed to break anything here as this is the normal behavior of that list.

Skrypt:
I tried by removing the if and it doesn't work.

```C#
                if (listableTypes.Any())
                {
                    query.With<ContentItemIndex>(x => x.ContentType.IsIn(listableTypes.Select(t => t.Name).ToArray()));
                }
```

src/OrchardCore.Modules/OrchardCore.Contents/Services/DefaultContentsAdminListFilter.cs
```diff
@@ -92,6 +92,23 @@ public async Task FilterAsync(ContentOptionsViewModel model, IQuery<ContentItem>
...
+                else
+                {
+                    // Make sure we remove ContentItems from non-listable ContentTypes if
+                    // user doesn't have any unrestricted listable ContentType
+                    var nonListableTypes = new List<ContentTypeDefinition>();
+                    foreach (var ctd in _contentDefinitionManager.ListTypeDefinitions())
+                    {
+                        if (!ctd.GetSettings<ContentTypeSettings>().Listable)
+                        {
+                            nonListableTypes.Add(ctd);
```
jtkech, line 104: First, above line 91 we are doing

    if (listableTypes.Any())
    {
        query.With<ContentItemIndex>(x => x.ContentType.IsIn(listableTypes.Select(t => t.Name).ToArray()));
    }

Normally it should work without `if (listableTypes.Any())` so that we don't need this else block

If we still need to use the above `if (listableTypes.Any())`, if we go in this else block we already know that all content types are not listable, so maybe resolve once `.ListTypeDefinitions()` and hold it in a var , then above try to reduce the collection, otherwise use the whole collection (if any value).

But i think we should not use `if (listableTypes.Any())` line 91, and if it doesn't work we have to fix it on YesSql side. In yesSql, e.g. for `IsIn`, if there is no value it uses the following that is always false

                    if (values.Count == 0)
                    {
                        builder.Append(" 1 = 0");
                    }

Skrypt: How would it know which Content Types are non-listable if we don't exclude them from that Query?
So here when we do IsIn() it adds on the Query : WHERE ContentType IN (1,2,3,4,5,6). But if we are a user that has access to none content type then the Query will not do WHERE ContentType IN ()... so it expects that we want all the ContentTypes that are in the ContentItemIndex table. Here, I don't understand how removing the if(listableTypes.Any()) condition would help at all. This listableTypes collection can have a count of 0 items even if you are an administrator on the website.

Skrypt: Ok so it will do exactly the same but the Query will be bigger in the end.

Skrypt: I'm sorry I don't see what you guys mean at all. 😄

jtkech: Oops, i missed the `_authorizationService.AuthorizeAsync()` thing ;)

Let me check it again

Skrypt: I just did a code optimization to make it more obvious and readable.

jtkech: Okay, same kind of remarks but not on the whole content type collection but those for a given user ;)

About resolving once this collection, you did it somewhere in your last change by only using one loop. But here, if a type is not authorized for an user it would need to be added to the `nonListableTypes`, even it is listable.

But again, maybe i missed something else, the `IsIn` should work on an empty collection, so that if there is no content type that is authorized and listable, we just have to filter with a `IsIn` on an empty collection.

E.g. with the last change there is still a path where no filter is done at all.

But maybe i'm wrong, just as i understand the code

Good candidate for a skipe if you want ;)

src/OrchardCore.Modules/OrchardCore.Contents/Services/DefaultContentsAdminListFilter.cs
```diff
@@ -81,17 +78,15 @@ public async Task FilterAsync(ContentOptionsViewModel model, IQuery<ContentItem>
                 {
                     if (ctd.GetSettings<ContentTypeSettings>().Listable)
                     {
-                        var authorized = await _authorizationService.AuthorizeAsync(_httpContextAccessor.HttpContext.User, Permissions.EditContent, await _contentManager.NewAsync(ctd.Name));
+                        var authorized = await _authorizationService.AuthorizeAsync(_httpContextAccessor.HttpContext.User, Permissions.EditContent, new ContentItem { ContentType = ctd.Name });
```
deanmarcussen, line 81: @Skrypt I don't think this is a good / necessary change?

Skrypt: I think it is appropriate. Here we fake a content item only and it is not necessary to use the contentmanager to do so. In fact the contentmanager will trigger the events which here we don't want to since it's just a fake content item.

Skrypt: In fact here this is the reason why content items we're not displaying properly. It was never evaluating if the current user has the EditContent permission globally on each content type. This is why the ListableContentType dictionary was always empty. The ContentTypeAuthorizationHandler was always returning false because it was expecting to receive a real ContentItem with a ContentItem.Owner != null. This is why I added a else statement in the handler. Here if you say that the contentManager was necessary because we wanted to throw the events on the CreateAsync method then I made a mistake but I don't think so since we should only be listing content items in that list.

So faking the ContentItem by using a new ContentItem instanciation here makes sense to me because it will be faster than using the ContentManager to create a fake ContentItem and throw it's events ...

Skrypt: Let me know if it breaks something I didn't thought of.

src/OrchardCore.Modules/OrchardCore.Contents/Security/ContentTypeAuthorizationHandler.cs
```diff
@@ -37,9 +38,19 @@ protected override async Task HandleRequirementAsync(AuthorizationHandlerContext
...
-                if (OwnerVariationExists(requirement.Permission) && HasOwnership(context.User, contentItem))
+                if (contentItem.Owner != null) // If we want to evaluate a specific content item authorization
                 {
-                    permission = GetOwnerVariation(requirement.Permission);
+                    if (OwnerVariationExists(requirement.Permission) && HasOwnership(context.User, contentItem))
+                    {
+                        permission = GetOwnerVariation(requirement.Permission);
+                    }
+                }
+                else // If we want to evaluate if a user has authorization on a ContentType globally
```
Skrypt, line 48: @sebastienros Here is what I added for allowing to not need to inject the IHttpContextAccessor everywhere we need to evaluate if the user can create content items of a certain content type. There was only 2 places in the code where we need this so far and people made mistake of not passing the ContentItem.Owner so it was failing to evaluate. Here, we also would evaluate the same value against each other because the "patch" would be to call the authorization service like this : AuthorizationService.AutorizeAsync(IUser, IPermission, _httpContext.User.Identity.Name). And you can see on line 129 that we evaluate ContentItem.Owner = User.Identity.Name ... which is always going to return true ...

src/OrchardCore.Modules/OrchardCore.Contents/Services/DefaultContentsAdminListFilter.cs
Skrypt: @sebastienros Here we are debating on if we should keep using the ContentManager to create a fake content item to evaluate. In case someone would have used the events to extend something when evaluating permissions when listing the content list ?!? Not sure this was a good call to use the ContentManager here so I removed it.

PR #7117, opened 2020-09-22
Enhance Docker story

Skrypt:
The issue we have is that the .dockerignore file doesn't allow to build a docker image from source code directly.
One of the common practices with Docker is to build your source code in an intermediary image.
Right now we only allow to use precompiled assets in the `/build/release/` folder so if I want to compile what's in `/src` I can't.

In this PR I : 

- Move CI's to use `Dockerfile-CI`.
- Move CI's to use `Dockerfile-CI.dockerignore`. https://github.com/moby/moby/issues/12886#issuecomment-480575928
- Make the default `Dockerfile` use an intermediary image to build from `/src` folder. This is what most people will want to do when trying to build their own container for testing locally. This allows to test builds against different OS targets locally.
- Adjust `.dockerignore` file to allow building OC in a Docker container by copying `/src` to intermediary image in Docker. 

It is not necessary to have prebuilt OC in `/build/release` to build an OC Docker image anymore.

Here some reading : https://testdriven.io/blog/faster-ci-builds-with-docker-cache/

## Comments

To remove intermediary containers you simply need to use `docker build --rm` command.
To remove intermediary images you simply need to prune them after building by labelling images in your Dockerfile and using `docker image prune` command.

```
docker build -t oc --rm .
docker image prune -f --filter label=stage=build-env
```

For documentation @sebastienros last time I did one on Docker it was so bad that I removed the PR 😉 
Do we really need one if I remove the `docker-compose.yml` file? I think the remaining is general knowledge.
And I'm not too much into doing a second documentation on this. 💔


.travis.yml
```diff
@@ -1,52 +1,52 @@
...
+    - if test "$TRAVIS_OS_NAME" == "linux"; then export PATH="/home/travis/.dotnet":"$PATH"; fi
+    - if test "$TRAVIS_OS_NAME" == "osx"; then export PATH="/Users/travis/.dotnet":"$PATH"; fi
+    - dotnet --info
+    - dotnet build -c Release
+    - if [[ "$TRAVIS_BRANCH" == "master" ]] && [[ "$TRAVIS_OS_NAME" == "linux" ]] && [[ "$TRAVIS_PULL_REQUEST" == "false" ]]; then
+          cd $TRAVIS_BUILD_DIR/src/OrchardCore.Cms.Web;
+          rm -rf ./App_Data;
+          dotnet publish -c Release -o $TRAVIS_BUILD_DIR/.build/release;
+          cd $TRAVIS_BUILD_DIR;
+          docker build -f Dockerfile-CI -t orchardproject/orchardcore-cms-linux:latest -t orchardproject/orchardcore-cms-linux:1.0.0-rc2 .;
```
Skrypt, line 39: building using Dockerfile-CI is the only real change here

docker-compose.yml
```diff
@@ -0,0 +1,44 @@
+version: '3.3'
+services:
+    web:
+        build: 
+            context: .
+            dockerfile: Dockerfile
```
Skrypt, line 6: `docker-compose` will use naturally Dockerfile. I added this to make sure or if anyone would like to change it as an example.

Dockerfile
```diff
@@ -1,12 +1,12 @@
-# This Docker file is intended for the CI
-# A prerequisite is a published application in the .build/release  
-
-FROM mcr.microsoft.com/dotnet/core/aspnet:3.1
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1 AS build-env
```
Skrypt, line 1: Uses an intermediary container to build OC in the same OS type (Windows/Linux) than Docker runs.

.dockerignore
```diff
@@ -1,5 +1,9 @@
-#ignore all files
-*/*
+#ignore all
+**
 
-#except release
-!/.build/release/*
```
Skrypt, line 5: Here not sure if CI uses this .dockerignore file at all. Else we need to add back this line.

agriffard:
Can you please review @jtkech?

.dockerignore
Skrypt: Yes it does. Issue fixed by using 2 a Dockerfile-CI.dockergnore file.

.dockerignore
```diff
@@ -1,5 +1,12 @@
...
\ No newline at end of file
+#except src for build
+!./src/*
+
+#ignore any App_Data folder
+**/App_Data/
```
Skrypt, line 8: This is not necessary but it is consistent with what we do in the CI.

deanmarcussen (commented):
I think it's fine based on last nights triage discussion. 
/cc @sebastienros 
I imagine that we are only able to test drive the CI by merging it, so suggest we see if Seb has any comments, and whoever merges it (@Skrypt ) needs to check the size (don't want a 2gb image again) and test drive image that ends up on dockerhub

Skrypt:
If the CI picks up the Dockerfile-CI.dockerignore file it should only allow to get files from the /build/release folder.
If it doesn't, then it will fallback to the .dockerignore file and then make the size of those images bigger which will be a good indicator.

agriffard:
Please solve the conflicts.

jptissot:
Do we still need this with the ci improvements? I guess it could be useful for folks that want to build locally with docker.

PR #7119, opened 2020-09-22
Media profiles feature

deanmarcussen:
Adds a media profile feature allowing you to specify resizing options.

A profile can then be called with the profile name `resize_url: profile: 'banner'` rather than having to specify all the resizing options that may apply.

- [x] liquid filters
- [x] razor helper
- [x] tag helpers
- [x] docs
- [x] recipes for agency and blog theme


src/OrchardCore.Modules/OrchardCore.Media/Razor/OrchardRazorHelperExtensions.cs
```diff
@@ -28,6 +32,33 @@ public static string AssetUrl(this IOrchardHelper orchardHelper, string assetPat
...
             resizedUrl = fileVersionProvider.AddFileVersionToPath(orchardHelper.HttpContext.Request.PathBase, resizedUrl);
         }
+
+        return resizedUrl;
+    }
+
+    /// <summary>
+    /// Returns the relative URL of the specified asset path for a media profile with optional resizing parameters.
+    /// </summary>
+    public static async Task<string> AssetProfileUrlAsync(this IOrchardHelper orchardHelper, string assetPath, string imageProfile, int? width = null, int? height = null, ResizeMode resizeMode = ResizeMode.Undefined, bool appendVersion = false, int? quality = null, Format format = Format.Undefined)
```
deanmarcussen, line 42: To make this non breaking, and async, I have made another helper called `AssetProfileUrlAsync`, which goes alongside the `AssetUrl` helper.

Otherwise it breaks the `AssetUrl` helper by making it `AssetUrlAsync`

deanmarcussen:
Marking this as ready again, as there is nothing from the cropping feature that needs to be applied to this pr.

It will need to go in before I continue with https://github.com/OrchardCMS/OrchardCore/pull/7209 or it will be merge conflict hell ;)

Skrypt:
The **alt** attribute can be used on the following elements: area, img, input.
The **alt** attribute is required only for the  element.
The **title** attribute is part of the Global Attributes, and can be used on any HTML element.

So, question is should we use MetaData or a fixed property in the code for this.
I think we could have a fixed Title property for everything but the Alt could be simply using the Title property.
Else MetaData is fine but we need to make sure that all the images in the Media renders an alt attribute with some text even if it's the image name itself.

Skrypt (commented):
Here 3 facts : 

- The **alt** attribute can be used only on the following elements: `<area>`, ``, `<input>`.
- The **alt** attribute is **_required_** only for the `` element.
- The **title** attribute is part of the Global Attributes, and can be used on any HTML element.

So, question is should we use MetaData or a fixed property in the code for this.
I think we could have a fixed Title property for everything but the Alt could be simply using the Title property.
Else MetaData is fine but we need to make sure that all the images in the Media renders an alt attribute with some text even if it's the image name itself.

deanmarcussen:
> So, question is should we use MetaData or a fixed property in the code for this.

That's all on the other pr @Skrypt 

But I'm always open to naming options. At the moment it's called alt text because that tends to be what people think of it as (even though technically it's only `alt`). But we could call it `Metadata`

Skrypt (approved):
LGTM. I've not seen anything sensible to break something. Just good additions.

PR #7120, opened 2020-09-22
Fix Docker example in README.md

lomholdt:
Expose port in Docker example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Here is a more detailed [roadmap](https://github.com/OrchardCMS/OrchardCore/wiki
 - Run `docker run --name orchardcms -p 8080:80 orchardproject/orchardcore-cms-linux:latest`
 
 Docker images and parameters can be found at <https://hub.docker.com/u/orchardproject/>  
-See Docker documentation https://docs.docker.com/engine/reference/commandline/run/#publish-or-expose-port--p---expose to expose different port.
+See [Docker documentation](https://docs.docker.com/engine/reference/commandline/run/#publish-or-expose-port--p---expose) to expose different port.
 
 ### Documentation
 
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Here is a more detailed [roadmap](https://github.com/OrchardCMS/OrchardCore/wiki
 
 ### Docker
 
-- Run `docker run --name orchardcms orchardproject/orchardcore-cms-linux:latest`
+- Run `docker run --name orchardcms -p 80:80 orchardproject/orchardcore-cms-linux:latest`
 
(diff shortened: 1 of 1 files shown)


Skrypt:
Not necessary since port 80 could be used by IISExpress or apache on the running machine. This is something that is documented in Docker documentation as to how to expose the service port. The example shown in our documentation should not specify -p 80:80 but something like -p SomeAvailablePortInt:SomeAvailablePortInt or something better. So adding the -p doesn't garantee it will work instantly in every context. Maybe add a comment instead under the command itself.

```
// See https://docs.docker.com/engine/reference/commandline/run/#publish-or-expose-port--p---expose to expose port
docker run --name orchardcms orchardproject/orchardcore-cms-linux:latest
```

lomholdt:
@Skrypt I believe we can never guarantee it will work in every context. We could also just expose a more random port? Maybe -p 8080:80 or something like that?

README.md
```diff
@@ -51,7 +51,7 @@ Here is a more detailed [roadmap](https://github.com/OrchardCMS/OrchardCore/wiki
 
 ### Docker
 
-- Run `docker run --name orchardcms orchardproject/orchardcore-cms-linux:latest`
+- Run `docker run --name orchardcms -p 80:80 orchardproject/orchardcore-cms-linux:latest`
```
Skrypt, line 54:
```suggestion
- Run `docker run --name orchardcms -p 8080:80 orchardproject/orchardcore-cms-linux:latest`
```

lomholdt: Awesome! I love it 👍

README.md
```diff
@@ -51,7 +51,7 @@ Here is a more detailed [roadmap](https://github.com/OrchardCMS/OrchardCore/wiki
 
 ### Docker
 
-- Run `docker run --name orchardcms orchardproject/orchardcore-cms-linux:latest`
+- Run `docker run --name orchardcms -p 8080:80 orchardproject/orchardcore-cms-linux:latest`
 
 Docker images and parameters can be found at <https://hub.docker.com/u/orchardproject/>
```
Skrypt, line 56:
```suggestion
Docker images and parameters can be found at <https://hub.docker.com/u/orchardproject/>
See Docker documentation https://docs.docker.com/engine/reference/commandline/run/#publish-or-expose-port--p---expose to expose different port.
```

README.md
```diff
@@ -51,9 +51,10 @@ Here is a more detailed [roadmap](https://github.com/OrchardCMS/OrchardCore/wiki
...
```
Skrypt, line 56:
```suggestion
Docker images and parameters can be found at <https://hub.docker.com/u/orchardproject/>  
```

Skrypt (approved):

PR #7121, opened 2020-09-22
Password visibility : small fix MS-Edge and IE

Skrypt:
Hides the automatically displayed eye icon on MS-edge and IE.


Skrypt:
@hishamco please merge if it's fine. Don't pass too much time on that one.

hishamco:
Hope to try it on Edge tonight, please let me know which version of IE that the fix is working on?

Skrypt:
I've had the issue on the latest Microsoft Edge installed in the new Windows 10 20H2. I believe though that the same CSS gets applied on IE11 and other Edge versions. Worst case scenario is that this custom CSS is not necessary for some IE versions.

PR #7129, opened 2020-09-23
Prevent disabling or removing administrator role

### LINKED ISSUE #7125: Admin user became disabled
Did QA with TryOrchardCore.Web src.
Repro steps:

1. Build SAAS recipe orchard core site
image

2. Login to admin
3. Go to Security - >  users, and edit user you specify on setup
image

4. Save the user
image

User disabled.
image

Skrypt:
Add validations to prevent disabling or removing administrator role on the last user account marked as so.

Closes #7125

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
@@ -194,7 +194,7 @@ namespace OrchardCore.Users.Drivers
                     foreach (var role in rolesToRemove)
                     {
                         // Make sure we always have at least one administrator account
-                        if(usersOfAdminRole.Count == 1 && usersOfAdminRole.First().UserName == user.UserName && role == "Administrator")
+                        if (usersOfAdminRole.Count == 1 && usersOfAdminRole.First().UserName == user.UserName && role == "Administrator")
                         {
                             _notifier.Warning(H["Cannot remove administrator role from unique administrator."]);
                             continue;
--- a/src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
@@ -83,6 +83,7 @@ namespace OrchardCore.Users.Drivers
         public override async Task<IDisplayResult> UpdateAsync(User user, UpdateEditorContext context)
         {
             var model = new EditUserViewModel();
+            var httpContext = _httpContextAccessor.HttpContext;
 
             if (!await context.Updater.TryUpdateModelAsync(model, Prefix))
             {
@@ -95,7 +96,7 @@ namespace OrchardCore.Users.Drivers
             var usersOfAdminRole = await _userManager.GetUsersInRoleAsync("Administrator");
             if (!model.IsEnabled && user.IsEnabled)
             {
-                if (user.UserName != _httpContextAccessor.HttpContext.User.Identity.Name)
+                if (user.UserName != httpContext.User.Identity.Name)
                 {
                     if(usersOfAdminRole.Count == 1 && usersOfAdminRole.First().UserName == user.UserName)
                     {
@@ -140,7 +141,7 @@ namespace OrchardCore.Users.Drivers
                 }
             }
 
-            if (model.UserName != user.UserName && user.UserName == _httpContextAccessor.HttpContext.User.Identity.Name)
+            if (model.UserName != user.UserName && user.UserName == httpContext.User.Identity.Name)
             {
                 context.Updater.ModelState.AddModelError(string.Empty, S["Cannot modify user name of the currently logged in user."]);
             }
--- a/src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
@@ -98,7 +98,7 @@ namespace OrchardCore.Users.Drivers
             {
                 if (user.UserName != httpContext.User.Identity.Name)
                 {
-                    if(usersOfAdminRole.Count == 1 && usersOfAdminRole.First().UserName == user.UserName)
+                    if (usersOfAdminRole.Count == 1 && usersOfAdminRole.First().UserName == user.UserName)
                     {
                         _notifier.Warning(H["Cannot disable unique administrator."]);
                     }
--- a/src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
@@ -29,6 +29,7 @@ namespace OrchardCore.Users.Drivers
         private readonly ILogger _logger;
         private readonly IStringLocalizer S;
         private readonly IHtmlLocalizer H;
+        private static readonly string _administratorRole = "Administrator";
 
         public UserDisplayDriver(
             UserManager<IUser> userManager,
@@ -48,7 +49,6 @@ namespace OrchardCore.Users.Drivers
             _notifier = notifier;
             _logger = logger;
             Handlers = handlers;
-
             H = htmlLocalizer;
             S = stringLocalizer;
         }
@@ -93,7 +93,7 @@ namespace OrchardCore.Users.Drivers
             model.UserName = model.UserName?.Trim();
             model.Email = model.Email?.Trim();
 
-            var usersOfAdminRole = await _userManager.GetUsersInRoleAsync("Administrator");
+            var usersOfAdminRole = await _userManager.GetUsersInRoleAsync(_administratorRole);
             if (!model.IsEnabled && user.IsEnabled)
             {
                 if (usersOfAdminRole.Count == 1 && usersOfAdminRole.First().UserName == user.UserName)
@@ -192,7 +192,7 @@ namespace OrchardCore.Users.Drivers
                     foreach (var role in rolesToRemove)
                     {
                         // Make sure we always have at least one administrator account
-                        if (usersOfAdminRole.Count == 1 && usersOfAdminRole.First().UserName == user.UserName && role == "Administrator")
+                        if (usersOfAdminRole.Count == 1 && usersOfAdminRole.First().UserName == user.UserName && role == _administratorRole)
                         {
                             _notifier.Warning(H["Cannot remove administrator role from the only administrator."]);
                             continue;
--- a/src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
@@ -194,7 +194,7 @@ namespace OrchardCore.Users.Drivers
                         // Make sure we always have at least one administrator account
                         if (usersOfAdminRole.Count == 1 && usersOfAdminRole.First().UserName == user.UserName && role == "Administrator")
                         {
-                            _notifier.Warning(H["Cannot remove administrator role from unique administrator."]);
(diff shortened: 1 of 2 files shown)


src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
```diff
@@ -79,11 +93,26 @@ public override async Task<IDisplayResult> UpdateAsync(User user, UpdateEditorCo
...
             if (!model.IsEnabled && user.IsEnabled)
             {
-                user.IsEnabled = model.IsEnabled;
-                var userContext = new UserContext(user);
-                await Handlers.InvokeAsync((handler, context) => handler.DisabledAsync(userContext), userContext, _logger);
+                if (user.UserName != httpContext.User.Identity.Name)
+                {
+                    if(usersOfAdminRole.Count == 1 && usersOfAdminRole.First().UserName == user.UserName)
+                    {
+                        _notifier.Warning(H["Cannot disable unique administrator."]);
```
hishamco, line 103: Cannot disable the super administrator.

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
```diff
@@ -79,11 +93,26 @@ public override async Task<IDisplayResult> UpdateAsync(User user, UpdateEditorCo
...
 
+            var usersOfAdminRole = await _userManager.GetUsersInRoleAsync("Administrator");
             if (!model.IsEnabled && user.IsEnabled)
             {
-                user.IsEnabled = model.IsEnabled;
-                var userContext = new UserContext(user);
-                await Handlers.InvokeAsync((handler, context) => handler.DisabledAsync(userContext), userContext, _logger);
+                if (user.UserName != httpContext.User.Identity.Name)
+                {
+                    if(usersOfAdminRole.Count == 1 && usersOfAdminRole.First().UserName == user.UserName)
```
hishamco, line 101: Can we move this condition at the beginning, because if there's one use we are 100% sure it's a super admin

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
```diff
@@ -21,22 +24,32 @@ public class UserDisplayDriver : DisplayDriver<User>
...
+        private readonly INotifier _notifier;
         private readonly ILogger _logger;
         private readonly IStringLocalizer S;
+        private readonly IHtmlLocalizer H;
 
         public UserDisplayDriver(
             UserManager<IUser> userManager,
             IRoleService roleService,
             IUserRoleStore<IUser> userRoleStore,
+            IHttpContextAccessor httpContextAccessor,
```
hishamco, line 37: I don't think we need to inject `IHttpContextAccessor` just checking the `user` will enough

hishamco (changes requested):
Checking the first user in Administrator roles with the updated or deleted user at the beginning may be useful instead of inner conditions

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
Skrypt: Hmm I thought that the User was the current data from the database and model is the one coming from the form. So if we want to check who's the current user we need to get it from the context?

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
Skrypt: Well technically it's not a super administrator but a simple administrator. I get what you mean though.

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
Skrypt: Which line you want me to move it to? Because here I want to display the proper notifier warning.

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
hishamco: Still I feel "unique" is annoying ;)

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
hishamco: I mean line 101, because when the admin users is 1, a notification should popup

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
```diff
@@ -157,7 +193,16 @@ public override async Task<IDisplayResult> UpdateAsync(User user, UpdateEditorCo
 
                     foreach (var role in rolesToRemove)
                     {
-                        await _userRoleStore.RemoveFromRoleAsync(user, _userManager.NormalizeName(role), default(CancellationToken));
+                        // Make sure we always have at least one administrator account
+                        if(usersOfAdminRole.Count == 1 && usersOfAdminRole.First().UserName == user.UserName && role == "Administrator")
```
agriffard, line 197: Please format document.

hishamco: Nice catch ;)

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
Skrypt: Ok reversed both `if` conditions priority.

Skrypt:
I'm thinking about his PR and we use the UserManager directly in the UserDisplayDriver but what will happen if we create/remove a user from a Workflow or from any event raised? I find this kind of tiedous that we have all the business logic directly in the DisplayDriver while we should have that baked in in the UserManager which could be extended? I need to check that option.

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
```diff
@@ -21,22 +24,32 @@ public class UserDisplayDriver : DisplayDriver<User>
...
             IStringLocalizer<UserDisplayDriver> stringLocalizer)
         {
             _userManager = userManager;
             _roleService = roleService;
             _userRoleStore = userRoleStore;
+            _httpContextAccessor = httpContextAccessor;
+            _notifier = notifier;
             _logger = logger;
             Handlers = handlers;
+
```
hishamco, line 51: Remove extra line

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
```diff
@@ -157,7 +191,16 @@ public override async Task<IDisplayResult> UpdateAsync(User user, UpdateEditorCo
 
                     foreach (var role in rolesToRemove)
                     {
-                        await _userRoleStore.RemoveFromRoleAsync(user, _userManager.NormalizeName(role), default(CancellationToken));
+                        // Make sure we always have at least one administrator account
+                        if (usersOfAdminRole.Count == 1 && usersOfAdminRole.First().UserName == user.UserName && role == "Administrator")
+                        {
+                            _notifier.Warning(H["Cannot remove administrator role from unique administrator."]);
```
hishamco, line 197: Still the "unique" is not make sense to me ;) but I don't mind if the its fine for the others

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
```diff
@@ -79,11 +93,26 @@ public override async Task<IDisplayResult> UpdateAsync(User user, UpdateEditorCo
             model.UserName = model.UserName?.Trim();
             model.Email = model.Email?.Trim();
 
+            var usersOfAdminRole = await _userManager.GetUsersInRoleAsync("Administrator");
```
hishamco, line 96:
```csharp
private static readonly string _administratorRole = "Administrator";
```

hishamco (changes requested):
Looks good after fixing the little tiny things that I mentioned

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
deanmarcussen, line 197:
```suggestion
                            _notifier.Warning(H["Cannot remove administrator role from the only administrator."]);
```

Skrypt:
Anything else guys? Approve the PR or merge it 😉

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
```diff
@@ -79,11 +93,26 @@ public override async Task<IDisplayResult> UpdateAsync(User user, UpdateEditorCo
...
+                {
+                    if (user.UserName != httpContext.User.Identity.Name)
+                    {
+                        user.IsEnabled = model.IsEnabled;
+                        var userContext = new UserContext(user);
+                        await Handlers.InvokeAsync((handler, context) => handler.DisabledAsync(userContext), userContext, _logger);
+                    }
+                    else
+                    {
+                        _notifier.Warning(H["Cannot disable current user."]);
```
hishamco, line 113: Cannot disable current **the** user.

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
```diff
@@ -157,7 +191,16 @@ public override async Task<IDisplayResult> UpdateAsync(User user, UpdateEditorCo
 
                     foreach (var role in rolesToRemove)
                     {
-                        await _userRoleStore.RemoveFromRoleAsync(user, _userManager.NormalizeName(role), default(CancellationToken));
+                        // Make sure we always have at least one administrator account
+                        if (usersOfAdminRole.Count == 1 && usersOfAdminRole.First().UserName == user.UserName && role == _administratorRole)
+                        {
+                            _notifier.Warning(H["Cannot remove administrator role from the only administrator."]);
```
hishamco, line 197: Cannot remove administrator role from the ~~only~~ **super** administrator.

hishamco (changes requested):
Looks good after modifying the messages

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
Skrypt: I don't think it's appropriate in that order. "Cannot disable the current user" would work but you can omit using "the" in this sentence.

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
Skrypt: In fact this message would be misleading as we can disable the first user account created on the backend as long as we log on a second administrator account. So we are not preventing deleting the "super administrator" account if you want to name it that way. If we should do ; then let's discuss about why.

hishamco: > we don't have a super administrator account.

What about the line 101 ;)

Skrypt: Line 101 says. You can't disable the last administrator account remaining. Which could be a different user account than the one created on setup. This doesn't prevent disabled the "super administrator" account because we have never defined one. Do we have one?

hishamco (approved):
LGTM

PR #7144, opened 2020-09-26
Setup : cleanup code

Skrypt:
The ConnnectionString now uses the same javascript method than the password to show/hide it's fa-eye icon. There was a duplication of methods for doing the same thing in the code (because of my last PR).
Also, fixing the icon for the connectionstring that was reversed ; it was showing fa-eye-slash when the password was displayed.


hishamco:
Hope it doesn't conflict with eye icon in Microsoft Edge

Skrypt:
lol, this should be covered by the previous CSS I added right?

hishamco:
I though setup has a different style file ;)

Skrypt:
Yeah, we are never too sure to be sure on anything 🙈 
But yeah you are right, good catch 👍

hishamco (approved):
LGTM

PR #7146, opened 2020-09-26
Missing null check in DatabaseShellsSettingsSources

jtkech:
So that we can register some application level services but only if the related options are provided by the application level configuration that we can retrieve through this new property.

- Maybe better to inject `IConfiguration` in the main startup, as it is done when you create a web app from VStudio, and pass it to `AddOrchardCms()` and `AddOrchardCore()` and then use it when building the `OrchardCoreBuilder`.

    But it would be a breaking change, so here this is the `OrchardCoreBuilder` itself that retrieve the `IConfiguration` from the application services collection, knowing that the related service descriptor `ImplementationFactory(null)` just return the application configuration instance without using the service provider parameter.

    It works with both a web host app, and with a generic host app as we are using, let me know if it is too hacky.

    Note: There is also a builder context that is registered and whose descriptor has an `ImplementationInstance`, and that also has an `IConfiguration` property, but for a generic host app it is an `HostBuilderContext` and for a web host app it is a `WebHostBuilderContext`, so we would have to check both types.

- Here i used it for `AddDatabaseShellsConfiguration()` that here is always called by `AddOrchardCms()`, so that we just have to provide the related options, e.g. in `appsettings.json`, to validate the functionality without changing the code.

- Tried the above on a fresh installation, first nothing is found in the database so the setup is run, then at the end of the setup the shell settings / configs are saved in the database.


src/OrchardCore/OrchardCore.Infrastructure/Shells.Database/Extensions/DatabaseShellsOrchardCoreBuilderExtensions.cs
```diff
@@ -7,14 +8,23 @@ namespace Microsoft.Extensions.DependencyInjection
...
     {
         /// <summary>
-        /// Host services to load shells settings and configuration from database.
+        /// Registers host services to load shells settings and configuration from the database,
+        /// but only if the related options are provided by the application level configuration.
         /// </summary>
         public static OrchardCoreBuilder AddDatabaseShellsConfiguration(this OrchardCoreBuilder builder)
         {
-            var services = builder.ApplicationServices;
+            var options = builder.Configuration
```
deanmarcussen, line 16: What concerns me a little here @jtkech is that it will just silently fail if the configuration is set incorrectly. 

And then you would think you are using it, only to discover that you aren't

src/OrchardCore/OrchardCore.Abstractions/Modules/Builder/OrchardCoreBuilder.cs
```diff
@@ -14,9 +15,14 @@ public class OrchardCoreBuilder
 
         public OrchardCoreBuilder(IServiceCollection services)
         {
+            Configuration = (IConfiguration)services
+                .FirstOrDefault(d => d.ServiceType == typeof(IConfiguration))
+                .ImplementationFactory(null);
```
Jetski5822, line 20: Does FirstOrDefault return Null here?

src/OrchardCore/OrchardCore.Application.Cms.Core.Targets/ServiceExtensions.cs
```diff
@@ -30,7 +30,10 @@ public static OrchardCoreBuilder AddOrchardCms(this IServiceCollection services)
 
                 .AddTheming()
                 .AddLiquidViews()
-                .AddCaching();
+                .AddCaching()
+
+                .AddDatabaseShellsConfiguration()
+                ;
```
Jetski5822, line 36: spacing

src/OrchardCore/OrchardCore.Abstractions/Modules/Builder/OrchardCoreBuilder.cs
jtkech: It should not, both the generic host and web host register a singleton with an `ImplementationFactory` that just returns the `IConfiguration` instance that has been built by the `HostBuilder` or the `WebHostBuilder`, and without using the passed service provider (so here we can pass null).

So here it may fail on an NRE, but only if there is an AspnetCore breaking change

src/OrchardCore/OrchardCore.Application.Cms.Core.Targets/ServiceExtensions.cs
jtkech: The `AddXxx()` things are already separated by groups

Here the new line is just to start a new group

And the `;` in a new line is just to make easier to add / remove an `AddXxx()`

If this is what you meant ;)

jtkech:
@deanmarcussen 

> What concerns me a little here @jtkech is that it will just silently fail if the configuration is set incorrectly.

> And then you would think you are using it, only to discover that you aren't

Yes i thought about it but was not sure, i will remove this part that registers some services only based on the application configuration, at least now we know how to do it if we need it.

At least i will keep this PR because it adds a missing null check

But do you think it is still worth to have and init a `Configuration` property on the `OrchardCoreBuilder`?

deanmarcussen:
> > it will just silently fail if the configuration is set incorrectly
> 
> Hmm, I first thought you meant e.g. a typo in the `OrchardCore_Shells_Database` section name, is it what you meant? Otherwise here we just check that the section exists before registering the services, but then the services will use the related options as before, and `if the configuration is set incorrectly` it will still fail as before.
> 
> Let me know what you think

Yes typo in the connection string would make it fail with an exception, which is helpful.

But also a missing section. 

For example currently I set these through Azure AppSettings - so environment variables.

So if I don't get the devops pipeline to set the section correctly, then I get an exception because it's expecting to find them, and they are empty. 

Others will use keyvault or other configuration tools for them, so it's not just about what we have in our `appsettings.json` files.

> But do you think it is still worth to have and init a Configuration property on the OrchardCoreBuilder?

If we had had this when we wrote `Shells_Database` we might have included a property like
`DisableShells = true` which would potentially have been useful

jtkech:
@deanmarcussen 

Okay i got it and good idea for the `DisableShells = true` option

Hmm, right now for this PR i will just keep the missing null check in `DatabaseShellsSettingsSources`.

Issue #6338, opened 2020-06-04
There is no preview part to add


devhomecali:
Configuring the Preview feature for Blog Posts [this guide](https://docs.orchardcore.net/en/dev/docs/guides/decoupled-cms/), you describe: 

- Edit the content definition of Blog Post, click Add Parts, then select Preview. Click Save.
- In the list of parts, for Preview, click on Edit to change its settings for this content type.
- In the Pattern textbox, enter /blogpost/{{ ContentItem.Content.AliasPart.Alias }} which is the way to generate the same URL as the route which is configured in the Razor page.
 

But there is no **Preview Part** at that list of Content Parts.

How can I add the preview part to PostBlog definition type?

[Screenshot](https://github.com/devhomecali/OrchardCore/blob/master/PreviwPart-NoFound.png)

deanmarcussen:
You probably need to upgrade to the preview package feed as I think we added the `Preview Part` after RC1

See the guide here
https://docs.orchardcore.net/en/dev/docs/getting-started/preview-package-source/

devhomecali:
@deanmarcussen Following your suggestion I installed the preview version: **_OrchardCore.Application.Cms.Tarjets v1.0.0-rc1-13380_**. But although the Preview part can now be added to the definition of the BlogPost, it did not solve the described problem. Doing the settings described in the guide for the preview does not have any effect, the custom url does not look: the standard url of the preview is shown.

Image 1: [Preview is added to BlogPost definition](https://github.com/devhomecali/OrchardCore/blob/master/PreviewPart-OnBlogPostDefinition-1.png)

Image 2: [Pattern for Preview](https://github.com/devhomecali/OrchardCore/blob/master/PreviewPart-OnBlogPostDefinition-2.png)

Image 3: [Normal URL for BlogPost content](https://github.com/devhomecali/OrchardCore/blob/master/PreviewPart-OnBlogPostDefinition-3.png)

Image4: [Preview url is not the slug alias pattern](https://github.com/devhomecali/OrchardCore/blob/master/PreviewPart-OnBlogPostDefinition-4.png)

sebastienros:
Can someone else confirm it's broken?

ns8482e:
@devhomecali `Preview` part allows you to configure `PreviewUrl` for `Preview` button in add/edit screen. 

It doesn’t replace ootb content management UI.

You are still able see `view` OOTB screen.

deanmarcussen:
[...]

PR #7148, opened 2020-09-27
Generate display-for link using url in PreviewPart

ns8482e:
ContentLinkTagHelper - Display-For   generates `View` based on `AdminController` in  `Content`.
However in decoupled scenario, the actual `view` is different and can be configured in `PreviewPart`.

Generating `display-for` link using url configured in PreviewPart.

Fixes: #6338

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentPreview/Handlers/PreviewPartHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentPreview/Handlers/PreviewPartHandler.cs
@@ -2,7 +2,6 @@ using System;
 using System.Linq;
 using System.Threading.Tasks;
 using Fluid;
-using Microsoft.AspNetCore.Routing;
 using OrchardCore.ContentManagement;
 using OrchardCore.ContentManagement.Handlers;
 using OrchardCore.ContentManagement.Metadata;
@@ -43,37 +42,18 @@ namespace OrchardCore.ContentPreview.Handlers
 
             if (!String.IsNullOrEmpty(pattern))
             {
-                string previewUrl = null;
-
-                var model = new PreviewPartViewModel()
+                await context.ForAsync<PreviewAspect>(async previewAspect =>
                 {
-                    PreviewPart = part,
-                    ContentItem = part.ContentItem
-                };
-
-                previewUrl = await _liquidTemplateManager.RenderAsync(pattern, NullEncoder.Default, model,
-                    scope => scope.SetValue("ContentItem", model.ContentItem));
-
-                previewUrl = previewUrl.Replace("\r", String.Empty).Replace("\n", String.Empty);
-
-                await context.ForAsync<PreviewAspect>(previewAspect =>
-                 {
-                     previewAspect.PreviewUrl = previewUrl;
-                     return Task.CompletedTask;
-                 });
-                await context.ForAsync<ContentItemMetadata>(metadata =>
-                {
-                    if (metadata.DisplayRouteValues == null)
+                    var model = new PreviewPartViewModel()
                     {
-                        metadata.DisplayRouteValues = new RouteValueDictionary {
-                            {"PreviewUrl", previewUrl}
-                        };
-                    }
-                    else
-                    {
-                        metadata.DisplayRouteValues.Add("PreviewUrl", previewUrl);
-                    }
-                    return Task.CompletedTask;
+                        PreviewPart = part,
+                        ContentItem = part.ContentItem
+                    };
+
+                    previewAspect.PreviewUrl = await _liquidTemplateManager.RenderAsync(pattern, NullEncoder.Default, model,
+                        scope => scope.SetValue("ContentItem", model.ContentItem));
+
+                    previewAspect.PreviewUrl = previewAspect.PreviewUrl.Replace("\r", String.Empty).Replace("\n", String.Empty);
                 });
             }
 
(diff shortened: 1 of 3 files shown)


src/OrchardCore.Modules/OrchardCore.ContentPreview/Handlers/PreviewPartHandler.cs
```diff
@@ -41,18 +43,37 @@ public override async Task GetContentItemAspectAsync(ContentItemAspectContext co
...
+                await context.ForAsync<PreviewAspect>(previewAspect =>
+                 {
+                     previewAspect.PreviewUrl = previewUrl;
+                     return Task.CompletedTask;
+                 });
+                await context.ForAsync<ContentItemMetadata>(metadata =>
+                {
+                    if (metadata.DisplayRouteValues == null)
+                    {
+                        metadata.DisplayRouteValues = new RouteValueDictionary {
```
deanmarcussen, line 68: This should probably call `contentManager.PopulateAspect<PreviewAspect>` inside this func, rather than the change you made to generate the preview url 

Aspects are intended to be extensible so another handler can jump in and replace the aspect values if it wants to later in the pipeline.

You'll need to resolve the contentmanager lazily from a service provider.

Otherwise seems like a good idea

ns8482e: @deanmarcussen yes, here, the code is using `GetContentItemAspectAsync` of `PreviewPartHandler`.  Here PreviewPartHandler is also adding "PreviewUrl" in route metadata.

`ContentLinkTagHelper` is using route metadata  for links and calling aspect as below.

`metadata = await _contentManager.PopulateAspectAsync<ContentItemMetadata>(DisplayFor);`

deanmarcussen: Yes, but you can't run this code outside the func, or it will run on many requests, rather than just when the aspect is called.

It would be ok if the code was in a private method that was shared between the two aspects, and called inside the func.

This shouldn't add to the route values either, as otherwise it will actually become a route value for other uses of `ContentItemMetadata`.

It would need to be a seperate property on `ContentItemMetadata`, maybe `DisplayUrl` which could then be checked seperate of the route values.

deanmarcussen: Or it shouldn't be a property on `ContentItemMetadata` at all, and just look for a `PreviewAspect` in the `display-for` helper

ns8482e: > It would be ok if the code was in a private method that was shared between the two aspects, and called inside the func.

Understood, will create private method.


> This shouldn't add to the route values either, as otherwise it will actually become a route value for other uses of `ContentItemMetadata`.

I opted for `ContentItemMetadata` as `ContentLinkTagHelper` is using it and I can't use `PreviewAspect` as its not good idea to add reference of `ContentPreview` module in  `OrchardCore.Contents.TagHelpers` for `PreviewAspect` type

ns8482e: @deanmarcussen If I move `PreviewAspect` class from `ContentPreview` module to `ContentManagement.Abstractions`  then I can use `PreviewAspect` in `display-for` tag helper.

src/OrchardCore/OrchardCore.Contents.TagHelpers/ContentLinkTagHelper.cs
```diff
@@ -82,6 +83,27 @@ public override async Task ProcessAsync(TagHelperContext tagHelperContext, TagHe
             if (DisplayFor != null)
             {
                 contentItem = DisplayFor;
+                var previewAspect = await _contentManager.PopulateAspectAsync<PreviewAspect>(contentItem);
+
+                if (!string.IsNullOrEmpty(previewAspect.PreviewUrl))
+                {
+                    var previewUrl = previewAspect.PreviewUrl;
+                    if (!previewUrl.StartsWith("~/", StringComparison.OrdinalIgnoreCase))
```
deanmarcussen, line 91: I think this would be useful as an extension method on `PreviewAspect` - it becomes obvious that it may need virtualizing.

`NormalizePath`

deanmarcussen (commented):
Generally looks good, I'm still not 100% sure that the `PreviewPart` is the right thing to use here, only because it is poorly named, when it is now being used for more than preview.

BUT I think we should go with it  (to hard to rename), and see how it works ouy

ns8482e:
> Generally looks good, I'm still not 100% sure that the `PreviewPart` is the right thing to use here, only because it is poorly named, when it is now being used for more than preview.
> 
> BUT I think we should go with it (to hard to rename), and see how it works out

If you wish, we can rename `PreviewAspect` and it's `PreviewUrl` property used here as It's not just PreviewUrl rather it's actually `DisplayUrl` for decoupled scenarios.

src/OrchardCore/OrchardCore.Contents.TagHelpers/ContentLinkTagHelper.cs
ns8482e: I agree!  Same logic I am also using for form's action in my contents

deanmarcussen: > If you wish, we can rename `PreviewAspect` and it's `PreviewUrl` property used here as It's not just PreviewUrl rather it's actually `DisplayUrl` for decoupled scenarios.

Let's not rename it at this point, but keep that as an idea / possibility. I suspect @sebastienros may have some thoughts Tuesday.

Lets get `NormalizePath` in and merge it for right now

PR #7158, opened 2020-09-29
Ensures shell terminating and disposing

jtkech:
When e.g. we enable a feature for a given shell through the admin, the shell is released in the context of a shell scope, and at the end of this scope (if it is the last one for this shell) the terminate events are called and then the shell is disposed.

But when e.g. we reload a tenant through the admin or when a shell releases its dependents, a shell may be released without having any current scope on it, **so it is disposed immediately but without calling the shell terminate events**.

So here, in that case **we create a new scope** (without using it) and dispose it **to let it manage the shell state as usual**.

Note: In that case the terminate events are called synchronously, we would need a shell `ReleaseAsync()`, but at least there are called. Another suggestion would be to get rid of the terminate events that are not used in the current code.


jtkech:
@sebastienros 

- So, because of the above, the shell terminating / terminated events may not be called asynchronously

- And we call them in the shell scope `DisposeAsync()` after the `BeforeDisposeAsync()` that call the session `CommitAsync()`, so if a terminate event re-uses the session, it will never be re-committed asynchronously

I could fix it but my question is, **can we get rid of the terminate events** that currently are never used in the code?

The code would be much simpler

jtkech:
So, in summary

- Here we ensure that the shell terminating / terminated events are called, even if the shell is not terminated by a last shell scope, e.g when the shell is released by dependency or when changed through the tenants admin UI.

    So, right now i kept the terminate events, but the question remains, could we get rid of them?

- Here we also call the shell terminating / terminated events before the `BeforeDispose` callbacks that we use to call the session commit asynchronously, which is better if a terminate event uses an `ISession`

jtkech:
Here just as a reminder, but i found other things

- When releasing a shell we dispose it if not in use (scope count == 0), otherwise we let the last scope doing it. At some point a given scope was decrementing the count before checking if equal to 0. But to prevent the shell from being disposed by another thread (seeing count == 0) while calling terminate events, we moved the decrement further.

    But this increases the possibility that 2 last scopes both think that it is not the very last one, which ends up by not calling terminate handlers and not disposing the shell

- When a shell is disabled while still in use, it is no more served but it is released by its last scope, so that its execution can complete, and if it get a child scope we don't create a new disabled shell that would have a null service provider

    But when a shell is released by dependency, it doesn't follow this rule

- Finally, a dependent shell is released before its dependency, but because a shell may be disposed at any time by a last scope, this doesn't prevent a dependent shell from being disposed before its dependency.

    Here a shell would need to be also aware of its dependencies, so in a dedicated PR if it is worth to do.

PR #7164, opened 2020-09-29
Update UseCDN  default value

hyzx86:
I think it should be set to false, because in some cases (such as internal networks), the CDN server is not always available.
For example, I am from China. My network problem directly caused me to be unable to access some key js file resources, which also caused me to be unable to click the login button. Even if the login is successful, I cannot expand the settings menu.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Settings/SiteSettings.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Settings/SiteSettings.cs
@@ -24,7 +24,7 @@ namespace OrchardCore.Settings
         public string SiteSalt { get; set; }
         public string PageTitleFormat { get; set; }
         public string SuperUser { get; set; }
-        public bool UseCdn { get; set; } = true;
+        public bool UseCdn { get; set; } = false;
         public string CdnBaseUrl { get; set; }
         public RouteValueDictionary HomeRoute { get; set; } = new RouteValueDictionary();
         public bool AppendVersion { get; set; } = true;


hishamco:
> I cannot expand the settings menu.

This is happen with me too during the low internet connection

src/OrchardCore.Modules/OrchardCore.Settings/SiteSettings.cs
```diff
@@ -24,7 +24,7 @@ public class SiteSettings : Entity, ISite
         public string SiteSalt { get; set; }
         public string PageTitleFormat { get; set; }
         public string SuperUser { get; set; }
-        public bool UseCdn { get; set; } = true;
+        public bool UseCdn { get; set; } = false;
```
hishamco, line 27:
```csharp
public bool UseCdn { get; set; }
```

No need to set `false`

hishamco (changes requested):
Looks good after apply my changes and if others are happy with this default

src/OrchardCore.Modules/OrchardCore.Settings/SiteSettings.cs
hyzx86: Ha, you're right, I was so focused on solving the problem that I forgot the default value

hyzx86:
> > I cannot expand the settings menu.
> 
> This is happen with me too during the low internet connection

But my network is 1GB bandwidth, 囧 。 There may be some difficulties in deploying CDN servers in China

deanmarcussen:
There is another pr which was just waiting for @stantoxt to update here https://github.com/OrchardCMS/OrchardCore/pull/5969

to add a selector for the cdn to the setup page, which would be better than just turning it off by default I think.

hishamco:
I will close this then, thanks @hyzx86

PR #7169, opened 2020-09-30
Admin UI : responsive behaviors

### LINKED ISSUE #5449: Menu : responsive behavior issues
The collapsed menu doesn't provide scrolling

image

Also when we are opening a sub-menu from the bottom of that collapsed menu the items are not accessible. We should open the sub-menu reversed or something like it.

### LINKED ISSUE #7222: Stray div in admin when using blog recipe
When using the SaaS setup recipe, no stray div tag is present. When using the Blog recipe, a stray div tag is present in the output. 

`Error: Stray end tag div.`

There are also other html errors. I think I should list them seperately.

Skrypt:
orchard

orchard

- [x] Use proper semantic for admin menu : https://stackoverflow.com/questions/1141639/how-to-semantically-provide-a-caption-title-or-label-for-a-list-in-html
- [x] Add a scrollbar on the collapsed admin menu (overflow: hidden) but keep displaying the pop-up submenus : https://css-tricks.com/popping-hidden-overflow/
- [x] If a pop-up menu opens up outside of the visible body of the page then make sure we can see all it's elements. https://pawelgrzybek.com/the-intersection-observer-api-explained/
- [x] Removing the usage of perfect-scrollbar on the admin menu in favor of styling the scrollbar with CSS now that it can be done.
- [x] Fix the media library responsive behavior and layout.
- [x] Fix the "Admin Menu feature" views align back to top.
- [x] Fix the "Edit password" view layout labels wrapping too quick by using properly `w-{breakpoint}-{size}` mixins.
- [x] Fix GraphiQL and Media Gallery top margin.

Fixes #5449
Fixes #7222


deanmarcussen:
@Skrypt before I forget since you are giving the menu some love.

Any chance you could make it so the selected menu item is opened on refresh or when you click to open in a new window?

Cheers

Skrypt:
It used to work but something broke that feature.

Skrypt:
Maybe I fixed without knowing it but I can't repro in my current working branch which is based on dev.

orchard

deanmarcussen:
> Edit : Maybe we are missing some places?

Yeah you're right. It is working, but not always in all places - that's what caught me. I will find where it's not working, think it's around deployment plans from memory

Skrypt:
Was it not because of something you had done custom on your side? Something like admin menu items from Taxonomies?

deanmarcussen:
> Was it not because of something you had done custom on your side? Something like admin menu items from Taxonomies?

No I'm not running that custom theme, you know what it is, it just gets lost on site restart, so I see it in development all the time when I restart the site, and refresh the page, and I'm like, where's my menu gone ;)

So pretty minor really.

jptissot:
IF you are playing with the Admin Menu, I noticed that there were Html validation errors related to the menu. Maybe we want to tackle those as part of this PR or maybe another. 

Repro: 
1. Run blog setup recipe
2. Go to the Admin interface and view source
3. Go to https://validator.w3.org/nu/#textarea and paste the source

Here are a few of the errors: 
- Error: The element button must not appear as a descendant of an element with the attribute role=button.
- Error: Attribute role not allowed on element label at this point.
- Error: Attribute aria-expanded not allowed on element label at this point.
- Error: Duplicate ID users.
- Error: Duplicate ID menu-Settings.
- Error: Stray end tag div.

Skrypt:
Yeah I've fixed some of them now but I can clearly see that we are not using semantics as they should be.
Example : itemprop should not be used to define if we open a confirm modal on an element for example. This one as been transported from Orchard 1.

jptissot:
@Skrypt Maybe we should create a bunch of seperate issues for some of these errors. I looked at what you are doing here and it's mostly styles. Maybe keep it specific to that. I can create / fix some of these html issues tonight.

Skrypt:
I think I covered most of them now but one that is redundant is with the itemprop.

Skrypt:
Ok time for review and merge.

src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsAdminList-Search.cshtml
```diff
@@ -1,4 +1,5 @@
 @model ContentOptionsViewModel
+@removeTagHelper Microsoft.AspNetCore.Mvc.TagHelpers.InputTagHelper, Microsoft.AspNetCore.Mvc.TagHelpers
```
deanmarcussen, line 2: Why do you need to do this?

src/OrchardCore.Modules/OrchardCore.Media/Assets/scss/_main-layout.scss
```diff
@@ -55,6 +62,8 @@
             height: 100%;
             width: 240px;
             overflow: auto;
+            scrollbar-width: thin;
+            scrollbar-color: #495057 #EAEAEA;
```
deanmarcussen, line 66: Should we do these with bootstrap variables? 
or just variables in the variables file (which are copied from bootstrap from memory)

deanmarcussen (commented):
Looks pretty good to me, just a couple of queries.

src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsAdminList-Search.cshtml
Skrypt: Because when you do asp-for="Model.Something" on an input there is a bug where it will append twice the type="search" attribute. This is because of the InputTagHelper in ASP.NET MVC.

See : https://stackoverflow.com/questions/42544961/asp-net-core-custom-input-tag-helper-rendering-duplicate-checkboxes which is somehow similar to that.

hishamco:
@Skrypt hopefully I will review this tonight ...

src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsAdminList-Search.cshtml
deanmarcussen: maybe I am being dumb.

If you've removed the `InputTagHelper` which TagHelper is rendering the 

`<input asp-for="DisplayText"` in line 14?

src/OrchardCore.Modules/OrchardCore.Media/Assets/scss/_main-layout.scss
Skrypt: More on this. These styles will work with Firefox only. We use now `::-webkit-scrollbar` for Chrome based browser. It won't work with Internet Explorer but it will still render an unstyled scrollbar. I don't think we should matter about it as now the new Edge (Chromium) is out.

Skrypt: I will add that the perfect-scrollbar was buggy. It was not retaining the proper height of elements all the time because we use absolute and relative positioning in that menu. I had no choice to remove it so that it works correctly.

jptissot (approved):

jptissot:
Tested with Firefox.

deanmarcussen (changes requested):
Small block

The removeTagHelper is the wrong fix.

The issue is with the `InputIsDisabledTagHelper` running twice on that code.

We need to fix that.

Probably just needs `base.Process(context, output);` removed

Skrypt:
Good catch, I would not have thought about looking at these. I tested by moving the `base.Process(context, output);` inside the `if` statement and it also worked. But removing it also work. So, it is not necessary.

PR #7175, opened 2020-09-30
Fix HtmlFieldSettings not showing up

### LINKED ISSUE #7112: Html field settings not editable
I am running version `1.0.0-rc2-14344`

image

jptissot:
Fixes #7112

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Settings/HtmlFieldSettingsDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Settings/HtmlFieldSettingsDriver.cs
@@ -18,7 +18,7 @@ namespace OrchardCore.ContentFields.Settings
                 model.SanitizeHtml = settings.SanitizeHtml;
                 model.Hint = settings.Hint;
             })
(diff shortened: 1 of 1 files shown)


sebastienros:
This is not the only settings driver that was targeting "Editor" . The change was made by @deanmarcussen in this commit (sanitization):
https://github.com/OrchardCMS/OrchardCore/commit/99cfa8e6c5447dfb7b27c7320586983b11409999#diff-3b391a96d2e285da64e4642f4b98b088

I think we need to understand why the change was done, and also fix it everywhere it's wrong, as there are other places that have this.

jptissot:
I think you need to target Editor for editor based options (TextFieldPredefinedList) but the common options for a field need to be rendered in the Content. I think this was just a typo.

jptissot:
I looked at the other field option drivers and they all looked fine.

deanmarcussen:
Yeah, I couldn't repro @jptissot issue but I think the fix is good.

As @jptissot general settings / common to all editors should be in `Content` and editor modes should be in `Editor`

Change originally made by @agriffard and I think just a typo.

Issue #7154, opened 2020-09-28
Errors in FlowPart.Edit.cshtml


MazurchakVova:
When using Trumbowyg inside FlowPart, it throws an error when adding shortcodes.

There are also errors in the DOM, the id attribute is duplicated (FlowPart_ContentTypes)
[DOM] Found 2 elements with non-unique id #FlowPart_ContentTypes
[DOM] Found 2 elements with non-unique id #FlowPart_Prefixes
input type="hidden" id="FlowPart_ContentTypes" name="FlowPart.ContentTypes" value="Paragraph"
input type="hidden" id="FlowPart_ContentTypes" name="FlowPart.ContentTypes" value="CollapsibleSection"

Use my recipe  [FlowDeploymentPlan.zip](https://github.com/OrchardCMS/OrchardCore/files/5294313/FlowDeploymentPlan.zip) to reproduce the error.

screenshot also add below
![FlowPartError](https://user-images.githubusercontent.com/12045907/94473035-51ae3d00-01d4-11eb-8d84-fd2c69bc9f21.png)

ns8482e:
I can reproduce,  It's due `initializeShortcodesApp` is called more than once

PR #7178, opened 2020-09-30
Initialize shortcodeApp only once and unique id attribute in dom for FlowPart

ns8482e:
Fixed: #7154

Initialize vue app  -  shortcodeApp only once
Removed duplicate id attribute in DOM


deanmarcussen (approved):
Ok I got it. I will add another pr with some other fixes later. But this fixes the main issue.

ns8482e:
> @ns8482e did you actually repro the shortcode error?
> 
> I am unable to, but I see a few other issues with it that I need to work on (maybe related / depends if you repro'd?)

 Yes `shortcodesApp`  is initialized when page loads. However when new content item is inserted via ajax, function `initializeShortcodesApp` is called once again that overrides var `shortcodesApp` that doesn't have `$ref`

PR #7190, opened 2020-10-02
Upgrade sortablejs

agriffard:
https://www.npmjs.com/package/sortablejs/v/1.12.0

vuedraggable seems to work still (Content picker).


agriffard:
Has been deprecated since.

PR #7193, opened 2020-10-02
Bump OpenIddict to 3.0.0-beta4.20502.55

kevinchalet:
OpenIddict 3.0 beta4 was released earlier today: https://kevinchalet.com/2020/10/02/introducing-quartz-net-support-and-new-languages-in-openiddict-3-0-beta4/

### DIFF
--- a/src/OrchardCore.Build/Dependencies.props
+++ b/src/OrchardCore.Build/Dependencies.props
@@ -30,8 +30,8 @@
     <PackageManagement Include="Newtonsoft.Json" Version="12.0.3" />
     <PackageManagement Include="NLog.Web.AspNetCore" Version="4.9.3" />
     <PackageManagement Include="NodaTime" Version="3.0.0" />
-    <PackageManagement Include="OpenIddict.AspNetCore" Version="[3.0.0-beta3.20403.69]" />
-    <PackageManagement Include="OpenIddict.Core" Version="[3.0.0-beta3.20403.69]" />
+    <PackageManagement Include="OpenIddict.AspNetCore" Version="[3.0.0-beta4.20502.55]" />
+    <PackageManagement Include="OpenIddict.Core" Version="[3.0.0-beta4.20502.55]" />
     <PackageManagement Include="OrchardCore.Translations.All" Version="1.0.0-rc2-108003" />
     <PackageManagement Include="Serilog.AspNetCore" Version="3.4.0" />
     <PackageManagement Include="Shortcodes" Version="1.0.0-beta-175913881" />
--- a/src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Stores/OpenIdAuthorizationStore.cs
+++ b/src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Stores/OpenIdAuthorizationStore.cs
@@ -235,6 +235,17 @@ namespace OrchardCore.OpenId.YesSql.Stores
             => throw new NotSupportedException();
 
         /// <inheritdoc/>
+        public virtual ValueTask<DateTimeOffset?> GetCreationDateAsync(TAuthorization authorization, CancellationToken cancellationToken)
+        {
+            if (authorization == null)
+            {
+                throw new ArgumentNullException(nameof(authorization));
+            }
+
+            return new ValueTask<DateTimeOffset?>(authorization.CreationDate);
+        }
+
+        /// <inheritdoc/>
         public virtual ValueTask<string> GetIdAsync(TAuthorization authorization, CancellationToken cancellationToken)
         {
             if (authorization == null)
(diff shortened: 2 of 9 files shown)


src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Stores/OpenIdAuthorizationStore.cs
```diff
@@ -359,12 +370,12 @@ public virtual async ValueTask PruneAsync(CancellationToken cancellationToken)
...
-                    authorization.AuthorizationId.IsNotIn<OpenIdTokenIndex>(
-                        token => token.AuthorizationId,
-                        token => token.Status == OpenIddictConstants.Statuses.Valid &&
-                                 token.ExpirationDate > DateTimeOffset.UtcNow))).Skip(offset).Take(1_000).ListAsync();
+                    authorization => authorization.CreationDate < threshold &&
+                                    (authorization.Status != OpenIddictConstants.Statuses.Valid ||
+                                    (authorization.Type == OpenIddictConstants.AuthorizationTypes.AdHoc &&
+                                     authorization.AuthorizationId.IsNotIn<OpenIdTokenIndex>(
+                                         token => token.AuthorizationId,
+                                         token => token.Id != 0)))).Skip(offset).Take(1_000).ListAsync();
```
kevinchalet, line 378: @sebastienros I'm not a huge fan of this workaround. Do you think it would make sense to have an `IsNotIn` overload that doesn't require a second argument? Or am I using this stuff incorrectly? :confused:

sebastienros: Didn't think about it. Very easy to handle an overload that would not have a constraint.

Code is here:

https://github.com/sebastienros/yessql/blob/2e4d526cff556dd41ec108e3837b59c94bd5c968/src/YesSql.Core/Services/DefaultQuery.cs#L254-L255

 You can file an issue. You can even just return `true` I assume. That should generate `WHERE 1` :/

kevinchalet: > You can file an issue.

Will do!

kevinchalet:
@sebastienros I tried to bump YesSql, but the build fails due to an error in a GraphQL test, that seems to use YesSql. Would you mind having a look when you have a moment? Feel free to merge once you're happy with the changes.

src/OrchardCore.Modules/OrchardCore.Queries/Sql/SqlParser.cs
```diff
@@ -140,7 +140,7 @@ private string EvaluateSelectStatement(ParseTreeNode selectStatement)
 
             if (!String.IsNullOrEmpty(_where))
             {
-                sqlBuilder.WhereAlso(_where);
+                sqlBuilder.WhereAnd(_where);
```
Skrypt, line 143: Could it be this line that makes it fail?

kevinchalet: Possible. Before making this change, I took a look at https://github.com/sebastienros/yessql/pull/272/files#diff-e3bd893e42f046e064d35d69baddd662R197, and it seems `WhereAnd` replaces `WhereAlso`, but I may be wrong.

hishamco:
@kevinchalet you can fix your issue with `IsNotIn` after merging #7199

kevinchalet:
@hishamco thanks, but I'm going to revert the YesSql changes for now, as I need to merge this PR for a blog post I'm writing (that I'd like to release today).

src/OrchardCore.Build/Dependencies.props
```diff
@@ -39,12 +39,12 @@
     <PackageManagement Include="xunit.analyzers" Version="0.10.0" />
     <PackageManagement Include="xunit" Version="2.4.1" />
     <PackageManagement Include="xunit.runner.visualstudio" Version="2.4.1" />
-    <PackageManagement Include="YesSql.Abstractions" Version="2.0.0-beta-1514" />
-    <PackageManagement Include="YesSql.Core" Version="2.0.0-beta-1514" />
-    <PackageManagement Include="YesSql.Provider.MySql" Version="1.0.0-beta-1514" />
-    <PackageManagement Include="YesSql.Provider.PostgreSql" Version="1.0.0-beta-1514" />
-    <PackageManagement Include="YesSql.Provider.SqLite" Version="1.0.0-beta-1514" />
-    <PackageManagement Include="YesSql.Provider.SqlServer" Version="1.0.0-beta-1514" />
+    <PackageManagement Include="YesSql.Abstractions" Version="2.0.0-beta-1541" />
```
hishamco: No need for this change because there are in my PR ;)

src/OrchardCore.Modules/OrchardCore.Queries/Sql/SqlParser.cs
hishamco: >  it seems WhereAnd replaces WhereAlso

That's what Seb tell but the question there are two versions now `WhereAnd` and `WhereOr` which the parser needs to differentiate between them, I will try to check this in my PR

PR #6618, opened 2020-07-09
Align content fields editors

hishamco:
Fixes #6615

Alignment

PR #7195, opened 2020-10-02
Fix sizing of html/markdowneditor settings to be consistent 

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6615

Closes https://github.com/OrchardCMS/OrchardCore/pull/6618

Related also to https://github.com/OrchardCMS/OrchardCore/pull/7175 

Bringing them all together for some consistency.

They all look like this now, without uneven padding, and shapes are consistently ordered.

Screenshot 2020-10-02 at 19 03 49

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Settings/HtmlFieldSettingsDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Settings/HtmlFieldSettingsDriver.cs
@@ -18,7 +18,7 @@ namespace OrchardCore.ContentFields.Settings
                 model.SanitizeHtml = settings.SanitizeHtml;
                 model.Hint = settings.Hint;
             })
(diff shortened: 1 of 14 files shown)


hishamco:
There's already PR opened for #6615 ;)

Skrypt (approved):
LGTM

deanmarcussen:
Yes @hishamco your pr was just missing some, so it was easier to start again.

However I see you managed to get the text field settings which I have missed, so if you want to update your pr after to just include those, to match this, that would be good

hishamco:
No problem @deanmarcussen the important thing is we 're fixing the issues, but what I have seen last time there are some duplicate PR that are solving the same issue, so instead of more efforts we can ask the author for the first one to make the suitable updates, but if there's no respond we can start a new

I will try to review this one if it's need, thanks as usual Dean

PR #7196, opened 2020-10-02
Fix Lucene numHits must be > 0

### LINKED ISSUE #5983: ArgumentException: numHits must be > 0; please use TotalHitCountCollector if you just need the total hit count

Skrypt:
Closes #5983, Closes #5984

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/AdminController.cs
@@ -363,8 +363,12 @@ namespace OrchardCore.Lucene.Controllers
                 {
                     var parameterizedQuery = JObject.Parse(tokenizedContent);
                     var docs = await _queryService.SearchAsync(context, parameterizedQuery);
-                    model.Documents = docs.TopDocs.ScoreDocs.Select(hit => searcher.Doc(hit.Doc)).ToList();
-                    model.Count = docs.Count;
+                    
+                    if(docs.TopDocs != null)
+                    {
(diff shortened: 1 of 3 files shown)


src/OrchardCore.Modules/OrchardCore.Lucene/Controllers/AdminController.cs
```diff
@@ -363,8 +363,12 @@ public async Task<IActionResult> Query(AdminQueryViewModel model)
                 {
                     var parameterizedQuery = JObject.Parse(tokenizedContent);
                     var docs = await _queryService.SearchAsync(context, parameterizedQuery);
-                    model.Documents = docs.TopDocs.ScoreDocs.Select(hit => searcher.Doc(hit.Doc)).ToList();
-                    model.Count = docs.Count;
+                    
+                    if(docs.TopDocs != null)
+                    {
+                        model.Documents = docs.TopDocs.ScoreDocs.Select(hit => searcher.Doc(hit.Doc)).ToList();
```
sebastienros, line 369: Is the collection not null if docs.TopDocs is null?

Skrypt: Don't merge it, I will do a second pass on that code. It looks strange to me needing to do this in the controller.

Skrypt: Yeah I fixed it. I looked also if it is possible to use a different override of context.IndexSearcher.Search() method and pass no int to that method instead. But in the end we will get the same result. I wished that this was handled directly in that method but at the same time it doesn't make sense to pass 0 to this method.

PR #6346, opened 2020-06-04
Coming soon theme fixes

deanmarcussen:
Applies fixes from @ns8482e pr

(I can't push to your pr Niraj, so to get RC2 on the way I do this one)

Fixes https://github.com/OrchardCMS/OrchardCore/issues/6305

I've applied the bootstrap-oc.scss in the other themes, and applied a class to the form wrapper to differentiate between it

PR #7197, opened 2020-10-02
FormPart: Add wrapper for content type

### LINKED ISSUE #7180: FormPart - Add wrapper for content type
`FormPart` is rendered using `Form.Wrapper.cshtml` for all content type containing `FormPart` 

Add wrapper in driver to allow customize form per content type.  

i.e. The custom Form content type  `SignupLead` containing `FormPart` can have shape template `Form-SignupLead.Wrapper.cshtml`.  where developer can provide custom form processing without affecting default implementation for `Form` content type or any other content type containing `FormPart`.

https://github.com/OrchardCMS/OrchardCore/issues/6326#issuecomment-670632804

ns8482e:
Fixes #7180  Add wrapper for content type

**(Updated)** For example, following wrapper template `Form-SignupLead.Wrapper.cshtml` for content type `SignupLead` enables form submitted to workflow via ajax - fetch api.

```cshtml
@using OrchardCore.Mvc.Utilities;


@{
    dynamic formPart = Model.ContentItem.Content.FormPart; 
}

<form action="@formPart.Action" 
      method="@formPart.Method"  enctype="@formPart.encType" 
      class="form-content form-@((Model.ContentItem.ContentType as string).HtmlClassify())"
      onsubmit="ajaxSubmit(event)">
    @if(formPart.EnableAntiForgeryToken.Value as bool? ?? false)
    {
        @Html.AntiForgeryToken()
    }

    <input type="hidden" name="customHidden1" value="value1" />
    <input type="hidden" name="customHidden1" value="value1" />

    <script at="Foot">
        function ajaxSubmit(e){

                e.preventDefault();

                const formJson = [].reduce.call(e.target.elements, function(data, element) {
                    data[element.name] = element.value;
                    return data;
                }, {});
                
                var url = e.target.getAttribute('action');  
                fetch(url, {
                    method: e.target.getAttribute('method'),
                    headers: {
                        'Content-Type': 'application/json'  
                    },
                    body: JSON.stringify(formJson) 
                }).then( function(resp){
                    // process response
                });
                
            }
    </script>
</form>
```

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Forms/Drivers/FormContentDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Forms/Drivers/FormContentDisplayDriver.cs
@@ -13,10 +13,14 @@ namespace OrchardCore.Forms.Drivers
         public override Task<IDisplayResult> DisplayAsync(ContentItem model, BuildDisplayContext context)
         {
             var formItemShape = context.Shape;
-            // If the content item contains FormPart add Form Wrapper.
+            // If the content item contains FormPart add Form Wrapper only in Display type Detail
             var formPart = model.As<FormPart>();
-            if (formPart != null)
+            if (formPart != null && context.DisplayType == "Detail")
             {
+                //Add wrapper for content type
+                formItemShape.Metadata.Wrappers.Add($"Form_Wrapper__{model.ContentType}");
+
+                // Add wrapper for <form> tag
                 formItemShape.Metadata.Wrappers.Add("Form_Wrapper");
             }
 
--- a/src/OrchardCore.Modules/OrchardCore.Forms/Drivers/FormContentDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Forms/Drivers/FormContentDisplayDriver.cs
@@ -17,11 +17,8 @@ namespace OrchardCore.Forms.Drivers
             var formPart = model.As<FormPart>();
             if (formPart != null && context.DisplayType == "Detail")
             {
-                // Add wrapper for content type
+                // Add wrapper for content type if template is not available it will fall back to Form_Wrapper
                 formItemShape.Metadata.Wrappers.Add($"Form_Wrapper__{model.ContentType}");
-
-                // Add wrapper for <form> tag
-                formItemShape.Metadata.Wrappers.Add("Form_Wrapper");
             }
 
             // We don't need to return a shape result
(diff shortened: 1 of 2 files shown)


src/OrchardCore.Modules/OrchardCore.Forms/Drivers/FormContentDisplayDriver.cs
```diff
@@ -13,10 +13,14 @@ public class FormContentDisplayDriver : ContentDisplayDriver
...
         {
             var formItemShape = context.Shape;
-            // If the content item contains FormPart add Form Wrapper.
+            // If the content item contains FormPart add Form Wrapper only in Display type Detail
             var formPart = model.As<FormPart>();
-            if (formPart != null)
+            if (formPart != null && context.DisplayType == "Detail")
             {
+                // Add wrapper for content type
+                formItemShape.Metadata.Wrappers.Add($"Form_Wrapper__{model.ContentType}");
```
deanmarcussen, line 21: Looks like this gets hit twice, and is generating two `<form>` tags as a result

ns8482e: All wrappers are rendered. here `Form_Wrapper` will generate form Tag and `Form_Wrapper__[content type]` will generate inner wrapper that allows to provide additional customization within that form. No need to define form tag again

ns8482e: Try with my example code above

deanmarcussen: ah, but have you tried it without creating the second wrapper template?

Try with the coming soon theme. two form tags...

ns8482e: @deanmarcussen  It's not related to this PR, May be you have left over old code for ComingSoon. You  have already fixed that earlier in PR #6346

deanmarcussen: Flow module fresh install.

Before 1 form tag:

Screenshot 2020-10-04 at 09 42 06

After 2 form tags:

Screenshot 2020-10-04 at 09 43 07

ns8482e: Wow! It seems wrapper fall back to parent shape

ns8482e: fixed

deanmarcussen: Now you got it ;)

Looks good. Why did you remove

`id="form_@Model.ContentItem.ContentItemId"`

I thought that looked useful?

ns8482e: I removed id as I feel that its better to use `FormElementPart` but It needs changes in migration so will create separate PR

deanmarcussen (approved):

PR #7199, opened 2020-10-03
Update YesSql.Core to 2.0.0-beta-1541

hishamco:


agriffard:
Build error: 
```
C:\projects\orchardcore\src\OrchardCore.Modules\OrchardCore.PublishLater\bin\Release\netcoreapp3.1\OrchardCore.PublishLater.Views.dll
Sql\SqlParser.cs(143,28): error CS1061: 'ISqlBuilder' does not contain a definition for 'WhereAlso' and no accessible extension method 'WhereAlso' accepting a first argument of type 'ISqlBuilder' could be found (are you missing a using directive or an assembly reference?) [C:\projects\orchardcore\src\OrchardCore.Modules\OrchardCore.Queries\OrchardCore.Queries.csproj]
```

hishamco:
I just saw the error, but the test is passed on YesSql repo

hishamco:
Seems @sebastienros removed `WhereAlso` in https://github.com/sebastienros/yessql/commit/68972b2325c0594f98491fd40e2c77bd75dd3b59#diff-437fac8b343ae4d5d77f4c992da01af0L37

I will update `OrchardCore.PublishLater` to react to the changes, but this kind of change is a breaking change in YesSql, unless you set the version to `2.0.0`

hishamco:
I just found it had been used in

- `OrchardCore.Indexing`

https://github.com/OrchardCMS/OrchardCore/blob/1879619df4418ff175a59ba70dcb36bab89fcbb5/src/OrchardCore.Modules/OrchardCore.Indexing/IndexingTaskManager.cs#L185-L207

here I think `Where` clause is enough

- `OrchardCore.Quieries`

https://github.com/OrchardCMS/OrchardCore/blob/1879619df4418ff175a59ba70dcb36bab89fcbb5/src/OrchardCore.Modules/OrchardCore.Queries/Sql/SqlParser.cs#L110-L188

@Skrypt I may need your help in this one

sebastienros:
WhereAlso just became WhereAnd.

hishamco:
> WhereAlso just became WhereAnd.

I notice a `WhereOr` too ;)

sebastienros:
Yeah, maybe I made a mistake, there is another AndAlso ... I might fix it at some point.

hishamco:
I got some unit tests errors that I'm trying to know why they are failing. @sebastienros `WhereAnd` and `WhereOr` are different, that's why I asked before, because they will lead to different result

nakamura2000:
Unit test error is cause by yessql new sql table name alias feature.Graphql module uses original table name when generating sql where clause. so sql error occurs in this test case and graphql query fails. I tried to fix this error in #7468.

hishamco:
@nakamura2000 the error that show up occurs on `OrchardCore.Indexing` when I tested last time

sebastienros:
We looked at it this morning and the issue is that dynamically generated SQL needs to know the alias tables. So GraphQL tests don't work. I will try to understand why graphql is not using the predicates to do it instead of relying on its own mix of SQL generation + predicates. And maybe also provide some public helpers to the yessql aliases tables if they are required.

sebastienros:
Fixed in another PR

PR #7202, opened 2020-10-03
locale shortcode

jptissot:
An idea I had for Localization using Shortcodes. This allows users to write localized content in a single field / part that supports shortcodes.


src/OrchardCore.Modules/OrchardCore.ContentLocalization/Startup.cs
```diff
@@ -129,4 +131,18 @@ public override void ConfigureServices(IServiceCollection services)
             services.Replace(ServiceDescriptor.Scoped<IContentItemsQueryProvider, LocalizedContentItemsQueryProvider>());
         }
     }
```
agriffard, line 133: newline

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Shortcodes/LocalizationShortcodeProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public ValueTask<string> EvaluateAsync(string identifier, Arguments arguments, string content, Context context)
+        {
+            if (identifier != ShortCodeIdentifier)
+            {
+                return Null;
+            }
+
+            var language = arguments.NamedOrDefault("lang");
+
```
hishamco, line 21: Remove extra line

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Shortcodes/LocalizationShortcodeProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+            var currentCulture = CultureInfo.CurrentCulture.Name;
+
```
hishamco, line 23: Also here ...

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Shortcodes/LocalizationShortcodeProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+            if(!string.Equals(language, currentCulture, StringComparison.InvariantCultureIgnoreCase))
+            {
+                return Null;
+            }
+            return new ValueTask<string>(content);
```
hishamco, line 28: Add a newline before

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Startup.cs
```diff
@@ -129,4 +131,18 @@ public override void ConfigureServices(IServiceCollection services)
...
+    public class ShortcodesStartup : StartupBase
+    {
+        public override void ConfigureServices(IServiceCollection services)
+        {
+            services.AddShortcode<LocalizationShortcodeProvider>("locale", d =>
+            {
+                d.DefaultValue = "[locale '{language_code}'] [/locale]";
+                d.Hint = "Conditionally render content in the specified language";
+                d.Usage = "[locale 'en'] English Text[/locale][locale 'fr'] French Text[/locale]";
+                d.Categories = new string[] { "Localization" };
```
hishamco, line 144: Can we add it to `Content` category too

hishamco (changes requested):
If all agreed about the principle of introducing such shortcode, which I find myself is useful, we need to do some formatting before merge

hishamco (approved):
LGTM

sebastienros:
I blocked it. I'll let you some time to think about why, it until @deanmarcussen had some time to come in. I am an ass like this :)

sebastienros:
Actually it's not localizing, but I still want to double check with @deanmarcussen. 

Also is it handling sub cultures?

hishamco:
> Also is it handling sub cultures?

Nope, it compares the `lang` with the current culture, we could add a new property `FallBackToParentCulture` or make this happen by default

src/OrchardCore.Modules/OrchardCore.Shortcodes/Views/ShortcodeDescriptor-Button-Actions.SummaryAdmin.cshtml
```diff
@@ -1,2 +1,2 @@
 @model ShapeViewModel<ShortcodeDescriptor>
-<button class="btn btn-primary btn-sm" v-on:click="insertShortcode('@Model.Value.DefaultValue')">@T["Add"]</button>
+<button class="btn btn-primary btn-sm" v-on:click="insertShortcode('@Model.Value.DefaultValue.Replace("'", @"\'")')">@T["Add"]</button>
```
sebastienros, line 2: Use `JavascriptEncoder.Encode` instead of replacing the single quote. It will handle all values.
another better way is to render it in an attribute, and read the attribute in javascript, this way Razor will encode it by itself. We do it in many views already to render urls.

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Shortcodes/LocalizationShortcodeProvider.cs
```diff
@@ -0,0 +1,30 @@
...
+        public ValueTask<string> EvaluateAsync(string identifier, Arguments arguments, string content, Context context)
+        {
+            if (identifier != ShortCodeIdentifier)
+            {
+                return Null;
+            }
+
+            var language = arguments.NamedOrDefault("lang");
+            var currentCulture = CultureInfo.CurrentCulture.Name;
+            if(!string.Equals(language, currentCulture, StringComparison.InvariantCultureIgnoreCase))
```
sebastienros, line 22: I think it should also fallback to parent cultures. So if the page is in en-US a [locale en] will also work

jptissot:
> 
> 
> I blocked it. I'll let you some time to think about why, or until @deanmarcussen has some time to comment. I am an ass like this :)

This is still missing a lot ;) This is similar to what WP was doing a while ago If I remember properly, they used [:en]  [en] [:fr] [fr] tags to save multilingual content in a single entry in the db.  

What do you guys think of adding some wrappers around our editors so they handle Multilingual content ? I've had comments from users that it was a bit hard to handle localization of content, especially when the content's structure is similar in both languages but only some text changes.

agriffard:
I guess he mainly blocked it because of the missing use of JavascriptEncoder.
Please do the suggested change.

deanmarcussen:
I'm not sure why @sebastienros blocked this, possibly something about encoding output (i.e. localized output). But I think that would be handled ok by the way we do html?

In terms of shortcodes we did decide at the start of it all not to provide too many oob, or if we did to provide them as shortcode templates via recipes, rather than in code, so it would be easier for users to modify behaviour.

This one would be harder to do as a shortcode template, as we don't have a recipe dedicated to localization.

But it would be important that anything done in the shortcode could be done in a liquid shortcode template, so it is overridable.

We did media in code, for the same reason.

sebastienros:
My concern is about encoding. At first I thought  that it was actually hooking up in the localizer. What is the rule with shortcodes? Shod it always have sanitized html, or will it be sanitized automatically. Or does the part using it sanitize them?

jptissot:
Thanks @deanmarcussen  I will look into it :)

jptissot:
Still missing docs.  Let me know if you think of a better name for the shortcode. I'm not a huge fan of `locale`.

src/OrchardCore.Modules/OrchardCore.Shortcodes/Views/ShortcodeDescriptor-Button-Actions.SummaryAdmin.cshtml
jptissot: Done

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Shortcodes/LocalizationShortcodeProvider.cs
jptissot: Done

test/OrchardCore.Tests/Html/HtmlSanitizerTests.cs
```diff
@@ -14,6 +14,7 @@ public class HtmlSanitizerTests
         [InlineData("<script>alert('xss')</script><div onload=\"alert('xss')\">Test<img src=\"test.gif\" style=\"background-image: url(javascript:alert('xss')); margin: 10px\"></div>", "<div>Test<img src=\"test.gif\" style=\"margin: 10px\"></div>")]
         [InlineData("<IMG SRC=javascript:alert(\"XSS\")>", @"<img>")]
         [InlineData("<a href=\"javascript: alert('xss')\">Click me</a>", @"<a>Click me</a>")]
+        [InlineData("<a href=\"[locale 'en']javascript: alert('xss')[/locale]\">Click me</a>", @"<a>Click me</a>")]
```
deanmarcussen, line 17: Interesting so it strips everything. cool

jptissot: Now I'm wondering if it strips the shortcode.. will it work ?

test/OrchardCore.Tests/Modules/OrchardCore.ContentLocalization/LocalizeShortcodeTests.cs
```diff
@@ -0,0 +1,49 @@
...
+        [InlineData("fr", "foo [locale 'en']bar[/locale][locale 'fr']far[/locale] baz", @"foo far baz")]
+        public async Task ShouldProcess(string currentCulture, string text, string expected)
+        {
+            var sanitizerOptions = new HtmlSanitizerOptions();
+            sanitizerOptions.Configure.Add(opt => opt.AllowedAttributes.Add("class"));
+
+            CultureInfo.CurrentUICulture = new CultureInfo(currentCulture);
+            CultureInfo.CurrentCulture = new CultureInfo(currentCulture);
+
+            var sanitizer = new HtmlSanitizerService(Options.Create(sanitizerOptions));
```
jptissot, line 38: Don't need the sanitizer here.

jptissot:
Anything else to be done for this ?

jptissot:
Should be ok to merge now.

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Shortcodes/LocalizationShortcodeProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+            else
+            {
+                if (currentCulture.Name.Equals(language, StringComparison.CurrentCultureIgnoreCase))
+                {
+                    return new ValueTask<string>(content);
+                }
+            }
+
+            return Null;
+
```
jptissot, line 49: Extra lines

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Shortcodes/LocalizationShortcodeProvider.cs
```diff
@@ -0,0 +1,53 @@
+using System;
+using System.Globalization;
+using System.Threading.Tasks;
+using Shortcodes;
+
+namespace OrchardCore.ContentLocalization.Shortcodes
+{
+    public class LocalizationShortcodeProvider : IShortcodeProvider
```
jptissot, line 8:
```suggestion
    public class LocaleShortcodeProvider : IShortcodeProvider
```

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Startup.cs
```diff
@@ -112,4 +114,19 @@ public override void ConfigureServices(IServiceCollection services)
             services.Replace(ServiceDescriptor.Scoped<IContentItemsQueryProvider, LocalizedContentItemsQueryProvider>());
         }
     }
+
+    [RequireFeatures("OrchardCore.Shortcodes")]
+    public class ShortcodesStartup : StartupBase
+    {
+        public override void ConfigureServices(IServiceCollection services)
+        {
+            services.AddShortcode<LocalizationShortcodeProvider>("locale", d =>
```
jptissot, line 123:
```suggestion
            services.AddShortcode<LocaleShortcodeProvider>("locale", d =>
```

test/OrchardCore.Tests/Modules/OrchardCore.ContentLocalization/LocalizeShortcodeTests.cs
```diff
@@ -0,0 +1,46 @@
...
+using OrchardCore.Media.Events;
+using OrchardCore.Media.Shortcodes;
+using OrchardCore.ResourceManagement;
+using OrchardCore.Shortcodes.Services;
+using Shortcodes;
+using Xunit;
+
+namespace OrchardCore.Tests.Modules.OrchardCore.ContentLocalization
+{
+    public class LocalizeShortcodeTests
```
jptissot, line 22:
```suggestion
    public class LocaleShortcodeTests
```

test/OrchardCore.Tests/Modules/OrchardCore.ContentLocalization/LocalizeShortcodeTests.cs
```diff
@@ -0,0 +1,46 @@
...
+        [InlineData("en-CA", "foo [locale 'en']bar[/locale][locale 'fr']far[/locale] baz", @"foo bar baz")]
+        [InlineData("en-CA", "foo [locale en false]bar[/locale][locale fr]far[/locale] baz", @"foo  baz")]
+        [InlineData("fr", "foo [locale 'en']bar[/locale][locale 'fr']far[/locale] baz", @"foo far baz")]
+        [InlineData("fr", "foo [locale en]bar[/locale][locale fr]far[/locale] baz", @"foo far baz")]
+        public async Task ShouldProcess(string currentCulture, string text, string expected)
+        {
+            CultureInfo.CurrentUICulture = new CultureInfo(currentCulture);
+            CultureInfo.CurrentCulture = new CultureInfo(currentCulture);
+
+            var localizationProvider = new LocalizationShortcodeProvider();
```
jptissot, line 37:
```suggestion
            var localeProvider= new localeShortcodeProvider();
```

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Shortcodes/LocaleShortcodeProvider.cs
```diff
@@ -0,0 +1,51 @@
+using System;
+using System.Globalization;
+using System.Threading.Tasks;
+using Shortcodes;
+
+namespace OrchardCore.ContentLocalization.Shortcodes
+{
+    public class LocaleShortcodeProvider : IShortcodeProvider
```
sebastienros, line 8: Why is it in `ContntLocalization`? I don't see the link with it.

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Shortcodes/LocaleShortcodeProvider.cs
```diff
@@ -0,0 +1,51 @@
...
+                        return new ValueTask<string>(content);
+                    }
+
+                    currentCulture = currentCulture.Parent;
+                }
+                while (currentCulture != CultureInfo.InvariantCulture);
+            }
+            else
+            {
+                if (currentCulture.Name.Equals(language, StringComparison.CurrentCultureIgnoreCase))
```
sebastienros, line 42: `StringComparison.InvariantCultureIgnoreCase`

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Shortcodes/LocaleShortcodeProvider.cs
```diff
@@ -0,0 +1,51 @@
...
+            // default value of true for the fallback
+            var fallback = argFallback == null ? true : Convert.ToBoolean(argFallback); ;
+            var currentCulture = CultureInfo.CurrentCulture;
+
+            if (fallback)
+            {
+                // fallback to parent culture, if the current culture is en-CA and the shortcode targets en, the html will be output
+                do
+                {
+                    if (currentCulture.Name.Equals(language, StringComparison.CurrentCultureIgnoreCase))
```
sebastienros, line 31: Same, don't use `CurrentCultureIgnoreCase`

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Shortcodes/LocaleShortcodeProvider.cs
```diff
@@ -0,0 +1,51 @@
...
+            if (identifier != ShortCodeIdentifier)
+            {
+                return Null;
+            }
+
+            var language = arguments.NamedOrDefault("lang")?.ToLower();
+            var argFallback = arguments.NamedOrAt("fallback", 1);
+            // default value of true for the fallback
+            var fallback = argFallback == null ? true : Convert.ToBoolean(argFallback); ;
+            var currentCulture = CultureInfo.CurrentCulture;
```
sebastienros, line 24: `CurrentCulture` or `CurrentUICulture`. I know we set both but I think for views we need to use the later.

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Shortcodes/LocaleShortcodeProvider.cs
jptissot: I was not sure where to put this shortcode. Do you have any suggestions?

hishamco: I think it's fine to be placed in `ContentLocalization` while the shortcode is related localize content

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Shortcodes/LocaleShortcodeProvider.cs
```diff
@@ -0,0 +1,51 @@
+using System;
+using System.Globalization;
+using System.Threading.Tasks;
+using Shortcodes;
+
+namespace OrchardCore.ContentLocalization.Shortcodes
+{
+    public class LocaleShortcodeProvider : IShortcodeProvider
+    {
+        public const string ShortCodeIdentifier = "locale";
```
hishamco, line 10: Add newline after

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Shortcodes/LocaleShortcodeProvider.cs
hishamco: Use `CurrentUICulture` as Seb mentioned

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Shortcodes/LocaleShortcodeProvider.cs
```diff
@@ -0,0 +1,51 @@
...
+                do
+                {
+                    if (currentCulture.Name.Equals(language, StringComparison.CurrentCultureIgnoreCase))
+                    {
+                        return new ValueTask<string>(content);
+                    }
+
+                    currentCulture = currentCulture.Parent;
+                }
+                while (currentCulture != CultureInfo.InvariantCulture);
```
hishamco, line 38:
```
while (currentCulture != currentCulture.Parent);
```

hishamco (changes requested):

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Shortcodes/LocaleShortcodeProvider.cs
sebastienros: It has nothing to do with Content, I don't understand where you get this idea from ;) It is related to current culture, and "text". Could either be in a Localization module or directly in the Shortcodes one.

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Startup.cs
```diff
@@ -112,4 +114,19 @@ public override void ConfigureServices(IServiceCollection services)
...
+    public class ShortcodesStartup : StartupBase
+    {
+        public override void ConfigureServices(IServiceCollection services)
+        {
+            services.AddShortcode<LocaleShortcodeProvider>("locale", d =>
+            {
+                d.DefaultValue = "[locale {language_code}] [/locale]";
+                d.Hint = "Conditionally render content in the specified language";
+                d.Usage = "[locale en]English Text[/locale][locale fr]French Text[/locale]";
+                d.Categories = new string[] { "Content", "Localization" };
```
sebastienros, line 128: Not related to content

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Shortcodes/LocaleShortcodeProvider.cs
hishamco: The "text" that need to be localized is the content itself

> Could either be in a Localization module 

I don't think so, the localization care about UI, dates, number, currencies .. etc, I think what @jptissot is suggest right, unless you see something I didn't aware of ;)

sebastienros: Then shortcodes module. I don't even think it requires a reference to Localization.

hishamco: If this will go to `Shortcodes` no need to localization at all. At the beginning I thought it will localize the content, but after I dig into the shortcode and usage seems it doesn't make sense to be in content localization or localization module at all. You are right Seb ;) I revoke my previous words ;)

hishamco: @jptissot I'm with Seb, please move it to Shorcodes modules and remove all the localization references, while its not needed at all

deanmarcussen: If it goes into shortcodes it’s fine.
But it needs RequiresFeature on its startup, possibly on localization.

It shouldn’t show up in the popup on single culture sites

jptissot: Ok for shortcode feature.

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Shortcodes/LocaleShortcodeProvider.cs
jptissot: That will not work. It works as is now :)

jptissot:
I don't understand why the build is failing.

jtkech:
It doesn't find the namespace `OrchardCore.Shortcodes.Providers` defined in

    src/OrchardCore.Modules/OrchardCore.ShortCodes/Providers/LocaleShortcodeProvider.cs

It seems that locally you created an `OrchardCore.ShortCodes` folder, try to rename it `OrchardCore.Shortcodes`

hishamco:
There is no folder named `ShortCode` I am checking this from my end

jtkech:
See the first one in the files changed of this PR

This is the only one whose path contains `.ShortCodes/` not a problem under windows but under unix os ...

hishamco:
Ya there was a folder but seems it's removed coz I didn't see it here. I knew that Linux is sensitive but let me find the folder first ;)

jtkech:
@hishamco wait 5 minutes, let me try to fix it under the github UI

jtkech:
Boom ;)

hishamco:
I expect you will recreate that file ;) file rename is problematic in git especially if you want to change the casing

hishamco (approved):
LGTM

jptissot:
> Boom ;)

Thank you !! I don't remember changing the casing on that one. Oh well you got it !! 👍🏼

PR #7203, opened 2020-10-03
React to change DisplayName for TitlePart

### LINKED ISSUE #6129: Changing DisplayName for TitlePart doesn't change label in edit.
I have a requirement to rename Title to different name.  e.g. "Subject."

I changed DisplayName in TitlePart to caption.

image

But when Add Part / Edit, It still shows `Title`

image

image

image

hishamco:
Fixes #6129

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Title/Views/TitlePart.Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Title/Views/TitlePart.Edit.cshtml
@@ -1,16 +1,21 @@
 @model TitlePartViewModel
+@using OrchardCore.ContentManagement.Metadata
+@using OrchardCore.ContentManagement.Metadata.Settings
 @using OrchardCore.ContentLocalization
 @using OrchardCore.Localization
 @using OrchardCore.Title.Models
 @using OrchardCore.Title.ViewModels;
 
+@inject IContentDefinitionManager ContentDefinitionManager
+
 @{
     var culture = await Orchard.GetContentCultureAsync(Model.TitlePart.ContentItem);
+    var displayName = ContentDefinitionManager.GetPartDefinition("TitlePart").DisplayName();
 }
 @if (Model.Settings?.Options != TitlePartOptions.GeneratedHidden)
 {
     <div class="form-group" asp-validation-class-for="Title">
-        <label asp-for="Title">@T["Title"]</label>
+        <label asp-for="Title">@displayName</label>
         <input asp-for="Title" class="form-control content-preview-text content-caption-text" disabled="@(Model.Settings?.Options == TitlePartOptions.GeneratedDisabled)" autofocus="autofocus" dir="@culture.GetLanguageDirection()" />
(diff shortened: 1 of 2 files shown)


hishamco:
FYI `TitlePart` are using a static value `Title` I presume this bug should happen for other parts, I will fix the `TitlePart` one, the I will fix the remaining once in another PR

src/OrchardCore.Modules/OrchardCore.Title/Views/TitlePart.Edit.cshtml
```diff
@@ -1,16 +1,21 @@
...
+
 @{
     var culture = await Orchard.GetContentCultureAsync(Model.TitlePart.ContentItem);
+    var displayName = ContentDefinitionManager.GetPartDefinition("TitlePart").DisplayName();
 }
 @if (Model.Settings?.Options != TitlePartOptions.GeneratedHidden)
 {
     <div class="form-group" asp-validation-class-for="Title">
-        <label asp-for="Title">@T["Title"]</label>
+        <label asp-for="Title">@displayName</label>
```
hishamco, line 18: This fix the issue, but it will break the localization in all the places where `title` is there, @sebastienros shall I fix the issue while it's make sense to react to rename title part or shall we defer this until we support data localization?

jptissot: I don't see any issue with this not being localized. I might be wrong, but isn't this one of the use-cases for dynamic localization ?

src/OrchardCore.Modules/OrchardCore.Title/Views/TitlePart.Edit.cshtml
```diff
@@ -1,16 +1,21 @@
...
         <input asp-for="Title" class="form-control content-preview-text content-caption-text" disabled="@(Model.Settings?.Options == TitlePartOptions.GeneratedDisabled)" autofocus="autofocus" dir="@culture.GetLanguageDirection()" />
         <span asp-validation-for="Title"></span>
         @if (String.IsNullOrWhiteSpace(Model.Settings?.Pattern) && Model.Settings?.Options == TitlePartOptions.Editable)
```
jptissot, line 21: Maybe another issue, but the hint below should, in my opinion, be customizable. We would often avoid using the TitlePart in our content types because we could not name it / modify the hint.

src/OrchardCore.Modules/OrchardCore.Title/Views/TitlePart.Edit.cshtml
hishamco: No @jptissot I don't see issue with this either ;) but there 're many places where `title` is there

src/OrchardCore.Modules/OrchardCore.Title/Views/TitlePart.Edit.cshtml
hishamco: I think someone else already started another PR to address the hint edit

Issue #4326, opened 2019-09-18
Allow to define alt text in MediaPickerField


jptissot:
We need to be able to define localizable alt text for Media. I thought the easiest place would be in the MediaPickerField. Thoughts ?

hishamco:
If there's ability to enter the `alt` text from the admin, the current implementation will unable to localize it, but when we add the dynamic string localization, this will be a good enhancement

jptissot:
Best would be to allow alt text to be defined with the media files and be localizable. Maybe it already is possible. If not, I am ok with the MediaPickerField holding the alt text. 
/cc @matiasmolleja

deanmarcussen:
@jptissot The way I've normally handled that, and I believe is the design intention, is that if we want metadata (alt tags) on a MediaField, we should wrap that field into an ImagePart, and place metadata on that, i.e. Caption on the Agency sample.

Then it's localizable right? 
but does that mean that you have to pick the image again?

jptissot:
I guess that could work. Create a ReusablePart called ImagePart and add a MediaField to it + a Alt Text field. 

I just thought the base MediaPickerField should have at least the alt text since it's pretty common and nessesary for accessibility

Skrypt:
I think the goal in OC is to not bind any metadata by default to the medias. It's easier to migrate medias that way. In 01 in order to be able to migrate medias you needed to have database records migrated for them to work. The goal in RC was to make media more independent of it's metadata.

What @deanmarcussen says is what needs to be documented for this specific usage.

jptissot:
What about my idea of adding a Textbox to the MediaPickerField for alt text? The metadata would be independent from the media but be modifyable from the ContentItem edit page ?

Skrypt:
Yeah, you should check with @matiasmolleja for this. I think we need to come up in this case with a procedure to add a Textbox to the MediaPickerField and a documentation on how to reproduce it. Because your TextField could be named "Apple" or "Oranges" ... and the Vue.js ViewComponent in that case needs to have some constants to look at to add this.

jptissot:
I was thinking something along the lines of what we do for the LinkField. It has 2 textbox (url and text) The MediaPickerField would have the Media picker and another Textbox for alt text. I dont think we need to modify the Vue component.

Skrypt:
Hmm ask @sebastienros what he thinks about it first.

matiasmolleja:
Something like this has been discussed before:
#908  #3404
[...]

Issue #6180, opened 2020-05-13
Image Editor


dodyg:
Does this functionality exist in Orchard Core? I found this blog post on Orchard 1.7 (https://www.davidhayden.me/blog/media-manager-and-image-editor-in-orchard-cms-1.7)

hishamco:
This is a good enhancement for Media module, which is suited after `1.0.0` release

dodyg:
I think this is the code in Orchard

https://github.com/OrchardCMS/Orchard/blob/dev/src/Orchard.Web/Modules/Orchard.MediaProcessing/Controllers/AdminController.cs

sebastienros:
Migrate it !
There might be better libraries than Caiman nowadays. But please keep it client-side only.

dodyg:
CamanJS is no longer maintained (https://github.com/meltingice/CamanJS). I will take a look for alternatives.

dodyg:
https://github.com/fabricjs/fabric.js

dodyg:
I am truly lost trying to solve this problem. Bleh.

dodyg:
Related discussion https://github.com/OrchardCMS/OrchardCore/issues/2784

PR #7209, opened 2020-10-05
Media Anchor Picker and Media Text Editor

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/4326

Allows extra meta data to be added to the media field(s) for alt text, and introduces a crop anchor picker where a crop center location / anchor can be selected and saved with the media field(s).

Is optional, and configurable through settings if alt-text is allowed, or crop picking.

Uses ImageSharp for the crop / pad anchoring, and stores data in the media field element (not hard typed, so optional).

TODO 
- [x] Merge https://github.com/OrchardCMS/OrchardCore/pull/7119
- [x] Add encryption option to prevent disc filling
- [x] ~~Add precision option~~ removed in favour of tokens
- [x] max-height the modal image
- [x] `MediaText` property hard type on the `MediaField`

@sebastienros I changed my mind ;)


deanmarcussen:
Related to The media profiles feature https://github.com/OrchardCMS/OrchardCore/pull/7119

but not reliant on it.

Marking both not ready as we should see if we want to integrate more with profiles, i.e. crops per profile. 

I think probably not, it might be over complicated (for users) to do that, but opinions are welcome...

agriffard:
Can it be related to #6180 ?

deanmarcussen:
> Can it be related to #6180 ?

Kinda of, but it isn't an image editor, I think we will still want an image editor, but that would be more for the media app itself, rather than the media field.

deanmarcussen:
What about calling it `MediaText` @Skrypt ?

Because `Metadata` people might think it had image size and other meta information in it?

Skrypt:
I'd try to make a distinction between Title and Alt. Here it's a Media Title. MediaText can work but it's less descriptive. Though, no big difference in the end.

agriffard:
What do you mean by 'Merge #7209' in the todos? This is the current one.

deanmarcussen:
> What do you mean by 'Merge #7209' in the todos? This is the current one.

Oops, meant https://github.com/OrchardCMS/OrchardCore/pull/7119 (updated)

Blocked on doing anything further as I'm to lazy to deal with the merge conflicts

deanmarcussen:
> I'd try to make a distinction between Title and Alt. Here it's a Media Title. MediaText can work but it's less descriptive. Though, no big difference in the end.

So for me, unless I'm misunderstanding something it could be used for Title / Alt / anything (you could even throw some json in it to give more data, alt, and title).

So it seems like `DisplayText` - which can mean `Title`, but doesn't always (Widgets for example).
Hence I'm thinking `MediaText`, as being generic enough that it could be used for anything.

I will make it a property on the field (`MediaTexts` - because it's an array), rather than an extension...

scleaver:
@deanmarcussen could this be used to store other things? Is it extensible? So currently I am using the media picker to select an mp3 file that is stored on Azure. I need to store the file size and actually the duration too automatically... I don't want to store this on save of the content item rather than the more costly exercise of getting this info from the file when displaying the content on the frontend.

I had actually added two properties to my content type to store the size and but it would seem better to do something like what you are doing.

deanmarcussen:
@scleaver this could be a little extensible, in that you could choose to store, for example, a comma delimited string with multiple values. 

But it would still be editable, and may not be the best user experience.

I suspect adding the two properties and managing it as a content type / part might be a better user experience.

Equally all the fields, including this one, support storing other values alongside them. 

Here there is also a `Centers` array which is not hardcoded on the `MediaField` but accessed and set through some extension methods.

So for what you are doing, if you want to do it automatically, you might implement a second mediafield driver (or a content type handler) which could check the size, and duration, if it is an mp3 file, during save, and update a hidden property on the `MediaField` which would be accessible in your templates, but not seen on the ui.

deanmarcussen:
Done - I think :)

`AltText` renamed to `MediaText`
Crop Centers renamed to `Anchor` as calling them `Center` was not quite clear enough. (it's not a center, when you offset a different part of the image, at that point it's a crop center, or pad center.)

But Anchor seemed to make sense as being specifically an Anchor point, whichever resizing technique is applied.

Tokens remain `token=` because I think they are actually a token, because they do contain the actual values of the resizing commands in the encrypted string. So it's not just obscufation, they do contain a tokenized value.

Screenshot 2020-11-01 at 12 52 44

Screenshot 2020-11-01 at 12 52 55

Issue #7215, opened 2020-10-06
System.InvalidCastException: Unable to cast object of type 'OrchardCore.ContentManagement.ContentHandleManager' to type 'OrchardCore.ContentManagement.IContentAliasManager'.


rjpowers10:
```
System.InvalidCastException: Unable to cast object of type 'OrchardCore.ContentManagement.ContentHandleManager' to type 'OrchardCore.ContentManagement.IContentAliasManager'.
   at OrchardCore.ContentManagement.ServiceCollectionExtensions.<>c.<AddContentManagement>b__0_0(IServiceProvider sp) in C:\projects\orchardcore\src\OrchardCore\OrchardCore.ContentManagement\ServiceCollectionExtensions.cs:line 32
   at ResolveService(ILEmitResolverBuilderRuntimeContext , ServiceProviderEngineScope )
   at ResolveService(ILEmitResolverBuilderRuntimeContext , ServiceProviderEngineScope )
   at Microsoft.Extensions.DependencyInjection.ActivatorUtilities.GetService(IServiceProvider sp, Type type, Type requiredBy, Boolean isDefaultParameterRequired)
   at lambda_method(Closure , IServiceProvider , Object[] )
   at Microsoft.AspNetCore.Mvc.Controllers.ControllerFactoryProvider.<>c__DisplayClass5_0.<CreateControllerFactory>g__CreateController|0(ControllerContext controllerContext)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.Next(State& next, Scope& scope, Object& state, Boolean& isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.InvokeInnerFilterAsync()
```

I'm not sure what the simplest way to reproduce this is, but I think I see the problem in [here](https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore/OrchardCore.ContentManagement/ServiceCollectionExtensions.cs).

The code is trying to cast an `IContentHandleManager` as an `IContentAliasManager`. But looking at [this](https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore/OrchardCore.ContentManagement.Abstractions/IContentHandleManager.cs), it looks to be the other way around. 

/cc @deanmarcussen

PR #7218, opened 2020-10-07
Obsolete IContentAliasManager

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/7215

@rjpowers10 we renamed this interface, and the backwards compatibility code didn't work.

I've just obsoleted it instead as making it totally backwards compatible just made for weird code.

So inject `IContentHandleManager`


rjpowers10:
Thank you for the fix!

PR #7226, opened 2020-10-08
PagerShape anchors "rel" attribute for SEO

Skrypt:
- Set PagerShape anchors rel attribute properly for SEO. Adding next, prev, canonical values to them by default.
- Allows to override all the "rel" attributes with "no-follow" value.
- Remove href attribute of anchors that are disabled.


agriffard:
It is always debatable to introduce a new default behaviour that forces into one direction.
May be some people will want it as it is currently for other reasons.

Here are some questions to narrow a solution: 
- Can this be set from the theme with an alternate if needed?
- Could it be settable with a Global setting, in an SEO module, ...?

deanmarcussen:
I'm not sure
https://www.contentkingapp.com/academy/pagination/

We're going through an seo audit on a site over the next couple of weeks so I'll see what they have to say.

I already know we need to add some canonicals in (on the home page /blog particularly as they shouldn't be seen as different)

Opinions on pager links appear to be divided. but nofollow wasn't something I would have thought to add on a pager link

Skrypt:
Ok, definitely need something to be done on this. On my part, I made those no-follow for a reason. We don't want to index our search result pages just like Google would not want to index it's search result pages. It does make sense to avoid robots to crawl every single page of your pagers most of the time. But let's make this configurable per pager then. Let's add an option that we can pass to the shape to be able to set the "rel" attribute of those links.

deanmarcussen:
> Let's add an option that we can pass to the shape to be able to set the "rel" attribute of those links.

Yup, all good if it's optional. But I think would already work just by adding to the `shape.Attributes`?

Skrypt:
Yeah, not sure about that @deanmarcussen, adding this to the Shape.Attributes will add it on which rendered item in the Pager? We need to define on each Anchor of the pager different "rel" so I need to check.

Skrypt:
Ok quick review on this. We have :  

Pager_First
Pager_Previous
Pager_CurrentPage
Pager_Next
Pager_Last

Which are all :  Pager_Link

Skrypt:
I'm going into an other direction with this. It is always possible to use a `<meta name="robots" content="noindex">` in the page header if you don't want robots to crawl all your pages. So, I'm updating the PagerShape anchors to at least have the proper rel attribute values set. If you don't want all your pages to be indexed then you have other options than setting the Pager links to no-follow.

deanmarcussen (approved):
Cool, so we get next and prev by default now as well, and no-follow is optional. Awesome

PR #7240, opened 2020-10-09
Remove duplicate LangVersion entry

hishamco:


sebastienros:
Where is the other one?

hishamco:
Here is a better look

https://github.com/OrchardCMS/OrchardCore/blob/e93de03087aa9e2919fcc52d3aa38a48e2120647/src/OrchardCore.Build/OrchardCore.Commons.props#L6-L17

agriffard:
Do we keep it explicit or with a version number?

In the 5.0 branch, we upgraded to C# 9.

hishamco:
I think the implicit version will get the latest version without any modifications everytime

deanmarcussen:
I would suggest keeping it explicit. 

If we're doing a 3.1 LTS release we don't want v9 code sneaking in accidentally if people are running betas.

hishamco:
OK Dean

deanmarcussen (approved):

PR #7241, opened 2020-10-09
Update README.md

hishamco:
- Add build history
- Separate packages feed from build status

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,19 +10,20 @@ Orchard Core consists of two distinct projects:
 [![Documentation](https://readthedocs.org/projects/orchardcore/badge/)](https://docs.orchardcore.net/)
 [![Crowdin](https://badges.crowdin.net/orchard-core/localized.svg)](https://crowdin.com/project/orchard-core)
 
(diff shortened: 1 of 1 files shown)


hishamco:
ReadMe

deanmarcussen:
I wish people would open issues about these things so they can be triaged.

So the change is adding some build history?

deanmarcussen:
My concern with this is understanding what happens when builds are cancelled because we merge a few things at once.

If it's listing these cancelled builds, as failed (which it does when looking through github), then it would be showing an inaccurate picture.

hishamco:
Dean frankly I'm not sure if it tracks the cancelled builds, but let me check if the gray bars means the build is cancelled

agriffard:
This is not the kind of information that should appear in the Readme.
Closing it.

hishamco:
> This is not the kind of information that should appear in the Readme.

Why? It is something I see in some repos for more detailed history about the builds, something like build status that we already show in README.md

Issue #7243, opened 2020-10-10
Add ShapeResult - ResultConditionAsync 


deanmarcussen:
@Skrypt has a pr to conditionally add `SummaryAdmin` shapes based on a users permission (using `AuthorizeAsync`)

PR here: https://github.com/OrchardCMS/OrchardCore/pull/7104

The problem is that `SummaryAdmin` shapes are added during `BuildDisplayAsync` so the conditions will be evaluated for every front end `BuildDisplayAsync` display, even though they're only intended to be placed in a `SummaryAdmin` display.

I propose adding a `Func` to the `ShapeResult` that can be invoked after placing a shape in a location, to evaluated whether it should be placed.

This would mean that for all of the `SummaryAdmin` shapes the `ResultConditionAsync` will only run after they have been placed in `SummaryAdmin`.

So the evaluation will not occur during a `Detail` or `Summary` view on the front end.

The `AuthorizeAsync` isn't a super heavy call, but we often want to conditionally evaluate in a driver whether to return a `ShapeResult` so I think it makes sense to have something that can be applied later in the `BuildDisplayAsync` pipeline, when we know that we are actually going to place the shape in a location.

The code changes from (this, plus the 4 other `AuthorizeAsync` evaluations)
```
public override async Task<IDisplayResult> DisplayAsync(ContentItem contentItem, IUpdateModel updater)
{
...
                var hasEditPermission = await _authorizationService.AuthorizeAsync(context.User, CommonPermissions.EditContent, contentItem);
                var hasPublishPermission = await _authorizationService.AuthorizeAsync(context.User, CommonPermissions.PublishContent, contentItem);
                if (hasEditPermission || hasViewPermission)
                {
                    results.Add(Shape("ContentsButtonEdit_SummaryAdmin", new ContentItemViewModel(contentItem)).Location("SummaryAdmin", "Actions:10"));
                }

```

To 
```
public override async Task<IDisplayResult> DisplayAsync(ContentItem contentItem, IUpdateModel updater)
{
...
                results.Add(
                    Shape("ContentsButtonEdit_SummaryAdmin", new ContentItemViewModel(contentItem)).Location("SummaryAdmin", "Actions:10")
                    .ResultConditionAsync(async () => {
                        var hasViewPermission = await _authorizationService.AuthorizeAsync(context.User, CommonPermissions.ViewContent, contentItem);
[...]

PR #7244, opened 2020-10-10
Add ShapeResult - RenderWhen

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/7243

@jtkech I put a bunch of details on the issue. 

Wouldn't mind to get your review on this as well if you have the time?

p.s naming is hard. I renamed this twice, and then ended up calling it `IfAsync` so if you have any better ideas they're welcome...

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement/Views/ShapeResult.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/Views/ShapeResult.cs
@@ -24,7 +24,7 @@ namespace OrchardCore.DisplayManagement.Views
         private Action<CacheContext> _cache;
         private string _groupId;
         private Action<ShapeDisplayContext> _displaying;
-        private Func<Task<bool>> _resultConditionAsync;
+        private Func<Task<bool>> _ifAsync;
 
         public ShapeResult(string shapeType, Func<IBuildShapeContext, ValueTask<IShape>> shapeBuilder)
             : this(shapeType, shapeBuilder, null)
@@ -106,7 +106,7 @@ namespace OrchardCore.DisplayManagement.Views
             }
 
             // If a condition has been applied to this result evaluate it only if the shape has been placed.
-            if (_resultConditionAsync != null && !(await _resultConditionAsync.Invoke()))
+            if (_ifAsync != null && !(await _ifAsync.Invoke()))
             {
                 return;
             }
@@ -297,11 +297,11 @@ namespace OrchardCore.DisplayManagement.Views
 
         /// <summary>
         /// Sets a condition that must return true for the shape to render.
-        /// The condition is only evaluated after placing the shape.
+        /// The condition is only evaluated if the shape has been placed.
         /// </summary>
-        public ShapeResult ResultConditionAsync(Func<Task<bool>> resultConditionAsync)
+        public ShapeResult IfAsync(Func<Task<bool>> ifAsync)
         {
-            _resultConditionAsync = resultConditionAsync;
+            _ifAsync = ifAsync;
             return this;
         }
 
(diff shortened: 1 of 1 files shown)


src/OrchardCore/OrchardCore.DisplayManagement/Views/ShapeResult.cs
```diff
@@ -104,6 +105,12 @@ private async Task ApplyImplementationAsync(BuildShapeContext context, string di
                 return;
             }
 
+            // If a condition has been applied to this result evaluate it only if the shape has been placed.
+            if (_ifAsync != null && !(await _ifAsync.Invoke()))
```
hishamco, line 109: Could we use `!(await _ifAsync?.Invoke())` instead?

src/OrchardCore/OrchardCore.DisplayManagement/Views/ShapeResult.cs
```diff
@@ -288,6 +295,16 @@ public ShapeResult Cache(string cacheId, Action<CacheContext> cache = null)
...
         }
 
+        /// <summary>
+        /// Sets a condition that must return true for the shape to render.
+        /// The condition is only evaluated if the shape has been placed.
+        /// </summary>
+        public ShapeResult IfAsync(Func<Task<bool>> ifAsync)
+        {
+            _ifAsync = ifAsync;
+            return this;
```
hishamco, line 305: Add new line before

src/OrchardCore/OrchardCore.DisplayManagement/Views/ShapeResult.cs
```diff
@@ -288,6 +295,16 @@ public ShapeResult Cache(string cacheId, Action<CacheContext> cache = null)
             return this;
         }
 
+        /// <summary>
+        /// Sets a condition that must return true for the shape to render.
+        /// The condition is only evaluated if the shape has been placed.
+        /// </summary>
+        public ShapeResult IfAsync(Func<Task<bool>> ifAsync)
```
hishamco, line 302: Naming is hard ;)

Skrypt:
I would have used ValidateAsync but it is fine as it is.

src/OrchardCore/OrchardCore.DisplayManagement/Views/ShapeResult.cs
jtkech: That's okay because we would need something like this

    await ( _ifAsync?.Invoke() ?? Task.FromResult(false) )

Here I suggest

    if (_ifAsync != null && !await _ifAsync())

src/OrchardCore/OrchardCore.DisplayManagement/Views/ShapeResult.cs
jtkech: About naming, maybe

    public ShapeResult DisplayWhen(Func<Task<bool>> displayPredicateAsync)
    {
        _displayPredicateAsync = displayPredicateAsync;

    public ShapeResult RenderWhen(Func<Task<bool>> renderPredicateAsync)
    {
        _renderPredicateAsync = renderPredicateAsync;

jtkech (commented):
See my comment about naming

src/OrchardCore/OrchardCore.DisplayManagement/Views/ShapeResult.cs
deanmarcussen: I like `RenderWhen` thanks @jtkech

src/OrchardCore/OrchardCore.DisplayManagement/Views/ShapeResult.cs
deanmarcussen: Don't really see why this is necesary @hishamco ? Have a look at the actual class.

It's not a hard and fast rule and it just wastes space in a builder like class like this.

hishamco: This is the format we use everywhere ;) no worry, the important thing is the name got pretty ;)

Skrypt:
LGTM

PR #7247, opened 2020-10-11
Workflows: jQuery package version

agriffard:
https://www.npmjs.com/package/@types/jquery/v/3.5.2

Do not upgrade to https://www.npmjs.com/package/@types/bootstrap/v/5.0.0 yet, stay on 4.5.0.
It will need a dependency on the new @popperjs/core package, I guess.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Workflows/package.json
+++ b/src/OrchardCore.Modules/OrchardCore.Workflows/package.json
@@ -5,6 +5,6 @@
     "bootstrap": "4.5.2",
     "@types/bootstrap": "4.5.0",
     "jsplumb": "2.14.6",
-    "@types/jquery": "3.5.1"
+    "@types/jquery": "3.5.2"
   }
 }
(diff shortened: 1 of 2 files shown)


src/OrchardCore.Modules/OrchardCore.Workflows/package.json
```diff
@@ -5,6 +5,6 @@
     "bootstrap": "4.5.2",
     "@types/bootstrap": "4.5.0",
     "jsplumb": "2.14.6",
-    "@types/jquery": "3.5.1"
+    "@types/jquery": "3.5.2"
```
deanmarcussen, line 8: Confused. I thought we were still on `jQuery` 3.5.1 ?

Is the types upgrade fixing some issues with the jquery 3.5.1 type definitions?

agriffard: The version of `@types/jquery` is not always tied to the version of jquery.

deanmarcussen (approved):

PR #7250, opened 2020-10-11
Add support for using remote source in MediaStep

lukaskabrt:
Right now `MediaStep` supports a file saved in the same file provider or a base64 encoded file as the source. This PR adds the third option - the file from a remote URL.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Media/Recipes/MediaStep.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Recipes/MediaStep.cs
@@ -107,9 +107,9 @@ namespace OrchardCore.Media.Recipes
 
             /// <summary>
             /// URL where the content is read from. Use when the file
-            /// will be available on the remote website.
-            /// If Base64 property or SourcePath property are set, thet take
-            /// precedence ovar SourceUrl.
+            /// will be available on a remote website.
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.Media/Recipes/MediaStep.cs
```diff
@@ -13,6 +14,7 @@ namespace OrchardCore.Media.Recipes
     public class MediaStep : IRecipeStepHandler
     {
         private readonly IMediaFileStore _mediaFileStore;
+        private readonly static HttpClient _httpClient = new HttpClient();
```
hishamco, line 17: LGTM except if we can get the `HttpClient` from DI, but it's fine

deanmarcussen (approved):

agriffard:
During the @sebastienros suggested to only use the SourcePath and check if it starts by `http`.

lukaskabrt:
@agriffard if that's agreed on I can change it ... `SourceUrl` just seemed like a cleaner solution.

agriffard:
Can you please create an issue and then a PR @lukaskabrt ?
@sebastienros also talked about the resiliency of the httpClient calling external resources.

PR #7261, opened 2020-10-13
Fix RSS format.

### LINKED ISSUE #7260: RSS Feed double encodes CDATA
The `FeedController` double encodes CDATA when it doesn't need too. For example, viewing source nodes using CDATA should look like this:

`<description><![CDATA[<p>My amazing description.</p>]]></description>`

Instead it is double encoded like this:
`<description>&lt;![CDATA[&lt;p&gt;My amazing description.&lt;/p&gt;]]&gt;</description>`

This messes up some feed readers which results in them displaying the description like: "My amazing description.]]"

scleaver:
Fixes #7260 

Not exactly sure how to fix the non-RSS format fallback section of the code as it doesn't use Xelement but uses the AddProperty method of the FeedBuilder interface which only uses strings. Any thoughts would be appreciated.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Feeds/CommonFeedItemBuilder.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Feeds/CommonFeedItemBuilder.cs
@@ -49,7 +49,12 @@ namespace OrchardCore.Contents.Feeds.Builders
                     feedItem.Element.SetElementValue("title", WebUtility.HtmlEncode(contentItem.DisplayText));
                     feedItem.Element.Add(link);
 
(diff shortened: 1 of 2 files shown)


deanmarcussen:
> Not exactly sure how to fix the non-RSS format fallback section of the code as it doesn't use Xelement

Can you use `XCData.ToString()` or does that only work with `XElement.ToString()`

hishamco:
I might try this too, coz I'm still wondering why `bodyAspect.Body` is encoded in `CData` only, I may use Arabic to see how encoding and everything is look like

hishamco:
I blocked this for now, please refer to my comment in the related issue

hishamco:
I just created a PR #7268 that fixed the issue, removes unnecessary checks and use `XCData` in other places

hishamco:
Fixed in #7268, thanks a lot Seth

PR #7268, opened 2020-10-13
Fix CDATA encoding issue

### LINKED ISSUE #7260: RSS Feed double encodes CDATA

hishamco:
Fixes #7260

- Avoid double encode CDATA
- Use XData

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListFeedQuery.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListFeedQuery.cs
@@ -79,7 +79,7 @@ namespace OrchardCore.Lists.Feeds
             else
             {
                 context.Builder.AddProperty(context, null, "title", contentItem.DisplayText);
-                context.Builder.AddProperty(context, null, "description", new XCData(bodyAspect.Body?.ToString() ?? String.Empty).ToString());
+                context.Builder.AddProperty(context, null, new XElement("description", new XCData(bodyAspect.Body?.ToString() ?? String.Empty)));
 
                 context.Response.Contextualize(contextualize =>
                 {
--- a/src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListFeedQuery.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListFeedQuery.cs
@@ -66,7 +66,7 @@ namespace OrchardCore.Lists.Feeds
                 context.Response.Element.SetElementValue("title", WebUtility.HtmlEncode(contentItem.DisplayText));
                 context.Response.Element.Add(link);
 
-                context.Response.Element.SetElementValue("description", bodyAspect.Body != null ? $"<![CDATA[{bodyAspect.Body?.ToString()}]]>" : String.Empty);
+                context.Response.Element.Add(new XElement("description", new XCData(bodyAspect.Body?.ToString() ?? String.Empty)));
 
                 context.Response.Contextualize(contextualize =>
                 {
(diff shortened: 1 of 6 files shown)


hishamco:
I may add a unit test for this one

Skrypt:
Looks good

src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListFeedQuery.cs
```diff
@@ -79,8 +79,7 @@ public async Task ExecuteAsync(FeedContext context)
             else
             {
                 context.Builder.AddProperty(context, null, "title", contentItem.DisplayText);
-
-                context.Builder.AddProperty(context, null, "description", bodyAspect.Body != null ? $"<![CDATA[{bodyAspect.Body?.ToString()}]]>" : String.Empty);
+                context.Builder.AddProperty(context, null, "description", new XCData(bodyAspect.Body?.ToString() ?? String.Empty).ToString());
```
scleaver, line 82: Same issue in the "else"  as for the CommonFeedItemBuilder here too.

scleaver (changes requested):
There are still some issues inside the else when using context.Builder.AddProperty. See comments on the code.

src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListFeedQuery.cs
scleaver: Also no need to out in empty CDATA if the body is empty.

hishamco:
@scleaver let me complete the unit tests and check your comment about `Builder.AddProperty()`

src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListFeedQuery.cs
hishamco: BTW I think there's no issue with `<![CDATA[]]>` I see this quite often

scleaver: > BTW I think there's no issue with `<![CDATA[]]>` I see this quite often

Correct there is no issue... personal preference ;)

Skrypt: It is better to have `<![CDATA[]]>` than not, as it enforce defining which type of data is contained in that parameter even if it's empty. If you use a "" value in that attribute then it means that the returned content could vary of type.

hishamco:
@scleaver you could try the latest changes

test/OrchardCore.Tests/Modules/OrchardCore.Contents/Feeds/RssFeedTests.cs
```diff
@@ -0,0 +1,63 @@
...
+            {
+                DisplayText = "News",
+                PublishedUtc = DateTime.UtcNow
+            });
+
+            // Act
+            await commonFeedItemBuilder.PopulateAsync(feedContext);
+
+            // Assert
+            Assert.Contains($"<![CDATA[{bodyContent}]]>", feedContext.Response.Items[0].Element.ToString());
```
deanmarcussen, line 50: This looks generally good.

But the test does not test the issue. 
Where is the test that checks whether the elements have been html entity escaped or not?

Skrypt (changes requested):
Assert that this : `<description><![CDATA[<p>My amazing description.</p>]]></description>`
Never becomes this : `<description>&lt;![CDATA[&lt;p&gt;My amazing description.&lt;/p&gt;]]&gt;</description>`

After that you can merge.

hishamco:
For sure @Skrypt, I will do it now

test/OrchardCore.Tests/Modules/OrchardCore.Contents/Feeds/RssFeedTests.cs
hishamco: @deanmarcussen I just checked whether the `<![CDATA[]>` becomes `&lt;![CDATA[&lt;]]]&gt;` or not ;) I will do the assert that requested by @Skrypt

PR #7269, opened 2020-10-13
Admin UI Permissions : rename

Skrypt:
Rename weird permission name to ListContent 😄 

The other name we need to change is `IsAuthorizedToAContentTypeAsync`
Please give me some ideas.

```C#
        /// <summary>
        /// Evaluate if we have a specific owner variation permission to at least one content type
        /// </summary>
        public static async Task<bool> IsAuthorizedToAContentTypeAsync(this IAuthorizationService service, ClaimsPrincipal user, Permission requiredPermission, IEnumerable<ContentTypeDefinition> contentTypeDefinitions, IContentManager contentManager)
        {
            foreach(var contentTypeDefinition in contentTypeDefinitions)
            {
                var permission = GetOwnerVariation(requiredPermission);
                var contentTypePermission = ContentTypePermissionsHelper.ConvertToDynamicPermission(permission);
                permission = ContentTypePermissionsHelper.CreateDynamicPermission(contentTypePermission, contentTypeDefinition);
                
                var contentItem = await contentManager.NewAsync(contentTypeDefinition.Name);
                contentItem.Owner = user.Identity.Name;

                if(await service.AuthorizeAsync(user, permission, contentItem))
                {
                    return true;
                }
            }

            return false;
        }
```

I was thinking about reversing it :  `AContentTypeIsAuthorizedToAsync`
But that's as bad 😢


src/OrchardCore.Modules/OrchardCore.Contents/Permissions.cs
```diff
@@ -25,7 +25,7 @@ public class Permissions : IPermissionProvider
         public static readonly Permission CloneContent = CommonPermissions.CloneContent;
         public static readonly Permission CloneOwnContent = CommonPermissions.CloneOwnContent;
         public static readonly Permission AccessContentApi = new Permission("AccessContentApi", "Access content via the api");
-        public static readonly Permission ViewNotOwnedItemsAdminContentList = new Permission("ViewNotOwnedItemsAdminContentList", "View content item(s) not owned by user in admin content list");
+        public static readonly Permission ListContent = new Permission("ListContent", "List content item(s) not owned by user");
```
deanmarcussen, line 28:
```suggestion
        public static readonly Permission ListContent = new Permission("ListContent", "List content item(s) owned by all users");
```

deanmarcussen:
How about `AuthorizeContentTypeDefinitionsAsync`

That's essentially what it does. The comment is and fact it returns a bool is enough to understand that it returns after finding one that has the permission.

PR #7280, opened 2020-10-15
Media Library: improve localization

### LINKED ISSUE #7277: Media Library: improve localization and little SCSS modifications
Hi all, 

I would like to ask if I can do a change in the Media Library interface to add localization improvement by add the "Media Library" in the up-left corner (in the first image look at the awful yellow square) to the dictionary. Also I would like todo a little SCSS code modification to fix some elements.
For example, media library before:
Inkedbefore_LI

Media Library after:
after

PiemP:
Fixes #7277


src/OrchardCore.Modules/OrchardCore.Media/Views/Admin/Index.cshtml
```diff
@@ -2,7 +2,7 @@
 <style asp-src="~/OrchardCore.Media/Styles/media.min.css" debug-src="~/OrchardCore.Media/Styles/media.css"></style>
 
 <script at="Foot">
-    initializeMediaApplication(true, '@Url.Action("MediaApplication", "Admin", new { area = "OrchardCore.Media" })', '@Context.Request.PathBase.ToString()');
+    initializeMediaApplication('@T["Media Library"]', true, '@Url.Action("MediaApplication", "Admin", new { area = "OrchardCore.Media" })', '@Context.Request.PathBase.ToString()');
```
deanmarcussen, line 5: This probably shouldn't be passed in directly, but read from a hidden div on the `MediaApplication.cshtml` view

PiemP: I Hope to have understand well your request 😅. I Hope to have understand well your request. I have opted to don't read the text from an hidden input but I have choose to read the localized string from the root folder link that already render it.

Thank you.

deanmarcussen (commented):
`.text()` makes sense.

@jtkech was there a reason the `var root` was outside the app? I think you added this recently?

but LGTM

jtkech:
@deanmarcussen 

> @jtkech was there a reason the var root was outside the app? I think you added this recently?

I changed the way the selected folder is initialized, but I didn't move the root variable declaration outside the app.

PiemP:
> @jtkech was there a reason the var root was outside the app? I think you added this recently?

effectively, is not a "safer approach". I have looked to the use of the `root` variable in code and seem related to the `mediaApp` object. But I admit to not tested well well the element and if you prefer I can move the `root` variable to the previous position and set only the `name` property with the localized string.

deanmarcussen (approved):
Cool just checking @jtkech 

@PiemP I also agree root seems scoped.

LGTM

Issue #4094, opened 2019-08-23
Forms Module - Select Widget not working/complete


jonvee:
When using the Forms Module, the Select Widget only has fields for Name and ID. There is no means to add options or add an actual select element to the form. 
My current workaround is adding a liquid widget to the form and hard coding the html.

I only briefly looked at the source code and it appears unfinished.

jptissot:
I had the same issue. I ended up creating a SelectInput type with a bag of SelectOption type and implementing the templates myself.

urbanit:
Still issue active

sebastienros:
PRs are open

Issue #3297, opened 2019-03-08
How can i add options to the SelectPart?


cellpaddding:
I'm using the forms module and work flow module to save form data.
When i add the select part to my form,i find that i can't add options to the select part.
How can i do it?
Thank you.

sebastienros:
I assume we are missing the SELECT and OPTION form elements.

https://github.com/OrchardCMS/OrchardCore/tree/dev/src/OrchardCore.Modules/OrchardCore.Forms/Models

Feel free to create a PR to add these, or even a "Custom" Form element where you could render custom tags.

PR #7281, opened 2020-10-15
Implement missing forms selectpart and options

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/3297
Fixes https://github.com/OrchardCMS/OrchardCore/issues/4094

Similar ui to the predefined text field list settings.

I chose not to do it with `SelectListItem` and / or `optgroup` to keep everything simple, and consistent with the rest of the forms module.

Screenshot 2020-10-14 at 15 42 31


src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/SelectPart.Fields.Edit.cshtml
```diff
@@ -0,0 +1,89 @@
...
+                <th scope="col" colspan="3">@T["Default?"]</th>
+            </tr>
+        </thead>
+        <draggable v-model="data.options" :tag="'tbody'">
+            <tr v-for="(option, index) in data.options" :key="index">
+                <td>
+                    <input type="text" class="form-control" v-model="option.text" placeholder="Enter the text" />
+                </td>
+                <td>
+                    <input type="text" class="form-control courier" v-model="option.value" placeholder="Enter a value" />
```
agriffard, line 26: @T["Enter a value"]

ns8482e:
@deanmarcussen I have started working on this too.

Can you add my request here? Support for `Type`, placeholder and optgroup? 

e.g. Dropdown list, radio list, List box, checkbox list?

src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/SelectPart.Fields.Edit.cshtml
deanmarcussen: done.

@agriffard we should probably update the `TextFieldPredefinedListEditorSettings.Edit` as that was copied straight over, and they're missing there too

deanmarcussen:
> Can you add my request here? Support for `Type`, placeholder and optgroup?
> 
> e.g. type being Dropdown list, radio list, List box, checkbox list?

I thought about those, but I saw this as a `SelectPart` and would see a radio group as a different part.

Maybe not?

it could be extended for optgroup, but I had to do this for a client, and wanted to keep it simple, so thought I'd push it up after.

Feel free to take over the pr, I am short of time to do anything much more to it

deanmarcussen:
> Technically if we map `SelectPart` to `<select>` tag than it seems different. 

Yes, I essentially agree, but currently the form part's are matched to a tag type.

i.e. the input part supports radio buttons etc.

Hence why I felt we should probably continue that. 

Otherwise we're changing our minds partway through, and they become a strange mix.

deanmarcussen:
Might be worth asking @sfmskywalker what the original design intentions of the forms module was?

ns8482e:
> i.e. the input part supports radio buttons etc.

OK, But IMO - using radio/checkbox with input requires and making list of choices and maintaining `ModelState` on postback requires some sort of `html` `asp.net` knowledge :).  

It's kinda too much expectation from non technical form designer.

ns8482e:
@deanmarcussen anyway continue with your PR, I'll create PR for my request.

src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/SelectPart.cshtml
```diff
@@ -0,0 +1,26 @@
...
+            fieldClass = "form-control input-validation-error";
+        }
+    }
+}
+<select id="@fieldId" name="@fieldName" class="@fieldClass">
+    @foreach(var option in Model.Value.Options)
+    {
+        <option value="@option.Value" selected="@(option.Value == fieldValue ? "selected" : null)">@option.Text</option>        
+    }
+</select>
```
ns8482e, line 26: EOL

src/OrchardCore.Modules/OrchardCore.Forms/Drivers/SelectPartDisplayDriver.cs
```diff
@@ -0,0 +1,65 @@
...
+                part.DefaultValue = viewModel.DefaultValue;
+                try
+                {
+                    part.Options = String.IsNullOrWhiteSpace(viewModel.Options)
+                        ? Array.Empty<SelectOption>()
+                        : JsonConvert.DeserializeObject<SelectOption[]>(viewModel.Options);
+                }
+                catch
+                {
+                    updater.ModelState.AddModelError(Prefix, S["The options are written in an incorrect format."]);
```
ns8482e, line 58: missing nameof(Options)

src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/SelectPart.cshtml
```diff
@@ -0,0 +1,26 @@
...
+    var formInputElementPart = Model.Value.ContentItem.As<FormInputElementPart>();
+    var elementId = formElementPart.Id;
+    var fieldName = formInputElementPart.Name;
+    var fieldId = !string.IsNullOrEmpty(elementId) ? elementId : !string.IsNullOrEmpty(fieldName) ? Html.GenerateIdFromName(fieldName) : default(string);
+    var fieldValue = Model.Value.DefaultValue;
+    var fieldClass = "form-control";
+
+    if (ViewData.ModelState.TryGetValue(fieldName, out var fieldEntry))
+    {
+        fieldValue = fieldEntry.AttemptedValue;
```
ns8482e, line 14: It could be  - a joined into a comma-separated string for multiple select

src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/SelectPart.cshtml
```diff
@@ -0,0 +1,26 @@
+@using OrchardCore.Forms.Models
+@model ShapeViewModel<SelectPart>
+@{
+    var formElementPart = Model.Value.ContentItem.As<FormElementPart>();
+    var formInputElementPart = Model.Value.ContentItem.As<FormInputElementPart>();
+    var elementId = formElementPart.Id;
+    var fieldName = formInputElementPart.Name;
+    var fieldId = !string.IsNullOrEmpty(elementId) ? elementId : !string.IsNullOrEmpty(fieldName) ? Html.GenerateIdFromName(fieldName) : default(string);
+    var fieldValue = Model.Value.DefaultValue;
+    var fieldClass = "form-control";
```
ns8482e, line 10: why not just assign directly to html `class="form-control"` and have `fielderror` variable instead?

src/OrchardCore.Modules/OrchardCore.Forms/Drivers/SelectPartDisplayDriver.cs
```diff
@@ -0,0 +1,65 @@
...
+using OrchardCore.Forms.ViewModels;
+
+namespace OrchardCore.Forms.Drivers
+{
+    public class SelectPartDisplayDriver : ContentPartDisplayDriver<SelectPart>
+    {
+        private static readonly JsonSerializerSettings SerializerSettings = new JsonSerializerSettings
+        {
+            ContractResolver = new CamelCasePropertyNamesContractResolver(),
+            Formatting = Formatting.Indented
```
ns8482e, line 19: Why Indented when stored into db?

src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/SelectPart.cshtml
```diff
@@ -0,0 +1,26 @@
+@using OrchardCore.Forms.Models
+@model ShapeViewModel<SelectPart>
+@{
+    var formElementPart = Model.Value.ContentItem.As<FormElementPart>();
+    var formInputElementPart = Model.Value.ContentItem.As<FormInputElementPart>();
+    var elementId = formElementPart.Id;
+    var fieldName = formInputElementPart.Name;
+    var fieldId = !string.IsNullOrEmpty(elementId) ? elementId : !string.IsNullOrEmpty(fieldName) ? Html.GenerateIdFromName(fieldName) : default(string);
```
ns8482e, line 8: May be default to `contentitemid_nameof(part)` instead of default(string)?

jptissot:
Why not just add a bag of SelectItems / OptGroups ?

Skrypt:
I like that Vue.js component I wonder who did it? 🙈 
Just a small comment here. We need to add a margin between those form elements. They all look stacked on top of each others. No clear different margin between a title and the previous elements group.

src/OrchardCore.Modules/OrchardCore.Forms/Assets/js/SelectOptionsEditor/selectOptionsEditor.js
```diff
@@ -0,0 +1,90 @@
...
+                selectOptionsModal.methods.showModal();
+            }
+        }
+    });
+
+}
+
+function IsNullOrWhiteSpace(str) {
+    return str === null || str.match(/^ *$/) !== null;
+}
```
Skrypt, line 90: eof

src/OrchardCore.Modules/OrchardCore.Forms/wwwroot/Scripts/selectOptionsEditor.js
```diff
@@ -0,0 +1,105 @@
...
+      showModal: function showModal() {
+        selectOptionsModal.methods.showModal();
+      }
+    }
+  });
+}
+
+function IsNullOrWhiteSpace(str) {
+  return str === null || str.match(/^ *$/) !== null;
+}
```
Skrypt, line 105: eof

Skrypt:
Yeah it could definitely use a SelectListItem / OptGroups but I believe the Vue.js component doesn't do it for now. We'd need to be able to add OptGroups when we click on the small edit icon and then display a list of those OptGroups as a new column in the Option items. I think it's fine enough for now while we wait on someone to work on adding this on the Vue.js component.

Skrypt (commented):
LGTM after small fixes

ns8482e:
> @ns8482e Is there a functional reason why you added this style?
To reduce white space - when used in smaller columns like 25%, 33%

jptissot:
> 
> 
> Yeah it could definitely use a SelectListItem / OptGroups but I believe the Vue.js component doesn't do it for now. We'd need to be able to add OptGroups when we click on the small edit icon and then display a list of those OptGroups as a new column in the Option items. I think it's fine enough for now while we wait on someone to work on adding this on the Vue.js component.

But why not simply use Widgets and a bag instead of writing a custom implementation ?

Skrypt:
@jptissot That would be for listing ContentItems right? Then this would be more like a ContentPicker.

jptissot:
No, I mean the Select has a Bag (named Options) of SelectItem / SelectOptGroup and you build those as Content Items instead of adding a custom editor for Options.

Skrypt:
Yes a content item picker...

deanmarcussen:
> why not simply use Widgets and a bag instead of writing a custom implementation ?

Yes, I saw your comment on the issue @jptissot , that you had resolved it that way.

It tends to be the way I build sites as well, bags, bagcontainers, and bagitems, to create lists inside content.
But I tend to think that approach is more appropriate to theme development, not our internal module development.

The reason I didn't for this was because you do end up with content items, which makes the templating more complex, and the part becomes not hard typed.

i.e. you can't just have an array on it, it becomes an array of content items.

As I said, I'm reasonably easy on this pr, I needed it for a client, so figured I'd push it back to OC afterwards.

If you guys think it should be done differently, then I don't mind closing it, but the issues have sat there for a year and a half without anyone doing anything to fix it ;)

ns8482e:
> No, I mean the Select has a Bag (named Options) of SelectItem / SelectOptGroup and you build those as Content Items instead of adding a custom editor for Options.

Then it’s kinda data source to select item

Skrypt:
I think we should have both options. These are 2 different option lists which makes sense to have.

src/OrchardCore.Modules/OrchardCore.Forms/wwwroot/Scripts/selectOptionsEditor.js
ns8482e: @Skrypt this file is gulp generated

Skrypt:
@jptissot Will it break your actual SelectPart because of naming? Maybe we could name this one as SimpleSelectPart and come up with a ContentItemSelectPart later on ...

ns8482e:
@Skrypt `SelectPart` and `Select` content type metadata are already defined in Migration

Skrypt:
Ok I understand now why we debate on the implementation.

https://github.com/OrchardCMS/OrchardCore/blob/7de0dbe962a20b518b94aa862ad674d46921707a/src/OrchardCore.Modules/OrchardCore.Forms/Migrations.cs#L76-L84

ns8482e:
I guess yes, you can have multiple drivers contribute in rendering for the same Part. 
See `CommonPart` for example.

Skrypt:
So here what I'm saying is that we should probably decide if that's the name we want to use for the Basic implement with uses the Vue.js viewcomponent.

ns8482e:
You mean `SelectPartDisplayDriver`?  Seems fine. For future enhancement - functionality can be added within that driver.
Also I guess it doesn't affect any custom modules, if it's using the same name for driver.

Skrypt:
Ok so I'm leaving you decide if this PR needs more work or not. I'm fine with it. Better have something than nothing too.

src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/SelectPart.cshtml
deanmarcussen: Not sure I understand what you mean here?

It supports only a single select, a multi select would be another field, with a different name

src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/SelectPart.cshtml
deanmarcussen: Just keeping it the same as all the other form templates.
We could review them as a batch at any time

src/OrchardCore.Modules/OrchardCore.Forms/Drivers/SelectPartDisplayDriver.cs
deanmarcussen: Just for debugging convenience

src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/SelectPart.cshtml
deanmarcussen: again same as the other templates - could be review as a batch

src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/SelectPart.cshtml
ns8482e: Nervermind

src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/SelectPart.cshtml
ns8482e: OK

src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/SelectPart.cshtml
ns8482e: ok

src/OrchardCore.Modules/OrchardCore.Forms/Drivers/SelectPartDisplayDriver.cs
ns8482e: ok

PR #7294, opened 2020-10-16
Lucene Indexation with LocalizationPart issue

### LINKED ISSUE #7278: Lucene indexation: wrong result for Localized ItemList
Hello,

I have a **localized** contentype with a **List behavior**, let's say "NewList".
The list items are localized too , let's say "NewsItem"

I have 2 Lucene Indices, one per culture
 
LuceneIdxNewsItemsIt : returns NewItems for Italian localization 
 
LuceneIdxNewsItemsEn: returns NewItems for English localization 

Here's what's happen.

If I modify a NewItem in a culture (eg IT)  (it is enough just saving it ) 
it appears in the other  LuceneIdxNewsItemsEN too.

If I rebuild the LuceneIdxNewsItemsEN the wrong NewItem disappear and everything works fine.

But I can't rebuild every Lucene Index after every modification.

Is there a way to fix it?

Thanks

Skrypt:
Fix issue with indexing a ContentType with a LocalizationPart in LuceneIndexingContentHandler which indexes directly when we publish or save a content item.

We are doing it properly in the LuceneIndexingService when it runs from a background task. So, if you had the Lucene background task enabled it was probably cleaning up those records. Also, if you we're running periodical rebuild / reset of indexes it would have cleaned up too.

Closes #7278

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
@@ -5,6 +5,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using OrchardCore.ContentLocalization;
 using OrchardCore.ContentManagement;
 using OrchardCore.ContentManagement.Handlers;
 using OrchardCore.ContentPreview;
@@ -82,7 +83,8 @@ namespace OrchardCore.Lucene.Handlers
 
                 foreach (var indexSettings in await luceneIndexSettingsService.GetSettingsAsync())
                 {
-                    bool ignoreIndexedCulture = indexSettings.Culture == "any" ? false : context.ContentItem.Content?.LocalizationPart?.Culture != indexSettings.Culture;
+                    var cultureAspect = await contentManager.PopulateAspectAsync(context.ContentItem, new CultureAspect());
+                    bool ignoreIndexedCulture = indexSettings.Culture == "any" ? false : cultureAspect?.Culture.Name != indexSettings.Culture;
                     if (indexSettings.IndexedContentTypes.Contains(context.ContentItem.ContentType) && !ignoreIndexedCulture)
                     {
                         if (!indexSettings.IndexLatest && !publishedLoaded)
--- a/src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
@@ -82,7 +82,8 @@ namespace OrchardCore.Lucene.Handlers
 
                 foreach (var indexSettings in await luceneIndexSettingsService.GetSettingsAsync())
                 {
-                    if (indexSettings.IndexedContentTypes.Contains(context.ContentItem.ContentType))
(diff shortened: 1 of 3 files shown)


src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
```diff
@@ -82,7 +82,8 @@ private static async Task IndexingAsync(ShellScope scope, IEnumerable<ContentCon
 
                 foreach (var indexSettings in await luceneIndexSettingsService.GetSettingsAsync())
                 {
-                    if (indexSettings.IndexedContentTypes.Contains(context.ContentItem.ContentType))
+                    bool ignoreIndexedCulture = indexSettings.Culture == "any" ? false : context.ContentItem.Content?.LocalizationPart?.Culture != indexSettings.Culture;
```
deanmarcussen, line 85: Does using the `CultureAspect` achieve this? 

We shouldn't be having to look at `ContentItem.Content` directly like this

Skrypt: You mean contentManager.PopulateAspectAsync() ?

Skrypt: My code was fine because we need to validate if the content item has a LocalizationPart else calling contentManager.PopulateAspectAsync() will always try to assign a culture to every content item without knowing if it's attached to a LocalizationPart.

Skrypt: Else, does the contentManager.PopulateAspectAsync(contentItem, new CultureAspect()) return the proper value all the time? I'm not sure, maybe the issue is in there. I need to take a closer look. For now, what I do here fix the issue explained above.

Skrypt: Ok new CultureAspect() sets by default the CultureAspect.Culture to CultureInfo.CurrentUICulture. Here, if we populate the CultureAspect of that content item while wanting to index it, we end up not indexing content items that have no LocalizationPart because it evaluates an inappropriate populated default culture.

jtkech (approved):

src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
deanmarcussen: Ok so culture aspect isn't the right one ;)

So for now I guess is fine, but there is a need a `LocalizationAspect` to return this kind of information.

see this pr too https://github.com/OrchardCMS/OrchardCore/pull/6316

Pop it in as an issue to do later if you prefer

sebastienros: I think it should have used Culture aspect. That's exactly what it's made for. The issue is that it still returns a value if no part is attached?  The maybe the aspect should have an extra property (IsDefaultCulture). Because if we have to use LocalizationPart then it makes the aspect useless. And it prevents users from extending it or creating other ways to define localization parts.

Skrypt: Let's do this in a different PR because it will take way more time to do. Unit tests and else. I just wanted to fix the indexation for now. I'm pretty sure we have other places that we use the LocalizationPart directly like this.

Skrypt: See #7306

PR #7095, opened 2020-09-17
Hide Content Type filter when using Admin Menus Content Type feature

PiemP:
Fixes #7094 

When contentTypeId have a value means the user want to list the items of a specific content type and the ContentType filter is not necessary. I have left the possiblity to set up ContentTypeOptions from the query string.

PR #6551, opened 2020-07-01
Extensible filters for contents admin list refactor

deanmarcussen:
This replaces https://github.com/OrchardCMS/OrchardCore/pull/6249

I've refactored this to use a `IDisplayDriver<ContentOptionsViewModel` on the `ContentOptionsViewModel`.

The previous pr used the `ShapeFactory` exclusively, which worked, but was very complex, and not as easily extensible. (too much checking if a zone exists, and trying to push things around zones.

Using an `IDisplayManager<ContentOptionsViewModel>` and associated drivers, has meant no need for any extra interfaces to manage the `RouteValueCollection` on posts to the `AdminController`, as the drivers are able to handle it on `Update` 

I'll leave both prs open as I would like to review which technique is better.

Fixes #6012
Fixes #4921

PR #7299, opened 2020-10-17
Add ListContent Permission as Content Type dynamic permission

### LINKED ISSUE #7289: Add ListContent Permission to each Content Type permission
image

This would allow to decide which Content Type we want to show only owned content or not.
If the ListContent permission is set on the OrchardCore.Contents this would allow it on every Content Type (override the each Content Type permission) just like we do with the other permissions.

Skrypt:
Adds the ability to set ListContent permission on each Content Type.
The global ListContent permission is of course prioritary over these.

Closes #7289, Fixes regression from #7095, #6551

screencapture-localhost-5000-Admin-Roles-Edit-Test-2020-10-18-16_28_06


Skrypt:
Ok finally it builds. Travis acting funny today.

src/OrchardCore.Modules/OrchardCore.Contents/Controllers/AdminController.cs
```diff
@@ -176,37 +176,33 @@ public async Task<IActionResult> List(ListContentsViewModel model, PagerParamete
                 new SelectListItem() { Text = S["Delete"], Value = nameof(ContentsBulkAction.Remove) }
             };
 
-            // When using the AdminMenus ContentTypes Feature and specifying a Content Type hide the 'Content Type' filter.
```
deanmarcussen, line 179: Ok, I tried this feature added by @PiemP and you are correct it doesn't work properly.

It was supposed to only hide the `ContentType` filter when using the `AdminMenus` `ContentTypes` menu listing (i.e. the one we have by default in the Blog recipe)

But as you say it is hiding them when selected from the normal contents listing (which is bad)

Is it possible to actually make the feature @PiemP wanted to add work, or are we getting rid of it, perhaps to be implemented correctly by @PiemP at a later stage.

Rest of pr looks fine

Skrypt: Yeah I don't think this PR should fix this issue. I reverted the changes and made sure that the Content Type is selected in all cases instead. Should we prevent switching from one Content Type to an other when we display a specific list of Content Types? I'm not sure if it's necessary. The list now populates with Content Types that the user has authorization so there's no risks that it tries to list something that the user can't.

Skrypt: I changed my mind, I'm going to fix it.

Skrypt: Done

PR #7307, opened 2020-10-17
Use CultureAspect in Lucene

### LINKED ISSUE #7306: Fix CultureAspect extensibility
We need to refactor code in places where we access directly the LocalizationPart to get the ContentItem current culture.
Instead we should use the `_contentManager.PopulateAspectAsync(contentItem, new CultureAspect());` or something similar so that if someone implements a different ContentPart that could alter the ContentItem culture ; it evaluates both.

## TODO : 

- [ ] Evaluate changes to be made with CultureAspect class.
- [ ] Create a Unit Test that test extensibility.

## Examples : 

https://github.com/OrchardCMS/OrchardCore/blob/cfe0237eb843c103f113d11dfe093c0343ca572b/src/OrchardCore.Modules/OrchardCore.Lucene/Services/LuceneIndexingService.cs#L189

https://github.com/OrchardCMS/OrchardCore/blob/cfe0237eb843c103f113d11dfe093c0343ca572b/src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs#L85

So far it is the only 2 places I found where we do this. But we probably also use a similar technique with Recipes.
Please report other places where it needs to be fixed here.

hishamco:
Fixes #7306

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
@@ -66,7 +66,6 @@ namespace OrchardCore.Lucene.Handlers
             var luceneIndexManager = services.GetRequiredService<LuceneIndexManager>();
             var luceneIndexSettingsService = services.GetRequiredService<LuceneIndexSettingsService>();
             var logger = services.GetRequiredService<ILogger<LuceneIndexingContentHandler>>();
-            var orchardHelper = services.GetService<IOrchardHelper>();
             // Multiple items may have been updated in the same scope, e.g through a recipe.
             var contextsGroupById = contexts.GroupBy(c => c.ContentItem.ContentItemId, c => c);
 
@@ -81,8 +80,9 @@ namespace OrchardCore.Lucene.Handlers
 
                 foreach (var indexSettings in await luceneIndexSettingsService.GetSettingsAsync())
                 {
-                    var culture = await orchardHelper.GetContentCultureAsync(context.ContentItem);
-                    var ignoreIndexedCulture = indexSettings.Culture == "any"? false : culture?.Name != indexSettings.Culture;
+                    var cultureAspect = await contentManager.PopulateAspectAsync<CultureAspect>(context.ContentItem);
+                    var culture = cultureAspect.HasCulture ? cultureAspect.Culture.Name : null;
+                    var ignoreIndexedCulture = indexSettings.Culture == "any"? false : culture != indexSettings.Culture;
 
                     if (indexSettings.IndexedContentTypes.Contains(context.ContentItem.ContentType) && !ignoreIndexedCulture)
                     {
--- a/src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
@@ -66,6 +66,7 @@ namespace OrchardCore.Lucene.Handlers
             var luceneIndexManager = services.GetRequiredService<LuceneIndexManager>();
             var luceneIndexSettingsService = services.GetRequiredService<LuceneIndexSettingsService>();
             var logger = services.GetRequiredService<ILogger<LuceneIndexingContentHandler>>();
+            var orchardHelper = services.GetService<IOrchardHelper>();
             // Multiple items may have been updated in the same scope, e.g through a recipe.
             var contextsGroupById = contexts.GroupBy(c => c.ContentItem.ContentItemId, c => c);
 
@@ -80,8 +81,8 @@ namespace OrchardCore.Lucene.Handlers
 
                 foreach (var indexSettings in await luceneIndexSettingsService.GetSettingsAsync())
                 {
-                    var cultureAspect = await contentManager.PopulateAspectAsync(context.ContentItem, new CultureAspect());
-                    var ignoreIndexedCulture = indexSettings.Culture == "any" ? false : cultureAspect.Culture.Name != indexSettings.Culture;
+                    var culture = await orchardHelper.GetContentCultureAsync(context.ContentItem);
+                    var ignoreIndexedCulture = indexSettings.Culture == "any"? false : culture?.Name != indexSettings.Culture;
 
                     if (indexSettings.IndexedContentTypes.Contains(context.ContentItem.ContentType) && !ignoreIndexedCulture)
                     {
--- a/src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
@@ -5,6 +5,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using OrchardCore.ContentLocalization;
 using OrchardCore.ContentManagement;
 using OrchardCore.ContentManagement.Handlers;
 using OrchardCore.ContentPreview;
@@ -60,14 +61,11 @@ namespace OrchardCore.Lucene.Handlers
         private static async Task IndexingAsync(ShellScope scope, IEnumerable<ContentContextBase> contexts)
         {
             var services = scope.ServiceProvider;
-
             var contentManager = services.GetRequiredService<IContentManager>();
             var contentItemIndexHandlers = services.GetServices<IContentItemIndexHandler>();
-
             var luceneIndexManager = services.GetRequiredService<LuceneIndexManager>();
             var luceneIndexSettingsService = services.GetRequiredService<LuceneIndexSettingsService>();
             var logger = services.GetRequiredService<ILogger<LuceneIndexingContentHandler>>();
-
             // Multiple items may have been updated in the same scope, e.g through a recipe.
             var contextsGroupById = contexts.GroupBy(c => c.ContentItem.ContentItemId, c => c);
 
@@ -82,7 +80,8 @@ namespace OrchardCore.Lucene.Handlers
 
                 foreach (var indexSettings in await luceneIndexSettingsService.GetSettingsAsync())
                 {
-                    var ignoreIndexedCulture = indexSettings.Culture == "any" ? false : context.ContentItem.Content?.LocalizationPart?.Culture != indexSettings.Culture;
+                    var cultureAspect = await contentManager.PopulateAspectAsync(context.ContentItem, new CultureAspect());
+                    var ignoreIndexedCulture = indexSettings.Culture == "any" ? false : cultureAspect.Culture != indexSettings.Culture;
 
                     if (indexSettings.IndexedContentTypes.Contains(context.ContentItem.ContentType) && !ignoreIndexedCulture)
                     {
--- a/src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
@@ -81,7 +81,7 @@ namespace OrchardCore.Lucene.Handlers
                 foreach (var indexSettings in await luceneIndexSettingsService.GetSettingsAsync())
                 {
                     var cultureAspect = await contentManager.PopulateAspectAsync(context.ContentItem, new CultureAspect());
-                    var ignoreIndexedCulture = indexSettings.Culture == "any" ? false : cultureAspect.Culture != indexSettings.Culture;
+                    var ignoreIndexedCulture = indexSettings.Culture == "any" ? false : cultureAspect.Culture.Name != indexSettings.Culture;
 
                     if (indexSettings.IndexedContentTypes.Contains(context.ContentItem.ContentType) && !ignoreIndexedCulture)
                     {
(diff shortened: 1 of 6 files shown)


hishamco:
@Skrypt before I move further into all other places if there are, could you please let me know how to test the localization change in Lucene coz this is the first time I will try this module ;)

Skrypt:
Create a custom Content Type.
Associate the LocalizationPart on it.
Create 2 different Lucene Index targetting 2 different specific cultures of your choice.
Create content items and check in which Index it gets indexed by using the match_all query.

hishamco:
Oops, seems `indexSettings.Culture` is a string ;)

Skrypt:
I think we added the default culture assignment to the CultureAspect because we presumed it would be used to get populated for the frontend only. We made a mistake by thinking so. I don't want to break the actual other implementations, so it will require some testing of Content Localization feature.

hishamco:
> Here when we index we don't want this to happen because we want that if a Content Item has no LocalizationPart that the culture returned/evaluated/populated be null.

I see, I will try to consider this, but my question what will happen if the default culture return? Seems it 'll be indexed with that culture, Am I right? how this will affect lucene search?

Skrypt:
It will get indexed in the wrong index 😉

hishamco:
> We made a mistake by thinking so. I don't want to break the actual other implementations, so it will require some testing of Content Localization feature (if we do any refactoring).

I think we shouldn't break anything at this stage, probably we can look into Content Localization after `1.0.0`, IMHO for now we can use `CultureAspect` whenever it's possible. I will try to make tweak this to make it work in Lucene

hishamco:
> Here when we index we don't want this to happen because we want that if a Content Item has no LocalizationPart that the culture returned/evaluated/populated be null.

@Skrypt is that also true if the index set for any culture?

Skrypt:
Yes, it needs to be equal to what is stored in a LocalizationPart. If there is no LocalizationPart then we should index a null value.

Skrypt:
Found what would work.

```C#
                                var contentItemCultureAspect = await contentManager.PopulateAspectAsync(context.ContentItem, new CultureAspect(){ Culture = null });
                                var ignoreIndexedCulture = settings.Culture == "any" ? false : contentItemCultureAspect.Culture?.Name != settings.Culture;

```

src/OrchardCore/OrchardCore.ContentLocalization.Abstractions/OrchardHelperExtensions.cs
```diff
@@ -21,7 +22,11 @@ public async static Task<CultureInfo> GetContentCultureAsync(this IOrchardHelper
             var contentManager = orchardHelper.HttpContext.RequestServices.GetService<IContentManager>();
             var cultureAspect = await contentManager.PopulateAspectAsync(contentItem, new CultureAspect());
 
-            return cultureAspect.Culture;
+            // Workaround to ignore the default culture that has been set to CultureAspect.Culture
+            // for the content item which doesn't have a LocalizationPart attached
+            return contentItem.Has<LocalizationPart>()
```
Skrypt, line 27: Here we end up with the same problem. We don't want to know about any part from here, it's the PopulateAspectAsync that should take care of returning the proper CultureAspect.

Skrypt: Here in this Method we would need to be able to say default the value to CultureInfo.CurrentUICulture or not.

hishamco:
@Skrypt frankly I already used `PopulateAspectAsync` before `GetContentCultureAsync()`, but I remembered @sebastienros suggest to use it long time back instead of `LocalizationPart`, which is make sense instead of repeating things, so fix it once and use it everywhere

If you complain about `contentItem.Has<LocalizationPart>` I already come up with a solution that works fine, which is to check if the retrieved cultures is supported or not

I suggest to continue with this PR unless the guys have another opinion

Thanks

src/OrchardCore/OrchardCore.ContentLocalization.Abstractions/OrchardHelperExtensions.cs
hishamco, line 27: We could use `ILocalizationService` to ensure whether the `CultureAspect.Culture` is supported or not, this way no need to use `LocalizationPart` at all

Skrypt:
I will take a look tomorrow. I don't have a preference for now. I agree though that if we could use the same method everywhere that be better. I tried fixing the GetContentCultureAsync method differently too but so far no luck. The only thing I don't want to see is the use of the LocalizationPart directly in these methods. The LocalizationPart should be affecting the result only when we do PopulateAspectAsync().

hishamco:
> I tried fixing the GetContentCultureAsync method differently too but so far no luck. The only thing I don't want to see is the use of the LocalizationPart directly in these methods.

I already mentioned that above in my comment [here](https://github.com/OrchardCMS/OrchardCore/pull/7307#discussion_r507515995)

Skrypt:
Yeah, sorry I'm tired. Going to sleep 😉

src/OrchardCore/OrchardCore.ContentLocalization.Abstractions/OrchardHelperExtensions.cs
Skrypt: I don't see what you mean. I need to check.

hishamco:
No problem dude, sleep well, it's morning here ;)

src/OrchardCore/OrchardCore.ContentLocalization.Abstractions/OrchardHelperExtensions.cs
Skrypt: Good thing is that we're getting closer to something.

Skrypt: Ok I can live with this for now but we need to find something better than looking for a specific ContentPart. Maybe more something like look for any ContentPart that inherits ILocalizable.

Skrypt: Ok I can live with this for now but we need to find something better than looking for a specific ContentPart. Maybe more something like look for any ContentPart that inherits ILocalizable.

hishamco: The little issue is occur in the unit tests, coz the default culture will be supported, but I need to do more test to check what's going on

The reason is if I'm not wrong that `CultureInfo.CurrentUICulture` equals `CultureInfo.InstalledUICulture.Name`, Let us think while I finish some tests

Skrypt: And also, I think it will break some Views where we look for a culture for defining RTL or LTR. Because now it could return a null value. That's mainly why I was not fixing it here.

hishamco: Did you check my comment [here](https://github.com/OrchardCMS/OrchardCore/pull/7307#discussion_r508064029)

Skrypt:
While what Hisham did here in his global method looks fine it will break some UI features so we keep this on hold for more testing and fixing if we use this method. I had a simpler fix which doesn't break anything but @hishamco wants to fix the main method that returns the content item culture which now can return a null value which wasn't before.

Skrypt (changes requested):
While what Hisham did here in his global method looks fine it will break some UI features so we keep this on hold for more testing and fixing if we use this method. I had a simpler fix which doesn't break anything but @hishamco wants to fix the main method that returns the content item culture which now can return a null value which wasn't before (this is what will break LocalizationPart UI views and else). 

This goes against the design of using the CultureAspect.Culture to be set to CultureInfo.CurrentUICulture by default.
Also, I believe that the CultureAspect was made to be used on frontend things and we return a default culture because a null culture doesn't exist for that matter. Though, here when we save the content item culture value it can be null. 

**We need to make a distinction between business logic and views/drivers logic.**

hishamco:
> While what Hisham did here in his global method looks fine it will break some UI features

Give me some UI that was broken

> so we keep this on hold for more testing and fixing if we use this method

Agree, I will try to update this PR once I have time and have some extra tests

Skrypt:
Try creating a Content Type that has no LocalizationPart. Create a content Item. Then add the LocalizationPart to it and try to edit it to see that it won't return a default culture to be saved against it. That's why we added the default culture on the CultureAspect I think.

hishamco:
@Skrypt I have an idea probably crazy, but it solve the issue with minimal changes, what if we add a property something like `IsOriginalCultureNull` or any better name, with that the `Culture` property will still the same with default value `CultureInfo.CurrentUICulture`, but the new property will be useful especially for lucene to indicate that the original culture is `null`

Skrypt:
Yes the GetContentItemCulture should have a param that says : here I want my default to be a null culture instead of the one provided by the CultureAspect.

hishamco:
May be we can rename it to `GetContentCultureOrDefaultAsync()`

Skrypt:
Nah I think the name is fine, just add a comment on the method.

hishamco:
The comment is useful for those who have access to the code only, I see similar pattern in Microsoft APIs as well as OC

Skrypt:
Comments on the method shows when you use the method elsewhere. You can add param comments. Of course inner comments won't show.

hishamco:
The main question here, if we didn't add metadata in `CultureAspect` we will reply on `HasPart` which is the thing that you don't like, so your thought

Skrypt:
We should not rely on any part. If you allow to use a different default culture, then you don't need to.

hishamco:
> When you do populateAspect() it triggers the proper "LocalizationPartHandlers" that will affect the content item culture and it's fine like that. It abstract this part.

Got it, but we use

```csharp
var cultureAspect = await contentManager.PopulateAspectAsync(contentItem, new CultureAspect());
```

The culture will be `CultureInfo.CurrentUICulture` by default, so no change to know if the original culture is `null`, hope you got my point

Skrypt:
I know this is quite difficult because it's a static class/method but something like this could work :  

``` C#
var cultureAspect = await contentManager.PopulateAspectAsync(contentItem, new CultureAspect(){ Culture = IsOriginalCultureNull ? null : CultureInfo.CurrentUICulture });
```

hishamco:
I thought you will do it without `IsOriginalCultureNull` ;) ok, just leave it to me and I will see what can I come up with, I will keep you  updated

src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
```diff
@@ -82,7 +81,8 @@ private static async Task IndexingAsync(ShellScope scope, IEnumerable<ContentCon
 
                 foreach (var indexSettings in await luceneIndexSettingsService.GetSettingsAsync())
                 {
-                    var ignoreIndexedCulture = indexSettings.Culture == "any" ? false : context.ContentItem.Content?.LocalizationPart?.Culture != indexSettings.Culture;
+                    var culture = await orchardHelper.GetContentCultureAsync(context.ContentItem);
```
deanmarcussen, line 84: Don't use the `IOrchardHelper` here. Use the `IContentManager.PopulateAspect`

The helper methods are for the ui/users of OC, not internal code.

deanmarcussen (commented):
Reviewing this, I think the problem, as you have already noted @hishamco is that the `CultureAspect` might not be always useful, in that it returns a default.

In a mono culture site, or varied culture (as in some content is localized, some isn't) this isn't always useful.

Personally I think the `CultureAspect` should have a second property `HasCulture` returning a bool, set to false initially, whereby you can check first to see if a content item has had a culture specified.

This feels non breaking, and I think would resolve the issue in lucene, where it is necesary to know if there is a culture assigned or not, first, before then trying to get that culture.

There could be helper methods on the `IOrchardHelper` as well, but I would suggest to focus on getting the `CultureAspect` useful for all scenarios.

hishamco:
IMHO we need to explicit assign `CultureInfo.CurrenUICulture` to `CultureAspect.Culture`, so `HasCulture` could be a ready only property, but for now it's fine, unless we know exactly where are the places that need to set culture by default

hishamco:
@Skrypt could you please review this?

src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
```diff
@@ -82,7 +80,9 @@ private static async Task IndexingAsync(ShellScope scope, IEnumerable<ContentCon
 
                 foreach (var indexSettings in await luceneIndexSettingsService.GetSettingsAsync())
                 {
-                    var ignoreIndexedCulture = indexSettings.Culture == "any" ? false : context.ContentItem.Content?.LocalizationPart?.Culture != indexSettings.Culture;
+                    var cultureAspect = await contentManager.PopulateAspectAsync<CultureAspect>(context.ContentItem);
+                    var culture = cultureAspect.HasCulture ? cultureAspect.Culture.Name : null;
+                    var ignoreIndexedCulture = indexSettings.Culture == "any"? false : culture != indexSettings.Culture;
```
deanmarcussen, line 85:
```suggestion
                    var ignoreIndexedCulture = indexSettings.Culture == "any" ? false : culture != indexSettings.Culture;
```

deanmarcussen (commented):
I would wait for @Skrypt to check it, but I think you have a better / more useful aspect api now

Skrypt:
I will review this soon. Looks good @hishamco

Skrypt (approved):

PR #7313, opened 2020-10-19
Move feeds to feature

ns8482e:
Move feeds to feature

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Lists/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lists/Startup.cs
@@ -68,9 +68,6 @@ namespace OrchardCore.Lists
             services.AddScoped<IContentItemIndexHandler, ContainedPartContentIndexHandler>();
             services.AddScoped<IContainerService, ContainerService>();
 
-            // Feeds
-            // TODO: Create feature
-            services.AddScoped<IFeedQueryProvider, ListFeedQuery>();
             services.AddContentPart<ListPart>()
                 .UseDisplayDriver<ListPartFeedDisplayDriver>()
                 .AddHandler<ListPartFeedHandler>();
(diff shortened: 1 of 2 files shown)


src/OrchardCore.Modules/OrchardCore.Lists/Startup.cs
```diff
@@ -121,4 +107,29 @@ public override void ConfigureServices(IServiceCollection services)
...
+    {
+        public override void ConfigureServices(IServiceCollection services)
+        {
+            // Feeds
+            services.AddScoped<IFeedQueryProvider, ListFeedQuery>();
+
+            services.AddContentPart<ListPart>()
+                .UseDisplayDriver<ListPartFeedDisplayDriver>()
+                .AddHandler<ListPartFeedHandler>();
+
```
hishamco, line 122: Just remove the two extra lines

hishamco (approved):
LGTM

deanmarcussen (changes requested):
Good, yes, but I need to check it is non breaking.

Please hold off on merging for a little bit

hishamco:
Sure, I need to try this too ;)

deanmarcussen:
Ok, so it's non breaking which is good.

But I think the comment actually meant to move the `ListPart` feeds provider to an actual feature, so that one could have feeds, without lists being a provider of feeds.

i.e. still providing custom feeds from other sources.

but better change than nothing, so ok with me.

Niraj if you wanted to move it to an actual feature, it would just need to enable the feature by default in a migration on Lists (to be non breaking)

ns8482e:
I changed to hide the following UI for users when `OrchardCore.Feeds` is disabled as it's returning `404`.

image

The change in this PR will only display FeedUrl options when when `OrchardCore.Feeds` is enabled.

deanmarcussen (approved):

PR #7314, opened 2020-10-19
Fix CultureAspect extensibility

Skrypt:
Here is the solution I came up with @hishamco 
Without changing the default assignment of the CultureAspect class Culture to CultureInfo.CurrentUICulture.
Feels like we should keep this since using a null culture can always be more problematic than anything.
But here in the Lucene Index we want to actually have the culture set to null if the content item has no LocalizationPart or other ContentPart that would set it's culture.

If we would store the CultureInfo.CurrentUICulture here it would be misleading if the website changes from one system to an other where the CurrentUICulture would be set from the actual OS culture. Null culture here means that it is the same culture than the default one which can vary. Also, it allows to store these values in the appropriate index all the time.

I have not changed the GetContentCultureAsync method because it is fine to be used mainly on UI things where we can't use a null culture.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
@@ -5,6 +5,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using OrchardCore.ContentLocalization;
 using OrchardCore.ContentManagement;
 using OrchardCore.ContentManagement.Handlers;
 using OrchardCore.ContentPreview;
@@ -82,7 +83,8 @@ namespace OrchardCore.Lucene.Handlers
 
                 foreach (var indexSettings in await luceneIndexSettingsService.GetSettingsAsync())
                 {
-                    var ignoreIndexedCulture = indexSettings.Culture == "any" ? false : context.ContentItem.Content?.LocalizationPart?.Culture != indexSettings.Culture;
+                    var contentItemCultureAspect = await contentManager.PopulateAspectAsync(context.ContentItem, new CultureAspect(){ Culture = null });
+                    var ignoreIndexedCulture = indexSettings.Culture == "any" ? false : contentItemCultureAspect.Culture?.Name != indexSettings.Culture;
 
                     if (indexSettings.IndexedContentTypes.Contains(context.ContentItem.ContentType) && !ignoreIndexedCulture)
                     {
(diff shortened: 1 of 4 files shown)


src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
```diff
@@ -82,7 +83,8 @@ private static async Task IndexingAsync(ShellScope scope, IEnumerable<ContentCon
 
                 foreach (var indexSettings in await luceneIndexSettingsService.GetSettingsAsync())
                 {
-                    var ignoreIndexedCulture = indexSettings.Culture == "any" ? false : context.ContentItem.Content?.LocalizationPart?.Culture != indexSettings.Culture;
+                    var contentItemCultureAspect = await contentManager.PopulateAspectAsync(context.ContentItem, new CultureAspect(){ Culture = null });
```
Skrypt, line 86: Here we don't need to call GetContentCultureAsync() we already have the contentManager to use it directly.

src/OrchardCore/OrchardCore.ContentLocalization.Abstractions/OrchardHelperExtensions.cs
```diff
@@ -15,11 +15,12 @@ public static class OrchardHelperExtensions
         /// </summary>
         /// <param name="orchardHelper">The <see cref="IOrchardHelper"/>.</param>
         /// <param name="contentItem">The <see cref="ContentItem"/> in which to get its culture.</param>
+        /// <param name="defaultCulture">The <see cref="CultureInfo"/> the default culture it should fallback to.</param>
         /// <returns></returns>
-        public async static Task<CultureInfo> GetContentCultureAsync(this IOrchardHelper orchardHelper, ContentItem contentItem)
+        public async static Task<CultureInfo> GetContentCultureAsync(this IOrchardHelper orchardHelper, ContentItem contentItem, CultureInfo defaultCulture = CultureInfo.CurrentUICulture)
```
Skrypt, line 20: Here I'm adding an optional defaultCulture param if anyone needs to do the same kind of thing somewhere else where the ContentManager is not injected e.g. a view template.

src/OrchardCore.Modules/OrchardCore.Lucene/Handler/LuceneIndexingContentHandler.cs
hishamco: Agree and disagree ;) nothing wrong with using `GetContentCultureAsync()`  even we already have the contentManager to use it directly. Fix it once and use it everywhere, IMHO we don't wanna to enforce ourselves or everyone to write the changes in multiple places, which is the aim of the issue that you wrote. Please have a look to my comment [here](https://github.com/OrchardCMS/OrchardCore/pull/7307#issuecomment-711699574)

Issue #5605, opened 2020-02-20
Officially supported darkmode theme


mawendel:
**I would like to see an officially supported darkmode admin theme. I'm hoping we can get more than a few to chime in and make this happen.**

After briefly looking at how to implement my own darkmode admin theme, my first thoughts are that it looks pretty tough coming from the outside, with no prio knowledge of how the admin theme is built up.

I tried using the compiled bootstrap.css from https://bootswatch.com/darkly/ with no luck, there are too many custom tags overriding what should otherwise take precedence. It seems like it would take me a lot of time to figure out what things are overriden, where it comes from and how to target the individual tags.

The maintaining of my own separate version could also become quite the chore, considering that any changes made to the official TheAdmin theme could be quite opaque. Then having to dig through thousands of lines of css tags to figure out how to make your admin theme work again..

_It does not have to be the Darkly skin we use, it was merely a way to test it. In fact im not sure im sold on the white background for all inputs._

Skrypt:
I'm surprised to hear that. Would you send us some screenshots of this attempt? The only part that is strongly designed is the left menu navbar to be honest. Maybe some other zones too like the header navbar. But generally we stick to BS4 classes for most of the work.

deanmarcussen:
It may just need to be compiled from sass with the rtl support from the gulpfile, that's generally what is changing the bootstrap tags

Skrypt:
I'm trying exactly that right now

Skrypt:
![image](https://user-images.githubusercontent.com/3228637/74965613-8f97e700-53e3-11ea-99fe-ce51478b0d43.png)

Just like what I thought

mawendel:
I didnt even get that. But i also only referenced the bootstrap.css in my Layout. If you used some custom sass compilation or something else that might explain why yours looks partly there?

Skrypt:
![image](https://user-images.githubusercontent.com/3228637/74970520-7e9fa380-53ec-11ea-989d-a5a27879fb6f.png)

Yes I used the sass files and created an alternate TheAdminDark.scss file with it.

sebastienros:
@Skrypt people are asking for it, maybe make it official in the admin theme, with a setting that is stored in a cookie. It should just be a scss file that we can update for this option.

hishamco:
We could support light and dark themes out of the box, making an option to add new themes and switches between them from within the admin site

mawendel:
[...]

PR #7326, opened 2020-10-20
Admin Theme Extensibility

Skrypt:
Work In Progress to answer : https://github.com/OrchardCMS/OrchardCore/issues/5605

Using Bootswatch "Darkly" as a base to do a proper admin dark theme at the same time.

orchard

- [x] Allow to extend admin sidebar menu through SASS vars.
- [x] Allow to extend general styles by following practices used in Bootswatch themes.
  - [x] Background headers should use bg-primary instead of bg-light.
- [x] Allow to extend our custom admin styles by defining colors used in SASS vars.
- [ ] Add an option on themes to allow selecting a different stylesheet "bootswatch".
- [ ] Add a user preference to allow selecting a prefered style on a theme. e.g. (dark mode). 
  - [x] LocalStorage based.
  - [ ] Cookie based.
- [ ] Allow extending module styles with admin SASS mixins, funcs, vars.
- [x] Automatically detect dark mode from OS dark mode with javascript. See `prefers-color-scheme: dark` media selector.
- [x] Add admin setting to enable/disable the dark mode admin theme feature.

## Idea

Module extensibility could use a node module that includes all modules SASS files to allow theming them. This way we would be able to use sass variables, functions, mixins to override existing module styles. Right now we have a css generated per module, an alternative would be to move all modules sass files to TheAdminTheme. Using both techniques would allow to properly be able to style every elements from a source code or nuget packages solution AND use SASS.

More on this. We could also add the TheAdmin mixins, func, vars to that node module to allow define our module styles. Right now we can't use the TheAdmin mixins within our modules for example.

### Dark Mode

Tracking Bootstrap Dark Mode feature : twbs/bootstrap#27514

While this should be the way to do this it is more likely to be done for BS 6 rather than 5 so for now we need to rely on a theme switcher for this unfortunately.

Closes : https://github.com/OrchardCMS/OrchardCore/issues/5605


Skrypt:
I think I did enough for this to get merged. Other features could be achieved in separate PR's. We should merge this sooner than later because of the SASS file structure changes.

src/OrchardCore.Themes/TheAdmin/Assets.json
```diff
@@ -3,9 +3,17 @@
     "generateSourceMaps": false,
     "generateRTL": true,
     "inputs": [
-      "Assets/scss/TheAdmin.scss"
+      "Assets/scss/admin.scss"
```
deanmarcussen, line 6: Do we have to change the name?

This actually breaks some of my custom admin themes (which rely on the existing name to pull in the normal admin.css code from `TheAdmin` theme, before applying other style overrides)

src/OrchardCore/OrchardCore.ResourceManagement/TagHelpers/StyleTagHelper.cs
```diff
@@ -5,16 +5,21 @@
 
 namespace OrchardCore.ResourceManagement.TagHelpers
 {
+    [HtmlTargetElement("style", Attributes = IdAttributeName)]
     [HtmlTargetElement("style", Attributes = NameAttributeName)]
     [HtmlTargetElement("style", Attributes = SrcAttributeName)]
     [HtmlTargetElement("style", Attributes = AtAttributeName)]
     public class StyleTagHelper : TagHelper
     {
+        private const string IdAttributeName = "id";
```
deanmarcussen, line 14: This should be `asp-id`.

It also shouldn't be part of the `HtmlTargetElement` attribute set (I think), as this would cause the tag helper to be activated when only using the `asp-id` attribute.

Activation should only happen on the name, src, or at attribute.

deanmarcussen (commented):
If I have time I'll have a look at the tag helper later.

I think there might be a nicer way to do that

src/OrchardCore.Themes/TheAdmin/Assets.json
Skrypt: Yes because you will be forced to transpile 2 different css now. One for dark mode else the UI will fail to load it. Changing the name forces people to simply reference the new 2 files. In the end you just have to change a name and create a second sass file to generate a second dark mode stylesheet.

src/OrchardCore/OrchardCore.ResourceManagement/TagHelpers/StyleTagHelper.cs
Skrypt: I'm fine with asp-id. I did not create a lot of those TagHelpers I don't know all the tricks yet.

src/OrchardCore.Themes/TheAdmin/Assets.json
Skrypt: Plus here, we have a new way to generate those styles without requiring to override on top of the TheAdmin.css file.

Skrypt:
Tracking Bootstrap Dark Mode feature : https://github.com/twbs/bootstrap/issues/27514

src/OrchardCore.Themes/TheAdmin/Assets.json
Skrypt: Reverted the name change. Problem solved, changed the javascript to refelct this change.

src/OrchardCore.Themes/TheAdmin/Assets/js/darkmode.js
```diff
@@ -0,0 +1,23 @@
+var darkmode = (darkmode === undefined) ? false : darkmode;
+
+$(function () {
+    if ($('body').hasClass('darkmode') || darkmode)
+    {
+        $('#admin-css').attr('href', $('#admin-css').attr('href').replace('/TheAdmin.css', '/TheAdmin-darkmode.css'));
```
deanmarcussen, line 6: Longer term we will want a way for these user preferences to be stored against the User property rather than in cookies, and for this to be a `Condition` applied to the resource tag. 
But I would not block the feature on this.

deanmarcussen (commented):
LGTM. But I think @agriffard pr for the regressions should be merged first, and this css rebuilt.

src/OrchardCore.Themes/TheAdmin/Assets/js/darkmode.js
Skrypt: Last thing I want to add is the automatic selection of the dark mode if your OS is on dark mode.
Will work on small details like these today then I will merge it when I think it's solid enough.

Skrypt: I think you meant User IEntity.Properties? Right now I use LocalStorage of the browser which cannot be read from server side. Maybe it should use a Cookie for all those AdminPreferences instead. 

Not sure what you mean with the Condition applied on the resource tag and what it would be for.

src/OrchardCore.Themes/TheAdmin/Views/Layout.cshtml
```diff
@@ -8,7 +14,11 @@
     <link type="image/x-icon" rel="shortcut icon" href="~/TheAdmin/favicon.ico" />
 
     <!-- Bootstrap CSS -->
-    <style asp-name="admin" asp-src="~/TheAdmin/Styles/TheAdmin.min.css" debug-src="~/TheAdmin/Styles/TheAdmin.css" depends-on="jQuery-ui"></style>
+    <style asp-id="admin-default" asp-media="all" asp-name="admin" asp-src="~/TheAdmin/Styles/TheAdmin.min.css" debug-src="~/TheAdmin/Styles/TheAdmin.css" depends-on="jQuery-ui"></style>
+    @if (adminSettings.DisplayDarkmode)
```
Skrypt, line 18: I don't like this. I tried using a @section with a Partial view but seems we can't do that.
Any alternative you guys might think of?

Skrypt:
@deanmarcussen @jtkech Please take a look and tell me if you understand why the CI doesn't pass the functional tests.

jtkech:
@Skrypt Sorry too late i really need to leave, i will take a look tomorrow night if dean didn't fix it in the meantime ;)

src/OrchardCore.Modules/OrchardCore.Admin/Admin/AdminSettings.cs
```diff
@@ -3,5 +3,7 @@ namespace OrchardCore.Admin.Models
     public class AdminSettings
     {
         public bool DisplayMenuFilter { get; set; }
+
+        public bool DisplayDarkmode { get; set; }
```
deanmarcussen, line 7: I think should be true by default.

Otherwise people will not know it's available

(Same with `DisplayMenuFilter` IMO @agriffard do you agree / disagree?)

src/OrchardCore/OrchardCore.ResourceManagement/TagHelpers/StyleTagHelper.cs
```diff
@@ -10,17 +10,25 @@ namespace OrchardCore.ResourceManagement.TagHelpers
     [HtmlTargetElement("style", Attributes = AtAttributeName)]
     public class StyleTagHelper : TagHelper
     {
+        private const string IdAttributeName = "asp-id";
```
deanmarcussen, line 13: I am flat out today, but I will fix this for you later (before merge), so that any attributes are automatically applied, and we don't need these `asp-id` attributes

deanmarcussen (commented):
@Skrypt I took a quick look, and fixed the layout a bit.

The problem with the CI is you've broken both the Setup and Login alternate layouts (they both display blank)

The good news is the mac picked it up automatically (when it was enabled)

src/OrchardCore/OrchardCore.ResourceManagement/TagHelpers/StyleTagHelper.cs
deanmarcussen: I tweaked this so that when normal html attributes are supplied to the tag helper it renders them (any of them).

so no special `asp-id` required, just `id` `media` or any other you might need.

src/OrchardCore.Themes/TheAdmin/Assets/scss/mixins/_grid.scss
```diff
@@ -0,0 +1,21 @@
...
+@mixin make-col-grouping($grouping) {
+    &.col-#{$grouping} {
+        @each $breakpoint in map-keys($grid-breakpoints) {
+            @include make-col-fix(#{$breakpoint});
+            @for $i from 1 through $grid-columns / 2 {
+                @include make-col-fix(#{$breakpoint}-#{$i});
+            }
+        }
+    }
+}
```
jptissot, line 21: eof

src/OrchardCore.Themes/TheAdmin/Views/_ViewImports.cshtml
```diff
@@ -7,3 +7,6 @@
 @addTagHelper *, Microsoft.AspNetCore.Mvc.TagHelpers
 @addTagHelper *, OrchardCore.DisplayManagement
 @addTagHelper *, OrchardCore.ResourceManagement
+
+@using OrchardCore.Entities
+@using OrchardCore.Admin.Models
```
jptissot, line 12: eof

Skrypt:
WTF ?? I never asked for a review from @sfmskywalker !!

sfmskywalker:
Someone called? 😁

sfmskywalker (approved):

Skrypt:
I wish I would have been the one who requested it now I'm asking myself who's been able to use my Github account! 😄

src/OrchardCore/OrchardCore.ResourceManagement/TagHelpers/StyleTagHelper.cs
Skrypt: Thank you !

jptissot:
We should just merge my cypress pr ;) that one works :P

netwavebe:
@Skrypt : FYI, just noticed selected tabs (like site settings) are now white text on a white background.

gulpfile.js
```diff
@@ -323,7 +323,7 @@ function buildJsPipeline(assetGroup, doConcat, doRebuild) {
             "** NOTE: This file is generated by Gulp and should not be edited directly!\n" +
             "** Any changes made directly to this file will be overwritten next time its asset group is processed by Gulp.\n" +
             "*/\n\n"))
-        .pipe(gulpif(generateSourceMaps, sourcemaps.write()))
+        .pipe(gulpif(generateSourceMaps, sourcemaps.write(assetGroup.outputDir)))
```
deanmarcussen, line 326: @Skrypt this change has broken building the media app 

or I think any assets that have a comma delimited listing

i.e.
```
    "inputs": [
      "./node_modules/blueimp-file-upload/js/jquery.fileupload.js",
      "./node_modules/blueimp-file-upload/js/jquery.iframe-transport.js",
      "Assets/js/**/*.js"
    ],
```

what's the change for?

Skrypt: To be able to generate a sourceMap file. This is not supposed to be related.

Skrypt: I think someone added a line in the gulpfile.cs so that the Media app be taken into consideration recently.

Issue #39, opened 2015-09-02
How unload extension, module


GVX111:
IExtensionManager has only 

<pre>
AvailableExtensions
AvailableFeatures
GetExtension
LoadFeatures
EnabledFeatures
</pre> 

How can i disable modules?

Jetski5822:
I think we need an UnloadFeatures method...

Also, it needs to monitor Module.txt files so if it gets renamed, it unloads the assembly.

gvx111:
Also LoadFeatures doesnt work.

Jetski5822:
Really!? that seems weird. I will take a look later this evening.

gvx111:
Did you look at this problem? @Jetski5822

Jetski5822:
Not yet, just finishing up on the event bus work I was looking at, and fixing the latest beta 8 changes.

Should get time tomorrow, sorry for the delay

-----Original Message-----

Jetski5822:
I have just been looking in to this, and decided to port the tests from Orchard over around this, and all go green?

Once the recipe story is done, then the MinimumShellDescriptor will be rewritten to be exactly what it says.

gvx111:
yes it works but maybe dependency not working because I have  Routes : IRouteProvider and Routes are not registered also IDependency. Meybe it must added in ShellContext

sebastienros:
Same as in Orchard1, when some feature change, or are enabled/disabled, the shell has to be reloaded on the next request, i.e. it's like a restart for the tenant. Nothing is implemented yet, neither is tenant management like disabling, restarting, or creating a new one. Will be part of the Orchard.MultiTenancy port.

agriffard:
Orchard.Modules now provides an admi UI that allows you to enable or disable the modules.

PR #7328, opened 2020-10-20
Fix double encoding of non-HTML strings in feed

### LINKED ISSUE #7323: Issues with double encoding in RSS of non-HTML strings
In issue #7261  and PR #7268 the double encoding of CDATA was sorted but unfortunately, I have just noticed that other non-HTML strings in the RSS where WebUtility.HtmlEncode is used double encoding occurs because the WebUtility.HtmlEncode encodes the string which is then encoded again by XDocument - it appears that WebUtility.HtmlEncode is not needed.

So a title like this:
What's happening now!

Becomes this in the RSS (must view source of RSS in Chrome to see):
<title>What&amp;#39;s happening now!</title>

Desired result should be just:
<title>What&#39;s happening now!</title>

scleaver:
Fixes #7323 

WebUtility.HtmlEncode is not required as this results in double encoding. Adds a test as well. Interestingly .Net built-in encoding for XDocument doesn't encode all pre-defined entity references for XML only the strickly illegal onces: https://www.w3schools.com/xml/xml_syntax.asp

### DIFF
--- a/test/OrchardCore.Tests/Modules/OrchardCore.Contents/Feeds/RssFeedTests.cs
+++ b/test/OrchardCore.Tests/Modules/OrchardCore.Contents/Feeds/RssFeedTests.cs
@@ -84,7 +84,7 @@ namespace OrchardCore.Tests.Modules.Contents.Feeds
             // Assert
             var title = feedContext.Response.Items[0].Element.Element("title").ToString();
 
-            Assert.NotEqual("<title>It&amp;#39;s a great title &amp;amp; so much &amp;gt; than anybody&amp;#39;s!</title>", title);
+            // Test to ensure that double encoding of title does not occur and complies with XML requirements
             Assert.Equal("<title>It's a great title &amp; so much &gt; than anybody's!</title>", title);
         }
 
--- a/test/OrchardCore.Tests/Modules/OrchardCore.Contents/Feeds/RssFeedTests.cs
+++ b/test/OrchardCore.Tests/Modules/OrchardCore.Contents/Feeds/RssFeedTests.cs
@@ -55,7 +55,7 @@ namespace OrchardCore.Tests.Modules.Contents.Feeds
         [Theory]
         [InlineData("rss")]
         [InlineData("non rss")]
-        public async Task AvoidDoubleEncodeStrings(string format)
+        public async Task ShouldOnlyHtmlEntityEscapeFeedTitle(string format)
         {
             // Arrange
             var contentManagerMock = new Mock<IContentManager>();
--- a/test/OrchardCore.Tests/Modules/OrchardCore.Contents/Feeds/RssFeedTests.cs
+++ b/test/OrchardCore.Tests/Modules/OrchardCore.Contents/Feeds/RssFeedTests.cs
@@ -52,6 +52,42 @@ namespace OrchardCore.Tests.Modules.Contents.Feeds
             Assert.Equal("<description><![CDATA[<p>The news description goes here ...</p>]]></description>", description);
         }
 
+        [Theory]
+        [InlineData("rss")]
+        [InlineData("non rss")]
+        public async Task AvoidDoubleEncodeStrings(string format)
+        {
+            // Arrange
+            var contentManagerMock = new Mock<IContentManager>();
+            var commonFeedItemBuilder = new CommonFeedItemBuilder(contentManagerMock.Object);
+            var feedContext = CreateFeedContext(format);
+
+            contentManagerMock.SetReturnsDefault(Task.FromResult(new ContentItemMetadata
+            {
+                DisplayRouteValues = new RouteValueDictionary()
+            }));
+
+            contentManagerMock.SetReturnsDefault(Task.FromResult(new BodyAspect
+            {
+                Body = new HtmlString("<p>The news description goes here ...</p>")
+            }));
+
+            feedContext.Builder.AddItem(feedContext, new ContentItem
+            {
+                DisplayText = "It's a great title & so much > than anybody's!",
+                PublishedUtc = DateTime.UtcNow
+            });
+
+            // Act
+            await commonFeedItemBuilder.PopulateAsync(feedContext);
+
+            // Assert
+            var title = feedContext.Response.Items[0].Element.Element("title").ToString();
+
+            Assert.NotEqual("<title>It&amp;#39;s a great title &amp;amp; so much &amp;gt; than anybody&amp;#39;s!</title>", title);
+            Assert.Equal("<title>It's a great title &amp; so much &gt; than anybody's!</title>", title);
+        }
+
         private static FeedContext CreateFeedContext(string format)
         {
             var feedContextMock = new Mock<FeedContext>(Mock.Of<IUpdateModel>(), format);
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Feeds/CommonFeedItemBuilder.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Feeds/CommonFeedItemBuilder.cs
@@ -46,7 +46,7 @@ namespace OrchardCore.Contents.Feeds.Builders
                         guid.Add(url);
                     });
 
-                    feedItem.Element.SetElementValue("title", WebUtility.HtmlEncode(contentItem.DisplayText));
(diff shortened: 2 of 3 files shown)


test/OrchardCore.Tests/Modules/OrchardCore.Contents/Feeds/RssFeedTests.cs
```diff
@@ -52,6 +52,42 @@ public async Task AvoidDoubleEncodeCDATA(string format)
...
+            });
+
+            // Act
+            await commonFeedItemBuilder.PopulateAsync(feedContext);
+
+            // Assert
+            var title = feedContext.Response.Items[0].Element.Element("title").ToString();
+
+            Assert.NotEqual("<title>It&amp;#39;s a great title &amp;amp; so much &amp;gt; than anybody&amp;#39;s!</title>", title);
+            Assert.Equal("<title>It's a great title &amp; so much &gt; than anybody's!</title>", title);
```
hishamco, line 88: Why "&amp;" and "&gt;" are show up, I'm expecting "&" and ">"

scleaver: Sorry @hishamco I don't quite follow what you mean?

hishamco: Why it's not *<title>It's a great title & so much > than anybody's!</title>*

Skrypt: It is HTML encoded which is fine.

hishamco: Oops, it's my bad ;) you are right Jasmin

hishamco (approved):
LGTM :shipit:

Skrypt:
Question. Do we remove the WebUtility.HtmlEncode here because this is previously already saved in the database as HTML encoded?

hishamco:
> I'm just trying to understand why it was like it was beforehand and if we are not potentially breaking anything else.

Good question, let us hold before we merge this one

Skrypt:
What are the use cases of this except for the Blog/List feeds?

hishamco:
That's what I'm trying to do

hishamco:
Just added `dontmerge`  label to avoid merge this accidently by someone else, I will remove it once we are sure that removing the encoding doesn't break anything else

hishamco:
> What are the use cases of this except for the Blog/List feeds?

AFAIK, just in those two places, but I will double check

hishamco:
Seems it was originally authored by Seb in 2016, after 3 years later @deanmarcussen added a title encoding in #3441. It's in your plate Dean ;)

scleaver:
> Question. Do we remove the WebUtility.HtmlEncode here because this is previously already saved in the database as HTML encoded? What would happen here if I'd pass a simple non pre-HtmlEncoded string? Question asked differently : do we assume it will always be HTML?
> 
> I'm just trying to understand why it was like it was beforehand and if we are not potentially breaking anything else.

We are removing WebUtility.HtmlEncoding because you don't need to HTMLEncode strings in this case because XDocument encodes the XML nodes when it outputs the stream AND XDocument only encodes characters that are not compatible with XML. 

So if you take a string like `It's a great title & so much > than anybody's!` then WebUtility.HtmlEncode encodes this to `It&#39;s a great title &amp; so much &gt; than anybody&#39;s!` then when the XDocument outputs the stream it encodes it again and it becomes `It&amp;#39;s a great title &amp;amp; so much &amp;gt; than anybody&amp;#39;s!` - double encoded.

XDocument already encodes illegal characters when it outputs the stream using XML safe encoding which is not the same as HTML encoding. So  `It's a great title & so much > than anybody's!`  becomes `It's a great title &amp; so much &gt; than anybody's!`

NOTE: that you don't have to encode HTML with XDocument either, just encase them in CDATA which is what @hishamco fixed in #7268

There are 5 entities in XML which should be encoded which:

`&lt;` | < | less than
-- | -- | --
`&gt;` | > | greater than
`&amp;` | & | ampersand
`&apos;` | ' | apostrophe
`&quot;` | " | quotation mark

Notice XDocument above does not encode apostrophe and quotation... it seems that the default implementation only encodes strictly illegal characters as a side note.

Skrypt:
Yes, everything you say there is fine. But what I'm asking you guys to look after is if covers all use cases we have in OC. Because, I can remember that this method has been refactored quite some times now. Else, if we have an exception here we will find it later on anyways.

hishamco:
I think we should wait for Dean

hishamco:
Sorry closed accidently when I'm using a mobile

Skrypt:
Yes keep this opened. @sebastienros reviewed it in today's meeting and looked fine. I just want to make sure that we don't fix it a fifth time.

deanmarcussen (approved):
This is a good fix.

Same issue as before where chrome was decoding xml automatically, so it looked like it was unencoded

sebastienros:
I commented on the issue as it wasn't showing double encoding. Yes this PR is fine, the XElement is encoding already. The only thing that is weird is the test on the previous wrong value. A comment what we are testing might be clearer, why would we test two different values, only one is correct. I could test an infinity of wrong values.

Skrypt:
The Unit Test has been added to prevent any changes in that method that would eventually not encode these special chars : 

`&lt;` | < | less than
-- | -- | --
`&gt;` | > | greater than
`&amp;` | & | ampersand
`&apos;` | ' | apostrophe
`&quot;` | " | quotation mark

You are right to say that we test something we don't need but it's there to say : here are the requirement for this method to work properly. If you make any eventual changes that don't pass these test then it will not pass.

hishamco:
@deanmarcussen shall we merge this?

test/OrchardCore.Tests/Modules/OrchardCore.Contents/Feeds/RssFeedTests.cs
```diff
@@ -52,6 +52,42 @@ public async Task AvoidDoubleEncodeCDATA(string format)
             Assert.Equal("<description><![CDATA[<p>The news description goes here ...</p>]]></description>", description);
         }
 
+        [Theory]
+        [InlineData("rss")]
+        [InlineData("non rss")]
+        public async Task AvoidDoubleEncodeStrings(string format)
```
deanmarcussen, line 58:
```suggestion
        public async Task ShouldOnlyHtmlEntityEscapeFeedTitle(string format)
```

test/OrchardCore.Tests/Modules/OrchardCore.Contents/Feeds/RssFeedTests.cs
```diff
@@ -52,6 +52,42 @@ public async Task AvoidDoubleEncodeCDATA(string format)
...
+                PublishedUtc = DateTime.UtcNow
+            });
+
+            // Act
+            await commonFeedItemBuilder.PopulateAsync(feedContext);
+
+            // Assert
+            var title = feedContext.Response.Items[0].Element.Element("title").ToString();
+
+            Assert.NotEqual("<title>It&amp;#39;s a great title &amp;amp; so much &amp;gt; than anybody&amp;#39;s!</title>", title);
```
deanmarcussen, line 87: Remove this `NotEqual`

deanmarcussen (commented):
Yes Seb is right, the test is a little off.

Have made suggestions just to rename it to be specific to the result it is expecting.

hishamco:
@scleaver please consider Dean's comment then we can merge

test/OrchardCore.Tests/Modules/OrchardCore.Contents/Feeds/RssFeedTests.cs
sebastienros: Same, just put a comment explaining what we are trying to prevent

sebastienros (approved):

hishamco:
@scleaver I'm waiting to consider the Dean & Seb comments, so I can merge once you do the changes

scleaver:
Made changes as per comments. Should be good to go.

hishamco:
Thanks a lot @scleaver

PR #7332, opened 2020-10-20
Add classes / id's to buttons in admin to allow easier selection in e2e tests

jptissot:
Add id's and classes to buttons in the admin interface. This will allow us to write better selectors for e2e testing. These selectors will be implemented in https://github.com/jptissot/cypress-orchardcore


src/OrchardCore.Modules/OrchardCore.Shortcodes/Views/Admin/Edit.cshtml
```diff
@@ -120,16 +120,16 @@
     </div>
     <div class="form-group">
         <div class="btn-group">
-            <button type="submit" name="submit" class="btn btn-success">@T["Save"]</button>
+            <button id="btn-save" type="submit" name="submit" class="btn btn-success">@T["Save"]</button>"Save"]</button>
```
hishamco, line 123: Seems there's an issue in few places ;(

hishamco:
What's the purpose of this PR?

agriffard:
> What's the purpose of this PR?

I guess it is to access to buttons when you use Cypress.

Please be careful to not have the same id multiple times on one page and that you also rename it when it is used with a jQuery selector.

hishamco:
Also the typo will break the things ;)

src/OrchardCore.Modules/OrchardCore.Shortcodes/Views/Admin/Edit.cshtml
jptissot: Oops will fix

jptissot: Actually, only in 1 spot ;)

jptissot:
> Please be careful to not have the same id multiple times on one page and that you also rename it when it is used with a jQuery selector.

Yes, I did not test all the pages but most of these have been tested with an html validator. Once this is merged I will create reusable commands for these buttons in cypress

src/OrchardCore.Modules/OrchardCore.Shortcodes/Views/Admin/Create.cshtml
```diff
@@ -112,16 +112,16 @@
 
     <div class="form-group">
         <div class="btn-group">
-            <button type="submit" name="submit" class="btn btn-success">@T["Save"]</button>
+            <button id="btn-save" id="btn-submit" type="submit" name="submit" class="btn btn-success">@T["Save"]</button>
```
jptissot, line 115: duplicate id

src/OrchardCore.Modules/OrchardCore.Shortcodes/Views/Admin/Create.cshtml
```diff
@@ -112,16 +112,16 @@
...
             <button type="button" class="btn btn-success dropdown-toggle dropdown-toggle-split" data-reference="parent" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
                 <span class="sr-only">@T["Toggle Dropdown"]</span>
             </button>
             <div class="dropdown-menu">
-                <button class="dropdown-item" name="submit" type="submit" value="SaveAndContinue">@T["and continue"]</button>
+                <button id="btn-save-continue" class="dropdown-item" name="submit" type="submit" value="SaveAndContinue">@T["and continue"]</button>["and continue"]</button>
```
jptissot, line 120: duplicate text

src/OrchardCore.Modules/OrchardCore.Shortcodes/Views/Admin/Edit.cshtml
```diff
@@ -120,16 +120,16 @@
...
     <div class="form-group">
         <div class="btn-group">
-            <button type="submit" name="submit" class="btn btn-success">@T["Save"]</button>
+            <button id="btn-save" type="submit" name="submit" class="btn btn-success">@T["Save"]</button>
             <button type="button" class="btn btn-success dropdown-toggle dropdown-toggle-split" data-reference="parent" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
                 <span class="sr-only">@T["Toggle Dropdown"]</span>
             </button>
             <div class="dropdown-menu">
-                <button class="dropdown-item" name="submit" type="submit" value="SaveAndContinue">@T["and continue"]</button>
+                <button id="btn-save-continue" class="dropdown-item" name="submit" type="submit" value="SaveAndContinue">@T["and continue"]</button>["and continue"]</button>
```
jptissot, line 128: duplicate text

Skrypt:
Could we not generate the id's on demand through javascript for each element that uses "btn" class instead?

hishamco:
Agree with @Skrypt, we can access them using classes instead, unless there's a value I don't know about!!

jptissot:
The issue with classes is that they can be changed to change the color of a button and then that breaks the selectors. Maybe this won't be an issue here but in past projects, the tests would fail because someone changed a classes to change the look and feel. There is no point of adding the id's with js as you could simply build the selectors to use the same syntax the id builder would use.

Skrypt:
The advantage of using javascript would be to avoid getting duplicate id's in the same page. Here nothing prevents it.

jptissot:
So maybe instead we could add a class or something to the wrapper and use  a convention for buttons (maybe button text or something). `cy.get('.oc-submit-actions').find('button:contains("Submit"))`

I agree that id's are a bit risky in the sense that they need to be unique and it can be hard in OC to see then whole page before rendering.

Skrypt:
Instead of using a id you could also use a data-something attribute simply.

jptissot:
Yes, my preferred approach is to use `data-cy` attributes but your argument sticks as well. Why would we add that only for testing ?

As long as we find a good way to target specific buttons and it's not too flaky. I'm open to anything. I just want to avoid flakiness in tests.

Skrypt:
Go for data-cy.

jptissot:
I think @sebastienros did not like that idea ;)

Skrypt:
I can see why. It would add specific HTML markup just for doing functional tests. That's why I was thinking about using something that is more common and that would not be too intrusive. But, these tests could be ran on any kind of elements so that makes less sense to rely only on role="button" for that.

jptissot:
Maybe using a class would be the least intrusive thing ?  like all save buttons would have a `save-button` class.

Skrypt:
If it's not styling anything let's not use styles to define a functionality. I'd rather use something like data-cy="save" or data-test="save" to make it more global and convince @sebastienros

sebastienros:
Yeah I don't like it ;) I prefer when the ids or classes are used. Or other roles.

jptissot:
I prefer classes to id's because of the reduced risk. Even if those classes are not styling anything.

Skrypt:
ANd I don't like classes because they are meant to style things. So let's use Roles. lol

jptissot:
What do you mean roles ? like role attribute ?

jptissot:
Ok, that sounds fine to me. I will amend this PR using the role attribute.

What about the dynamic values. Can I use an ID there ?

jptissot:
What about the dynamic lists ? Role as well or ID is ok there ? https://github.com/OrchardCMS/OrchardCore/pull/7332/files#diff-0ab32711c2d5b6a2ccd9444337c472d7391778050c052ee45ccf23d873a619f2R64

Skrypt:
<a role="button save @Html.GenerateIdFromName(recipe.Name)">Save</a>

jptissot:
ok I will try this.

jptissot:
It works.. but I get html validator errors when using non standard roles.

`Error: Discarding unrecognized token import from value of attribute role. Browsers ignore any token that is not a defined ARIA non-abstract role.`

Skrypt:
Damn it. I've seen that we can use custom values but seems that they we're wrong.

jptissot:
Why not use a class for styling and for this ? 

Write some sass for the buttons:

```
.publish-button {
  @extend .btn;
  @extend .btn-primary;
}
...
```

Also use the same class to easily select the button using cypress. This would also allow us to be really consistent between pages without having to think about it.

Skrypt:
Here when you use @extend it removes the advantages of using Bootstrap classes directly. The thing is that when you use extend you copy all the style contained in those other class in your own class which creates a lot of redundancy. We use this to override the grid for example but if we need this for every kind of buttons we have I must disagree.

jptissot:
I don't think that is accurate if we extend 1 style instead of two. What will end up happening is that the output css will just add the new class as an alias to the same color. 

```
.btn-primary, .btn-publish {
 ... 
}
```

This seems to be the best alternative to id's or data attributes. The output css is marginally bigger and the html output is the exact same size.

Skrypt:
Why not consider using a taghelper so that we can use a data attribute that would not render on prod?
I mean why change the way we style things because of functional tests?
I would rather use a data attribute for this over any style or id. That's what the data attribute is meant for.
I can see that we don't want it to render all the time ... but using styles and adding extra styles for this.
I mean, if we end up adding a button-republish we will need to add a sass style for functional tests and styles to work ???

jptissot:
I think this tag-helper idea or env idea is not a good move for perf reasons. Why add processing to every request to remove an attribute used for testing. (I might be wrong about this, just a hunch)

My preferred approach is to use a data-cy attribute and just leave it there. In a ReactJS application I wrote last year, we we had a flag that we would pass at build time to strip these attributes so that was acceptable. 

I think here the best approach is to use a specific class for the buttons. We would not be adding styles, just adding an alias to the style that would also easily be targeted by an automated test. 

I am also ok with using an id even if it's more risky.

Skrypt:
At this point it's a question of taste. I can't argue with taste. I don't think it would add any big major perf hit.
The idea is also to strip those unnecessary attributes on prod. If we'd strip these attributes with javascript afterward then we'd add also processing overhead on every requests on the client browser. It's a no win game on either side.

The only thing that doesn't add processing is leaving those attributes there. If we leave them there would it not allow to make it easier for a robot to automate triggering some buttons based on these attribute values? This is mainly why I think it should not be kept on a production website. Of course, there a million other ways to trigger a button from a robot but I think we should at least not make it easier than it is.

As for using an `id` I can't agree because we can't assert they will be unique.

For a `class`, I can't agree unless they would be removed and not required to be styled with aliases.
If we use class then it's more complex to define which one are used for functional tests and needs to be cleaned up.
Same thing goes for an id ...
I see why you guys want to use class, because you want to keep it there. But we have Bootstrap classes and now we want to extend them because we want to use some different naming?

Also, just like we saw that we can't use the role attribute and modify it's **_purpose_** ; we should not use the `id` or `class` ...

That, makes sense to me.

I think you should use `data-cy` and we will think about if we should clean them or not afterward.
If that's the only remaining thing to think about don't block on it just move forward with other things.

In the end, it's just my opinion, you can do what you like. 😉

jptissot:
I vote for classes + extending bootstrap classes. Simple, no processing,  reduced risk.

Skrypt:
Why do you need to extend these classes ?

jptissot:
Fine with that too.

Skrypt:
I'm sorry for being an ass and analyzing everything. I sometimes have a strong opinion.
I may be wrong and right at the same time ... do what you like it has no big incidence in the end.
It is always complicated to make everyone happy or agree on something. 

Everyone has different taste and we can't argue on tastes ...

jptissot:
That's fine man, it's good to be challenged. We usually come up with a better solution together. 

I think in this case, classes are the way to go. Adding classes can also help with custom themes without having to completely rewrite the template. I know we have bootstrap classes but sometimes it could be useful to allow custom admin themes to target some specific elements. 

I will amend the pr to add the short classname you suggested. The selectors in my cypress testing framework will be much cleaner as a result :)

src/OrchardCore.Modules/OrchardCore.AdminMenu/Views/Menu/List.cshtml
```diff
@@ -9,7 +9,7 @@
     </div>
     <div class="form-group col mb-n1">
         <div class="dropdown float-right order-md-1">
-            <a asp-route-action="Create" class="btn btn-primary float-right" role="button">@T["Add Admin Menu"]</a>
+            <a asp-route-action="Create" class="btn btn-primary float-right create">@T["Add Admin Menu"]</a>
```
jptissot, line 12: re-add role here..

jptissot:
I think I might change this up a bit. The class names should probably be prefixed with `btn-` to avoid style clashing / confusion.

Skrypt:
Honestly no one should be doing a `.save {}` class for example.

Skrypt:
By the way Bootstrap uses dashes for when their base components are altered through mixins/functions. Here an example with .btn is .btn-primary which uses the base color classes which are defined with sass vars $primary, $secondary ... and so on. But as you can see they also have .btn which we could say the same than .publish ...

So contrarly to what people might think, it's not about a naming convention but about when a base style is passed in mixins that they use suffixes e.g. "-something".

jptissot:
Cool, I was not aware of how they implemented it in the background. I was going off what I remember about the bem css naming convention.

jptissot:
A bit off topic, but interesting nonetheless:

Harry Roberts identified another key benefit of using a syntax like BEM when he writes about improving developer confidence:

This is the main reason we end up with bloated code bases, full of legacy and unknown CSS that we daren’t touch. We lack the confidence to be able to work with and modify existing styles because we fear the consequences of CSS’ globally operating and leaky nature. Almost all problems with CSS at scale boil down to confidence (or lack thereof): People don’t know what things do any more. People daren’t make changes because they don’t know how far reaching the effects will be.

Likewise, Philip Walton argues that this problem can be fixed if enough developers stick to the principles of BEM:

While 100% predictable code may never be possible, it’s important to understand the trade-offs you make with the conventions you choose. If you follow strict BEM conventions, you will be able to update and add to your CSS in the future with the full confidence that your changes will not have side effects.

From https://css-tricks.com/bem-101/

Skrypt:
Taken from the linked page : 

```css
/* Btn Module with State */
.btn.is-collapsed { }
```

jptissot:
> 
> 
> Taken from the linked page :
> 
> ```css
> /* Btn Module with State */
> .btn.is-collapsed { }
> ```

This is not bem, it's SMACSS but i don't mind it too much. If you don't think it will cause issues, then let's go for it :)

src/OrchardCore.Modules/OrchardCore.Shortcodes/Views/Admin/Create.cshtml
```diff
@@ -112,16 +112,16 @@
...
                 <span class="sr-only">@T["Toggle Dropdown"]</span>
             </button>
             <div class="dropdown-menu">
-                <button class="dropdown-item" name="submit" type="submit" value="SaveAndContinue">@T["and continue"]</button>
+                <button class="dropdown-item save-continue" name="submit" type="submit" value="SaveAndContinue">@T["and continue"]</button>
             </div>
         </div>
 
-        <a class="btn btn-secondary" asp-route-action="Index">@T["Cancel"]</a>
+        <a class="btn btn-secondary cancel" asp-route-action="Index">@T["Cancel"]</a>
```
jptissot, line 124: @Skrypt Should all these cancel buttons have role="button" ? I did not think so because they are links to a page. But maybe for a11y?

Skrypt: Role is an attribute for accessibility so I think we should have it.

jptissot: Ok I will add them all in this pr then.

agriffard:
Please solve the conflicts.

Skrypt (approved):

Skrypt:
I merged the dev branch and fixed the merge issues.

PR #7336, opened 2020-10-20
Make content items action bar sticky on scrolling

hishamco:
Sticky


Skrypt:
I don't want to push it but could we not also have the table header row stacking right underneat that as a second sticky element?

hishamco:
Ok, let me try this one too ...

hishamco:
You know what, before I begin I realize the `action-bar` can't be used here because the top of both are different

Skrypt:
filter-bar

hishamco:
My question is we need a better way instead of setting `top` every time, we need something like a stack layout

hishamco:
ContentItemsSticky

Skrypt:
Looks good. Will review when I get time.

src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsAdminList-Create.cshtml
```diff
@@ -1,20 +1,24 @@
 @model ContentOptionsViewModel
-<div class="form-group">
+<div class="form-group d-inline-flex float-right mb-0">
     <div class="btn-group">
         @if (Model.CreatableTypes.Any())
         {
             @if (Model.CreatableTypes.Count == 1)
             {
-                <a class="btn btn-sm btn-success" href="@Url.RouteUrl(new { area = "OrchardCore.Contents", controller = "Admin", action = "Create", id = Model.CreatableTypes.First().Value, returnUrl = FullRequestPath })">@T["New {0}", Model.CreatableTypes.First().Text]</a>
+                <a class="btn btn-sm btn-secondary" href="@Url.RouteUrl(new { area = "OrchardCore.Contents", controller = "Admin", action = "Create", id = Model.CreatableTypes.First().Value, returnUrl = FullRequestPath })">@T["New {0}", Model.CreatableTypes.First().Text]</a>
```
Skrypt, line 8: Why are you changing this button color?

hishamco: > Why are you changing this button color?

To make it suited with the filter color scheme

> No need to do 5 different PR's just change those that are easy to do for now in all other places you can find.

You 're right, after this PR, coz we cover all the cases, I can merge content types and content parts into a single PR, but as always Seb said make small manageable PRs, so it could be easy to review ;)

agriffard:
Please resolve css conflicts (merge dev)

hishamco:
This file make Fork client crash during merge ;)

hishamco:
Resolve the conflict via GitHub

Skrypt:
There is some weird issue going on with the menu dropdowns while using position: sticky. We need to find a solution to that if we want to merge.

hishamco:
Can you provide a screenshot? If it's a z-index issue I can increase the value in action-bar

hishamco:
I'm able to reproduce the issue, the fix is coming soon ...

Menu

Skrypt:
Here is one : 

image

I tried with a z-index of 50 and it fixed this issue since the left menu is at 1000. I think we should try with a value at least under 1000.

hishamco:
> I think we should try with a value at least under 1000

I think 1040 is fine, it's lower the `action-bar` z-index and higher than 1030 for the nav-bar

hishamco:
Nah, there's another one with the New menu, let me have a try

hishamco:
Seems z-index 1070 for a menu is fine too

hishamco:
NewMenu

Skrypt:
If that works ... 😄

hishamco:
> Make sure the top navbar has a higher z-index that the left-menu.

I think that already calculated with `z-index: calc($zindex-sticky + 1);`

hishamco:
Seems it looks good at my side, have a try if you have a time

NewMenu

hishamco:
BTW I notice another UI bug from the screenshot, the site title disappear when the menu is collapse, I will fix this in another PR

Skrypt:
The site name disapear intentionally. I can clearly see that the menu gets on top of the top navbar which is not good.

hishamco:
> I can clearly see that the menu gets on top of the top navbar which is not good.

I notice that is on top of the shadow if I'm not wrong, but I TRUST this line `z-index: calc($zindex-sticky + 1);` ;)

hishamco:
Seems it's overridden somewhere else

ZIndex

Skrypt:
```
z-index: calc(#{$zindex-sticky} + 1);
```

hishamco:
UpdatedNewMenu

Skrypt:
Don't merge yet I want to make more tests. I don't like what happens when we have 2 bars stacked. The dropdown menus are acting really badly getting underneat the other sticky content. I will try to put some more time reviewing this to make sure everything is fine.

hishamco:
No problem, but dropdown issue got fixed as shown [here](https://github.com/OrchardCMS/OrchardCore/pull/7336#issuecomment-716276701)

Skrypt:
Try the dropdown menus from the second bar when your windows height is small and that they open on top instead of down.

hishamco:
Good catch that will raise another UI issue, but here like a deadlock ;) the drop down on the first should appear on top of the second bar, similarly the drop down on the bottom should appear on top of the first bar especially when the menu opened from bottom to top. Simple solution for this to give the drop down a higher z-index so it will be the top most all the time

Skrypt (approved):

PR #7345, opened 2020-10-21
Media gallery : background-color

Skrypt:
Sets a background-color to the media gallery based on main $body-bg.
This will prevent displaying the notifier that is underneat.
We use also a background-color on the GraphiQL explorer page for the same reason.
The INotifier zone should be displayed differently on these pages.
Removing duplicate classes in favor of importing Bootstrap variables and functions.

Sans titre

Found the issue while testing UserId migration branch.


src/OrchardCore.Modules/OrchardCore.Media/Assets/scss/_variables.scss
```diff
@@ -1,26 +1,6 @@
...
-$gray-900: #212529 !default;
-
-$blue: #007bff !default;
-
-$primary: $blue;
-$secondary: $gray-600;
-
-$zindex-fixed: 1030 !default;
+@import "../../node_modules/bootstrap/scss/functions";
+@import "../../node_modules/bootstrap/scss/variables";
```
jptissot, line 2: The only risk I can see with this is that if we modify the bootstrap variables in the admin there, the modifications will not get used here. This assumes we will always use the stock bootstrap variables.

Skrypt: No it's just using the bootstrap variable declaration file. To get them in the Media gallery style. See what it transpile into. We don't add any vars just use them. In fact we don't want to modify them just use them for consistency.

Skrypt: TheAdmin theme uses the same technique of importing the Bootstrap variables. When you just want to use the function, vars and mixins you can just import those without importing all the bootstrap classes. So, if you want to customize TheAdmin theme you can change colors by overriding vars just like they do in Bootswatch.

Skrypt: The only way we could affect these variables from the TheAdmin theme would be to add a _variable file import from the TheAdmin theme in here. I'm working on TheAdminDark theme on a different PR maybe we could do that. But it would mean that this module would be now dependent on TheAdmin theme which probably should not.

We need to be able to generate these styles per admin theme if that's what you are suggesting and right now we could'nt and we still can't

Skrypt: Food for thoughts on a different PR.

jptissot: @Skrypt I think there is a way to make a custom theme as a separate project / folder. We could take parts of what is in TheAdmin and make it available to any other projects that references it. This way, our admin interfaces from all the different modules could easily reference the admin varialbes and other functions.

Skrypt: Yeah let's move this discussion over to : https://github.com/OrchardCMS/OrchardCore/pull/7326

PR #7354, opened 2020-10-22
Use camel case for content picker field search results

### LINKED ISSUE #7273: Problem with content menu item content picker
EDIT: OC version is 1.0.0-rc2-14774
EDIT 2: I forgot to mention that this had been working at one point, back when I initially created my menu some number of weeks ago. It was only yesterday that I noticed this issue when I went to add a new menu item.

I'm not sure if this is a bug or something that's gone wrong in my test site, but I haven't been able to solve the problem and I'm unfamiliar with Vue.

Basically, there seems to be a problem with the content picker field when I create a new content menu item. All I see is "not published" for each item.

image

I've debugged DefaultContentPickerResultProvider and confirmed that it is returning the results that I expect, and that everything in the result list has been published (ContentPickerResult.HasPublished = true).

I've also debugged vue-multiselect-wrapper.js, and it seems like I'm getting the correct JSON (again, everything has HasPublished set to true).

But for some reason, the dropdown doesn't seem to be rendering correctly.

I tried a fresh OC site using the blog recipe, and it seems to be working just fine there.

image

I'm just looking for any tips on how to figure out what is going on. Thanks!

rjpowers10:
Fixes #7273

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Controllers/ContentPickerAdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Controllers/ContentPickerAdminController.cs
@@ -2,6 +2,8 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
 using OrchardCore.Admin;
 using OrchardCore.ContentFields.Settings;
 using OrchardCore.ContentFields.ViewModels;
@@ -57,7 +59,10 @@ namespace OrchardCore.ContentFields.Controllers
                 PartFieldDefinition = partFieldDefinition
             });
 
-            return new ObjectResult(results.Select(r => new VueMultiselectItemViewModel() { Id = r.ContentItemId, DisplayText = r.DisplayText, HasPublished = r.HasPublished }));
+            return new ObjectResult(
+                JsonConvert.SerializeObject(
+                    results.Select(r => new VueMultiselectItemViewModel() { Id = r.ContentItemId, DisplayText = r.DisplayText, HasPublished = r.HasPublished }),
+                    new JsonSerializerSettings { ContractResolver = new CamelCasePropertyNamesContractResolver() }));
         }
     }
 }


jptissot:
I suggest to instead remove the camelCasing and adjust the VueJs template to use PascalCase. 

You will also need to update the LocalizationSetContentPickerField's template

deanmarcussen:
Something about this issue doesn't seem right.

OOB ASP.Net Core serialization returns camelCase.

And an OOB OrchardCore setup correctly returns camelCase from this controller already.

Is it possible you have changed the default serializer @rjpowers10  ?

rjpowers10:
> Something about this issue doesn't seem right.

Yes, I agree. Something is off here and I don't have an explanation for it yet. This may be treating the symptom and not the cause.

> OOB ASP.Net Core serialization returns camelCase.
> 
> And an OOB OrchardCore setup correctly returns camelCase from this controller already.
> 

Now I think I'm going crazy. I had thought that I confirmed that the OOB site was returning pascal case (and was somehow still working despite that), but sure enough, when I look at it again, I see that it is indeed returning camel case.

> Is it possible you have changed the default serializer @rjpowers10 ?

I suppose it is possible, though I don't recall doing this. At least not on purpose.

deanmarcussen:
I think you are indeed treating the sympton.

Have you added something like this somewhere?
```
services.AddMvc()
        .AddJsonOptions(options => options.SerializerSettings.ContractResolver = new DefaultContractResolver());
```

rjpowers10:
@deanmarcussen 

Oh no! I've been betrayed by one of my own team members...look at what I found in our code:

```
services.AddControllers()
    .AddXmlSerializerFormatters()
    .AddNewtonsoftJson(options =>
    {
        options.UseMemberCasing();
    });
```

I removed this and everything started working. As for my PR, I'll don't feel strongly about it either way. I could maybe see an argument for explicitly using camel case here, to avoid problems like this in the future, but I also won't feel bad it you just want to call this "user error" and drop the PR.

I'm sorry for wasting everyone's time on this, but I really appreciate the help!

jtkech:
@rjpowers10 so i assume we can close it ;) feel free to re-open an issue it if you still have any one

Issue #7128, opened 2020-09-23
Full Text Index for Widgets/Bag Items


deanmarcussen:
For pages that are built with a BagPart or a Widget what would be the best approach to include the content of these Contained Items in the FullTextIndex?

I have a rudimentary BodyAspect handler working which looks for all HtmlFields inside the BagPart content items, but it has limitations.

Principally with this rudimentary approach, only HtmlFields are covered, so titles / captions for media fields (which in this case are textfields, and I could include, but haven't at this stage).

The two options I'm seeing is
- What I have which has limitations 
- Calling BuildDisplay on all the ContentItems contained, and executing these shapes into an HtmlContentBuilder

BuildDisplay is going to potentially end up giving me a lot of html tags / theme related output, but it was also mentioned that we're stripping html tags before they go into the lucene index?

The key issue here is that as the pages are built from many contained content items, they don't have a standard html body to build an aspect from.

/cc @Skrypt

Skrypt:
What you mean by limitations?

deanmarcussen:
I think I probably need to be calling recursive fulltextaspects on each content item in the bags so I can take advantage of the liquid templating for custom index fields on each of the content types inside the bag.

The limitation is that the bags have more than just html fields, sometimes nested bags, but I think recursive fulltextaspects will sort it.

Will see tomorrow

Skrypt:
I thought you meant that you we're hitting some performance bottleneck but yes using the FullTextAspect on each content items inside the bags should be the way to go. This way it is configurable and each content item should already be indexed unless it's broken. @hswami did that part where we index the bagparts. I will let you work on it and take a look when you push some draft PR.

Skrypt:
Also after thinking about this twice. There is probably no need to create an indexation on something already indexed. Probably what you need is a proper custom ElasticSearch Query unless we are missing some parts which needs to be added to the index.

deanmarcussen:
ok, so got this working.

Yes the BagParts are indexed, but not as part of the `FullText` index.

So it's either add a lot of fields to the default search index settings for all the bag parts or ...

Have a liquid filter for the `FullTextAspect` so that in the `CustomFullText` settings we can do
```
{% for contentItem in Model.Content.BagPart.Contentitems %}
  {{ contentItem | full_text_aspect }}
{% endfor %}
```

hswami:
[...]

PR #7357, opened 2020-10-22
Add full_text_aspect liquid filter

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/7128

Adds a `full_text_aspect` filter, which can be used in the `FullTextSettings` to allow Widgets or Bags items contribute towards the full text search index.

Usage 
```
{{ Model.Content.FlowPart.Widgets | full_text_aspect }}

or 

{% for contentItem in Model.Content.FlowPart.Widgets %}
  {{ contentItem | full_text_aspect }}
{% endfor %}
```

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Handlers/FullTextAspectContentHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Handlers/FullTextAspectContentHandler.cs
@@ -16,13 +16,13 @@ namespace OrchardCore.Contents.Handlers
     /// <summary>
     /// Provides the content for FullTextAspect based on FullTextAspectSettings
     /// </summary>
-    public class FullTextAspectSettingsHandler : ContentHandlerBase
+    public class FullTextAspectContentHandler : ContentHandlerBase
     {
         private readonly IContentDefinitionManager _contentDefinitionManager;
         private readonly ILiquidTemplateManager _liquidTemplateManager;
         private readonly IServiceProvider _serviceProvider;
 
-        public FullTextAspectSettingsHandler(
+        public FullTextAspectContentHandler(
             IContentDefinitionManager contentDefinitionManager,
             ILiquidTemplateManager liquidTemplateManager,
             IServiceProvider serviceProvider
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Indexing/FullTextContentIndexHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Indexing/FullTextContentIndexHandler.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Threading.Tasks;
 using OrchardCore.ContentManagement;
 using OrchardCore.ContentManagement.Models;
@@ -21,10 +22,13 @@ namespace OrchardCore.Contents.Indexing
             // Index each segment as a new value to prevent from allocation a new string
             foreach (var segment in result.Segments)
             {
-                context.DocumentIndex.Set(
-                    IndexingConstants.FullTextKey,
-                    segment,
-                    DocumentIndexOptions.Analyze | DocumentIndexOptions.Sanitize);
+                if (!String.IsNullOrEmpty(segment))
+                {
+                    context.DocumentIndex.Set(
+                        IndexingConstants.FullTextKey,
+                        segment,
+                        DocumentIndexOptions.Analyze | DocumentIndexOptions.Sanitize);
+                }
             }
         }
     }
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Liquid/BuildDisplayFilter.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Liquid/BuildDisplayFilter.cs
@@ -45,9 +45,19 @@ namespace OrchardCore.Contents.Liquid
                 throw new ArgumentException("Services missing while invoking 'shape_build_display'");
             }
 
+            var serviceProvider = (IServiceProvider)services;
+
+            var buildDisplayRecursionHelper = serviceProvider.GetRequiredService<BuildDisplayRecursionHelper>();
+
+            // When {{ Model.ContentItem | shape_build_display | shape_render }} is called prevent recursion.
+            if (buildDisplayRecursionHelper.IsRecursive(contentItem))
+            {
+                return new ValueTask<FluidValue>(NilValue.Instance);
+            }
+
             var displayType = arguments["type"].Or(arguments.At(0)).ToStringValue();
-            var displayManager = ((IServiceProvider)services).GetRequiredService<IContentItemDisplayManager>();
-            var updateModelAccessor = ((IServiceProvider)services).GetRequiredService<IUpdateModelAccessor>();
+            var displayManager = serviceProvider.GetRequiredService<IContentItemDisplayManager>();
+            var updateModelAccessor = serviceProvider.GetRequiredService<IUpdateModelAccessor>();
 
             var task = displayManager.BuildDisplayAsync(contentItem, updateModelAccessor.ModelUpdater, displayType);
             if (task.IsCompletedSuccessfully)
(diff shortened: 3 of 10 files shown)


src/OrchardCore.Modules/OrchardCore.Contents/Indexing/FullTextContentIndexHandler.cs
```diff
@@ -21,10 +22,13 @@ public async Task BuildIndexAsync(BuildIndexContext context)
             // Index each segment as a new value to prevent from allocation a new string
             foreach (var segment in result.Segments)
             {
-                context.DocumentIndex.Set(
-                    IndexingConstants.FullTextKey,
-                    segment,
-                    DocumentIndexOptions.Analyze | DocumentIndexOptions.Sanitize);
+                if (!String.IsNullOrEmpty(segment))
```
deanmarcussen, line 25: I added this check, because by default Widgets often don't have display text, or a body aspect, so many of the segments were returning empty, and it seemed worthwhile to not place empty strings into the index.

src/OrchardCore.Modules/OrchardCore.Contents/Handlers/FullTextAspectContentHandler.cs
```diff
@@ -16,13 +16,13 @@ namespace OrchardCore.Contents.Handlers
     /// <summary>
     /// Provides the content for FullTextAspect based on FullTextAspectSettings
     /// </summary>
-    public class FullTextAspectSettingsHandler : ContentHandlerBase
+    public class FullTextAspectContentHandler : ContentHandlerBase
```
deanmarcussen, line 19: Just renamed because I couldn't find it to begin with.

src/OrchardCore.Modules/OrchardCore.Contents/Liquid/FullTextAspectFilter.cs
```diff
@@ -0,0 +1,86 @@
...
+
+                var aspects = new List<FullTextAspect>();
+
+                foreach(var contentItem in contentItems)
+                {
+                    aspects.Add(await contentManager.PopulateAspectAsync<FullTextAspect>(contentItem));
+                }
+                
+                // When returning segments seperate them so multiple segments are indexed individually.
+                return new ArrayValue(aspects.SelectMany(x => x.Segments).Where(x => !String.IsNullOrEmpty(x)).Select(x => new StringValue(x + ' ')));
```
deanmarcussen, line 51: Seemed the best way to make sure that multiple widgets had their values seperated.
Might be better ways?

sebastienros:
Potential stack overflow to fix

deanmarcussen:
@jtkech might ask you to take a look at this, as I know you did a shape recursion pr a while back.

Here I have to use the `ContentItemId` to evaluate against as there is too much `JObject.ToObject` to use object equivalency.

I also found it possible to stackoverflow a liquid `Content` template, by recursively building displays.

Could not find a way to break it with any of the other helpers.

Have not interfaced them, as they are really intended to be used internally.

jtkech:
@deanmarcussen 

> @jtkech might ask you to take a look at this, as I know you did a shape recursion pr a while back.

Okay, i will take a look this night

sebastienros (approved):

src/OrchardCore.Modules/OrchardCore.Contents/Liquid/BuildDisplayFilter.cs
```diff
@@ -45,9 +45,19 @@ static async ValueTask<FluidValue> Awaited(Task<IShape> task)
...
             }
 
+            var serviceProvider = (IServiceProvider)services;
+
+            var buildDisplayRecursionHelper = serviceProvider.GetRequiredService<IContentItemRecursionHelper<BuildDisplayFilter>>();
+
+            // When {{ Model.ContentItem | shape_build_display | shape_render }} is called prevent recursion.
+            if (buildDisplayRecursionHelper.IsRecursive(contentItem))
+            {
+                return new ValueTask<FluidValue>(NilValue.Instance);
```
jtkech, line 55: Are we sure here that it can't be called more than once, not necessarily through the same model / shape, but on the same item, this in a given request scope?

jtkech:
@deanmarcussen 

Too late, already merged ;) I still did a little review

Yes i did some recursions checking, as i remember it was to prevent a shape / model to render itself too many times, by comparing by ref the passed model to the current one, max recursions being equal to 3. The use case is

    {{ Model | shape_render }}

I still did a little review, the main remark is to register a generic helper

    services.AddScoped(typeof(IContentItemRecursionHelper<>), typeof(ContentItemRecursionHelper<>));

One question, the recursion helper is also used in the `BuildDisplayFilter`, are we sure that we will never need to render more than once, not necessarily through the same model / shape, but the same content item? This in a given rq scope.

src/OrchardCore.Modules/OrchardCore.Contents/Startup.cs
```diff
@@ -138,6 +138,10 @@ public override void ConfigureServices(IServiceCollection services)
             services.AddLiquidFilter<ContentItemFilter>("content_item_id");
             services.AddLiquidFilter<DisplayTextFilter>("display_text");
             services.AddLiquidFilter<DisplayUrlFilter>("display_url");
+            services.AddLiquidFilter<FullTextFilter>("full_text");
+
+            services.AddScoped<IContentItemRecursionHelper<FullTextFilter>, ContentItemRecursionHelper<FullTextFilter>>();
+            services.AddScoped<IContentItemRecursionHelper<BuildDisplayFilter>, ContentItemRecursionHelper<BuildDisplayFilter>>();
```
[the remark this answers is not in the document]
deanmarcussen: I like that @jtkech didn't even know about it.

Will do some docs shortly, so will change that then.

src/OrchardCore.Modules/OrchardCore.Contents/Liquid/ContentItemRecursionHelper.cs
```diff
@@ -0,0 +1,37 @@
...
+    public class ContentItemRecursionHelper<T> : IContentItemRecursionHelper<T> where T : ILiquidFilter
+    {
+        private HashSet<string> _contentItemIds = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+        /// <inheritdocs />
+        public bool IsRecursive(ContentItem contentItem)
+        {
+            if (_contentItemIds.Contains(contentItem.ContentItemId))
+            {
+                return true;
```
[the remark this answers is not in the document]
deanmarcussen: Yes, I couldn't because there is too much `JObject.ToObject<ContentItem>` which breaks by ref

deanmarcussen:
> One question, the recursion helper is also used in the `BuildDisplayFilter`, are we sure that we will never need to render more than once, not necessarily through the same model / shape, but the same content item? This in a given rq scope.

Tricky @jtkech , in a single request scope if you accidentally call `| shape_build_display | shape_render` on the same content item as the template you are in, you'll always end up back in that template.

So preventing that.

But it is possible you might want to render items twice.
Perhaps in a different display mode. 

I have done that sometimes for a list view and a tile view... But in Razor.

Hmm I will think if there is a better way to keep it safe, but not preventing scenarios either.

src/OrchardCore.Modules/OrchardCore.Contents/Liquid/ContentItemRecursionHelper.cs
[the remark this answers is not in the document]
jtkech: Cool then

jtkech:
@deanmarcussen Yes i meant in the same request scope, so now in the same template context, but from 2 different models / shapes but displaying the same content item, e.g. in 2 different places, yes maybe in 2 display modes.

But that's okay for me ;)

Otherwise maybe just a max recursion counter with a max recursion value

Hmm, one solution would be to also pass to your helper the current `Model` (accessible by any liquid filter) and then in your helper having a set of `ContentItemId` but per model e.g. with a dictionary of `<object, HashSet<string>>`, hmm if the model can be null maybe with a null key if it is possible. Then you could avoid the 1st recurssion under the same model.

PR #7363, opened 2020-10-23
Update Nuget packages

agriffard:
HtmlSanitizer 5.0.355
NodaTime 3.0.3


mariojsnunes:
Is the new YesSql version ready to be updated as well? There is the new "IsIn" feature from a few days ago that I would like to start using :)

agriffard:
> Is the new YesSql version ready to be updated as well? There is the new "IsIn" feature from a few days ago that I would like to start using :)

See #7199

PR #7369, opened 2020-10-24
Make filter bar in content parts sticky on top while scrolling

hishamco:
ContentPartsSticky


hishamco:
Just a little note to myself, we need to find away to make filter bar or action bar has a certain `top` property when appear once in the page, @Skrypt if you have an idea please shoot it to me ;)

PR #7379, opened 2020-10-24
Improve Create Module App Guide

davidhayden:
Changed the guide to align with the sample app and module in the repository as well as improved clarity and grammar.

### DIFF
--- a/src/docs/guides/create-modular-application-mvc/README.md
+++ b/src/docs/guides/create-modular-application-mvc/README.md
@@ -2,10 +2,10 @@
 
 ## What you will build
 
-You will build a modular ASP.NET Core MVC web application similar to the sample "Hello World" application included with Orchard Core. It includes a web application and a module. The web application provides the layout while the module registers a route and responds to homepage requests. You can refer to the following projects in Orchard Core for more information.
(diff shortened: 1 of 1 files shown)


src/docs/guides/create-modular-application-mvc/README.md
```diff
@@ -2,83 +2,94 @@
 
 ## What you will build
 
-You will build an application that is made of modules. The module will provide a Controller and a View while the Layout will
-be provided by the main application project.
+You will build a modular ASP.NET Core MVC web application similar to the sample "Hello World" application included with Orchard Core. It includes a web application and a module. The web application provides the layout while the module registers a route and responds to homepage requests. You can refer to the following projects in Orchard Core for more information.
+
+- [OrchardCore.Mvc.Web](../../../OrchardCore.Mvc.Web)
```
deanmarcussen, line 7: I don't think these will work well on mkdocs because there are no readme.md docs files in the root of these folders.

It might be better just to link to github perhaps?

davidhayden: I didn't realize that. Thank you.

If that's the case, I can only go as far as linking to <https://github.com/OrchardCMS/OrchardCore>, and the reader will need to navigate appropriately.

Is that what you're thinking?

deanmarcussen: Yes, you're probably right a link to https://github.com/OrchardCMS/OrchardCore might be as much as we can do here - otherwise I guess we're pointing at branches which should be up to the reader to select.

PR #7380, opened 2020-10-24
Add support for sending email attachments

mariojsnunes:
Adds a way to add email attachments via the SmtpService.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Email/Services/SmtpService.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Email/Services/SmtpService.cs
@@ -128,6 +128,23 @@ namespace OrchardCore.Email.Services
                 body.TextBody = message.Body;
             }
 
+            if (message.Attachments != null)
+            {
+                foreach (var attachment in message.Attachments)
+                {
+                    if (attachment.Stream == null)
+                    {
+                        // Gets the file from the File System. The filename must be the system path to the file. "App_Data/tenants.json" works.
+                        body.Attachments.Add(attachment.Filename);
+                    }
+                    else
+                    {
+                        // Gets the file from the Stream
+                        body.Attachments.Add(attachment.Filename, attachment.Stream);
+                    }
+                }
+            }
+
             mimeMessage.Body = body.ToMessageBody();
 
             return mimeMessage;
--- a/src/OrchardCore.Modules/OrchardCore.Email/Services/SmtpService.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Email/Services/SmtpService.cs
@@ -132,14 +132,9 @@ namespace OrchardCore.Email.Services
             {
                 foreach (var attachment in message.Attachments)
                 {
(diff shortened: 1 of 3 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=7380) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=7380) before we can accept your contribution.<br/><br/>:x: mariojsnunes [sign now](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=7380)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OrchardCMS/OrchardCore?pullRequest=7380) it.</sub>

hishamco:
Thanks @mariojsnunes for your contribution, I already started at the past to support this feature, but IMHO I suggest to defer this feature after `1.0.0` release, because there 're many things and consideration should we care about, such as attachment number, size, also security consideration which is very important

mariojsnunes:
@hishamco thanks for that. Do you have a list of considerations we should take? I would like to understand them, and I could keep work on that since I need this for my project.
On a side note, since this only allows adding attachments programmatically, and there is no UI for it, is it still an issue?

hishamco:
> Do you have a list of considerations we should take?

This is an open question for core devs and community, but I will try to share mine ;)

> On a side note, since this only allows adding attachments programmatically, and there is no UI for it, is it still an issue?

That's one of the questions that I will ask, I don't think there is a problem at all, the Email module should provide a necessary services that can be used in Orchard CMS or any other web application that using ASP.NET Core, but in case of OC, the UI is flexible and extendable for adding this if it's needed

src/OrchardCore.Modules/OrchardCore.Email/Services/SmtpService.cs
```diff
@@ -128,6 +128,23 @@ private MimeMessage FromMailMessage(MailMessage message)
                 body.TextBody = message.Body;
             }
 
+            if (message.Attachments != null)
+            {
+                foreach (var attachment in message.Attachments)
+                {
+                    if (attachment.Stream == null)
+                    {
+                        // Gets the file from the File System. The filename must be the system path to the file. "App_Data/tenants.json" works.
```
deanmarcussen, line 137: I don't think we should support this. Too much security risk. What if you email the `appsettings.json` file with the sql server passwords in it?

At some point, someone will want to add ui to this, and then it will be compromised.

I think setting via a `Stream` only is fine. 

It then becomes the callers responsibility to resolve a stream correctly, which would mean when a workflow feature gets added, that feature is responsible for delivering a safe stream.

Should it also have a max size setting?
Probably, but you might have to read the entire stream to get the length.
Most mail providers will set their own limit, so possibly can be managed there.

PR #7381, opened 2020-10-25
Update menu item according to its  current type definition

### LINKED ISSUE #7355: LinkMenuItem text field cannot be saved
Added new Text Field "IconClass" to LinkMenuItem but editing the text in field it can not be saved even after publishing the changes.

using Orchard Core Package
<PackageReference Include="OrchardCore.Application.Cms.Targets" Version="1.0.0-rc2-14969" />

jtkech:
Fixes #7355


Skrypt:
What does it fix visually?

jtkech:
Repro with the blog theme

- Through the admin UI, add a text field to the `LinkMenuItem` type definition

- Edit an **existing** menu item in the current main menu, fill the new text field, then publish

- Edit again the same existing menu item, **you see that it was not persisted**

PR #7383, opened 2020-10-25
Better responsive behaviour for all editors of NumericField

netwavebe:
Since row has `margin-right` and `margin-left -15px` and col has `padding-right` and `padding-left 15px` an element should not contain both classes. Columns should always be child elements of a row.

`col-sm-3` was changed to `col-md-6 col-lg-4` to make it usable on small screens.

The hint was moved outside of the column so it can be full width.


agriffard:
Thank you.
If you can attach some before/after screenshots when you are making visual changes, it will help us.

netwavebe:
Sure Antoine!

Before:
image

After:
image

Before:
image

After:
image

The margin I indicated with the red lines was due to the row & columns being set on the same element.

PR #7384, opened 2020-10-25
Better responsive behaviour for DateField

netwavebe:
Since row has margin-right and margin-left -15px and col has padding-right and padding-left 15px an element should not contain both classes. Columns should always be child elements of a row.

The hint was moved outside of the column so it can be full width.


netwavebe:
Some before/after screenshots.

Before:
image

After:
image

Before:
image

After:
image

hishamco:
@deanmarcussen I think you already did something similar if it's the same thing for alignment

PR #7392, opened 2020-10-26
Make recipes action bar sticky while scrolling

hishamco:


hishamco:
We can add any other pages that may need this class to stick the action bar on top in this PR

PR #7393, opened 2020-10-26
Document Options and Compression

jtkech:
**Compression**

- Redis latency and throughput depend on the number of keys, keys sizes, values sizes. I skip the details but generally it's better to keep a value size under 100Kb, maybe a few Mb if not read / write too frequently as with our multi level cache.

    Let say an autoroute takes 300 bytes (as seen in my tests), for a given tenant 100 000 items will end up to an `AutorouteDocument` of 30Mb. So i tried to compress the data with `GZip` which is very effective on this kind of data, the size is divided by around 6, and the reduction factor increases as the document increases.
    
- So here i added a **config option to compress or not the data of a given documen**t type in the distributed cache. And a **marker interface to set this option to true if not already configured** for this document type.

    **So here i added this interface to the `AutorouteDocument`** (and to the `LocalizationDocument`), fragmenting the `AutorouteDocument` in one cache entry per autoroute would end up in too many keys. **So here, for a given tenant, 100 000 items ends up to an `AutorouteDocument` of 30Mb but only 5Mb in the distributed cache** if compressed.

    Note: Here i didn't talk about pipelining, what i can say is that we don't use it explicitly but it is naturally done when multiple async http requests on one or multiple OC instances are accessing the cache. Then if we want to support, for a given tenant, millions of items, there are other solutions on redis side as e.g. redis cluster for auto sharding.

**Atomicity**

- On concurrent updates, we try to be thread safe and preserve the document consistency (in some race condition we may explicitly throw), and keep it in sync between the store (if not volatile) and the shared cache, but some updates may be missed, it would involve other mechanisms to lock or allow a shared editing, so, at least for now, that's okay if a race condition only concerns one document at a time.

    **But with the `AutorouteDocument`** (and the `LocalizationDocument`), even if the concurrent updates are all related to different items, the same `AutorouteDocument` is updated concurrently, **so here atomicity is more important**, notice that before, for the same reason, we were using a local concurrent dictionary.

- By nature an individual Redis operation is atomic, but when reading some cached data, doing some updates, and then writting to the redis cache, the whole operation is not atomic.

    **So here i plan to suggest (not yet done) the 1st usage of our distributed lock when updating the `AutorouteDocument`** (and the `LocalizationDocument`). I think our `RedisLock` is safe to use, to acquire a lock we pass a timeout and an expire time, if an instance fails to release a lock, it is auto released on the redis side after the expire time.

    I also thought about content definitions, but editing any content type is like editing the same single document.


mariojsnunes:
Hello, great work!
Have you considered using Brotli instead of GZip?
[https://devblogs.microsoft.com/dotnet/introducing-support-for-brotli-compression/](https://devblogs.microsoft.com/dotnet/introducing-support-for-brotli-compression/)

sebastienros:
How much does it take when using MessagePack? As I assume it would be much faster for most structures. Since you are adding interfaces to define how something can be serialized, maybe we should use this to use MessagePack too, and fallback to json serialization for structures that don't work. We could even implement ISerializable if that helps. For instance let's assume an autoroute is just `url|id;url|id` ... don't need to repeat the property names, the format can be very simple. If it is critical in terms of perf.

Then maybe make this pipeline extensible such that someone could write their own serialization logic to optimize for specific cases. It's fine if the default one would come with custom marker interfaces, but for each type a module could take over and do its own logic.

Another optimization if memory utilization, and maybe being able to split the serialized content in multiple blocks < 80KB might be important.

sebastienros:
and MessagePack even has a mode to compress, and to split the buffer in blocks < 80KB as I mentioned: https://github.com/neuecc/MessagePack-CSharp#lz4-compression

sebastienros:
Also, if some types can't be serialized with MsgPack, then it can be extended with custom formatters. For instance here is a library that adds some formatter for the NodaTime types: https://github.com/ARKlab/MessagePack/tree/master/MessagePack.NodaTime

jtkech:
@mariojsnunes thanks for the feedback, as said by @sebastienros, maybe an exension point allowing a module to provide its own serializer / compressor for some document types, maybe not in this PR.

@sebastienros thanks for the feedback

- **I agree with what you said about perf with MessagePack and extensibility**, but here, knowing that with our multi level cache, a redis value is only read / write on initialization and on update, **the first of my 2 concerns** is not to find the faster serializer / compressor (even it is good), **but just to not have a too big redis value exchanged on the network** e.g. for the `AutorouteDocument`.

    That's why i opted for the simplest way by using a compressor as GZip that already reduces the `AutorouteDocument` by 80%. I agree that would be good to use MessagePack in compressor mode (didn't know it) if the compression is smaller than GZip, but just to say that in this PR my concern is not exactly how fast data are serialized, but still good to consider maybe in another PR, but first, here just reduce the size of the redis value on the network.

- **The 2nd concern is atomicity with the `AutorouteDocument`**. Maybe with our distributed lock (easy to use and seems to be safe as i have tried). But i still have a problem, when we mutate a volatile document (not persisted in the store) it is still updated in our multi level cache only after the ambient session is committed, e.g. the autoroutes document is only updated if the related content item is successfully persisted (we use checkConcurrency), this also allow to have multiple updates in the same scope and update the cache in one pass at the end of the scope.

    **So, for good reasons but we can't update the `AutorouteDocument` atomically by just using our distributed lock**. because the actual update is deferred.

- In fact there is another way, when a volatile document (as autoroutes) is first built from other data from the database, one solution is to not try to be atomic on updating the cache, but just invalidate it so that it is rebuilt on a next request, and from the up to date data from the database. But here another problem ;) rebuilding the autoroute document involves to request again all the published items from the `AutoroutePartIndex` table, so not once on first init.

- Yes i also thought about fragmentation but i felt that consisency and atomicity would be harder to do. Hmm, at this point maybe another approach, maybe a mix of a fragmentation (but not at the serialization level) and the invalidate pattern as described above to fix the atomicity concern, and then maybe without distributed lock ;)

    Note: Again, here we only talk about a document that may result in a too big redis value, that needs many other data from the database to be first built (so we want to request them once), and also that contains data related to many other documents (so updated when any of the other documents is updated), as the `AutorouteDocument`

So, maybe i will close this PR and open a new one with another approach, then maybe another PR for the extensibility point for serializers, maybe MessagePack by default and so on, but in another PR as i think it is important to first find a good compromise for my 2 concerns before releasing version 1.0

---

**For info**

> let's assume an autoroute is just url|id;url|id ... don't need to repeat the property names, the format can be very simple

Not so simple as an autoroute entry may include contained items routes and jsonPaths

> and MessagePack even has a mode to compress, and to split the buffer in blocks < 80KB as I mentioned

I didn't know that, i will take a look, thanks

> Also, if some types can't be serialized with MsgPack, then it can be extended with custom formatters

I know, this is what i already used e.g. for the case sensitive dico, but was blocked because of the polymorphim issue

jtkech:
@sebastienros just for infos

- **Okay to re-use MessagePack** and compression mode, custom formatters (but harder to define a custom document), maybe extensibility (not sure), **but less simple so maybe in another PR**.

   So, because data are only serialized when refreshing the cache, **here i just want to reduce the size of the data over the network**, even it is good to have a faster serializer.

    **And i found a simple way to automatically compress the data when > 10Kb** without having to use any additional options and marker interface. In my tests using the blog recipe and adding some routes

      Autoroutes 18025 bytes => 2535 bytes    ratio = 7
      Definitions  16394 bytes => 1955 bytes    ratio = 8

      // Other documents are not compressed as their sizes are < 10Kb

- **My last concern is atomicity** for the special case of the Autoroutes document, i'm working on

jtkech:
@sebastienros 

- Okay, i will allow to provide custom serializers, maybe through the existing configurable options per document, because it is easy to do if it is just to fallback to the default existing one if another serializer is not provided.

    So right now the default one would be `NewtonSoft` + `GZip` if need to be compressed. Then with this option i will try to re-use `MessagePack` + its compression mode for the `AutorouteDocument` and see what happens.

    Not yet done, maybe in another PR, will see.

- In the meantime i worked on atomicity, for and only for volatile documents, i added an atomic update using our distributed lock (a local lock by default), and i tried it with the `AutorouteDocument`.

jtkech:
@sebastienros for info

- I refactored how `DocumentOptions` is configured per document type, so that we still takes into account configuration sources, but now we can easily override these options also by code with regular extensions.

    I didn't do it before because it was not possible to inject an `IOptions<>` **per document type** by only using `IConfigureOptions<>`. But now we use `IConfigureNamedOptions` and `IOptionsSnapshot<>` that has a `Get()` method with a name parameter that we use to pass the document type.

    **So now it is possible and easy to override an option also by code for a given document type**

- **I added the `Serializer` as an option**, so it is very easy to provide a custom serializer for a given document type. I used it for the autoroute module, by passing the document name as we register here an `IConfigureNamedOptions`.

            services.Configure<DocumentOptions>(typeof(AutorouteDocument).FullName, options =>
            {
                options.Serializer = AutorouteDocumentSerializer.Instance;
            });


- So for testing i used an `AutorouteDocumentSerializer` based on `MessagePack`. Yes, without compression `MessagePack` is much more faster, with compression it is still faster but the data size (data size over the network was my concern here) is higher than if you use `Newtonsoft + GZip`. So maybe i will remove this custom serializer, but still keeping the extensibily to easily override the serializer, so that i will be able to work more on `MessagePack` in a future PR.

    I did some test with **1 Million of autoroutes, a whole update** (get/update/set and UI time) takes around 30s with `MessagePack + Lz4 compress` (data size = 90Mb), around 60s with `Newtonsoft + GZip` (data size = 50Mb).

    For **100_000 autoroutes, a whole update** (get/update/set and UI time) takes around 6s with `MessagePack + Lz4 compress` (data size = 9Mb), around 10s with `Newtonsoft + GZip` (data size = 5Mb).

    **So anyway, if we want to support e.g. 1 Million of autoroutes per tenant, we need another solution**, maybe fragmentation, i will work on it, but first **one question is how many autoroutes do we want to support?** 

- About atomicity see my previous comment, but it may depend on the answer of the above question

deanmarcussen:
Interesting figures @jtkech 

My thoughts on fragmentation / autoroute
- It would be difficult to impose a limit on how many autoroute entries we should support - people will always go over. 
- Autoroutes are generally long term (i.e. static) after a period of time
- Updates are more often around new autoroutes, while content is being finalized, so become static soon after creation.
- Updates shouldn't become slower over time (as the entry list grows)
- Fragmentation of the distributed document (internally, so not visible, but with the local memory cache version not fragmented?) seems like a way to workaround this, without having to impose any limits.

Good work getting `MessagePack` back in. 

I still didn't understand where it fell over due to polymorphism? What's the document that causes problems for that?

jtkech:
@deanmarcussen thanks for the feedback

> It would be difficult to impose a limit on how many autoroute entries we should support

> Updates shouldn't become slower over time (as the entry list grows)

Yes i felt it like this but was not sure ;) So fragmentation looks like a way to go, i was too lazy to start on it before the other concerns, it would decrease the atomicity concern, the maximum fragmentation would be one entry per autoroute but then too many keys. Anyway, with any medium fragmentation there is always a limit but higher and yes without imposing it

> cache version not fragmented?

Yes i think, at least for now

> I still didn't understand where it fell over due to polymorphism? What's the document that causes problems for that?

The problem is e.g. with a list of a given base type but that contains different derived types, when deserialized the list is only populated with objects of the base type. E.g. with a list of `Query` that each may be a `LuceneQuery` or a `SqlQuery`

jtkech:
I prepared this PR to be merged by only keeping safe improvments, this by excluding an usage related to atomicity as we need another approach (e.g. fragmentation) that i prefer to start in a separate PR, idem for an usage of `MessagePack`, but by keeping the extensibilities so that it will be easy to re-use them in separate PRs.

- **Improves document options (per document type) management** allowing to use the regular options model

- **Makes the serializer an option allowing to define a custom serializer** (per document type), e.g. `MessagePack`

    Here i only used it for autoroutes, so i removed this usage but i kept the option to be easily used in a future PR

- For volatile documents i introduced a new `UpdateAtomicAsync()`

    Here i only used it for autoroutes, so i removed this usage but i kept this new method to be used in a future PR.

- Here **the default serializer compresses data with GZip but only if size > configurable threshold**

    As said the serializer is configurable so the compression part is also overridable

PR #7398, opened 2020-10-27
Add proper icon while switching between themes

hishamco:
DarkMode


Skrypt:
This won't display the proper icon if the theme is automatically set. I intentionally did not add this feature because the script that automatically applies the dark mode if your OS is set to darkmode is ran before the document is ready. It doesn't know if there is a button yet.

Skrypt:
Also it is not sun mode but dark mode so the moon makes sense to be kept ...

Skrypt:
Unless you take time to make it work in both use cases we can't merge this yet.

Skrypt (changes requested):
Unless you take time to make it work in both use cases we can't merge this yet.

hishamco:
> This won't display the proper icon if the theme is automatically set

When I enable the dark mode it doesn't take the affect unless I click the button so how it will be automatically applied?

> Also it is not sun mode but dark mode so the moon makes sense to be kept ..

There's no notion of sun mode nor moon mode but this common in many sites for light and dark theme

Skrypt:
See this :  

https://github.com/OrchardCMS/OrchardCore/blob/62a69463b3288aaf498cae9435f37a3ee7e22baa/src/OrchardCore.Themes/TheAdmin/Assets/js/header/darkmode.js#L13

hishamco:
Now is working on both: loading the page when the theme is dark and clicking the dark theme button

PR #7400, opened 2020-10-27
NumericField (Range): get continuous updates while dragging the slider

netwavebe:
onchange is triggered when the user releases the mouse. To get continuous updates, you should use the oninput event, which will capture live updates in Firefox, Safari and Chrome, both from the mouse and the keyboard.

However, oninput is not supported in IE10, so it's best to combine the two event handlers.

Before:

RangeBefore

After:

RangeAfter


Skrypt:
We don't need to support IE10 so it's fine.

Issue #6770, opened 2020-07-25
MySqlException: Table 'exam.identifiers' doesn't exist


uussoft:
When I installed orchardcore, this exception occurred when selecting MySQL database

![微信图片_20200725093805](https://user-images.githubusercontent.com/3772962/88446223-b9ee4480-ce5a-11ea-8756-fcc328abb905.png)

uussoft:
My database version is mysql-5.5

hishamco:
Please provide us with the steps to reproduce

hodamz:
@uussoft was you able to fix this?

sebastienros:
I'd suggest we add the current functional tests running with other databases.

hishamco:
@agriffard do you know how to run the functional test multiple time, so I can finalize this one

Skrypt:
You will need Docker containers for doing this. I've started working on a PR for testing the UserId migration which could run wathever we need. We might want to build / test this also directly with the CI's later on.

Skrypt:
Should we close this one if functional test works fine on the CI? I assume that the issue comes from a custom IndexProvider since the table concerned is "exam" and we don't have this in OC. Looks like a failed migration that happened.

lengockyquang:
i'm having this issue with mariadb version 10.1.47. Any ideas to fix it ?

Skrypt:
Repro steps?

lengockyquang:
i just normally created empty database then put connection string at the setup setup. After i click save, it throw error that Identifiers does not exist. When i checked database, only table Document was created. I will try to create video for this issue tomorrow if i have free time :) almost 11pm from vn now

scil:
I use mysql-5.7.4-m14-winx64.zip and encounter the same error.

Skrypt:
Generally, we have issues under SQL Server when the database table is accessed with a different Schema. But under MySQL, a schema is a synonym for a table. Also, "exam.identifiers" here seems like a bad prefix issue if you prefixed your table names with "exam" then it should be "exam_identifiers". Need to see details.

scil:
> @Skrypt I used mysql-5.7.38 and everything went well. Today I use mysql-5.7.4-m14-winx64.zip and encounter this error.
> 
> mysql 5.7.38 vs 5.7.4

my collegues have fixed it. 

just change mysql  character from utf8mb4_unicode_ci to utf8_unicode

PR #7406, opened 2020-10-27
CI Testing across SQL MySql and Postgres databases

deanmarcussen:
Related to https://github.com/OrchardCMS/OrchardCore/issues/6770

Implements some `Theory` attributes which skips tests when Connection strings are not supplied by the environment (CI or Local)

The idea is to run the integration tests against the SQL databases we support, rather than just Sqlite.

Primarily I'd like to do this because
- Sqlite doesn't support max column lengths, so we've had to fix a number of issues when users get to production. These could be better addressed by testing against a production database server.
- Table name lengths. Same issue - works on Sqlite, but MySql, and Postgres have smaller limits (here I've had to write a limited (read quick) table generator, because a guid is too long and MySql and Postgres both fail to create tables.)
- I'd like to do some optimisation of the indices on the index tables (including `DocumentId` and others) which are a lot easier to do and prove when one can run a test.

This might be controversial as @sebastienros had suggested to do this with the functional tests, rather than the integration tests.

Future extensions might include some test recipes (easy to keep in the test project without them leaking into nuget) to activate features like open id etc (again trying to test things like table lengths, and general setup), whereas the work @jptissot is doing with cypress would suit testing other pieces.

Would also be nice to spin up two servers and redis to test distributed... but we'll see.

I did have to downgrade xunit to get around this issue as well https://github.com/xunit/xunit/issues/1782

Now we'll see if the ci passes...


appveyor.yml
```diff
@@ -15,6 +15,11 @@ skip_commits:
     - mkdocs.yml
     - src/docs/**/*
 
+services:
+  - mssql2017
```
Skrypt, line 19: Nothing more recent? 2019?

Skrypt: https://www.appveyor.com/docs/services-databases/#sql-server-2019

Skrypt:
So basically these unit tests won't execute on Travis because we won't supply a connectionstring.

Some thought : 

I'm not sure if we should still be testing for MacOS on Travis. It is slowing down the CI a lot.

appveyor.yml
Skrypt: I can see why now.

test/OrchardCore.Tests/Apis/Context/Attributes/MySqlDataAttribute.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        public MySqlDataAttribute()
+        {
+            if (String.IsNullOrEmpty(Environment) && String.IsNullOrEmpty(Skip))
+            {
+                Skip = "MySql test skipped by environment";
+            }
+        }
+
+		public override IEnumerable<object?[]> GetData(MethodInfo testMethod)
```
Skrypt, line 31: indentation is strange here and other DataAttribute inherited class.

test/OrchardCore.Tests/Apis/Context/Attributes/PostgreSqlDataAttribute.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        public PostgreSqlDataAttribute()
+        {
+            if (String.IsNullOrEmpty(Environment) && String.IsNullOrEmpty(Skip))
+            {
+                Skip = "PostgreSql test skipped by environment";
+            }
+        }
+
+		public override IEnumerable<object?[]> GetData(MethodInfo testMethod)
```
Skrypt, line 31: indentation

deanmarcussen: Don't know why it's having such trouble with formatting here...

Anyway still trying to get sql server working

deanmarcussen:
Ok, I'm happy this is passing reliably on appveyor.

I've now only tested each standard recipe against all 4 databases

(and removed one test which is covered by other tests, and creates unnecessary log entries during the test process)

More tests (I would like some) may come later on another pr

Skrypt (approved):
LGTM

PR #7411, opened 2020-10-28
Travis : builds taking too long

Skrypt:
I commented out the OSX specific build because I don't think this build is necessary.
If it's building under Linux it should also build under OSX. This is a none issue in almost all cases.

Though, since we are not running any unit/functional tests from the Travis builds we can't really know if there are specific things that would fail under OSX/Linux. 

**I propose we remove the OSX build since it's only slowing down the entire experience.**

And to be honest I really question the utility of using Travis if we are not running any tests on it's builds.
Also, I think we could easily run those builds in AppVeyor.

Right now, every commits takes around 45 minutes to get built ...


jptissot:
Why don't we move to Github actions ? I have it successfully setup for my repos and it's very easy to use.

sebastienros:
When I checked the build yesterday, it was the Linux step which was queued, not macos.

We should try github actions, step by step, and verify it can support database providers (docker?).

Skrypt:
Ok I agree let's drop Travis completely. No need at all.

Skrypt:
https://github.com/actions/virtual-environments
Github actions supports MacOS 11 in preview mode.

Skrypt:
@jptissot You mean this? https://github.com/jptissot/cypress-orchardcore/blob/master/.github/workflows/pipeline.yml

jptissot:
That is one of my github actions yes. We also use it on this repo: https://github.com/StatCan/StatCan.OrchardCore/tree/master/.github/workflows

deanmarcussen:
> We should try github actions, step by step, and verify it can support database providers (docker?).

If we like the idea of https://github.com/OrchardCMS/OrchardCore/pull/7406 I don't mind slowly stepping the concept over to github actions.

Appveyor is just crashing everytime at the moment (randomly) on it, but it is working fine for me locally with docker containers.

Github actions seems a lot faster generally for me, and has support to run the containers it needs

Skrypt:
I just saw last night that we use Travis to create the public docker images. We can't really drop the Travis build unless we move it to something else like Github Actions.

PR #7413, opened 2020-10-28
Update RazorPage.cs

### LINKED ISSUE #7414: Saving a liquid page with a custom layout should not throw Object null ref runtime error
1. Create a new liquid page
2. Provide and save: {% layout 'ThreeJSLayout' %}
3. On save 'Object reference null' runtime exception thrown.

image

Adding Context?. 'seems' to fix it - Page gets saved successfully.

Please review PR: https://github.com/OrchardCMS/OrchardCore/pull/7413

If you don't set the layout (in step 2), error doesn't occur as Context seems to be not null in those usual scenarios.

scriptmaster:
### Why?
Saving a liquid page with a custom layout throws null ref runtime error.

1. Create a new liquid page
2. Provide and save: `{% layout 'ThreeJSLayout' %}`
3. On save 'Object reference null' runtime exception thrown.

### Observations:

1. Page gets saved successfully.
2. Layout template name does exist and renders after save.
3. Adding Context?. 'seems' to fix it - please review/check.
1. If you don't set the layout (in step 2), error doesn't occur as Context seems to be not null in those usual scenarios.

Fixes #7414

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
@@ -107,7 +107,7 @@ namespace OrchardCore.DisplayManagement.Razor
             {
                 if (_themeLayout == null)
                 {
-                    _themeLayout = Context.Features.Get<RazorViewFeature>()?.ThemeLayout;
+                    _themeLayout = Context?.Features.Get<RazorViewFeature>()?.ThemeLayout;
                 }
 
                 return _themeLayout;


scriptmaster:
image

jtkech:
@scriptmaster 

Good catch, yes i can repro, this is when we render a view without any ambient view context, e.g. in a background task, through GraphQL, and **here when Lucene indexing request the body aspect** of the liquid part.

Idem with an html part, we don't see it because of sanitization, but if we disable the sanitization in the html part setting, it also fails, of course if there is a liquid tag / filter using the `Context` property of the razor page (here a derived liquid page). **Note: There is no sanitization settings for a Liquid part**. Also it doesn't fail if you disable Lucene.

In `RazorPage.cs` we would need to do a null check everywhere we use the `Context` property, **in fact the right fix is to not have a null `Context` property in these special cases**.

So, when there is no view context, we build one on the fly, for razor it doesn't fail as somewhere we use the mvc `view.RenderAsync(viewContext)` where we pass the view context we created on the fly. **So for liquid, the fix is to just init the view context of the page** (the http context is resolved from the view context) with the one we created on the fly.

I will do a PR soon, thanks again because you found the issue ;)

jtkech:
@scriptmaster so i close this PR but thanks again because you found the issue ;)

scriptmaster:
> to init the view context of the page

makes sense.  deleting my patch as #7424 is a better solution.

PR #7416, opened 2020-10-28
Fix toggle password style in site login

### LINKED ISSUE #7408: Login page with Blog Theme
If you check 'Use site theme for login page' in the settings of a site using theBlogTheme, the login page looks like this:

image

The button to show the password is not looking good.

hishamco:
Fixes #7408

Login


hishamco:
I just inline the style to ensure that will work across all site themes

deanmarcussen (commented):
Anything like this should be at theme level, rather than in the view.

Unfortunately here you are adding a style which is theme dependant, and while it may provide a fix for the samples themes, does not fit well with the concept of theming in general

Skrypt:
I will take a look.

hishamco:
> Anything like this should be at theme level, rather than in the view.

That means we should add it in all the site themes

agriffard:
Add it to bootstrap-oc.scss may be.

hishamco:
Seems `bootstrap-oc.scss` available in all the themes

Thanks

hishamco:
@agriffard it works!! so it's one from my side and ready to merge

PR #7417, opened 2020-10-28
Remove Handlers dependency from Display Manager

### LINKED ISSUE #7377: Display Driver now depends on the handler
Relating to pr https://github.com/OrchardCMS/OrchardCore/pull/6195

comment from @sebastienros 
> Can it be an issue that the Display Driver now depends on the handler?

The comment is correct, the `Validating` code added to the `ContentItemDisplayManager` should be moved to the `ContentItemDisplayCoordinator` as it currently creates an unwanted dependency between the display manager and drivers.

/cc @MichaelPetrinolis

MichaelPetrinolis:
closes #7377


hishamco:
Please fix the build

src/OrchardCore.Modules/OrchardCore.Alias/Models/AliasPart.cs
```diff
@@ -4,6 +4,9 @@ namespace OrchardCore.Alias.Models
 {
     public class AliasPart : ContentPart
     {
+        // Maximum length that MySql can support in an index under utf8 collation.
+        public const int MaxAliasLength = 767;
```
deanmarcussen, line 8: Not here. 
I understand why you want to move it, but on the part it becomes part of the JSON I suspect. Plus is also in the views.

`Startup`  maybe as a central spot?

src/OrchardCore.Modules/OrchardCore.Alias/Models/AliasPartExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.ComponentModel.DataAnnotations;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.Extensions.Localization;
+using OrchardCore.Alias.Indexes;
+using YesSql;
+
+namespace OrchardCore.Alias.Models
+{
+    internal static class AliasPartExtensions
```
deanmarcussen, line 11: I'm not sure these should be internal. 
If you need to swap the driver out (which is supported), for another reason - i.e. not validation, then you're going to have to copy these extensions to the project, for no good reason - or you write your own which are wrong ;)

src/OrchardCore.Modules/OrchardCore.Autoroute/Drivers/AutoroutePartDisplay.cs
```diff
@@ -110,12 +112,29 @@ public override async Task<IDisplayResult> UpdateAsync(AutoroutePart model, IUpd
                     await updater.TryUpdateModelAsync(model, Prefix, t => t.SetHomepage);
                 }
 
-                var errors = await context.ValidateAsync(model);
-                updater.ModelState.BindValidationResults(Prefix, errors);
+                updater.ModelState.BindValidationResults(Prefix, model.ValidatePathFieldValue(S));
+
+                // This can only validate the path if the Autoroute is not managing content item routes or the path is absolute.
+                if (!String.IsNullOrEmpty(model.Path) && (!settings.ManageContainedItemRoutes || (settings.ManageContainedItemRoutes && model.Absolute)))
```
deanmarcussen, line 118: Yeah so this is why we were always undecided about sharing the validation code.

The logic is similar between the handler and here (maybe even the same) but is used in different ways.

src/OrchardCore.Modules/OrchardCore.Autoroute/Models/AutoroutePartExtensions.cs
```diff
@@ -0,0 +1,29 @@
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Linq;
+using Microsoft.Extensions.Localization;
+
+namespace OrchardCore.Autoroute.Models
+{
+    internal static class AutoroutePartExtensions
+    {
+        internal static IEnumerable<ValidationResult> ValidatePathFieldValue(this AutoroutePart autoroute, IStringLocalizer S)
```
deanmarcussen, line 10: We need to check that the po extractor picks this up.

We know it finds, `var S = serviceProvider.Get<...>` but we don't know if it finds `S` in a method parameter.

Maybe @agriffard will check it?

Maybe @agriffard would also have to update all the po files so we don't lose the translations either.

src/OrchardCore.Modules/OrchardCore.Autoroute/Models/AutoroutePart.cs
```diff
@@ -4,6 +4,9 @@ namespace OrchardCore.Autoroute.Models
 {
     public class AutoroutePart : ContentPart
     {
+        public static char[] InvalidCharactersForPath = ":?#[]@!$&'()*+,.;=<>\\|%".ToCharArray();
+        public const int MaxPathLength = 1024;
```
deanmarcussen, line 8: idem same

MichaelPetrinolis: @deanmarcussen thanx for the review. I moved field validation constants to Startup, although they shouldn't be part of the JSON. Maybe introducing a class like AutoroutePart.Fields would be a better option? Or in Migrations?

jtkech: > they shouldn't be part of the JSON

Maybe a `[JsonIgnore]` attribute

MichaelPetrinolis: Thanx for the tip @jtkech. I really prefer having the constants in Part model class instead of startup. They are ContentPart's characteristics and they are available on all places that use the part. I did not decorate them with [JsonIgnore] because it is only applicable on properties. Also, JSON.Net does not serialize constants/static members out of the box unless they are decorated with [JsonProperty] attribute. Could not find any custom converter for ContentPart in source that would force serialization of consts/static members. So I moved them back for now.

agriffard:
Can you please merge dev?

Issue #2266, opened 2018-08-23
Numeric field editors are not compatible with localization


sebastienros:
From @jtkech 

@sebastienros cool, i will take a look more in depth but first as @agriffard i also saw some little issues i'm working on, this around **decimal** number formats according to the culture.

@Skrypt when the current issues will be fixed, i will be open to do the same but for content parts.

### E.g with the html5 `type="number"` in the spinner editor

- We would need to always pass string values in invariant culture to get it first working.

- Then it displays the value according to the browser culture but which may not be the same as the server side `CurrentUICulture`.

- Then the values are always posted back in invariant culture and the driver which use the `CurrentUICulture` can also fail to parse...

As i remember that's why in O1 we always use `type="text"`. So maybe we would need to get rid of the spinner editor (`type="number"` is also used for the scale settings but here there is no decimal part).

Or implement it with a script which will render the value according to the server side `CurrentUICulture`. But the script would need to be culture aware, so not so easy. Note: [i did a PR in O1](https://github.com/OrchardCMS/Orchard/pull/5511) which was using `Globalize.js` to do this kind of things but i don't remember the details.

- So, right now i will first check all other number editors when using a decimal part.

**UPDATE**

- That's ok for the `Number` editor, i just did a little cleanup that i will commit. And that's ok for the `Select` editor.

- But same remark for the html5 `type="range"` as above for `type="number"`.

- Then same problems with the `Slider` editor which uses `bootstrapSlider()` in a script which is not culture aware.

- So, for me there are only the `Number` and the `Select` editors which are orchard culture compatible, unless we do client side culture aware scripts. This doesn't mean that i don't like other editors ;)

Hmm, one possible solution would be to display a decimal separator and implement separately on the left the integer part and on the right the decimal part (if exists) but which would be both integers ...

hishamco:
FYI input `type="number"` can display comma or dot as separator based on the current culture, but not accept it as a valid value

netwavebe:
@agriffard pointed me to this issue.

Noticed this too when working on the editors of NumericField. Some are broken when scale > 0 due to localization (point/comma).

CurrentUICulture is used on the server for validation:
https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.ContentFields/Drivers/NumericFieldDisplayDriver.cs#L72

[...]

PR #7423, opened 2020-10-28
NumericField Range: fixed localization

netwavebe:
The 'Range' NumericField did not use localized values when scale > 0. Fixed this by using the same technique as in the 'Spinner' NumericField.

Before:

RangeLocBefore

After:

RangeLocAfter


netwavebe:
Again, https://github.com/OrchardCMS/OrchardCore/issues/2266 discusses the problems.

PR #7426, opened 2020-10-29
NumericField Slider: fixed localization & layout improvements

netwavebe:
The 'Slider' NumericField did not use localized values when scale > 0. Fixed this by using the same technique as in the 'Spinner' NumericField.

Because it used `type="number"` localization seemed okay, but you could not save a content item.

Before:

SliderBefore

After:

SliderAfter

I also changed the layout to match the other fields (like min/max).

The slider itself is not responsive, it has hardcoded values in its stylesheet. I moved it out of the column so it can be bigger than the input-element (like the hint).

I changed the tooltip to 'show' instead of 'always' so it's only visible when hovering over the slider, as it could overlap the input-element.

When entering non-numeric values, there are no more javascript-exceptions.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Views/NumericField-Slider.Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Views/NumericField-Slider.Edit.cshtml
@@ -8,6 +8,8 @@
     decimal min = settings.Minimum.HasValue ? settings.Minimum.Value : 0;
     decimal max = settings.Maximum.HasValue ? settings.Maximum.Value : 10000;
     string id = Html.IdFor(m => m.Value);
+    string javascriptDecimalSeparator = CultureInfo.InvariantCulture.NumberFormat.NumberDecimalSeparator;
+    string serverDecimalSeparator = CultureInfo.CurrentUICulture.NumberFormat.NumberDecimalSeparator;
 }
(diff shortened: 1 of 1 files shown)


netwavebe:
Again, https://github.com/OrchardCMS/OrchardCore/issues/2266 discusses the problems.

src/OrchardCore.Modules/OrchardCore.ContentFields/Views/NumericField-Slider.Edit.cshtml
```diff
@@ -18,31 +20,58 @@
...
+            <span class="ml-3 mr-1">@max</span>
+        }
+    </div>
     @if (!String.IsNullOrEmpty(settings.Hint))
     {
         <span class="hint">@settings.Hint</span>
     }
 </div>
+
 <script at="Foot">
```
Skrypt, line 56: asp-name ?

Skrypt: Yeah ok it initializes the bootstrapSlider per field {id}

netwavebe: Exactly, I need this script to render for each field.

PR #7430, opened 2020-10-29
NumericField Range: keep slider in sync with input values

netwavebe:
The 'Range' NumericField should keep the slider in sync with input values, just like the 'Slider' NumericField.

Before:

RangeSyncBefore

After:

RangeSyncAfter

Non-numeric input values are defaulted to 0. A range can handle values outside min/max boundaries.


Skrypt (approved):
Apart than the fact that the onchange method could be set on each inputs that are Range numeric fields this is LGTM. I'm not requesting any changes here because it works fine that way too.

``` javascript
object.addEventListener("change", myScript);
```

PR #7434, opened 2020-10-29
GraphQL 15.4.0

agriffard:
https://github.com/graphql/graphql-js/releases/tag/v15.4.0


agriffard:
Now that Codeowners are defined, it automatically requests a review to team members according to their associated modules/folders.

PR #7443, opened 2020-10-30
Use Advanced markdown extensions

### LINKED ISSUE #7422: Markdown Strikethrough
I'm using the version 1.0.0-rc2-13450 from nuget following this [guide](https://docs.orchardcore.net/en/dev/docs/guides/decoupled-cms/) for creating a decoupled CMS Website (great btw). When parsing the strikethrough's markdown in the side by side panel is correct, but not in the website.

image
image

hishamco:
Fixes #7422

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Markdown/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Markdown/Startup.cs
@@ -49,7 +49,11 @@ namespace OrchardCore.Markdown
             services.AddLiquidFilter<Markdownify>("markdownify");
 
             services.AddOptions<MarkdownPipelineOptions>();
-            services.ConfigureMarkdownPipeline((pipeline) => pipeline.DisableHtml());
+            services.ConfigureMarkdownPipeline((pipeline) =>
+            {
+                pipeline.DisableHtml();
+                pipeline.UseAdvancedExtensions();
+            });
 
             services.AddScoped<IMarkdownService, DefaultMarkdownService>();
         }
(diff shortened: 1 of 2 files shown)


deanmarcussen (commented):
Looks good, but the docs need updating as well, as they specifically explain how to add these extensions, so just need to be updated to say the extensions are there by default, and show an example of how to add a different extension 
https://docs.orchardcore.net/en/dev/docs/reference/modules/Markdown/

agriffard:
Could we imagine a setting to configure the extensions?

https://github.com/lunet-io/markdig/blob/751c79fce4e92857f861d8af0854a8eff0b29071/src/Markdig/MarkdownExtensions.cs#L539

With `nohtml+advanced` by default.

hishamco:
> With nohtml+advanced by default.

We could but it call the same extensions at the end, no need for extra step IMHO

deanmarcussen:
In what respect would you prefer a setting @agriffard ?

i.e. are you talking about on a per field basis, or the pipeline as a whole?

If you mean pipeline as a whole, then potentially the two of these could be features, which are on by default, and can be disabled as a feature, rather than having to disable them via options?

Or are you talking about appsettings?

agriffard:
> Or are you talking about appsettings?

Yes, appsetting.

deanmarcussen:
> Yes, appsetting.

I'm not against it, if you want to do it.

deanmarcussen (approved):
For the meanwhile this LGTM

hishamco:
I can start a new PR if all agree about disable these features from appsettings

Issue #7420, opened 2020-10-28
Admin UI : Improve admin menu filter


Skrypt:
I'm opening an issue about this because an attemps has been made here : #5769
But since I refactored the admin menu using proper semantics the PR doesn't work anymore.
I will take some time to refactor this PR to get the intended results by @agriffard

deanmarcussen:
I still think both of these, the admin menu filter, and dark mode should be on by default.

Both good features, don't break things, and can then be turned off. 
Otherwise noone even knows they exist.

hishamco:
Totally agree with Dean, that they should be on by default

PR #7444, opened 2020-10-30
Menu filter & switch themes should be on by default

hishamco:
Address the comment https://github.com/OrchardCMS/OrchardCore/issues/7420#issuecomment-718506915

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Admin/Admin/AdminSettings.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Admin/Admin/AdminSettings.cs
@@ -2,8 +2,8 @@ namespace OrchardCore.Admin.Models
 {
     public class AdminSettings
     {
-        public bool DisplayMenuFilter { get; set; }
+        public bool DisplayMenuFilter { get; set; } = true;
 
-        public bool DisplayDarkMode { get; set; }
+        public bool DisplayDarkMode { get; set; } = true;
     }
 }


agriffard:
Will the default theme will be the light one?

hishamco:
> Will the default theme will be the light one?

Yep, this just will show the button that enable you to switch between the themes

src/OrchardCore.Modules/OrchardCore.Admin/Admin/AdminSettings.cs
```diff
@@ -2,8 +2,8 @@ namespace OrchardCore.Admin.Models
 {
     public class AdminSettings
     {
-        public bool DisplayMenuFilter { get; set; }
+        public bool DisplayMenuFilter { get; set; } = true;
 
-        public bool DisplayDarkMode { get; set; }
+        public bool DisplayDarkMode { get; set; } = true;
```
hishamco, line 7: @agriffard as the property said display the dark mode but not enable it, so we 're safe ;)

agriffard:
Please wait that #7420 is fixed (admin menu filter) before merging this.

Skrypt:
The default theme is defined by your OS current dark mode. If you are under Mac OS and you enabled dark mode on your OS it will show the dark admin theme by default ... If you click on the button then it stores your preference in the LocalStorage of your browser. So, once we enable the dark mode the default is not the light theme. If you don't enable it then the default is the light theme.

hishamco:
> If you are under Mac OS for example and you enabled dark mode on your OS it will show the dark admin theme by default

How the theme is OS related?!!

Skrypt:
Dude I showed you the line that does that and I added a comment on the PR. It's a new feature that has been added to browser that started by Safari on Mac OS. When they added the Mac OS X dark mode they also added an option to have a dark mode on websites to trigger automatically.

https://css-tricks.com/a-complete-guide-to-dark-mode-on-the-web/

Now Bootstrap doesn't support it yet but it's planned for version 6 isssh.

hishamco:
Thanks, I didn't know about this feature, but can we enable the light theme instead of set the dark theme one by default, at least as @deanmarcussen suggested the user will notice there's a new feature. I agreed with him to show the feature but not enable it by default

Skrypt:
And this is why I did not enable it by default 😄

hishamco:
You 're the man ;)

What if disable the dark mode then and persist it `darkMode = false;` this should work?

Skrypt:
When you disable the dark mode the javascript doesn't apply changes so even if the LocalStorage does hold the `darkmode = true` it won't apply it.

Issue #7395, opened 2020-10-26
If you apply multiple taxonomy filters ,  no results.


nakamura2000:
steps to reproduce.

1. enable "Taxonomies" feature and "Taxonomies Contents List Filters" feature.
2. create content type "term" which have title body part.
3. create content item "color" as "taxonomy" type, and it has "red" and "blue" (terms) .
4. create content item "shape" as "taxonomy" type, and it has "square" and "round" (terms).
5. create content type "object" and it has "color_field" and "shape_field" as taxonomy field.
6. create content item "dish" for "object" , and it has "red"  term and  "round" term. 
7. create Taxonomy Filters for "color" and "shape".
8. filter content by "color" : "red" , and "dish" content item is successfully filtered and shown.
9. filter content by "shape" : "round" , and "dish" content item is successfully filtered and shown.
10. but If you filter content by both "color" :"red" and "shape" : "round" , there is no result.

This issue may be already addressed to solve. 

https://github.com/sebastienros/yessql/commit/68972b2325c0594f98491fd40e2c77bd75dd3b59#diff-25a6634263c1b1f6fc4697a04e2b9904ea4b042a89af59dc93ec1f5d44848a26

I have applied latest yessql from dev source code 
(and fix some changed method) and I confirmed it works.

In TaxonomyContentsAdminListFilter.cs FilterAsync

current:

   query.With<TaxonomyIndex>(x => x.TermContentItemId == viewModel.SelectedContentItemId);

to:
   query.All(
       x => query.With<TaxonomyIndex>(x => x.TermContentItemId == viewModel.SelectedContentItemId)
   );

deanmarcussen:
Yes, it was never designed to work like that.

However the Yes.Sql improvements should work but I suspect just using the `IsIn()` extension would work just as well.

sebastienros:
The query in code needs to change to handle AND. Either with the .All or the IsIn(), whichever makes sense.

sebastienros:
@nakamura2000 do you want to try and create a PR for this?

nakamura2000:
I confirmed that taxonomiIndex is aliased in generated sql and filter is applied for each aliased table with above code.

I will create PR but current linked nuget yessql version is 1514 so I had to add version attribute 1614 in csproj.

SELECT count(distinct [Document].[Id]) 
FROM [Document] 
[...]

PR #7468, opened 2020-11-01
YesSql Upgrade and fix multiple taxonomy filter issue 

nakamura2000:
Apply yessql 1565 and use new method (query.All) to fix multiple taxonomy filter issue (#7395).
yessql 1565 have some breaking change , so I fixed that method to corresponding method.
Graphql library is also affected by the yessql change.Handling “where” part of Graphql generates partial sql “where” clause and It needs modification to apply sql table alias name.

taxonomy filter modification part  (TaxonomyContentsAdminListFilter.cs) is my main pull request.


nakamura2000:
fixed GraphQL error and pass test cases. But this fix uses internal method by reflection.

sebastienros:
Took the freedom to update this PR with the new release of yessql where I added .GetTableAlias as a public method. This allows to remove the private reflection. Just to discover that we could also use a `With(Type)` instead of the generic one. So I added some public reflection for now.

sebastienros:
Was very impressed by the way you managed to understand how the grapql code is building the dynamic queries, then understand how yessql internally creates the custom aliases for each bound table. I hope you'll be able to keep contributing.

nakamura2000:
Thank you for your comment. I'm glad that I could contribute to this project.

src/OrchardCore.Modules/OrchardCore.MiniProfiler/Startup.cs
```diff
@@ -33,7 +33,16 @@ public override void Configure(IApplicationBuilder app, IEndpointRouteBuilder ro
             var store = serviceProvider.GetService<IStore>();
 
             var factory = store.Configuration.ConnectionFactory;
-            store.Configuration.ConnectionFactory = new MiniProfilerConnectionFactory(factory);
+            var profilerFactory = new MiniProfilerConnectionFactory(factory);
```
deanmarcussen, line 36: @sebastienros @nakamura2000 

The new feature is great, I wanted it to do this when I wrote it but couldn't.

Happy with this in `Startup`, rather than the Factory?

Re: YesSql updates  must remember to update the `CreateIndex` methods to use the new one.

nakamura2000: > Happy with this in Startup, rather than the Factory?

Sorry but I couldn’t understand your suggestion point.

MiniProfiler has to change because `SqlDialectFactory.SqlDialects` became not public.

If `CommandInterpreterFactory.For` has Type parameter method overload , the code will be more simple.

sebastienros: @nakamura2000 `SqlDialectFactory` has this: `public static ISqlDialect For(Type dbConnectionType)`, what else do you need?

src/OrchardCore.Modules/OrchardCore.MiniProfiler/Startup.cs
```diff
@@ -33,7 +33,16 @@ public override void Configure(IApplicationBuilder app, IEndpointRouteBuilder ro
...
+            store.Configuration.ConnectionFactory = profilerFactory;
+            var dialect = SqlDialectFactory.For(factory.DbConnectionType);
+            SqlDialectFactory.Register(profilerFactory.DbConnectionType, dialect);
```
sebastienros, line 39: Can there be an issue if we have tenants that have different databases and have the mini profiler enabled?  Since each tenant will want to register a different dialect?

src/OrchardCore.Modules/OrchardCore.MiniProfiler/Startup.cs
nakamura2000: @sebastienros 
I think `CommandInterpreterFactory` have to add `CommandInterpreterFactory.For(Type dbConnectionType)`

sebastienros: Doing already, should be there soon

src/OrchardCore.Modules/OrchardCore.MiniProfiler/Startup.cs
jtkech: Before there is

            var factory = store.Configuration.ConnectionFactory;
            var profilerFactory = new MiniProfilerConnectionFactory(factory);

So it seems that it only depends on the store, **so seems to be ok** as the store is a **tenant** singleton

src/OrchardCore.Modules/OrchardCore.MiniProfiler/Startup.cs
deanmarcussen: @nakamura2000 @sebastienros has added the missing feature to YesSql - he just needs to merge it to master to get it on NuGet.

Then we should be able to finish this one

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/7395

src/OrchardCore.Modules/OrchardCore.MiniProfiler/Startup.cs
sebastienros: I am just worried that anywhere inside yessql the dialect is resolved using the connection type, and in this case it's ProfileDbConnection, for all the tenants, even if this type wraps the actual connection type.

deanmarcussen: Yes, it seems that a tenant with sqlite and a tenant running sqlserver each reset the `SqlDialectFactory` `ProfileDbConnection` type as they startup with their own dialects. 

That being said, it all seems to work ok still. And was also happening previously, but based on the magic string `profiledbconnection`. 

So perhaps not getting resolved anywhere inside YesSql?

sebastienros: I checked the code and it might be resolved before we change the factory. So that's fine. Also very limited scope since we use it for devs. I might need to change yessql to at least be able to support that just in case. Too much "static" here. So GTG.

I also did all the pending changes.

PR #7471, opened 2020-11-01
Control MarkdownOptions via appsettings

hishamco:
Related to PR #7443

/cc @deanmarcussen


agriffard:
I was specifically talking about the Configure method that takes a string with extensions separated by `+`:
https://github.com/lunet-io/markdig/blob/751c79fce4e92857f861d8af0854a8eff0b29071/src/Markdig/MarkdownExtensions.cs#L539

hishamco:
> I was specifically talking about the Configure method that takes a string with extensions separated by +

I knew that, after that, the dialog goes to `appsetting`. Am I right?

agriffard:
The whole goal for me was to be able to enable any extensions with only one appsetting property, not to create one property for all the possible extensions.

hishamco:
> to enable any extensions with only one appsetting property

What if someone want to enable Emoji and keep html disabled?

agriffard:
> What if someone want to enable Emoji and keep html disabled?

`nohtml+emojis`

hishamco:
I got what you intend to do, seems I misunderstood you from the beginning  ;) I will update the PR

hishamco:
@agriffard seems we need to point to all possible values or features that could be used in the docs

src/OrchardCore.Modules/OrchardCore.Markdown/MarkdownOptions.cs
```diff
@@ -0,0 +1,14 @@
...
+    public class MarkdownOptions
+    {
+        /// <summary>
+        /// Gets or sets a list of the features to be used in <see cref="Markdig.MarkdownPipelineBuilder"/>.
+        /// </summary>
+        /// <remarks>
+        /// "nohtml" will call DisableHtml()
+        /// "advanced" will call UseAdvancedExtensions()
+        /// </remarks>
+        public string Features { get; set; } = "nohtml+advanced";
```
agriffard, line 12: `Extensions` maybe, instead of Features.

agriffard:
> @agriffard seems we need to point to all possible values or features that could be used in the docs

We could add a table with each extension/description in the OrchardCore.Markdown documentation.

hishamco:
> We could add a table with each extension/description in the OrchardCore.Markdown documentation

Could be, but we need to reduce the effort of updating the docs whenever an extension is added or removed, IMHO point to that function may useful, but all the options are open

src/OrchardCore.Modules/OrchardCore.Markdown/Startup.cs
```diff
@@ -22,14 +24,27 @@ namespace OrchardCore.Markdown
...
+            _shellConfiguration = shellConfiguration;
+        }
+
         public override void ConfigureServices(IServiceCollection services)
         {
+            var markdownOptionsConfig = _shellConfiguration.GetSection("OrchardCore_Markdown");
+            var markdownOptions = new MarkdownOptions();
+            markdownOptionsConfig.Bind(markdownOptions);
+
+            services.Configure<MarkdownOptions>(markdownOptionsConfig);
```
jtkech, line 46: We don't need this line if we only use the options to configure the markdown pipeline below

If so we could just do in one line

    var markdownOptions = _shellConfiguration.GetSection("OrchardCore_Markdown")
        .Get<MarkdownOptions>() ?? new MarkdownOptions();

deanmarcussen: We actually shouldn't have a `MarkdownOptions` class at all here, because it is being passed directly into the `.ConfigureMarkdownPipeline` and captured there. This means that `services.Configure<MarkdownOptions>` is unused, because the `IOptions<MarkdownOptions>` is itself never resolved.

In this case it would be better not to have the class at all, and just resolve the string directly from `IShellConfiguration` 

e.g. `_shellConfiguration.GetValue("OrchardCore_Markdown", "nohtml+advanced");`

jtkech: yes indeed if the options is now only a string ;)

hishamco: @deanmarcussen I was think about that before, but this will allow anyone to resolve `IOptions <MarkdownOptions>` instead on working with magic string

deanmarcussen: But it will be useless if resolved because the value of it has already been passed directly into the `MarkdownPipelineOptions`, and that is the only place it is useful.

hishamco: You are right because it's already configured, I will update the PR

deanmarcussen:
> We could add a table with each extension/description in the OrchardCore.Markdown documentation.

Looks good, just needs docs.

src/OrchardCore.Modules/OrchardCore.Markdown/Startup.cs
```diff
@@ -22,12 +24,21 @@ namespace OrchardCore.Markdown
 {
     public class Startup : StartupBase
     {
+        private static readonly string DefaultMarkdownOptions = "nohtml+advanced";
```
deanmarcussen, line 27:
```suggestion
        private static readonly string DefaultMarkdownExtensions = "nohtml+advanced";
```

src/OrchardCore.Modules/OrchardCore.Markdown/Startup.cs
```diff
@@ -51,8 +62,8 @@ public override void ConfigureServices(IServiceCollection services)
             services.AddOptions<MarkdownPipelineOptions>();
             services.ConfigureMarkdownPipeline((pipeline) =>
             {
-                pipeline.DisableHtml();
-                pipeline.UseAdvancedExtensions();
+                var extensions = _shellConfiguration.GetValue("OrchardCore_Markdown:Extensions", DefaultMarkdownOptions);
```
deanmarcussen, line 65:
```suggestion
                var extensions = _shellConfiguration.GetValue("OrchardCore_Markdown:Extensions", DefaultMarkdownExtensions);
```

src/docs/reference/modules/Markdown/README.md
```diff
@@ -166,6 +166,48 @@ To disable sanitization:
 @await Orchard.MarkdownToHtmlAsync((string)Model.ContentItem.Content.MarkdownParagraph.Content.Markdown, false)
 ```
 
+## Markdown Settings
```
deanmarcussen, line 169:
```suggestion
## Markdown Configuration
```

src/docs/reference/modules/Markdown/README.md
```diff
@@ -166,6 +166,48 @@ To disable sanitization:
...
+
+Enabling the `OrchardCore.Markdown` module will allow the user to set the following settings:
+
+| Setting | Description |
```
deanmarcussen, line 173: The table for supported extensions looks great!

Suggest instead of the table here for the settings it would be better to just show the configuration json (for the default values). So exactly what you have added to the `appsettings.json` file without the comments.

(Otherwise from reading this, I have no idea where to put the `Extensions` setting)

Example here https://docs.orchardcore.net/en/dev/docs/reference/modules/Media/#configuration

Also I think the https://docs.orchardcore.net/en/dev/docs/reference/modules/Markdown/#markdown-pipeline
pipeline options should be shown second, after this, as this is the more convenient way to configure.

So happy for you to move the pipeline options section to after this.

hishamco: Thanks Dean for your comments and suggestion, I will update the PR ASAP

PR #7473, opened 2020-11-01
Fix CulturePickerOptions settings

hishamco:
I think there's an issue with the `CulturePickerOptions` section name, I did a fix but I need to be 100% sure that is work fine, @jptissot when the `ContentCulturePickerController` is called, AFAIK there should be a culture picker in the site that trigger this, but I didn't see anything, did I miss something?


hishamco:
Now I'm able to trigger the controller after adding `ContentCulturePicker` shape, time to test the PR ;)

hishamco:
Everything looks good after testing

src/OrchardCore.Cms.Web/appsettings.json
```diff
@@ -44,8 +44,10 @@
     //  "CreateContainer": true // Creates the container during app startup if it does not already exist.
     //},
     // Uncomment to configure content localization.
-    //"OrchardCore_ContentLocalization_CulturePicker": {
```
deanmarcussen, line 47: This change is wrong.

The current section name is correct. It should match the `FeatureId`. Refer the `Manifest.cs` for more information.

We already spoke about this on your original pr.

Here you are confusing Features with namespaces, and class names. Which are not connected.

jptissot: I think the current code is ok no ?

hishamco:
> This change is wrong.

It's working fine Dean

> The current section name is correct. It should match the FeatureId. Refer the Manifest.cs to understand what a feature is.

I will have a look

> I think the current code is ok. No need for this PR

I will double check

src/OrchardCore.Cms.Web/appsettings.json
hishamco: > This change is wrong.

It's working fine Dean

> The current section name is correct. It should match the FeatureId. Refer the Manifest.cs to understand what a feature is.

I will have a look

> I think the current code is ok. No need for this PR

I will double check

jptissot:
@deanmarcussen Can this be closed ? I am not familiar with the structure of the settings. If it's based on the Feature, then close, if it's based on module structure, then maybe consider this. From your comment, it sounds like the current implementation is correct and this can be closed.

hishamco:
I'm not sure why we need to close this? I did similar thing in Markdown settings and the current PR works well

deanmarcussen:
The current naming is correct and makes sense.

hishamco:
Still wondering, what will happen if we add a new setting to `Content Localization`?!!

jptissot:
From what I gather, it would use another section. Sections are configured per feature and not per namespace.

hishamco:
IMHO settings should be grouped per module

PR #7475, opened 2020-11-02
Modify frontend links to open in new window

### LINKED ISSUE #7472: Any "View" action should open in new window
A small issue but from the Admin UI, when I click a "View" button in relation to a content type it feels like a new window should be rather than it being in the same window and being take away from the editing UI - like the preview button does.

Reasons:
- Consistency - all buttons that are going to show you what the content looks like, "View" and "Preview", perform the same and the user knows what to expect.
- Loss of contextual actions - say I am on a content list and I am selecting a couple of items to do a bulk action and I think "is this the content item I am thinking of" and I click the "view" button to check, you then loose whatever actions you have done that's aren't saved.

scleaver:
Fixes #7472 I have tested all links and they work as required but was not able to find in which situation OrchardCore.Sitemaps/Views/Admin/Display.cshtml is used.

@deanmarcussen - I think you created this can you shed some light?

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsButtonEdit_SummaryAdmin.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsButtonEdit_SummaryAdmin.cshtml
@@ -17,5 +17,5 @@
 
 @if (hasPublished && hasViewPermission)
 {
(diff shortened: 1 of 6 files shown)


scleaver:
Can anyone think of any other links that should be added to this list?

Skrypt:
Search for @T["View"] with Visual Studio Code.

Skrypt:
Seems like you covered all those.

PR #7484, opened 2020-11-02
Default value for AllowMediaText Setting

deanmarcussen:
It is incorrectly merged by Newtonsoft when used via `WithSettings`

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Media/Settings/MediaFieldSettings.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Settings/MediaFieldSettings.cs
@@ -9,6 +9,8 @@ namespace OrchardCore.Media.Settings
 
         [DefaultValue(true)]
         public bool Multiple { get; set; } = true;
+        
+        [DefaultValue(true)]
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.Media/Settings/MediaFieldSettings.cs
```diff
@@ -9,6 +9,8 @@ public class MediaFieldSettings
 
         [DefaultValue(true)]
         public bool Multiple { get; set; } = true;
+        
+        [DefaultValue(true)]
```
hishamco, line 13: This is the first time I saw this attribute, why it's useful, while we assign a value to the property itself?

deanmarcussen: It's required by the serializer for content type / part settings.

https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_DefaultValueHandling.htm

hishamco (approved):

Issue #7427, opened 2020-10-29
Update DefaultMediaFileStoreCacheFileProvider to delete files on request aborted


deanmarcussen:
Discussed on gitter

In
`SetCacheAsync(Stream stream, IFileStoreEntry fileStoreEntry, CancellationToken cancellationToken)`

we should try catch the File Write, and when connection aborted, or exception thrown, we should delete the file.

It may have been created with a zero length, or just aborted, which will prevent the next read from rereading it.

It should rethrow the exception.

PR #7489, opened 2020-11-02
Cleanup partial blob storage cache download

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/7427

I am not checking the cancellation token or for an `OperationCanceledException` as this is not what azure storage will throw if the blob storage request fails due to network conditions or...

So on any exception seemed better,

### DIFF
--- a/src/OrchardCore/OrchardCore.Media.Core/DefaultMediaFileStoreCacheFileProvider.cs
+++ b/src/OrchardCore/OrchardCore.Media.Core/DefaultMediaFileStoreCacheFileProvider.cs
@@ -56,9 +56,19 @@ namespace OrchardCore.Media.Core
                 Directory.CreateDirectory(directory);
             }
 
-            using (var fileStream = File.Create(cachePath))
+            // A file download may fail, so a partially downloaded file should be deleted so the next request can reprocess.
+            try
             {
-                await stream.CopyToAsync(fileStream, StreamCopyBufferSize, cancellationToken);
+                using (var fileStream = File.Create(cachePath))
+                {
+                    await stream.CopyToAsync(fileStream, StreamCopyBufferSize, cancellationToken);
+                }
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "Error saving file {Path}", cachePath);
+                File.Delete(cachePath);
+                throw;
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/OrchardCore/OrchardCore.Media.Core/DefaultMediaFileStoreCacheFileProvider.cs
```diff
@@ -56,9 +56,19 @@ public async Task SetCacheAsync(Stream stream, IFileStoreEntry fileStoreEntry, C
                 Directory.CreateDirectory(directory);
             }
 
-            using (var fileStream = File.Create(cachePath))
+            // A file download may fail, so a partially downloaded file should be deleted so the next request can reprocess.
+            try
+            {
+                using (var fileStream = File.Create(cachePath))
```
sebastienros, line 62: Check if the file exists before creating it, then delete it. That should save some exceptions.

src/OrchardCore/OrchardCore.Media.Core/DefaultMediaFileStoreCacheFileProvider.cs
```diff
@@ -56,9 +56,19 @@ public async Task SetCacheAsync(Stream stream, IFileStoreEntry fileStoreEntry, C
...
+                {
+                    await stream.CopyToAsync(fileStream, StreamCopyBufferSize, cancellationToken);
+                }
+            }
+            catch (Exception ex)
             {
-                await stream.CopyToAsync(fileStream, StreamCopyBufferSize, cancellationToken);
+                _logger.LogError(ex, "Error saving file {Path}", cachePath);
+                File.Delete(cachePath);
```
sebastienros, line 70: Check if the file exists. And do a try catch also on the File.Delete, just logging the error.
Maybe add a time based token too in case the deletion takes too long (lock, or file system issue).

deanmarcussen: ok. have one. didn't add a token because it felt like it would need a file system watcher, which seems like more than what we want. The issue looks to be less around file system, and more around network failures during transit from blob storage. (so partially completed). 
We might be able to track that better through the `BlobFileStore` but I think this is the better place to handle it

sebastienros: You can create a custom cancelation token that is based on time, the idea is that if the file is locked and you can't delete it, it won't block for too long: https://docs.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource.cancelafter?view=net-5.0#System_Threading_CancellationTokenSource_CancelAfter_System_Int32_

PR #7438, opened 2020-10-29
Converted "row col-xx-xx" to "w-xx-xx" and moved hints outside the column

netwavebe:
A lot of xxxSettings.Edit.cshtml views had this markup in place:

```
<div class="row col-medium">
</div>
```

Since row has margin-right and margin-left -15px and col has padding-right and padding-left 15px an element should not contain both classes. Columns should always be child elements of a row.

I converted the markup to:

```
<div class="row">
    <div class="col-medium">
    </div>
</div>
```

I also moved any hints placed inside a column so that they are not restricted by the width of the input-element.

furthermore I nomalized all hint inputs to `col-medium`.

Random screenshot of Numeric field settings:

Before:

image

After:

image

Right margin is now always the same.

PR #7495, opened 2020-11-03
Add padding on other inputs on FieldSettings forms

Skrypt:
Follow up on #7438 
Some inputs are now larger than the one's we fixed.
Adding pr-md-3 to them so that everything be the same size.

/cc @netwavebe


netwavebe:
Tried this with several fields, I think everything is now 100% correct aligned in all cases! :ok_hand:

PR #7498, opened 2020-11-03
Fix html encoding in LiquidViewFilters

### LINKED ISSUE #7497: Wrong html entities in <title> tag after PR 7463
We have spotted an issue with [PR 7463](https://github.com/OrchardCMS/OrchardCore/pull/7463).

Before this change, our <title> tags with special characters were rendered just fine (plain raw text). Now we have them shown as hex html entities but the **&** character is being escaped, so they are not rendered properly:

image

If you go to the actual html output of the page it has:

image

but it should have:
`<title>P&#xE1;gina de prueba - my-site</title>`

Summing up, the **&** character in the `&#xE1;` entity should not be escaped to `&amp;xE1;`

BTW, the <title> tag in TheBlogTheme is rendered with (which is fine):
`<title>{{ "PageTitle" | shape_new | shape_stringify }}</title>`

//cc @jtkech

hishamco:
Fixed #7497

@jagbarcelo could please try the PR

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/LiquidViewFilters.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/LiquidViewFilters.cs
@@ -109,7 +109,12 @@ namespace OrchardCore.DisplayManagement.Liquid.Filters
                 StringValue value;
                 using (var writer = new StringWriter())
                 {
-                    var htmlEncoder = ShellScope.Services.GetRequiredService<HtmlEncoder>();
+                    if (!context.AmbientValues.TryGetValue("Services", out var services))
+                    {
+                        throw new ArgumentException("Services missing while invoking 'LiquidView'");
+                    }
+
+                    var htmlEncoder = ((IServiceProvider)services).GetRequiredService<HtmlEncoder>();
                     task.Result.WriteTo(writer, htmlEncoder);
                     value = new StringValue(writer.ToString());
                 }
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/LiquidViewFilters.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/LiquidViewFilters.cs
@@ -2,12 +2,14 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Runtime.CompilerServices;
+using System.Text.Encodings.Web;
 using System.Threading.Tasks;
 using Fluid;
 using Fluid.Values;
 using Microsoft.AspNetCore.Html;
 using Microsoft.AspNetCore.Mvc.Localization;
-using Microsoft.AspNetCore.Razor.TagHelpers;
+using Microsoft.Extensions.DependencyInjection;
+using OrchardCore.Environment.Shell.Scope;
 using OrchardCore.Mvc.Utilities;
 
 namespace OrchardCore.DisplayManagement.Liquid.Filters
@@ -107,7 +109,8 @@ namespace OrchardCore.DisplayManagement.Liquid.Filters
                 StringValue value;
                 using (var writer = new StringWriter())
                 {
-                    task.Result.WriteTo(writer, NullHtmlEncoder.Default);
+                    var htmlEncoder = ShellScope.Services.GetRequiredService<HtmlEncoder>();
+                    task.Result.WriteTo(writer, htmlEncoder);
                     value = new StringValue(writer.ToString());
                 }
 


src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/LiquidViewFilters.cs
```diff
@@ -107,7 +109,8 @@ static async ValueTask<FluidValue> Awaited(Task<IHtmlContent> task)
                 StringValue value;
                 using (var writer = new StringWriter())
                 {
-                    task.Result.WriteTo(writer, NullHtmlEncoder.Default);
+                    var htmlEncoder = ShellScope.Services.GetRequiredService<HtmlEncoder>();
```
jptissot, line 112: You should use the AmbientValues of the context to get a service. Not the ShellScope.

Look at the ShortCodeFilter for an example.

jptissot (changes requested):

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/LiquidViewFilters.cs
hishamco: > You should use the AmbientValues of the context to get a service. Not the ShellScope.

Is there any difference?

jagbarcelo:
I've applied the same changes (including the use AmbientValues) to my dev machine and it still shows the same output:

image

I think it is being encoded twice, thus the `&amp;` instead of a simple `&`. Might it be re-encoded elsewhere, up in the pipeline?

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/LiquidViewFilters.cs
jptissot: I am not aware what the difference is, but it removes the need to depend on the ShellScope. I can only assume it allows for use in different contexts and also better testing (without having to implement a shell scope).

jtkech:
Before the regression, `shape_render` and `shape_stringify` were both returning a `StringValue`, not encoded for `shape_render`, and encoded for `shape_stringify`, that's why i still encoded it in my PR, Notice that `StringValue` is encoded by default so no need to encode it again. After the regression, `shape_render` and `shape_stringify` were both returning an `IHtmlContent` and in fact were doing exactly the same thing.

So currently, when `shape_stringify` is used for direct rendering there are 2 workarounds

In place of using `shape_stringify` we can just use `shape_render` that is doing the same thing as the previous stringify

    <title>{{ "PageTitle" | shape_new | shape_render }}</title>

Or add the raw filter to prevent double encoding

    <title>{{ "PageTitle" | shape_new | shape_stringify | raw }}</title>

Then we would have to update the layout in our themes.

---

**But stringify is intended to be used as a string not for direct rendering, e.g. to be passed as a parameter to another shape, so maybe before the regression there was another issue by encoding it. So, finally here i suggest to just replace**

                    value = new StringValue(writer.ToString());

With

                    value = new StringValue(writer.ToString(), false);

I tried it, it works for the related issue.

sebastienros:
I agree with the fact that `stringify`, if it's returning html, should not be encoded. But not sure that the result string should know about it, or actually its usage in the title and then use `| raw`. Need to find other examples of stringify, and what the doc states. Or watch the meeting where it was introduced. I assume I wrote this one, or you maybe JT?

Skrypt:
@sebastienros It was written because we had issues with the Lucene/SQL Query templates (liquid). We needed something to get a raw value from a form posted value.

sebastienros:
Take a look at this sample:

https://github.com/OrchardCMS/OrchardCore/blob/95b33018bad36c431cadf7c638851a4237a4903e/src/OrchardCore.Themes/TheBlogTheme/Views/Content-BlogPost.Summary.liquid#L12-L13

It's using `shape_stringify` and then calls `| raw` of the result. So it's expecting it to be html, but not to be a `StringValue(encode: false)`.

jtkech:
Yes i saw this and tried it, here even we change it to `StringValue(..., false)` we still need the raw filter

Keep in mind that before #7463 that i did to fix an issue comented on gitter, `shape_render` and `shape_stringify` were doing exactly the same thing, this after this commit https://github.com/OrchardCMS/OrchardCore/commit/44754f60b42a2bd7f8151eb461654ea6a8bda549#diff-8141fd33486da211684152b5d06ea47dc755a10fd523d5fceeff0f0e8ff2d112, so at least one of them was useless.

Before the above commit there were both reurning a `StringValue` but `shape_stringify` was (maybe wrongly) encoded, that's why i kept it encoded in #7463

hishamco:
If no encoding is required I will close this PR

jtkech:
@sebastienros about `{{ "on {0}" | t: dateTime | raw }}`

> Well it's a translation argument ... might be even more complex

Yes just tried, in fact here the raw is not applied to the dateTime argument, but to the whole, here this is the `t` filter that encodes, anyway the `t` filter uses the inner `_value` of the passed argument regardless it is marked to be encoded.

`shape_stringify` should not be marked to be encoded as it is the representation of a shape, so normally already encoded, hmm, so not sure that the result of a shape stringify is a good candidate for a `t` parameter, as it is already encoded, that's why here we need anyway the raw filter.

`shape_render` return an `HtmlContentValue` holding a `ViewBuffer` whose `WriteTo()` doesn't encode again, that's why it was working for #7497 with the previous `shape_stringify`, before i did #7463, because it was doing exactly the same thing, so useless.

But there was another issue on gitter where someone really want to work with the returned value expecting it is a string (not a `ViewBuffer`), as it was [before this commit](https://github.com/OrchardCMS/OrchardCore/commit/44754f60b42a2bd7f8151eb461654ea6a8bda549#diff-8141fd33486da211684152b5d06ea47dc755a10fd523d5fceeff0f0e8ff2d112L75) (LiquidViewFilters.cs line 75)

That's why i did #7463 to really return a string, but unfortunately before the above commit `shape_stringify` was wrongly returning a string marked to be encoded again, that's why i just did the new #7503 PR to fix it.

I think there are other filters that wrongly return a string marked to be encoded, i'm thinking about those that generate urls, maybe currently it works because the browser is aware of this and html decodes them before url encoding them.

agriffard:
Fixed by #7503

PR #7499, opened 2020-11-03
CodeMirror min-height to match bootstrap

deanmarcussen:
The minimum height for a code mirror field has been annoying me for a while. 

This just sets the min-height to match the same height as a standard bootstrap input field, which means the fields look better when displayed together on a content type (and values are empty)

Before

Screenshot 2020-11-03 at 10 45 53

After

Screenshot 2020-11-03 at 10 46 16


src/OrchardCore.Themes/TheAdmin/Assets/scss/admin.scss
```diff
@@ -11,8 +11,8 @@
 @import "modules/admin/_grid.scss";
 @import "modules/admin/_tabs.scss";
 @import "modules/admin/_login.scss";
-@import "modules/admin/_switch.scss";
 @import "modules/admin/_codemirror.scss";
+@import "modules/admin/_switch.scss";
```
deanmarcussen, line 15: Moved because the switch overrides the default boostrap input height

Issue #7451, opened 2020-10-30
Apply a recursion counter to the recent build_display filter recursion limiter


deanmarcussen:
Feature added here https://github.com/OrchardCMS/OrchardCore/pull/7357

Need to add counter to it to allow shapes to be built more than once (think bag items, in different display modes).

But still limit the amount of times this can happen to prevent stackoverflows when accidentally called recursively from within the same content template.

jeffolmstead:
As commented on Gitter, my breaking example is:

```
{% for blogPost in recentBlogPosts %}
  {{ blogPost | shape_build_display: "Featured" | shape_render }}
  {{ blogPost | shape_build_display: "Thumbnail" | shape_render }} 
{% endfor %}
```

Thanks for tackling this.

sebastienros:
how is Jeff's bug related to full text? I assumed the recursion limit was only applied to full text generation which is recursive. Unless it has also been implemented to shape display, in which case the key should take the display type into account. But I assume a counter works too, as long as it's a setting.

deanmarcussen:
It slipped in because I tested the other content filters at the same time, and discovered the issue with the `build_display` filter.

I used a counter rather than a display type, as there's been times (in Razor) I've built the display type twice, and then mutated the shapes to display differently (breadcrumbs, tiles, and list items).

So I set it at 20 - which feels really high, but this was about preventing accidental mistakes bringing the site down. It quickly becomes obvious that a mistake in the template has been made.

Could be an Option so it's configurable.

PR #7500, opened 2020-11-03
Limit content item recursions to 20

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/7451

@jtkech I ended up using both a dictionary and hashset, because comparing by ref wasn't possible, because when it is recursing like this it's creating new shapes everytime, so new objects getting spawned.

I picked 20 as the number, as @jeffolmstead has already experienced the issue you spotted, trying to build multiple display types, on the same content item.

The other option is to remove recursion checking here on this filter completely.

But a bit of bad code, i.e. `{{ Model.ContentItem | shape_build_display | shape_render }}` does stackoverflow the system.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Liquid/BuildDisplayFilter.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Liquid/BuildDisplayFilter.cs
@@ -14,6 +14,8 @@ namespace OrchardCore.Contents.Liquid
 {
     public class BuildDisplayFilter : ILiquidFilter
     {
+        private const int MaxContentItemRecursions = 20;
+
         public ValueTask<FluidValue> ProcessAsync(FluidValue input, FilterArguments arguments, TemplateContext ctx)
         {
             static async ValueTask<FluidValue> Awaited(Task<IShape> task)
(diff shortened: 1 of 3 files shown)


jtkech:
@deanmarcussen 

Okay for the counters per items, i would not have used an `HashSet<string>` but only a dictionary which is a kind of hashtable and fast enough, otherwise LGTM.

Not so important but can you remind me why we can't compare the item by ref, i understand if it is a new instance built from a JObject, but i didn't find how to do an infinite recursion on the same item id but never the same item instance. So okay, with the following i can see a recursion but always on the same item instance

    {{ Model.ContentItem | shape_build_display | shape_render }}

Then i tried the following with the agency theme, indeed, because built from a JObject i can see different instances, but the recursions count is only equals to the explicit calls that we do, it's not an automatic infinite recursion

            {% for service in Model.ContentItem.Content.Services.ContentItems %}
                {{ service | shape_build_display | shape_render }}
                {{ service | shape_build_display | shape_render }}
                {{ service | shape_build_display | shape_render }}

The only way i can find to do automatically an infinite recursion, is when this is the current shape that renders a JObject item, or its parent shape, that is recursively called, so normally our existing recursion checking, that compares by ref the current model / shape, would have seen that it renders itself (max recursions = 3).

deanmarcussen:
@jtkech 

by ref does work! because as you say the only time this can happen is when it is the `Model.ContentItem` which is always the same object, not serialized as new from JSON. 

Validated with both `full_text` and `shape_build_display`

Also added as @sebastienros wanted it configurable is a `max_recursions` argument.

If you can't see any other issues can you merge this tonight, as it is a little broken right now. Cheers.

jtkech:
@deanmarcussen okay

So, if `max_recursions` is an argument that may have different values, potentialy the counter may be already greater than a current passed value, so maybe i will replace `if (counter == maxRecursions)` with `if (counter >= maxRecursions)`

Or maybe with `if (counter > maxRecursions)` because the first call is not a first recursion ;)

jtkech (approved):

PR #7502, opened 2020-11-03
Revert code mirror min-height

deanmarcussen:
This turned out to be a bad idea
https://github.com/OrchardCMS/OrchardCore/pull/7499

It stalls CodeMirror when the document is over 20 lines (i.e. the agency template), and the code mirror window doesn't resize properly until you click on it. And even then it's still not the entire height of the page.


jptissot (approved):

Skrypt:
Open an issue so that I remember to look at it.

Issue #5494, opened 2020-02-08
IsInRole Layer rule


agriffard:
There should be a `IsInRole(...)` layer rule that allow you to return true if the current user is in a specific role.

matiaskusack:
Hi I was working in a ContetType rule (like Orchard 1) bit I was stocked on finding the current ContentItem.
I might be able to implement this togheter with ContentType rule I think.

PR #7507, opened 2020-11-04
Add 'isInRole' layer rule & make 'is_in_role' liquid filter case insensitive

netwavebe:
This PR adds an 'isInRole' layer rule. I also updated the documentation.

I noticed `IsInRole(role)` is case SENSITIVE. So `isInRole('editor')` does not match role `Editor`. I don't know if it should stay this way, or be made case insensitive?

This fixes https://github.com/OrchardCMS/OrchardCore/issues/5494

/cc @agriffard

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Layers/Services/DefaultLayersMethodProvider.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Layers/Services/DefaultLayersMethodProvider.cs
@@ -12,6 +12,7 @@ namespace OrchardCore.Layers.Services
         private readonly GlobalMethod _isHomepage;
         private readonly GlobalMethod _isAnonymous;
         private readonly GlobalMethod _isAuthenticated;
+        private readonly GlobalMethod _isInRole;
         private readonly GlobalMethod _url;
         private readonly GlobalMethod _culture;
 
@@ -47,6 +48,16 @@ namespace OrchardCore.Layers.Services
                 {
                     var httpContext = serviceProvider.GetRequiredService<IHttpContextAccessor>().HttpContext;
                     return httpContext.User?.Identity.IsAuthenticated == true;
+                })
+            };
+
+            _isInRole = new GlobalMethod
+            {
+                Name = "isInRole",
+                Method = serviceProvider => (Func<string, bool>) (role =>
+                {
+                    var httpContext = serviceProvider.GetRequiredService<IHttpContextAccessor>().HttpContext;
+                    return httpContext.User?.IsInRole(role) == true;
                 })
             };
 
@@ -88,7 +99,7 @@ namespace OrchardCore.Layers.Services
                 })
             };
 
-            _allMethods = new[] { _isAnonymous, _isAuthenticated, _isHomepage, _url, _culture };
+            _allMethods = new[] { _isAnonymous, _isAuthenticated, _isInRole, _isHomepage, _url, _culture };
         }
 
         public IEnumerable<GlobalMethod> GetMethods()
--- a/src/OrchardCore.Modules/OrchardCore.Layers/Services/DefaultLayersMethodProvider.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Layers/Services/DefaultLayersMethodProvider.cs
@@ -1,8 +1,11 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Identity;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
 using OrchardCore.Scripting;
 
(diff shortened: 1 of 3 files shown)


src/OrchardCore.Modules/OrchardCore.Layers/Services/DefaultLayersMethodProvider.cs
```diff
@@ -50,6 +51,16 @@ public DefaultLayersMethodProvider()
                 })
             };
 
+            _isInRole = new GlobalMethod
+            {
+                Name = "isInRole",
+                Method = serviceProvider => (Func<string, bool>) (role =>
+                {
+                    var httpContext = serviceProvider.GetRequiredService<IHttpContextAccessor>().HttpContext;
+                    return httpContext.User?.IsInRole(role) == true;
```
agriffard, line 60: I am not a big fan of `== true`.

netwavebe: It's needed because the `?` after `User` makes it a `bool?` instead of a `bool`.

hishamco: What about 

```
httpContext.User?.IsInRole(role) ?? false;
```

netwavebe: Makes it less readable imho. Also, `== true` is in line with the other rules. 

https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.Layers/Services/DefaultLayersMethodProvider.cs#L49

MichaelPetrinolis: There are already defined some liquid filters in Users module. We could add the scripting functions in that module, be in parity with what provided for users in liquid and even reuse some code. see https://github.com/OrchardCMS/OrchardCore/blob/54658bc8a427bbe1f6f3eb1f9f325a38204b3849/src/OrchardCore.Modules/OrchardCore.Users/Liquid/IsInRoleFilter.cs#L11

Skrypt:
The case sensitivity issue should be reported on the asp.net core repository if not already reported.

src/OrchardCore.Modules/OrchardCore.Layers/Services/DefaultLayersMethodProvider.cs
netwavebe: > If User is null, I think that the httpContext.User?.IsInRole(role) will throw NRE

No, the `?` handles that.

netwavebe:
@Skrypt So you agree it should be case insensitive?

There are a few ways to fix this. The easiest one is probably something like this:

`return httpContext.User?.Claims.Where(c => c.Type == ClaimTypes.Role).Any(c => c.Value.Equals(role, StringComparison.InvariantCultureIgnoreCase)) == true;`

Skrypt:
UserNames are normalized from the asp.net core UserManager so I think it should be case insensitive.

jtkech:
Also if possible use `OrdinalIgnoreCase`in place of `InvariantCultureIgnoreCase`.

src/OrchardCore.Modules/OrchardCore.Layers/Services/DefaultLayersMethodProvider.cs
sebastienros: I agree with @MichaelPetrinolis on the NRE. `httpContext.User?.IsInRole()` is not the same as `httpContext?.User?.IsInRole`

netwavebe: @sebastienros @MichaelPetrinolis User can be `null` but isn't it safe to assume httpContext is never `null` in OC? If not, I need to update `isAnonymous` and `isAuthenticated` too.

netwavebe:
Made it case insensitive. Also used `IOptions<IdentityOptions>` to get role claim type.

netwavebe:
Also made the '_is_in_role' liquid filter case insensitive in the same way.

MichaelPetrinolis:
If Microsoft's default API functions HasClaim and IsInRole are case sensitive, and we search for the same kind of information in a case insensitive manner, might lead to side effects.

netwavebe:
@MichaelPetrinolis Check @Skrypt's link. Stores are case sensitive, but managers are not. So `UserManager.IsInRole("administrator")` would return true even if the role was named `Administrator`.

When it comes to layers (and liquid), I don't think there should be a difference between `IsIsRole("Editor")` and `IsInRole("editor")` as there can only be one role named `Editor` or `editor`.

MichaelPetrinolis:
Looks good then. from scripting and liquid is like calling the manager APIs

PR #7511, opened 2020-11-04
Admin Menu Permissions

### LINKED ISSUE #7284: Admin menu: unable to set Permission over item from recipe
Hi All, 

I have created and admin menu with this item: 
```
{
              "$type": "OrchardCore.AdminMenu.AdminNodes.LinkAdminNode, OrchardCore.AdminMenu",
              "LinkText": "Libreria Multimediale",
              "LinkUrl": "Admin/Media",
              "IconClass": "far fa-images",
              "UniqueId": "[js:uuid()]",
              "Enabled": true,
              "Text": null,
              "Id": "media-first-level",
              "Href": null,
              "Url": null,
              "Position": null,
              "LinkToFirstChild": true,
              "LocalNav": false,
              "Culture": null,
              "Resource": null,
              "Items": [],
              "RouteValues": null,
              "Permissions": [{ "name": "ManageMediaContent" }],
              "Classes": ["media-first-level"]
}
```
but when I run my recipe I receive this error: 
```
2020-10-15 19:09:54.6967|Default||1b4aa20d-49cd3a81eecc7d6b.||Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware|ERROR|An unhandled exception has occurred while executing the request. Newtonsoft.Json.JsonSerializationException: Unable to find a constructor to use for type OrchardCore.Security.Permissions.Permission. A class should either have a default constructor, one constructor with arguments or a constructor marked with the JsonConstructor attribute. Path '[0].MenuItems[9].Permissions[0].name'.
```

Skrypt:
WIP

screencapture-localhost-5001-Admin-AdminMenu-Node-Edit-2020-11-04-13_09_11

Fixes #7284

/cc @PiemP


Skrypt:
I've not tested recipes do they work?

src/OrchardCore.Modules/OrchardCore.AdminMenu/AdminNodes/LinkAdminNodeDriver.cs
```diff
@@ -23,20 +41,73 @@ public override IDisplayResult Edit(LinkAdminNode treeNode)
...
+                            treeNode.Permissions.Add(perm);
+                        }
+                    }
+                }
             };
 
             return Edit(treeNode);
         }
+
+        private async Task<IEnumerable<Permission>> GetInstalledPermissionsAsync()
```
Skrypt, line 94: This could be moved in an abstraction module.

PiemP:
> I've not tested recipes do they work?

Yes, I have test it whit the recipe in the issue

Skrypt:
Yeah I just saw you marked the Permission with a JsonContructor attribute that might be the fix.

src/OrchardCore.Modules/OrchardCore.AdminMenu/Controllers/NodeController.cs
```diff
@@ -344,5 +352,27 @@ public async Task<IActionResult> Toggle(string id, string treeNodeId)
 
             return Ok();
         }
+
+
+        public async Task<IActionResult> SearchPermissions(string query)
```
Skrypt, line 357: This needs to be moved in an abstraction for reusability I think too. And this needs to be secured. Maybe adding the [admin] attribute.

Skrypt: Nevermind the controller has it.

src/OrchardCore.Modules/OrchardCore.AdminMenu/AdminNodes/LinkAdminNodeDriver.cs
PiemP: `OrchardCore.Infrastructure.Abstractions`? I create a service to do this Job and load the service where it need? About `vue-multiselect-wrapper.js`, what did you think? I have copied it from the `OrchardCore.ContentFields` module but now I believe exist 2 or 3 copy of this file in the project.

src/OrchardCore.Modules/OrchardCore.AdminMenu/Controllers/NodeController.cs
```diff
@@ -344,5 +352,27 @@ public async Task<IActionResult> Toggle(string id, string treeNodeId)
 
             return Ok();
         }
+
+
```
deanmarcussen, line 356: you don't need to make an ajax call at all. or limit the permissions returned. they're static (so it's not an open ended sql call), so just return them all from the driver in one hit.

src/OrchardCore.Modules/OrchardCore.AdminMenu/Controllers/NodeController.cs
Skrypt: Probably something like a static class in OrchardCore.Infrastructure.Abstraction/Security/Services/PermissionsHelper.cs
But for now I guess it's fine.

src/OrchardCore.Modules/OrchardCore.AdminMenu/AdminNodes/LinkAdminNodeDriver.cs
Skrypt: Yes that would make sense PermissionsService.cs

src/OrchardCore.Modules/OrchardCore.AdminMenu/Controllers/NodeController.cs
Skrypt: I agree on that too.

Skrypt:
It misses the permission picker on the Placeholder menu item and also on the menu themselves.
I've not tested yet but I assume that the Content Type menu items should also follow the content type permissions by using `_authorizationService.AuthorizeContentTypeDefinitionsAsync`

PiemP:
> So the logic is that even the menu itself is a **MenuItem**. If we follow this logic then ; it misses the permission picker on the **Placeholder** menu item and also on the **menu** themselves.
> 
> I've not tested yet but I assume that the Content Type menu items should also follow the content type permissions by using `_authorizationService.AuthorizeContentTypeDefinitionsAsync`.

OK, I'm agree to add permission picker to placeholder and seem this object already support the Permissions system but I don't have understood what did you mean with "and also on the **menu** themselves"? Did you want to add permission picker to the entire menu?

PiemP:
If I allow user to add multiple permission to a `MenuItem`, how have I evaluate this permissions? In AND way (the user must have all the permissions to see the element) or in OR way (the user need to have at least one permission to see the link)? Or I have to create a checkbox to allow a choice about this.

If I have to create a checkbox I need to add ad-hoc property in the `MenuItem` class to store this value.

Skrypt:
Let me review this a little more because I did not have time yet. I will get back to you with some answers. I think we agree to add the permission picker on the placeholder. For the Menu itself I need to check if it is a root MenuItem. But feels like it should be to me. Else, if it's not then each Menu Items permissions should be enough. Yes, somehow, we should only display Menu items that the user has permission on. Else clicking on the Menu Item will return a 403 error page which we want to avoid.

PiemP:
> For the Menu itself I need to check if it is a root MenuItem. But feels like it should be to me. Else, if it's not then each Menu Items permissions should be enough. 

I have made a couple of check and I believe that doesn't need to add permission picker to an admin menu because already exist this functionality. In the edit role interface you can find a  permission to see all the admin menues or a specific admin menu.

> Evaluation of multiple permissions on a MenuItem "not sure yet". Maybe we could add this in a second PR?

In this moment the code in this PR already manage multiple permissions related to a `MenuItem` and the code evaluate this permissions with a logical AND between all the permissions selected.

Skrypt:
Ok, I will test it one more time today to make sure everything is working. Looking good so far. 😉

src/OrchardCore.Modules/OrchardCore.AdminMenu/AdminNodes/LinkAdminNodeDriver.cs
```diff
@@ -23,17 +36,61 @@ public override IDisplayResult Edit(LinkAdminNode treeNode)
...
+
+                    model.SelectedItems.Add(new VueMultiselectItemViewModel
+                    {
+                        Id = permission.Name,
+                        DisplayText = $"{permission.Name} - {permission.Description}"
+                    });
+                }
+                model.PermissionIds = string.Join(",", nameList);
+                
+                foreach (var permission in _permissionService.GetInstalledPermissionsAsync().Result)
```
jtkech, line 55: For info, if it fails exceptions are better managed by using `.GetAwaiter().GetResult()` in place of `.Result`

But here better to override `DisplayAsync()` and then use a regular async / await

src/OrchardCore.Modules/OrchardCore.AdminMenu/AdminNodes/PlaceholderAdminNodeDriver.cs
```diff
@@ -22,16 +34,59 @@ public override IDisplayResult Edit(PlaceholderAdminNode treeNode)
...
+
+                    model.SelectedItems.Add(new VueMultiselectItemViewModel
+                    {
+                        Id = permission.Name,
+                        DisplayText = $"{permission.Name} - {permission.Description}"
+                    });
+                }
+                model.PermissionIds = string.Join(",", nameList);
+                
+                foreach (var permission in _permissionService.GetInstalledPermissionsAsync().Result)
```
jtkech, line 53: For info, if it fails exceptions are better managed by using `.GetAwaiter().GetResult()` in place of `.Result`

But here better to override `DisplayAsync()` and then use a regular async / await

src/OrchardCore.Modules/OrchardCore.AdminMenu/Views/Items/PlaceholderAdminNode.Fields.TreeEdit.cshtml
```diff
@@ -64,6 +66,59 @@
     </div>
 </div>
 
+<div class="form-group">
+    @{
+        var selectedItems = JsonConvert.SerializeObject(Model.SelectedItems, new JsonSerializerSettings { ContractResolver = new CamelCasePropertyNamesContractResolver() });
+        var allItems = JsonConvert.SerializeObject(Model.AllItems, new JsonSerializerSettings { ContractResolver = new CamelCasePropertyNamesContractResolver() });
+    }
+
+    <script asp-name="vue-multiselect-wrapper" asp-src="~/OrchardCore.AdminMenu/Scripts/vue-multiselect-wrapper.js" at="Foot" depends-on="vuejs, vue-multiselect, sortable, vuedraggable"></script>
```
deanmarcussen, line 75: We shouldn't use the same name here.
It won't matter much on this page, but it's technically a conflict with the content fields resource name.

So should be named differently.

src/OrchardCore/OrchardCore.Infrastructure.Abstractions/Security/Services/IPermissionsService.cs
```diff
@@ -0,0 +1,51 @@
...
+            _permissionProviders = permissionProviders;
+            _typeFeatureProvider = typeFeatureProvider;
+        }
+
+        public async Task<IEnumerable<Permission>> GetInstalledPermissionsAsync()
+        {
+            var installedPermissions = new List<Permission>();
+            foreach (var permissionProvider in _permissionProviders)
+            {
+                var feature = _typeFeatureProvider.GetFeatureForDependency(permissionProvider.GetType());
```
deanmarcussen, line 37: unusued variable

src/OrchardCore/OrchardCore.Infrastructure.Abstractions/Security/Services/IPermissionsService.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    /// <summary>
+    /// Service that returns the installed permissions
+    /// </summary>
+    public interface IPermissionsService
+    {
+        Task<IEnumerable<Permission>> GetInstalledPermissionsAsync();
+    }
+
+    public class BasePermissionsService : IPermissionsService
```
deanmarcussen, line 18:
```suggestion
    public class DefaultPermissionsService : IPermissionsService
```

deanmarcussen: Also this class shouldn't be in the abstractions project.

The interface should, but the implementation should be in the `Infrastructure` project

src/OrchardCore/OrchardCore.Navigation.Core/NavigationManager.cs
```diff
@@ -227,12 +227,16 @@ private async Task<List<MenuItem>> AuthorizeAsync(IEnumerable<MenuItem> items, C
                 }
                 else
                 {
+                    //evaluate permissions in AND way
```
deanmarcussen, line 230: I don't totally understand this change. Can you expand the comment?

deanmarcussen (commented):
Just a quick look @Skrypt and @PiemP 

I want to check something else about the `JsonConstructor` and how it will related to object references - tomorrow though :) 

Also can you format the files with Ctrl K + D @PiemP ?

Thanks

src/OrchardCore/OrchardCore.Navigation.Core/NavigationManager.cs
Skrypt: Because the Permission selector is a multi-permission selector. Maybe it should be a single permission selector only.

deanmarcussen: Ah, I see.

I think it should probably be a single selector (that's worked perfectly well up to now? I think?)

Unless @PiemP you can find me an example where a menu should be displayed based on two different permissions?

PiemP: Honestly I don't have an example. If someone create a module that do something strange over a the items of a specific content type and the user want to show the menu item only if the user can publish and see all the content item of this content type...I don't know...Probably it's better to ask it to the guy that have create the `MenuItem` object with a structure for the property `Permissions`. In this moment I don't have this needs in my site.

For my opinion the only issue about that is about the evaluation of multiple permissions: that is done only in "AND way" and the user can't choose to change this behavior. We can only loose a possibility, if we remove this code. And ff we limit the permission picker to select only one permission? In the mean time we can think about what do with this.  

Thank you guys, I hope to be explained me well. 

PS. this code fix an issue. Before, if someone add plus than one permission to a MenuItem, this menu item was show multiple time.

Skrypt: Let's keep it and see then. I totally agree that having it is better than not having it. We will see over time how people will use it.

src/OrchardCore.Modules/OrchardCore.AdminMenu/Views/Items/PlaceholderAdminNode.Fields.TreeEdit.cshtml
```diff
@@ -64,6 +66,59 @@
...
+
+    <script asp-name="vue-multiselect-permission-picker" asp-src="~/OrchardCore.AdminMenu/Scripts/vue-multiselect-permission-picker.js" at="Foot" depends-on="vuejs, vue-multiselect, sortable, vuedraggable"></script>
+    <style asp-name="vue-multiselect-permission-picker" asp-src="~/OrchardCore.AdminMenu/Styles/vue-multiselect-permission-picker.min.css" debug-src="~/OrchardCore.AdminMenu/Styles/vue-multiselect-permission-picker.css" depends-on="vue-multiselect"></style>
+
+    <label asp-for="PermissionIds">@T["Permissions"]</label>
+
+    <div id="PermissionPicker" class="vue-multiselect" data-editor-type="ContentPicker" data-selected-items="@selectedItems" data-all-items="@allItems" data-multiple="true">
+
+        <div class="form-group">
+            <ul class="list-group w-xl-50 content-picker-default__list" v-show="arrayOfItems.length" v-cloak>
```
deanmarcussen, line 83: `content-picker-default__list`

src/OrchardCore.Modules/OrchardCore.AdminMenu/Views/Items/PlaceholderAdminNode.Fields.TreeEdit.cshtml
```diff
@@ -64,6 +66,59 @@
...
+        var selectedItems = JsonConvert.SerializeObject(Model.SelectedItems, new JsonSerializerSettings { ContractResolver = new CamelCasePropertyNamesContractResolver() });
+        var allItems = JsonConvert.SerializeObject(Model.AllItems, new JsonSerializerSettings { ContractResolver = new CamelCasePropertyNamesContractResolver() });
+    }
+
+    <script asp-name="vue-multiselect-permission-picker" asp-src="~/OrchardCore.AdminMenu/Scripts/vue-multiselect-permission-picker.js" at="Foot" depends-on="vuejs, vue-multiselect, sortable, vuedraggable"></script>
+    <style asp-name="vue-multiselect-permission-picker" asp-src="~/OrchardCore.AdminMenu/Styles/vue-multiselect-permission-picker.min.css" debug-src="~/OrchardCore.AdminMenu/Styles/vue-multiselect-permission-picker.css" depends-on="vue-multiselect"></style>
+
+    <label asp-for="PermissionIds">@T["Permissions"]</label>
+
+    <div id="PermissionPicker" class="vue-multiselect" data-editor-type="ContentPicker" data-selected-items="@selectedItems" data-all-items="@allItems" data-multiple="true">
```
deanmarcussen, line 80: `data-editor-type="ContentPicker"`

src/OrchardCore.Modules/OrchardCore.AdminMenu/Views/Items/LinkAdminNode.Fields.TreeEdit.cshtml
```diff
@@ -71,6 +73,59 @@
...
+        var selectedItems = JsonConvert.SerializeObject(Model.SelectedItems, new JsonSerializerSettings { ContractResolver = new CamelCasePropertyNamesContractResolver() });
+        var allItems = JsonConvert.SerializeObject(Model.AllItems, new JsonSerializerSettings { ContractResolver = new CamelCasePropertyNamesContractResolver() });
+    }
+
+    <script asp-name="vue-multiselect-permission-picker" asp-src="~/OrchardCore.AdminMenu/Scripts/vue-multiselect-permission-picker.js" at="Foot" depends-on="vuejs, vue-multiselect, sortable, vuedraggable"></script>
+    <style asp-name="vue-multiselect-permission-picker" asp-src="~/OrchardCore.AdminMenu/Styles/vue-multiselect-permission-picker.min.css" debug-src="~/OrchardCore.AdminMenu/Styles/vue-multiselect-permission-picker.css" depends-on="vue-multiselect"></style>
+
+    <label asp-for="PermissionIds">@T["Permissions"]</label>
+
+    <div id="PermissionPicker" class="vue-multiselect" data-editor-type="ContentPicker" data-selected-items="@selectedItems" data-all-items="@allItems" data-multiple="true">
```
deanmarcussen, line 87: idem

src/OrchardCore.Modules/OrchardCore.AdminMenu/Views/Items/LinkAdminNode.Fields.TreeEdit.cshtml
```diff
@@ -71,6 +73,59 @@
...
+
+        <div class="form-group">
+            <ul class="list-group w-xl-50 content-picker-default__list" v-show="arrayOfItems.length" v-cloak>
```
deanmarcussen, line 90: idem

src/OrchardCore.Modules/OrchardCore.AdminMenu/Assets/js/vue-multiselect-permission-picker.js
```diff
@@ -0,0 +1,91 @@
...
+                    self.arrayOfItems.push(selectedOption);
+                },
+                remove: function (item) {
+                    this.arrayOfItems.splice(this.arrayOfItems.indexOf(item), 1)
+                }
+            }
+        })
+        
+        /*Hook for other scripts that might want to have access to the view model*/
+        var event = new CustomEvent("vue-multiselect-created", { detail: { vm: vm } });
```
deanmarcussen, line 88: event collision with the content picker

src/OrchardCore.Modules/OrchardCore.AdminMenu/Assets/js/vue-multiselect-permission-picker.js
```diff
@@ -0,0 +1,91 @@
+function debounce(func, wait, immediate) {
```
deanmarcussen, line 1: collision

src/OrchardCore.Modules/OrchardCore.AdminMenu/Assets/js/vue-multiselect-permission-picker.js
```diff
@@ -0,0 +1,91 @@
...
+            timeout = null;
+            if (!immediate) func.apply(context, args);
+        };
+        var callNow = immediate && !timeout;
+        clearTimeout(timeout);
+        timeout = setTimeout(later, wait);
+        if (callNow) func.apply(context, args);
+    };
+};
+function initVueMultiselect(element) {
```
deanmarcussen, line 15: `initPermissionsPicker`

deanmarcussen (commented):
@Skrypt @PiemP 

All good, but this has a lot of references to the content picker js left in it, and a lot of room for naming collisions.

It would be good to fix those I think

PiemP:
Sorry @deanmarcussen, 

My fault. I have prepared a branch to fix this naming conflict. have I create a new PR or can I proceed in the way used by me for update this PR? 

Thank you.

Skrypt:
New PR

PR #7512, opened 2020-11-04
Removed wrong code and fixed code that set permissions to LinkAdminMenu

PiemP:
Removed the property PermissionIds because is not needed

Added the code to effectively set the permissions in the LinkAdminMenu


PiemP:
some fixes for #7511

Skrypt:
Can I merge?

PiemP:
Yes

PR #7517, opened 2020-11-05
Make { "text" | t } work in autoroute pattern.

Skrypt:
Reinsert this feature which was not working before.

See #5626

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
@@ -11,11 +11,13 @@ using Newtonsoft.Json.Linq;
 using OrchardCore.Autoroute.Drivers;
 using OrchardCore.Autoroute.Models;
 using OrchardCore.Autoroute.ViewModels;
+using OrchardCore.ContentLocalization;
 using OrchardCore.ContentManagement;
 using OrchardCore.ContentManagement.Handlers;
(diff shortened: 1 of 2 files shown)


hishamco:
Similar to what I did before, what we are missing before?!!

Skrypt:
There's an other part coming later on. The fix was appropriate but it still doesn't work because we can't localize a Liquid template.

Skrypt:
I looked and we use a IViewLocalizer which inherits from a IHtmlLocalizer which has a method .WithCulture which has been deprecated. They say : 

```
[System.Obsolete("This method is obsolete. Use `CurrentCulture` and `CurrentUICulture` instead.")]
public Microsoft.AspNetCore.Mvc.Localization.IHtmlLocalizer WithCulture (System.Globalization.CultureInfo culture);
```

So basically what I did locally. But I just found out that if you add a LocalizationPart to the LiquidPage Content Type for example. It doesn't take the LocalizationPart into consideration. Right now a LiquidPage renders in all cultures and so is a LiquidTemplate. It just takes the CultureInfo.CurrentUICulture all the time. Now I assume that what @sebastienros was saying is to assign the LiquidViewTemplate.Context.CultureInfo before rendering the Liquid template. But nowhere in the code we use this Context.CultureInfo when rendering the Liquid template.

Skrypt:
So basically, in the same request we should be able to render multiple Liquid Templates in a different culture. The current page culture has nothing to do with it. The LiquidTemplate culture should be defined by the CultureAspect but we probably don't call _contentManager.PopulateAsync() on every time because the Liquid template could be also file based.

PR #7521, opened 2020-11-05
Refactor mail attachments

deanmarcussen:
Feedback from meeting just to refactor it a little bit

/cc @mariojsnunes


mariojsnunes:
Great, thanks that looks better

PR #7528, opened 2020-11-06
Autoroute Pattern localization

### LINKED ISSUE #5599: Autoroute Pattern and {{ | t }} filter

Skrypt:
Fixes #5599


Skrypt:
I looked at this tonight with @jtkech and this is the final solution we agreed on.

src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
```diff
@@ -390,13 +391,20 @@ private async Task GenerateContainerPathFromPattern(AutoroutePart part)
...
                 var cultureAspect = await _contentManager.PopulateAspectAsync(part.ContentItem, new CultureAspect());
-                LiquidViewTemplate.Context.CultureInfo = cultureAspect.Culture;
+
+                CultureInfo.CurrentUICulture = cultureAspect.Culture;
 
                 part.Path = await _liquidTemplateManager.RenderAsync(pattern, NullEncoder.Default, model,
                     scope => scope.SetValue("ContentItem", model.ContentItem));
 
+                // We reassign the proper culture to the current execution flow
+                CultureInfo.CurrentUICulture = currentCulture;
```
hishamco, line 406: We could introduce something like `CultureScope` if this could be used in other places, instead of cache the culture for certain execution

Skrypt: It's already thread safe to do this.

Skrypt: Not sure if this would be necessary. These CultureInfo vars are already thread safe?

hishamco: I can do this if it's useful, which its IMHO ;)

hishamco: > These CultureInfo params are already thread safe?

It save us time instead of doing this over and over in many places

Skrypt: I don't think it is necessary. It would probably just produce "cleaner" code. Propose the changes in a draft PR or ask @sebastienros first if you don't want to waste too much time on this.

hishamco: I will propose it in a draft PR. Regarding `WithCulture` they proposed using `CurrentCulture` because they doing the same, but in our case we need to it for a particular time, then revert the changes, I already have seen similar solution but using attributes  in ASP.NET Core localization repo

jtkech: @Skrypt 

In the meantime, maybe at least a try block and then a finally block to restore the ui culture in case the rendering fails

But i'm not saying to do it yourself and / or now ;)

Skrypt: If the rendering fails at this point won't it throw an exception and simply require to throw a new request which will be set to the proper culture? It's only for AutoroutePartHandler here so nothing to rush about. Let's wait on @hishamco PR to get merged which will do the same.

Skrypt: Here it is. Merge it if you want it : #7539

jtkech: You're right, i was just thinking of the remaining part of the request, e.g. maybe showing the exception in the right culture. You're right let's wait for the PR of @hishamco

Issue #7530, opened 2020-11-06
Support for .Net Core 5?


4kvikram:
Hi All,

We have a project in pipeline and during evaluation, we tried to run OrchardCore with .Net Core 5 and we faced issues.

Question:

Is there a planned update for OrchardCore in near future to support Net Core 5?

jptissot:
See https://github.com/OrchardCMS/OrchardCore/pull/6999

PR #6999, opened 2020-09-02
Upgrade to .NET 5

agriffard:
Todos:
- [ ] Rename .NET Core to .NET in the ReadMe, GitHub baseline, docs, ....
- [ ] Release notes and breaking changes.
- [ ] Specify the minimum version of Visual Studio to use when we are sure. 
- [ ] Pre-list package dependencies that have planned to move to .NET 5 quickly after the release is available.
- [ ] Determine when exactly Azure App Service is able to host .NET 5 applications: https://aspnetcoreon.azurewebsites.net/
- [ ] Tag and branch a version of the code that will be kept on 3.1 for future potential patches. 
- [X] Use Docker mcr.microsoft.com/dotnet/sdk:5.0 and mcr.microsoft.com/dotnet/aspnet:5.0

Use Framework version `5.0.2`.

Replace TargetFramework `netcoreapp3.1` by `net5.0`.

In Dependencies.AspNetCore.props:

`Microsoft.Extensions.Caching.StackExchangeRedis` and `Microsoft.Extensions.Http.Polly` has another version than the other ones.

In PortableObjectStringLocalizer.cs, remove:

```
        [Obsolete("This method will be removed in the upcoming ASP.NET Core major release.")]
        public IStringLocalizer WithCulture(CultureInfo culture) => this;
```

In NullHtmlLocalizerFactory.cs, remove:

`            IHtmlLocalizer IHtmlLocalizer.WithCulture(CultureInfo culture) => Instance;`

PR #7532, opened 2020-11-06
Fix NullHtmlLocalizer for Core5

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/7530

This just makes the interface for `WithCulture` which is obsolete on the `IHtmlLocalizer` interface explicit rather than implicit.

This means you can use a ASP.NET Core 3 nuget pack of OrchardCore in a .Net 5 application safely.

Packed locally and tested, works fine.

```
<Project Sdk="Microsoft.NET.Sdk.Web">

  <PropertyGroup>
    <TargetFramework>net5.0</TargetFramework>
    <TieredCompilation>true</TieredCompilation>
    <AspNetCoreHostingModel>InProcess</AspNetCoreHostingModel>
    <PreserveCompilationReferences>true</PreserveCompilationReferences>
  </PropertyGroup>

  <ItemGroup>
    <Folder Include="wwwroot\" />
    <Folder Include="Localization\" />
  </ItemGroup>

  <ItemGroup>
    <PackageReference Include="OrchardCore.Logging.NLog" Version="1.0.0-rc2-mypack" />
    <PackageReference Include="OrchardCore.Application.Cms.Targets" Version="1.0.0-rc2-mypack" />
  </ItemGroup>

</Project>
```

@sebastienros I know we talked about using some compiler directives, but at this point they're not necessary, unless we at some point do a multi target version.

I'm not sure we should - or if we do, we might want to keep the lang version to C# 8 so we don't have to many compiler directives to worry about.

We might put the compiler directives around this to prevent compiler warnings (I think this will actually still compile fine, but there might be a build analyser warning) for people using the source code, and moving the entire OrchardCore solution to .Net 5, but for a separate pr, if we decide we should support that

### DIFF
--- a/src/OrchardCore/OrchardCore/Localization/NullHtmlLocalizerFactory.cs
+++ b/src/OrchardCore/OrchardCore/Localization/NullHtmlLocalizerFactory.cs
@@ -56,7 +56,8 @@ namespace OrchardCore.Localization
             public LocalizedString GetString(string name, params object[] arguments) =>
                 NullStringLocalizerFactory.NullLocalizer.Instance.GetString(name, arguments);
 
-            IHtmlLocalizer IHtmlLocalizer.WithCulture(CultureInfo culture) => Instance;
+            [Obsolete("This method will be removed in the upcoming ASP.NET Core major release.")]
+            public IHtmlLocalizer WithCulture(CultureInfo culture) => Instance;
         }
     }
 }


src/OrchardCore/OrchardCore/Localization/NullHtmlLocalizerFactory.cs
```diff
@@ -56,7 +56,8 @@ public IEnumerable<LocalizedString> GetAllStrings(bool includeParentCultures)
             public LocalizedString GetString(string name, params object[] arguments) =>
                 NullStringLocalizerFactory.NullLocalizer.Instance.GetString(name, arguments);
 
-            IHtmlLocalizer IHtmlLocalizer.WithCulture(CultureInfo culture) => Instance;
+            [Obsolete("This method will be removed in the upcoming ASP.NET Core major release.")]
```
hishamco, line 59: I'm not sure why we need while the Localization APIs already warn about this. Second thing I'm not sure how this fix related to the referred issue, @agriffard is already fixed most if not issues in the PR #6999

deanmarcussen: This pr is not about moving to .Net Core 5 @hishamco - If you read the comments on the pr, it is about enabling usage of OrchardCore nugets in a solution targeting .Net Core 5.

@agriffard pr is about actually moving the target framework to .Net Core 5, which is a different thing.

I put the `Obsolete` in, although not required, so that when we move to .Net Core 5, we will catch it to remove it.

hishamco: > This pr is not about moving to .Net Core 5

Ya, but I'm not sure what's the issue if we didn't add this attribute?

Skrypt: It's going to warn those using it that they should stop using this method because it could be eventually be removed.

hishamco: I'm expected ASP.NET Core localization APIs will going to warn insted

sebastienros:
I don't think the C# version matters at all. And we can't use C# 9 if we don't use the dotnet 5 sdk which is not public yet.

deanmarcussen:
> I don't think the C# version matters at all. And we can't use C# 9 if we don't use the dotnet 5 sdk which is not public yet.

A comment about the possible issues around multi targeting only.

This pr is purely about getting the ability to run the current Orchard Core nugets, inside a webapp targeting the dotnet 5 pre release framework

sebastienros:
C# version is independent from dotnet version, when you multi target it's with dotnet versions. Once an assembly is built, the language version doesn't matter anymore.

sebastienros:
Note that we can remove `<TieredCompilation>true</TieredCompilation>` since it's a default for many versions now

PR #7537, opened 2020-11-06
Introduce Culture Scope

hishamco:
It's related to the PR #7528 and my comments there

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
@@ -21,6 +21,7 @@ using OrchardCore.ContentManagement.Routing;
 using OrchardCore.DisplayManagement.Liquid;
 using OrchardCore.Environment.Cache;
 using OrchardCore.Liquid;
+using OrchardCore.Localization;
 using OrchardCore.Settings;
 using YesSql;
 
@@ -391,19 +392,16 @@ namespace OrchardCore.Autoroute.Handlers
                     ContentItem = part.ContentItem
                 };
 
-                // We keep the current culture as a reference for later
-                var currentCulture = CultureInfo.CurrentUICulture;
+                using (var cultureScope = new CultureScope(CultureInfo.CurrentUICulture))
+                {
+                    _contentManager ??= _serviceProvider.GetRequiredService<IContentManager>();
+                    var cultureAspect = await _contentManager.PopulateAspectAsync(part.ContentItem, new CultureAspect());
 
-                _contentManager ??= _serviceProvider.GetRequiredService<IContentManager>();
-                var cultureAspect = await _contentManager.PopulateAspectAsync(part.ContentItem, new CultureAspect());
+                    CultureInfo.CurrentUICulture = cultureAspect.Culture;
 
-                CultureInfo.CurrentUICulture = cultureAspect.Culture;
-
-                part.Path = await _liquidTemplateManager.RenderAsync(pattern, NullEncoder.Default, model,
-                    scope => scope.SetValue("ContentItem", model.ContentItem));
-
-                // We reassign the proper culture to the current execution flow
-                CultureInfo.CurrentUICulture = currentCulture;
+                    part.Path = await _liquidTemplateManager.RenderAsync(pattern, NullEncoder.Default, model,
+                        scope => scope.SetValue("ContentItem", model.ContentItem));
+                }
 
                 part.Path = part.Path.Replace("\r", String.Empty).Replace("\n", String.Empty);
 
--- a/src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
@@ -392,13 +392,11 @@ namespace OrchardCore.Autoroute.Handlers
                     ContentItem = part.ContentItem
                 };
 
-                using (var cultureScope = new CultureScope(CultureInfo.CurrentUICulture))
+                _contentManager ??= _serviceProvider.GetRequiredService<IContentManager>();
+
+                var cultureAspect = await _contentManager.PopulateAspectAsync(part.ContentItem, new CultureAspect());
+                using (var cultureScope = new CultureScope(cultureAspect.Culture))
                 {
-                    _contentManager ??= _serviceProvider.GetRequiredService<IContentManager>();
-                    var cultureAspect = await _contentManager.PopulateAspectAsync(part.ContentItem, new CultureAspect());
-
-                    CultureInfo.CurrentUICulture = cultureAspect.Culture;
-
                     part.Path = await _liquidTemplateManager.RenderAsync(pattern, NullEncoder.Default, model,
                         scope => scope.SetValue("ContentItem", model.ContentItem));
                 }
--- a/src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
@@ -395,7 +395,7 @@ namespace OrchardCore.Autoroute.Handlers
                 _contentManager ??= _serviceProvider.GetRequiredService<IContentManager>();
 
                 var cultureAspect = await _contentManager.PopulateAspectAsync(part.ContentItem, new CultureAspect());
-                using (var cultureScope = new CultureScope(cultureAspect.Culture))
+                using (CultureScope.Create(cultureAspect.Culture))
                 {
                     part.Path = await _liquidTemplateManager.RenderAsync(pattern, NullEncoder.Default, model,
                         scope => scope.SetValue("ContentItem", model.ContentItem));
--- a/src/OrchardCore/OrchardCore.Abstractions/Localization/CultureScope.cs
+++ b/src/OrchardCore/OrchardCore.Abstractions/Localization/CultureScope.cs
@@ -35,6 +35,7 @@ namespace OrchardCore.Localization
         {
             SetCultures(_originalCulture, _originalUICulture);
         }
+
         private static void SetCultures(CultureInfo culture, CultureInfo uiCulture)
         {
             CultureInfo.CurrentCulture = culture;
--- a/src/OrchardCore/OrchardCore.Abstractions/Localization/CultureScope.cs
+++ b/src/OrchardCore/OrchardCore.Abstractions/Localization/CultureScope.cs
@@ -5,17 +5,17 @@ namespace OrchardCore.Localization
 {
     public sealed class CultureScope : IDisposable
     {
-        private readonly CultureInfo _currentCulture;
-
         public CultureScope(CultureInfo culture)
         {
-            _currentCulture = culture;
+            Culture = culture;
             CultureInfo.CurrentUICulture = culture;
         }
 
+        public CultureInfo Culture { get; private set; }
+
         public void Dispose()
         {
-            CultureInfo.CurrentUICulture = _currentCulture;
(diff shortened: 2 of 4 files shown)


src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
```diff
@@ -391,19 +392,16 @@ private async Task GenerateContainerPathFromPattern(AutoroutePart part)
...
                 };
 
-                // We keep the current culture as a reference for later
-                var currentCulture = CultureInfo.CurrentUICulture;
-
-                _contentManager ??= _serviceProvider.GetRequiredService<IContentManager>();
-                var cultureAspect = await _contentManager.PopulateAspectAsync(part.ContentItem, new CultureAspect());
-
-                CultureInfo.CurrentUICulture = cultureAspect.Culture;
+                using (var cultureScope = new CultureScope(CultureInfo.CurrentUICulture))
```
sebastienros, line 395: This is kind of ironic to set the CurrentUICulture to CurrentUICulture ;)

So I will just assume that here the goal is to be able to restore the previous value correctly. Kind of over-engineering to create a scope object when a try/finally will just work here.

But otherwise I have nothing against it, let's assume that it could be used somewhere else.

src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
```diff
@@ -391,19 +392,16 @@ private async Task GenerateContainerPathFromPattern(AutoroutePart part)
...
+                {
+                    _contentManager ??= _serviceProvider.GetRequiredService<IContentManager>();
+                    var cultureAspect = await _contentManager.PopulateAspectAsync(part.ContentItem, new CultureAspect());
 
-                part.Path = await _liquidTemplateManager.RenderAsync(pattern, NullEncoder.Default, model,
-                    scope => scope.SetValue("ContentItem", model.ContentItem));
+                    CultureInfo.CurrentUICulture = cultureAspect.Culture;
```
sebastienros, line 400: What's the point of creating a helper to do that, and now you set it manually. Why not call the aspect before and use the scope with the result of the aspect?

src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
jtkech: Yes, i saw both, try/finally or using block

Hmm, maybe useful to have this helper class for each future case you want to render a template but not in a request context, e.g. in the background to send an email

Yes, the usage would be

    using (var cultureScope = new CultureScope(cultureAspect))

Hmm, maybe name it `CultureSwitcher` or if only for UI `UICultureSwitcher`

src/OrchardCore/OrchardCore.Abstractions/Localization/CultureScope.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Globalization;
+
+namespace OrchardCore.Localization
+{
+    public sealed class CultureScope : IDisposable
+    {
+        public CultureScope(CultureInfo culture)
+        {
+            Culture = culture;
+            CultureInfo.CurrentUICulture = culture;
```
sebastienros, line 11: What is the rule to either update CurrentCulture, CurrentUICulture, or both?
When would it make sense to not update the two with the same value?

jtkech: Just for info, maybe useful ;)

https://github.com/joakimriedel/culture-switcher/blob/master/src/CultureSwitcher.cs

Skrypt: CurrentUICulture worked fine but why not having both. Felt like this was related to UI only on my PR. Now, the helper is kind of over-engineering like seb says but I'm fine with it if done right.

src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
hishamco: Wait a minute, seems I did something wrong here ;) No need to set the CurrentUICulture in the constructor, seems I forgot to delete from the constructor

src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
hishamco: You are right that's why you saw CurrentUICulture to CurrentUICulture ;)

src/OrchardCore/OrchardCore.Abstractions/Localization/CultureScope.cs
```diff
@@ -0,0 +1,23 @@
+using System;
+using System.Globalization;
+
+namespace OrchardCore.Localization
+{
+    public sealed class CultureScope : IDisposable
+    {
+        private CultureScope(CultureInfo culture)
```
jtkech, line 8: Maybe `UICultureScope` as we only set the UI culture, or as suggested by Seb we could set both cultures

src/OrchardCore/OrchardCore.Abstractions/Localization/CultureScope.cs
```diff
@@ -0,0 +1,23 @@
...
+        {
+            Culture = culture;
```
jtkech, line 10: This is wrong, should be `Culture = CultureInfo.CurrentUICulture` otherwise you will not reset correctly the ui culture in the dispose method

src/OrchardCore/OrchardCore.Abstractions/Localization/CultureScope.cs
```diff
@@ -0,0 +1,23 @@
...
+            CultureInfo.CurrentUICulture = culture;
+        }
+
+        public CultureInfo Culture { get; private set; }
```
jtkech, line 14: I would not use a property, just a private field

hishamco: Just a second you will see where could be used ;)

src/OrchardCore/OrchardCore.Abstractions/Localization/CultureScope.cs
```diff
@@ -0,0 +1,44 @@
...
+            => Create(CultureInfo.GetCultureInfo(culture), CultureInfo.GetCultureInfo(uiCulture));
+
+        public static CultureScope Create(CultureInfo culture) => Create(culture, culture);
+
+        public static CultureScope Create(CultureInfo culture, CultureInfo uiCulture) => new CultureScope(culture, uiCulture);
+
+        public void Dispose()
+        {
+            SetCultures(_originalCulture, _originalUICulture);
+        }
```
Skrypt, line 37: new line

hishamco: Nice catch

src/OrchardCore/OrchardCore.Abstractions/Localization/CultureScope.cs
```diff
@@ -0,0 +1,45 @@
...
+            UICulture = uiCulture;
+            _originalCulture = CultureInfo.CurrentCulture;
+            _originalUICulture = CultureInfo.CurrentUICulture;
+
+            SetCultures(culture, uiCulture);
+        }
+
+        public CultureInfo Culture { get; }
+
+        public CultureInfo UICulture { get; }
```
jtkech, line 23: @hishamco i will approve your PR

But just one thing, i don't see any usage of these 2 public properties

test/OrchardCore.Tests/Localization/PortableObjectStringLocalizerTests.cs
```diff
@@ -199,30 +202,32 @@ public void LocalizerReturnsOriginalTextForPluralIfTranslationDoesntExist(string
...
         [InlineData("zh-Hans", "球", 1, new string[] { "球" })]
         [InlineData("zh-Hans", "球", 2, new string[] { "球" })]
         public void LocalizerReturnsCorrectTranslationForPluralIfNoPluralFormsSpecified(string culture, string expected, int count, string[] translations)
         {
-            var currentCulture = CultureInfo.GetCultureInfo(culture);
-            CultureInfo.CurrentUICulture = currentCulture;
+            using (var cultureScope = CultureScope.Create(culture))
+            {
+                // using DefaultPluralRuleProvider to test it returns correct rule
+                TryGetRuleFromDefaultPluralRuleProvider(cultureScope.Culture, out var rule);
```
hishamco, line 222: > But just one thing, i don't see any usage of these 2 public properties

It's used here

src/OrchardCore/OrchardCore.Abstractions/Localization/CultureScope.cs
hishamco: I point to where it's used, also we might to expose the original culture because may it's useful in some cases

Skrypt (approved):

test/OrchardCore.Tests/Localization/PortableObjectStringLocalizerTests.cs
jtkech: First, based on the removed line, here maybe `cultureScope.UICulture`

Okay, so you also use the culture scope as a CultureInfo helper to get a cultureInfo from a string

Then because the culture scope set the async local cultures you can just use `CultureInfo.CurrentUICulture`, **that's the goal of this scope** to set cultures that can be retrieved from everywhere called in the context of this scope

            using (CultureScope.Create(CultureInfo.GetCultureInfo(culture)))
            {
                // using DefaultPluralRuleProvider to test it returns correct rule
                TryGetRuleFromDefaultPluralRuleProvider(CultureInfo.CurrentUICulture, out var rule);

Or if you want to keep the helper part creating a cultureInfo from a string

            using (CultureScope.Create(culture))
            {
                // using DefaultPluralRuleProvider to test it returns correct rule
                TryGetRuleFromDefaultPluralRuleProvider(CultureInfo.CurrentUICulture, out var rule);

hishamco: > First, based on the removed line, here maybe cultureScope.UICulture

Ok, but it's the same in this case

> Okay, so you also use the culture scope as a CultureInfo helper to get a cultureInfo from a string

There's already an overload to do that

jtkech: Okay will see tomorrow if it is not yet merged ;)

The main thing is that you are using a scope to set the static async local CultureInfo properties, and then for your tests in place of retrieving these values from this same async local CultureInfo you are retrieving them from the scope you just created, so here you just test the properties of this scope, you are not testing what it is intended to do.

E.g. when using this scope for liquid rendering (when not in a real view context), the template rendering will still retrieve the cutures from the static CultureInfo, that's the goal

hishamco: I may add some unit tests

test/OrchardCore.Tests/Localization/CultureScopeTests.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            }
+
+            // Assert
+            Assert.Equal(culture, CultureInfo.CurrentCulture);
+            Assert.Equal(uiCulture, CultureInfo.CurrentUICulture);
+        }
+
+        [Fact]
+        public void CultureScopeRetreivesTheOrginalCulturesIfExceptionOccur()
```
hishamco: > E.g. when using this scope for liquid rendering (when not in a real view context)

@jtkech no need to test a fake `ILiquidTemplateManager` this unit tests ensures if the `Render` throws an exception it will restore the original cultures

PR #7539, opened 2020-11-06
Autoroute pattern localization try/finally

Skrypt:
Adding try/finally


src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
```diff
@@ -391,19 +391,23 @@ private async Task GenerateContainerPathFromPattern(AutoroutePart part)
...
-                CultureInfo.CurrentUICulture = cultureAspect.Culture;
-
-                part.Path = await _liquidTemplateManager.RenderAsync(pattern, NullEncoder.Default, model,
-                    scope => scope.SetValue("ContentItem", model.ContentItem));
+                // We keep the current culture as a reference for later
+                var currentCulture = CultureInfo.CurrentUICulture;
 
-                // We reassign the proper culture to the current execution flow
-                CultureInfo.CurrentUICulture = currentCulture;
+                CultureInfo.CurrentUICulture = cultureAspect.Culture;
```
Skrypt, line 400: I wanted to make sure the whole process won't throw. Though here the finally is only necessary if you actually change the CultureINfo.CurrentUICulture so that makes sense.

Skrypt: We should probably also make sure that cultureAspect.Culture is never null.

jtkech: Yes you're right if it could be null

@hishamco will have to take it into account in its PR if it updates this part

Skrypt: The PR that changes the `await orchardHelper.GetContentCultureAsync` might do that for when we want to index with Lucene. Though we don't use `await orchardHelper.GetContentCultureAsync` here either and we should probably too.

Skrypt: I think we are due to fix those issues all at once else we do small bits here and there and it is not consistent.

Skrypt: Right now cultureAspect.Culture can't be null because of the default of the CultureAspect.

src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
```diff
@@ -391,19 +391,23 @@ private async Task GenerateContainerPathFromPattern(AutoroutePart part)
...
+                try
```
hishamco, line 401: Can I add the try block im PR or you could add it after merge my PR

jtkech: If we use your PR with an using block we would not have to use a try / catch

hishamco: That's why I think before, but what will happen if an exception occurs during the rendering process?

BTW my PR is ready for review

jtkech: Your dispose() will be still called, an using block uses try / finally blocks internally

PR #7544, opened 2020-11-07
Remove CultureScope public culture properties

jtkech:
The culture scope is not intended to provide cultures but to switch the current cultures with provided values

Not so important because through these properties we retrieve the same values

**But i think it may be misleading**. E.g. in the unit tests we are checking / using these public properties in place of checking / using the current cultures that the culture scope was intended to switch.

E.g. if in the culture scope you remove the lines that update the current cultures, most of its unit tests still work.

But as said, not so important, just to show to @hishamco what i meant ;)

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.Globalization;
 using System.Linq;
 using System.Threading.Tasks;
 using Fluid;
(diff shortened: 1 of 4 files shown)


hishamco:
I never mind to remove these properties but sometimes it's useful to retrieve the current or old culture inside the scope

jtkech:
Okay cool, and accessing a regular property has a lower cost than accessing again a current culture async state, **so i will close this PR**, sorry @hishamco ;) it was just to show that **in some unit tests** we are using e.g. the `UICulture` property that we pass when creating the scope, **not the `CurrentUICulture` that the scope was intended to switch**.

hishamco:
> it was just to show that in some unit tests we are using e.g. the UICulture property that we pass when creating the scope, not the CurrentUICulture that the scope was intended to switch.

CurrenUICulture will be set to what we passed in in the scope, so they are equivalent, Am I right?

jtkech:
Yes you're right, better to check that it has set its own property than checking that the current culture was switched

Issue #7526, opened 2020-11-05
Is placement possible in the User editor?


craftyweb:
Hi,

I made a CustomerModel to attach to the User entity. I was wondering if it is possible to place the editor of this model in separated tabs like you can do with fields and parts in placement.

Now it’s just showing below the default fields of the User.

![AC3F8EB6-1514-4415-AB6C-B6EED8FEC52D](https://user-images.githubusercontent.com/61787638/98290439-f7aa4f80-1fa9-11eb-94c8-6da33e7708b0.jpeg)

Thanks!

deanmarcussen:
I don't see why not. Have you tried it?

craftyweb:
> I don't see why not. Have you tried it?

@deanmarcussen 
Yes I've tried it by placing the name of the shape defined in the driver:

Customer_Edit

And place this in placement.json like 

"Customer_Edit": [
    {
      "place": "#Customer:0"
    }
  ]

I think there is something wrong with the way I want it to place in placement.json so I was wondering if you guys could help me with the right solution (if it is possible).

deanmarcussen:
For placement to work there you need a `Match`. Just specifying `place` is not enough.

But we don't have any differentiators by default on `User` shapes, or anything to `Match` on.

Plus, I was working on something yesterday, and realised tabs don't quite work with users, because they don't have a `ContentItemId`

However, this pr fixes that https://github.com/OrchardCMS/OrchardCore/pull/7556
and if you look at the `CustomUserSettingsDisplayDriver` you'll see how I've specified the tab via the `Location` property.

```
 .Location($"Content:10#{contentTypeDefinition.DisplayName}")`
```

So when that is merged, tabs will work, and I would suggest using `.Location` as you have written a driver.

craftyweb:
@deanmarcussen Thanks! this works very well. Closing this now.

Issue #1025, opened 2017-09-26
Add IEntity support to User


sebastienros:
And make the current implementation based on a driver.

paynecrl97:
I'm going to have a go at doing this.

Am I good to base this on the way `ISite` works in the `OC.Settings` module, and then create 2 drivers- one for user setting & one for roles?

sebastienros:
Yes, I actually explained this to @sfmskywalker yesterday. You resolve a `IDisplayManager<IUser>` and it will let you use drivers for `IUser`. Then you can use `SectionDisplayDriver` to handle parts-like drivers.

So yes, if you understand how `ISite` works, it should be super easy to do it for Users. You could optionally implement `Display`, with shapes targeting the front end and others for the admin. And use it in the current view that displays the list of users (SummaryAdmin display type).

paynecrl97:
Oh yeah, didn't think about using SummaryAdmin to display the list. I'll add that in too.

hishamco:
@paynecrl97 are you still interested to do this?

deanmarcussen:
Fixed by https://github.com/OrchardCMS/OrchardCore/pull/1054

PR #1054, opened 2017-10-02
Makes IUser extensible

paynecrl97:
#1025 

This PR adds `IEntity` support to `IUser` and moves user create/edit logic from the `AdminController` and into a driver. This allows the `IUser` object to be extended.

I'll raise a further PR at some point shortly to change the admin list view of users to use drivers as well.

PR #7556, opened 2020-11-09
Custom User Settings

deanmarcussen:
I don't think we have an issue about this, but we've talked about it a few times.

Custom User Settings, much like Custom Site Settings.

Implement a content type with a stereotype `CustomUserSettings` and it'll show up on a tab in the user properties.

Screenshot 2020-11-09 at 14 21 54

Screenshot 2020-11-09 at 14 23 10

Screenshot 2020-11-09 at 14 23 16

I've just done this as a feature of `Users` as it's a small number of files required.


src/OrchardCore.Modules/OrchardCore.Contents/Views/AdminTabs.cshtml
```diff
@@ -9,7 +9,7 @@
     <ul class="nav nav-tabs flex-column flex-md-row" role="tablist">
         @foreach (var tab in tabs)
         {
-            var tabId = $"tab-{tab}-{Model.ContentItem.ContentItemId}".HtmlClassify();
+            var tabId = $"tab-{tab}-{Model.ContentItem?.ContentItemId}".HtmlClassify();
```
deanmarcussen, line 12: An id here is not required, it is just useful if you want to target these with javascript.

And for user shapes, there isn't an easy id, but we don't need it.

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/7526

jptissot:
I would move this in a seperate assembly as I think the User's module is not a CMS module ? It's classified under the OrchardCore.Modules in the VS solution. I don't know enough about the dependency tree, just going with the VS solution's structure.

agriffard:
Is it possible to set those fields in a Register and an Edit Profile page?

agriffard:
Some people could also want to use them on the front instead of the username.
You explained in the doc in Liquid. 
Can you please provide a Razor sample?

Could it be possible to redefine an admin template to override the displayed user name in admin? 
Ex: Display Avatar and First name if set.

sebastienros:
How do permissions work? Can a user access its own profile? Or just part of it?

What about having a feature that would have a custom controller/view to render a user profile on the front-end, with the goal that devs could customize this page and not worry about security, or loading the user. Obviously I'd prefer it to be totally decoupled but that could be a separate guide ;)

deanmarcussen:
Re permissions.

- This issue should probably be closed https://github.com/OrchardCMS/OrchardCore/issues/1025
- The pr that added `IEntity` support for Users, has some good comments about permissions that we could add, that never quite got done I suspect https://github.com/OrchardCMS/OrchardCore/pull/1054

As there isn't permissions other than `ManageUsers` as yet, this feature doesn't have any permission handlers.

As far as guides go... there's room for plenty on users - how to add a custom claim factory (for providing a profile picture via a claim), and then use it in the admin would be a good one @agriffard There's demo code for some of it in the `Demo` project.

And a user profile feature / controller would also work.

Also not far behind is a `UserPickerField`.

davidhayden:
This looks really cool, and I can't wait to try it out!

I couldn't tell from the conversation, but is it possible to make this available on the frontend where a user can manage their own profile?

The UserProfile example in the demo code isn't ideal, because it doesn't work with content editors and requires coding each time a change is made to the profile. This solution, if available on the frontend, would be similar to the Profile Part in O1 that allowed one to simply add fields to a content part that each user was then able to securely access and update on the frontend.

sebastienros:
@davidhayden Right now this PR doesn't have any front end capability. I was suggesting we should add a controller to "display" and only display the profile, such that you can create a custom template file to change it. But no edition since the settings are edited using standard fields which are meant for the admin, and handling front-end edition is too complex in terms of UI. 

My suggestion would be to provide your own form and controller/razor page on the front-end if you need it. At least the current PR provides a way to define the forms on the admin so you can store, see and update custom information for any user.

Or optionally use the forms module and a workflow to let users edit very specific parts of their profile. Custom widgets can also read the profile data of the current logged users. These are just ideas of the possibilities...

hishamco:
From design perspective I'd like to have 3 tabs as following:

1. Content

Which is shown as screenshot above

2. Profile

Which contains: FirstName, LastName, Photo, Email .. etc

3. Others

Which contains other custom user properties something like Mobile, Address, Social media links ... etc. Such properties could be extensible and created programmatically

So far what has been done is great but it needs a little organization as I mentioned above. Another thing which may I create a PR for it, is to create a menu for user profile in the admin theme before logout menu, that allow user to edit his profile

deanmarcussen:
It's great how you can create them all through the admin ui and get complete freedom on how they are designed @hishamco 

Next time I'm preparing a couple of demo screenshots I'll make sure to highlight that they're only an example of what can be done ;)

hishamco:
Great!! waiting for your demos as usual ;)

agriffard:
Can you please add a page to edit your user Profile from the admin, as discussed during the meeting?

jptissot:
> Can you please add a page to edit your user Profile from the admin, as discussed during the meeting?

I think the suggestion was around having a link in the menu to edit your own user.

hishamco:
@agriffard I think what you asking is similar to what I suggested before, I think it should be in another PR I can create a menu item that link to the current user edit page, no need to create another page

jptissot:
> @agriffard I think what you asking is similar to what I suggested before, I think it should be in another PR I can create a menu item that link to the current user edit page, no need to create another page

I think we also need a seperate page. In case the user has access to the admin but does not have permissions to edit users (and roles)

hishamco:
@jptissot I don't think we need a new page for the profile, while the user able to edit his info

> In case the user has access to the admin but does not have permissions to edit users (and roles)

In this we can disable non user profile settings, so the user will not be able to edit his roles

deanmarcussen:
Yes, I will handle the authorization / permissions / display aspects discussed in the meeting as part of this pr.

- separate drivers and views for the current user driver into user editable sections / admin only sections
- changes to user permissions for `EditOwnUser` 
- user information edit controller action for managing the driver
- dynamic permissions for the dynamic sections added via content types

@davidhayden just to clarify what @sebastienros discusses.

- This pr is focusing on the admin, but we will look towards front end display, probably not front end editing (though maybe providing an example or guide), in another pr.
- The custom user settings, while creatable through the admin, are pretty similar to the examples provided in the `OrchardCore.Demo` module, in that for a custom user claim, you'll still need to implement a custom claims factory, to map the user settings data, to a claim.
- So it's mostly about making it easier to create these settings through the admin interface, rather than code, for the moment

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
```diff
@@ -121,46 +128,8 @@ public override async Task<IDisplayResult> UpdateAsync(User user, UpdateEditorCo
...
-                var userWithSameNameId = await _userManager.GetUserIdAsync(userWithSameName);
-                if (userWithSameNameId != model.Id)
-                {
-                    context.Updater.ModelState.AddModelError(string.Empty, S["The user name is already used."]);
-                }
-            }
-
-            if (model.UserName != user.UserName && user.UserName == httpContext.User.Identity.Name)
-            {
-                context.Updater.ModelState.AddModelError(string.Empty, S["Cannot modify user name of the currently logged in user."]);
```
deanmarcussen, line 146: @Skrypt I need to understand why you added this, as I have removed it, and forced a new cookie when editing the current user... so allowing the user to change his name.

hishamco: AFAIK this has been added when the user edit his info (username & role), so it restricts the editing to specific information

deanmarcussen:
From the original pr https://github.com/OrchardCMS/OrchardCore/pull/1054

> Just a little request, could we add new permissions named ManageUserInformation and ManageOwnUserInformation (which is implied by the first one)? When editing a user we then check for ManageUserInformation. The ManageOwnInformation is provided by a permission handler (like for content item Ownership permissions).

So Permissions added, plus dynamic ones for the custom user settings.

Here's some admin profile views

`ManageAllUsers`
Screenshot 2020-11-13 at 16 09 47

`ManageOwnUserInformation` + Manage just that custom setting.
Screenshot 2020-11-13 at 16 10 08

Screenshot 2020-11-13 at 16 10 17

davidhayden:
Thanks for the clarification @sebastienros  and @deanmarcussen 

This looks **awesome** and I look forward to the frontend integration as well.

PR #7563, opened 2020-11-10
Fixing unique user id verification

### LINKED ISSUE #7560: Can't create new users anymore, stuck in loop
Just updated to the latest bits on dev, can't create new users anymore.

Code is stuck in this loop:
https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs#L87

Not sure why. A new UserId is generated correctly but the query still finds it.

sebastienros:
Fixes #7560


sebastienros:
I tested it by creating a new user, and stepped into the code to be sure it was going through this change.
Ideally, I would have added a functional test.

PR #7570, opened 2020-11-10
Introduce User Picker

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6204

Something else we have discussed regarding users. 

The idea being to offer different ways of connecting content items to users.

This, for example, allows multiple users to be selected as co-authors of a Content Item.

This could also be done automatically through a workflow.

We will need to decide if this introduces any security implications, by making the `User` more or less accessible.

Noting that the `PasswordHash` is stored on the `User` model.

The concepts of a display profile that this could potentially link to are all up for discussion and ideas.

The Autoroute templating feature we have already talked about could solve some of these ideas as well.

i.e. imagine a content item with an `Autoroute.Path` of `/profile/{username}`

It might involve some display permissions for users / parts of a user entity as well as edit permissions.

Screenshot 2020-11-10 at 13 43 50

Screenshot 2020-11-10 at 13 44 18


src/OrchardCore.Modules/OrchardCore.ContentFields/Fields/UserPickerField.cs
```diff
@@ -0,0 +1,10 @@
+using System;
+using OrchardCore.ContentManagement;
+
+namespace OrchardCore.ContentFields.Fields
+{
+    public class UserPickerField : ContentField
+    {
+        public string[] UserIds { get; set; } = Array.Empty<string>();
```
deanmarcussen, line 8: Should we store the user name here as well, possibly as a less well known property like tag names?

Or as a concrete value?

This would be necesary to get the value into the lucene index

src/OrchardCore.Modules/OrchardCore.ContentFields/Indexing/UserPickerFieldIndexHandler.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            var options = DocumentIndexOptions.Store;
+
+            if (field.UserIds.Length > 0)
+            {
+                foreach (var userId in field.UserIds)
+                {
+                    foreach (var key in context.Keys)
+                    {
+                        context.DocumentIndex.Set(key, userId, options);
```
deanmarcussen, line 19: Needs the username property on the field to be able to set it in the lucene index here.

src/OrchardCore.Modules/OrchardCore.ContentFields/Views/UserPickerField.cshtml
```diff
@@ -0,0 +1,32 @@
...
+@using OrchardCore.Users.Indexes
+@using YesSql.Services;
+@using YesSql;
+
+@inject YesSql.ISession Session
+
+@{
+    var name = (Model.PartFieldDefinition.PartDefinition.Name + "-" + Model.PartFieldDefinition.Name).HtmlClassify();
+    var users = await Session.Query<User, UserIndex>().Where(x => x.UserId.IsIn(Model.UserIds)).ListAsync();
+}
```
deanmarcussen, line 13: Should we expose this as a `Session.Query` ?

Or extend the `IUserService` to have a `GetMany` method.

Or be more like the media field, and just provide the user id in the view?

For liquid that will mean writing a SQL / Lucene query, or should we expose part of the `IUserService`

src/OrchardCore.Modules/OrchardCore.Users/Startup.cs
```diff
@@ -187,8 +187,6 @@ public override void ConfigureServices(IServiceCollection services)
 
             services.AddScoped<IDisplayDriver<ISite>, LoginSettingsDisplayDriver>();
 
-            services.AddScoped<ILiquidTemplateEventHandler, UserLiquidTemplateEventHandler>();
```
deanmarcussen, line 190: this was just being registered twice, once in the wrong startup.

src/OrchardCore.Modules/OrchardCore.ContentFields/Fields/UserPickerField.cs
sebastienros: Ok to store usernames too.

src/OrchardCore.Modules/OrchardCore.ContentFields/Indexing/UserPickerFieldIndexHandler.cs
sebastienros: sure

src/OrchardCore.Modules/OrchardCore.ContentFields/Views/UserPickerField.cshtml
sebastienros: helper+ filter to get a user from userid

deanmarcussen:
@jptissot lets catch up later, and see if I covered all the bases.

Will leave the how to get a generic user claim (for say profile pic) from the `ClaimsPrincipal` for another pr.

Docs to come Ag ;)

src/OrchardCore.Modules/OrchardCore.Users/Startup.cs
```diff
@@ -201,6 +201,8 @@ public override void ConfigureServices(IServiceCollection services)
             services.AddLiquidFilter<HasClaimFilter>("has_claim");
             services.AddLiquidFilter<IsInRoleFilter>("is_in_role");
             services.AddLiquidFilter<UserEmailFilter>("user_email");
+            services.AddLiquidFilter<UserIdClaimFilter>("user_id_claim");
+            services.AddLiquidFilter<UserFilter>("user_id");
```
jptissot, line 205: I am not a fan of this name. I know it's similar to the `content_item_id` filter but this is confusing. I would rename to load_user or something else ? 

If I understand properly, we will need to do this to load the current user : 
`{% assign user =  User | user_id_claim | user_id %}`

Thinking about it I would change it up so that `user_id` loads the id claim and `get_user` to load user(s) ?

src/OrchardCore.Modules/OrchardCore.ContentFields/Assets/js/vue-multiselect-userpicker.js
```diff
@@ -0,0 +1,92 @@
+function debounceUserPicker(func, wait, immediate) {
```
jptissot, line 1: For another PR, but maybe we could make this guy even more generic so we would not need to copy it around.

src/OrchardCore.Modules/OrchardCore.Users/Startup.cs
deanmarcussen: > I am not a fan of this name. I know it's similar to the content_item_id filter but this is confusing

Me neither. I'll have a look at see what shopify are calling things like that.

src/OrchardCore.Modules/OrchardCore.ContentFields/Controllers/UserPickerAdminController.cs
```diff
@@ -0,0 +1,66 @@
...
+        public UserPickerAdminController(
+            IContentDefinitionManager contentDefinitionManager,
+            IEnumerable<IUserPickerResultProvider> resultProviders
+            )
+        {
+            _contentDefinitionManager = contentDefinitionManager;
+            _resultProviders = resultProviders;
+        }
+
+        public async Task<IActionResult> SearchUsers(string part, string field, string query)
```
jptissot, line 32: This is using the same pattern as the ContentPicker. I was wondering if there was a way to secure this controller with proper permissions ?

deanmarcussen: what would the proper permissions be?

We have `ManageUsers` but we have nothing for `DisplayUsers`

jptissot: I am not sure to be honest. Just pointing it out. Do you think this could be a security issue where someone who has the visit admin permission be able to get a list of  users ?

src/OrchardCore.Modules/OrchardCore.Users/Startup.cs
jptissot: How about `user_id` to get the user's id and `users` to get the user(s)

jptissot: @sebastienros namingmaster how should we name this ? ;)

src/OrchardCore.Modules/OrchardCore.ContentFields/Controllers/UserPickerAdminController.cs
deanmarcussen: We might need a `ViewUsers` permission. 

Because this is related to content editing, I'm not sure. It might be better here to restrict this access to Edit Content.

As you couldn't use it, unless you were editing a content item.

Skrypt: This is the kind of field where it should only allow to search by UserName.

deanmarcussen: I made it extensible @Skrypt so that someone could implement a search by email, or any of the other custom properties that they may have relating to users.

Like for example they might have a first name / last name to search, so they could search for Jasmin, rather than having to search for @Skrypt 

@jptissot I think I will tie the permission to `EditContent` for that content type, rather than a new permission for users or for this field.

jptissot: Makes sense! Thanks @deanmarcussen

jptissot:
I think this also fixes https://github.com/OrchardCMS/OrchardCore/issues/7669

src/OrchardCore.Modules/OrchardCore.Users/Startup.cs
deanmarcussen: > How about user_id to get the user's id and users_info to get the user(s) information

Renamed to the above

src/OrchardCore.Modules/OrchardCore.ContentFields/Controllers/UserPickerAdminController.cs
deanmarcussen: Permissions done

jptissot: LGTM

jptissot:
I pushed a couple of tweaks to the code and added some liquid docs. I left a TODO in the docs for the text of the field itself.

src/docs/reference/modules/Users/CustomUserSettings/README.md
```diff
@@ -17,10 +17,14 @@ Once created, open the Setting menu item and each of these sections should appea
...
-The Custom User Settings are available on the `{{ User.Properties }}` object.  
+The Custom User Settings are available when loading the user from the database. 
+```liquid 
+{% assign user = User | user_id | get_users %}
+{{user.Properties}}
+```
 Each section is made available using its name.
 
-For instance for a custom settings section named `UserProfile`, with a `TextField` named `FirstName` would be accessible using `{{ User.Properties.UserProfile.Content.UserProfile.FirstName.Text }}`.
+For instance for a custom settings section named `UserProfile`, with a `TextField` named `FirstName` would be accessible using `{{ user.Properties.UserProfile.Content.UserProfile.FirstName.Text }}`.
```
jptissot, line 27:
```suggestion
For instance for a custom settings section named `UserProfile`, with a `TextField` named `FirstName` would be accessible using `{{ user.Properties.UserProfile.UserProfile.FirstName.Text }}`.
```

jptissot (approved):

PR #7579, opened 2020-11-10
Admin Menu AuthorizeContentTypeDefinitionsAsync should use UserId

### LINKED ISSUE #7577: Admin Menu permissions not evaluated correctly (UserId regression)
Now that we have the UserId we should be evaluating it instead of the UserName.

Skrypt:
Fixes #7577


Skrypt:
image

Huhhhghh !

jtkech (approved):

jtkech:
@Skrypt 

In the `AdminController` of `OC.Contents`, there are 5 other places where an owner is set to `User.Identity.Name`

Do we need to also replace them with `User.FindFirstValue(ClaimTypes.NameIdentifier)`?

Skrypt:
Yes I think so.

PR #7581, opened 2020-11-11
Switch CI to github actions

jptissot:
Replace travis and Appveyor and run Functional tests on all databases.

## Todo
- ~SQL Server~
- ~Do not run if only docs modified~
- ~Docker images publish~


Skrypt:
@sebastienros I think @jptissot needs you to configure Github Action "secrets" for the Docker User/Password.

jptissot:
I am adding a bunch of stuff to this pr. Namely running the e2e tests on all databases. Also running the tests on all OS

jptissot:
I opted to make the SaaS tenant pre-configured to remove the dependency between tests.

.github/workflows/test.yml
```diff
@@ -0,0 +1,152 @@
...
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: ${{ matrix.dotnet }}
+    - name: Build and test
+      run: |
+        dotnet build --configuration Release
+        dotnet test -c Release --no-build ./test/OrchardCore.Tests/OrchardCore.Tests.csproj
+  test_functional_mvc:
+    env:
+      DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
```
jptissot, line 33: I think these can be moved a level and be used for all steps

Skrypt:
I think @deanmarcussen also had some Unit Tests targetting specific databases which we're requiring a global var to be set so that it gets triggered. Maybe later for that one.

jptissot:
> I think @deanmarcussen also had some Unit Tests targetting specific databases which we're requiring a global var to be set so that it gets triggered. Maybe later for that one.

I think those run on appveyor so we might be ok for now.

jtkech:
Always happy when we see that everything is green ;)

jptissot:
> I think @deanmarcussen also had some Unit Tests targetting specific databases which we're requiring a global var to be set so that it gets triggered. Maybe later for that one.

I guess I went a bit further and now we don't need appveyor for the dev branch. Maybe we need a bit more in this PR to completely remove AppVeyor.

@deanmarcussen I did not re-run the unit tests using database in the ci. Maybe we can fix it later ?

Can someone explain to me what the difference is between the windows docker image and the linux one ? I think they are identical except that one is built on windows and the other on linux. I don't think we actually use any "windows containers" here. So why have 2 different images ?

jptissot:
@sebastienros I think after this is merged you should enable a check on the branches settings of github.

image

test/Functional/cypress-commands/src/tenants.js
```diff
@@ -37,14 +37,27 @@ Cypress.Commands.add('newTenant', function(tenantInfo) {
...
+    const db = $body.find("#DatabaseProvider");
+    // if a database provider is already specified by an environment variable.. leave it as is
+    // this assumes that if you set the provider, you also set the connectionString
+    if (db.length > 0 && db.val() == "") {
+      db.val('Sqlite');
+    } else {
+        //set the tablePrefix to the name.
+        const prefix = $body.find("#TablePrefix");
+        if(prefix.length > 0){
+            prefix.val(name);
```
deanmarcussen, line 58: Here you are using `name` for the table prefix, rather than `prefix`- I know they're calculated the same, so call me Mr Picky ;)

test/OrchardCore.Tests.Modules/BaseThemeSample/BaseThemeSample.csproj
```diff
@@ -6,6 +6,7 @@
 
   <PropertyGroup>
     <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework>
+    <IsPackable>false</IsPackable>
```
deanmarcussen, line 9: Was this just for a safety net, or are you running pack on the whole project folder, rather than just the src folder?

deanmarcussen (commented):
Looks great.

Totally happy to do the integration tests at a different time, or perhaps even drop them completely. 

Might have to try writing a cypress test to see if I can get results I was looking for with them.

No SQL Server ? Another pr?

Can I ask one thing though, I'd really like to get a zip of the `App_Data` with any log failures produced as an artefact of the tests, as I don't think the cypress output of failures, is going to be easy to see what the problem is. But any setup errors / migration errors / database problems will show in the logs.

test/Functional/cypress-commands/src/tenants.js
jptissot: i needed to keep the prefix really small because I hit an issue with postgresql table size. Some of our indexes are long ;)

test/OrchardCore.Tests.Modules/BaseThemeSample/BaseThemeSample.csproj
jptissot: I was running pack on the whole project. like was done before ?

jptissot: Ah I just saw it. Everything is packed but only projects under src are kept as artifacts by appveyor: 
https://github.com/OrchardCMS/OrchardCore/pull/7581/files#diff-92ab9a36df5d8e9f7076f2fdec59492d1ac2d9cf27ea046767a7fc4d542ef3dcL90

jptissot:
>Can I ask one thing though, I'd really like to get a zip of the App_Data with any log failures produced as an artefact of the tests, as I don't think the cypress output of failures, is going to be easy to see what the problem is. But any setup errors / migration errors / database problems will show in the logs.

For sure. I will add this !

jptissot:
> No SQL Server ? Another pr?

I just have to figure out how to host it using the github actions services. Should be relatively easy.

test/Functional/cypress-commands/src/tenants.js
deanmarcussen: Yes, I ran into the same problem :) Same with MySql

What I meant was you have a prefix variable, and a name variable, (which are calculated to the same value), but instead of using the prefix variable here you use the name. 

Ah but I suddenly got it. The prefix variable is the url prefix, not the table prefix.

Too many prefixes... confused me

jptissot:
@agriffard seems the assets need to be rebuilt ;) 
https://github.com/OrchardCMS/OrchardCore/pull/7581/checks?check_run_id=1390462531

test/Functional/cypress-commands/src/tenants.js
jptissot: Ahhhh I get it. The only reason it's there was to allow flexibility for folks that want a different prefix. I guess it is a bit redundent. prefix should be renamed to urlPrefix to avoid confusion. Maybe also introduce a tablePrefix variable ? :P

sebastienros:
SQL Server is also available as a docker image. I assume that's what you use for other dbs, so do the same. Works on linux and windows.

I already answered on skype, but since you asked here too the windows containers has native windows bits. Same for linux. Meaning the content is different.

test/OrchardCore.Tests.Modules/BaseThemeSample/BaseThemeSample.csproj
sebastienros: I think we should revert these changes since you are now filtering on the artifacts, and being able to package might be necessary to test the modularity

sebastienros:
At that point we can also remove appveyor and handle master too, it's just about adding the nuget step
You can create a variable `isReleaseBranch` to group everything that has to be done on master,dev or release/.

jptissot:
@sebastienros SQL Server done ;)

jptissot:
Also, I have seen this unit test fail sometimes: https://github.com/OrchardCMS/OrchardCore/pull/7581/checks?check_run_id=1391788666

sebastienros:
I still think that everything that is tested in the db integration tests could also be done in the functional tests. Meaning we would not need this extra "kind" of test with the required setup. Unless these integration tests run on all oses, which we don't do for functional tests.

deanmarcussen:
I am coming round to agreeing about dropping the cross database integration tests, now that cypress is here, and its easier to write functional tests that prove something useful

PR #7583, opened 2020-11-11
Tenant link should open in new window

scleaver:
Quick one which I missed in #7475  - tenant link should open in a new window to be consistent


hishamco (approved):
LGTM

PR #7590, opened 2020-11-12
Allow multiple email authors

hishamco:

PR #7593, opened 2020-11-12
ReplyTo must have author addresses if not set

hishamco:
According to the MailKit docs http://www.mimekit.net/docs/html/P_MimeKit_MimeMessage_ReplyTo.htm

> When the list of addresses in the Reply-To header is empty, replies should be sent to the mailbox(es) specified in the From header


hishamco:
This should be merged after #7592 like #7590

PR #7602, opened 2020-11-12
Documentation: Added precedence info for placement

NogginBox:
There was no info to explain what order placement was applied in and it was not clear that a theme template could not change the placement of editor shapes.


NogginBox:
@deanmarcussen here is the documentation change we discussed. Mostly taken from your explanation to me. I've flipped the order so that the most likely to be used is at the top. I thought that was clearer?

PR #7604, opened 2020-11-12
Refactor GitHub actions

jptissot:
Pull requests targeting dev:
- Build & test on linux and windows
- Run functional tests on linux + sqlite

Dev branch: 
- Build & Tests on linux & windows
- Push nuget packages from linux
- Push docker linux and windows image

Manual steps
- MacOS unit tests
- Multiple DB functional tests
- Frontend assets build validation

### DIFF
--- a/.github/workflows/release_ci.yml
+++ b/.github/workflows/release_ci.yml
@@ -1,4 +1,4 @@
-name: Dev - CI
+name: Release - CI
 on:
   push:
     paths-ignore:
--- a/.github/workflows/assets_validation.yml
+++ b/.github/workflows/assets_validation.yml
@@ -1,5 +1,6 @@
 name: Assets build validation
-on: 
+on:
+# manual trigger
   workflow_dispatch:
 jobs:
   test-npm-build:
--- a/.github/workflows/assets_validation.yml
+++ b/.github/workflows/assets_validation.yml
@@ -1,4 +1,4 @@
-name: Assets build validation
+name: Frontend Assets build validation
 on:
 # manual trigger
   workflow_dispatch:
--- a/.github/workflows/assets_validation.yml
+++ b/.github/workflows/assets_validation.yml
@@ -1,0 +1,29 @@
+name: Assets build validation
+on: 
+  workflow_dispatch:
+jobs:
+  test-npm-build:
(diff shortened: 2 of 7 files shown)


.github/workflows/release_ci.yml
```diff
@@ -0,0 +1,75 @@
...
+        os: [ubuntu-latest, windows-latest]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup .NET Core on ${{ matrix.os }}
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 3.1.404
+    - name: Set build number 
+      if: matrix.os == 'ubuntu-latest'
+      run: echo "BuildNumber=$(( $GITHUB_RUN_NUMBER + 15471 ))" >> $GITHUB_ENV
```
deanmarcussen, line 29: @jptissot cloudsmith has two builds of 15472 (latest).

I am guessing this is because you made a new workflow, and the `GITHUB_RUN_NUMBER` started from 0 again for the workflow?

So another merge should increment it?

But will wait for you to check and confirm this, before we merge anything else.

jptissot: Hmmm strange. Maybe that's why it did that. Because I changed the name of the workflow. We should make sure this number gets incremented properly. Or maybe it's because the number starts at 0?

jptissot: Actually I think it's the symbols. Maybe I should disable them? Look at the last ci run. I'm building the symbols along with the packages.

deanmarcussen: Docs say it starts at 1 https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables

But is unique for each workflow name, so because the previous workflow also started at 1, it's going to add up to `1 + 15471` = `15472` on for both of the two runs on the different workflow names.

I suspect it just needs another small test ci pr to be merged to go to the correct version number.

But you wrote it so I let you decide :)

deanmarcussen: The fastest gun in the west @agriffard has decided to test this theory :)

Looks like it' going to `15437` so just the one set of packages duplicated, and they're old now so all g.

PR #7605, opened 2020-11-13
Dark mode V2 (cookie based)

### LINKED ISSUE #7519: Admin menu link - icon picker: background single icon dark
If you try to add an icon to a admin menu link I see this page:
admin_menu_icon

Skrypt:
- [x] Cookie based settings storage.
- [x] Use only one CSS to prevent flickers and animations when switching from light to dark mode and vice versa.
- [x] Use data-theme on html element like suggested by @mdo in Bootstrap issue.
- [x] Fix #7519, #7700, modals
- [x] Fix GraphiQL styles (Codemirror global styles conflict)


hishamco:
Is the aim of this PR to have a different strategy for tracking the theme mode or replace the old one?

Skrypt:
A little bit of both. It fixes issues and refactor some things because that's how I should have done it the first time but it seemed more complex to achieve. Mainly because it took me some time to figure out how to combine 2 themes in one .css file. I used the proposed method by the guy who works on Bootstrap instead of using 2 different .css files. The issue about that is that when you switch from one .css to the other it tries to execute css animations which creates some visual flickering. Also, I kept the LocalStorage but now I combine it with a cookie to store the `adminPreferences`. This way I can access the cookie value from server side and render the HTML properly without needing to use javascript to switch it right after the document.ready (more flickering).

Skrypt:
@agriffard don't merge yet I need to understand why I get so much updated assets and package.lock.json files.

hishamco:
Seems it's merge too early ;)

Skrypt:
You found an issue yet? lol

hishamco:
I was waiting to mark it as ready to test it, but seems it's merged so fast, I will try to test

Skrypt:
Oh well you know it's the same thing just done differently. Nothing new in the sky 😄 
The CSS is more strict this way and I like it. The result though is the same than before.

Jetski5822:
Dark Mode!!! Awesome :)

Skrypt:
Enjoy! 😄

PR #7613, opened 2020-11-13
Only use email in ResetPassword

### LINKED ISSUE #7540: UserService.FindByUsernameOrEmailAsync could find a user other than the one intended
[UserService.FindByUsernameOrEmailAsync](https://github.com/OrchardCMS/OrchardCore/blob/7a2e676b44dd1c9070c9b8495e6a6c07aa99c941/src/OrchardCore/OrchardCore.Users.Core/Services/UserService.cs#L213) first tries to find a user by username, then by email address.

However, what should happen in this scenario, where the email address of one user is the username of another? I was able to create these users in a fresh site using the blog theme.

image

It seems like `UserService.FindByUsernameOrEmailAsync` will always prefer the user with the name over the user with the email.

At a minimum, I know the reset password feature uses this method, so the reset email could be sent to the wrong user.

agriffard:
Fixes #7540


src/OrchardCore/OrchardCore.Users.Core/Services/UserService.cs
```diff
@@ -179,7 +179,7 @@ public async Task<bool> ResetPasswordAsync(string userIdentifier, string resetTo
                 return result;
             }
 
-            var user = await FindByUsernameOrEmailAsync(userIdentifier) as User;
+            var user = await _userManager.FindByEmailAsync(userIdentifier) as User;
```
rjpowers10, line 182: Should userIdentifier be renamed to email? I believe there is also a check above that says "A user name or email is required.", which isn't accurate anymore.

rjpowers10 (commented):
I'm not sure this completely resolves #7540 

`UserService.GetForgotPasswordUserAsync` has the same problem. Seems to me like `UserService.FindByUsernameOrEmailAsync` is just a flawed method and needs to be removed. `GetForgotPasswordUserAsync` would also call `UserManager.FindByEmail`.

rjpowers10 (commented):
Sorry, I think there are a few last pieces.

- [ForgotPassword.cshtml](https://github.com/OrchardCMS/OrchardCore/blob/c953909cf9cb399df78fd2eb1cff21d28dfe11ff/src/OrchardCore.Modules/OrchardCore.Users/Views/ResetPassword/ForgotPassword.cshtml#L15) still says "Email or username"
- Might want to update [ForgotPasswordViewModel](https://github.com/OrchardCMS/OrchardCore/blob/c953909cf9cb399df78fd2eb1cff21d28dfe11ff/src/OrchardCore.Modules/OrchardCore.Users/ViewModels/ForgotPasswordViewModel.cs#L8) similar to [ResetPasswordViewModel](https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.Users/ViewModels/ResetPasswordViewModel.cs) in that it prompts for and validates the email.

Thank you for taking this!

src/OrchardCore.Modules/OrchardCore.Users/ViewModels/ForgotPasswordViewModel.cs
```diff
@@ -1,10 +1,27 @@
...
+        public string Email { get; set; }
+
+        public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
+        {
+            var S = validationContext.GetService<IStringLocalizer<ForgotPasswordViewModel>>();
+            if (string.IsNullOrWhiteSpace(Email))
+            {
+                yield return new ValidationResult(S["Email is required."], new[] { nameof(Email) });
+            }
+            else if (Email.IndexOf('@') == -1 || !MailboxAddress.TryParse(Email, out var emailAddress))
```
hishamco, line 21: We already have `IEmailAddressValidator` on Infrastructure APIs that could be used here

hishamco (changes requested):

PR #7618, opened 2020-11-13
await registrationEvents

giannik:


src/OrchardCore.Modules/OrchardCore.Users/Controllers/ControllerExtensions.cs
```diff
@@ -84,7 +84,7 @@ internal static async Task<IUser> RegisterUser(this Controller controller, Regis
                             await signInManager.SignInAsync(user, isPersistent: false);
                         }
                         logger.LogInformation(3, "User created a new account with password.");
-                        registrationEvents.Invoke((e, user) => e.RegisteredAsync(user), user, logger);
+                        await registrationEvents.InvokeAsync((e, user) => e.RegisteredAsync(user), user, logger);
```
jptissot, line 87: I don't think this will properly await everything /cc @jtkech

jtkech: @jptissot this is the rught fix

@giannik good catch

jtkech (approved):

src/OrchardCore.Modules/OrchardCore.Users/Controllers/ControllerExtensions.cs
jptissot: Perfect thanks !

PR #7621, opened 2020-11-13
Fix custom settings to be able to support parts

### LINKED ISSUE #7620: Custom settings needs a fix to support parts
as we recently had to do for menus, taxonomies

deanmarcussen:
Fixes #7620


agriffard:
Does it fix an existing issue?

jptissot (approved):

sebastienros:
Was it already handled in the Custom Settings code?

PR #7634, opened 2020-11-14
Changed GetCenters() to GetAnchors() in media documentation

netwavebe:


Skrypt:
There is a CI issue here. It did not trigger the CI build probably because you only modified documentation.
/cc @jptissot

hishamco:
@Skrypt I noticed that in my previous two PRs

jptissot:
That is wanted behavior, no? Docs modifications should not trigger the ci!

Skrypt:
For pull request true. But the issue then is that we can't merge it if the CI doesn't confirm a sucessful build. I guess only @agriffard can.

jptissot:
Ahh I see. Can you file an issue ?

PR #7643, opened 2020-11-15
AutoroutesEntries Atomicity and Size

### LINKED ISSUE #7664: Autoroute Document Atomicity and Size
- Updating any content item (with an autoroute) also update the same shared `AutorouteDocument`, **so here atomicity is important**, locally and across running instances.

- For the same reason the `AutorouteDocument` **may become too big, and then saving a content item may be too long**, e.g. with one million of content items, updating only one content item would take around one minute.

jtkech:
Fixes #7664


src/OrchardCore/OrchardCore.Infrastructure/Documents/VolatileDocumentManager.cs
```diff
@@ -55,9 +55,16 @@ public Task UpdateAtomicAsync(Func<Task<TDocument>> updateAsync)
...
-                updated.Identifier ??= IdGenerator.GenerateId();
-                await SetInternalAsync(updated);
+
+                TDocument document = null;
+                foreach (var d in _updateDelegateAsync.GetInvocationList())
+                {
+                    document = await ((UpdateDelegate)d)();
+                }
+
+                document.Identifier ??= IdGenerator.GenerateId();
```
agriffard, line 65: I didn't know the null-coalescing assignment operator `??=`:
https://docs.microsoft.com/fr-fr/dotnet/csharp/language-reference/operators/null-coalescing-operator

sebastienros:
CQRS?

What is this fixing? That looks so complicated, nobody will ever be able to change this or fix something without breaking something else.

jtkech:
@sebastienros 

> CQRS?

Yes, for now this PR (marked as not ready) is just for testing another approach to fix the following concerns

> What is this fixing?

- **Atomicity**: Updating concurrently the same content item happens much less often than updating concurrently different ones, but in the last case we update the same `AutorouteDocument` for all items, **so here atomicity is important**.

    Note: Before using a shared `Document`, on a given instance at least we were using a local lock

- **Size**: The `AutorouteDocument` (related to all items) grows up as we create new content items, so the time to retrieve / update it may become too long. **E.g. with one million of items you will need one minute to save one content item**.

    So here the idea is to **not share the whole document but only the n last events**. E.g. in place of getting the whole, we get a much lighter document that is used, but only if its identifier has changed, to update a bigger local document.

    Idem on updating, in place of updating the whole, we update a lighter document composed of the n last events

src/OrchardCore/OrchardCore.Infrastructure/Documents/VolatileDocumentManager.cs
jtkech: Yes, we already use it in some places

jtkech:
CQRS

https://docs.microsoft.com/fr-fr/azure/architecture/patterns/cqrs

> Some implementations of CQRS use the Event Sourcing pattern. With this pattern, application state is stored (in our case in the distributed cache as autoroute data are not persisted) as a sequence of events (here renamed commands). Each event (command) represents a set of changes to the data. The current state (local autoroutes entries of a given instance) is constructed by replaying the events (here the commands that are not yet processed) ...

By taking into account what I added in parentheses, this is something like this that i'm doing here.

jtkech:
@sebastienros 

- So we planed to use the `AutoroutePartIndex` table to sync autoroute entries, and the `ContentItemIndex` table to know when a content item is soft removed, and having 2 cursors on these tables.

    Here i had a question, because there is no yesSql `LastOrDefautAsync()`, to get the last id of the `ContentItemIndex` table i used `OrderByDescending()` + `FirstOrDefaultAsync()`, **is it a problem to use this ordering on a big table?**

- Anyway, I found a way to only use the `AutoroutePartIndex` table by updating the index provider so that a record is added to the table when an item is soft removed (not all versions) OR when the autoroute is disabled OR the path is null.

    So now we resolve the autoroute entries from only one index table and always from this table. So i could remove some code in the part handler, in place of resolving new local entries from the item being updated, we now always resolve them from the new index table records (after the cursor), and only if the session commit is successful.

@deanmarcussen because autoroutes now have contained items routing, autoroute part disabled setting, and so on, even if this PR is not yet completed, can you try it at some point to see if i didn't break anything ;)

jtkech:
Marked as `notready` again but just because i started another PR with the same pattern but for `LocalizationEntries`

And i want to make them consistent with each other

src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
```diff
@@ -65,30 +62,14 @@ public class AutoroutePartHandler : ContentPartHandler<AutoroutePart>
...
-                    new AutorouteEntry(part.ContentItem.ContentItemId, part.Path)
-                };
-
-                if (part.RouteContainedItems)
-                {
-                    _contentManager ??= _serviceProvider.GetRequiredService<IContentManager>();
-
-                    var containedAspect = await _contentManager.PopulateAspectAsync<ContainedContentItemsAspect>(context.PublishingItem);
-
-                    await PopulateContainedContentItemRoutes(entriesToAdd, part.ContentItem.ContentItemId, containedAspect, context.PublishingItem.Content as JObject, part.Path, true);
```
deanmarcussen, line 88: @jtkech 

So `PopulateContainedContentItemRoutes` is still used by Pattern generation, but no longer here on published, if I am reading correctly.

There is something inside the method about homeroutes, which from memory allows a contained item to still be set as the home route.

So might still have to run some of it, for no purpose other than that?

src/OrchardCore.Modules/OrchardCore.Autoroute/Models/AutoroutePart.cs
```diff
@@ -20,6 +20,11 @@ public class AutoroutePart : ContentPart
         /// </summary>
         public bool Disabled { get; set; }
 
+        /// <summary>
+        /// Whether the related content item was removed.
+        /// </summary>
+        public bool ContentItemRemoved { get; set; }
```
deanmarcussen, line 26: This property is just to track it inside the index provider? 

And then not persisted?

Would it be better to have a scoped service to track removed, rather than adding a property?

Or too much...

deanmarcussen (commented):
Let me know when you're happy with this, and I'll test drive it

src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
jtkech: @deanmarcussen thanks for the feedback

From memory too ;) as i also started the same kind of PR for localization entries having their own singularities, so i will focus on it again this night, in the meantime,

> So PopulateContainedContentItemRoutes is still used by Pattern generation, but no longer here on published, if I am reading correctly.

Yes but the `AutoroutePartIndex` table is up to date and then re-used from its current cursor to update the entries.

> There is something inside the method about homeroutes, which from memory allows a contained item to still be set as the home route.

From memory i kept the part related to the homeroute for the container, about still allowing a contained item to be set as the home route, can't right now but i will check it this night

src/OrchardCore.Modules/OrchardCore.Autoroute/Models/AutoroutePart.cs
jtkech: Yes

> And then not persisted?

I was thinking about retrieving an item version without this state, e.g. with audit trail

> Would it be better to have a scoped service to track removed, rather than adding a property?

> Or too much...

Yes, here we just need a scoped state, i also thought about using our ShellScope items / features
But yes, maybe too much to have a service
i will think about it this night

jtkech:
@deanmarcussen 

- About the `part.ContentItemRemoved` state, i had a discussion with @sebastienros which suggested to use the scoped `IContentManagerSession`, currently it caches `_itemByVersionId` and `_publishedItemsById`, the idea is to also cache `_removedItemsById`. I will do it tomorrow

- So i didn't have so much time to review it, but i just did a fix about the **home route from a contained item**, this by calling a method that only checks if a contained item is defined as the home page, and breaking the loop if it finds one.

    With the current dev branch there is an issue, after having set a contained item as the home page, we can then set another regular item as the home, but it doesn't work because the previous `JsonPath` was kept in the home route settings, i fixed it by also passing an empty `JsonPath`

                await SetHomeRouteAsync(part, homeRoute =>
                {
                    homeRoute[_options.ContentItemIdKey] = context.ContentItem.ContentItemId;
                    homeRoute[_options.JsonPathKey] = "";
                });

deanmarcussen:
@jtkech sounds good

- About the home route. Good catch. The only reason I allowed it / implemented it, was because it was easier to allow it, than hide the button from the ui ;)

- About using the session. Makes sense. 
Would it make sense to make it more command orientated? 
or are we trying to stay away from that architectural style?

Instead of `_removedItemsById` it could be `_itemCommands = List<string, object>` where object is `RemovedItemCommand`

Then it could also be used for `SetHomeRouteCommand` for example, and we could lose the weird `SetHomePage` property - only because I've had to explain a few times to users why it doesn't stay set to true.

And then it is also easily extensible for other users to do other arrangements like you are doing here?
(maybe useful for the localization entries as well?)

Just an idea, maybe too much ;)

jtkech:
@deanmarcussen 

> Would it make sense to make it more command orientated?
or are we trying to stay away from that architectural style?

Maybe you are thinking about what i first did with a command list Document (shared through e.g. Redis), this is still a valid pattern when we can't share a whole document and then need to refresh it through another source of data. And this is what we are still doing with the `AutoroutePartIndex` table, that is not really a list of commands but a set of data from wich we produce them, but in that case @sebastienros prefers to share such data through the database. There is also the dedicated `IndexingTask` table where the `Type` column is a real command (Update = 0, Delete = 1)

> Instead of _removedItemsById it could be _itemCommands = List<string, object> where object is RemovedItemCommand

Above i was talking about using this pattern through a shared Document, and yes, if we need it locally, it is still a good pattern to use, but then i think not through `IContentManagerSession` that only caches items. But here removing an item is already a command that triggers handlers, and here we only need a new cached state for the index provider.

So here i think it is a specific case, just because an `IScopedIndexProvider` can't be also an `IContentHandler`, otherwise it would be able to cache removed items itself on the related event.

> Then it could also be used for SetHomeRouteCommand for example, and we could lose the weird SetHomePage property - only because I've had to explain a few times to users why it doesn't stay set to true.

Yes i agree, better to have a dedicated command on a given item, without having to check it when editing it. E.g. when listing the content items through the UI, it would be part of the available actions on a given item.

src/OrchardCore/OrchardCore.ContentManagement/DefaultContentManagerSession.cs
```diff
@@ -48,10 +49,26 @@ public bool RecallPublishedItemId(string id, out ContentItem item)
             return _publishedItemsById.TryGetValue(id, out item);
         }
 
+        public void AddRemovedContentItem(ContentItem item)
```
deanmarcussen, line 52: @jtkech for infos.

This is fine, what I was suggesting, is that maybe if it was more generic, we could use it for more things (i.e. SetHomePage) - and external users could use it as well...

Something like
```
private readonly List<ContentCommand>() _commands;

public abstract class ContentCommand
{
  public ContentItem ContentItem { get; set; }
}

public class RemovedContentItemCommand : ContentItemCommand
{
}

public class SetHomepageContentItemCommand : ContentItemCommand
{
}

public bool IsCommand<T>(ContentItem contentItem)
{
  return _commands.Any(x => x.ContentItem.ContentItemId = contentItem.ContentItemId && x is typeof(T));
}
```

That kind of thing. But was just an idea, and this is fine.

jtkech: @deanmarcussen 

Okay i understand, that's a good idea, the manager session might not be the better place to cache a kind of event / command, can you suggest a change in this PR, or in a separate PR if you don't have time.

Note: We need to compare by ref on the `ContentItem` instance, not by only using `ContentItemId`, because when removing an item, multiple versions may be removed, and we want to add only one record in the index table.

A list of commands is useful for history e.g. for syncing, to deffer some works e.g. to avoid duplicate commands or in a given order, maybe other cases, otherwise an event handler is enough. So, here is it only for index providers that can't be event handlers, and e.g. `SetHomepageContentItemCommand` would not be related to an index provider.

**Update**: In fact a scoped index provider can be an `IContenHandler`, it only needs to not resolve `ISession` as it is resolved while the session itself is resolved, but not so easy as we can't inherit from the base class `ContentHandlerBase` as we already derive from the base class `IndexProvider<T>` so we have to define all the many methods. So, what i did is to use `IndexProvider` instead and then just added this

        public string CollectionName { get; set; }

        void IIndexProvider.Describe(IDescriptor context)
        {
            Describe((DescribeContext<ContentItem>)context);
        }

        public Type ForType() => typeof(ContentItem);

Then i could derive from `ContentHandlerBase` and then just override `RemovedAsync()`, just tried it works, when deleting an item this `RemovedAsync()` is executed, right now it does nothing but i will try to use it this week end.

So maybe we could revert the session manager changes, but we would have to do the same for another index provider needing it, e.g. the localization index provider.

jtkech:
Hi @deanmarcussen 

Just to say that i like your idea to manage more abstract content item commands, but not sure if you meant to still use the manager session or another scoped service, and if you meant to still add a removed command in the content manager `RemoveAsync()` method, and then if we need to do the same in others content manager methods or after as needed.

In the meantime, i was able to **derive the autoroute par index from `ContentHandlerBase`**, not so difficult, so that it can enlist itself removed items (only if holding an autoroute part) and then check its own removed list on mapping.

It works and i could revert the changes related to `IContentManagerSession`, this is what I would have done first but I was thinking that it was not possible, so now i think that it's okay

But if you don't think so, feel free to update this PR with your suggestion if you have time ;)

deanmarcussen:
Hi @jtkech 

> In the meantime, i was able to derive the autoroute par index from ContentHandlerBase, not so difficult, so that it can enlist itself removed items (only if holding an autoroute part) and then check its own removed list on mapping.

I like this even better :)

My other thought was just a thought, maybe it is useful another time, for managing the `SetHomePage` better - but for another pr, and better not to be introducing something if we don't need it.

Plus it's kind of weird, to have both events, and commands.

Will try and try it, but will be tomorrow

jtkech:
@deanmarcussen Cool then, i let you try it when you have time

Note: After my last updates using `ContentHandlerBase`, I didn't retry all use cases

src/OrchardCore.Modules/OrchardCore.Autoroute/Drivers/AutoroutePartDisplay.cs
```diff
@@ -117,7 +115,7 @@ public override async Task<IDisplayResult> UpdateAsync(AutoroutePart model, IUpd
                 // This can only validate the path if the Autoroute is not managing content item routes or the path is absolute.
                 if (!String.IsNullOrEmpty(model.Path) && (!settings.ManageContainedItemRoutes || (settings.ManageContainedItemRoutes && model.Absolute)))
                 {
-                    var possibleConflicts = await _session.QueryIndex<AutoroutePartIndex>(o => o.Path == model.Path).ListAsync();
+                    var possibleConflicts = await _session.QueryIndex<AutoroutePartIndex>(o => (o.Published || o.Latest) && o.Path == model.Path).ListAsync();
```
deanmarcussen, line 118: @jtkech as far as I know the index only ever contains published or latest. so do we need the extra check here?

deanmarcussen (commented):
@jtkech test drove it. seems all good 👍 

Is the localization one ready as well? I didn't try that

jtkech:
@deanmarcussen 

> Is the localization one ready as well? I didn't try that

Yes it is ready

src/OrchardCore.Modules/OrchardCore.Autoroute/Drivers/AutoroutePartDisplay.cs
jtkech: @deanmarcussen 

Yes we need it as we now also have removed items (only when soft removed, not all versions)

So with Published = Latest = false

deanmarcussen: Ah ok, so soft removed is allowing for if a document that is not latest, or published, to also be removed?

In which case the table now has Published = false and Latest = false

Sorry i'm being dumb here, but I just want to check what will end up in the table (permanently)

Does that mean it then stays in the table permanently, as an item that has been removed?
So the index is now tracking more document history? (like the `ContentItemIndex` does, but not items that have been superceded - so not removed, but replaced by the another draft)

jtkech: Yes, but the `AutoroutePartIndex` table will have much less record than `ContentItemIndex`, not all versions e.g. when you saved a draft and / or each time you publish an item, there is no new record.

For a given item, it will have just one more record and only if you soft remove it

jtkech: > Sorry i'm being dumb here

No problem ;)

jtkech: > Ah ok, so soft removed is allowing for if a document that is not latest, or published, to also be removed?

Sorry, when i say soft remove, it is as we already do, set published and latest to false, soft because the version is still in the database.

So this is only at the time we removed an item (we know that through the handler part) that it is still indexed.

E.g. when we publish a draft, the draft version is removed, but it is not indexed as we don't trigger the removed handler event

jtkech: So, for a given item, it will have just one more record

And only if you remove it, and only if it has an autoroute part

deanmarcussen (approved):

src/OrchardCore.Modules/OrchardCore.Autoroute/Indexes/AutoroutePartIndex.cs
```diff
@@ -43,18 +44,37 @@ public class AutoroutePartIndex : MapIndex
         public string JsonPath { get; set; }
     }
 
-    public class AutoroutePartIndexProvider : IndexProvider<ContentItem>, IScopedIndexProvider
+    public class AutoroutePartIndexProvider : ContentHandlerBase, IIndexProvider, IScopedIndexProvider
     {
         private readonly IServiceProvider _serviceProvider;
-
+        private readonly List<ContentItem> _removed = new List<ContentItem>();
```
sebastienros, line 50: Hashset is better here

jtkech: Done

Issue #927, opened 2017-08-20
Should Orchard be using email instead of username


scleaver:
It appears to be the more common practice to use email address nowdays rather than a username to login with especially if you want to implement third party auth (like social login).

Should Orchard Core be following this trend rather than using username?

Skrypt:
Both is fine as long as a email is required.

scleaver:
So users can sign in with email instead or username?

sebastienros:
See https://github.com/OrchardCMS/Orchard2/issues/739#issuecomment-323624200

sebastienros:
Maybe we just need to do two lookups in the controller ... I can see how it could be confusing.

scleaver:
I agree Sébastien - I think the login username field should be "Username or password". A lookup would work although you may need to also prevent a email address being used in the username field in the registration process so you don't get the situation where someone uses an email address in the username... this may cause issues? A lot of sites/platforms seem to have dropped username altogether but your solution allows for choice of either I guess.

PR #2744, opened 2018-11-27
Enable username or email as login credentials

petedavis:
Update to the account controller to enable logging in with either username or email.

I haven't indicated this on the template, but it also enables sites to use a login form that shows just email/password form, which is reasonably common practice these days.

PR #7645, opened 2020-11-15
Login Via Email or User Name

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/5409
Fixes https://github.com/OrchardCMS/OrchardCore/issues/927

Replaces https://github.com/OrchardCMS/OrchardCore/pull/2744 (please leave open there are many valuable comments)

This pr replaces the default ASP.NET Identity `UserValidator` with a `UserAccountValidator`.

It validates that a username and email are unique as a whole, enabling the idea of safe(r) usage of the two values.

It moves away from the defaults of identity, enforcing required Orchard Core rules around email and user name usage. 

Obviously this validator could be replaced again to break the system, but so could many parts of Orchard Core.

There's a bit more work to be done (in a later pr) if we decide to take this, as there is a lot of various validation rules incorrectly scattered around controllers (Registration controller, User Drivers etc) performing adhoc validations that should all be rationalised into this validator.

But before we go refactoring all of that, I figure lets try and keep this pr clean, and focused on the issue.

/cc @kevinchalet @sebastienros


src/OrchardCore/OrchardCore.Users.Core/Services/UserAccountValidator.cs
```diff
@@ -0,0 +1,112 @@
...
+    {
+        private readonly IdentityErrorDescriber Describer;
+        private readonly IEmailAddressValidator _emailAddressValidator;
+
+        public UserAccountValidator(IdentityErrorDescriber identityErrorDescriber, IEmailAddressValidator emailAddressValidator)
+        {
+            Describer = identityErrorDescriber;
+            _emailAddressValidator = emailAddressValidator;
+        }
+
```
agriffard, line 24: 2 empty lines

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserInformationDisplayDriver.cs
```diff
@@ -63,46 +42,14 @@ public override async Task<IDisplayResult> UpdateAsync(User user, UpdateEditorCo
         {
```
agriffard, line 42: Format document.

agriffard:
You don't set or force `IdentityOptionsUser.RequireUniqueEmail` ? 
It is not a setting?

The fact that you can use an email does not appear in any UI or doc?

kevinchalet:
> It validates that a username and email are unique as a whole, enabling the idea of safe(r) usage of the two values.

Safer, not sure, as it allows squatting email addresses, as I described here: https://github.com/OrchardCMS/OrchardCore/pull/2744#issuecomment-723302715

deanmarcussen:
> Safer, not sure, as it allows squatting email addresses, as I described here: [#2744 (comment)](https://github.com/OrchardCMS/OrchardCore/pull/2744#issuecomment-723302715)

Hmm interesting @kevinchalet . 

So a change to the ruleset to, if the `UserName` is an email address, it must match the `EmailAddress` then changes the entire issue, of can a username be used to spoof an email. As then validation fails when the `name != email`

tbc.

@agriffard no `RequireUniqueEmail` setting is ignored (as it is currently, by a lot of the driver code and registration code - as they validate all on their own)

Docs, and UI to come. When / if satisfied with the validation.

kevinchalet:
@deanmarcussen one "simple" way to avoid all these concerns would be to remove `@` from the allowed username chars 😅

deanmarcussen:
@kevinchalet I wish it was that simple ;) 
Too many users are already using email = username to get around the problem of having a login that matches their email address (I know I am for some front end users/members, and I think @Skrypt is too, so if we are, there are sure to be others.)

Skrypt:
Well you know I don't mind what @kevinchalet is suggesting but then allow to login by using the real email field then.

deanmarcussen:
Updates/summary @kevinchalet and @sebastienros 

Design

1. I think we should support being able to log with a username or an email, and it might just be the only option
2. I don't think we should support being able to log with only the email or only the username (i.e. prevent the other one) as long as we don't expose if the entry exists in the db when the login fails, to prevent lookups
3. Unicity:
usernames should be unique (and case insensitive to prevent case spoofing)
emails should be unique (and case insensitive to prevent case spoofing)
4. Usernames and Emails should be unique, as a whole, such that an existing email can't be used as a new username
5. Like this PR we should be able to enforce the email to be valid before enabling an account, as a setting (in the admin)
6. Usernames should not be changed if they are used to represent a user reference (they are). Or we should introduce a new identity for users and use it, and also store usernames for historical data, not reference integrity (like we need to for tenants)
7. Emails can be changed as they are not used as user identity (and they shouldn't, because they can be changed)
8. I don't see why the username and the email should not have the same value, even if they can be changed individually, as long as the value doesn't collide with another existing user. There should then be no validation on the username, but validate the email.

Actual
- login via email or username
- username is unique and case insensitive
- emails are required, and unique
- when username is an email address, it must be the same email as the email field
    - this enforces unicity and makes it impossible to try and create an account that would be matched against another users email.
    - it also prevents account squatting
    - this will also mean that if the user does this, e.g. name=email, then if they change their email, they must change their username to match.
- all these rules are also applied during editing / updating username and email address

sebastienros:
I would suggest to remove the `@` from the `AllowedUserNameCharacters` by default such that emails and usernames can't match.

Any existing username with an `@` should be updated to remove it. Then we can authenticate with either username or email.

sebastienros:
For our front-end registration UI, we could add an option to require the username and/or the email, and generate a username automatically if not required, using a pluggable logic.

deanmarcussen:
Ok, so updated as discussed, to force `IdentityOptions` to require a unique email, and that the UserName can no longer be an email.

Migration added to migrate any existing user names from `foo@foo.com` to `foo+foo.com`

@kevinchalet if there are still concerns that someone can change `IdentityOptions` without realising the impact, we could potentially throw an exception from somewhere like the `UserStore` if the options are changed away from the requirements.

i.e.
```
if(options.User.AllowedUserNameCharacters.Contains('@') || options.User.RequireUniqueEmail == false) 
{
     throw etc...
}
```

test/OrchardCore.Tests/OrchardCore.Users/UserValidatorTests.cs
```diff
@@ -0,0 +1,150 @@
...
+            Assert.False(result.Succeeded);
+            Assert.Equal(describer.DuplicateEmail(user.Email).Code, result.Errors.First().Code);
+        }
+
+        [Fact]
+        public async Task ShouldRequireUserNameIsNotAnEmailAddress()
+        {
+            // Arrange
+            var describer = new IdentityErrorDescriber();
+            var userManager = UsersMockHelper.MockUserManager<IUser>();
```
deanmarcussen, line 108: I kept the tests, although identity also has these tests.

Mostly just to indicate that we have specified a requirement.

They could go if not wanted

agriffard:
Conflict to solve.

agriffard:
@kevinchalet Unless you have some remarks to add, we are going to merge this in the next few days.

fzadrazil:
FYI, not every username which contains @ must be email address. This PR just broke our system because all user names are in the form of name@ldap_domain, but it is not an email. And as we are synchronizing user accounts from Shibboleth which is used in the organization, we cannot have different usernames (with + instead of @) in Orchard.

Skrypt:
You can decide to allow @ symbols by allowing it in the ASP.NET Core Identity middleware configuration.

https://github.com/OrchardCMS/OrchardCore/blob/f0d3aa0be7b6971705a37808be01fc2f47a05c7c/src/OrchardCore.Modules/OrchardCore.Users/Startup.cs#L144

Though, I don't recommend this. We removed @ symbols in the user names for a security reason.

fzadrazil:
Yep, we have already made this change and reversed the migrations as it has rewritten all our usernames (@ => +) which is not acceptable for our scenario. I just wanted to point out that replacing @ with + is rather extreme and can be a breaking change for some organizations.

sebastienros:
We only identified a few sites using @ in usernames, and migrating was fine for these.
Can you confirm if you have mitigated that issue? We had to make this the new default since it's a security concern. As long as you don't allow custom registration on your system I assume that's not an issue.

fzadrazil:
> As long as you don't allow custom registration on your system I assume that's not an issue.

Exactly. We don't allow registration, we just synchronize the user accounts from another system (many different identity providers on our portals). I have just allowed @ and reversed the migrations, which works for us, but will complicate future Orchard updates, so it will be the best if we could configure this for each tenant via Dashboard.

Issue #6364, opened 2020-06-07
OpenId Deployment and Recipes have some issues


deanmarcussen:
There are a number of issues with the OpenId Recipe and Deployment Steps

- The Deployment Step for the Server Settings produces **completely incorrect** JSON to match what the Recipe Step is expecting (the settings directly, when the step expects a Model)

- Some of the Steps use Models, some use `CreateViewModels`

- Most of the Steps are storing their data directly in the step property, where they should be keyed. (Minor issue)

- The Application Step can only be run once, as it does not check to see if the application already exists, so throws an exception the second time.

Tsjerno:
Added some code to fix the issues
See https://github.com/OrchardCMS/OrchardCore/pull/7646

kaipm:
I'm working on the view model refactor, but I got stuck a little. It's more a decision, rather than a technical problem.

Due to the recipe using a view model, currently only the following format works in tests:
```
  "steps": [
    {
      "name": "OpenIdApplication",
      ...
      "roleEntries": [
        {
          "name": "role1",
          "selected": true
        }
      ]
    }
  ]
```
`selected` missing or equal to `false` results in the role beeing ignored.

All the other list-properties use a string, delimited by space like this:
```
...
"redirectUris": "https://uri1 https://uri2",
...
```

I'm not sure how I should fix this now. I could:
- replicate this format, so existing recipes don't break
- introduce breaking changes:
  - make the property a string, delimited by space, to match other properties
  - eventually also rename it to `roles` or `clientCredentialRoles`
- try to support both formats, with the caveat of additional complexity

Any advice would be appreciated!

sebastienros:
I understand the issue, the step is using a "View Model" that is designed for checkboxes, and in aspnet it needs a bool to map the state of the checkbox. In the step it just need a presence. So maybe change the step logic by removing the `Select(r => r.Selected == true)` and take all the steps that are listed. Or use a different class that also has `RoleEntries` but no `selected` field. And stop using the View model class.

kaipm:
Right, I'll do that then, thank you!

kaipm:
> There are a number of issues with the OpenId Recipe and Deployment Steps
> 
> * The Deployment Step for the Server Settings produces **completely incorrect** JSON to match what the Recipe Step is expecting (the settings directly, when the step expects a Model)
> * Some of the Steps use Models, some use `CreateViewModels`
[...]

PR #7646, opened 2020-11-15
Update of open id recipes (issue #6364)

Tsjerno:
Update of openID application Step 
- not using viewmode
 - recipe can be run multiple times
Update of openID scope Step
 - not using viewmode
 - recipe can be run multiple times ScopeController

Update of openID scope page
 - Give user feedback when spaces are added in the name


src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/ScopeController.cs
```diff
@@ -122,6 +122,11 @@ public async Task<IActionResult> Create(CreateOpenIdScopeViewModel model, string
                 ModelState.AddModelError(nameof(model.Name), S["The name is already taken by another scope."]);
             }
 
+            if(model.Name.Contains(' '))
```
kevinchalet, line 125: Consider moving that to the view model class using `IValidatableObject`.

src/OrchardCore.Modules/OrchardCore.OpenId/Recipes/OpenIdApplicationStep.cs
```diff
@@ -64,29 +80,56 @@ public async Task ExecuteAsync(RecipeExecutionContext context)
...
+                    from uri in model.RedirectUris
+                    select new Uri(uri, UriKind.Absolute));
+                }
+            }
 
-            await _applicationManager.CreateAsync(descriptor);
+            var application = await _applicationManager.FindByClientIdAsync(descriptor.ClientId);
+            if (application != null)
+            {
+                await _applicationManager.UpdateAsync(application, descriptor);
```
kevinchalet, line 127: If you want to support UPSERT, you'll need to make additional changes to this class to remove the permissions/redirects_uris/post_logout_redirect_uris. Take a look at how it's done in the app controller.

src/OrchardCore.Modules/OrchardCore.OpenId/Recipes/OpenIdScopeStep.cs
```diff
@@ -27,20 +28,29 @@ public async Task ExecuteAsync(RecipeExecutionContext context)
...
             {
-                descriptor.Resources.UnionWith(model.Resources.Split(new[] { " " }, StringSplitOptions.RemoveEmptyEntries));
+                descriptor.Resources.UnionWith(model.Resources);
             }
 
-            await _scopeManager.CreateAsync(descriptor);
+            var scope = await _scopeManager.FindByNameAsync(model.ScopeName);
+            if (scope != null)
+            {
+                await _scopeManager.UpdateAsync(scope, descriptor);
```
kevinchalet, line 48: Same remark, the current code is not suitable for UPSERT scenarios.

kevinchalet (changes requested):
Thanks for your PR!

src/OrchardCore.Modules/OrchardCore.OpenId/Recipes/OpenIdScopeStep.cs
Tsjerno: I get you're point have been looking at the code. Did test this easier solution and it seems to work.
At least, I expect that executing a recipe does a reset of the scope. This code resets the scope to the value's of the recipe.
My solution starts with an empty descriptor and then adds the values of the model. The controller starts with a descriptor based on the model, resets it and then fills it with the values of the model.

Tsjerno: Do I need to add explicit the current tenant to the resources, as is done in the controller? Was not sure about that.

kevinchalet: The problem with starting with an empty descriptor is that every custom permission/property won't be preserved as the recipe won't re-add it. It would be better to populate the descriptor and amend it as in the app/scopes controller.

src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/ScopeController.cs
Tsjerno: Changed the validation using an IValidatableObject

src/OrchardCore.Modules/OrchardCore.OpenId/Recipes/OpenIdScopeStep.cs
Tsjerno: Ok, will do that, what about the current tenant, which is added in the scope controller

kevinchalet: That's pretty much the opposite: the scope controller filters out the current tenant as it's always included later by the authorization controller: https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/ScopeController.cs#L145

That said, I think it would be nice to also have a `Tenants` property in the recipe and use the same logic.

Tsjerno: Added a TentantNames property to the recipe

agriffard:
@kevinchalet Can you please review this PR?

src/OrchardCore.Modules/OrchardCore.OpenId/Recipes/OpenIdApplicationStep.cs
```diff
@@ -29,24 +29,65 @@ public async Task ExecuteAsync(RecipeExecutionContext context)
                 return;
             }
 
-            var model = context.Step.ToObject<CreateOpenIdApplicationViewModel>();
+            var model = context.Step.ToObject<OpenIdApplicationStepModel>();
+            var application = await _applicationManager.FindByClientIdAsync(model.ClientId);
 
-            var descriptor = new OpenIdApplicationDescriptor
+            var descriptor = new OpenIdApplicationDescriptor();
+            if (application!=null)
```
kevinchalet, line 36: nit: spaces.

src/OrchardCore.Modules/OrchardCore.OpenId/Recipes/OpenIdApplicationStep.cs
```diff
@@ -29,24 +29,65 @@ public async Task ExecuteAsync(RecipeExecutionContext context)
...
             {
-                ClientId = model.ClientId,
-                ClientSecret = model.ClientSecret,
-                ConsentType = model.ConsentType,
-                DisplayName = model.DisplayName,
-                Type = model.Type
-            };
+                await _applicationManager.PopulateAsync(application, descriptor);
 
+                descriptor.Permissions.Remove(OpenIddictConstants.Permissions.GrantTypes.ClientCredentials);
```
kevinchalet, line 40: It would have been better to use the same `if`/`else` logic as the one used in the controllers to make things more consistent and avoid discrepancies between recipes and controllers, but it shouldn't be blocking to merge this PR,

src/OrchardCore.Modules/OrchardCore.OpenId/Recipes/OpenIdApplicationStep.cs
```diff
@@ -29,24 +29,65 @@ public async Task ExecuteAsync(RecipeExecutionContext context)
...
+            }
             if (model.AllowAuthorizationCodeFlow)
             {
                 descriptor.Permissions.Add(OpenIddictConstants.Permissions.GrantTypes.AuthorizationCode);
             }
             if (model.AllowClientCredentialsFlow)
             {
                 descriptor.Permissions.Add(OpenIddictConstants.Permissions.GrantTypes.ClientCredentials);
+
+                if (model.Roles!= null && model.Roles.Any())
```
kevinchalet, line 87: nit: formatting

src/OrchardCore.Modules/OrchardCore.OpenId/Recipes/OpenIdApplicationStep.cs
Tsjerno: Is it better to create an OpenIdApplicationService with an Create and Edit function with an ApplicationModel?
Then use this service in the recipe and the controller?

kevinchalet: It's an option, but the problem is that the step and view models are 2 different CLR classes, so we'd to either merge the classes or introduce some interfaces. Dunno if it's worth it or not.

If anything, it would be better to put that in `IOpenIdApplicationManager`.

deanmarcussen:
I think you had a bad merge here as there are over 1200 file changes.

Can you reset the pr please?

PR #7653, opened 2020-11-16
Generate correct urls for assets with no resizing commands

### LINKED ISSUE #7652: Version of AssetUrl within OrchardHelper for non-image assets
Need a helper extension equivalent of AssetUrl but for non-image assets mike PDFs and such so they don't unnecessarily get a token added to thier url. Would be good to include the option to append version.

Not sure what to call it... AssetUrl would have been good and then use ImageUrl specifically for images but bit late to change :)  - maybe MediaUrl.

scleaver:
Resolves #7652

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenService.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenService.cs
@@ -43,6 +43,13 @@ namespace OrchardCore.Media.Processing
             var pathParts = path.Split('?');
 
             var parsed = QueryHelpers.ParseQuery(pathParts.Length > 1 ? pathParts[1] : string.Empty);
+
+            // If no commands or only a version command don't bother tokenizing.
+            if (parsed.Count == 0 || parsed.Count == 1 && parsed.ContainsKey(ImageVersionProcessor.VersionCommand))
+            {
+               return path;
+            }
+
             var processingCommands = new Dictionary<string, string>();
             var otherCommands = new Dictionary<string, string>();
 


deanmarcussen:
I don't think we need another helper for this.

It's handled already here https://github.com/OrchardCMS/OrchardCore/blob/aa7a6078899c651404ed8e2b656da865e71cad49/src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaImageSharpConfiguration.cs#L79

So the issue you have is that you want the query string to say `v=` rather than `token=`, when it only has the version supplied and no resizing arguments?

It's more of a visual on the query string, rather than an actual issue, I take it? 
as the token in this case, will just be a tokenized version of the `v=` value, so just as useful for cache busting.

I would suggest just doing the same logic in the `IMediaTokenService` and not returning a tokenized string, if there is only one query string value, and it is `v`

scleaver:
No probs. Yeah I guess it is just a visual thing. The old v= token was quite short and I'm just not a fan of unnecessarily long Paramus. 

When you go to share a link to a PDF URL via text or something it and it's like `https://www.mydomain.com.au/media/terms.pdf?token=CfDJ8LmBhn5Sz39GkzTyHvqUgEiVc3CtD0MLuB7z3C56hAk5JOgXmB6e6Cdd20lk8D8iAuGHFMY1RzZaSQKwxTG79iQcenlMGGIzFem6r1640aPCJtB4YeGwI2tqOnOgG3z6iz-D7ubfr9qOnoa0PM5ZBp4` - it seems a bit over the top. 

But it's just pure personal preference. No probs we can just close this PR.

deanmarcussen:
> I would suggest just doing the same logic in the IMediaTokenService and not returning a tokenized string, if there is only one query string value, and it is `v`

This change should be simple, and then problem solved?

scleaver:
Ok I will attack it that way.

scleaver:
Ok so I have investigated and you are right... it's already handled and in fact if there is only a `v` param the token service doesn't get hit and v= remains BUT I'm not going crazy - this is only working in liquid:

Liquid: `{{ project.Project.Image.Paths.first | asset_url | append_version }}`
Result: `https://localhost:5001/media/portfolio/02-full.jpg?v=QAhc-s4PczoDn15UXbLcNz3wCOn-srf6-HLyrcfTlgE`

C#: `_orchardHelper.AssetUrl(mediaField.Paths[0], appendVersion: true)`
Result: `https://localhost:5001/media/portfolio/02-full.jpg?token=CfDJ8AmYzF57vn5KpWiOixoQgbwEmC_2dkWnVkHqaGIoQidUM81-Ihd3bgwMuwWLpXvDQWsbauPqvSlUoUe6aDYgXUckvc1cRbVpXO3nMoiEDIwMZmjMEeITqgDyJZkuh-8pA_B-LdWEUigh5js_PksUxEGrt4GLGe6XuIzfpncEAY2cTnzCTDW9VZrDf-D94Nhyxg`

So I will investigate what the difference between the liquid helper and the orchardhelper is cause at the moment it doesn't look consistent.

deanmarcussen:
`{{ project.Project.Image.Paths.first | asset_url | append_version }}` I might have made a mistake here. Let me try and remember.

deanmarcussen:
No it's correct.

It's just the difference between liquid, which is a chained filter, i.e. `asset_url | append_version | resize_url` where things are added sequentially, and the `OrchardHelper.AssetUrl` where it's not a chain of events, but one single event.

With tag helpers it works because tag helpers can have an order, but with liquid the order is how the filters are placed.

So this order `asset_url | resize_url: width... | append_version` would not work either, as it'll get stripped by the command processor, because it skips the tokenization. But `asset_url | append_version | resize_url: width... ` is good :)

If that makes sense.

For me the good fix is still in `IMediaTokenService`

```
if (processingCommands.Count == 1 && processingCommands.ContainsKey("v") 
{
    otherCommands["v"] = processingCommands["v"];

    return QueryHelpers.AddQueryString(pathParts[0], otherCommands);
}
```

or in `AssetUrl` after adding the version.
`if (width == null && height == null &&... && ...)
{
  return resolvedAssetPath;
}

Either should do the trick. 
Maybe the last one looks worse codewise, but better placed in terms of it making sense as to where it's responsibility lies?

scleaver:
I noticed that doing simply `@Orchard.AssetUrl("about/1.jpg")` also unnecessarily caused a token.

Since I think what is needed is a consistent approach I decided to implement the change into the IMediaTokenService and do the check early in the method straight after the query string parsing as no change to the path is needed if the tests do not pass.

Helper: `@Orchard.AssetUrl("about/1.jpg")`
Result: `/media/about/1.jpg`

Helper: `@Orchard.AssetUrl("about/1.jpg", appendVersion: true, width: 100)`
Result: `/media/about/1.jpg?v=H5M_6JYJCnGUA5XLHiruNPyGHTEhztKQSUPVdfRte9o `

Helper: `@Orchard.AssetUrl("about/1.jpg")`
Result: `/media/about/1.jpg?token=CfDJ8B8ozCR_ikFFma4hCzJHSctszVhzB3s0X5nzNa6HBA4JTQuKLvbPT_VecAre1veQCyIPMx3azaEyPhDt1pSc1R2c18Ig3nCzsZmSzCvQA9WGNwt-orRplpa39uj0frz6QuNicC7PrUaXTIfMOYlsXIHAHwfT8Qt1CNJ-G2MwmLSBOdzLhdYnZA8Gdh-7-VnxlKHjaWPYKX632cVM42LAu-0 `

deanmarcussen (approved):
Ideal

PR #7662, opened 2020-11-16
Prevent creating folders with trailing space & check for invalid characters

netwavebe:
Folders should not have leading or trailing whitespace characters.

Creation succeeds but it's impossible to upload anything to these folders. They also can't be removed easily (at least not on Windows).

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
@@ -380,7 +380,7 @@ namespace OrchardCore.Media.Controllers
             var invalidFolderNameCharacters = new char[] { '\\', '/' };
             if (invalidFolderNameCharacters.Any(invalidChar => name.Contains(invalidChar)))
             {
-                return StatusCode(StatusCodes.Status403Forbidden, S["Cannot create folder because the folder name contains invalid characters"]);
+                return StatusCode(StatusCodes.Status400BadRequest, S["Cannot create folder because the folder name contains invalid characters"]);
             }
 
             var newPath = _mediaFileStore.Combine(path, name);
--- a/src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
@@ -377,7 +377,9 @@ namespace OrchardCore.Media.Controllers
                 path = "";
             }
 
-            var newPath = _mediaFileStore.Combine(path, name.Trim());
+            name = name.Replace('\\', '-').Replace('/', '-'); // Don't create nested folders
+
+            var newPath = _mediaFileStore.Combine(path, name);
 
             if (!await authorizationService.AuthorizeAsync(User, Permissions.ManageMedia)
                 || !await authorizationService.AuthorizeAsync(User, Permissions.ManageAttachedMediaFieldsFolder, (object)newPath))
--- a/src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
@@ -377,7 +377,11 @@ namespace OrchardCore.Media.Controllers
                 path = "";
             }
 
-            name = name.Replace('\\', '-').Replace('/', '-'); // Don't create nested folders
+            var invalidFolderNameCharacters = new char[] { '\\', '/' };
+            if (invalidFolderNameCharacters.Any(invalidChar => name.Contains(invalidChar)))
+            {
+                return StatusCode(StatusCodes.Status403Forbidden, S["Cannot create folder because the folder name contains invalid characters"]);
+            }
 
             var newPath = _mediaFileStore.Combine(path, name);
 
--- a/src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
@@ -377,7 +377,7 @@ namespace OrchardCore.Media.Controllers
                 path = "";
             }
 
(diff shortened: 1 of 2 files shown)


Skrypt:
Slugify them instead

netwavebe:
Mmm, interesting idea!

Pros: always valid urls for media.
Cons: can't have spaces between words nor use PascalCasing.

"ProductImages" would become "productimages"
"Product Images" would become "product-images"

Still, I like this approach. Any reason not to slugify folders @deanmarcussen ?

Skrypt:
Casing could be also slugified then.

netwavebe:
The same could be said for filenames.

deanmarcussen:
> Slugify them instead

I'm ok with just trimming here.

Slugify could be a feature.

But then someone will want pascal. and someone else will want camel...

Slugify would also strip diacritics in its current implement I think - which you made me fix @Skrypt :)

I do think this should be at service level, not in the controller though.

Blob storage accepts trailing spaces fine (but I've not checked the media app works with them. Imagine it does)

deanmarcussen:
> I do think this should be at service level, not in the controller though.

`Combine` I meant. Or the `NormalizePath` extension.

netwavebe:
@deanmarcussen Thanks for the feedback.

I found another issue when using a slash in the new folder name, like 'clients/suppliers'. This (obviously) creates a folder "clients" with another folder "suppliers" inside.   When naming the new folder 'clients / suppliers' I'm still getting folders with trailing/leading spaces... 😞

Don't think this is something I can fix in `Combine` or `NormalizePath`.

netwavebe:
@deanmarcussen Please check the last commit when you have time.

Had to check for `\` and `/` in the controller as `_mediaFileStore.Combine(path, name);` should be able to combine `aaa/bbb` and `ccc/ddd`. However, in this particular case the second part should not contain nested folders.

src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
```diff
@@ -377,6 +377,8 @@ public async Task<IActionResult> MoveMediaList(string[] mediaNames, string sourc
                 path = "";
             }
 
+            name = name.Replace('\\', '-').Replace('/', '-'); // Don't create nested folders
```
deanmarcussen, line 380: I'm not sure I agree with this change.

I would have thought it creating nested folders was a logical conclusion?

Seems like something we could discuss endlessly :)

Skrypt:
@deanmarcussen Why would we want to keep filenames without slugifying them?
What does Wordpress do in his media gallery? Do they slugify filenames and paths?

Azure Blog Storage allows to use trailing spaces on folder/filenames ; means in url's they will be html encoded to `%20`. As far as I remember, those are not SEO friendly. Maybe we could have a global media gallery setting to slugify all paths/files then. This way we could keep natural file/folder names or not. But let's see what other CMS's does too.

src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
netwavebe: It does create nested folders on Windows. For example, adding a folder named ´clients/suppliers' works fine (I get a folder 'clients' with another nested folder 'suppliers' inside).

Problem is when I add a folder named 'clients / supplier'. This adds a folder 'clients ' (with a trailing space!) with another nested folder ' suppliers' (with a leading space!) inside.

Folders with leading spaces are no problem, but trailing spaces are! You can't upload anything to it and you can't delete it (not in media library, not in file explorer). You would need a funky command like `rd "\\?\C:\path\media\clients "` to get rid of it.

I could split on `/`, trim each part and join again. Then I thought it probably wasn't the users intention to create nested folders, hence the change above.

Skrypt:
Ok let's not get carried away and fix the trailing space issue for now. Forget the slugify thing.

src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
netwavebe: @deanmarcussen FYI: even in the case when valid nested folders are created (valid = without trailing spaces) the media library doesn't handle this correctly. It shows the deepest folder as a direct descendant of the parent. We should prevent creating nested folders imho.

netwavebe: @deanmarcussen You're right!

netwavebe:
`Create Folder` now shows a message if the folder name contains invalid characters:

createfolder

I agree this solution is much better.

src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
```diff
@@ -377,6 +377,12 @@ public async Task<IActionResult> MoveMediaList(string[] mediaNames, string sourc
                 path = "";
             }
 
+            var invalidFolderNameCharacters = new char[] { '\\', '/' };
+            if (invalidFolderNameCharacters.Any(invalidChar => name.Contains(invalidChar)))
+            {
+                return StatusCode(StatusCodes.Status403Forbidden, S["Cannot create folder because the folder name contains invalid characters"]);
```
deanmarcussen, line 383: Looks great!

Call me Mr Picky but isn't model state errors `BadRequest` ?

https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.statuscodes.status400badrequest?view=aspnetcore-5.0

netwavebe: @deanmarcussen Yes, it is, I changed it.

deanmarcussen (approved):

deanmarcussen:
@Skrypt @netwavebe I still like the idea of slugifying, just as a feature if anyone wants to do it.

(I would use it!)

netwavebe:
@deanmarcussen @Skrypt I would use it too when starting on a new website. Might do this.

PR #7670, opened 2020-11-17
Add user_settings liquid filter

### LINKED ISSUE #7669: CustomUserSettings not accessible in liquid
According to the docs, accessing the UserProfile content type with a CustomUserSettings steretotype should be possible by using the following line : `{{ User.Properties.UserProfile.Content.UserProfile.FirstName.Text }}` But this does not work. 

I think the issue is that only the `HttpContext.User` is loaded and not the user entity.

/cc @deanmarcussen

jptissot:
Used to access CustomUserSettings

fix #7669

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Users/Liquid/UserPropertiesFilter.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Liquid/UserPropertiesFilter.cs
@@ -1,0 +1,31 @@
+using System.Security.Claims;
+using System.Threading.Tasks;
+using Fluid;
+using Fluid.Values;
+using Microsoft.AspNetCore.Identity;
+using OrchardCore.Liquid;
+using OrchardCore.Users.Models;
+
+namespace OrchardCore.Users.Filters
+{
+    public class UserPropertiesFilter : ILiquidFilter
+    {
+        private readonly UserManager<IUser> _userManager;
+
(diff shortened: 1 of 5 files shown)


deanmarcussen (changes requested):
I don't like it. Sorry going to be the code police :(

Should be through claims.

We shouldn't be retrieving the user from the database for this, it defeats the entire purpose of claims

jptissot:
I agree, but what about the use case where you want a piece of the Properties that you don't want exposed as claims ? This would suport this use-case ? Maybe I misunderstand the use of claims

sebastienros:
DEAN THE CODE POLICE !!!!! 👮 

What about renaming it to user_settings or user_profile to make it clear it's for profile information. And there is no option but to do a database query in this case. All example should just use things from custom profile settings.

Skrypt:
One doesn't prevent having the other too.

jptissot:
I think there is a use case for this filter. And another one for claims. That's why I create another issue.

Skrypt:
Ok was just thinking out loud. 😉

deanmarcussen:
We do need something but not sure this is it (yet).

Some feedback on the comments on this pr https://github.com/OrchardCMS/OrchardCore/pull/7570

As in how should we resolve users through liquid as a whole

jptissot:
This will be handled by the #7570 PR

Issue #7629, opened 2020-11-14
Remove the localized error descriptions from the authorization controller


jardg:
I'm using an API controller with authentication. My OC site default culture is spanish. If I make a request using as authorization header an expired access token I get a 500 internal error as status code instead of the 401 with the localized error_description in the header. 

This is the error I get:

> System.InvalidOperationException: Invalid non-ASCII or control character in header: 0x00E1
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpHeaders.ThrowInvalidHeaderCharacter(Char ch)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpHeaders.ValidateHeaderValueCharacters(StringValues headerValues)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpResponseHeaders.SetValueFast(String key, StringValues value)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpHeaders.Microsoft.AspNetCore.Http.IHeaderDictionary.set_Item(String key, StringValues value)
   at Microsoft.AspNetCore.Http.ParsingHelpers.SetHeaderUnmodified(IHeaderDictionary headers, String key, Nullable`1 values)
   at Microsoft.AspNetCore.Http.ParsingHelpers.AppendHeaderUnmodified(IHeaderDictionary headers, String key, StringValues values)
   at OpenIddict.Validation.AspNetCore.OpenIddictValidationAspNetCoreHandlers.AttachWwwAuthenticateHeader`1.HandleAsync(TContext context)
   at OpenIddict.Validation.OpenIddictValidationDispatcher.DispatchAsync[TContext](TContext context)
   at OpenIddict.Validation.OpenIddictValidationDispatcher.DispatchAsync[TContext](TContext context)
   at OpenIddict.Validation.AspNetCore.OpenIddictValidationAspNetCoreHandler.HandleChallengeAsync(AuthenticationProperties properties)
   at Microsoft.AspNetCore.Authentication.AuthenticationHandler`1.ChallengeAsync(AuthenticationProperties properties)
   at Microsoft.AspNetCore.Authentication.AuthenticationService.ChallengeAsync(HttpContext context, String scheme, AuthenticationProperties properties)
   at Microsoft.AspNetCore.Authentication.AuthenticationHandler`1.ChallengeAsync(AuthenticationProperties properties)
[...]

PR #7675, opened 2020-11-17
Bump OpenIddict to 3.0.0-rc1.20567.82

kevinchalet:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/7629.

https://kevinchalet.com/2020/11/17/introducing-openiddict-3-0-s-first-release-candidate-version/

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/AccessController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/AccessController.cs
@@ -10,7 +10,6 @@ using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Localization;
 using OpenIddict.Abstractions;
 using OpenIddict.Server.AspNetCore;
 using OrchardCore.Environment.Shell;
@@ -24,6 +23,8 @@ using static OpenIddict.Abstractions.OpenIddictConstants;
 
 namespace OrchardCore.OpenId.Controllers
 {
+    // Note: the error descriptions used in this controller are deliberately not localized as
+    // the OAuth 2.0 specification only allows select US-ASCII characters in error_description.
     [Authorize, Feature(OpenIdConstants.Features.Server)]
     public class AccessController : Controller
     {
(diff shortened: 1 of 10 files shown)


hishamco:
Build is failed

kevinchalet:
I know. That's because it requires the .NET 5 SDK, that OC doesn't use it yet. We're discussing this change on Gitter ATM.

Skrypt:
Should it not target then https://github.com/OrchardCMS/OrchardCore/tree/ag/net5 instead?

kevinchalet:
@Skrypt no: according to @agriffard, this branch - that makes a full migration to .NET 5.0 (SDK + TFMs + packages dependencies) will only be merged after OC 1.0 ships.

In this case, I don't need OC to target .NET 5.0. I just need the 5.0 SDK to fix a bug that affects nullable references.

src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/AccessController.cs
```diff
@@ -99,7 +97,7 @@ public async Task<IActionResult> Authorize()
                     {
                         [OpenIddictServerAspNetCoreConstants.Properties.Error] = Errors.ConsentRequired,
                         [OpenIddictServerAspNetCoreConstants.Properties.ErrorDescription] =
-                            S["The logged in user is not allowed to access this client application."]
+                            "The logged in user is not allowed to access this client application."
```
agriffard, line 100: No more `S[]` ?

kevinchalet: It's deliberate. See https://github.com/OrchardCMS/OrchardCore/pull/7675/files#diff-44ca9283b8f821a762e5d0c72ad06292720058093eeb604fb872af0aee4f77a3R26-R27 for the explanation.

kevinchalet:
I ended up tweaking the `.yml` files to install both the .NET 5.0 and .NET Core 3.1 SDKs. Technically, we don't need the 3.1 SDK, but we need the 3.1 runtime to run the tests, and installing the entire SDK is the likely simplest option.

If OC used Arcade, it would take care of downloading the runtimes automatically. That's what we use in OpenIddict: https://github.com/openiddict/openiddict-core/blob/dev/global.json#L2-L11

sebastienros (approved):

PR #7695, opened 2020-11-20
Handle Trumbowyg empty options

### LINKED ISSUE #7450: Trumbowyg settings issues
I found two issues with the trumbowyg editor.

First, if you setup a content type with the trumbowyg editor and don't specify the options (via the ContentDefinitionManager or via a recipe), the page throws a javascript error because the output js is invalid.

```
...
var settings = ;
...
```

Second, I don't think it's a good idea to output javascript directly in the browser like [this](https://github.com/OrchardCMS/OrchardCore/blob/d1d8eade9540251744ea3a5ab17f25758db3f475/src/OrchardCore.Modules/OrchardCore.ContentFields/Views/HtmlField-Trumbowyg.Edit.cshtml#L56). I am pretty sure this is a security issue?

I suggest we pass the settings via a `data-` attribute.  

/cc @agriffard

agriffard:
Related to #7450

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Views/HtmlField-Trumbowyg.Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Views/HtmlField-Trumbowyg.Edit.cshtml
@@ -55,14 +55,18 @@ else
     $(function () {
         @if (!string.IsNullOrEmpty(trumbowygSettings.Options))
         {
+            <text>
             var settings = @Html.Raw(trumbowygSettings.Options);
             settings['lang'] = '@culture.TwoLetterISOLanguageName';
+            </text>
         }
         else
         {   
+            <text>
             var settings = { 
                 lang: '@culture.TwoLetterISOLanguageName'
             }
+            </text>
         }
 
         $('#@Html.IdFor(m => m.Html)').trumbowyg(settings).on('tbwchange', function () {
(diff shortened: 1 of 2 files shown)


jptissot:
This fixes part of the issue but we are still outputting JS directly from the admin.

src/OrchardCore.Modules/OrchardCore.ContentFields/Views/HtmlField-Trumbowyg.Edit.cshtml
```diff
@@ -53,8 +53,21 @@ else
     }
 }    
     $(function () {
-        var settings = @Html.Raw(trumbowygSettings.Options);
-        settings['lang'] = '@culture.TwoLetterISOLanguageName';
+        @if (!string.IsNullOrEmpty(trumbowygSettings.Options))
+        {
+            <text>
+            var settings = @Html.Raw(trumbowygSettings.Options);
```
Skrypt, line 59: @agriffard 
Render the Options on the textarea directly in a data-options attribute like this : 

```
<textarea asp-for="Html" class="form-control" data-options="@Html.Raw(trumbowygSettings.Options);"></textarea>
```

And then here get the options value by doing : 

```
var settings = $('#@Html.IdFor(m => m.Html)').data("options");
```

jptissot: I tried doing this last time. encoding issues.

Skrypt: Json.Parse(settings); ???

Issue #7508, opened 2020-11-04
Custom Recipe Parameters


nuno-tiago-reis:
Hello, I was reading the pull request #4922 and I have a question. I'm trying to use this feature to change the value of my OpenId Connection authority according to the environment I'm on. Does this allow me to retrieve configurations from the main appsettings.json file or only the one that is generated by Orchard and is placed in the App_Data folder?

Another question is that whenever I use the configuration helper I get the following error:

````
JavaScriptException: configuration is not defined
````

The way I'm using it in the recipe is as follows:

````
"variables":
{
	"authority": "[js: configuration('IdentityManager:Host')]"
}
````

deanmarcussen:
it works for all the `appsettings.json` files.

But it only works for single key, i.e. `"authority": "[js: configuration('Host')]"

It would be a nice pr to improve it to allow access to section keys.

nuno-tiago-reis:
Okay but then the error itself is strange, shouldn't it say that there is not matching key?

deanmarcussen:
@nuno-tiago-reis I realised why you're getting the `configuration is not defined` error

Currently it will only work inside the recipe steps section, not the variables section.

I will tweak that, along with the key accessor limitations I already mentioned.

Issue #4889, opened 2019-11-26
Feature Request: Custom Recipe Parameters support


DariuS231:
Would it be possible to have somehow support for Custom parameter values in the recipe? 
Something like:

in the appsettings.json

``` JSON
{
  "RecipeName": "Blank",
  "Secret": "...",
  "ConnectionString": "...",
  "DatabaseProvider": "Sqlite",
  "TablePrefix": "Test",
  "CustomParamAppID": "86eb5541-ba2b-4255-9344-54eb73cec375",
  "CustomParamTenantID": "4cc363b6-5254-4b8c-bc1b-e951a5fc85ac"
}
```
in the recipe

``` JSON
{
  "name": "Blank",
  "displayName": "Blank site",
  "description": "Creates a blank site with content management features.",
  "author": "The Orchard Team",
  "website": "https://orchardproject.net",
  "version": "2.0",
  "issetuprecipe": true,
  "categories": [ "default" ],
  "tags": [ "blank" ],

  "variables": {
  },
  "steps": [
    ... ,
     {
	"name": "azureADSettings",
	"appId": "[js: parameters('CustomParamAppID')]",
	"tenantId": "[js: parameters('CustomParamTenantID')]",
	"displayName": "Orchard Core AD App",	
	"callbackPath": "/signin-oidc"
      }
   ]
}
```
That way we can have features fully configured after tenant setup.

MichaelPetrinolis:
you could create a custom module that executes the recipe passing a custom environment.
check IRecipeExecutor.ExecuteAsync

DariuS231:
@MichaelPetrinolis agree, but I think there could be a lot of value having this feature natively.

sebastienros:
There might be an open issue for that, we have suggested it already.
In a way that we could even plug it with some kind of Forms module, or at least with your own controller. Right now the simplest solution would be to just merge the two json documents, one with the recipe, and one with the variables, then execute this merged json document. Should be very easy.

What I had looked into was to actually describe the expected variables in the recipe itself, such that we could validate them before running the recipe. Or someone to even generate a form based on specific fields.

But the recipe service already accepts custom parameters AFAIR, and this is probably how we plugged the Setup form with the selected recipe.

Now, each variable can have a custom prefix, like `js:` to evaluate the variable as JavaScript. You could create a custom provider for `settings:[name]` that would take the value from the settings.

hishamco:
> There might be an open issue for that,

[...]

PR #7704, opened 2020-11-20
Add AdminOptionsMethod for recipes.

### LINKED ISSUE #7702: Recipes need to account for AdminUrlPrefix when creating Admin Menus
The blog theme recipe creates a few admin menu links that don't work if the settings have specified a custom AdminUrlPrefix. This is because "Admin" is hardcoded in the URL. Here is a snippet from the [blog link](https://github.com/OrchardCMS/OrchardCore/blob/df011d387465e352ff86a2da79c678648fb9c766/src/OrchardCore.Themes/TheBlogTheme/Recipes/blog.recipe.json#L1314).

`"LinkUrl": "[js: 'Admin/Contents/ContentItems/' +  variables('blogContentItemId') + '/Display']",`

It's not a huge deal to change this in the backend for one instance, but if I specify a custom admin prefix for all tenants at the root settings, all admin menu links created by recipes in all tenants will fail.

If recipes don't have access to the admin prefix, I believe they need it. And, if they do have access to the admin prefix, it needs to be accounted for in the recipes bundled with OC.

Skrypt:
Fixes #7702

/cc @davidhayden

### DIFF
--- a/src/OrchardCore.Themes/TheBlogTheme/Recipes/blog.recipe.json
+++ b/src/OrchardCore.Themes/TheBlogTheme/Recipes/blog.recipe.json
@@ -1311,7 +1311,7 @@
             {
               "$type": "OrchardCore.AdminMenu.AdminNodes.LinkAdminNode, OrchardCore.AdminMenu",
               "LinkText": "Blog",
-              "LinkUrl": "[js: 'Admin/Contents/ContentItems/' +  variables('blogContentItemId') + '/Display']",
+              "LinkUrl": "[js: adminurlprefix() + '/Contents/ContentItems/' +  variables('blogContentItemId') + '/Display']",
               "IconClass": "fas fa-rss",
               "UniqueId": "7b293d57056a4eebb3713f07f12c65d8",
               "Enabled": true,
@@ -1332,7 +1332,7 @@
             {
               "$type": "OrchardCore.AdminMenu.AdminNodes.LinkAdminNode, OrchardCore.AdminMenu",
               "LinkText": "Main Menu",
-              "LinkUrl": "[js: 'Admin/Contents/ContentItems/' +  variables('menuContentItemId') + '/Edit']",
+              "LinkUrl": "[js: adminurlprefix() + '/Contents/ContentItems/' +  variables('menuContentItemId') + '/Edit']",
               "IconClass": "fas fa-sitemap",
               "UniqueId": "5118cecfde834dacb26ac08980f1b5a7",
               "Enabled": true,
@@ -1369,7 +1369,7 @@
                 {
                   "$type": "OrchardCore.AdminMenu.AdminNodes.LinkAdminNode, OrchardCore.AdminMenu",
                   "LinkText": "Content Items",
-                  "LinkUrl": "Admin/Contents/ContentItems/",
+                  "LinkUrl": "[js: adminurlprefix() + '/Contents/ContentItems/']",
                   "IconClass": null,
                   "UniqueId": "7b293d57056a4eebb3713f07f12c65d9",
                   "Enabled": true,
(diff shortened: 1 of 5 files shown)


src/OrchardCore.Themes/TheBlogTheme/Recipes/blog.recipe.json
```diff
@@ -1311,7 +1311,7 @@
             {
               "$type": "OrchardCore.AdminMenu.AdminNodes.LinkAdminNode, OrchardCore.AdminMenu",
               "LinkText": "Blog",
-              "LinkUrl": "[js: 'Admin/Contents/ContentItems/' +  variables('blogContentItemId') + '/Display']",
+              "LinkUrl": "[js: adminurlprefix() + '/Contents/ContentItems/' +  variables('blogContentItemId') + '/Display']",
```
jtkech, line 1314: @Skrypt

We already have a `ConfigurationMethodProvider` so as i remember maybe we can already use something like this

    "[js:configuration('OrchardCore_Admin').AdminUrlPrefix]"

Skrypt: We can't assume that someone will have provided a AdminUrlPrefix in a appsettings.json file. By exposing the AdminOptions class we then get the AdminUrlPrefix that defaults to `/admin`

Skrypt: Else we would need to be able to use a ?? operator here.

Skrypt: `[js: configuration('OrchardCore_Admin').AdminUrlPrefix ?? '/admin']`

Skrypt: I'm getting an error when doing that

ParserException: Line 1: Unexpected token ?
Esprima.JavaScriptParser.ThrowUnexpectedToken<T>(Token token, string message)

jtkech: In the related issue i said

> Hmm, maybe we could change the `ConfigurationMethodProvider` to do a bind on an object instance, so that, if there is no config, we get the default value that is `Admin`. But not sure we can do that here as we would need a typed object

Maybe a null check in javascript but not so good

Skrypt: Esprima.NET support only ECMAScript 2016 and the null coalescing operator is 2020.

Skrypt:
This needs refactoring based on @sebastienros comments https://github.com/OrchardCMS/OrchardCore/issues/4889#issuecomment-558760938

src/OrchardCore.Themes/TheBlogTheme/Recipes/blog.recipe.json
deanmarcussen: Yes, I think @jtkech is right we should use the `ConfigurationMethodProvider`.

But it has a few limitations, that I've known about for a while, but are easily fixed.

`"[js:configuration('OrchardCore_Admin').AdminUrlPrefix]"` doesn't work, because the object returned isn't accessible. it only works when you return a single string value...
i.e.
```
"[js:configuration('Foo')]"
```
where 'Foo' is
```
{
  "Foo": "/MyAdmin" 
}
```

I mentioned it recently here, https://github.com/OrchardCMS/OrchardCore/issues/7508 but time...

it's an easy fix (non breaking on how it works already), so I will change it to work better, i.e.
`"[js:configuration('OrchardCore_Admin:AdminUrlPrefix, 'Default')]"`

Skrypt: Don't waste time on this. We need to remove this code. It's not secure.

PR #7705, opened 2020-11-21
Added feature to slugify media folders and files to make them SEO-friendly

netwavebe:
This is a follow up on https://github.com/OrchardCMS/OrchardCore/pull/7662.

The idea is to have a feature that would slugify folders and files to make them SEO-friendly.

Here's how it works (first without the feature enabled):

mediaslugify

I made seperate methods for folders and files, as I can imagine someone might implement a custom `IMediaNameNormalizerService` to change filenames (e.g. add copyright notice) but not folders.

I was looking for a place to add this to the documentation as well, but it doesn't seem to fit anywhere. https://docs.orchardcore.net/en/dev/docs/reference/modules/Media/ is all about code, not the media library.

I'd like some feedback on this PR.

/cc @deanmarcussen @Skrypt

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
@@ -22,6 +22,7 @@ namespace OrchardCore.Media.Controllers
         
         private readonly HashSet<string> _allowedFileExtensions;
         private readonly IMediaFileStore _mediaFileStore;
+        private readonly IMediaNameNormalizerService _mediaNameNormalizerService;
         private readonly IAuthorizationService _authorizationService;
         private readonly IContentTypeProvider _contentTypeProvider;
         private readonly ILogger _logger;
@@ -29,6 +30,7 @@ namespace OrchardCore.Media.Controllers
 
         public AdminController(
             IMediaFileStore mediaFileStore,
+            IMediaNameNormalizerService mediaNameNormalizerService,
             IAuthorizationService authorizationService,
             IContentTypeProvider contentTypeProvider,
             IOptions<MediaOptions> options,
@@ -37,6 +39,7 @@ namespace OrchardCore.Media.Controllers
             )
         {
             _mediaFileStore = mediaFileStore;
+            _mediaNameNormalizerService = mediaNameNormalizerService;
             _authorizationService = authorizationService;
             _contentTypeProvider = contentTypeProvider;
             _allowedFileExtensions = options.Value.AllowedFileExtensions;
@@ -179,10 +182,12 @@ namespace OrchardCore.Media.Controllers
                     continue;
                 }
 
+                var fileName = _mediaNameNormalizerService.NormalizeFileName(file.FileName);
+
                 Stream stream = null;
                 try
                 {
-                    var mediaFilePath = _mediaFileStore.Combine(path, file.FileName);
+                    var mediaFilePath = _mediaFileStore.Combine(path, fileName);
                     stream = file.OpenReadStream();
                     mediaFilePath = await _mediaFileStore.CreateFileFromStreamAsync(mediaFilePath, stream);
 
@@ -196,7 +201,7 @@ namespace OrchardCore.Media.Controllers
 
                     result.Add(new
                     {
-                        name = file.FileName,
+                        name = fileName,
                         size = file.Length,
                         folder = path,
                         error = ex.Message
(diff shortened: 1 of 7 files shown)


Skrypt:
All good. Only thing I can see is if we should have this service in the controller or in a lower lever like the store.

Skrypt:
Ok nevermind I changed my mind it's fine as it is.

src/OrchardCore.Modules/OrchardCore.Media/Services/DefaultMediaNameNormalizerService.cs
```diff
@@ -0,0 +1,9 @@
+namespace OrchardCore.Media.Services
+{
+    public class DefaultMediaNameNormalizerService : IMediaNameNormalizerService
```
deanmarcussen, line 3:
```suggestion
    public class NullMediaNameNormalizerService : IMediaNameNormalizerService
```

netwavebe:
That does describe it better, @deanmarcussen.

src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
```diff
@@ -179,10 +182,12 @@ public async Task<ActionResult<object>> GetMediaItem(string path)
                     continue;
                 }
 
+                var fileName = _mediaNameNormalizerService.NormalizeFileName(file.FileName);
```
deanmarcussen, line 185: How does this handle duplicate file names?

Now we are altering the file name, so `Apple` becomes `apple` as does ` duplicate file names will be more common, and users more likely to get into a scenario where they can't upload a file without renaming it because it has been normalized?

I think all we would get is an exception, with no information

netwavebe: It handles duplicate file names after normalization in the same way as without the feature enabled.

I uploaded a file `Apple 123.jpg`. This is what happens when I try to upload another file named `Apple-123.jpg`:

image

Looks good to me?

deanmarcussen: All good for developers, (they will understand), but how will end users understand that their file name changed?

Odd looking apple ;)

We handle this with autoroute by looking for the existing and adding `-1` to the end of the file name, until it is unique. So no error, due to normalizing.

Maybe it is not an issue, but I think of users, who have a lot less understanding of these things than developers.

netwavebe: Depends on the situation I guess. If I upload the exact same file, I don't want to add it again but rather have the system reject it. In other cases, it might be better to add `-1` to the slug. Problem is: how to know if it's the exact same file after normalizing? By comparing file content?

The same thing is true without slugify enabled. If I upload 2 different files with the same name `team.jpg`, should the second be rejected or renamed `team-1.jpg`?

deanmarcussen: Yes, it's tricky, that's why I wanted to think about it first, but it's merged so...

At least there is a clear error message - I was unsure that would be the case.

Perhaps it's just better handled as a note in the docs about the feature.

Maybe @agriffard will suggest a good location in the docs, for this feature?

agriffard: You can add a new folder/README.md in docs/reference/modules for your modules.
In the TOC, we could start making a Media menu with sub-menus (Media, Azure, Slugify).

Skrypt: As far as I remember uploading a file with the same name twice has always returned that error message for us.

Skrypt: @deanmarcussen I confirm that uploading a file with the same name will just return an error with or without the feature enabled.

Without :   

orchard

With :    

orchard

And, I think that when someone tries to upload a file with the same name a modal should open to give the option to overwrite the current file, rename with a number suffix or cancel the operation. Else, when we want to replace a media file we need to delete it then reupload it. 

Also, doing this caused cached media file issues which I'm not sure 100% are fixed. So, right now we don't allow to overwrite a file simply.

Skrypt: Bébé Yoda

deanmarcussen: > I think that when someone tries to upload a file with the same name a modal should open to give the option

That would be the nice option, and nice improvement :)

Issue #7603, opened 2020-11-12
How to handle the owner id with recipemigrations


deanmarcussen:
Pulled over from the discussion here https://github.com/OrchardCMS/OrchardCore/discussions/7575#discussioncomment-127219

There's a couple of scenarios we may need to take into account with the `UserId` / `Owner`

During a setup recipe, the `AdminUserId` is available as a js property.

However people are also trying to use `RecipeMigrator` to setup content this is fine, but it potentially leaves the content without an owner.

In this case there is no context object provided, so the  `"Owner": "[js: parameters('AdminUserId')]"` is not available.

Also migrations are run during site startup so there is no `HttpContext` from which to get the current user.

However we do have the `ISite.SuperUser` property available, which could be used to pass to the context, always providing at least an `[js: parameters('AdminUserId')]` which would make sense for content being imported during a migration.

The other scenario is with deployment plans, and how the owner id an be maintained when deploying through sites.

This does also open up a conversation about migrating users between sites - which maybe better covered when I get that secrets/encryption module finished off.

Skrypt:
Question is also if we should have a "system" user account for migrating things like Workflows which can also end up most of the time not having any "Owner" assigned to. Globally, a content item should never end up having a "null" Owner else it becomes an orphan and it's more complex to reassign those to a specific Owner from UI afterward. Of course there will always be exceptions so we do not enforce the database Owner to never be null but at least we could assign them a special user account so that we can evaluate permissions on them.

sebastienros:
We need to use SuperUser  in this case

agriffard:
I'd like to remind a feature that would be really useful:
Allow to declare parameters from a the setup (like ARM templates) and use them in the recipe.

See also #3966 to use the culture and #495 for a Setup Wizard.

deanmarcussen:
@jtkech might have to reach out for some ideas on this.

It is trivial to retrieve the SuperUser in the `IRecipeMigrator` but it is more complicated to do so during a setup process, as it is the values for the Site set until after the recipes have run.

Ideally this information would be useful to provide to the `IRecipeMigrator` during both setup and normal ongoing migrations, as from users reports, they are using data migrations to create content items, which will then run during setup.

I had thought it might be possible to use the `ShellScope.Items` for this, and set a value in the setup service, that could later be caught in the `IRecipeMigrator`.

[...]

Issue #3966, opened 2019-07-24
Add setup culture in available recipe parameters


agriffard:
Retrieve the culture specified in the setup (ex: by using this in the querystring ?culture=fr-FR or by using the appsettings default culture) and make it available in the recipe parameters.

It will allow to use it in a recipe step that adds and sets the default culture in the Localization settings dynamically.

hishamco:
> and make it available in the recipe parameters

I didn't get what you mean here

> It will allow to use it in a recipe step that adds and sets the default culture in the Localization settings dynamically.

AFAIK the default culture and supported cultures are added to the recipes. Is this what you mean or there's something else?

sebastienros:
@hishamco there is a set of special variables in the recipes that are passed by the called, so you can use these values in the recipe. When executed from the setup controller, we pass all the form values to the recipes. I assume @agriffard means that we should also pass the "current" culture that was used, so the recipe could use it to configure the default culture of the site. It could have been a js method, but maybe easier to provide it as a parameter.

Skrypt:
To add the current configured cultures in a recipe they should be added normally to the SiteSettings.Properties. Right now, if I test to export these cultures by exporting the SiteSettings.Properties it doesn't export them. I've not looked at the code yet but I'm guessing that we need to handle exporting these IEntity.Properties from the OrchardCore.Settings SettingDeploymentSource.cs.

Also, the same issue must happen in SettingsStep.cs.

deanmarcussen:
This issue is different to exporting localization / culture settings.

I think what @agriffard wants to do is pass the selected query string value of the setup culture chosen, through to a recipe as a parameter. When the `Properties` dictionary is added from this pr https://github.com/OrchardCMS/OrchardCore/pull/7885 that value can easily be added, and made available to a recipe.

Remind me @agriffard when it is merged and I will show you how.

PR #7706, opened 2020-11-21
Improve configuration method provider

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/7508

Note: This is a scoped method provider only for recipes, so doesn't give global access to configuration.

Improvements to it.
- available in the `variables` section now as well.
- can access keys, as the existing was only returning strings.
- has a default value option (not required - so is non breaking)

/cc @Skrypt @jtkech 

Note: I haven't updated the blog recipe, there may still be a value in having a method provider specific. 

This is mostly just fixes to remove some limitations, and correct the docs, as they were misleading.

### DIFF
--- a/src/OrchardCore/OrchardCore.Recipes.Core/ConfigurationMethodProvider.cs
+++ b/src/OrchardCore/OrchardCore.Recipes.Core/ConfigurationMethodProvider.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using Microsoft.Extensions.Configuration;
 using OrchardCore.Environment.Shell.Configuration;
 using OrchardCore.Scripting;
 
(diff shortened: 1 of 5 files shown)


agriffard:
Does it fix #7603?

agriffard:
I'd like to remind a feature that would be really useful:
Allow to declare parameters from a the setup (like ARM templates) and use them in the recipe.

See also #3966 to use the culture and #495 for a Setup Wizard.

deanmarcussen:
> Does it fix #7603?

Not quite yet, I would wait to see if @Skrypt agrees, and @jtkech also agrees with the change to the accessor (as I don't actually know why the existing one doesn't work, whether it's Jint related, or due to `IConfiguration` objects.)

Then just the blog recipe could be updated to use the feature, by default.

> I'd like to remind a feature that would be really useful:

Would be awesome. Let's try and get back to secrets first. then laters... also related to https://github.com/OrchardCMS/OrchardCore/issues/396

jtkech:
@agriffard @deanmarcussen 

The default value of the admin prefix is `Admin` not `/Admin`, i know that we trim the slashes before using it to build route patterns (that should not start with a leading slash), but for admin menu links, an url starting with a slash is considered as absolute and then used as is, in place of taking into account the pathBase that may include a tenant prefix.

**So i think that the examples in the doc of the admin prefix value (or default) should be `Admin` with no leading slash**

Skrypt:
"LinkUrl": "[js: 'Admin/Contents/ContentItems/' +  variables('blogContentItemId') + '/Display']",

PR #7707, opened 2020-11-21
Add UseCdn Picker to Setup Page 

hyzx86:
#5969


Skrypt:
We have a discussion next week about this. Your PR doesn't cover all scenarios. For example, here it will maybe work in the setup page. But it should be set before getting to the setup page. And even if the setup works afterward if you go to /admin the CDN issue will still remain.

hyzx86:
This problem has been going on for a year, 
let's fix it first.
There are 1.4 billion people who don't have access to these CDN resources. A large number of new entrants have struggled with this problem. If there are any other problems, let's submit a new PR to solve them.

Thanks!

Skrypt:
Duplicate PR see : https://github.com/OrchardCMS/OrchardCore/pull/7599

sebastienros:
Closing since there is another PR with the same goal. And my preference is to just have local assets by default.

hyzx86:
Hi @sebastienros  , Has the problem not been solved yet?
I found that the latest version still requires manual code to avoid loading resources from the CDN.

Issue #7598, opened 2020-11-12
Custom query schema parsing not working


tropcicstefan:
I think method BuildSchemaBasedFieldType inside LuceneQueryFieldTypeProvider.cs and SqlQueryFieldTypeProvider.cs is wrong because [official json ](https://json-schema.org/learn/miscellaneous-examples.html)schema should look something like this&nbsp;
 ```
{
    "type": "object",   
    "properties": {  
        "firstName" : {
            "type" : "string"
        },
        "age": {
            "type": "integer"
        }
     } 
} 
 ```
and line where code tries to get type specifically "child["type"]" will throw error of "System.InvalidOperationException: 'Cannot access child value on Newtonsoft.Json.Linq.JProperty.'"
```
foreach (var child in properties.Children())
    {
        var name = ((JProperty)child).Name;
        var nameLower = name.Replace('.', '_');
        var type = child["type"].ToString();
...
```
fix should be something along the lines of
```
var properties = querySchema["properties"];

foreach (JProperty child in properties.Children())
{
    var name = child.Name;
    var nameLower = name.Replace('.', '_');
    var type = child.Value["type"].ToString();
```
(and properties key should probably be lowercase)

sebastienros:
Honestly I don't think anyone has ever tried it. I was even for removing this, or at least moving the field to somewhere less visible. If you think you can fix it please provide a PR, and if you can contribute a little bit of documentation that would help. Thanks.

PR #7708, opened 2020-11-21
Custom query schema fix

tropcicstefan:
added documentation about custom schema in queries, fixed getting type property of individual fields, added support for description

PR for [7598](https://github.com/OrchardCMS/OrchardCore/issues/7598)

I wanted it to have same behavior for lucene and sql so I added null checks inside build methods in sql provider. Looking back I think it should be changed to throw logged errors when properties (of a schema) is nonexistent or there is no such content type.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Lucene/GraphQL/LuceneQueryFieldTypeProvider.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lucene/GraphQL/LuceneQueryFieldTypeProvider.cs
@@ -54,20 +54,22 @@ namespace OrchardCore.Queries.Lucene.GraphQL.Queries
                         continue;
                     }
                     var type = querySchema["type"].ToString();
+                    FieldType fieldType;
 
                     if (query.ReturnContentItems &&
                         type.StartsWith("ContentItem/", StringComparison.OrdinalIgnoreCase))
                     {
                         var contentType = type.Remove(0, 12);
-                        var queryField = BuildContentTypeFieldType(schema, contentType, query);
-                        if (queryField != null)
-                        {
-                            schema.Query.AddField(queryField);
-                        }
+                        fieldType = BuildContentTypeFieldType(schema, contentType, query);
                     }
                     else
                     {
-                        schema.Query.AddField(BuildSchemaBasedFieldType(query, querySchema));
+                        fieldType = BuildSchemaBasedFieldType(query, querySchema);
+                    }
+
+                    if (fieldType != null)
+                    {
+                        schema.Query.AddField(fieldType);
                     }
                 }
                 catch (Exception e)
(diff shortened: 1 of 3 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=7708) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=7708) before we can accept your contribution.<br/><br/>:x: tropcicstefan [sign now](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=7708)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OrchardCMS/OrchardCore?pullRequest=7708) it.</sub>

src/OrchardCore.Modules/OrchardCore.Lucene/GraphQL/LuceneQueryFieldTypeProvider.cs
```diff
@@ -77,38 +79,45 @@ public async Task BuildAsync(ISchema schema)
...
             {
-                var name = ((JProperty)child).Name;
+                var name = child.Name;
                 var nameLower = name.Replace('.', '_');
-                var type = child["type"].ToString();
+                var type = child.Value["type"].ToString();
+                var description = child.Value["description"]?.ToString();
 
-                if (type == "String")
+                if (type == "string")
```
deanmarcussen, line 102: Why do you change the capitalization here from `String` to `string` ?

Would it be better just to do a `String.Equals("string", type, StringComparison.OrdinalIgnoreCase)` case insensitive comparison? 

To be non breaking on existing queries?

tropcicstefan: I changed it so it's according to [official schema](https://json-schema.org/learn/miscellaneous-examples.html). And regarding to 'non breaking change' this commit is actually to fix bug because I wasn't even able to run this feature, i.e. it didn't work at all

deanmarcussen: ok makes sense.

I updated the docs a little, so mkdocs will flow them properly.

Good job fixing this, I feel like it's been 'fixed' two or three times in the last year or two.

I agree about the logging it would be a nice improvement.

Do you want to do it on this pr, or prefer to add another time?

Skrypt:
@deanmarcussen I let you review and merge this one. I've not looked at details neither tested it but if it works then we should maybe merge it. That's better than something broken 😉

Skrypt (approved):

src/OrchardCore.Modules/OrchardCore.Lucene/GraphQL/LuceneQueryFieldTypeProvider.cs
tropcicstefan: Actually not sure what to do. Maybe it's more efficient to remove null checking on properties and typetype variable. And let `properties.Children()` and `schema.Query.Fields.OfType<ContentItemsFieldType>().First(x => x.Name == contentType)` throw errors (and let buildAsync to catch it) because those cases should happen rarely and there is no reason to have checking for null that will happen almost never.

And secondly what're the rules around here about unit testing are you avoiding them or there just isn't enough time.

agriffard:
Close and reopen

Skrypt:
@tropcicstefan How does that work with a custom database table?
I've tested for example creating a Test table in SQLite and then creating a SQL Query that returns a single column of that table. Used a Schema on the Query and I can see it is added to the GraphiQL explorer but it seems to consider it as a Content Item.

image

image

Skrypt:
Actually I tried this a second time and it works.

tropcicstefan:
I think you have content type of name 'test'

Skrypt:
Exactly. This is why that Query tries to mix up with my custom table Query.
But after adding a Test2 Query it doesn't mix up.

PR #7730, opened 2020-11-24
CI cleanup and e2e runner tweaks

jptissot:
- Reinstate paths-ignore for the Pull Request workflow
- e2e tests now use the App_Data_Test folder
- e2e tests now create the SaaS tenant like before
- Replaced the `clean` function with the `deleteDirectory` function
- Removed the `setup` function
- Add ability to specify the App_Data location for the host function

/cc @sebastienros


.github/workflows/release_ci.yml
```diff
@@ -64,14 +64,10 @@ jobs:
         password: ${{ secrets.DOCKERHUB_PASSWORD }}
     - name: Deploy preview docker image for linux
       if: matrix.os == 'ubuntu-latest'
-      # Buildkit Only working in linux
-      env:
-        DOCKER_BUILDKIT: 1
```
Skrypt, line 69: This should normally stay there unless you want also to remove the Dockerfile-CI.dockerignore file.

jptissot: I was not sure if it was being used as we were copying the build files a couple lines below.

Skrypt: Dockerfile-CI doesn't need to use the DOCKER_BUILDKIT env var. This env var says to Docker itself to enable it's Docker Buildkit feature which allows to use alternate dockerignore files.

PR #7734, opened 2020-11-24
Sitemap Custom Path

PiemP:
Resolve #7689 

added `CustomUrlSitemapSource.cs` in the lib `OrchardCore.Sitemap.Abstractions`.

added driver, interfaces, view model, views, builder and handler to allow the management of the custom url, in the sitemap (module `OrchardCore.Sitemaps`).


src/OrchardCore.Modules/OrchardCore.Sitemaps/ViewModels/CustomUrlSitemapSourceViewModel.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace OrchardCore.Sitemaps.ViewModels
+{
+    /// <summary>
+    /// A sitemap source for managing custom url.
+    /// </summary>
+    public class CustomUrlSitemapSourceViewModel
+    {
+        /// <summary>
+        /// the custom url.
```
agriffard, line 13: If you want to add a summary to each property, please start with `Gets or sets ...`

PiemP:
I'm pretty new to the sitemap in general.
I have a couple of question (and probably stupid) about it:

 - have sense to add URL that point to another site? If no, It's better to use `CustomPathSitemap...` instead of `CustomUrlSitemap...` in the name of the classes and views?
 - why the sitemap URL creation use `siteSettings.BaseUrl` to get the host? (I have used the `UrlHelperExtensions` in the `CustomUrlsSitemapSourceBuilder` to get host and protocol)
 - If we are in SaaS scenario, when we create a tenant, If we add more than one host in the **Host name** parameter we have to create an url for each host in the sitemap? 

Thank you

deanmarcussen:
@PiemP worth having a read of the sitemaps protocol here https://www.sitemaps.org/protocol.html

- it is possible to point at another site, but you have to prove ownership, so it is more complicated.
- I chose to prefer `BaseUrl` over url helper to enforce an active choice with the user as to what url would be produced. The reasoning behind this is how google considers urls different, ie http is different to https, and there can be a risk of a sitemap being produced with a url that the site owner doesn't want google to be aware of, i.e. azurewebsites.net/mysite instead of the real domain. When they leak into google it can take a long time to get rid of them.
- not sure what mean in the third question. From a search engine point of view you would be preferring to promote one url to the search engine, I would have thought?

Maybe then CustomPath would make more sense, than CustomUrl.

Still thinking, need to find time to look at what you have made here :)

src/OrchardCore.Modules/OrchardCore.Sitemaps/Handlers/CustomPathSitemapSourceUpdateHandler.cs
```diff
@@ -0,0 +1,56 @@
...
+        private readonly ISitemapManager _sitemapManager;
+
+        public CustomPathSitemapSourceUpdateHandler(ISitemapManager sitemapManager)
+        {
+            _sitemapManager = sitemapManager;
+        }
+
+        public async Task UpdateSitemapAsync(SitemapUpdateContext context)
+        {
+            var contentItem = context.UpdateObject as ContentItem;
```
deanmarcussen, line 21: This won't work because there will never be a `ContentItem` that is related to this sitemap, because it is a custom path.

The function of this method is to clear the sitemape cache if a content item changes. (it's slightly not working see: https://github.com/OrchardCMS/OrchardCore/issues/8087 )

But with a custom path I don't think there will be a way to detect that the custom path has changed. So probably best to just remove this update handler

src/OrchardCore.Modules/OrchardCore.Sitemaps/ViewModels/CustomPathSitemapSourceViewModel.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// <summary>
+    /// A sitemap source for managing custom url.
+    /// </summary>
+    public class CustomPathSitemapSourceViewModel
+    {
+        /// <summary>
+        /// Gets and sets the custom url.
+        /// </summary>
+        [Required]
+        [RegularExpression(@"^~?\/[^?#]+(\?([^#]*))?(#(.*))?")]
```
deanmarcussen, line 16: I don't think we should be using RegEx here.

Better to use the same as the `AutorouteDriver` rules.

src/OrchardCore.Modules/OrchardCore.Sitemaps/Views/CustomPathSitemapSource.Edit.cshtml
```diff
@@ -0,0 +1,53 @@
...
+                    <label asp-for="Url">@T["Url"] </label>
+                    <div class="input-group">
+                        @if (!String.IsNullOrWhiteSpace(Site.BaseUrl))
+                        {
+                            <div class="input-group-prepend">
+                                <div class="input-group-text">@Site.BaseUrl</div>
+                            </div>
+                        }
+                        <input asp-for="Url" class="form-control" autofocus />
+                        <span asp-validation-for="Url" class="text-danger">@T["The Url is required or have used a wrong format"]</span>
```
deanmarcussen, line 24: To make it easy on the translators try and get the same text as is used to validate the sitemap.path

src/OrchardCore.Modules/OrchardCore.Sitemaps/Views/CustomPathSitemapSource.Edit.cshtml
```diff
@@ -0,0 +1,53 @@
...
+                    </div>
+                    <span class="hint">@T["The URL to add to the sitemap. ex. /path_to_something or ~/path/to/something"]</span>
```
deanmarcussen, line 26: Same as autoroute. describe path without leading `/`. so path = `blog/post`

src/OrchardCore.Modules/OrchardCore.Sitemaps/Builders/CustomPathSitemapSourceBuilder.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            return Task.FromResult(false);
+        }
+
+        private bool BuildUrl(SitemapBuilderContext context, CustomPathSitemapSource source, XElement url)
+        {
+            if (string.IsNullOrEmpty(source.Url))
+                return false;
+
+            if(!(context.HostPrefix?? "").EndsWith("/") && !source.Url.StartsWith("/") && !source.Url.StartsWith("~"))
```
deanmarcussen, line 42: If you assume `blog/post` no need for extra checks here. Just prepend `~/` virtual path to all I suspect

src/OrchardCore.Modules/OrchardCore.Sitemaps/Drivers/CustomPathSitemapSourceDriver.cs
```diff
@@ -0,0 +1,57 @@
...
+                View("CustomPathSitemapSource_SummaryAdmin", sitemapSource).Location("SummaryAdmin", "Content"),
+                View("CustomPathSitemapSource_Thumbnail", sitemapSource).Location("Thumbnail", "Content")
+            );
+        }
+
+        public override IDisplayResult Edit(CustomPathSitemapSource sitemapSource, IUpdateModel updater)
+        {
+            
+
+            return Initialize<CustomPathSitemapSourceViewModel>("CustomPathSitemapSource_Edit", model =>
```
deanmarcussen, line 28: Formatting (extra lines... everywhere)

src/OrchardCore/OrchardCore.Sitemaps.Abstractions/Models/CustomPathSitemapSource.cs
```diff
@@ -0,0 +1,31 @@
+using System;
+
+namespace OrchardCore.Sitemaps.Models
+{
+    /// <summary>
+    /// A sitemap source for managing custom url.
+    /// </summary>
+    public class CustomPathSitemapSource : SitemapSource
+    {
+        public static char[] InvalidCharactersForPath = ":?#[]@!$&'()*+,.;=<>\\|%".ToCharArray();
```
deanmarcussen, line 10: Can you add `{` and `}` to this. They're also invalid characters in route templates, which we haven't got round to adding to autoroute yet

src/OrchardCore.Modules/OrchardCore.Sitemaps/Drivers/CustomPathSitemapSourceDriver.cs
```diff
@@ -0,0 +1,68 @@
...
+                    m => m.Priority,
+                    m => m.ChangeFrequency
+                ))
+            {
+                sitemap.Url = model.Url;
+                sitemap.Priority = model.Priority;
+                sitemap.ChangeFrequency = model.ChangeFrequency;     
+                sitemap.LastUpdate = DateTime.Now;
+
+                context.Updater.ModelState.BindValidationResults(Prefix, sitemap.ValidateUrlFieldValue(S));
```
deanmarcussen, line 62: This doesn't need to be an extension method, it can just be done in the driver. We just did it in autoroute, as there is shared validation between the handlers. Simple if it is kept together

deanmarcussen (commented):
Thanks for updating. Couple more comments.
You can take it out of draft, it's nearly ready.

src/OrchardCore.Modules/OrchardCore.Sitemaps/Views/CustomPathSitemapSource.Edit.cshtml
```diff
@@ -0,0 +1,48 @@
...
+                        }
+                        <input asp-for="Url" class="form-control" autofocus />
+                    </div>
+                    <span asp-validation-for="Url"></span>
+                    <span class="hint">@T["The URL to add to the sitemap. ex. path/to/something"]</span>
+                </div>
+                <div class="form-group" asp-validation-class-for="ChangeFrequency">
+                    <label asp-for="ChangeFrequency">@T["Change Frequency"] </label>
+                    <select asp-for="ChangeFrequency" class="form-control w-25" asp-items="Html.GetEnumSelectList<ChangeFrequency>()"></select>
+                    <span class="hint">@T["The default frequency with which these content types are likely to change."]</span>
```
deanmarcussen, line 26:
```suggestion
                    <span class="hint">@T["The default frequency with which this path is likely to change."]</span>
```

src/OrchardCore.Modules/OrchardCore.Sitemaps/Views/CustomPathSitemapSource.Edit.cshtml
```diff
@@ -0,0 +1,48 @@
...
+                        <option value="3">@T["0.3"]</option>
+                        <option value="4">@T["0.4"]</option>
+                        <option value="5">@T["0.5"]</option>
+                        <option value="6">@T["0.6"]</option>
+                        <option value="7">@T["0.7"]</option>
+                        <option value="8">@T["0.8"]</option>
+                        <option value="9">@T["0.9"]</option>
+                        <option value="10">@T["1.0"]</option>
+                    </select>
+                    <span class="hint">@T["The default priority of these content types relative to other content on your site."]</span>
```
deanmarcussen, line 43:
```suggestion
                    <span class="hint">@T["The default priority of this path relative to other content on your site."]</span>
```

src/OrchardCore.Modules/OrchardCore.Sitemaps/Views/CustomPathSitemapSource.Thumbnail.cshtml
```diff
@@ -0,0 +1,4 @@
+@model ShapeViewModel<CustomPathSitemapSource>
+
+<h4 class="card-title">@T["Sitemap Custom Path"]</h4>
+<p>@T["Add sitemap entries for a site path."]</p>
```
deanmarcussen, line 4:
```suggestion
<p>@T["Add a sitemap entry for a path."]</p>
```

src/OrchardCore/OrchardCore.Sitemaps.Abstractions/Models/CustomPathSitemapSource.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// </summary>
+    public class CustomPathSitemapSource : SitemapSource
+    {
+        public static char[] InvalidCharactersForPath = ":?#[]@!$&'()*+,.;=<>\\|%{}".ToCharArray();
+        public const int MaxPathLength = 1024;
+
+        /// <summary>
+        /// Gets and sets the custom url.
+        /// </summary>
+        public string Url { get; set; }
```
deanmarcussen, line 16:
```suggestion
        public string Path { get; set; }
```

src/OrchardCore/OrchardCore.Sitemaps.Abstractions/Models/CustomPathSitemapSource.cs
```diff
@@ -0,0 +1,31 @@
+using System;
+
+namespace OrchardCore.Sitemaps.Models
+{
+    /// <summary>
+    /// A sitemap source for managing custom url.
```
deanmarcussen, line 6:
```suggestion
    /// A sitemap source for managing a custom path.
```

src/OrchardCore.Modules/OrchardCore.Sitemaps/Startup.cs
```diff
@@ -80,6 +81,13 @@ public override void ConfigureServices(IServiceCollection services)
             services.AddContentPart<SitemapPart>()
                 .UseDisplayDriver<SitemapPartDisplayDriver>()
                 .AddHandler<SitemapPartHandler>();
+
+            //custom url in sitemap
```
deanmarcussen, line 85:
```suggestion
            // Custom sitemap path.
```

For infos comments are //spaceCapitalLetterFullstop

src/OrchardCore.Modules/OrchardCore.Sitemaps/Builders/CustomPathSitemapSourceBuilder.cs
```diff
@@ -0,0 +1,78 @@
...
+            url.Add(changeFreq);
+
+            var priority = new XElement(Namespace + "priority");
+            priority.Add(priorityValue);
+            url.Add(priority);
+        }
+
+        private void PopulateLastMod(CustomPathSitemapSource source, XElement url)
+        {
+            // Last modified is not required. Do not include if content item has no modified date.
```
deanmarcussen, line 69:
```suggestion
            // Last modified is not required. Do not include if the path has no modified date.
```

src/OrchardCore.Modules/OrchardCore.Sitemaps/Builders/CustomPathSitemapSourceBuilder.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            return Task.FromResult(false);
+        }
+
+        private bool BuildUrl(SitemapBuilderContext context, CustomPathSitemapSource source, XElement url)
+        {
+            if (string.IsNullOrEmpty(source.Url))
+                return false;
+
+            //add ~/ to the path, because the user insert it without leading /
```
deanmarcussen, line 42:
```suggestion
            // Add ~/ to the path, because the it is inserted without leading /.
```

deanmarcussen (commented):
Found lots of references in this still to content types, so suggesting to go through a refer to it as the Path, not a content type.

PiemP:
Thank you @deanmarcussen for your reviews and sorry for my stupid errors

deanmarcussen (approved):

src/OrchardCore.Modules/OrchardCore.Sitemaps/Views/CustomPathSitemapSource.Edit.cshtml
```diff
@@ -0,0 +1,48 @@
+@model CustomPathSitemapSourceViewModel
```
Skrypt, line 1: This require @using OrchardCore.Sitemaps.Builders namespace or add the using to the _ViewImports.cshtml file of that module. Else when building it won't find the reference to these ViewModels.

PiemP: @Skrypt I believe this @using is not necessary. This class is 'CustomPathSitemapSourceViewModel' and it is under **ViewModel** folder, not under **Builders** folder.

Skrypt: Well maybe I did not put my comment on the right line. But definitely you need to add the `@using OrchardCore.Sitemaps.Builders` in the _ViewImport.cshtml file else it won't publish your website. So this breaks "publishing" of Orchard Core.

src/OrchardCore.Modules/OrchardCore.Sitemaps/Views/CustomPathSitemapSource.SummaryAdmin.cshtml
```diff
@@ -0,0 +1,5 @@
+@model ShapeViewModel<CustomPathSitemapSource>
```
Skrypt, line 1: Here is the issue reported previously

src/OrchardCore.Modules/OrchardCore.Sitemaps/Views/CustomPathSitemapSource.Edit.cshtml
Skrypt: Actually I think you are right. This is maybe a bad build/publish on my side.

PR #7735, opened 2020-11-24
Queries documentation - Small fix to make description match code sample

NogginBox:
The description talks about a query called LatestBlogPosts, but the code sample was running a query called AllContent. These should match and I thought LatestBlogPosts was a nicer sounding query, so I kept that one.

Update: Now using actual query from blog recipe called `RecentBlogPosts`.


agriffard:
Blog recipe: `RecentBlogPosts`

PR #7745, opened 2020-11-25
Github Actions : use Chrome instead of Electron with Cypress

### LINKED ISSUE #7744: Github Actions : Use Chrome for functional tests
Capture d’écran de 2020-11-24 21-28-06

As you can see here in this screenshot we are using Electron to make our functional tests. We should probably use Chrome to avoid specific exceptions and to cover a wider range of browsers.

/cc @jptissot

Skrypt:
Fixes #7744

### DIFF
--- a/test/Functional/cypress-commands/src/test-runner.js
+++ b/test/Functional/cypress-commands/src/test-runner.js
@@ -57,7 +57,7 @@ export function e2e(dir, assembly, { appDataLocation='./App_Data', dotnetVersion
   deleteDirectory(path.join(dir, "App_Data_Tests"));
   var server = host(dir, assembly, { appDataLocation, dotnetVersion });
 
-  let test = child_process.exec("npx cypress run");
+  let test = child_process.exec("npx cypress run --browser chrome");
   test.stdout.on("data", data => {
     console.log(data);
   });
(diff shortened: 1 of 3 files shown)


test/Functional/cypress-commands/src/test-runner.js
```diff
@@ -57,7 +57,7 @@ export function e2e(dir, assembly, { appDataLocation='./App_Data', dotnetVersion
   deleteDirectory(path.join(dir, "App_Data_Tests"));
   var server = host(dir, assembly, { appDataLocation, dotnetVersion });
 
-  let test = child_process.exec("npx cypress run");
+  let test = child_process.exec("npx cypress run --browser chrome");
```
jptissot, line 60: Can you make the --browser flag a parameter of the method defaulting to chrome ?

jptissot (changes requested):

test/Functional/cypress-commands/src/test-runner.js
Skrypt: By default it always uses Electron.

https://docs.cypress.io/guides/guides/launching-browsers.html#Electron-Browser

PR #7747, opened 2020-11-25
Hide connection string when already set for tenant

### LINKED ISSUE #7691: Database connection strings should be not be displayed in the admin
When a connection string is defined in an environment variable, the value is displayed in the Tenants creation page. 

```
$env:OrchardCore__ConnectionString="server=localhost;uid=root;database=db;password=my-secret-pw;"
$env:OrchardCore__DatabaseProvider="MySql"
```
image

Connection strings should not be displayed if coming from an environment variable.

jptissot:
fixes #7691

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Tenants/ViewModels/EditTenantViewModel.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Tenants/ViewModels/EditTenantViewModel.cs
@@ -23,8 +23,7 @@ namespace OrchardCore.Tenants.ViewModels
 
         public IEnumerable<RecipeDescriptor> Recipes { get; set; }
 
-        public bool CanSetDatabasePresets { get; set; }
-        public bool ConnectionStringFromConfiguration { get; set; }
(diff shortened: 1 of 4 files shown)


src/OrchardCore.Modules/OrchardCore.Tenants/ViewModels/EditTenantViewModel.cs
```diff
@@ -24,5 +24,7 @@ public class EditTenantViewModel
         public IEnumerable<RecipeDescriptor> Recipes { get; set; }
 
         public bool CanSetDatabasePresets { get; set; }
+        public bool ConnectionStringFromConfiguration { get; set; }
+        public bool DatabaseProviderFromConfiguration { get; set; }
```
Skrypt, line 28: I think we should assume that both should be passed at the same time. We should probably make sure of that and use only one boolean here then.

jptissot: Ok done :)

Skrypt (approved):

PR #7749, opened 2020-11-25
Brackets no longer removed in recipe executor

### LINKED ISSUE #7623: Bracket can get removed from field when running a recipe
If you fill a TextField starting with `[` and ending with `]` and export it. Add it to a recipe in your module. When the field is imported, the starting `[` and last `]` are removed.

Before export:
image

After import:
image

jptissot:
fix #7623


hishamco:
I think this could be simplified, I will review this

jptissot:
Probably haha it was late

hishamco (commented):
I create a new PR #7776, your thought

Issue #6966, opened 2020-08-26
ContentItemIndex should have a content type index


deanmarcussen:
There is no index on the `ContentItemIndex` for the `ContentType` property - commonly used for searching on.

Probably it needs to have an index for
`ContentType`
`ContentType, Published, Latest`

Possibly just the second one is enough.

Skrypt:
If you mean having a DB Index then yes.

deanmarcussen:
Yes a DB Index (naming is hard )

Ok, so I reviewed execution plans of some common queries, i.e. `Query<Document, Index>()`

None of the indexes we currently have are being used because none of them include the `DocumentId`

Here's an execution plan for opening the content items list

<img width="914" alt="Screenshot 2020-08-27 at 14 00 22" src="https://user-images.githubusercontent.com/13782679/91446311-35cf1880-e86f-11ea-9777-cf995af5744a.png">

Note the Clustered Scan Index on the `ContentItemIndex` table (meaning it's seeking on the primary key, and then hitting the table)

Here it is after adding an Index `DocumentId`, ContentType`, `Published`, `ModifiedUtc`

<img width="892" alt="Screenshot 2020-08-27 at 14 00 49" src="https://user-images.githubusercontent.com/13782679/91446361-4b444280-e86f-11ea-9715-f038cd4fc965.png">

Now seeking against an Index Scan

My point is that none of our indexes include the `DocumentId` so for most queries that join on the document table they're not being used.

A query direct against the index, i.e. `QueryIndex<ContentItemIndex` with a `Where` would use them.

So needs a bigger review.

carlwoodhouse:
Yeah all indexes should really include the document id: it speeds up publishing perf loads (yes sql deletes get super slow on large indexes without it)

I did suggest that yessql should just do it a while back ;) 

https://github.com/sebastienros/yessql/issues/218

deanmarcussen:
Have you been adding custom indexes @carlwoodhouse ?
 
It would be helpful to know what you've tended to add?

I haven't had the time to properly eval all that I feel we should tweak here, but my general take is that we should (where possible, due to key length issues on index tables like the `Autoroute`) have indexes like

- DocumentId (as you say for the update queries)
- DocumentId, ContentItemId, Published
- DocumentId, ContentItemId, Latest
- DocumentId, ContentType, Published, ModifiedUtc (common query)
- DocumentId, ContentType, Published, Latest, ModifiedUtc (another commonish query)

[...]

PR #7757, opened 2020-11-25
LocalizationEntries Atomicity and Size

jtkech:
Does the same as #7643 but for `LocalizationEntries`


jtkech:
@jptissot 

As this PR is related to content localization, can you give it a try if you have time, to be sure i didn't break anything ;)

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Sitemaps/LocalizedContentItemsQueryProvider.cs
```diff
@@ -70,7 +70,7 @@ public async Task GetContentItemsAsync(ContentTypesSitemapSource source, Content
                     var queryResults = await _session.Query<ContentItem>()
                          .With<ContentItemIndex>(ci => ci.ContentType == source.LimitedContentType.ContentTypeName && ci.Published)
                          .OrderBy(ci => ci.CreatedUtc)
-                         .With<LocalizedContentItemIndex>()
+                         .With<LocalizedContentItemIndex>(i => i.Published || i.Latest)
```
deanmarcussen, line 73: Think it should only be Published here, draft items shouldn't in the sitemap.
Although they should already be excluded by the `ContentItemIndex` part of the query.

So an extra predicate here might not be necessary at all?

src/OrchardCore/OrchardCore.ContentLocalization.Abstractions/Records/Migrations.cs
```diff
@@ -45,5 +52,19 @@ public int UpdateFrom1()
...
+
+        // This code can be removed in a later version.
+        public int UpdateFrom2()
+        {
+            SchemaBuilder.AlterTable(nameof(LocalizedContentItemIndex), table => table
+                .AddColumn<bool>("Latest", c => c.WithDefault(false))
+            );
+
+            SchemaBuilder.AlterTable(nameof(LocalizedContentItemIndex), table => table
+                .CreateIndex("IDX_LocalizationPartIndex_State", "Published", "Latest")
```
deanmarcussen, line 64: Because the query is against the index the internal index needs all properties to be used
so it needs an index for
`Published`, `Latest`, `ContentItemId` 

and another index for
`Published`, Latest`, `LocalizationSet`

Otherwise SQL Server will not use the index, but fall back to a clustered scan (so it is useless). many of our indexes have this issue... slowly trying to fix them.

For infos

If it was a `Query<ContentItem, LocalizationPartIndex>(x => predicate)` it should also include the `DocumentId` as the first property on the index, as that gets used for the join part of the sql predicate.

They also have to be in order of how they are being queried.

src/OrchardCore.Modules/OrchardCore.Lists/Handlers/ContainedPartLocalizationHandler.cs
```diff
@@ -27,8 +27,8 @@ public override async Task LocalizingAsync(LocalizationContentContext context, L
             // todo: remove this check and change the handler to targe t ContainedPart when issue 3890 is fixed
             if (containedPart != null)
             {
-                var list = await _session.QueryIndex<LocalizedContentItemIndex>(x => x.ContentItemId == containedPart.ListContentItemId).FirstOrDefaultAsync();
-                var localizedList = await _session.QueryIndex<LocalizedContentItemIndex>(x => x.LocalizationSet == list.LocalizationSet && x.Culture == context.Culture).FirstOrDefaultAsync();
+                var list = await _session.QueryIndex<LocalizedContentItemIndex>(i => (i.Published || i.Latest) && i.ContentItemId == containedPart.ListContentItemId).FirstOrDefaultAsync();
+                var localizedList = await _session.QueryIndex<LocalizedContentItemIndex>(i => (i.Published || i.Latest) && i.LocalizationSet == list.LocalizationSet && i.Culture == context.Culture).FirstOrDefaultAsync();
```
deanmarcussen, line 31: Don't know enough about this handler to know whether it should be including `Latest` as well?

src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
```diff
@@ -93,7 +90,11 @@ public override async Task PublishedAsync(PublishContentContext context, Autorou
 
             if (!String.IsNullOrWhiteSpace(part.Path) && !part.Disabled && part.SetHomepage)
             {
-                await SetHomeRoute(part, homeRoute => homeRoute[_options.ContentItemIdKey] = context.ContentItem.ContentItemId);
+                await SetHomeRoute(part, homeRoute =>
```
deanmarcussen, line 93: maybe this is on the other pr?

jtkech: Yes, this is the exact same change, maybe i needed it here to test some use cases

src/OrchardCore.Modules/OrchardCore.Lists/Handlers/ContainedPartLocalizationHandler.cs
jtkech: Before, even there was no `Latest` column (only the `Published` column), all latest items was indexed

                    // Inthe index provider
                    if (!contentItem.Latest)
                    {
                        return null;
                    }
                    // All latest are indexed (and that have a loc part and non null Set and Culture)

So, before the query was returning all the latest (published or not) that was matching the predicate

But, because now when we removed an item, and only when we removed an item, new records are added with published = latest = false, so we need to filter these new records to return the same ones as before

As done for autoroutes

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Sitemaps/LocalizedContentItemsQueryProvider.cs
jtkech: See my next comment for why i added this globaly

But yes, here seems useless as above there is `ci.Published` on the `ContentItemIndex`

Good catch, i will remove it

src/OrchardCore/OrchardCore.ContentLocalization.Abstractions/Records/Migrations.cs
jtkech: Okay, just did the same as for autoroute migrations when the new `Latest` column was added

            SchemaBuilder.AlterIndexTable<AutoroutePartIndex>(table => table
                .CreateIndex("IDX_AutoroutePartIndex_State", "Published", "Latest")
            );

Hmm, but there were already these indexes


            SchemaBuilder.AlterIndexTable<LocalizedContentItemIndex>(table => table
                .CreateIndex("IDX_LocalizationPartIndex_LocalizationSet_Culture", new[] { "LocalizationSet", "Culture" })
            );

            SchemaBuilder.AlterIndexTable<LocalizedContentItemIndex>(table => table
                .CreateIndex("IDX_LocalizationPartIndex_ContentItemId", "ContentItemId")
            );

            // In a migration step
            SchemaBuilder.AlterIndexTable<LocalizedContentItemIndex>(table => table
                .AddColumn<bool>(nameof(LocalizedContentItemIndex.Published)));

So, do we need to drop some indexes before adding new ones ?

**Maybe simpler that you update the PR with what you are suggesting**

deanmarcussen: No problem, I'll update it.

Yes the existing indexes are not the best, I've been avoiding Dropping indexes, because it's been tricky with the one time I did that on Autoroute (doesn't work on SQL Server 2014 and below). So have been mostly, just improving them on new setups, and trying to get the new ones we're adding to be good.

jtkech: Okay cool then, thanks

Just need to fix unit tests before leaving

jtkech: Okay it was just a build error on the new `AlterIndexTable`

Feel free to update the PR

jtkech:
@deanmarcussen 

About adding the `Published, Latest, ContentItemId` and `Published, Latest, LocalizationSet` indexes as you suggested, just to be sure i fully understood

- After a quick look on the indexed queries we are using (i may have miss some)

    Why not `Published, Latest, Culture`?

    And, for the query only using `Published`, do we still need another index but only on `Published` (as we have)

- For Autoroutes there is an index on `JsonPath` but there is no query using it

    Maybe we would have need an index on `Path` instead

    And now i assume an index on `Published, Latest, Path`

But are you sure? If we have an index on A1, A2, ..., An and we do a query using A1, A2, ..., An+1, i thought that the index would be still used to first get a reduced set of records, this before checking An+1.

On the opposite, if we have an index on A1, A2, ..., An+1 and we do a query only using A1, A2, ..., An, does the index would be used? Or do we also need an index only on A1, A2, ..., An?

deanmarcussen:
@jtkech I've just been looking back at my notes here https://github.com/OrchardCMS/OrchardCore/issues/6966 and it seems mostly right, but a comment @carlwoodhouse made about the need for the `DocumentId` in general for performance on YesSql deletes was something I had forgotten about.

It's been a while since I was a dba, and key indexes are tricky to get completely right, because you have to guess at the most common queries being used for them ;)

But they can always be added manually if the configuration / query behaviour is different for the user, than what we have guessed, and/or are using internally. Azure SQL is quite good at giving recommendations for improvements.
SQL Profiler is the best tool we can use to determine query execution path, and validate what we are doing is good.

For Autoroutes, the `Path` column is too long (1000 characters?) to put into a key index, but there are other improvements we need to make for it., re:document id.

For this pr.

I will have some time tomorrow, so happy to look at the migration and run them against sql profiler.

I'm still not completely sure I have all my assumptions correct ;)

jtkech:
@deanmarcussen 

Okay, I did some research, many things are said and sometimes contradictory, so i did some tests

As i saw, if we have an unclustered index on A1, A2, An, the only way to get is used is when all the required columns for the produced SELECT OR for a produced JOIN are included in A1, A2, An

Then, and only if the above condition is true, in the WHERE / OR / AND parts, an unclustered index may be used if it includes some of the colums that are checked, and / or a clustered one for the other columns

---

So, when we do a `QueryIndex(...)` as the produced select is on all columns, unclustered indexes are never used and seem to be useless, unless we define an index including all columns

Then, when we do a query on `Document` with an inner join on `LocalizedContentItemIndex`, so on `DocumentId`, an unclustered index may be used if it inclues `DocumentId`

So here in the migrations, i let the indexes that are already defined, i just added the new `Latest` column, and the following index including `DocumentId`, starting with the most selective column to the less selective ones

    .CreateIndex("IDX_LocalizationPartIndex_DocumentId_Set_Culture_State", "DocumentId", "LocalizationSet", "Culture", "Published", "Latest")

I didn't add `ContentItemId` as it is only used with a `QueryIndex()`, never with a `Query()` and an inner JOIN

Here the result where we can see that only the unclustered index is used for the localization index table

Unclustered Index

Maybe better to not include in this index `Published` and `Latest` so that it doesn't include to much columns, and becasue they are not so selective, in that case the new unclustered index is still used, but also the clustered one

But on another side, `Published` and `Latest` don't add too much data, so for now i kept them in this new index.

deanmarcussen:
@jtkech you bet me too it :)

But I reckon your research agrees with my research.

Column lengths are good for me. MySql has the lowest limit (see Alias for how I stretched it right at to the limit)

`Published` and `Latest` are only 1 bit each, so well worth keeping.

Agree we need to review the other indexes - not just Autoroute, but the other ones, so we have a good pattern.

It's been on my list, but haven't got there yet.

src/OrchardCore/OrchardCore.ContentLocalization.Abstractions/Records/Migrations.cs
```diff
@@ -25,17 +25,24 @@ public int Create()
                 .Column<string>("LocalizationSet", col => col.WithLength(26))
                 .Column<string>("Culture", col => col.WithLength(16))
                 .Column<string>("ContentItemId", c => c.WithLength(26))
+                .Column<bool>("Published")
+                .Column<bool>("Latest")
             );
 
             SchemaBuilder.AlterIndexTable<LocalizedContentItemIndex>(table => table
-                .CreateIndex("IDX_LocalizationPartIndex_LocalizationSet_Culture", new[] { "LocalizationSet", "Culture" })
+                .CreateIndex("IDX_LocalizationPartIndex_LocalizationSet_Culture", "LocalizationSet", "Culture")
```
deanmarcussen, line 33: JT should we just drop this one (from the code, not from existing sites). It looks useless now?

jtkech: Okay

deanmarcussen:
@jtkech I added the `ContentItemId` back in.

It gets used in one the the handlers `ContainedPartLocalizationHandler` and feels like it makes sense to have it.

Size of index we're still quite good, cause all of these are 26 char's, or bits, and I checked in on MySql, and it's all working good.

Merge if you are ok with my change :)

jtkech:
@deanmarcussen

Okay to include `ContentItemId` as the index size is kept low, but just because the index will contain all columns ;)

I didn't include it as it is only used with a `QueryIndex()`, but if we include it the index will contain all columns, and then, as said (see below), it will be also used with a `QueryIndex()`. So, that's the best we can do ;)

> So, when we do a QueryIndex(...) as the produced select is on all columns, unclustered indexes are never used and seem to be useless, **unless we define an index including all columns**

About selectivity i would have put it just after `DocumentId`, as a compromise i used the order of the columns, i also used a shorter index name. One rule could be to include `DocumentId`, then all columns but excluding those of too large size.

    .CreateIndex("IDX_LocalizationPartIndex_DocumentId", "DocumentId", "LocalizationSet", "Culture", "ContentItemId", "Published", "Latest")

deanmarcussen:
> Update: One rule could be to include DocumentId, then all columns but excluding those of too large size

Or maybe better to just include the DocumentID as a rule, in one index, and then make specific indexes as we are using them? 

This PR LGTM if you are ready with it @jtkech 

(tried myself y/day and entries seem to be working correctly when updated)

PR #7776, opened 2020-11-26
Don't remove brackets if it's not a script expression

### LINKED ISSUE #7623: Bracket can get removed from field when running a recipe

hishamco:
Fixes #7623 and replaced #7749:

- Allow Shortcodes to be used inside the recipes

- Ability to not run invalid script expression (The one who has  invalid prefix, e.g. `[1js: xxx]`)


src/OrchardCore/OrchardCore.Recipes.Core/Services/RecipeExecutor.cs
```diff
@@ -197,21 +198,29 @@ private void EvaluateJsonTree(IScriptingManager scriptingManager, RecipeExecutio
...
                         EvaluateJsonTree(scriptingManager, context, property.Value);
                     }
-                    break;
 
+                    break;
                 case JTokenType.String:
-
+                    const char scriptSeparator = ':';
                     var value = node.Value<string>();
+                    var scriptSeparatorIndex = value.IndexOf(scriptSeparator);
```
jptissot, line 213: This should be inside the while loop

hishamco: It was there, seems I copied accidently, thanks

deanmarcussen:
Still doesn’t work if you include a : inside the shortcode content? Maybe

hishamco:
I will do a final test to handle this case

hishamco:
@deanmarcussen after the added unit test seems everything works as expected

deanmarcussen:
Try `[sc text='some : text'/]`
or some html content with json in it i.e.
```
<code>
[
	{
		color: "red",
		value: "#f00"
	}
]
</code>
```

I'm just curious is all...

hishamco:
@deanmarcussen these work as well ;)

hishamco:
Seems Linux complaining about `\r` :(

src/OrchardCore/OrchardCore.Recipes.Core/Services/RecipeExecutor.cs
```diff
@@ -197,21 +198,29 @@ private void EvaluateJsonTree(IScriptingManager scriptingManager, RecipeExecutio
...
                 case JTokenType.String:
-
+                    const char scriptSeparator = ':';
                     var value = node.Value<string>();
 
                     // Evaluate the expression while the result is another expression
                     while (value.StartsWith('[') && value.EndsWith(']'))
                     {
+                        var scriptSeparatorIndex = value.IndexOf(scriptSeparator);
+                        // Don't remove brackets if it isn't a script expression [js:xxx], so shortcodes for instance could be used here
```
deanmarcussen, line 218:
```suggestion
                        // Only remove brackets if this is a valid script expression, e.g. '[js:xxx]', or '[file:xxx]'
```

test/OrchardCore.Tests/Recipes/RecipeExecutorTests.cs
```diff
@@ -0,0 +1,104 @@
...
+        [InlineData("recipe5", "[sc text='some : text'/]")]
+        [InlineData("recipe6", @"<code>
+[
+	{
+		color: ""red"",
+		value: ""#f00""
+	}
+]
+</code>")]
+        public async Task RemoveBracketsIfTheScriptExpressionProcessed(string recipeName, string expected)
```
deanmarcussen, line 38:
```suggestion
        public async Task ShouldTrimValidScriptExpression(string recipeName, string expected)
```

test/OrchardCore.Tests/Recipes/RecipeExecutorTests.cs
```diff
@@ -0,0 +1,104 @@
...
+                    Settings = shellSettings
+                })));
+            var recipeEventHandlers = new List<IRecipeEventHandler> { new RecipeEventHandler() };
+            var loggerMock = new Mock<ILogger<RecipeExecutor>>();
+            var recipeExecutor = new RecipeExecutor(shellHostMock.Object, shellSettings, recipeEventHandlers, loggerMock.Object);
+            var executionId = Guid.NewGuid().ToString("n");
+            var recipeDescriptor = new RecipeDescriptor { RecipeFileInfo = GetRecipeFileInfo(recipeName) };
+
+            // Act
+            try
```
deanmarcussen, line 57: why would you try catch this?

deanmarcussen (commented):
Ok, I had time to look at the code.

Was concerned you were not taking into account other script engines, but it looks good.

Just a couple of suggestions on naming.

Just drop the last test that is failing, we don't need it.

test/OrchardCore.Tests/Recipes/RecipeExecutorTests.cs
hishamco: Regarding this I forgot to tell you guys that execute throw an exception for accessing a disposable object `IServiceProvider` but it works, if someone can fix this, will be better

test/OrchardCore.Tests/Recipes/RecipeExecutorTests.cs
```diff
@@ -0,0 +1,90 @@
...
+        [InlineData("recipe4", "[locale en]This text contain a colon ':' symbol[/locale][locale fr]Ce texte contient un symbole du côlon ':'[/locale]")]
+        [InlineData("recipe5", "[sc text='some : text'/]")]
+        public async Task ShouldTrimValidScriptExpression(string recipeName, string expected)
+        {
+            // Arrange
+            var shellHostMock = new Mock<IShellHost>();
+            var shellSettings = new ShellSettings { Name = "Test" };
+            var services = new ServiceCollection()
+                .AddScripting();
+            shellHostMock.Setup(h => h.GetScopeAsync(It.IsAny<ShellSettings>()))
```
deanmarcussen, line 37: i suspect it's how you're building the shellhost mock here. I'll have a look tomorrow

deanmarcussen: Yeah this isn't going to work as a mock, because the ShellScope is a static asynclock.

Have a look at how @jtkech rebuilt the unit tests for the `AutorouteEntriesTests` when that moved to needing a `ShellScope`

hishamco: I tried to make the code looks like

```csharp
var shellContext = CreateShellContext();
await shellContext.CreateScope().UsingAsync(async scope =>
{
    // Arrange
    var shellHostMock = new Mock<IShellHost>();
    shellHostMock.Setup(h => h.GetScopeAsync(It.IsAny<ShellSettings>())).Returns(Task.FromResult(scope));

    var recipeEventHandlers = new List<IRecipeEventHandler> { new RecipeEventHandler() };
    var loggerMock = new Mock<ILogger<RecipeExecutor>>();
    var recipeExecutor = new RecipeExecutor(shellHostMock.Object, shellContext.Settings, recipeEventHandlers, loggerMock.Object);
    var executionId = Guid.NewGuid().ToString("n");
    var recipeDescriptor = new RecipeDescriptor { RecipeFileInfo = GetRecipeFileInfo(recipeName) };

    // Act
    await recipeExecutor.ExecuteAsync(executionId, recipeDescriptor, new object(), CancellationToken.None);

    // Assert
    var recipeContext = (recipeEventHandlers.Single() as RecipeEventHandler).Context;
    var recipeStep = recipeContext.Step;
    Assert.Equal(expected, recipeStep.SelectToken("data.[0].TitlePart.Title").ToString());
});
```

but still the tests fail :( @jtkech any idea?

jtkech: @hishamco @deanmarcussen 

Okay, i will take a look on it this week end

hishamco:
Thanks @jtkech I planned to use DI to register the required services instead of mocking, but I find the mock could be easier instead of doing a lot of registration ;)

But I still I need to know why the mocking throws that exception, while the result is fine

jtkech:
@hishamco 

First, because `ShellScope` is an `AsyncLocal`, we need to use the `ShellScope.UsingAsync()` helper that init this async local state. Then, because the `shellHostMock` was always returning the same scope, each recipe was using the same `ShellScope`, in place of a child one, and was disposing it at the end, so the second recipe step was using an already disposed shell scope and then with a disposed service provider. One solution was to only use the  `content` step in the recipes.

But i could fix it by using in `shellHostMock.Setup().Returns()` a method in place of a value, so that it is dynamic.

hishamco:
Thanks so much @jtkech, your approve guys before we merge

jptissot (approved):

PR #7784, opened 2020-11-27
Update ServiceCollectionExtensions namespace

hishamco:
For fast discovery the extensions for `IServiceCollection` should be in `Microsoft.Extensions.DependencyInjection` namespace, I found there 're many namespaces need to be modified, @jtkech if sounds is good to rename I will continue to rename the other once

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentLocalization/ServiceCollectionExtensions.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentLocalization/ServiceCollectionExtensions.cs
@@ -1,5 +1,5 @@
-using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
+using OrchardCore.ContentLocalization;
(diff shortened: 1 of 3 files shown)


jtkech:
@hishamco 

The problem is that, even there are not hidden to provide more flexibility, we don't want to expose some of them so easily, as normally they are not intended to be used directly, but by the modules / core project that define them, or through more visible helpers.

hishamco:
> we don't want to expose some of them so easily, as normally they are not intended to be used directly, but by the modules / core project that define them, or through more visible helpers.

Make sense, but for those who can be accessed publicly they should be easily discoverable IMHO

jtkech:
But some that we don't want to expose so easily are still public to still provide advanced extensibilty

One example is `AddExtensionManagerHost()`

hishamco:
> But some that we don't want to expose so easily are still public to still provide advanced extensibilty

Agree, my aim to change the namespace for those are our intend to be used in users application

PR #7787, opened 2020-11-27
MediaField: Display

agriffard:
- [X] Display img in card if the path as an image file extension.
- [X] Do not display anything if there is no media.
- [X] Use media text if set.


Skrypt (commented):
Format the entire document in Visual Studio.

Skrypt (approved):

PR #7802, opened 2020-11-28
Order Shortcodes categories

### LINKED ISSUE #7801: Shortcodes categories: Order alphabetically
image

agriffard:
Fixes #7801


deanmarcussen (approved):
Should we do the same for the workflows modal, or does it do it already?

agriffard:
> Should we do the same for the workflows modal, or does it do it already?

Workflows categories are already order by name: https://github.com/OrchardCMS/OrchardCore/blob/fbe6dd5b69e9082d45dccb1d0b931a34b2823c45/src/OrchardCore.Modules/OrchardCore.Workflows/Services/ActivityLibrary.cs#L23

PR #7808, opened 2020-11-29
Rename Workflow Types into Workflows

agriffard:
Rename also Workflows into instances where needed (breadcrumbs, titles).


agriffard:
Change some buttons classes:

image

sfmskywalker (approved):

Issue #1610, opened 2018-03-14
Consider introducing ShellSettings.Id


kevinchalet:
We're currently using `ShellSettings.Name` as a stable tenant identifier for the Data Protection services (it's also heavily used in my last PR to determine token audiences).

Unfortunately, this is not an ideal candidate as the tenant name can be accidentally re-assigned to another tenant (after deleting the existing one). It has security implications if the tenant is managed by someone else, as data encrypted by the removed tenant could be decrypted by the new tenant (since they share the same name).

To fix that, I suggest introducing a new opaque `Id` property (typically... a GUID) we can safely use for isolation purposes.

/cc @sebastienros

sebastienros:
We actually have private keys in this files, or at least we have an issue to generate it. I can show you how it is in O1. I think we should use it, but the idea is exactly the same so I agree with the initiative.

sebastienros:
It's not about privacy of the value, it's that we already need a key for secrets anyhow, and it's stored in the same file, ~~and we already have a field for it~~. 

Well apparently we didn't create the field yet, so it's fine creating a new Id in the shell settings.
Around here: https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.Setup/Services/SetupService.cs#L116

Just add a property to `ShellSettings` and it should be stored automatically.

sebastienros:
As part of your PR?

PinpointTownes:
What should we use? A GUID? Or a shorter thing?

sebastienros:
you can use the `IIdGenerator` if it's available (shorter, readable guid). Otherwise guid.

PinpointTownes:
Nice. Looks like `OrchardCore.Entities` is indirectly referenced by the setup module, so that's fine.

sebastienros:
The workflow branch is moving it to a more common package. This won't be an issue either.

PinpointTownes:
I'll do that in a future PR, as this is a quite massive change that deserves a dedicated PR.

PinpointTownes:
Note: now that tenants have their own key ring, doing that is way less interesting (since the keys wouldn't be preserved after re-creating a tenant with the same name, the new tenant couldn't decrypt payloads encrypted by the former tenant)

sebastienros:
Close if you think that's not necessary anymore.

PinpointTownes:
[...]

PR #7812, opened 2020-11-30
Replace ShellSettings Identifier by VersionId

jtkech:
The ShellSettings Identifier is used for consistency checking and we may introduce a real ID property later, so as suggested by @kevinchalet https://github.com/OrchardCMS/OrchardCore/issues/1610#issuecomment-733021441, see below, here we use a `VersionId` property instead.

> @jtkech hum, then maybe Version, Timestamp or ConcurrencyCheck would have been better names? If we decide to introduce a real ID property later, having both Id and Identifier will be very confusing.

### DIFF
--- a/src/OrchardCore/OrchardCore.Abstractions/Shell/ShellSettings.cs
+++ b/src/OrchardCore/OrchardCore.Abstractions/Shell/ShellSettings.cs
@@ -38,10 +38,10 @@ namespace OrchardCore.Environment.Shell
 
         public string Name { get; set; }
 
-        public string Identifier
+        public string VersionId
         {
-            get => _settings["Identifier"];
-            set => _settings["Identifier"] = value;
+            get => _settings["VersionId"];
+            set => _settings["VersionId"] = value;
         }
 
         public string RequestUrlHost
--- a/src/OrchardCore/OrchardCore/Shell/ShellHost.cs
+++ b/src/OrchardCore/OrchardCore/Shell/ShellHost.cs
@@ -141,7 +141,7 @@ namespace OrchardCore.Environment.Shell
 
         public async Task UpdateShellSettingsAsync(ShellSettings settings)
         {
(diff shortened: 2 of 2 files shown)


Skrypt:
Not sure about that one @jtkech I think that `Identifier` defines exactly what it is. It's not a version but really an identifier. Just saying, but I don't know all the story.

jtkech:
@Skrypt 

I introduced it myself for consistency checking ;)

            settings = await _shellSettingsManager.LoadSettingsAsync(settings.Name);

            // Consistency: We may have been the last to add the shell but not with the last settings.
            if (settings.Identifier == currentIdentifier)
            {
                return;
            }

kevinchalet:
Sounds like a good compromise 👍

deanmarcussen (commented):
LGTM.

I assume for migrations the `VersionId` will just return null, not be equal to the new one, and get a new `VersionId` generated.

A real identifier for use in the tenant cookie would be great @jtkech 

So annoying when you're working off localhost with a few sites and keep having to login/logout because the cookie is conflicting - to the point I am often overriding the `CookieAuthenticationOptions` to provide a specific cookie name for the particular project.

jtkech:
@kevinchalet thanks ;)

@deanmarcussen 

> I assume for migrations the VersionId will just return null, not be equal to the new one, and get a new VersionId generated.

Yes, it will be generated only when the settings will be saved again, just before, but that's okay as we only check it when we read back the settings we just saved with therefore a new VersionId

> A real identifier for use in the tenant cookie would be great @jtkech

Okay, but in a separate PR as i may need more info and thinkings as a tenant can be pre-created by a config source.

> So annoying when you're working off localhost with a few sites and keep having to login/logout because the cookie is conflicting - to the point I am often overriding the CookieAuthenticationOptions to provide a specific cookie name for the particular project.

I agree ;)

deanmarcussen (approved):

PR #7824, opened 2020-12-01
Support complex json token as parameter name

### LINKED ISSUE #7823: Allow json token  as parameter name in recipe
Currenty in recipes you can only  use simple json parameter names  passed in as parameter to parameters function 

eg 

```
      "Text": "[js: parameters('FirstName')]"
```

I would like to use complex json names like 

```
    "Text": "[js: parameters('Properties.FirstName')]"
     "Text": "[js: parameters('Properties.LastName')]"
```

my recipe  I want to use looks like this  : 

```
"LandingPage": {
            "FullName": {
              "Text": "[js: parameters('Properties.FirstName')]"
            },
            "ProfessionalTitle": {
              "Text": "[js: parameters('Properties.ProfessionalTitle')]"
            }
          },

```

submitting a pr to support this.

giannik:
fixes #7823

### DIFF
--- a/src/OrchardCore/OrchardCore.Recipes.Core/ParametersMethodProvider.cs
+++ b/src/OrchardCore/OrchardCore.Recipes.Core/ParametersMethodProvider.cs
@@ -18,7 +18,7 @@ namespace OrchardCore.Recipes
                 Name = "parameters",
                 Method = serviceprovider => (Func<string, object>)(name =>
                {
-                   return environmentObject[name].Value<string>();
+                   return environmentObject.SelectToken(name)?.Value<string>();
                })
             };
         }


deanmarcussen:
Out of curiosity how are you populating these complex objects?

From a custom setup screen or...?

giannik:
Yeah dean, 
I have  a custom workflow task "CreateCustomTenant" in which when executing i run the following 
```
 foreach (var keyValuePair in setUpProperties)
                    {
                        if (!string.IsNullOrEmpty(keyValuePair.Key))
                        {
                            setupContext.Properties[keyValuePair.Key] = keyValuePair.Value;
                        }

                    }
```


In setupContext I have added the additional property to store these values : 

```

       /// Gets and additional key/value info.
        /// </summary>
        public IDictionary<string, string> Properties { get; } = new Dictionary<string, string>();
```


Then when my custom recipe runs it picks them up from Properties bag.


Its a generic bag to pass in additional properties when setting up a tenant.

Can submit a pr for this as well if you think its not too specific.

deanmarcussen (approved):
Interesting, want to open another issue with the properties explanation, and we can look at it?

giannik:
sure , will do over the weekend.

PR #7832, opened 2020-12-02
Fix MySql casing in functional tests

deanmarcussen:


deanmarcussen:
@jptissot So the issue was that the provider was case sensitive, so `Mysql` wasn't working, but the change to `MySql` gets it going.

Another issue, so maybe you have time to take over this pr, is that the `App_Data` folder is still being used, not the `App_Data_Tests` folder, so I had to tweak the CI to see why it was failing.

jptissot (approved):

PR #7850, opened 2020-12-04
jquery-ui dependency

giannik:


Skrypt:
I don't think this one is necessary. We don't use any styles from JQuery-ui.

PR #7854, opened 2020-12-04
Remove duplicate Scripts folder

jtkech:
Under `TheAgencyTheme/wwwroot/` there are 2 folders `Scripts` and `scripts`

Here we remove the unused one


jptissot (approved):

Skrypt:
Did you do a `gulp build` to see that it gets generated back from the gulpfile standing in that folder?

jtkech:
@Skrypt 

These 2 files also exist in the `scripts` folder that still exists

On my widows machine they can't be added back in a `Scripts` folder as a `scripts` still exists.

I checked that the `package.json` only refers to the `scripts` folder

These 2 files were added back through e741810

PR #7859, opened 2020-12-04
Update Autoroutes indexes

jtkech:
Related to the last comments of #7757 about `LocalizedContentItemIndex`, but here applied to `AutoroutePartIndex`

> So, when we do a QueryIndex(...) as the produced select is on all columns, unclustered indexes are never used and seem to be useless, unless we define an index including all columns

> Then, when we do a query on Document with an inner join on LocalizedContentItemIndex, so on DocumentId, an unclustered index may be used if it includes DocumentId

But `AutoroutePartIndex` is only queried through `QueryIndex()`, so here we can remove by code the existing indexes, so that there are not created on new sites, and without having to create a new one including `DocumentId`.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Autoroute/Migrations.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Autoroute/Migrations.cs
@@ -1,4 +1,3 @@
-using OrchardCore.Autoroute.Drivers;
 using OrchardCore.Autoroute.Models;
 using OrchardCore.ContentManagement.Metadata;
 using OrchardCore.ContentManagement.Metadata.Settings;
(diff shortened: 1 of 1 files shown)


deanmarcussen:
I think we still have one for `DocumentId` because underlying YesSql uses it every time a document is saved, and the indexes are deleted, and recreated.

(such a shame the `Path` is too long for the Index)

Here's the delete sql
`exec sp_executesql N'delete from [AutoroutePartIndex] where [DocumentId] = @Id',N'@Id int',@Id=106`

Without index

Screenshot 2020-12-06 at 16 59 23


With Index

Screenshot 2020-12-06 at 16 59 35


I reckon we should have a DocumentId index on all our indexes. It makes much more sense now why updating a document on a remote sql server can be a bit slow sometimes ;)

jtkech:
> I think we still have one for DocumentId because underlying YesSql uses it every time a document is saved, and the indexes are deleted, and recreated.

Yes you're right, but as i saw, this unclustered index including `DocumentId` is used when an index record is deleted, not when inserted with new values and then updated with the right `DocumentId`. But yes, still useful on updating as we also need to delete previous record(s).

So because when deleting a record we just have a where clause on `DocumentId`, i just added

    .CreateIndex("IDX_AutoroutePartIndex_DocumentId", "DocumentId")

> (such a shame the Path is too long for the Index)

:)

deanmarcussen:
LGTM @jtkech

PR #7861, opened 2020-12-04
Admin section in appsettings

agriffard:
Add Admin section in appsettings.json so that people understand that they can change it easily.

Plus, adding it at the bottom as an benefit: You can uncomment any other settings section without needing to remove the comma the end.


hishamco (approved):
LGTM

PR #7866, opened 2020-12-05
Update dependencies and postcss

agriffard:
Upgraded postcss to version 8.
https://www.npmjs.com/package/postcss

It was making an error on gulp rebuild:
`Uncaught TypeError: Cannot read property 'unprefixed' of undefined`
so I added `rtlcss` as suggested here:
https://github.com/vkalinichev/postcss-rtl/pull/73#issuecomment-737784695

@jptissot @hishamco Could you help to use `rtlcss` in gulp instead of `postcss-rtl`?


jptissot:
I did try a simple change. Wait until merging as I don't think it works properly.

jptissot:
I think you should wait to merge this. Or at least revert my commits. And when you change a dependency like this, you need to run `npm run rebuild` and not `npm run build`

agriffard:
@jptissot: I reverted your changes and installed and rebuilt as you suggested.

PR #7867, opened 2020-12-05
Admin: Setting to display titles in top bar

agriffard:
New setting to display titles in the top bar in Admin:
image

Default:
image

Enabled:
image

Collapsed menu:
image

Dark:
image

RTL:
image


src/OrchardCore.Modules/OrchardCore.OpenId/Views/Application/Create.cshtml
```diff
@@ -6,7 +6,7 @@
 @{
     var settings = ViewData[nameof(OpenIdServerSettings)] as OpenIdServerSettings;
 }
-<h1>@RenderTitleSegments(T["Create a new application"])</h1>
+<titleZone><h1>@RenderTitleSegments(T["Create a new application"])</h1></titleZone>
```
kevinchalet, line 9: FWIW, I've never been a huge fan of creating tag helpers for HTML nodes that are not standard.
Wouldn't `<h1 oc-zone="title">...</h1>` be a better option?

agriffard: The TitleZone tagHelper contains the logic to determine in which zone it will be displayed based on the setting.

kevinchalet: I understood that.

My question was more: can we have a tag helper that does the same thing using an attribute and doesn't require a `<titleZone>` node, which will look like a non-standard thing to people who don't know it's backed by a tag helper.

agriffard: Ok, I get your point.
I also had this concern in mind doing all the changes: 
If people forget to surround their titles with the taghelper or the addtaghelper in _ViewImports.cshtml, it will not be effective but without any errors, just a `<titlezone>` element around it that doesn't mean anything in Html.

This is my way to do it so that nothing prevents you from adding something else than a `<h1>`.

Still need to document it.

agriffard: Finally chose to use `<zone Name="Title"></zone>` because it wouldn't have worked on production (no reference to OrchardCore.Admin).
The place to render the Title zone is done in the Layout.cshtml based on the setting value.

sfmskywalker (approved):

hishamco:
Looks good, but make bold will be better from UI/UX perspective

agriffard:
> Looks good, but make bold will be better from UI/UX perspective

image

hishamco (approved):

Skrypt:
Missing one here : 

image

hishamco:
Related #7874

PR #7376, opened 2020-10-24
Webpack 5

jptissot:
/cc @agriffard 

I am testing to make sure GraphiQL still works.

PR #7872, opened 2020-12-05
Use Webpack 5 in GraphQl

agriffard:
Replaces #7376


MikeKry:
@agriffard 

Graphiql fails on error

image

I think it is related to this change, any clue how to fix this?

Issue #2246, opened 2018-08-18
Dashboard Templated Items


agriffard:
Idea: Allow to add dashboard items that appear in the /admin page (instead of the Welcome text).

If we allow to use Liquid syntax, people will be able to render many dynamic things.

Ex: A blog exists, then thanks to a recent blog posts query, list the last blog posts created. 
Add a quick link to the blog (instead of having an AdminMenu suggested here #2220)

sebastienros:
There might already be an issue for it, or at least it was mentioned during a meeting. A simple implementation would be to use any content item as the dashboard page. Like the home page works, but for the admin. This way you can create a flow page and use any widgets. And obviously modules could provide widgets that are dedicated for the dashboard. It might require a custom stereotype or category, but it's already doable very easily.

There might be better ways too, we'll need to brainstorm.

agriffard:
For example, it would be interesting to add some 'Get started' links panel in the 'welcome admin page' to easily redirect to: 

- Edit your content items
- Enable new features
- List content types
- Manage the settings
...

weedkiller:
Since it was merged here, I am adding what I listed before for reference, I strongly recommend clicking on the link to get a sample view.. [this dynamic dashboard in ASP ](http://bit.ly/2G4hAkM)
______
Hello, I would like to request a much needed feature to allow developers to build widgets and have a drag drop widget based dashboard. Something like or similar to [this dynamic dashboard in ASP ](http://bit.ly/2G4hAkM)

Its been a long wait for something like this.
It would be easy to add charts, graphs, stats, summaries, timelines etc.

Also as @agriffard mentioned, having widgets/component/partialviews/ViewComponents or Orchard Components, provide a quick and easy way to build data, link up the data to UI like, chart, graph, timeline, Rich text Editor, table etc... would make it easier for orchard developers.

Basically allows to build component quicker and easier let the dashboard expose settings, drag drop, role based access etc.


![image](https://user-images.githubusercontent.com/21086190/69381649-3fde3e00-0c83-11ea-9110-9f5778dbec2e.png)

![image](https://user-images.githubusercontent.com/21086190/69381768-79af4480-0c83-11ea-94cb-9f4cb0bb2079.png)

Thanks

papyr:
Hello there, I saw this was being tracked.

Team this is _not just for the Admin_, **we are requesting/talking about Data/Graphical Widgets Content Types for users, managers as well for Front End and logged in users.**

[...]

PR #7875, opened 2020-12-06
Admin dashboard

agriffard:
Use a jumbotron.

Rename title.

Load shapes starting with `AdminDashboardCard` from modules:

image


agriffard:
image

hishamco:
I'm still have a doubt using color variations in buttone, it would be nice to use one color, frankly there's no mean to use different color here

Overall looks good, hope it's mobile friendly which I think

hishamco:
Please make sure it's 100% fine with the dark theme

deanmarcussen:
Hmm, what will happen if any of these modules are disabled?

Hint: this is why we have drivers

hishamco:
Good point Dean, what if we follow certain convensions like what we did in `AdminMenu`, so every module could participate on dashboard. So whenever a module has a view named `Dashboard.chtml` it could be placed into a dashboard module, that we we could fix the issue that Dean mentioned above

agriffard:
Ok, keeping it minimal for the moment.

hishamco:
We could keep the cards as its, then we could ability to load them dynamically

agriffard:
Did an example in Users with the count retrieved from a Query.

We could also display the N latest modified items or some stats.

Will try to take the permissions into account before displaying a card or a link.

Skrypt:
Maybe remove the title and make it part of the title in the jumbotron

screencapture-localhost-5001-Admin-2020-12-06-17_56_12

hishamco:
The first jumbotron is good looking

agriffard:
Title removed.
Didn't change the padding or it should be done in a specific css for the admin jumbotron.

Kept the word site:

image

Dashboard is a bit to long on small screens:

image

Compared to:

image

src/OrchardCore.Modules/OrchardCore.Admin/Views/AdminDashboard.cshtml
```diff
@@ -1,4 +1,23 @@
-<div class="alert alert-success">
-    <h4>@T["Welcome to Orchard"]</h4>
-    <p>@T["Feel free to browse the menu and discover all its possibilities."]</p>
+@using OrchardCore.DisplayManagement;
+@inject OrchardCore.DisplayManagement.Descriptors.IShapeTableManager ShapeTableManager
+
+@{
+    var shapeTable = ShapeTableManager.GetShapeTable("TheAdmin");
```
deanmarcussen, line 5: I would suggest this is done with Drivers, and a 
`IDisplayManager<AdminDashboard>()` (making an identifier class called `AdminDashboard` in `Admin.Abstractions` to drive the display with)

Here this code will fail if `GetShapeTable("TheAdmin")` a different admin theme was used.

If you do it with drivers, then you get placement - the ability to hide them / move them around. All the cool things that modular display management does.

Skrypt (approved):

agriffard:
Demoed during the meeting.

Suggestions:
- Make it a feature?
- Allow to add and organize Widgets by drag and drop and store states like in a zone of a layer.

agriffard:
Initial issue: #2246

hishamco:
Can we merge this while it's not break anything, then you can continue on widgets and make it feature as Seb suggested in the standup

agriffard:
No, it was just a POC.

We made a design meeting and the implementation will be completely different (and more powerful hopefully):
https://github.com/OrchardCMS/OrchardCore/issues/2246#issuecomment-742687838

Closing this one.
Replaced by https://github.com/OrchardCMS/OrchardCore/pull/7943

weedkiller:
@agriffard I saw you closed this, please share info/sample on creating charts/d3 visuals/tables/calendar, with UI pattern to wire data source

Also please help ensure they responsive if we have SVG element inside.

fasteddys:
Can you also add a graph or chart / calendar or some non text widget here

PR #7877, opened 2020-12-06
Update placement GUI.

microposmp:
I've been thinking about learning more about the placement files to get a better understanding about how they work. When I saw the new module for working with them in the GUI by @TFleury I've spent the day experimenting. Really nice to be able to add placement settings in the GUI.

During the experiments I did some changes to the module:

- Updated the GUI to appear like the other modules under the Design menu. I think these changes are made recently.
- Removed the Empty check as most if not all of the values are optional. When testing I've only used the **place** and **contentType** placement values. All other values have been left blank (null). When leaving the other values blank I couldn't save. Hope I've understood the logic/rules here correctly and don't break anything by this change.

Shape Type: MarkdownBodyPart_Edit

```json
[
  {
    "place": "Parts:0#Content;0",
    "displayType": null,
    "differentiator": null,
    "alternates": null,
    "wrappers": null,
    "shape": null,
    "contentType": [
      "Article"
    ]
  }
]
```


microposmp:
I noticed now that I broke some functionality. The placement links from Part definition pages.

image

I'll revert the broken change and update the PR. Only tested by creating the placements from the Design/Placements page.

microposmp:
Reverted the changes I made that broke functionality and added an item for adding/editing admin placement to the dropdowns.

TFleury:
Hi @microposmp, and thank you for your interest.

> Updated the GUI to appear like the other modules under the Design menu.

Why not, it's more consistent with other modules, even if I don't like to have paging on those lists and prefer client side filtering as it's more efficient.

> Removed the Empty check as most if not all of the values are optional

The `IsEmpty` check is here to keep placements clean and automatically remove void rules.
I added it because when using shortcuts in content definition, it automatically adds a new entry with relevent filters setted up. I think it's important to remove this entry if it's not used.

If a rule defines no "place", neither "shape", "wrappers" or "alternates", it has no effect. So there is no reason to keep it.

> Added an item for adding/editing admin placement

Great.

microposmp:
Hi @TFleury, thanks for reviewing. 

Added back the IsEmpty check and can't see the issue I was experiencing, so the error must have been with the data I saved. Sorry.

agriffard:
Please merge dev and fix the conflict.

microposmp:
Synced with latest updates. Search box is updated.

image

src/OrchardCore.Modules/OrchardCore.Placements/Controllers/AdminController.cs
```diff
@@ -24,40 +29,72 @@ public class AdminController : Controller
...
-                {
-                    ShapeType = entry.Key
-                })
-            });
+                ShapeType = entry.Key
+            }).ToList();
+
+            if (!string.IsNullOrWhiteSpace(options.Search))
+            {
+                shapeList = shapeList.Where(x => x.ShapeType.Contains(options.Search)).ToList();
```
deanmarcussen, line 75:
```suggestion
                shapeList = shapeList.Where(x => x.ShapeType.Contains(options.Search, StringComparison.OrdinalIgnoreCase)).ToList();
```

src/OrchardCore.Modules/OrchardCore.Placements/Views/Admin/Index.cshtml
```diff
@@ -1,72 +1,143 @@
...
+                        <a asp-action="Edit" asp-route-shapeType="@entry.ShapeType" asp-route-returnUrl="@FullRequestPath">@entry.ShapeType</a>
+                    </div>
+                </li>
+            }
+        }
+        else
+        {
+            <li class="list-group-item">
+                <div class="alert alert-info mb-0" role="alert">
+                    @T["<strong>Nothing here!</strong> There are no placments at the moment."]
```
Skrypt, line 79:
```suggestion
                    @T["<strong>Nothing here!</strong> There are no placements at the moment."]
```

Skrypt (approved):

PR #7878, opened 2020-12-06
Correct controller context

giannik:


agriffard:
Thank you. Good catch.

PR #7885, opened 2020-12-07
Add properties to setup context

### LINKED ISSUE #7882: Pass in arbitrary data to setup context
When setting up a tenant or site sometimes you need to pass in some custom data  and use it in your setup recipes.

Example : 
1. When a user registers on my site he submits a form with firstname, lastname  etc.  I then call a workflow  that creates the tenant and executes a setup recipe. In this setup recipe I create a landing page and i want  to assign the fisrtname , lastname  to be set as the displaytext of the landing page content item. 
2. Another use case would be to populate the custom  user profile settings during setup.

This is currently not possible.

submitting a pr that adds to  OrchardCore.Setup.Services.SetupContext an additional bag property : 

```
    /// <summary>
        /// Gets and additional key/value info.
        /// </summary>
        public IDictionary<string, string> Properties { get; } = new Dictionary<string, string>();
```
that can be consumed by recipes using the syntax : 

```
 "Text": "[js: parameters('Properties.FirstName')]"
```
that was  aplied by pr #7824

It is up to developer to populate the properties bag from his workflow task or a custom setup screen.

@deanmarcussen

giannik:
fixes #7882


src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs
```diff
@@ -202,7 +202,8 @@ await scope
                     context.AdminPassword,
                     context.DatabaseProvider,
                     context.DatabaseConnectionString,
-                    context.DatabaseTablePrefix
+                    context.DatabaseTablePrefix,
+                    Properties=context.Properties
```
deanmarcussen, line 206:
```suggestion
                    Properties = context.Properties
```

deanmarcussen:
Would like @sebastienros to review this.

It's cool, because (if I understood correctly) @giannik is using Forms and Workflows to preconfigure Tenants, so the Workflow / Form is able to provide custom properties that the recipes are able to use.

Hence why it's called a `Properties` dictionary, because that is similar to what we use in Workflows

agriffard:
Still expecting a review by @sebastienros

src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs
```diff
@@ -202,7 +202,8 @@ await scope
                     context.AdminPassword,
                     context.DatabaseProvider,
                     context.DatabaseConnectionString,
-                    context.DatabaseTablePrefix
+                    context.DatabaseTablePrefix,
+                    Properties = context.Properties
```
sebastienros, line 206: Can you change the `ExecuteAsync` call to use a `Dictionary<string, object>` instead of an anonymous object like today, and merge the `Properties` values inside this new dictionary?

It won't change how `parameters()` is called since the object is transformed in `JObject`, but your properties will also be available directly like `parameters('FirstName')`.

src/OrchardCore/OrchardCore.Setup.Abstractions/SetupContext.cs
```diff
@@ -74,5 +74,9 @@ public class SetupContext
         /// Gets or sets the errors during set up if there is.
         /// </summary>
         public IDictionary<string, string> Errors { get; set; }
+        /// <summary>
+        /// Gets additional key/value info.
+        /// </summary>
+        public IDictionary<string, string> Properties { get; } = new Dictionary<string, string>();
```
sebastienros, line 80: Can it be a `<string, object>` since I don't see any reason to limit the type of parameters.

sebastienros: Bonus: Could you remove all the "string" properties in this class and use the Properties dictionary to pass them? This way the context will be more generic and won't need to be changed every time we thing the setup screen needs to be changed. The properties like ShellSettings, EnabledFeatures, Recipe, Errors should remain first class properties.

src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs
giannik: done. Please review.

src/OrchardCore/OrchardCore.Setup.Abstractions/SetupContext.cs
giannik: done. Please review.

giannik:
@deanmarcussen please review after commits based on @sebastienros comments.

deanmarcussen:
Just rerunning the ci @giannik looks like it failed somewhere

giannik:
> Just rerunning the ci @giannik looks like it failed somewhere

Tests fixed.

src/OrchardCore.Modules/OrchardCore.Settings/Services/SetupEventHandler.cs
```diff
@@ -17,23 +18,15 @@ public SetupEventHandler(ISiteService siteService)
...
+            IDictionary<string, object> properties,
             Action<string, string> reportError
             )
         {
             // Updating site settings
             var siteSettings = await _siteService.LoadSiteSettingsAsync();
-            siteSettings.SiteName = siteName;
-            siteSettings.SuperUser = userId;
-            siteSettings.TimeZoneId = siteTimeZone;
+            siteSettings.SiteName = properties["SiteName"]?.ToString();
```
hishamco, line 27: Can we define these as `static readonly` fields in a class, so we can avoid magic string issues that may occurs

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs
```diff
@@ -249,17 +249,20 @@ public async Task<ActionResult> Setup(SetupApiViewModel model)
...
-                AdminPassword = model.Password,
                 Errors = new Dictionary<string, string>(),
                 Recipe = recipeDescriptor,
-                SiteTimeZone = model.SiteTimeZone,
-                DatabaseProvider = selectedProvider.Value,
-                DatabaseConnectionString = connectionString,
-                DatabaseTablePrefix = tablePrefix
+                Properties = new Dictionary<string, object>()
+                {
+                    {"SiteName", model.SiteName},
```
hishamco, line 257: Format, make a space before and after the brackets

src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs
```diff
@@ -220,17 +209,9 @@ void reportError(string key, string message)
...
-                    context.AdminEmail,
-                    context.AdminPassword,
-                    context.DatabaseProvider,
-                    context.DatabaseConnectionString,
-                    context.DatabaseTablePrefix,
-                    context.SiteTimeZone,
+                    context.Properties,
                     reportError
-                ), context, logger);
+                    ), context, logger);
```
hishamco, line 214: Extra whitespace

src/OrchardCore.Modules/OrchardCore.Settings/Services/SetupEventHandler.cs
giannik: done

src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs
giannik: done

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs
giannik: Can you fix the formatting ? Did'nt quite get it.

jtkech: Add a space after / before eah bracket `{ "SiteName", model.SiteName }`

jtkech: Under Visual Studio, you can fomat your file by doing ctrl+K+D

src/OrchardCore.Modules/OrchardCore.Settings/Services/SetupEventHandler.cs
jtkech: If so, why not just using `nameof(SetupContext.SiteName)`, in place of creating a class for this

Note: There are many other places where we use "magic" strings, but that may be okay when they are quite obvious.

hishamco: > nameof(SetupContext.SiteName)

I was thinking about that, nothing but there are few cases like `siteTimeZone` not equivalent to `TimeZoneId`

jtkech: So here it would just be

    siteSettings.TimeZoneId = properties[nameof(SetupContext.SiteTimeZone)]?.ToString()

In place of

    siteSettings.TimeZoneId = properties[SetupConstants.SiteTimeZone]?.ToString()

Here the advantage of the nameof is that if you change a setup context prop name, it will not compile.

hishamco: Both are fine, for 

```csharp
siteSettings.TimeZoneId = properties[SetupConstants.SiteTimeZone]?.ToString()
```

if the field has been changed, the value still the same ;) I never mind to use anyone of them instead of magic strings

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs
hishamco: Exactly!! CRTL+K+D is my favorite ;)

giannik: formatting done.  Im using Rider as IDE so it has diffrent opinions for formatting.

src/OrchardCore.Modules/OrchardCore.Settings/Services/SetupEventHandler.cs
giannik: SetupContext has been stripped down and SiteName property and all other string properties have been removed and palced in the Properties bag.

src/OrchardCore/OrchardCore.Recipes.Abstractions/Services/IRecipeExecutor.cs
```diff
@@ -6,6 +7,6 @@ namespace OrchardCore.Recipes.Services
 {
     public interface IRecipeExecutor
     {
-        Task<string> ExecuteAsync(string executionId, RecipeDescriptor recipeDescriptor, object environment, CancellationToken cancellationToken);
+        Task<string> ExecuteAsync(string executionId, RecipeDescriptor recipeDescriptor, Dictionary<string,object> environment, CancellationToken cancellationToken);
```
deanmarcussen, line 10:
```suggestion
        Task<string> ExecuteAsync(string executionId, RecipeDescriptor recipeDescriptor, IDictionary<string,object> environment, CancellationToken cancellationToken);
```

src/OrchardCore/OrchardCore.Recipes.Core/Services/RecipeExecutor.cs
```diff
@@ -39,7 +39,7 @@ public class RecipeExecutor : IRecipeExecutor
             _logger = logger;
         }
 
-        public async Task<string> ExecuteAsync(string executionId, RecipeDescriptor recipeDescriptor, object environment, CancellationToken cancellationToken)
+        public async Task<string> ExecuteAsync(string executionId, RecipeDescriptor recipeDescriptor, Dictionary<string,object> environment, CancellationToken cancellationToken)
```
deanmarcussen, line 42:
```suggestion
        public async Task<string> ExecuteAsync(string executionId, RecipeDescriptor recipeDescriptor, IDictionary<string,object> environment, CancellationToken cancellationToken)
```

src/OrchardCore/OrchardCore.Abstractions/Setup/SetupConstants.cs
```diff
@@ -0,0 +1,15 @@
+namespace OrchardCore.Abstractions.Setup
+{
+    public static class SetupConstants
+    {
```
deanmarcussen, line 4: I'm not a big fan of having static classes for constants.

It forces modules that really shouldn't need to, to take references to projects, where a Magic String would have sufficed.

I think @hisham you might be a bit obsessed with the idea that magic strings are bad, where Orchard Core uses them a lot, for the reasons of avoiding references where not necessary, and leaving the code more readable.

We don't attract bugs from this, we have had more issues where people have used the wrong constant, because it becomes less clear in the code, what the constant means.

The whole point of this change, was to move to a properties dictionary, where we don't need to update the code to change a property. Now we end up with that, but have a class where we have to update the code to change the property.

I'm not going to argue against it strongly, but I'm not a fan.

deanmarcussen (commented):
@giannik mostly I suggest `IDictionary` on the interfaces, and `Dictionary` on the implementations.

The constants, well, just an an opinion.

src/OrchardCore.Modules/OrchardCore.Settings/Services/SetupEventHandler.cs
```diff
@@ -17,23 +19,15 @@ public SetupEventHandler(ISiteService siteService)
...
+            IDictionary<string, object> properties,
             Action<string, string> reportError
             )
         {
             // Updating site settings
             var siteSettings = await _siteService.LoadSiteSettingsAsync();
-            siteSettings.SiteName = siteName;
-            siteSettings.SuperUser = userId;
-            siteSettings.TimeZoneId = siteTimeZone;
+            siteSettings.SiteName = properties[SetupConstants.SiteName]?.ToString();
```
deanmarcussen, line 28:
```suggestion
            if (properties.TryGetValue(SetupConstants.SiteName, out string siteName))
            {
                siteSettings.SiteName = siteName;
            }
```

While it's expected that these properties are supplied, this would avoid any key not found errors in the dictionary. 
Where `properties[SetupConstants.SiteName]? would throw a key not found exception

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs
```diff
@@ -289,6 +293,8 @@ public async Task<ActionResult> Setup(SetupApiViewModel model)
             return Ok(executionId);
         }
 
+
```
deanmarcussen, line 296: remove extra newlines

src/OrchardCore.Modules/OrchardCore.Users/Services/SetupEventHandler.cs
```diff
@@ -18,28 +20,20 @@ public SetupEventHandler(IUserService userService)
...
+            IDictionary<string, object> properties,
             Action<string, string> reportError
             )
         {
             var user = new User
             {
-                UserName = userName,
-                UserId = userId,
-                Email = email,
+                UserName = properties[SetupConstants.AdminUsername]?.ToString(),
```
deanmarcussen, line 29: as above `TryGetValue`

src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs
```diff
@@ -125,20 +126,20 @@ private async Task<string> SetupInternalAsync(SetupContext context)
...
             var shellSettings = new ShellSettings(context.ShellSettings);
 
             if (string.IsNullOrEmpty(shellSettings["DatabaseProvider"]))
             {
-                shellSettings["DatabaseProvider"] = context.DatabaseProvider;
-                shellSettings["ConnectionString"] = context.DatabaseConnectionString;
-                shellSettings["TablePrefix"] = context.DatabaseTablePrefix;
+                shellSettings["DatabaseProvider"] = context.Properties[SetupConstants.DatabaseProvider]?.ToString();
+                shellSettings["ConnectionString"] = context.Properties[SetupConstants.DatabaseConnectionString]?.ToString();
+                shellSettings["TablePrefix"] = context.Properties[SetupConstants.DatabaseTablePrefix]?.ToString();
```
deanmarcussen, line 137: as above TryGetValue. But here we should still set the shellSettings key even if there is no value provided I think @jtkech is it required that the shellsettings directionary have an empty key for `TablePrefix` ? Seems odd. But don't want to risk the breaking of anything

deanmarcussen (commented):
@jtkech would check with you about this 

```
shellSettings["TablePrefix"] = context.Properties[SetupConstants.DatabaseTablePrefix]?.ToString();
```

required to have any empty key for `TablePrefix` ?

src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs
jtkech: It sets values in the memory config source of the shell settings that can hold null values

Then, as i remember in the code we check if the string is null or empty

So i think that's okay

jtkech:
@deanmarcussen 

About `shellSettings["TablePrefix"] = context.Properties[SetupConstants.DatabaseTablePrefix]?.ToString();`

It sets values in the memory config source of the shell settings that can hold null values

Then, as i remember in the code we check if the string is null or empty

So i think that's okay

Yes, this is the part on the right that may throw a key not found exception

deanmarcussen (approved):

PR #7892, opened 2020-12-07
Show Widget ContentType

serverside-is:
Show the Widget `ContentType` next to the Widget title (and edit link), on the Widget list (just like in Orchard 1).

Before:

before

After:

after


agriffard:
Then, we could may be remove the "Widget".

or else, it would be a "{0} Widget" because the type will not alway be before in some languages.

serverside-is:
I removed the "Widget".  Makes sense. The user already knows that every content item on that page is a Widget.
no-widget

Your right about the languages. In my own language, Portuguese, the type would appear at the end: "Widget de RawHtml".

PR #7893, opened 2020-12-07
Added Source Link Package to Dependencies.props 

### LINKED ISSUE #7890: Sourcelink is an unmanaged package reference
From the recent nuget pr `SourceLink` is now giving a warning that is an unmanaged package reference.

Probably just needs to be added to the dependencies

ns8482e:
Fixes #7890

### DIFF
--- a/src/OrchardCore.Build/Dependencies.props
+++ b/src/OrchardCore.Build/Dependencies.props
@@ -47,6 +47,7 @@
     <PackageManagement Include="YesSql.Provider.PostgreSql" Version="1.0.0-beta-1620" />
     <PackageManagement Include="YesSql.Provider.SqLite" Version="1.0.0-beta-1620" />
     <PackageManagement Include="YesSql.Provider.SqlServer" Version="1.0.0-beta-1620" />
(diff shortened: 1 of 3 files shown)


deanmarcussen:
I think the projecttemplates project is still throwing a warning.

@jtkech might have some thoughts on that. 
Is it weird that we’re publishing that to nuget?

Because it doesn’t have the .targets file in it so doesn’t understand what PackageManagement is

ns8482e:
@deanmarcussen  Removed `SourceLink` nuget package from `OrchardCore.ProjectTemplates`

jtkech:
@deanmarcussen @ns8482e 

> Because it doesn’t have the .targets file in it so doesn’t understand what PackageManagement is

Yes, it only import `OrchardCore.Commons.props`

But that's okay to just remove the `SourceLink` reference from `OrchardCore.ProjectTemplates` as we don't need it for this project, we don't need it too in the project files of the templates that are produced for a given user.

jtkech (approved):

deanmarcussen:
> @deanmarcussen Removed `SourceLink` nuget package from `OrchardCore.ProjectTemplates`

Cool, good choice

deanmarcussen (approved):

PR #7899, opened 2020-12-08
Settings Group Ids

agriffard:
All Group Ids starting with lowercase letter.

Do not suffix with `Settings`.


deanmarcussen (approved):
Not sure what issue its fixing, but isn't breaking anything

PR #7930, opened 2020-12-10
OrchardHelperExtensions

agriffard:
Do not use namespace in a OrchardHelperExtensions class.


deanmarcussen (approved):
Probably didn't need renaming but all good :)

PR #7933, opened 2020-12-10
Display drivers

agriffard:
Some drivers (Part and Workflows) are called `...Display` and some `...DisplayDriver`.

Trying to be explicit (instead of keeping the names short).


src/OrchardCore.Modules/OrchardCore.Contents/Workflows/Drivers/ContentCreatedEventDisplayDriver.cs
```diff
@@ -0,0 +1,13 @@
+using OrchardCore.ContentManagement.Metadata;
```
hishamco, line 1: Is the aim of this PR to rename or add new display driver(s)  too?

hishamco (commented):
LGTM but this is sort of breaking changes, but it's acceptable while `1.0.0` not release, waiting for other members comment

src/OrchardCore.Modules/OrchardCore.Contents/Workflows/Drivers/ContentCreatedEventDisplayDriver.cs
agriffard: I don't know why those ones haven't been dtected as renamed like the others.

agriffard:
Why would it be breaking changes?

hishamco:
> Why would it be breaking changes?

Because there are public types

sfmskywalker (approved):

sfmskywalker:
@hishamco Although technically speaking you are correct, in practice though I reckon that the odds of custom modules directly referencing a public display driver are really small. So seems like a non-issue to me.

hishamco:
So if it's like that let us merge

PR #7943, opened 2020-12-11
Dashboard widgets

agriffard:
Related to https://github.com/OrchardCMS/OrchardCore/issues/2246#issuecomment-742687838


hishamco:
Could please share some screenshoots, the dashboard got pretty than before, Thanks @agriffard

agriffard:
It is a wip, nothing to show yet.

Expecting some help from @deanmarcussen and @sebastienros to connect the dots.

agriffard:
@jtkech Is there a way to make it a higher priority to be used as the `/admin` route?

jtkech:
@agriffard 

So it depends on `OrchardCore.Admin` but we would need to run its `Configure()` before

In the dashboard module startup try

        public override int ConfigureOrder => -10;

sebastienros (commented):
Only thing is to provide a shape to edit the position of the widget in its editor. And load the widget ordered by position.

agriffard:
Edit Position:

media

src/OrchardCore.Modules/OrchardCore.AdminDashboard/Assets/scss/admin-dashboard.scss
```diff
@@ -0,0 +1,58 @@
+.fade-in-pager {
```
sebastienros, line 1: Please remove any rule that is not "required" for this page.

src/OrchardCore.Modules/OrchardCore.AdminDashboard/Controllers/DashboardController.cs
```diff
@@ -0,0 +1,87 @@
...
+            New = shapeFactory;
+            _notifier = notifier;
+            S = stringLocalizer;
+            H = htmlLocalizer;
+            _logger = logger;
+        }
+
+        public async Task<IActionResult> Index()
+        {
+            if (!await _authorizationService.AuthorizeAsync(User, Permissions.AccessAdminDashboard) || !await _authorizationService.AuthorizeAsync(User, Permissions.ManageAdminDashboard))
```
sebastienros, line 70: Just check AccessDashboard, since it's implied by ManageDashboard.

src/OrchardCore.Modules/OrchardCore.AdminDashboard/Migrations.cs
```diff
@@ -0,0 +1,43 @@
...
+        public int Create()
+        {
+            SchemaBuilder.CreateMapIndexTable<DashboardPartIndex>(table => table
+               .Column<double>("Position")
+            );
+
+            _contentDefinitionManager.AlterPartDefinition("DashboardPart", builder => builder
+                .Attachable()
+                .WithDescription("Provides a way to add widgets to a dashboard.")
+                );
```
ns8482e, line 31: How about Weld `DashboardPart` in Driver, instead of defining in migraion?

Any content item created with stereotype Dashboard item should automatically weld to this part, Like we have as in FlowMetatadata and WidgetMetadata

sebastienros: We talked about it, but I think the reason is that there was not much advantage to weld it instead of doing it on purpose, either by code or UI. It's just a requirement. Welding makes things too magical (imo).

sebastienros: We are only hoping that you can take this feature and make a nice editor when it's merged.

sebastienros (approved):

ns8482e:
I am working to add drag drop and resizing effect to dashboard - WIP. Let me know your suggestions.

dashboard-2

src/OrchardCore.Modules/OrchardCore.AdminDashboard/Migrations.cs
ns8482e: Only reason, I was thinking on to welding automatically to `Dashboard` stereotype is because, it's should common behavior of all Dashboard content item (more like metadata to dashboard functionality than manageable content part), and shouldn't require user to explicitly define part on every content type definition only defining stereotype `Dashboard ` should be enough.

Also editing Position( and Width, Height for resizing) property is done on manage page (when drag drop is implemented) and  editing these meta data properties on edit content screen becomes less significant.

agriffard:
Nice example. I merged it so you can give it a look.
Some drag N drop and resize glyphs could be interesting.
Do you use Html or Liquid templates? Did you just add a media field?
Is it only for users with admin permission? (Where are the edit and delete buttons?)
Not a big fan of the bottom/right yellow border (Bootstrap Cards styles should be sufficient).

sebastienros:
@ns8482e I would use a "snap grid" (which you might already do) to keep things aligned. 

Editing options:
1- Have a button to go in edit/view mode, and the edit mode being protected by a permission. In edit mode we can add new widgets, move, edit and delete them. 
2- Don't have an edit mode and show the "cogs" icon on each widget to edit them, and all widgets can always be moved or resized. In that case have an "undo" alter when there is a change. 

I think having an edit mode is better since most of the time widget will be read-only. So less risky to make mistake.

About the welded DashboardPart, creating a dashboard widget isn't something that is done commonly, and having to add a part to a content type is not much harder that knowing we have to set a stereotype, so I'd rather not have it automatic for now.

ns8482e:
@agriffard  @sebastienros why we have `DetailAdmin` shape and not using the default `Detail` shape?

PR #7944, opened 2020-12-11
AdminMenuService: Correct namespace

agriffard:
Use `.Services` in namespace.


hishamco (approved):
LGTM be sure there's no dependency somewhere that requires to change the using namespace

hishamco:
I just verified .. it's fine to merge

Issue #7935, opened 2020-12-10
Media Tokens invalidating browser / cdn cache


deanmarcussen:
The Media Tokens feature has a few little problems :(

Underlying it is using DP to encrypt the command dictionary. 

It is cached, but on server restart, it generates a different string (as DP does on each encrypt).

This is invalidating the browsers cache, every time a server is restarted, and also then a CDN cache.

I don't think it's contributing to any growth in the IS cache, (but have to check again now)

Probably what it needs to do us use a signature hash, instead of encrypting the entire string, salted by the system, so it is consistent, and obscufated. Obviously not quite as secure to always be presenting the same hash value for a given combination of commands, but this was always an obscufation requirement, rather than full blown DP.

Other ideas welcome.

```
media/home-bg.jpg?token=CfDJ8LUblBuI4cxMuTWlnqoGNrdxQTBj7aDU-kfEI9ZDfvacIrBsc3R1fOK_wdNd5_lzM_GLvDSzxkNPfpTD-tDqWcZukbRogejppGh0j390_UFcT80e-pwDbMa7s2tqUVbRgKZlX_wtGd3F2damIaqU02Yh1eTx8IcTtA9gZCmG3o6CoRkA41IA476bhLPVyZLZ_Q
```

after restart, same image, different query string
```
media/home-bg.jpg?token=CfDJ8LUblBuI4cxMuTWlnqoGNrfbU_5nqMClxDpRkRLY9VBim36vmJzyv-wISKVKR0fP8hkAJGtQQHdTUQ00RwXQqa2-GEQ4WDhm9D54JKhW_Dwuspiwi9om7oRRhFWNJ3VcoeEqLq99YLsl65e9T2PMo3V5KuZ849CivBUXxQL8PlsmRRyHGI5HBm-xov2kkgOVqQ
```

Issue #7941, opened 2020-12-11
Race conditions running Orchard migrations on multiple instances


jersiovic:
Hello

We are facing race conditions on migration when we use two instances in an Azure App Service with Container.

We receive exceptions cause another instance executed a migration and the changes on migrations are attempeted to be executed by both instances. This causes problems cause one of the two instances didn't complete its initialization.

Taking a look at the code we saw Orchard uses SemaphoreSlim to avoid two threads excuting at the same time migrations for a tenant. Is there any artifact like TaskLease at Orchard 1? Should it be used to avoid problems on the scenario I mention?

jtkech:
We have an `IDistributedLock` that we don't use yet, the default implementation is a `LocalLock` based on multiple `SemaphoreSlim` per tenant, and we have a concrete implementation based on Redis.

The interface allows to `TryAcquireLockAsync()` a named lock (for a given tenant) in a given timeout and that will be valid for a given expiration, allowing to do e.g. atomic operations in a distributed env. There is also an `IsLockAcquiredAsync()` allowing to do exclusive tasks by just doing nothing if a named lock is already acquired.

Maybe we will use it e.g. for Workflow Singleton in a distributed env, Exclusive BackgroundTask ...

About migrations, when they are executed on a fresh setup, normally it is on only one instance. Through the admin UI, by enabling features, executing recipes, yes it may happen concurrently. When the code is updated with some migration steps, yes it may happen concurrently if multiples instances are starting, is it your scenario?

So, maybe we will need to also use an `IDistributedLock` for migrations, but maybe only if it is not during a fresh setup, hmm but in this case the lock resolved would be a simple `LocalLock`, hmm maybe not at some point if you enable `RedisLock` in the setup recipe. Anyway, i think it is better to not use a lock on setup (`TenantState.Initializing`), but yes maybe we have to use it on migrations if the tenant is already `TenantState.Running`.

> We are facing race conditions on migration when we use two instances

Does it throws exceptions? Or just let the 2 instances not synchronized.

jersiovic:
> We have an `IDistributedLock` that we don't use yet, the default implementation is a `LocalLock` based on multiple `SemaphoreSlim` per tenant, and we have a concrete implementation based on Redis.
> 
[...]

PR #7953, opened 2020-12-12
Refactor media tokens

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/7935

Replaces using Data Protection to encrypt tokens, as the encrypted token changed every server restart (because of course encryption should never produce the same value twice, for a given input - dup!).

This caused the browser and/or CDN to refresh all the images every time the server is restarted/deployed.

The is-cache wasn't affected by this, just the browser/cdn caches.

Uses instead an HMAC, with a unique randomly generated tenant key.

It becomes obfuscation, rather than encryption.

I've used a migration to set the key on existing installations, and a feature handler to set the key on new installations, OR when the media module is disabled, and re-enabled. 
Consider it a shortcut to exposing a settings driver for it, as it didn't feel like it should be super exposed.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenOptionsConfiguration.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenOptionsConfiguration.cs
@@ -1,0 +1,39 @@
+using System.Security.Cryptography;
+using Microsoft.Extensions.Options;
+using OrchardCore.Entities;
+using OrchardCore.Settings;
+
+namespace OrchardCore.Media.Processing
+{
+    public class MediaTokenOptionsConfiguration : IConfigureOptions<MediaTokenOptions>
+    {
+        private readonly ISiteService _siteService;
+
+        public MediaTokenOptionsConfiguration(ISiteService siteService)
+        {
+            _siteService = siteService;
+        }
+
+        public void Configure(MediaTokenOptions options)
+        {
+            var mediaTokenSettings = _siteService.GetSiteSettingsAsync()
+                .GetAwaiter().GetResult()
+                .As<MediaTokenSettings>();
+
+            if (mediaTokenSettings.HashKey == null)
+            {
+                var siteSettings = _siteService.LoadSiteSettingsAsync().GetAwaiter().GetResult();
+
+                var rng = RandomNumberGenerator.Create();
+
+                mediaTokenSettings.HashKey = new byte[64];
+                rng.GetBytes(mediaTokenSettings.HashKey);
+                siteSettings.Put(mediaTokenSettings);
+
+                _siteService.UpdateSiteSettingsAsync(siteSettings).GetAwaiter().GetResult();
+            }
+
+            options.HashKey = mediaTokenSettings.HashKey;
+        }
+    }
+}
--- a/src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenOptionsConfiguration.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenOptionsConfiguration.cs
@@ -1,4 +1,3 @@
-using System.Security.Cryptography;
 using Microsoft.Extensions.Options;
 using OrchardCore.Entities;
 using OrchardCore.Settings;
@@ -16,24 +15,10 @@ namespace OrchardCore.Media.Processing
 
         public void Configure(MediaTokenOptions options)
         {
-            var mediaTokenSettings = _siteService.GetSiteSettingsAsync()
+            options.HashKey = _siteService.GetSiteSettingsAsync()
                 .GetAwaiter().GetResult()
-                .As<MediaTokenSettings>();
-
-            if (mediaTokenSettings.HashKey == null)
-            {
-                var siteSettings = _siteService.LoadSiteSettingsAsync().GetAwaiter().GetResult();
-
-                var rng = RandomNumberGenerator.Create();
-
-                mediaTokenSettings.HashKey = new byte[64];
-                rng.GetBytes(mediaTokenSettings.HashKey);
-                siteSettings.Put(mediaTokenSettings);
-
-                _siteService.UpdateSiteSettingsAsync(siteSettings).GetAwaiter().GetResult();
-            }
-
-            options.HashKey = mediaTokenSettings.HashKey;
+                .As<MediaTokenSettings>()
+                .HashKey;
         }
     }
 }
--- a/src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenService.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenService.cs
@@ -1,34 +1,31 @@
 using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.Linq;
-using System.Web;
-using Microsoft.AspNetCore.DataProtection;
+using System.Security.Cryptography;
+using System.Text;
 using Microsoft.AspNetCore.WebUtilities;
 using Microsoft.Extensions.Caching.Memory;
-using Microsoft.Extensions.Primitives;
-using Newtonsoft.Json;
-using OrchardCore.Modules;
+using OrchardCore.Entities;
+using OrchardCore.Settings;
 using SixLabors.ImageSharp.Web.Processors;
 
 namespace OrchardCore.Media.Processing
 {
     public class MediaTokenService : IMediaTokenService
     {
-        private readonly IDataProtector _dataProtector;
-        private readonly IMemoryCache _memoryCache;
-        private readonly HashSet<string> _knownCommands = new HashSet<string>();
-
         private const string TokenCacheKeyPrefix = "MediaToken:";
-        private const string CommandCacheKeyPrefix = "MediaCommands:";
+        private readonly IMemoryCache _memoryCache;
+        private readonly ISiteService _siteService;
+
+        private readonly HashSet<string> _knownCommands = new HashSet<string>();
+        private readonly byte[] _hashKey;
 
         public MediaTokenService(
-            IDataProtectionProvider dataProtectionProvider,
             IMemoryCache memoryCache,
+            ISiteService siteService,
             IEnumerable<IImageWebProcessor> processors)
         {
-            _dataProtector = dataProtectionProvider.CreateProtector("MediaProcessingTokenService");
             _memoryCache = memoryCache;
+            _siteService = siteService;
             foreach (IImageWebProcessor processor in processors)
             {
                 foreach (string command in processor.Commands)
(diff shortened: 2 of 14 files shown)


src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenService.cs
```diff
@@ -1,44 +1,45 @@
...
             {
                 foreach (string command in processor.Commands)
                 {
                     _knownCommands.Add(command);
                 }
             }
+
+            var siteSettings = _siteService.GetSiteSettingsAsync().GetAwaiter().GetResult();
+            var mediaTokenSettings = siteSettings.As<MediaTokenSettings>();
+            _hashKey = mediaTokenSettings.HashKey;
```
deanmarcussen, line 39: @jtkech I opted here, as it is a singleton, to retrieve the value in the constructor, rather than making every method async, and checking in the methods, whether the value was null, and retrieving it if null.

(some methods that use this would be breaking to make async as well)

jtkech: @deanmarcussen 

Yes `ISiteService` is an exception (see below) as it is a singleton using a scoped `ISession`, so yes you can inject it in a singleton, but at least you should not call its methods outside a scope which may happen in a singleton constructor. 

It works as long as `IMediaTokenService` is injected only by scoped services, but i think it's not good to do that.

Also `SiteSettings` is a shared document but here `mediaTokenSettings` may not be in sync with other running instances. E.g. if a migration completes later, or if you re-generate the key on one instance, currently it is done by disabling / re-enabling the feature so the tenant is rebuilt, but here you would need the distributed tenants feature.

So maybe better to get the key from the cached settings on demand, and maybe generate it if it doesn't exist yet.

Also not sure it's good to use a migration step, maybe a shell activated handler to generate the hash key if it doesn't exist yet. OR as said above just generate the key on demand. As a side note, there are remaining things as migrations that would need to be atomic / exclusive, see #7941, i'm working on, e.g. on WF Singleton.

Then okay for the `IFeatureEventHandler` if the only way to re-generate a key is to disable / re-enable the feature, but why not a custom site settings driver / service to do that, e.g. through a `Generate a new key` button, but still generate it automatically on demand if it doesn't exist yet.

---

The `ISiteService` singleton is an exception as it uses a scoped `ISession`, it is injected by the `HomeRouteTransformer` singleton, hmm just tried to register this tansformer as a scoped service, it works. But it is also injected by an `IConfigureOptions<>` so somewhere by an `IOptions<>` singleton, here we would need to create a scope on the fly, or just retrieve the current one as we do in `ISiteService`.

Maybe worth to try to make `ISiteService` a scoped service to avoid bad usage, i will try it.

deanmarcussen: Ok, so sounds like it's better to get the site settings on demand once, and store the value locally in the singleton, but retrieve outside of the constructor.

Could do shell activating, rather than a migration, but the migration was only a one hit **for existing sites**.

I preferred the `IFeatureEventHandler` because I didn't want to expose it through a settings driver (figuring then people will expect a recipe step, and deployment step for it, when it should really be randomly generated per server - and rarely changed)

src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenService.cs
```diff
@@ -61,54 +60,75 @@ public string TokenizePath(string path)
...
 
-                return null;
-            });
+        private void GetHashKey()
+        {
+            if (_hashKey == null)
+            {
+                var siteSettings = _siteService.GetSiteSettingsAsync().GetAwaiter().GetResult();
+                var mediaTokenSettings = siteSettings.As<MediaTokenSettings>();
+                _hashKey = mediaTokenSettings.HashKey;
```
deanmarcussen, line 130: So @jtkech I moved it outside the constructor.

It's not totally on demand though, still held locally for the instance. 
I prefer this over making the methods async, and fully on demand, as it would break the `Orchard.AssetUrl` helper.

I left the migration for the moment, but it was intended as a one hit migration for existing sites. And again trying to avoid to much `.GetAwaiter()` work inside this service.

Could change that if you feel strongly about it ;)

src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenOptionsConfiguration.cs
```diff
@@ -0,0 +1,39 @@
...
+                var rng = RandomNumberGenerator.Create();
+
+                mediaTokenSettings.HashKey = new byte[64];
+                rng.GetBytes(mediaTokenSettings.HashKey);
+                siteSettings.Put(mediaTokenSettings);
+
+                _siteService.UpdateSiteSettingsAsync(siteSettings).GetAwaiter().GetResult();
+            }
+
+            options.HashKey = mediaTokenSettings.HashKey;
```
deanmarcussen, line 36: @jtkech all felt wrong still.

So I made an options class, handled the migration generation there, and made the `IMediaTokenService` scoped.

Kept the feature event handler for refreshing the key

src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenService.cs
jtkech: > Could do shell activating, rather than a migration, but the migration was only a one hit for existing sites.

Yes i know, maybe because i was thinking about making migrations atomic (related to #7941) and didn't want to add complexity here. Note: We already use a local lock for activating a shell (where auto migrations happen)

I was also thinking about another instance generating the key, so anyway we would need to first check if the key already exists before generating a new one, so i thought about the handler to have a better chance to be in sync.

> I preferred the IFeatureEventHandler because I didn't want to expose it through a settings driver

Okay

> when it should really be randomly generated per server - and rarely changed

**Do you mean a different key per server even they all serve the same web app?** If so, we would need to store the key per server, so not through a shared document as site settings. Otherwise, if the key need to be in sync across nodes for a given web app, maybe better to check a shared identifier as we do through shared documents.

Can't right now but i will focus on it this night, but need to know if the key needs to be in sync

src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenOptionsConfiguration.cs
jtkech: @deanmarcussen 

Okay no problem i will work on it this night

As a reminder: Not a problem to use `GetAwaiter().GetResult()` for site settings, as i remember they are aleady get through a filter and then cached for the whole scope lifetime.

But as said above, really need to know the key needs to be in sync across instances serving the same we app

src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenService.cs
deanmarcussen: Thanks @jtkech 

Same key per server. 

Different key for development etc (so it is not important to be able to set the key through a recipe)

It would be a rare requirement to want to change it. (changing it would be a result of a security concern, as it would cause every image to refresh in the browsers cache. Just the browsers cache, not the is-cache. So you would only change it as a result of an attempted attack, that targeted this particular area, i.e. disk filling through images)

So I think site settings is ok for this.

It's random rather than something we set in code, so that it is a per site key (and per tenant).

Now in terms of being in sync between servers.

I assume that activating the feature across a multi-node site would probably trigger the reset off the key twice, but through the shared document sync, only the first one would win? As you have already avoided races there.

jtkech: @deanmarcussen okay cool i understand

> I assume that activating the feature across a multi-node site would probably trigger the reset off the key twice

Yes but only if you enable the Distributed Tenants feature which needs an `IDistributedCache` implementation as `RedisCache`, but that is a separate feature.

> but through the shared document sync, only the first one would win?

Maybe yes if in a given usage we save a new value only if there is no existing one, but it may still happen concurrently. And no if 2 instances want to explicitly save a new value, the last win. That's why we use identifiers (one as a cache entry and one embedded in the document itself) to keep in sync the 2 instances, but each instane need to get data on demand from the multi level cache to keep up to date values.

That's the main goal, keep in sync instances between them and the data store. If we would want that the first win, we would need to use a distributed lock to do an exclusive work.

> As you have already avoided races there.

That's okay here, normally it is thread safe, but as said atomicity / exclusive jobs and so on (not all services need them) are other concerns i'm working on

jtkech:
@deanmarcussen Okay i could work on it

So okay, `ISiteService` is already used in different `IConfigureOptions<>` involving an `IOptions<>` singleton, so if we plan to make `ISiteService` a scoped service, anyway we would have to upate them too.

So okay, there are already singleton options that need the tenant to be rebuilt if they change on another node, the hash key (that is rarely changed) being just an additional one. So, to keep these options in sync across nodes, e.g. the `RedisCache` is not sufficient, we also need the `OC.Tenants.Distributed` feature, but that's okay.

Otherwise, as discussed we would need to get the hash from the synchronized document on demand, as we do for other services, and in that case it would be better to make methods async, but yes because it is rarely changed that's okay.

- So for now, `IConfigureOptions<MediaTokenOptions>` uses a valid pattern, but i removed the site settings updating, so that it uses the exact same pattern as others `IConfigureOptions<>` that use `ISiteService`.

- Then, i kept `MediaTokenSettingsUpdater` as an `IFeatureEventHandler` so that a new key is generated on enabling the media feature, just added a check on the feature name.

- And i also made `MediaTokenSettingsUpdater` a `ModularTenantEvents` so that when the shell is activated, and only if the media feature is enabled, the hash key is generated if it doesn't exist yet.

So i approved this PR and i let you have a look and try it before merging if you agree with the changes.

jtkech (approved):

src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenSettingsUpdater.cs
```diff
@@ -0,0 +1,77 @@
...
+            _siteService = siteService;
+        }
+
+        public override async Task ActivatedAsync()
+        {
+            var mediaTokenSettings = (await _siteService.GetSiteSettingsAsync()).As<MediaTokenSettings>();
+
+            if (mediaTokenSettings.HashKey == null)
+            {
+                var siteSettings = _siteService.LoadSiteSettingsAsync().GetAwaiter().GetResult();
```
deanmarcussen, line 30: @jtkech looks good. Haven't tried it yet ;)

Will just need to change these to await (I'll do it shortly, after I try it)

As you say still does not solve the multiple instance issue.

But we will need to have a better solution for that, for migrations as well, so we will remember to apply the same solution here.

jtkech: @deanmarcussen 

> Will just need to change these to await

Oops forgot it

> As you say still does not solve the multiple instance issue.

Not really an issue, that's okay if the `OC.Tenants.Distributed` feature (the one i demoed) is enabled, so that when a node is released / reloaded, other nodes are also released / reloaded

jtkech: Done

PR #7954, opened 2020-12-12
Recipes: Order by display name

agriffard:
In each group category, order the recipes by display name.

Also: Refactor the UI to display more condensed information.


agriffard:
Before:

image

After:

image

agriffard:
Currently with my sample recipes:

image

Skrypt:
Looks good but what are these numbers?

agriffard:
> Looks good but what are these numbers?

This is how I named my recipes, in order to display them in that order and the user that setup a site after a blank recipe install can execute them one after another.

Skrypt:
I would have thought that they would have been ordered by number since it's their "DisplayName". Strange unless I don't understand. Else the PR looks good.

agriffard:
> I would have thought that they would have been ordered by number since it's their "DisplayName". Strange unless I don't understand. Else the PR looks good.

Precisely. It is how it looks like with the latest dev package, not with the PR change.

PR #7956, opened 2020-12-12
AdminMenus: Search with ignore case

agriffard:
Use `Contains `with `StringComparison.OrdinalIgnoreCase` to filter the admin menus or else, you won't be able to find a menu with a capital case if you search with lower case.


deanmarcussen:
Looks good, can you do the other 5 or 6 document searches @Skrypt added recently too?

Search for `Contains(options.Search` in the solution - only the ones where it's a document dictionary that it's searching, not the YesSql queries.

sfmskywalker (approved):

deanmarcussen:
TemplateController still to do I think

agriffard:
https://github.com/OrchardCMS/OrchardCore/blob/22a27aec06b43063e53b3d81f29019e7367a904e/src/OrchardCore.Modules/OrchardCore.Templates/Controllers/TemplateController.cs#L85

PR #7957, opened 2020-12-12
Search Filters

### LINKED ISSUE #7968: Search in Lucene indices and Queries
If you type a search and press Enter, the resuls are not filtered on those admin pages.
It only works by clicking Search.

/cc @Skrypt

agriffard:
Remove Filters prepended input group.
Remove appended Search button: All search boxes work with Enter (bulkAction hidden button moved below).
Use search icon andd css classes. Ex: https://codepen.io/gungorbudak/pen/ooKNpz
Use `btn-success` for Add/Create/New buttons.

Before:

image

After:

image

Fixes #7968


agriffard:
Content Filters drodpdown at the end of the search box.

image

Skrypt:
hmm not sure Antoine. I kept a "search" button where there is an actual form POST action so that you can either click the button or press enter to submit the form. Though, if the list of items is filtered with javascript (when there is no pager exceptionally) then there is no "search" button which make sense because it is filtered on each key clicked.

Also, here, I think @hishamco was right about using a btn-secondary because the action of adding a record in the database here is not made from that "Add something" button but from the form where it links to. So, here, this is only an redirect to some page action thus I don't think it should be a btn-primary. But, you know I can change my mind on that one pretty easily. I just think it looks better without using the multi-color buttons. The only buttons that should not be btn-secondary (grey) is the button to delete something.

But here your changes are pretty much running in circle from what we did previously. It a question of taste which I can't argue with.

Skrypt (commented):
hmm not sure Antoine. I kept a "search" button where there is an actual form POST action so that you can either click the button or press enter to submit the form. Though, if the list of items is filtered with javascript (when there is no pager exceptionally) then there is no "search" button which make sense because it is filtered on each key clicked.

Also, here, I think @hishamco was right about using a btn-secondary because the action of adding a record in the database here is not made from that "Add something" button but from the form where it links to. So, here, this is only an redirect to some page action thus I don't think it should be a btn-primary. But, you know I can change my mind on that one pretty easily. I just think it looks better without using the multi-color buttons. The only buttons that should not be btn-secondary (grey) is the button to delete something.

But here your changes are pretty much running in circle back from what I did previously. It a question of taste which I can't argue with. My changes we're intentional ...

One thing I like is that search icon though.

agriffard:
It is indeed a question of taste. I like minimalism.
Filters preprend and Search button don't bring much value for me.

The Add buttons are using btn-success, not primary. 
Edit buttons already have a green button.
According to me, btn-secondary should be only used for Cancel and Actions button.

Skrypt:
The edit buttons always used btn-primary just like Github does. The only difference is that Github uses a green color as their primary color. So they are not using btn-success on their primary action buttons just because it is green.

Here is the proof : The Squash and merge button HTML 

image

agriffard:
image

agriffard:
image

Skrypt:
Well I'm reviewing this and it requires more work because it would require to do these changes everywhere for consistency. I think we need to have a guideline to stick to with buttons colors. Because everyone has their taste so let's do a guideline and stick to it.

agriffard:
Ok, I reverted all the create buttons to `btn-secondary`

Skrypt:
Moved back the hidden form inputs right after the `<form>` declaration.

Skrypt (approved):

PR #7962, opened 2020-12-12
List: Change background color on hover item

agriffard:
media


Skrypt (approved):

hishamco:
LGTM nothing but why we didn't use tables bootstrap already have built-in classes for hover and alternate rows

Skrypt:
Because tables are generally not recommended to do this anymore. Tables takes some extra time to render because they evaluate height which a common div or list doesn't. You should try to avoid using Tables most of the time.

PR #7964, opened 2020-12-12
Content Types and Parts list

### LINKED ISSUE #7960: Filters: Top border disappear on search
When you filter a page (ex: Content types), the first item can disappear so there is no top border anymore:

media

agriffard:
Use list-item-group (less borders):

media

Fixes #7960


Skrypt:
I think it should be fixed by using CSS because here you use a workaround. I saw that one coming last week. I'll take a look.

PR #7971, opened 2020-12-13
Tenants States

agriffard:
Display the State on the title line.

Purpose: Display 10 tenants and show the pager without having to scroll.

image


hishamco:
Show the badge underneath will looks better

agriffard:
> Show the badge underneath will looks better

By underneath, do you mean below?

My goal is to condense the display.

hishamco:
> By underneath, do you mean below?

Yes, coz the status in the badge are not equal in terms of characters length, that's why I see the UI is need to be organized little bit

Skrypt:
Put them between the name of the tenant and the url

agriffard:
Between title and url:

image

Skrypt:
How about using a grid here with a col-4 which would collapse on md breakpoint to col-md-12?

Skrypt:
Again here let's just see what Github does : 

image

hishamco:
We can move them below as we already did in content items or use circle icons as @Skrypt suggested

agriffard:
image

hishamco:
@agriffard this could be simplfied by #7987

agriffard:
Replaced by #7987

PR #7976, opened 2020-12-13
Tenant recipes

agriffard:
Sort recipes by display name in tenant create/edit.


hishamco:
Can we do the sort on the controller, instead of the each view?

agriffard:
> Can we do the sort on the controller, instead of the each view?

OK, done.

hishamco (approved):

PR #7977, opened 2020-12-13
Fields: Display definition name

### LINKED ISSUE #7920: Fields display
The idea is to use the display name everywhere in fields display and to provide some default positions for each kind of field, except for the Html and Markdown.

Sorry, I pushed the changes to dev (Forgot to make a branch): https://github.com/OrchardCMS/OrchardCore/commit/4044d7d38f825179b94a6929f1bad0efdd8b55e6

agriffard:
Display the definition name in missing fields.

Do not display if no value.

Fixes #7920

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Views/BooleanField.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Views/BooleanField.cshtml
@@ -1,10 +1,12 @@
 @model OrchardCore.ContentFields.ViewModels.DisplayBooleanFieldViewModel
-@using OrchardCore.Mvc.Utilities;
+@using OrchardCore.Mvc.Utilities
+@using OrchardCore.ContentManagement.Metadata.Models
 
 @{
     var name = (Model.PartFieldDefinition.PartDefinition.Name + "-" + Model.PartFieldDefinition.Name).HtmlClassify();
 }
 
 <div class="field field-type-booleanfield field-name-@name">
-    @Model.Value
+    <label for="@name">@Model.PartFieldDefinition.DisplayName()@T[":"]</label>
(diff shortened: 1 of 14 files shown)


deanmarcussen (commented):
I'm against this one @agriffard 

The beauty of the OC templates, is that by and large, they are light, and non intrusive. 
This means that you don't need to template many of them when building sites. They can be used directly, and styled with css.

Here by changing to show display names everywhere, it means that to actually use these fields with display management, every template must be overridden, to get back to just the field values.

Because in a site you would never (very rarely) want to use the display name anywhere.

e.g. Subtitle on the blog recipe. 
You wouldn't have a template that says `Subtitle: my subtitle`

So for me it's quite breaking.

Now I still don't know why you want to do this, i.e. what advantage does it bring?

Maybe we can solve that problem instead?

agriffard:
Is it possible in Liquid to get the equivalent of `@Model.PartFieldDefinition.DisplayName()` when you want to make a template that override a field's rendering?

deanmarcussen:
I did expose `PartFieldDefinition` to liquid a while back for tags - see
`BlogPost-Tags-TaxonomyField-Tags.Display.liquid`

`{% assign name = Model.PartFieldDefinition.PartDefinition.Name | append: '-' | append: Model.PartFieldDefinition.Name | html_class %}`

However that might not be the display name? Worth checking as it maybe the technical name.

You should still be able to get it by accessing the Settings property (like the extension method does) in liquid.

agriffard:
Ok, indeed because the DisplayName() is an extension method.

deanmarcussen:
`Model.PartFieldDefinition.PartDefinition.Settings.ContentPartFieldSettings.DisplayName` 

or something similar with any luck

PR #7987, opened 2020-12-14
Inline tenant status

hishamco:


agriffard:
Nice, I like it.

Can you add a `title="@T[...]"` with the state and use `fa-xs`?

hishamco:
I forgot the title, also I was looking for something like `fa-xs` ;)

Thanks

hishamco:
Seems localization will enforce me to return the switch back :(

agriffard:
May be init the variable cssClass and state in one switch and then use them in the html.

hishamco:
> May be init the variable cssClass and state in one switch and then use them in the html.

It will nothing the localization extractor will unable to find them, while they are dynamic

Issue #7715, opened 2020-11-22
Do we want important features dependent on shortcodes feature?


davidhayden:
I decided to disable the shortcodes feature on a client website because I wanted to see if it was responsible for mangling sample code in a knowledge base that included brackets ("[" and "]"), only to realize that disabling shortcodes meant I had to disable the Content Fields and HTML features that have a dependency on it.

Is there a reason we have a hard dependency on the nice-to-have shortcodes feature in other key CMS features? Can't we instead enable features in those modules when and if the shortcodes feature is enabled? Either way, I don't believe I have an option to opt-out like I do with the HTML sanitizer. 

![shortcodes-dependency](https://user-images.githubusercontent.com/96801/99916968-778b2600-2cdb-11eb-87f7-c2d553bd0aa1.png)

Skrypt:
I think the best approach to this would be to add something like a `<code>[ something ]</code>`  that would escape the shortcodes to try to render.

davidhayden:
If it ignored `code` blocks that would certainly help.

I think the problem is bigger though.

I should be able to disable the shortcodes feature without disabling all my CMS editors. If I disable shortcodes, I disable all Content Fields, HtmlBodyPart, MarkdownPart, and possibly other important features.

I should be able to opt-out of shortcode processing in a field, like HTML Sanitizing. This is how I fixed it for my client really quickly, but I think this is just a bandaid for the bigger problem above.

![shortcode-settings](https://user-images.githubusercontent.com/96801/99984606-46fac900-2d7b-11eb-9886-2463dc07661e.png)

deanmarcussen:
For infos @davidhayden 

The reason we (I) integrated shortcodes, as a non optional feature, was primarily because they provide a default behaviour to include media files into content.

i.e. without them you cannot use media from the media library inside any html content blocks.

To be fair, when we first implemented all we had was the `[image]` tag, which wouldn't mangle other `[` ...

davidhayden:
Maybe something changed or I'm experiencing something strange @deanmarcussen , because attributes in all sample code are completely swallowed up by the shortcode processor. If I have an attribute in sample code in an HTML Field, like `[HttpGet]`, the shortcode processor removes it completely from the code. In order to display it, I have to escape it as `[[HttpGet]]`, which I shouldn't have to do.

[...]

PR #7988, opened 2020-12-14
Upgrade Shortcodes to 1.0.0

### LINKED ISSUE #7986: Shortcodes 1.0.0
Update Shortcodes to version 1.0.0:
https://www.nuget.org/packages/Shortcodes

hishamco:
Fixes #7986 #7715

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Shortcodes/Providers/LocaleShortcodeProvider.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Shortcodes/Providers/LocaleShortcodeProvider.cs
@@ -9,7 +9,7 @@ namespace OrchardCore.Shortcodes.Providers
     {
         public const string ShortCodeIdentifier = "locale";
 
-        private static readonly ValueTask<string> Null = new ValueTask<string>((string)null);
+        private static readonly ValueTask<string> Null = new ValueTask<string>(String.Empty);
 
         public ValueTask<string> EvaluateAsync(string identifier, Arguments arguments, string content, Context context)
         {
@@ -21,7 +21,7 @@ namespace OrchardCore.Shortcodes.Providers
             var language = arguments.NamedOrDefault("lang")?.ToLower();
             var argFallback = arguments.NamedOrAt("fallback", 1);
             // default value of true for the fallback argument
-            var fallback = argFallback == null ? true : Convert.ToBoolean(argFallback); ;
+            var fallback = argFallback == null ? true : Convert.ToBoolean(argFallback);
             var currentCulture = CultureInfo.CurrentUICulture;
 
             if (fallback)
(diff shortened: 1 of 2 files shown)


agriffard:
Not so simple, the tests will fail.

hishamco:
> Not so simple, the tests will fail.

No worry I will fix them

hishamco:
Seems the only test that fails is locale shortcode, so I need to figure out if it's a shortcodes bug after the recent commits, because it was working before

hishamco:
Everything is working fine in Shortcodes, the issue occurs from `LocaleShortcodeProvider.Null`, problem solved

Skrypt:
hmm yeah Null is probably a reserved word.

src/OrchardCore.Modules/OrchardCore.Shortcodes/Providers/LocaleShortcodeProvider.cs
```diff
@@ -9,19 +9,19 @@ public class LocaleShortcodeProvider : IShortcodeProvider
...
 
-        private static readonly ValueTask<string> Null = new ValueTask<string>((string)null);
+        private static readonly ValueTask<string> Empty = new ValueTask<string>(String.Empty);
 
         public ValueTask<string> EvaluateAsync(string identifier, Arguments arguments, string content, Context context)
         {
             if (identifier != ShortCodeIdentifier)
             {
-                return Null;
+                return Empty;
```
deanmarcussen, line 18: This should return Null as before

src/OrchardCore.Modules/OrchardCore.Shortcodes/Providers/LocaleShortcodeProvider.cs
```diff
@@ -46,7 +46,7 @@ public ValueTask<string> EvaluateAsync(string identifier, Arguments arguments, s
                 }
             }
 
-            return Null;
+            return Empty;
```
deanmarcussen, line 49: This should return Empty

deanmarcussen (commented):
The change in behaviour with shortcodes is to render the `[blah]` if it is not registered.

I assume (hope) that means somewhere it's evaluated against null for not registered, and empty for is registered, but didn't return a value.

Can you try with the suggested change to see if it passes, and if not we might have to look at shortcodes.

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/7715

src/OrchardCore.Modules/OrchardCore.Shortcodes/Providers/LocaleShortcodeProvider.cs
hishamco: Dean if you look before my last commit is was `Null` the issue comes from `null` value

deanmarcussen: I checked shortcodes. This is wrong.

Please apply the suggestions, return Null initially, and Empty if it has no value.

hishamco: Let me try your suggestions

deanmarcussen (approved):

PR #8004, opened 2020-12-15
Darkmode could not be disabled

Skrypt:


hishamco:
> Darkmode could not be disabled

Is there a way to disable one of the available modes?

deanmarcussen (approved):

Skrypt:
You mean disable one of the theme? When you disable the darkmode it will always only use the "default" theme. 

The issue here was that it was not taking in consideration if the setting was enabled/disabled.

PR #8015, opened 2020-12-15
Tenants UI

agriffard:
- Badge state: icon + text
- Buttons icon with tooltips
- Open external link
- Filter on state click
- Display url host (can have multiple domains)
- Database and Recipe badges
- Vertically align description with title

image


hishamco:
> Bigger state icon

It was bigger at the begging ;)

hishamco (approved):

Skrypt:
Try the filters you will see that there is something to fix.

deanmarcussen:
I don't think we should be using dots w/colors to indicate things.

This applies more to the admin listing, but the patterns should be similar everywhere.

I think we should be using coloured icons to represent states. 

It's gone a little too minimilist to get good quick readability.

So suggestion is same colors, apply an icon to the color scheme.

Example from devops. Ticks and Crosses

Screenshot 2020-12-16 at 09 30 17

Second suggestion. The buttons are quite small. A little bigger maybe?

agriffard:
image

agriffard:
image

deanmarcussen:
Prefer the tags w/badges 👍 

(it helps keep the ui extensible easier as well)

Think the Open action was better on the right 

Maybe not bold text is better for the actions `Running`, not sure.

Whats the lowercase a mean?

Skrypt:
lowercase a is the tenant description written quickly like asdf

Skrypt:
Here is one thing that should be fixed and it is the alignment of the description of the tenant with the tenant name. Maybe this could be fixed in this PR with a solution that could be also duplicated in all lists using where we have the same issue.

agriffard:
4px margin added.

Skrypt:
Sans titre

agriffard:
Ok for the margin.
Why don't you have a disable icon on the running tenant?

Skrypt:
I think it's by design no?

Skrypt:
Ok I like it

Skrypt:
Maybe some work on the badges colors.

Skrypt:
The disable tenant button requires a confirm modal.

scleaver:
All list UIs aren't moving this "icon" based direction are they? Personally I am not a fan of too many icons for actions as they require interpretation by the user. 

If so can you apply the icons via a class applied to the button rather than inserting the icon markup into the button itself... this way those that want to can remove the icons via CSS rather than having to use a view in our theme to override 

Also for a saas installation are you really going to have different tenants on different database types?

Skrypt:
I kind of agree that icons only are quite small compared with buttons.

Skrypt:
Here I exagerated a little :  

image

Skrypt:
And here I think it needs some adjustment for mobile : 

image

Skrypt:
Sans titre

This is why a margin-left is not enough. It needs flexboxes.

agriffard:
@scleaver Icons have a Bootstrap tooltip with the same text they used to have.
I will only change the content items and the tenants for the moment.

agriffard:
2 columns as the content items:

image

src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Index.cshtml
```diff
@@ -129,67 +129,74 @@
...
-                            {
-                                <a asp-action="Enable" asp-route-id="@entry.Name" class="btn text-success btn-sm" itemprop="UnsafeUrl" data-toggle="tooltip" title="@T["Enable"]"><i class="fa fa-toggle-on"></i></a>
-                            }
-                            @if (entry.ShellSettings.State == TenantState.Uninitialized)
-                            {
-                                <a class="btn text-info btn-sm" id="btn-setup-@entry.Name" href="@GetEncodedUrl(entry, originalPathBase)" data-toggle="tooltip" title="@T["Setup"]"><i class="fa fa-cog"></i></a>
-                            }
-                            <a asp-action="Reload" asp-route-id="@entry.Name" class="btn text-secondary btn-sm" itemprop="UnsafeUrl" data-toggle="tooltip" title="@T["Reload"]"><i class="fa fa-redo"></i></a>
+                    <div class="row">
+                        <div class="col-xl-6 col-sm-12 title">
```
Skrypt, line 133: col-lg-6 would be sufficient

Skrypt: I tried quickly with Flexboxes and I've not found a perfect solution with it yet. Quicker with a BS grid.

Skrypt:
Fixed margin with Flexbox
Added text breaking if there is anyone using long tenant names
Merged latest dev and fixed merge issue.

image

PR #8023, opened 2020-12-16
Workflow Atomicity

jtkech:
By using `IDistributedLock`, whose default implementation uses local semaphores, this per WF type / instance

- Configurable per workflow type through a lock timeout and an auto expiration in milliseconds, **by default there are equal to 0 and in that case it works as before without any locking**.

- Prevents scope recursions per workflow type id when starting new WF instances, and prevents scope recursion per WF instance id when resuming WF instances.

- Acquire and use a lock (if both settings are > 0) per workflow type id when starting new WF instances, Acquire and use a lock (if both settings of the related WF type are > 0) per WF instance id when resuming WF instances.

    - This is when we create a new WF instance that we pass to it the 2 lock settings of its related WF type.

    - After a lock is acquired we need to check some current states to see if a WF can be still started / resumed, so here we do a session flush when a WF instance is persisted after having been started / resumed.

---

So i did some tests for a WF Singleton where a WF instance can't be started if a current one of the the same WF type is not yet completed.

- I first tried this Workflow marked as a **Singleton** and no Locking Times (= 0), so without any locking

    - Http event as a **Starting activity**

    - Then a blocking activity waiting for any event

    With apache i did 10 concurrent requests triggering this WF, the expected result was to start one WF instance (on the first request) until the blocking activity, and to not start other instances, **because it is a Singleton** and there is a current instance that is not completed yet. But in this test, wothout any locking, **7 instances was created**.

    Then i did the same tests with locking (Lock Times > 0), only **one instance was created**, as expected for a Singleton.

- Then i replaced the blocking activity with a non blocking one (process time = T1), and did the same tests, in both cases 10 WF instances was created, which is the expected result, but by using Locking times (> T1) the whole was taking more time because **the instances was created one after one atomically, like with a job queue**

- There are different scenarios we may want to implement, e.g. in place of starting each instance after each other, **as with a job queue** where all will do the job, we may want to start only one instance, and just reject the other requests in place of letting them waiting while the current instance is not completed, **more like a debouncer**.

    This is already possible by using a Lock expiration >> T1 (the WF process time), and by using a Lock tiemout << T1. In that case other requests fail to acquire a lock before the timeout, and they do nothing as they were rejected.


sfmskywalker (approved):

deanmarcussen:
Are you ready with this @jtkech ?

jtkech:
@deanmarcussen Yes i am

src/OrchardCore.Modules/OrchardCore.Workflows/Services/WorkflowManager.cs
```diff
@@ -105,57 +113,79 @@ public Task<ActivityContext> CreateActivityExecutionContextAsync(ActivityRecord
...
-
-            // If no workflow matches the event, do nothing.
-            if (!workflowTypesToStart.Any() && !haltedWorkflows.Any())
+            foreach (var workflow in haltedWorkflows)
             {
-                return;
+                // Don't allow scope recursion per workflow instance id.
+                if (_recursions.TryGetValue(workflow.WorkflowId, out var count) && count > 0)
+                {
+                    continue;
```
deanmarcussen, line 129: Should we log some of this? Even at `Information` or `Debug` as you prefer?

src/OrchardCore/OrchardCore.Workflows.Abstractions/Models/Workflow.cs
```diff
@@ -31,12 +31,27 @@ public class Workflow
         public WorkflowStatus Status { get; set; }
         public string FaultMessage { get; set; }
 
+        /// <summary>
+        /// The timeout in milliseconds to acquire a lock before executing this workflow instance.
+        /// </summary>
+        public int LockTimeout { get; set; }
```
deanmarcussen, line 37: My initial thoughts was

is this only if running distributed?
In which case we should only see it if we are running distributed.
But then I see that it is also about atomicity when not distributed from your initial comments above.

Am I correct there @jtkech ?

In which case we just need to document (another pr, I know you don't write docs ;) what it all means for both an ordinary single instance site, and / or a distributed site.

src/OrchardCore.Modules/OrchardCore.Workflows/Services/WorkflowManager.cs
jtkech: Yes, i hesitated, a WF listens to events that may be triggered by other WFs (through handlers), but in a given scope, it is not listening to its own events by design, like a message bus that just ignore a message sent by yourself.

Anyway, i will add it as it may be useful

src/OrchardCore/OrchardCore.Workflows.Abstractions/Models/Workflow.cs
jtkech: @deanmarcussen 

> is this only if running distributed?

No, also if the current distributed lock implementation is a local lock and / or if there is only one running instance, this because concurrent requests / WF events may still happen, **but only if both locking times are > 0**.

Note: In my PR related to background tasks atomicity, it's not the same because for a given running instance a background task can't be executed concurrently

> Am I correct there @jtkech ?

Yes, it's not exclusive to a distributed env, it just uses the current distributed lock implementation that may be a distributed / local lock.

> In which case we just need to document (another pr, I know you don't write docs ;) what it all means for both an ordinary single instance site, and / or a distributed site.

;)

jtkech:
@sebastienros 

- Yes i agree, if a workflow type is a `Singleton`, it should be a singleton without having to input locking times

    **I did this to allow to work without any locking as before, as i know you prefer to not use locking as it is possible ;)**

    But i will update the PR as described below

- **Workflow Type Singleton**: Only check the property as before

    Then we will use default lock timeout and expiration of 30s when **starting** a new instance of this workflow type

- **Atomic Workflow Instance**: A new checkbox

    And a dropdown (tied to the checkbox) to show the locking timeout and expiration inputs

    Then we will use these values when **resuming** a given workflow instance

jtkech:
@sebastienros just for info, here the context and in the next section what i opted to in my last commit

- **Starting a WF**: Yes, we need to do it atomically for a WF singleton, also when the starting activity is a Timer event, as it is hydrated every minutes and we don't want to create a new instance each minute, we already support it as an exclusive event and there is no atomicity concern on a single node, but on a distributed environment it may happen concurrently.

    In the 2 above cases, should we use default locking time values (e.g. 20s), or configurable ones e.g. through the UI?

- **Resuming a WF**: I think it's good to resume any WF instance atomically

    But maybe good to still allow to not use any locking, e.g. when resuming a WF atomically, after the lock is acquired we do an extra query to see if the WF is still correlated and still halted on the right activity.

    So does resuming a WF instance atomically should be a configurable option?

    And here also, should we use default locking time values (e.g. 20s), or configurable ones e.g. through the UI?

- Finally, if configurable, it makes sense to use the same locking times for both **starting** and **resuming** a WF instance

---

**For now, in my last commit i opted for the following** (we may change it on request in another PR)

- We always start a WF Singleton instance (or having an exclusive event as a starting activity) atomically, and by using default locking times of 20s

- We always resume any WF instance atomically, and by using default locking times of 20s

    Note: We use a lock per WF instance, so 2 different instances of the same type can still be executed at the same time

- I removed the configurable locking times from the UI, so it is as before

- So we lose the ability to configure a low timeout to just reject extra events without letting them waiting, but on another side when a event will execute it will see the WF state persisted by the previous one, so normally it will do nothing unless the WF contains 2 consecutive events of the same type, so it may have waited for nothing, but it doesn't bother me

    And we lose the ability e.g. to resume a WF instance without any locking as before, but it doesn't bother me

jtkech:
@sebastienros i changed my mind a little

-  So, as said above, we need locking times to resume any WF instance, and also to start a new WF instance if its type is a Singleton or if it has an exclusive starting event (e.g. Timer), and would be good to be able to configure them

- So i re-intoduced the ability to configure the locking times while editing a WF type, but in a different TAB and with a default value of 10_000 ms if not updated, also if no value is provided e.g. by existing WF types / instances

Here the properties as before in the first TAB

Edit WF 1

Here the locking times in an Advanced TAB, with a default value of 10_000 ms that is also used when starting / resuming a WF instance if no value is provided, e.g. by existing WF types that was not re-edited and by their existing WF instances.

Edit WF 2

jtkech:
@sebastienros Last change as i think it is the best compromise

- When **starting** a WF **Singleton** instance we always use a lock and always with locking times of 20_000 ms, so it fixes the original "issue" as you were suggesting, and it prevents to re-introduce the issue by inputing low locking times.

- Then there are still 2 configurable locking times in a separate TAB, but only used for **resuming** a given WF instance of this type, and with a default value of 0 meaning to not use any locking, so existing WFs are resumed as before.

WF2

PR #8024, opened 2020-12-16
Allow ShapeMetadata access in liquid

### LINKED ISSUE #8021: Unable to build liquid Form_Wrapper
It seems it's not possible to override `Form_Wrapper` with Liquid.

The default Razor template is :
```razor
...
<form id="@formId" action="@formPart.Action" method="@formPart.Method"  enctype="@encType" class="form-content @formContentTypeClassName">
    @if (formPart.EnableAntiForgeryToken)
    {
        @Html.AntiForgeryToken()
    }

</form>
```

So, the Liquid equivalent should be :
```liquid
...
<form action="{{ formPart.Action }}" method="{{ formPart.Method }}" enctype="{{ encType }}" class="form-content {{ formContentTypeClassName }}">
    {% if formPart.EnableAntiForgeryToken %}
        {% antiforgerytoken %}
    {% endif %}

  {{ Model.Metadata.ChildContent | shape_render }}
</form>
```

But `Model.Metadata.ChildContent` doesn't evalutes in Liquid, and it's not `IShape` but `IHtmlContent`.

TFleury:
and allow to render IHtmlContent with shape_render filter

Fixes #8021

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/LiquidViewFilters.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/LiquidViewFilters.cs
@@ -124,7 +124,8 @@ namespace OrchardCore.DisplayManagement.Liquid.Filters
                 return new HtmlContentValue(await task);
             }
 
-            if (input.ToObjectValue() is IShape shape)
+            var shape = input.ToObjectValue();
+            if (shape != null)
             {
                 if (!context.AmbientValues.TryGetValue("DisplayHelper", out var item) || !(item is IDisplayHelper displayHelper))
                 {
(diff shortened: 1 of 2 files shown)


src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/LiquidViewFilters.cs
```diff
@@ -124,7 +124,8 @@ static async ValueTask<FluidValue> Awaited(Task<IHtmlContent> task)
                 return new HtmlContentValue(await task);
             }
 
-            if (input.ToObjectValue() is IShape shape)
+            var shape = input.ToObjectValue();
+            if (shape != null)
```
jtkech, line 128: I would suggest to check if it is an `IShape` or an `IHtmlContent`

But i saw that it is checked in `IHtmlContent CoerceHtmlString(object value)`

And if it is another object type, we return `StringHtmlContent(value.ToString())`

So that's okay ;)

jtkech (approved):

PR #8040, opened 2020-12-17
Content type badges

agriffard:
Use badges in Content Type: Display mode, Editor.
Collapsable Options (Creatable, ...), Indexing (Full-text).
Move Shortcuts at bottom.

image

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentTypes/Views/Admin/Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.ContentTypes/Views/Admin/Edit.cshtml
@@ -29,13 +29,6 @@
 
     <input type="hidden" asp-for="Name" />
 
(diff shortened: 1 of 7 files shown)


agriffard:
The goal was to have a more condensed UI, to focus on Fields and Parts which are the most common tasks.

@sebastienros was confused by this new UI.

Should we revert the changes?

hishamco:
IMHO the new UI in general is better than the old

agriffard:
The Creatable, Listable, ... badges may be confusing because they are not linked to checkbox state and they are a repetition of the checkboxes.

My goal was to display them as a summary so you don't have to open the Options.

deanmarcussen:
Yes this one should be reverted.

Suggestion of tabs for a future rather than the drop downs sounded better

agriffard:
ok, reverted.

PR #8041, opened 2020-12-17
Admin Pager icons

agriffard:
Before:

image

After:

image

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsAdminList.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsAdminList.cshtml
@@ -44,7 +44,12 @@
         </li>
     }
 </ul>
-
+@{
+    Model.Pager.FirstText = T["<i class='fa fa-xs fa-chevron-left'></i><i class='fa fa-xs fa-chevron-left'></i>"];
+    Model.Pager.PreviousText = T["<i class='fa fa-xs fa-chevron-left'></i>"];
+    Model.Pager.NextText = T["<i class='fa fa-xs fa-chevron-right'></i>"];
+    Model.Pager.LastText = T["<i class='fa fa-xs fa-chevron-right'></i><i class='fa fa-xs fa-chevron-right'></i>"];
+}
 @await DisplayAsync(Model.Pager)
 
 <script at="Foot">


Skrypt:
Maybe add this directly to the PagerShape?

agriffard:
The default texts are <<, <, >, >>.
That would be specific to the admin.
Font awesome cannot be the defaults.

Skrypt:
Yes but then you need to add these on every PagerShape in every view.

agriffard:
Please show me how.

Skrypt:
I'm trying something

src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsAdminList.cshtml
```diff
@@ -44,7 +44,12 @@
         </li>
     }
 </ul>
-
+@{
+    Model.Pager.FirstText = T["<i class='fa fa-xs fa-chevron-left'></i><i class='fa fa-xs fa-chevron-left'></i>"];
+    Model.Pager.PreviousText = T["<i class='fa fa-xs fa-chevron-left'></i>"];
```
hishamco, line 49: Why they are localized?!!

Skrypt: Because the PagerShape requires an object to be passed on these. Not a string. So a StringLocalizer object.
When we try to pass an icon here it works but we should not do this.

Skrypt: We should probably have Model.Pager.FirstIcon ...

Skrypt: I'm looking at something about the CoerceHtmlString method in that PagerSahpes

Skrypt: So basically T here is a IViewLocalizer which returns an IHtmlContent "I think". This is why it works from the view when doing this. Because it pass in the CoerceHtmlString method and returns a proper IHtmlContent result. Else if he would pass a string, it would return a StringHtmlContent(value.ToString())

Skrypt: Like this : 

image

hishamco:
I'm not sure what's the advantage to use font awesome icons while we can do this using string

<<
<
>
>>

Skrypt:
Icons are prettier?

agriffard:
See #8047

Issue #7990, opened 2020-12-14
Resources CDN replaces inline script src


deanmarcussen:
When using an external CDN for locally hosted scripts the `CdnBaseUrl` property replaces the inline script with the url src of the cdn.

What should be rendered as 
<script>
... some javascript
</script>

becomes
<script src="https://mycdn.azureedge.net"</script>

jtkech:
Oops ;)

PR #8042, opened 2020-12-17
Correctly generate cdn base url for local scripts

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/7990

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ResourceUrlFilter.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ResourceUrlFilter.cs
@@ -40,7 +40,8 @@ namespace OrchardCore.DisplayManagement.Liquid.Filters
                 // Don't evaluate with Uri.TryCreate as it produces incorrect results on Linux.
                 !resourcePath.StartsWith("https://", StringComparison.OrdinalIgnoreCase) &&
                 !resourcePath.StartsWith("http://", StringComparison.OrdinalIgnoreCase) &&
-                !resourcePath.StartsWith("//", StringComparison.OrdinalIgnoreCase))
+                !resourcePath.StartsWith("//", StringComparison.OrdinalIgnoreCase) &&
+                !resourcePath.StartsWith("file://", StringComparison.OrdinalIgnoreCase))
             {
                 resourcePath = _options.CdnBaseUrl + resourcePath;
             }
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ResourceUrlFilter.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ResourceUrlFilter.cs
@@ -36,12 +36,11 @@ namespace OrchardCore.DisplayManagement.Liquid.Filters
             }
 
             // Don't prefix cdn if the path includes a protocol, i.e. is an external url, or is in debug mode.
-            if (!_options.DebugMode
-                && !String.IsNullOrEmpty(_options.CdnBaseUrl)
+            if (!_options.DebugMode && !String.IsNullOrEmpty(_options.CdnBaseUrl) &&
                 // Don't evaluate with Uri.TryCreate as it produces incorrect results on Linux.
-                && !resourcePath.StartsWith("https://", StringComparison.OrdinalIgnoreCase)
-                && !resourcePath.StartsWith("http://", StringComparison.OrdinalIgnoreCase)
-                && !resourcePath.StartsWith("//", StringComparison.OrdinalIgnoreCase))
+                !resourcePath.StartsWith("https://", StringComparison.OrdinalIgnoreCase) &&
+                !resourcePath.StartsWith("http://", StringComparison.OrdinalIgnoreCase) &&
+                !resourcePath.StartsWith("//", StringComparison.OrdinalIgnoreCase))
             {
                 resourcePath = _options.CdnBaseUrl + resourcePath;
             }
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ResourceUrlFilter.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ResourceUrlFilter.cs
@@ -38,6 +38,7 @@ namespace OrchardCore.DisplayManagement.Liquid.Filters
             // Don't prefix cdn if the path includes a protocol, i.e. is an external url, or is in debug mode.
             if (!_options.DebugMode
                 && !String.IsNullOrEmpty(_options.CdnBaseUrl)
+                // Don't evaluate with Uri.TryCreate as it produces incorrect results on Linux.
                 && !resourcePath.StartsWith("https://", StringComparison.OrdinalIgnoreCase)
                 && !resourcePath.StartsWith("http://", StringComparison.OrdinalIgnoreCase)
                 && !resourcePath.StartsWith("//", StringComparison.OrdinalIgnoreCase))
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ResourceUrlFilter.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ResourceUrlFilter.cs
@@ -35,10 +35,10 @@ namespace OrchardCore.DisplayManagement.Liquid.Filters
                 resourcePath = _httpContextAccessor.HttpContext.Request.PathBase.Add(resourcePath.Substring(1)).Value;
             }
 
-            // Don't prefix cdn if the path is absolute, or is in debug mode.
+            // Don't prefix cdn if the path includes a protocol, i.e. is an external url, or is in debug mode.
             if (!_options.DebugMode
                 && !String.IsNullOrEmpty(_options.CdnBaseUrl)
-                && !Uri.TryCreate(resourcePath, UriKind.Absolute, out var uri))
+                && !resourcePath.Contains("//", StringComparison.OrdinalIgnoreCase))
             {
                 resourcePath = _options.CdnBaseUrl + resourcePath;
             }
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ResourceUrlFilter.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ResourceUrlFilter.cs
@@ -38,7 +38,9 @@ namespace OrchardCore.DisplayManagement.Liquid.Filters
             // Don't prefix cdn if the path includes a protocol, i.e. is an external url, or is in debug mode.
             if (!_options.DebugMode
                 && !String.IsNullOrEmpty(_options.CdnBaseUrl)
-                && !resourcePath.Contains("//", StringComparison.OrdinalIgnoreCase))
+                && !resourcePath.StartsWith("https://", StringComparison.OrdinalIgnoreCase)
+                && !resourcePath.StartsWith("http://", StringComparison.OrdinalIgnoreCase)
+                && !resourcePath.StartsWith("//", StringComparison.OrdinalIgnoreCase))
             {
                 resourcePath = _options.CdnBaseUrl + resourcePath;
             }
--- a/test/OrchardCore.Tests/ResourceManagement/ResourceDefinitionTests.cs
+++ b/test/OrchardCore.Tests/ResourceManagement/ResourceDefinitionTests.cs
@@ -37,6 +37,22 @@ namespace OrchardCore.Tests.ResourceManagement
         }
 
         [Theory]
+        [InlineData("/base")]
+        [InlineData("https://base.com")]
+        public void GetScriptResourceWithBasePath(string basePath)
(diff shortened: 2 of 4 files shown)


src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ResourceUrlFilter.cs
```diff
@@ -35,10 +35,10 @@ public ValueTask<FluidValue> ProcessAsync(FluidValue input, FilterArguments argu
                 resourcePath = _httpContextAccessor.HttpContext.Request.PathBase.Add(resourcePath.Substring(1)).Value;
             }
 
-            // Don't prefix cdn if the path is absolute, or is in debug mode.
+            // Don't prefix cdn if the path includes a protocol, i.e. is an external url, or is in debug mode.
             if (!_options.DebugMode
                 && !String.IsNullOrEmpty(_options.CdnBaseUrl)
-                && !Uri.TryCreate(resourcePath, UriKind.Absolute, out var uri))
+                && !resourcePath.Contains("//", StringComparison.OrdinalIgnoreCase))
```
sebastienros, line 41: `Contains` or `Startswith`

sebastienros: And even if you mean `https://` I'd rather check `Starswith("http://") || Startswith("https://")`

deanmarcussen: Was allowing for 
- `https://`
- `http://` or
- `//` i.e. protocol agnostic

if you prefer, can do `StartsWith` for all 3

sebastienros: Please, I fear unknown consequences if a url had a `//` in the middle. Beginning with the most obvious, `https://`

sebastienros: You can make a new local private method to make it simpler

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ResourceUrlFilter.cs
```diff
@@ -35,10 +35,12 @@ public ValueTask<FluidValue> ProcessAsync(FluidValue input, FilterArguments argu
...
             }
 
-            // Don't prefix cdn if the path is absolute, or is in debug mode.
+            // Don't prefix cdn if the path includes a protocol, i.e. is an external url, or is in debug mode.
             if (!_options.DebugMode
                 && !String.IsNullOrEmpty(_options.CdnBaseUrl)
-                && !Uri.TryCreate(resourcePath, UriKind.Absolute, out var uri))
+                && !resourcePath.StartsWith("https://", StringComparison.OrdinalIgnoreCase)
+                && !resourcePath.StartsWith("http://", StringComparison.OrdinalIgnoreCase)
+                && !resourcePath.StartsWith("//", StringComparison.OrdinalIgnoreCase))
```
jtkech, line 43: @deanmarcussen 

Just tried, it seems that `System.Uri.TryCreate()` does the same checkings

It returns true for `System.Uri.TryCreate("https://toto.com")`

And it also returns true for `System.Uri.TryCreate("//toto")`

deanmarcussen: @jtkech but it also returns true for `/foo.js` 

That was the problem :(

Unless there is a better way? I tried with `IsWellFormed` as well, but wasn't correct.

jtkech: @deanmarcussen No, just tried, return false in that case ;)

deanmarcussen: https://dotnetfiddle.net/t6gb6x

Which `UriKind` @jtkech ?

returns true with UriKind.Absolute for me

jtkech: if (System.Uri.TryCreate("/foo.js", System.UriKind.Absolute, out var uri))
            {
                ; // <= Doesn't go there
            }

jtkech: Yes, under fiddle it creates `file:///foo.js` ???

deanmarcussen: Hmm. Interesting. 

`file:///foo.js`
 
That's what I get on the mac as well. 

Did we find a bug?

hishamco, line 43: What about `file://`?

jtkech: Yes maybe ;)

And your PR would be a good workaround

I'll take a look at the source code

hishamco: Seems @jtkech mention that above ;)

jtkech: @deanmarcussen 

Okay it's because, e.g. some grabbed comments

> A relative Uri wins over implicit Unix path unless uriKind == Absolute

> In .NET new Uri ("/foo", UriKind.RelativeOrAbsolute) returns a relative Uri whereas Mono assumes it is an absolute file path.

> Mono diverges from .NET because UNIX file paths starting with ‘/’ are absolute file paths.

So your fix is right

jtkech (approved):

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ResourceUrlFilter.cs
deanmarcussen: ok, interesting, and of course that makes sense.

There's tests now on these, so it should fail on the CI if we tried to change it.

test/OrchardCore.Tests/ResourceManagement/ResourceDefinitionTests.cs
```diff
@@ -10,157 +11,223 @@ namespace OrchardCore.Tests.ResourceManagement
...
         {
             _resourceManifest = new ResourceManifest();
         }
 
-        [Fact]
-        public void GetScriptResourceWithUrl()
+        // The application path is empty if it is the host site
+        // or contains the base path of a tenant.
+        [Theory]
+        [InlineData("")]
```
deanmarcussen, line 24: @jtkech i added the correct values for the application path here (base path)
but while I am in this piece of code what is the correct value for a virtual path? 
and I will add it as well.

leasing slash? Or without?

jtkech: /virtual/tenant

deanmarcussen: @jtkech I updated this to add a test for the `WithBasePath` / `SetBasePath`

I think that feature, despite having some notation of being a virtualPath is disconnected from what we would actually call a virtualPath (i.e. under IIS).

Not sure how useful the `WithBasePath` actually is, but it's there, and works.

jtkech: @deanmarcussen 

That's okay it's just a question of semantics, as you want ;)

About the base path i was thinking about the httpRequest `PathBase` that may contain a virtual path (i.e. under IIS) and or a tenant prefix, the tenant prefix being a kind of virtual path, but at the end the whole being a base path.

So forget the virtual term, just use basePath, and yes better to have

    ResourceDefinition SetBasePath(string basePath)

That said in the tests, for me `/tenant` and `/virtual/tenant` are not `applicationPath`, but `basePath`, and then in `/tenant/scripts/script.js`, `/scripts/script.js` is a path

---

Okay just saw in the `ResourceDefinition` code the usage of `GetTagBuilder(..., string applicationPath, ...)` which is called somewhere by `GetHtmlContent(string appPath)` which is called by `GetHtmlContent(_options.ContentBasePath)`, and where `_options.ContentBasePath` is initialized with the httpRequest `PathBase` ;)

            options.ContentBasePath = _httpContextAccessor.HttpContext.Request.PathBase.Value;

So okay, we could say that `applicationPath` is the same as `basePath`, but here i prefer `basePath`, at least `tenantPath` as an application can have multiple tenants, but each one being a kind of an isolated app ;)

---

`https://base.com` is not a `basePath`, nor an `applicationPath`, but a full **Uri** (Url is okay), and e.g. in the full uri `https://base.com/virtual/tenant/scripts/script.js`, `/virtual/tenant` is a basePath and `/scripts/script.js` is a path

Maybe good to have a test with `/virtual/tenant`, just to do a test with a basePath having 2 segments

But as said it's just a question of terminology, so as you want ;)

deanmarcussen: Yes, I agree @jtkech .

The terminology inside the resource definition is not quite what we are used to. So it was confusing even figuring out what they should be.

I've added `/virtual/tenant` for the two segments (makes sense)

And remove the full uri `https://base.com`. While it would work, I'm just not sure it makes sense.

Equally (and I think this is probably a different issue, but noone has ever raised it so I am ignoring it.)

I'm not sure how a base path would work in combination with a tenant prefix.

But I'm not sure exactly what you would use the `basePath` for anyway.

So I just test that it's behaviour is the **same as it is currently** so if it is being used by anyone we know there is no regression.

jtkech: Cool then, i'm fine with it

Why not checking if it starts with `file:///`, or just `file://`, or just `file:` ;)

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ResourceUrlFilter.cs
```diff
@@ -35,10 +35,13 @@ public ValueTask<FluidValue> ProcessAsync(FluidValue input, FilterArguments argu
                 resourcePath = _httpContextAccessor.HttpContext.Request.PathBase.Add(resourcePath.Substring(1)).Value;
             }
 
-            // Don't prefix cdn if the path is absolute, or is in debug mode.
+            // Don't prefix cdn if the path includes a protocol, i.e. is an external url, or is in debug mode.
             if (!_options.DebugMode
                 && !String.IsNullOrEmpty(_options.CdnBaseUrl)
-                && !Uri.TryCreate(resourcePath, UriKind.Absolute, out var uri))
+                // Don't evaluate with Uri.TryCreate as it produces incorrect results on Linux.
+                && !resourcePath.StartsWith("https://", StringComparison.OrdinalIgnoreCase)
```
jtkech, line 42: Formatting: in the aspnetcore repo they put the `&&` at the end of the upper line

But there is no universal rule ;)

deanmarcussen: Done. I tend to the other for easier readability, but that's just me ;)

test/OrchardCore.Tests/ResourceManagement/ResourceDefinitionTests.cs
deanmarcussen: I don't think doing `file://` is a good idea. That would suggest that using the resource manager to load files was a good idea. Which it's not. Wouldn't work anyway the browser would block it - I would hope ;)

jtkech: @deanmarcussen No problem, that's okay ;)

Just for info, using `file:///C:/Sources/OrchardCore/src/OrchardCore.Cms.Web/App_Data/Sites/Default/Media/about-bg.jpg` works on my machine

File Url

deanmarcussen: Yes, but try loading a script that way, on an actual site ;) 
How would I get `about-bg` to my computer. and in `C:` when I don't even have a `C:`

jtkech: > Wouldn't work anyway the browser would block it - I would hope ;)

Yes, it was just to show that the browser doesn't block an unc path starting with `file://`
It is on my local machine but i thought about something like `file://someserver/somepath`

deanmarcussen: Ok I give in ;)

Added `file://`

jtkech (approved):

PR #8047, opened 2020-12-17
Admin Pager icons (In PagerShapes)

Skrypt:
Related to #8041

Making some progress to see how this could be done from the PagerShape without breaking defaults to text values.
Trying to make this not dependent on anything UI related would be also better.


src/OrchardCore.Modules/OrchardCore.Navigation/PagerShapes.cs
```diff
@@ -170,7 +175,8 @@ public PagerShapes(IStringLocalizer<PagerShapes> localizer)
             object GapText,
             bool ShowNext,
             string ItemTagName,
-            IDictionary<string, string> ItemAttributes
+            IDictionary<string, string> ItemAttributes,
+            bool LinkIcons
```
Skrypt, line 179: `Bool LinkIcons = true` here never worked. But that would have been the appropriate thing to do to have it as default without requiring to add it everywhere.

Issue #8019, opened 2020-12-16
Liquid: support multiple matching TagHelpers


arkadiuszwojcik:
This story resulted from issue #8012 . Currently when using liquid script like this: `{% helper "input" ... %}` underneath implementation call at some point: `LiquidTagHelperFactory.GetActivator("input" ...)` which return first matching TagHelper from registred TagHelpers list, so only one matching TagHelper will be returned. Desired behavior would be to return all matching TagHelper and applay them one by one in respect of `TagHelper.Order` property.

jtkech:
Cool, feel free to open a PR if you want to work on this

PR #8055, opened 2020-12-17
InputIsDisabledTagHelper fix

### LINKED ISSUE #8012: Multiple liquid tag helpers not supported
Today I was experimenting with liqid helpers and tried to make use of `input`. Liquid script I had was:

`{% helper "input", type: "text", value: "some_val", for: "some_field" %}`

I could not make this code work, what was rendered was empty input control: `<input>`

What I found out debuging Orchard source code was that: `LiquidTagHelperFactory.GetActivator("input" ...)` is not returning `InputTagHelper` as I would expect but `InputIsDisabledTagHelper` which is registred earlier. Both tag helpers use same name: `input`.

Edit:

Maybe `LiquidTagHelperFactory.GetActivator("input" ...)` should be prepared to return multiple tag helpers with same name? Such filters should be then sorted by `TagHelper.Order `property.

arkadiuszwojcik:
This change should close #8012 

cc: @jtkech


jtkech:
Cool thanks

I think that for now we can say that it also closes #8019 (not planned for perf reasons, only if requested again)

jtkech (approved):

PR #8058, opened 2020-12-17
EditTemplate buttons stick to top while scrolling

### LINKED ISSUE #1087: Edit Template Editor Suggestions
- Support Ctrl+S to save the template and continue editing. The current workflow is limited.
- Use CodeMirror or Monaco to support syntax highlighting (At least HTML)

hishamco:
Fixes #1087

EditTemplate

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Templates/Views/Template/Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Templates/Views/Template/Edit.cshtml
@@ -51,7 +51,7 @@ else
         <span class="hint">@T["The Liquid template."]</span>
     </div>
 
-    <div class="form-group">
+    <div class="form-group position-sticky action-bar">
         <div class="btn-group">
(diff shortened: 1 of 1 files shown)


deanmarcussen:
This is a good start, and a nice idea, but it needs more thinking.

You can't have buttons that are sometimes in one place, and sometimes in another place.

Please can it either be **reverted** or **reworked** 

My suggestion is the idea of sticky buttons is good. Move them to the right and top permanently. Make them sticky there.

Don't show them in one place and then another.

Don't have them overhang the text. 

Example from devops that works quite well.

Screenshot 2020-12-18 at 09 27 49

/cc @hishamco @agriffard

agriffard:
Reverted.

@hishamco Can you please make a PR with the buttons at the top?

hishamco:
@deanmarcussen I was thinking to move them to the right, but seems @agriffard merged quickly, I will start a new PR

hishamco:
The PR #8067 is created

PR #8067, opened 2020-12-18
Move edit template action buttons to the top right

hishamco:
EditTemplate

/cc @deanmarcussen

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Templates/Views/Template/Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Templates/Views/Template/Edit.cshtml
@@ -19,7 +19,7 @@
 
 @if (Model.AdminTemplates)
 {
-    <zone Name="Title"><h1 class="display-inline float-left">@RenderTitleSegments(T["Edit Admin Template"])</h1></zone>
+    <zone Name="Title"><h1>@RenderTitleSegments(T["Edit Admin Template"])</h1></zone>
 }
 else
 {
--- a/src/OrchardCore.Modules/OrchardCore.Templates/Views/Template/Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Templates/Views/Template/Edit.cshtml
@@ -28,7 +28,7 @@ else
 
 <div asp-validation-summary="ModelOnly"></div>
 <form asp-action="Edit" asp-route-returnUrl="@ViewData["returnUrl"]" method="post">
-    <div class="form-group position-sticky action-bar text-right">
+    <div class="form-group position-sticky action-bar">
         <div class="btn-group">
             <button type="submit" name="submit" class="btn btn-primary save">@T["Save"]</button>
             <button type="button" class="btn btn-primary dropdown-toggle dropdown-toggle-split" data-reference="parent" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
@@ -48,6 +48,7 @@ else
         {
             <a class="btn btn-info" role="button" href="@Url.Action("Index", "Preview", new { area = "OrchardCore.Templates" })" target="_blank">@T["Preview"]</a>
         }
+        <a href="#editContent" class="btn" data-toggle="tooltip" title="@T["Go to content"]"><i class="fa fa-edit"></i> @T["Content"]</a>
     </div>
 
     <input type="hidden" value="@Model.Name" name="sourceName" />
@@ -63,6 +64,7 @@ else
     <div class="form-group" asp-validation-class-for="Description">
         <label asp-for="Description">@T["Description"]</label>
         <input asp-for="Description" class="form-control" autofocus />
+        <div id="editContent">&nbsp;</div>
         <span class="hint">@T["The description of the template."]</span>
     </div>
 
--- a/src/OrchardCore.Modules/OrchardCore.Templates/Views/Template/Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Templates/Views/Template/Edit.cshtml
@@ -28,7 +28,7 @@ else
 
 <div asp-validation-summary="ModelOnly"></div>
 <form asp-action="Edit" asp-route-returnUrl="@ViewData["returnUrl"]" method="post">
-    <div class="form-group position-sticky action-bar float-right">
+    <div class="form-group position-sticky action-bar text-right">
         <div class="btn-group">
             <button type="submit" name="submit" class="btn btn-primary save">@T["Save"]</button>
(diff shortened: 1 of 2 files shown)


deanmarcussen:
Looking much better @hishamco 

But it shouldn't float on top of things. It needs to be like an action bar. Like you did for the lists.

Try turning the title bar setting on as well, and then we can see it with the title in the breadcrumb / nav area, and the buttons below. (and that saves the cost of screen real estate wasted by the title, but costing with adding a bar)

So suggest to move the name below the button bar.

102608487-005e7b80-413b-11eb-8d88-f53e1a7c5926

This will make it work better for any of the views, that I imagine you will want to put it on later.

hishamco:
The issue comes from `float-right` I'm trying to fix it now

jptissot:
Should it be at the same level as the title ? like the title and buttons could be  an action bar ?

src/OrchardCore.Modules/OrchardCore.Templates/Views/Template/Edit.cshtml
```diff
@@ -48,6 +48,7 @@ else
         {
             <a class="btn btn-info" role="button" href="@Url.Action("Index", "Preview", new { area = "OrchardCore.Templates" })" target="_blank">@T["Preview"]</a>
         }
+        <a href="#editContent" class="btn" data-toggle="tooltip" title="@T["Go to content"]"><i class="fa fa-edit"></i> @T["Content"]</a>
```
hishamco, line 51: Why?

agriffard:
Button to go to the content:

media

hishamco:
it not looking good, seems align the bottons with the title make sense, but I need to check how it looks on other pages

agriffard:
> Should it be at the same level as the title ? like the title and buttons could be an action bar ?

> it not looking good, seems align the bottons with the title make sense, but I need to check how it looks on other pages

The title is in another zone. Remove the zone if needed.

hishamco:
> it not looking good, seems align the bottons with the title make sense, but I need to check how it looks on other pages

I think no need to do that, the content items page has similar look and feel except it has a search bar, so I think we need to revert the last commit

agriffard:
> > it not looking good, seems align the bottons with the title make sense, but I need to check how it looks on other pages
> 
> I think no need to do that, the content items page has similar look and feel except it has a search bar, so I think we need to revert the last commit

Revert if you want. Sorry, I liked the result so I pushed it.

hishamco:
Edit Template

Now everything looks good

deanmarcussen (changes requested):
I personally quite like it.

However it represents a significant change to how we display and place buttons in edit screens (and I know you will want to do it everywhere).

Please do not merge until demo'd on Tuesday (either AG or I will demo).

agriffard:
I think this is the only Edit screen where we need sticky buttons, because of the possible long height of the code mirror area.

It also reminded me that I did a PR a long time ago about a SideBar in which you can put some Actions (Buttons) and Meta (Created, Owner, State).
With BS 9-3 or 8-4 columns, on large screens, the sidebar is on the right and on small screens, at the end.

hishamco:
> I think this is the only Edit screen where we need sticky buttons,

Also list pages that expect a long entries such as Content Items, Features .. etc

Skrypt:
image

This doesn't look right to me.
Consider creating a top sticky bar for buttons instead. It will always be there, just like the search bar is in content list. Just keep it sticky there at all times.

image

hishamco:
You are right @Skrypt, I already did that but it looks ugly if you didn't scoll the page, there will be a wider space underneath the title

Skrypt:
Have you tried faking the space by adding a `margin-bottom: -value` ?

hishamco:
Is adding a bar enough to close this PR?

deanmarcussen:
> Is adding a bar enough to close this PR?

It's certainly going to improve its chances when it gets demo'd tonight ;)

hishamco:
This is how it looks like after adding action bar

ActionsBar

src/OrchardCore.Modules/OrchardCore.Templates/Views/Template/Edit.cshtml
```diff
@@ -28,6 +28,29 @@ else
 
 <div asp-validation-summary="ModelOnly"></div>
 <form asp-action="Edit" asp-route-returnUrl="@ViewData["returnUrl"]" method="post">
+    <div class="card mb-3 position-sticky action-bar">
+        <div class="card-body bg-primary p-3 text-right">
```
Skrypt, line 32:
```suggestion
        <div class="card-body bg-primary p-2 text-right">
```

Skrypt: Please add the same thing for the create view.

hishamco: Ok

agriffard:
Did you make the requested changes @hishamco ?

Skrypt:
Not necessary. Actually the size are consistent. It's just an optic illusion because there is less content in that bar. We might add some more content later on.

Skrypt:
The only thing remaining is to do the same thing with the create.cshtml file of that module.

hishamco:
> Did you make the requested changes @hishamco ?

AFAIK, YES!!

agriffard:
Please try to use the same code mirror addons on both pages.

hishamco:
> What you mean by code mirror addons, if the intend to do the same in the Create page, I will do

agriffard:
Look at the styles and scripts at the top of the Edit and apply the same on Create.

hishamco:
Ok, but what they will add to what I did in the PR?

agriffard:
> Ok, but what they will add to what I did in the PR?

Make it identical between Create and Edit.

hishamco:
Ok Antoine

hishamco:
@agriffard Done!!

src/OrchardCore.Modules/OrchardCore.Templates/Views/Template/Edit.cshtml
```diff
@@ -19,7 +19,7 @@
 
 @if (Model.AdminTemplates)
 {
-    <zone Name="Title"><h1>@RenderTitleSegments(T["Edit Admin Template"])</h1></zone>
+    <zone Name="Title"><h1 class="display-inline float-left">@RenderTitleSegments(T["Edit Admin Template"])</h1></zone>
```
deanmarcussen, line 22: Why this for edit, but not create?

I don't think you actually need this at all anymore, as it is all in a card right?

hishamco: Seeems there are not needed anymore, let me double check and confirm

Skrypt (approved):

deanmarcussen (approved):

Issue #6796, opened 2020-07-31
Consider use of IAsyncEnumerable<T> for our services


rserj:
https://docs.microsoft.com/en-us/archive/msdn-magazine/2019/november/csharp-iterating-with-async-enumerables-in-csharp-8
I think it will be beneficial in terms of performance for services like
- IFileStore.GetDirectoryContentAsync
- YesSQL?
- other services where we return Task<IEnumerable<T>>

PR #8073, opened 2020-12-18
Migrate media storage to IAsyncEnumerable

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6796

Migrates the blob storage system to make use of the async enumerable pages we recieve from it.

Unfortunately means a change in the return value of the interfaces, but better to do now than later.

Also brings in `System.Linq.Async` as a direct dependency - it was indirect through other packages, OpenIddict etc.

Ultimately this won't make a massive performance change as any controller will still resolve the `IAsyncEnumerable` to a list before returning it, but it opens the door to using the blob storage system in a more efficient way, particularly if you work with the services directly, rather than through a controller.


kevinchalet:
> Also brings in `System.Linq.Async` as a direct dependency - it was indirect through other packages, OpenIddict etc.

FYI OpenIddict no longer references this package as it causes conflicts when used with EF Core >= 3: https://github.com/openiddict/openiddict-core/issues/980 (we only use it in the tests projects).

deanmarcussen:
> FYI OpenIddict no longer references this package as it causes conflicts when used with EF Core >= 3: [openiddict/openiddict-core#980](https://github.com/openiddict/openiddict-core/issues/980) (we only use it in the tests projects).

Interesting thanks @kevinchalet had not caught up on that change

sebastienros (approved):

PR #8092, opened 2020-12-19
Differentiator for shape to target specifically in placement

giannik:
else all taxonomy filters have same name and cannot specifically target individual shape.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Taxonomies/Drivers/TaxonomyContentsAdminListDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Taxonomies/Drivers/TaxonomyContentsAdminListDisplayDriver.cs
@@ -100,7 +100,7 @@ namespace OrchardCore.Taxonomies.Drivers
                     })
                     .Location("Actions:40." + position.ToString())
                     .Prefix("Taxonomy" + taxonomy.ContentItemId)
-                    .Differentiator("Taxonomy" + taxonomy.ContentItemId)
+                    .Differentiator(GetDifferentiator(taxonomy))
                 );
 
                 position += 5;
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.Taxonomies/Drivers/TaxonomyContentsAdminListDisplayDriver.cs
```diff
@@ -100,6 +100,7 @@ public override async Task<IDisplayResult> EditAsync(ContentOptionsViewModel mod
                     })
                     .Location("Actions:40." + position.ToString())
                     .Prefix("Taxonomy" + taxonomy.ContentItemId)
+                    .Differentiator("Taxonomy" + taxonomy.ContentItemId)
```
deanmarcussen, line 103: I don't think we should use the `ContentItemId`.

I would suggest the `AliasPart` and normalized as is done in the `MenuShapes`, as I think the alias is on a taxonomy by default. If not Alias, `DisplayText` like the `MenuShapes`.

src/OrchardCore.Modules/OrchardCore.Taxonomies/Drivers/TaxonomyContentsAdminListDisplayDriver.cs
```diff
@@ -156,6 +158,60 @@ private static void PopulateTermEntries(List<FilterTermEntry> termEntries, IEnum
                     PopulateTermEntries(termEntries, children, level + 1);
                 }
             }
+        }
+               private static string GetDifferentiator(ContentItem contentItem)
+        {
+            var identifier = contentItem.Content.AliasPart?.Alias?.ToString() ?? contentItem.DisplayText ?? contentItem.ContentItemId;
+            var encodedIdentifier = EncodeAlternateElement(identifier);
+            var differentiator = FormatName(encodedIdentifier);
```
deanmarcussen, line 166: Just `FormatName` - `EncodeAlternate` is for if we're making shape alternates.

Formatting too ;)

PR #8094, opened 2020-12-19
respect isEnabled flag when authenticating via third party Idp

### LINKED ISSUE #7999: IsEnabled flag for User is not respected when authenticating via third party identity providers
I am not clear if the local account property of IsEnabled should be respected or not when authenticating via a third party provider. Since there is an account linkage I believe it should be and so we ca use this flag for pending accounts to be processed via workflows, etc.

Secondarily, my workaround is to check this in a custom AuthorizationHandler in my Razor Pages and then redirect a use to a 403 page. We should also potentially store the IsEnabled flag as a claim so this property does not need to be retrieved every request from the UserManager (db call). This would then allow users to access it easily in pages for custom views and actions.

MichaelPetrinolis:
closes #7999


MichaelPetrinolis:
no descriptive comments in commit

PR #8095, opened 2020-12-19
Check if user isEnabled when authenticating with external Idp

### LINKED ISSUE #7999: IsEnabled flag for User is not respected when authenticating via third party identity providers

MichaelPetrinolis:
use tempdata to send errors from externallogincallback to login

closes #7999

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Users/Controllers/AccountController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Controllers/AccountController.cs
@@ -381,7 +381,7 @@ namespace OrchardCore.Users.Controllers
             if (remoteError != null)
             {
                 _logger.LogError("Error from external provider: {Error}", remoteError);
-                ModelState.AddModelError("", S["An error occured in external provider"]);
+                ModelState.AddModelError("", S["An error occurred in external provider."]);
                 return RedirectToLogin(returnUrl);
             }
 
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.Users/Controllers/AccountController.cs
```diff
@@ -374,22 +381,24 @@ public async Task<IActionResult> ExternalLoginCallback(string returnUrl = null,
             if (remoteError != null)
             {
                 _logger.LogError("Error from external provider: {Error}", remoteError);
-                return RedirectToAction(nameof(Login));
+                ModelState.AddModelError("", S["An error occured in external provider"]);
```
agriffard, line 384: occured => occurred + ends with `.`

src/OrchardCore.Modules/OrchardCore.Users/Controllers/AccountController.cs
```diff
@@ -374,22 +381,24 @@ public async Task<IActionResult> ExternalLoginCallback(string returnUrl = null,
...
+                return RedirectToLogin(returnUrl);
             }
 
             var info = await _signInManager.GetExternalLoginInfoAsync();
             if (info == null)
             {
                 _logger.LogError("Could not get external login info.");
-                return RedirectToAction(nameof(Login));
+                ModelState.AddModelError("", S["An error occured in external provider"]);
```
agriffard, line 392: occured => occurred + ends with `.`

Skrypt (approved):

PR #8098, opened 2020-12-20
Skrypt/data localization

Skrypt:
Work in progress


src/OrchardCore/OrchardCore.Abstractions/Localization/DataLocalizedString.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public class DataLocalizedString
+    {
+        public DataLocalizedString(string context, string name)
+        {
+            Context = context;
+            Name = name;
+        }
+
+        public string Context { get; }
```
sebastienros, line 11: This should be a `LocalizedString` so that the UI can be displayed in multiple languages

src/OrchardCore.Modules/OrchardCore.Localization/Recipes/TranslationsStep.cs
```diff
@@ -0,0 +1,39 @@
...
+            }
+
+            if (context.Step.Property("DynamicDataTranslations").Value is JObject translations)
+            {
+                foreach (var property in translations.Properties())
+                {
+                    var name = property.Name;
+                    var value = property.Value.ToObject<Translation>();
+
+                    await _translationsManager.UpdateTranslationAsync(name, value);
```
sebastienros, line 34: Is that actually saving the document every time a new entry is added? Shouldn't there be a method that accepts a document that has been mutated with many changes instead?

src/OrchardCore/OrchardCore.Abstractions/Localization/ILocalizationDataProvider.cs
```diff
@@ -0,0 +1,9 @@
+using System.Collections.Generic;
+
+namespace OrchardCore.Localization
+{
+    public interface ILocalizationDataProvider
+    {
+        IEnumerable<DataLocalizedString> GetAllStrings();
```
sebastienros, line 7: This is a bad name. More like GetDescriptors. It's confusing with the GetAllStrings of the data localizer

Skrypt:
@hishamco Moved things added in OrchardCore.Localization to OrchardCore.Localization.Dynamic as a separate feature dependent on OrchardCore.Localization to keep it clean. Also, using dynamic data localization is not necessary a requirement for everyone.

Skrypt:
Here is the issue I reported in @hishamco PR.

This is happening on this PR branch.

orchard

This is what is happening in current dev branch (and how it should be).

orchard

test/OrchardCore.Tests/Localization/PortableObjectStringLocalizerFactoryTests.cs
```diff
@@ -13,9 +13,10 @@ namespace OrchardCore.Tests.Localization
 {
     public class PortableObjectStringLocalizerFactoryTests
     {
-        [Fact]
-        public async Task LocalizerReturnsTranslationFromInnerClass()
```
Skrypt, line 17: Probably related with Date format translation issue.

hishamco: No, this another one, basically to resolve the issue when the translation in inner class

hishamco:
Thanks @Skrypt for reporting this, really I didn't find time to follow up my PR since it was plnned for `1.1`. BTW there was 1 failed test AFAIK

hishamco:
> Also, because using dynamic data localization is not necessary a requirement for everyone.

I think it's because content types, content fields will use it

Skrypt:
> I think it's because content types, content fields will use it

They **could use it** but it's dependent on the feature. So moved it to the feature. Removed dependency on OrchardCore.Localization that was added on OrchardCore.ContentType. It should have a dependency only on OrchardCore.Localization.Dynamic feature instead which makes sense to me.

Skrypt:
@hishamco Moving discussion here.
Yes, this PR should be for OC. Last time I discussed about this with Sebastien it's now been a month.
We did a quick overview of what would be good for the UI and I think there is still some refactoring/renaming needed in the classes / structure.

hishamco:
So, I will closed my PR, as I remembered you made a fork to implemented for one of your websites. Anyhow can I know the final thoughts about data localization features coz I was interested on this from the beginning ;)

Skrypt:
I've not used it on any website yet because I think it needs some work. I'm just side-tracked with everything else 😉

hishamco:
@Skrypt I hcharged my enrgy to continue working on this ;) could you please let me know what else remaining other than refactoring and renaming. or shall I  continue based on our thought on what are mentioned in my previous PR?

Skrypt:
Probably a combination of comments on this PR and yours. First thing I would say is that it needs to work without breaking other localizations which is the main issue I noticed.

Skrypt:
I'm trying to use it but so far I did not find the proper combination or JSON values 😢 

```json
{
  "name": "",
  "displayName": "",
  "description": "",
  "author": "",
  "website": "",
  "version": "",
  "issetuprecipe": false,
  "categories": [],
  "tags": [],
  "steps": [
    {
      "name": "DynamicDataTranslations",
      "DynamicDataTranslations": {
        "Dealer": {
          "Context": "Dealer",
          "Key": "Dealer",
          "Values": {"fr": "Marchand", "en": "Dealer"}
        },
        "MagazineAd": {
          "Context": "MagazineAd",
          "Key": "MagazineAd",
          "Values": {"fr": "Annonce revue", "en": "Magazine Ad"}
        }
      }
    }
  ]
}
```

Here is what I tried in the end to see if I can get it from the CultureDictionary. Will try again tomorrow.
I added a D method in the RazorPage class so that we can use D["Something"]. I did not commit here yet though.
Need to test more before.

PR #8102, opened 2020-12-21
Fix modal semantics

### LINKED ISSUE #7225: HTML Semantics issue : confirm modals
image

On all our confirm modals we need to change this to use a different attribute (maybe data-action) as we are not using itemprop as it should.

/cc @jptissot

Skrypt:
Fixes #7225

`itemprop` attribute replaced by `data-url-af` "af for antiforgery".

JQuery selector has been replaced too, so it requires that everyone updates their current custom module implementations if we merge this.


src/OrchardCore.Themes/TheAdmin/wwwroot/Scripts/TheAdmin.js
```diff
@@ -7136,7 +7136,7 @@ $(function () {
   var magicToken = $("input[name=__RequestVerificationToken]").first();
 
   if (magicToken) {
-    $("body").on("click", "a[itemprop~='UnsafeUrl'], a[data-unsafe-url]", function () {
```
Skrypt, line 7139: Here there was a selector for a[data-unsafe-url] which has been removed because it's not used anywhere.

Skrypt: Fixed

deanmarcussen:
Is it possible to keep the `itemprop` js code for backward compatability and making it non breaking?

With notes that that code can be removed at a later time?

Skrypt:
Done

deanmarcussen (approved):
So I removed the breaking changes, because it's not anymore?

PR #8106, opened 2020-12-21
Admin UI : Fix trumbowyg font-color on fullscreen in darkmode

### LINKED ISSUE #8105: Admin UI : Trumbowyg full screen text color in darkmode
Text color is white in dark mode which makes text invisible.

orchard

Skrypt:
Fixes #8105


src/OrchardCore.Themes/TheAdmin/Assets/scss/modules/theme/darkmode/_bootswatch.scss
```diff
@@ -36,6 +37,10 @@
     }
 }
 
+.trumbowyg-fullscreen .trumbowyg-editor{
```
hishamco, line 40: Why we didn't introduce trumbowyg override to track all the WYSIWYG styles

Skrypt: We don't need to really. It works fine as it is. Don't change what ain't broken they say 😉

Issue #7915, opened 2020-12-09
Security issue: Managing Users permission


MikeKry:
When defining role for user that can manage users (add/edit) it allows also to delete users with administrator role, or to change role of administrator user, even though managing roles / assign roles is not allowed.

1.0.0-rc2-15584

hishamco:
This is already done by @Skrypt since month or two ago in the dev branch

MikeKry:
@hishamco If it is in dev branch, then it should work on my version, I have updated nuggets yesterday.

jtkech:
On nuget this is the master branch

The dev branch is on the cloudsmith feed

deanmarcussen:
This wasn't fixed by @Skrypt recently. @Skrypt fixed another issue.

I opened an issue around this https://github.com/OrchardCMS/OrchardCore/issues/7619

The assign roles permission is currently unused / unchecked so rather useless right now.

The idea being that we need to 
- use the permission
- probably move it to the users module, rather than roles
- update the user role editing view to come from another driver, which can check those permissions

hishamco:
> @hishamco If it is in dev branch, then it should work on my version, I have updated nuggets yesterday. (but it does not work)

Nope, please refer to @jtkech comment

@deanmarcussen I will have a look to the issues that you refer to

Thanks

MikeKry:
@jtkech 

I am using nuget feed from cloudsmith :)

jtkech:
Ah okay, sorry i thought you were using nuget from nuget ;)

hishamco:
Sorry seems we thought you are using a NuGet feed ;)

I will have a look to the refered issue

sebastienros:
Custom permissions (dynamic) to add/remove a specific role, implied by "Manage Roles". Only check for Manage Roles, and granted if the specific one is one of the allowed permissions.

Issue #7619, opened 2020-11-13
Role editing permissions


deanmarcussen:
From some notes on https://github.com/OrchardCMS/OrchardCore/pull/1054

> Also the driver should check for specific permission in Edit and Update methods, for instance Roles can't be edited but by someone with a specific Roles permission.

PR #8117, opened 2020-12-22
Dynamic user permissions

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/7915
Fixes https://github.com/OrchardCMS/OrchardCore/issues/7619

Adds dynamic permissions to managing users, on a per role basis.

Screenshot 2020-12-29 at 11 31 04


deanmarcussen:
Think I should demo this first.

And if anyone wants to check or question any of the logic (it is complicated), please do ;)

scleaver:
Hey @deanmarcussen - I watched the video on this and had a question. I noticed that if a user had "manage user in roles - editor" it appeared that they couldn't then see any users outside that role. But what if I wanted to add an existing user to the role I manage? I can't see them in the user list as it is not in the role yet? In my systems I have used I a person with such position can still see all users it's just that they can't do anything but add/remove them to a that role.

Typical scenario - I work in corporate public relations and I manage the PR role. While I am away on holidays I want another person in the organisation to be able to edit PR related content.

Maybe I am missing something?

scleaver:
Oh hang on... you are grappling with that at the end of the video... sorry

deanmarcussen:
@scleaver yes, I think changed probably to what you are saying.

- all users in all roles visible
- manage permission gives ability to enable/disable change password etc
- assign roles permission to assign roles

features that can't be done, based on restricted permissions, are greyed out, or not visible. 
e.g. can't assign administrator role, can still assign others, but can't change/remove the administrator role...

bulk actions, and user filters, another pr ;)

src/Templates/OrchardCore.ProjectTemplates/content/OrchardCore.Templates.Cms.Module/.template.config/template.json
```diff
@@ -23,8 +23,8 @@
           "description": "Target net5.0"
         }
       ],
-      "replaces": "netcoreapp3.1",
-      "defaultValue": "netcoreapp3.1"
+      "replaces": "$(TemplateTargetPackageFramework)",
```
deanmarcussen, line 26: @agriffard something has gone slightly of with the merge here. Can you check that what you expect to be in these template modules is still there ok?

PR #8131, opened 2020-12-23
Tenants buttons

agriffard:
image


hishamco:
@agriffard did the team agreed to use the buttons again? Because using many colors is confusing and doesn't make sense, there should be a set of colors according the color scheme of the theme

agriffard:
People seem to be used to Bootstrap buttons and colors.

hishamco:
I don't have issue with the buttons, but introducing more colors without meaning doesn't make sense, that's my point

deanmarcussen:
I'm not a fan of the slider icon for the disable / enable.

It suggests you can swipe it, as if it was a checkbox selector.

What about tick and cross?

agriffard:
image

agriffard:
Ok, no more icons. 
Screenshot updated in first comment.

PR #8142, opened 2020-12-23
Create ManageMediaOptions Permission

### LINKED ISSUE #8141: Create new permission for Manage Media Options
Currently if you have the ManageMedia permission you can ~edit~ see the Media Options.
It should be a separate permission ~as it allows to change the base path of media folder~.

Skrypt:
And assign it to it's controller action.

Fixes #8141

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
@@ -445,7 +445,7 @@ namespace OrchardCore.Media.Controllers
 
         public async Task<IActionResult> Options()
         {
-            if (!await _authorizationService.AuthorizeAsync(User, Permissions.ManageMediaOptions))
+            if (!await _authorizationService.AuthorizeAsync(User, Permissions.ViewMediaOptions))
             {
                 return Forbid();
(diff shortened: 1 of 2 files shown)


Skrypt:
Other option would be to use ManageAdminSettings.

deanmarcussen:
Yes, I'm ok with separating this permission out.

Suggest it is named `ViewMediaOptions` as was done for the Azure module, as there is no Management involved. They're a read only view of the options.

https://github.com/OrchardCMS/OrchardCore/blob/894ab0c931a944ef9db1d56a5bbd903b450f94ee/src/OrchardCore.Modules/OrchardCore.Media.Azure/Permissions.cs#L10

hishamco:
@Skrypt how this controls the media options in the UI?

Skrypt:
Seems like there is no edit button so you can't edit them. But still, doesn't make sense to me to allow seeing these options when you are a content editor for example.

hishamco:
> But still, doesn't make sense to me to allow seeing these options when you are a content editor for example.

I was expecting the PR solving this ;)

Skrypt:
You expecting too much of me 😉

deanmarcussen (approved):

PR #8125, opened 2020-12-23
Code Mirror assets

agriffard:
Text field with Code Mirror and Markdown field look ok:

image

image

/cc @Skrypt

PR #8150, opened 2020-12-25
Code Mirror: Cloudflare cdn

agriffard:
https://cdnjs.com/libraries/codemirror

Related to #8125

@giannik Can you please test and tell me if you have less errors with this cdn?

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Resources/ResourceManifest.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Resources/ResourceManifest.cs
@@ -143,7 +143,7 @@ namespace OrchardCore.Resources
             manifest
                 .DefineStyle("codemirror")
                 .SetUrl("~/OrchardCore.Resources/Styles/codemirror/codemirror.min.css", "~/OrchardCore.Resources/Styles/codemirror/codemirror.css")
-                .SetCdn(codeMirrorUrl + "lib/codemirror.min.css", codeMirrorUrl + "lib/codemirror.css")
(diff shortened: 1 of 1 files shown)


giannik:
getting 
File not found: https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.59.0/lib/codemirror.css

agriffard:
> getting
> File not found: https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.59.0/lib/codemirror.css

Ok, css url fixed.

giannik:
worked for me.

Skrypt:
I can't asset that this work. It's a CDN network issue thing here. Are we changing for one that will be accessible for @giannik and not for other people? We're playing russian roulette 😉

Skrypt (commented):
I can't asset that this work. It's a CDN network issue thing here. Are we changing for one that will be accessible for @giannik and not for other people? We're playing russian roulette 😉

deanmarcussen:
Is this issue only occuring because of an upgrade to codemirror? i.e. if it was fine on 5.58, but isn't on 5.59 we would be better to revert the version change until it is sorted by the existing cdn.

agriffard:
I had used jsdelivr because cloudflare took many days before providing the 5.59.0 version of Code Mirror.

agriffard:
Cloudflare seems to be a standard cdn while jsdelivr only exposes the npm package content.

Skrypt:
@deanmarcussen It's not about the issue with codemirror we had. It's just a CDN change. If it works. merge it. No big difference to me if nothing breaks.

PR #8154, opened 2020-12-25
jQuery cdn

agriffard:
https://cdnjs.com/libraries/jquery
https://cdnjs.com/libraries/jqueryui

Related to #8125


Skrypt (commented):
I can't assert that this work. It's a CDN network issue thing here. Are we changing for one that will be accessible for @giannik and not for other people? We're playing russian roulette 😉

giannik:
Agree. Should all be from the same source. I don't use cdn , but I think
some people in china had issues using some cdn network and then it was
changed to the current one.

Virus-free.
www.avast.com

PR #8156, opened 2020-12-25
bootstrap select and slider style resource

### LINKED ISSUE #8155: bootstrap-select and bootstrap-slider style Resources
According to documentation in OrchardCore.Resources 
there should exist the relevant style resource for the matching script resource.

The style resource does not exist for : 

- bootstrap-select
- bootstrap-slider

making it not possible to use the library in the front end.

These should continue to exist regardless of bundling done for the admin theme.

giannik:
fixes #8155


agriffard:
Their styles don't seem to be used for the moment.
Can you please test if adding the style where the script is already used does change the look and feel, and add screenshots before/after?

giannik:
in front end part i need to add
```
<script asp-name="bootstrap-select"  at="Foot"></script>
<style asp-name="bootstrap-select" at="Head"></style>
```

without the style tag the bootstap before/after is like : 


before

image


after
image

agriffard:
What about in the admin? 
When you use a text field with those editors.

Skrypt:
You know that you can add these resources to your own ResourceManifest file included in your own theme?

giannik:
Yes I know , but it's inconsistent.
It breaks Orchards clean and tidy organization.
Anyway , what issue does it create if it is included , why are these 2 so
special they are excluded and not others ?


Virus-free.
www.avast.com

Skrypt:
The reason it's been removed is that these Bootstrap components are now bundled directly in TheAdmin.css styles. They are extensions of Bootstrap classes. With the darkmode admin theme, I had no choice but to bundle them to get color variants for each of our color modes. Now, I probably removed them from the resources to avoid using them as standalone resources in the admin theme. But, it is true that the ResourceManager is shared with the frontend assemblies if you add it's dependency. I guess we can add them back but it would be only for frontend usage. This is why I think it's not relevant to have them by default. The scripts are required, so they make sense.

agriffard (approved):
Ok, for frontend use.

PR #8158, opened 2020-12-25
Remove icon before See documentation

### LINKED ISSUE #8120: See documentation
Should we display an icon before the `See documentation` link or remove it?

agriffard:
Added a seedoc css class if we ant to skin it someday.

Fixes #8120


Skrypt (commented):
Maybe you could keep the icon by adding a style?

PR #8161, opened 2020-12-25
Shell Activation Atomicity

### LINKED ISSUE #7941: Race conditions running Orchard migrations on multiple instances

jtkech:
Fixes #7941 related to multiple instances starting and then executing migration steps concurrently.

- Here, the solution is to use a distributed lock when activating the shell, knowing that the default implementation uses local semaphores, **so, as long as we don't enable a distributed lock (e.g. `RedisLock`), it works as before**.

- **If the shell is initializing** (during a setup), we force the usage of the default using local semaphores, **so as before**.

Todo: Make locking times configurable

### DIFF
--- a/src/OrchardCore/OrchardCore.Abstractions/Locking/ILocalLock.cs
+++ b/src/OrchardCore/OrchardCore.Abstractions/Locking/ILocalLock.cs
@@ -1,0 +1,6 @@
+namespace OrchardCore.Locking
+{
+    public interface ILocalLock : ILock
+    {
+    }
+}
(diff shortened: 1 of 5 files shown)


jtkech:
Just to say that here the default locking timeout and expiration are configurable through a config source e.g. appsettings.json, and the default value for both is 30s

Issue #1261, opened 2017-11-29
Need documentation on creating site settings


websitewill:
Need to add to the OrchardCore documentation for making site settings as part of modules and themes.

PR #8165, opened 2020-12-27
Doc recipe custom settings

arkadiuszwojcik:
Updated Recipe documentation to include custom-settings deployment step example. This is based on my experience, I dind't know `ContentType` was required to make this step work.


src/docs/reference/modules/Recipes/README.md
```diff
@@ -325,12 +325,27 @@ The Deployment step allows you to create a deployment plan with deployment steps
...
 
+### CustomSettings Step
+
+The CustomSettings step allows you to populate your custom settings with initial values.
+
+```json
+    {
+      "name": "custom-settings",
+      "BlogSettings": {
+        "ContentType": "BlogSettings",
```
deanmarcussen, line 336: It's actually expected this will be the full json of a content item, not the limited parts you have included here.
This would work, but for an example, but for an example it would be better if you could can create one using the `CustomSettings` deployment step.

(this is missing any fields for example)

arkadiuszwojcik: @deanmarcussen sorry but I don't undestand last sentence. So I should also add also `ContentItemId` and some custom fields (i.e. TextField) right?

deanmarcussen: No prob. 
Go to import/export -> deployment plans.
Make a deployment plan with the `CustomSetting` in it.
Execute the plan.
You will see the full format of the custom setting.

It would not hurt if somewhere on the docs we explained that you can by and large see and understand a recipe steps format, by making a deployment plan. (so much easier than trying to guess at it!)

agriffard:
See also #1261

arkadiuszwojcik:
@agriffard I will add site settings to this PR. Should I include all posible field or just few and put `...` for all the rest? This PR is not site settings module documentation but just recipe deployment steps so few fields should be enough?

agriffard:
You can do a separate PR for the site settings. 
I was just mentionning it because it seems related.

arkadiuszwojcik:
@deanmarcussen is it better now?

deanmarcussen (approved):
LGTM

PR #8172, opened 2020-12-28
Serve module static files without creating a scope

### LINKED ISSUE #1634: Too many scopes created
We are going in the `ModularTenantContainerMiddleware` and creating a service scope even for serving a static file.

jtkech:
Fixes #1634 

What is funny is that it is already the case for the `Default` tenant if it has no url prefix, in that case they are already served by the application level static files middleware ;)


jtkech:
Just to say that here we have only one module static files middleware for all tenants

So at least one good thing in case there are many tenants

jtkech:
I will redo it if it is requested again

PR #8173, opened 2020-12-28
Sitemap Init and Cleanup

### LINKED ISSUE #8087: Sitemap cache handling improvments
@jtkech think there's a regression from the distributed pr where the sitemap caches are not being cleared when content items are updated / added, so it's serving stale comment.

I see that you're generating a new id, which I assume is for the document to refresh.

But the local cache itself is no longer cleared.

https://github.com/OrchardCMS/OrchardCore/pull/5249/files#r546233524

I haven't dug further into it.

But I guess we will need to clear the cache locally, and when the id / document is regenerated clear it on the others as well?

**Edited**

Aha! I have missed what you did with `Identifiers`

So the cache is no longer cleared, the identifier is used to resolve whether it is a valid sitemap.

But this will mean the cache is not used if the server is restarted, as it is only collecting identifiers during running operations?

There might just be some changes to be done in the interfaces, to drop some no longer used.

I have to go but there might be a better way we can store the file based on the identifier, and retrieve it that way.

jtkech:
Fixes #8087 

@deanmarcussen i could work on it, so that it still use the cached files on restart, and i quickly done a background task to cleanup the cached files, here the idea

- Still uses the cached files if the server is restarted

    This by just populating the related identifiers from the shared document on startup, before checking them

- Cleanup background task as a separate feature

    It's a flat directory, so for now we just compare the sitemap path to the file info name, assuming that the path doesn't contains any slash, e.g. a leading one in the path but not in the name. Then we compare the last modified time of the file with the sitemap last modified time (as a new property set on updating).

    If it works well, maybe better to not define it as a separate feature, then maybe remove the sitemap cache controller, notice that in a distributed env, from the admin we only list the cached files of the node that received the request.

   One problem i saw is that i could enter paths like `/sitemap1.xml`, `test/sitemap1.xml`, and the sitemap cache was throwing, so i added the forward slash to the `InvalidCharactersForPath` from the UI, i may be wrong let me know

---

But quickly done so need more thinkings, e.g. in the cleanup task we could just remove the unused files without checking any last modified time, and let the sitemap controller update the files in use based on the related identifiers. It would prevent some race conditions and be more consistent.

A time based solution is not so consistent, e.g. you update a sitemap and then just before caching the file, another node update the same sitemap, you will have a cached file not in sync even its modified time > last update time. Again here it's an atomicity problem of the whole, we solved it for a document by embedding an identifier in the document itself.

But if we now re-use the cached files on startup, we still need to remove them from the cleanup task. So, because a cached file can't embed its related identifier, **i'm now thinking about embedding the identifier in the file name itself**.

So that we could detect unused files whose name don't match a sitemap `path + identifier`, and when the sitemap controller will create a new file it will have a different name (with a new identifier) preventing some race conditions.

Will see tomorrow ;)


deanmarcussen:
> But if we now re-use the cached files on startup, we still need to remove them from the cleanup task. So, because a cached file can't embed its related identifier, **i'm now thinking about embedding the identifier in the file name itself**.

I like the sound of this.

Technically `test/sitemap1.xml` is valid. (never tested that it works w/the cache though!).

What it means for a sitemap is that `test` is the delegated path on the sitemap that it is allowed to include files for...

From https://www.sitemaps.org/protocol.html
The location of a Sitemap file determines the set of URLs that can be included in that Sitemap. A Sitemap file located at http://example.com/catalog/sitemap.xml can include any URLs starting with http://example.com/catalog/ but can not include URLs starting with http://example.com/images/.

If you have the permission to change http://example.org/path/sitemap.xml, it is assumed that you also have permission to provide information for URLs with the prefix http://example.org/path/. Examples of URLs considered valid in http://example.com/catalog/sitemap.xml include:

jtkech:
@deanmarcussen 

Some cache method only work with a file name, not with a path with multiple segments. But that's okay, sitemap paths and cache file names now are decoupled, sitemaps can be still served with any path, and the cache is always use as a flat directory where a cache file name for a given sitemap is

    /// <summary>
    /// Sitemap unique file name used for caching.
    /// </summary>
    public string CacheFileName => Name + "_" + Identifier + Sitemap.PathExtension;

And by embedding the identifier in the cache file name (as we embed an identifier in a shared document), we could get rid of the identifiers collection in the sitemap controller, much more simpler and more consistent. It seems to work fine, i can see every minute the cleanup background task removing obsolete cache files

Sitemap cache

jtkech:
Hi @deanmarcussen 

Did you see my previous comment?

I'm ready with this PR, if you want to give it a try and see if it fixes your issue

deanmarcussen:
@jtkech I like it.

Tried, and all works. Much better to have the Ids in the file name, and makes the whole arrangement way better.

I added a dependency on `BackgroundTasks` to the `Sitemaps.CleanUp` feature, because it didn't trigger for me initially, as I hadn't turned background tasks on.

If that change is ok with you, all LGTM. 

ty

jtkech:
@deanmarcussen Cool !

> I added a dependency on BackgroundTasks to the Sitemaps.CleanUp feature, because it didn't trigger for me initially, as I hadn't turned background tasks on.

You don't need it as the background tasks feature is just to be able to override settings from the UI, just tried, it works without this dependency

When you enable the feature, the tenant is rebuilt and the bg service will first re-init schedulers so you may have to wait one minute more to get it first executed

deanmarcussen:
> When you enable the feature, the tenant is rebuilt and the bg service will first re-init schedulers so you may have to wait one minute more to get it first executed

Ok. Must've just been slow, I've seen that before, when it is starting, that background tasks take a while to kick in. (or I am just always impatient)

Cool. Does it need to run every minute? I know it's pretty light weight, but most of the time it will do nothing? Not too concerned though

jtkech:
> Does it need to run every minute?

I asked me the same question, but because it's not a real concern, i kept it for debugging

But yes always better to not do unnecessary calls

What are you suggesting? 5 minutes, one hour, every day?

jtkech:
@deanmarcussen let me know the value you want to use for the bg task

deanmarcussen:
shall we say 5min? at least then anyone watching to see if they get deleted will see it before tomorrow ;)

jtkech:
Okay cool, 5 min, done

jtkech:
I let you approve it so that we can merge it

deanmarcussen (approved):

PR #8174, opened 2020-12-28
Flex summary admin

deanmarcussen:
Moving the summary admin display to a single row caused a few hiccups, and the changes to how the Zones were displayed has caused a few extensibility issues, as the zones are now in different places, so can no longer contain the items that were added to those Zones.

Additionally wrapping was a problem, not just on smaller screens, and with more text than our default recipes have.

In the interests of moving forward, rather than reverting back, here we move towards flexbox as a way of managing tags and selectors, and meta.

Notes
- Buttons are a seperate pr/issue. They've never stacked well. Suggest we move the `related` classes towards flexbox, and away from floats.
- Zones now stack instead of just adding to a weird difficult to predict wrap. Might be controversial. Might get changed back if not liked.
- To get a new line, the `Content` zone is now needed instead of the `Meta` zone. However it will render underneath the buttons. This was probably the biggest regresion for me, as having that on a seperate line made more sense, with the extra `Meta` some content items have.
- There is still a visual lack of alignment due to the light grey of the tags. Technically they now line up, but you can't easily see that, because the badge grey is so light. Minor.
- We lose the - between title and content type as it looks weird when things are stacked.
- Shown here as `ListPart` content to see how the selectors needed to be their own effective column.
- The two columns become auto cols so there is more room to display extra meta data without everything having to start wrapping.
- Brings back some of the zone classes that were removed, and used for targetting.

TODO
- Widgets

**What I'm not sure about is whether this further reduces the Zone extensibility by forcing stacking (and therefore items that are small enough to suit a stack) or not. So opinions from anyone using Zones please**

Before @ 1920
Before1920
After @ 1920

After1920

Before @ Ipad

BeforeiPad

After @ ipad

AfterIpad

Before @ Iphone X

BeforeIphoneX

After @ iphonex

AfterIphoneX

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.SummaryAdmin.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.SummaryAdmin.cshtml
@@ -6,43 +6,55 @@
 }
 
 <div class="row">
-    <div class="col-lg-6 col-sm-12 title">
-        @if (Model.Selectors != null)
-        {
-            <div class="selectors d-inline cursor-pointer">
-                @await DisplayAsync(Model.Selectors)
+    <div class="col-lg col-12 title d-flex">
+        <div class="selectors-container d-flex">
+            @if (Model.Selectors != null)
+            {
+                <div class="selectors cursor-pointer">
+                    @await DisplayAsync(Model.Selectors)
+                </div>
+            }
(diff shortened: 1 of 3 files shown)


Skrypt:
I think it looks better for the desktop sizes. Mobile sizes needs some work still. But that's a good step forward.

Issue #6751, opened 2020-07-22
Deleteing a content-type that is inuse causes an exception


CrestApps:
When creating a content-type "i,e. Widget", then adding it to a Widget. Then deleting the content-type, the system throws an exception, and is hard to use the GUI to recover from it.

**Solution:** prior to performing the "delete" operation content-type or content-part, check that it isn't being used. If it is used, you might want to throw an alert to the user and prevent the delete.

hishamco:
Showing a warning alert may good for this case, @CrestApps would you like to push a PR for this?

CrestApps:
I could mess with it. But it won't be any time soon.

hishamco:
Which version you are using? I just deleted an Article content type successfully

KaloyanIT:
I'm interested in contributing. Could you provide a little more detail on exactly what should be accomplished to resolve this issue?

hishamco:
Glad to hear that @KaloyanIT!! the idea basically to avoid deleting the content type who has already some content items, unless we show a warning alert or confirm message telling that the current content type has associated content types. @CrestApps still waiting for you

sebastienros:
Will be hard for types that are contained (widgets, bag part content items, menu items...)

I assume we should just log the issue, and ignore the content items without throwing exceptions.

deanmarcussen:
Related https://github.com/OrchardCMS/OrchardCore/issues/1014

PR #8178, opened 2020-12-29
Flow / Bag Null References when using deleted content type definitions

deanmarcussen:
Related to https://github.com/OrchardCMS/OrchardCore/issues/6751

and the many other issues when deleting content type definitions.

This fixes null reference exceptions thrown by bag parts and /or flow parts when the list of contained content types includes a type which has been deleted.

Does not fix throwing null references, if the content type is currently in use inside the bag, that's another pr.

However this is a common place error, when refactoring content type definitions.

The contained content types list on the settings is self healing already.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Flows/Drivers/FlowPartDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Flows/Drivers/FlowPartDisplayDriver.cs
@@ -99,9 +99,17 @@ namespace OrchardCore.Flows.Drivers
                 return _contentDefinitionManager.ListTypeDefinitions().Where(t => t.GetSettings<ContentTypeSettings>().Stereotype == "Widget");
             }
 
-            return settings.ContainedContentTypes
-                .Select(contentType => _contentDefinitionManager.GetTypeDefinition(contentType))
-                .Where(t => t.GetSettings<ContentTypeSettings>().Stereotype == "Widget");
+            var ctds = new List<ContentTypeDefinition>();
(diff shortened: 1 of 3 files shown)


src/OrchardCore.Modules/OrchardCore.Flows/Drivers/FlowPartDisplayDriver.cs
```diff
@@ -101,7 +101,7 @@ private IEnumerable<ContentTypeDefinition> GetContainedContentTypes(ContentTypeP
 
             return settings.ContainedContentTypes
                 .Select(contentType => _contentDefinitionManager.GetTypeDefinition(contentType))
-                .Where(t => t.GetSettings<ContentTypeSettings>().Stereotype == "Widget");
+                .Where(t => t != null && t.GetSettings<ContentTypeSettings>().Stereotype == "Widget");
```
hishamco, line 104: Is this case senstive or this should be lower case, coz I see the above query using `widget`

deanmarcussen: Technically content type definitions, and stereotypes, are one of the few things that are case sensitive.
I'm not sure where you see `widget` in the query anywhere, it is always `Widget`

hishamco: Sorry, seems I was wrong when I was in my mobile, just leave my comment

Issue #8184, opened 2020-12-30
Data : User email migration


Skrypt:
### Describe the bug
We recently merged a PR that migrates user names data and replaces the @ with a + char to disallow usage of a email as a user name. The migration is working on the Document table on the UserName field but it should also be applied on the NormalizedUserName field of a `public class User : Entity, IUser`. Also the UserIndex table should be updated accordingly.

**To Reproduce**

- Checkout 53edb2d6811b50bf8ded4ec48dd7f96c34fcf1fd commit on dev branch
- Create a new tenant
- Create a user with a email as user name
- Checkout ee1fb5f1f30aa353e406028c2333c7668b5851b3 commit on dev branch "or latest dev"
- Recompile and restart the website to execute the migration

**Expected behavior**

UserIndex and Document table should have migrated the email data properly in all fields it's required. See fields named previously.

PR #8187, opened 2020-12-30
Fix user migration

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/8184

Has been made to rerun on anyone who has already run the migration from previous pr.

/cc @Skrypt

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Users/Migrations.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Migrations.cs
@@ -66,8 +66,8 @@ namespace OrchardCore.Users
                .Column<string>(nameof(UserByClaimIndex.ClaimValue)),
                 null);
 
-            // Return 6 here to skip migrations on new database schemas.
-            return 6;
+            // Return 8 here to skip migrations on new database schemas.
+            return 8;
         }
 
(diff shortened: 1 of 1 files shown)


Skrypt:
Ok, so for me it doesn'T seem to update the UserIndex table it might be an issue on my side with some handler failing. I will take a look, but the PR seems appropriate.

deanmarcussen:
Another failing migration perhaps? What does your migration record say

Skrypt:
```
    {
      "DataMigrationClass": "OrchardCore.Users.Migrations",
      "Version": 7
    },
```

deanmarcussen:
Ok so if you have 7 this next migration which returns 8 hasn’t run.

The index is refreshed by the document save.

but previously as normalized user name wasn’t set, the index refresh produced same output.

Skrypt:
Ok merge it. It's required ... we will see if it fixes my issues for the UserIndex from nuget packages afterward. I don't want to spend too much time testing it with full source code.

Skrypt (approved):

Skrypt:
Ok, just saw that the UserIndex indexes NormalizedUserName. So this is the appropriate fix. Merged.

PR #8200, opened 2020-12-31
Admin: Display New menu setting

agriffard:
Display New menu:

image

Do not display New menu:

image


Skrypt (approved):

deanmarcussen (commented):
Wouldn't we just move it to a feature? instead of having another setting for it

Skrypt:
Depends if you want to make it a standard for this admin menu or if you want it to just be an "extension" of the admin menu.
Here, it would require to register yet an other service just for adding this "New" admin menu item, vs accessing one that is already registered and cached which doesn't hurt any perf. I think, it is fine to use a setting here since this is an admin UI customization and that's what these settings are for.

This one is just a matter of taste which I can't argue on I'd say.

Issue #6593, opened 2020-07-06
"where placed" ResourceLocation is needed


scleaver:
On the odd occasion, it is necessary to have a stylesheet or script registered in the HTML exactly where is currently placed in the HTML. If you want to use all the features of the resource manager and/or the tag helpers, there is no way to prevent Orchard Core from loading it in the head or at the foot and still get all the goodness like versioning, debug, cdn and cache busting.

Use case: if you are working with HTML generated by Orchard Core in a SPA and you are leveraging the shadow DOM this requires that you inject styles at exactly the right place in the HTML.

sebastienros:
In the templates we just register resources to be rendered later. This might need another attribute (tag is too heavy) to change the behavior of the tags such that they render the resource inline.

Something line `<script asp-name="bootstrap" at="inline" />` which would not use the ResourceManager.Register call, but render the resource inline.

ns8482e:
If you want to render inline you could use plain html `<script>` or `<style>` tag

scleaver:
> If you want to render inline you could use plain html `<script>` or `<style>` tag

@ns8482e - If you are using the resource management features like named resources, use-cdn etc then a plain HTML tag is not going to cut it unfortunately.

@sebastienros - I like your suggestion. I will have a go at creating a PR for this.

ns8482e:
> @ns8482e - If you are using the resource management features like named resources, use-cdn etc then a plain HTML tag is not going to cut it unfortunately

@scleaver if you are using cdn url and named resource then there’s no benefit  adding it at inline as it’s already an external url loaded separately from the document html

scleaver:
@ns8482e - you seem to be suggesting that there is another way around my need without creating an at="inline" enhancement but your not explaining your alternative solution. I would like to hear you out and understand your alternative.

First let me explain again in my use case - I want to use managed resources and managed resource features but I do not want the output (the link) of my managed resource (in this case a css file￼, which leverages named resources, cdn, debug versions and versions) to be placed in the head or foot but in the exact position where I place it within the HTML. Why you might ask? Well I am generating HTML templates from orchard for use in Vue.Js. The Vue.Js app needs to leverage shadow dom. In order to style HTML loaded in the shadow dom you need to load the css file within the shadow dom itself NOT in the header and footer of the page.

ns8482e:
IMHO having named resource defined with `inline` won't help,

As,

[...]

PR #2973, opened 2019-01-04
Prevent liquid tag helper conflicts.

jtkech:
Related to #2598.

- With the `liquid script` tag helper when using the `src` attribute (not `source`), because of [this line in aspnet](https://github.com/aspnet/AspNetCore/blob/release/2.2/src/Mvc/src/Microsoft.AspNetCore.Mvc.Razor/TagHelpers/UrlResolutionTagHelper.cs#L47), this is the `UrlResolutionTagHelper` which is selected (at least now, maybe since 2.2).

        [HtmlTargetElement("script", Attributes = "[src^='~/']")]

- So, this doesn't affect e.g the `liquid style` tag helper, and it works with our `razor script` tag helper which fully uses `asp-src`, not only `src`.

- **The result is that in our themes some scripts are not rendered in the right place and with the minified version because e.g `debug-src` and `at` are not used and rendered as html attributes**.

---

- So, for the `liquid script` the rule would be to use `source`.

- Then trying to solve the #2598 issue, i removed the `at:"Foot"` argument. Then the script tag helper was not selected because `source` is not an `HtmlTargetElement`. So i added it and it was selected.

- But but but, still nothing was rendered because i think it's by design. No `resources` tag helper targetting `HeadScript` or `FootScript` can render a script without any location.

    Note: With the `liquid style` tag helper, if no location it always defaults to `Head`. For the `liquid script` it defaults to `Foot` but only if neither `source` neither `name` is used. **Let me know if we need to do something for the `liquid script` if no location**, at least here it would be selected.

- Then, to harmonize things in liquid, for the `style` tag helper, i also added `source` as an `HtmlTargetElement`. So that, when no location is specified, it is still working when using `source`.

PR #8211, opened 2021-01-01
Support Inline location for styles and scripts

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6593

Adds support for `at="Inline"` for both styles and scripts.

We partially had support for this in scripts (to fix an older issue), but have extended it to support `at="Inline"` which allows it work with `depends-on` and naming. So full support, rather than partial.

Think you added the partial support here https://github.com/OrchardCMS/OrchardCore/pull/2973 @jtkech 

By adding `Inline` as an option, this is non breaking, as to do it without, i.e. use `Unspecified` would break existing `style` tags.

Obviously `Inline` is only useful as an option if using the tag helpers. It would do nothing if registered directly.

### DIFF
--- a/src/OrchardCore/OrchardCore.ResourceManagement/TagHelpers/ScriptTagHelper.cs
+++ b/src/OrchardCore/OrchardCore.ResourceManagement/TagHelpers/ScriptTagHelper.cs
@@ -131,7 +131,7 @@ namespace OrchardCore.ResourceManagement.TagHelpers
                     setting.SetAttribute(attribute.Name, attribute.Value.ToString());
                 }
 
-                if (At == ResourceLocation.Unspecified)
+                if (At == ResourceLocation.Unspecified || At == ResourceLocation.Inline)
                 {
                     _resourceManager.RenderLocalScript(setting, output.Content);
                 }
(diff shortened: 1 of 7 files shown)


src/OrchardCore/OrchardCore.ResourceManagement/TagHelpers/ScriptTagHelper.cs
```diff
@@ -183,19 +183,25 @@ public override async Task ProcessAsync(TagHelperContext context, TagHelperOutpu
...
                 {
                     var childContent = await output.GetChildContentAsync();
+                    // Named inline declaration.
                     if (!childContent.IsEmptyOrWhiteSpace)
                     {
                         // Inline content definition
                         _resourceManager.InlineManifest.DefineScript(Name)
                             .SetInnerContent(childContent.GetContent());
                     }
+                    if (At == ResourceLocation.Inline)
```
jtkech, line 197: Add a new line before

src/OrchardCore/OrchardCore.ResourceManagement/TagHelpers/ScriptTagHelper.cs
```diff
@@ -183,19 +183,25 @@ public override async Task ProcessAsync(TagHelperContext context, TagHelperOutpu
...
-                if (At == ResourceLocation.Unspecified)
-                {
-                    _resourceManager.RenderLocalScript(setting, output.Content);
-                }
-                else
+                // Allow Inline to work with both named scripts, and named inline scripts.
+                if (At != ResourceLocation.Unspecified)
                 {
                     var childContent = await output.GetChildContentAsync();
+                    // Named inline declaration.
```
jtkech, line 190: Move up the comment before line 189

src/OrchardCore/OrchardCore.ResourceManagement/TagHelpers/StyleTagHelper.cs
```diff
@@ -68,6 +63,11 @@ public override async Task ProcessAsync(TagHelperContext context, TagHelperOutpu
                     setting.AtLocation(ResourceLocation.Head);
                 }
 
+                foreach (var attribute in output.Attributes)
+                {
+                    setting.SetAttribute(attribute.Name, attribute.Value.ToString());
```
jtkech, line 68: Just for curiosity why moving this here from the above

deanmarcussen: Literally no reason @jtkech I'd moved the above `AtLocation` to later when testing, but when i didn't need it later, I moved it back, but forgot where it was originally ;)

I just moved it back

jtkech (approved):

Issue #8189, opened 2020-12-30
Setup screen fails when user has a previous failed setup


TalMcMahon:
Setup screen errors with "The user name is required" even though it is properly filled in (image below).

* If you re-enter the username and submit, the create fails and you end up on the "The requested tenant is currently initializing" screen (also image below)

I have not found an easy way to recover from this state.

This is a tough thing to reproduce, I will do my best to describe it. It seems to be related to the Username being re-used as the Super User.  

I was working on a recipe and wanted to apply it to a "Blank Site" Logged in as the super user (my email address).    My recipe failed when I tried to execute it from the recipes screen. I was unable to do anything else as my recipe really gummed it up.  It was after this point I came across the bug described above.  Here are the steps I followed after that:

**Steps to Reproduce**
1. I closed the browser. (which stops the debugger/iisexpress)
3. Deleted the Bin, Obj, and App_Data folders to get a clean start. (I am using the SqlLite option. so this should be all of the data)
3. Hit F5 to start a brand new site from the create screen.
4. Selected everything the same from my previous setup (Chrome auto filled it)
5. Used the same email address as my Super User Name
6. Hit submit Saw the bug described above.

**Work Around** 
1. I stopped the debugger
2. Deleted the Bin, Obj, and App_Data folders to get a clean start.
3. Hit F5 to start a brand new site from the create screen.
4. Selected everything the same from my previous setup (Chrome auto filled it)
5. Used a **something unique** as my Super User Name
6. Site was created as expected.

**Expected behavior**
I expected that if I delete the Bin, Obj, and App_Data Folders, I would have an entire reset of the app and would be able to start again from scratch. 

**Other information**
 The problem seems cache or cookie related, I have ruled out cookie with the following tests
     * Ran into the same issue in "Incognito" mode on Chrome (Chrome is my default browser)
     * Ran into the same issue running it with Edge Browser.
 
**Screenshots**
The error on the screen:
![image](https://user-images.githubusercontent.com/3964594/103374124-5d8ff080-4a9c-11eb-8433-cc217538d32f.png)

The only screen that will display after the above error occurs:
![Version1 0 0-rc2-15720-Fail2](https://user-images.githubusercontent.com/3964594/103374245-ad6eb780-4a9c-11eb-947e-7e700144cd01.png)

jtkech:
Yes, the `@` char is no more allowed for an user name

But the validation message is not well descriptive, someone needs to fix it

[...]

Issue #1665, opened 2018-04-08
Setup page throws error on bad user-name, despite creating db tables


MikaelElkiaer:
Using the latest Docker image for linux, I am experiencing an error when filling out the setup page.
I entered the username: 'MikaelElkiær' and upon submit it spent some time creating the tables, and then it comes back with an error: 'User name is requried'. Re-opening the page will confirm that everything was created successfully though, so the error is false.

**Update:** When trying to sign in with the user 'MikaelElkiær' I am met with 'Invalid login attempt'.

**Update:** Same is true for user-name with space(s), e.g. 'Mikael Elkiaer'.

Skrypt:
Try cleaning your cookies and let me know if it fixed the issue.

sebastienros:
TODO: We need to try and repro this

sebastienros:
I can still repro. In the admin when trying to create a new user it fails at server-side validation, I assume from aspnet identity. We should do the same validation on javascript and on the server at least to prevent getting into a bad state.

hishamco:
I will try to do the validation on both client & server, but I think we need to support Unicode characters in username after 1.0.0 at least

hishamco:
![Setup](https://user-images.githubusercontent.com/3237266/64666195-0b1d6180-d45e-11e9-8302-d06a3d84baff.png)

This is a related issue when the validation is show up, I see the pre-filled configuration fields are show up which is totally wrong

hishamco:
I fixed the issue that mentioned above, nothing but I don't think the client-side validation is a good idea, what we should do is to force `TenantState` to be `Uninitialized` if a validation error occurs

/cc @jtkech

jtkech:
Normally, this is what we do in `SetupAsync`.

                if (context.Errors.Any())
                {
                    context.ShellSettings.State = initialState;
                }

Also, just saw in the setup controller

            if (!ModelState.IsValid)
            {
                CopyShellSettingsValues(model);
                return View(model);
            }

Need to be tried but maybe we could just do

            if (!ModelState.IsValid)
            {
                return View(model);
            }

So that the user could retrieve the form fields that have been filled.

Otherwise here, as @sebastienros said, it seems that there are some missing checks on the user name that we would need to do before attempting to run the setup recipe.

jtkech:
For infos, in aspnetcore identity i saw this options class

    public class UserOptions
    {
        /// <summary>
[...]

PR #8213, opened 2021-01-01
Fix setup user name validation

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/8189

Fixes https://github.com/OrchardCMS/OrchardCore/issues/1665

Fixes the user name validation for setup/api/tenants

Tidies up some of the email validation, and now using the same localization strings as are in the view. So less work for translators.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Setup/Controllers/SetupController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Setup/Controllers/SetupController.cs
@@ -3,10 +3,12 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.DataProtection;
+using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Mvc;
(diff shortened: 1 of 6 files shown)


hishamco:
@deanmarcussen I think your PR will fix #1665 too

jtkech:
@deanmarcussen 

I tried with a bad connection string, it works, you can correct it, it's not too late

For the last errors checking, it's too late as soon as some tables have been created through migrations, nothing that we can really do. But I still added the code to set the setup context with the current shell settings (registered in memory), so that at least it is reverted back to the `Uninitialized` state.

jtkech:
@hishamco yes you're right

Fixes #1665

deanmarcussen:
@jtkech 

Cool, makes sense. Yes too late on the last error checking, but good to set the `ShellSettings` back at least.

For infos JT & @hishamco I discovered the other day that PR's only close issues if Fixes (or Closes etc) are in the first comment, so you have to edit the comment, to add another issue that is fixed by the PR. Then github links it properly.

Skrypt (approved):

PR #8227, opened 2021-01-03
Fix ListPart.DetailAdmin data-url wrongly renamed to data-url-af

### LINKED ISSUE #8226: ListPart.DetailAdmin data-url wrongly renamed to data-url-af
Same as #8225

Skrypt:
Fixes #8226


jtkech:
I was just checking if there were other places to update ;)

PR #8244, opened 2021-01-05
Fix notification locale when the site is restart

### LINKED ISSUE #4915: Alert localization issue
Since few months ago I notice a weird bug that I forgot to log, that when you change the culture from the admin site, the first two alerts shows with the previous culture, but the last one will use the current culture

I know the main reason for that we are stored a localized messages then after the post they will show up!! and that's why we have this behavior

IMHO we can store the message but not the localized version to fix this issue, @sebastienros if you have a better idea shoot it to me :) 

Localization

hishamco:
Fixes #4915

This fixes an old localization bug, especially when the default culture is set, some of the notification are localized with the culture before the new culture is taking a place

As I remembered Seb noted that in a very old Orchard Core Standup and said "This is weird as usual" ;)

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Settings/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Settings/Controllers/AdminController.cs
@@ -77,7 +77,9 @@ namespace OrchardCore.Settings.Controllers
                 var culture = site.Properties
                     .GetValue("LocalizationSettings")
                     .Value<string>("DefaultCulture");
-                using(CultureScope.Create(culture))
+
+                // We create a transient scope with the newly selected culture to create a notification that will use it instead of the previous culture
+                using (CultureScope.Create(culture))
                 {
                     _notifier.Success(H["Site settings updated successfully."]);
                 }
--- a/src/OrchardCore.Modules/OrchardCore.Settings/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Settings/Controllers/AdminController.cs
@@ -5,6 +5,7 @@ using Microsoft.AspNetCore.Mvc.Localization;
 using OrchardCore.DisplayManagement;
 using OrchardCore.DisplayManagement.ModelBinding;
 using OrchardCore.DisplayManagement.Notify;
+using OrchardCore.Localization;
 using OrchardCore.Settings.ViewModels;
 
 namespace OrchardCore.Settings.Controllers
@@ -73,7 +74,13 @@ namespace OrchardCore.Settings.Controllers
             {
                 await _siteService.UpdateSiteSettingsAsync(site);
 
-                _notifier.Success(H["Site settings updated successfully."]);
+                var culture = site.Properties
+                    .GetValue("LocalizationSettings")
+                    .Value<string>("DefaultCulture");
+                using(CultureScope.Create(culture))
+                {
+                    _notifier.Success(H["Site settings updated successfully."]);
+                }
 
                 return RedirectToAction(nameof(Index), new { groupId });
             }
--- a/src/OrchardCore.Modules/OrchardCore.Localization/Drivers/LocalizationSettingsDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Localization/Drivers/LocalizationSettingsDisplayDriver.cs
@@ -118,6 +118,7 @@ namespace OrchardCore.Localization.Drivers
                     // We always release the tenant for the default culture and also supported cultures to take effect
                     await _shellHost.ReleaseShellContextAsync(_shellSettings);
 
+                    // We create a transient scope with the newly selected culture to create a notification that will use it instead of the previous culture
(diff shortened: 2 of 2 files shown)


src/OrchardCore.Modules/OrchardCore.Settings/Controllers/AdminController.cs
```diff
@@ -73,7 +74,13 @@ public async Task<IActionResult> IndexPost(string groupId)
             {
                 await _siteService.UpdateSiteSettingsAsync(site);
 
-                _notifier.Success(H["Site settings updated successfully."]);
+                var culture = site.Properties
+                    .GetValue("LocalizationSettings")
+                    .Value<string>("DefaultCulture");
+                using(CultureScope.Create(culture))
```
hishamco, line 80: @jtkech this a new usage for `CultureScope` that had been introduced ;) please your review, I knew that may there are a better way to access "LocalizationSettings", I tried not use `As<LocalizationSettings>()`  to avoid having a dependency on Localization module. Any suggestion or this is fine?

jtkech: @hishamco 

About the dependency that's okay as it only references the localization abstractions project, which is already the case through another referenced project, because you didn't need to add a reference in the settings module project file.

First, look at the `LocalizationService.GetDefaultCultureAsync()` that does a null check, if the default culture is null it returns `CultureInfo.InstalledUICulture.Name`. Then we could inject and use `ILocalizationService` but we may get the previous value, as the new one is not yet updated.

Anyway, i don't think that using the default culture is right, and here for any site settings update. E.g. in a given request, the selected culture is not necessarily the default culture. As i remember, the default culture is only a fallback when no culture provider (called by priority) has selected a given supported culture, it would be too easy ;)

hishamco: > About the dependency that's okay as it only references the localization abstractions project, which is already the case through another referenced project, because you didn't need to add a reference in the settings module project file.

`LocalizationSettings` require a localization module dependency

> Then we could inject and use ILocalizationService but we may get the previous value, as the new one is not yet updated.

I already did that but I didn't get the updated value

> Anyway, i don't think that using the default culture is right, and here for any site settings update.

I'm agree in the principle nothing but such setting will be applied after a proper restart, so the default culture is right in this case especially if you set it from the admin site

hishamco:
I may ping @Skrypt because he worked on the cultures selector ;)

src/OrchardCore.Modules/OrchardCore.Settings/Controllers/AdminController.cs
jtkech: > LocalizationSettings requires a localization module dependency

But you don't use the type, only an hard coded string "LocalizationSettings" (without saying that's is good), otherwise you would need to add the reference in the settings module project file

> I already did that but I didn't get the updated value

Okay

> I'm agree in the principle nothing but such setting will be applied after a proper restart, so the default culture is right in this case especially if you set it from the admin site

Okay


Hmm, need more thinkings, would be better to just update the async local culture in `LocalizationSettingsDisplayDriver`, and because we are at the end of a request, we know that it is a section of the site settings, so we know that the tenant will be rebuilt, maybe in this driver just update the culture but without using the culture scope, so that this new culture is also used for the next notification.

But not sure, need more thinkings, for me make sense to display the first notification in the culture that is still the current one. E.g. if the session commit fails, the default culture will be still the "previous" one.

But just some thoughts, i'm not against changing it.

hishamco: > But you don't use the type, only an hard coded string "LocalizationSettings"

I thought there is a better approach, adding a reference to localization module is not allowed here if the intend is using the type

Regarding changing the culture in driver, I already this attempt but seems one of the messages not localized with that I used the scope .. it may need thinking but at the end changing the settings in this driver will 100% enforce the app to restart that's why I used eager loading for the new culture, because we know that the notification will be localized first, then the action filter will be executed

@sebastienros your thoughts

hishamco:
@sebastienros please have a look to this if you have a time, I'm sure you will remember this old bug while change the default culture

src/OrchardCore.Modules/OrchardCore.Settings/Controllers/AdminController.cs
```diff
@@ -73,7 +74,13 @@ public async Task<IActionResult> IndexPost(string groupId)
             {
                 await _siteService.UpdateSiteSettingsAsync(site);
 
-                _notifier.Success(H["Site settings updated successfully."]);
+                var culture = site.Properties
```
sebastienros, line 77: Please add a comment to explain why we create a new scope with a custom culture.
Something like "We create a transient scope with the newly selected culture to create a notification that will use it instead of the previous culture". In both places.

sebastienros (approved):

PR #8247, opened 2021-01-06
Search icon

### LINKED ISSUE #8246: Purpose of huge magnifying glass lurking behind the search field
### Describe the bug
There is a huge invisible magnifying icon lurking behind the search field in some of the admin pages. I have used a few different devices to see if there it has a purpose but it doesn't seem to have one.

Maybe it's an easter egg with a $1M prize money to the person who finds it?

It is represented by the orange box in the screen shot... because the SVG has oveflow:hidden it is currently not showing (unless you create your own admin theme and it appears). 

**To Reproduce**
Go to one of the content items list, view the source around the search field.

**Expected behavior**
No huge icon should be there - I think it is a relic from a bygone era of Orchard Core.

**Screenshots**
Screen Shot 2021-01-06 at 6 56 16 pm

Can I get rid of it and submit a PR?

agriffard:
image

Fixes #8246

/cc @scleaver


scleaver:
@agriffard so no prizemoney then???￼

agriffard:
> @agriffard so no prizemoney then???￼

There is maybe another easter egg to find.

PR #8259, opened 2021-01-08
Shortcode / Liquid "Content" conflict

### LINKED ISSUE #8241: Wysiwyg: Content item link
When user inserts link to a content item through wysiwyg, it will become invalid when document permalink changes. 

It would be great, if we could link content items directly (e.g. like in menu), through contentitemid / alias

jtkech:
Fixes #8241 

When processing a custom **shortcode template** defined through the admin, and using **liquid syntax**

- There is a liquid scope property name conflict, in the `ContentLiquidTemplateEventHandler` we do

      context.SetValue("Content", new LiquidContentAccessor());

    But in the `ShortcodeTemplatesDocument` we do

      return await _liquidTemplateManager.RenderAsync(template.Content, _htmlEncoder, model,
          scope =>
          {
              scope.SetValue("Content", model.Content); <== Here the conflict
              scope.SetValue("Args", model.Args);
              scope.SetValue("Context", model.Context);
          });

    So, in this context the `LiquidContentAccessor` doesn't work, e.g. `Content.ContentItemId["xxxxxx"]`

- I suggest to use a lower case for the shortcode properties `args, content, context`, to prevent conflicts with other liquid properties already there in the scope, and where the rule would be to use an upper case for liquid properties.

    it's a little breaking change, but if we agree i will update the doc

- I also  pass the liquid model when available

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Liquid/Filters/ShortCodeFilter.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Liquid/Filters/ShortCodeFilter.cs
@@ -22,10 +22,11 @@ namespace OrchardCore.Liquid.Filters
             // TODO This provides no context to the shortcode service.
             // It could take a content item as an argument to provide some context.
 
-            return new StringValue(await shortcodeService.ProcessAsync(input.ToStringValue(), new Context
-            {
-                ["Model"] = ctx.LocalScope.GetValue("Model")
-            }));
+            return new StringValue(await shortcodeService.ProcessAsync(input.ToStringValue(),
+                new Context
+                {
+                    ["Model"] = ctx.LocalScope.GetValue("Model")
+                }));
         }
     }
 }
--- a/src/OrchardCore.Modules/OrchardCore.Liquid/Filters/ShortCodeFilter.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Liquid/Filters/ShortCodeFilter.cs
@@ -21,7 +21,7 @@ namespace OrchardCore.Liquid.Filters
             // TODO This provides no context to the shortcode service.
             // It could take a content item as an argument to provide some context.
 
-            return new StringValue(await shortcodeService.ProcessAsync(input.ToStringValue());
+            return new StringValue(await shortcodeService.ProcessAsync(input.ToStringValue()));
         }
     }
 }
--- a/src/OrchardCore.Modules/OrchardCore.Liquid/Filters/ShortCodeFilter.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Liquid/Filters/ShortCodeFilter.cs
@@ -4,6 +4,7 @@ using Fluid;
 using Fluid.Values;
 using Microsoft.Extensions.DependencyInjection;
 using OrchardCore.Shortcodes.Services;
+using Shortcodes;
 
 namespace OrchardCore.Liquid.Filters
 {
@@ -21,7 +22,10 @@ namespace OrchardCore.Liquid.Filters
             // TODO This provides no context to the shortcode service.
             // It could take a content item as an argument to provide some context.
 
-            return new StringValue(await shortcodeService.ProcessAsync(input.ToStringValue()));
+            return new StringValue(await shortcodeService.ProcessAsync(input.ToStringValue(), new Context
+            {
+                ["Model"] = ctx.LocalScope.GetValue("Model")
+            }));
         }
     }
 }
--- a/src/OrchardCore.Modules/OrchardCore.Liquid/Filters/ShortCodeFilter.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Liquid/Filters/ShortCodeFilter.cs
@@ -3,8 +3,8 @@ using System.Threading.Tasks;
 using Fluid;
 using Fluid.Values;
 using Microsoft.Extensions.DependencyInjection;
-using OrchardCore.Shortcodes.Services;
 using OrchardCore.DisplayManagement.Shapes;
+using OrchardCore.Shortcodes.Services;
 using Shortcodes;
 
 namespace OrchardCore.Liquid.Filters
@@ -22,10 +22,15 @@ namespace OrchardCore.Liquid.Filters
 
             var context = new Context();
 
+            // Retrieve the 'ContentItem' from the ambient liquid scope.
             var model = ctx.LocalScope.GetValue("Model").ToObjectValue();
             if (model is Shape shape && shape.Properties.TryGetValue("ContentItem", out var contentItem))
             {
                 context["ContentItem"] = contentItem;
+            }
+            else
+            {
+                context["ContentItem"] = null;
             }
 
             return new StringValue(await shortcodeService.ProcessAsync(input.ToStringValue(), context));
--- a/src/OrchardCore.Modules/OrchardCore.Liquid/Filters/ShortCodeFilter.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Liquid/Filters/ShortCodeFilter.cs
@@ -4,6 +4,8 @@ using Fluid;
 using Fluid.Values;
 using Microsoft.Extensions.DependencyInjection;
 using OrchardCore.Shortcodes.Services;
+using OrchardCore.DisplayManagement.Shapes;
+using Shortcodes;
 
 namespace OrchardCore.Liquid.Filters
 {
@@ -18,10 +20,15 @@ namespace OrchardCore.Liquid.Filters
 
             var shortcodeService = ((IServiceProvider)services).GetRequiredService<IShortcodeService>();
 
-            // TODO This provides no context to the shortcode service.
-            // It could take a content item as an argument to provide some context.
+            var context = new Context();
 
-            return new StringValue(await shortcodeService.ProcessAsync(input.ToStringValue()));
+            var model = ctx.LocalScope.GetValue("Model").ToObjectValue();
+            if (model is Shape shape && shape.Properties.TryGetValue("ContentItem", out var contentItem))
+            {
+                context["ContentItem"] = contentItem;
+            }
+
+            return new StringValue(await shortcodeService.ProcessAsync(input.ToStringValue(), context));
         }
     }
 }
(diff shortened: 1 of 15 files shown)


Skrypt:
I suggest that we use `scope.SetValue("Shortcode", model);` to make this more strict.

deanmarcussen:
It is quite breaking. 

Is there not a way we can wrap the shortcode `Content` in a variant of the `LiquidContentAccessor` - it is just used as a marker is it not?

Obviously my bad on this one, never thought we would want to be hitting those accessors through shortcodes.

Skrypt:
I made some tests last night with @jtkech and one is obviously making these shortcodes to be called recursively which creates an infinite loop. We need to add an exception or a limit for this. I'm not sure right now what we should do because I don't know the entire what other things might affect this Context. The thing is that anyone can alter other contexts by affecting the same key/value pair from anywhere. Making them more specific might be easier for maintainability.

jtkech:
@deanmarcussen 

The best solution is what @Skrypt suggests, it would prevent future conflicts by using e.g. `Shortcode.Content`, but still a little breaking change, but do you think there are many people already using custom shortcode templates?

> Is there not a way we can wrap the shortcode Content in a variant of the LiquidContentAccessor

I will try it

@Skrypt about recursivity, yes, i will take a look on it this night

For the fun, with @Skrypt we tried to do a custom shortcode template (so using liquid) that uses another shortcode piped by the shortocode filter, the last shorcode being also a custom shortcode template (so also using liquid), it was working ;)

jtkech:
@Skrypt @deanmarcussen 

- I reused the original shortcode properties names so that it is no more a breaking change

- Then i updated `LiquidContentAccessor` so that it works as before, but it also wraps a string `Content` that is returned when it is read as a string, , here used to wrap the shortcode content.

deanmarcussen:
Nice one @jtkech 

I don’t think we should be adding the model as you’ve done.

the problem is when you do the short code is only usable in that context and we wanted to keep them with as little context as possible, or make the context as close as possible, so they are more universal.

does that make sense?

jtkech:
@deanmarcussen

Yes, we first added it for testing, then i thought about removing it, but because it is only related to shortcode template, only related to `TemplateShortcodeProvider`, and that uses liquid syntax, i kept it.

But i'm not against removing it, yes lighter and cleaner.

Oops, forgot to see about recursion, if i can add something

Skrypt:
> does that make sense?

Yes, I understand your point now. I did not analyze this code properly when I looked at it. So, yes, what @jtkech did works for me. Though, if we would not had to keep backward compatibility ; I would have probably prefered to have a strict identifier.

Now, from what I understand the "content" context here for the shortcode is everything contained between it's brackets. Ex: "[shortcode arg1="" arg2=""]content[/shortcode]". And for a Liquid template ; it's content is either a ContentField, ContentPart, ContentItem. So in this case "Model.Content" represent something totally different than what we have in a shortcode.

I'm just writing to keep some track on the logic too. Correct me if I'm wrong.

jtkech:
@Skrypt For liquid the `Content` property is a `LiquidContentAccessor` that in fact allows to use the content manager when using a property access on it, e.g. by doing `Content.Latest(...)`. It is already there in the liquid scope, and we don't set it in the shortcode context, but we prevent to override this `Content` property when we add properties to the liquid scope, this when rendering a shortcode template using liquid. We can discuss about it on skype if you want

@deanmarcussen 

But i saw this comment in the `ShortcodeFilter` liquid filter

            // TODO This provides no context to the shortcode service.
            // It could take a content item as an argument to provide some context.

So, at least in this filter i added the `ContentItem` if it exists, from the ambient liquid model to a shortcode context.

Skrypt:
Yeah, you mean that this PR prevents to override the `LiquidContentAccessor` I think.

jtkech:
So, as a summary, only 3 files updated, and as commented in the code

    // Retrieve the 'ContentItem' from the ambient liquid scope.

    // Check if a shortcode template is recursively called.

    // Don't conflict with the liquid scope 'Content' property.

src/OrchardCore.Modules/OrchardCore.Shortcodes/Services/TemplateShortcodeProvider.cs
```diff
@@ -44,7 +50,21 @@ public async ValueTask<string> EvaluateAsync(string identifier, Arguments argume
             return await _liquidTemplateManager.RenderAsync(template.Content, _htmlEncoder, model,
                 scope =>
                 {
-                    scope.SetValue("Content", model.Content);
+                    // Used for recursion checking.
+                    scope.SetValue(identifier, "");
+
+                    // Don't conflict with the liquid scope 'Content' property.
+                    var content = scope.GetValue("Content").ToObjectValue();
+                    if (content is LiquidContentAccessor contentAccessor)
```
deanmarcussen, line 58: Looks good @jtkech 

But. Need to check now, how will this work with a nested shortcode? 
Is it a local scope it is attaching too? I think so, but checking.

e.g.
```
[foo]foocontent[bar]barcontent[/bar][/foo]
```

src/OrchardCore.Modules/OrchardCore.Liquid/Filters/ShortCodeFilter.cs
```diff
@@ -18,10 +20,20 @@ public async ValueTask<FluidValue> ProcessAsync(FluidValue input, FilterArgument
...
 
-            // TODO This provides no context to the shortcode service.
-            // It could take a content item as an argument to provide some context.
+            var context = new Context();
 
-            return new StringValue(await shortcodeService.ProcessAsync(input.ToStringValue()));
+            // Retrieve the 'ContentItem' from the ambient liquid scope.
+            var model = ctx.LocalScope.GetValue("Model").ToObjectValue();
+            if (model is Shape shape && shape.Properties.TryGetValue("ContentItem", out var contentItem))
+            {
```
deanmarcussen, line 28: yep, interesting approach. Will only work from the context of the liquid filter, not with the razor extension, that does the same thing?

Which was why the TODO suggested it as an argument to the filter, so the razor extension could also take an optional argument.
Unless we can attach to the current rendering scope in the razor filter to provide the same? 

That'd be pretty cool.

src/OrchardCore.Modules/OrchardCore.Shortcodes/Services/TemplateShortcodeProvider.cs
jtkech: @deanmarcussen it works as each get and use its own local scope

src/OrchardCore.Modules/OrchardCore.Liquid/Filters/ShortCodeFilter.cs
jtkech: @deanmarcussen I added exactly the same to the razor extension and tried it

Skrypt:
Hi, guys. This morning I've been thinking about these kind of implementations. There's nothing wrong about it but let me explain my point of view. Here, some part of the code is doing a "patch" validation to support current shortcodes that people did. This is why we are doing all this. So basically, I see these changes as some kind of override to support Liquid with shortcode templates. Here is just one thing.

The other thing is that, like I said to @jtkech, it is not a lot of code, but if we use this approach all the time, it adds up ; whereas using the other approach of changing the key/value pair to "Shortcode" would have been cleaner and require less code.

Here, I'm thinking that it looks a lot like the UserId PR where we wanted to support old data. The thing here is that I'm probably too perfectionist and see this as extra code that could be simply not required if we would have had a tool to migrate the Document table JSON data. 

So, generally, it's easier for myself to execute a small SQL script to migrate my data than composing extra code to support old and new data. That's basically what I do everytime I require a structural change on my website, because that's how it is and can't avoid it at some point.

So, based on this, my vision of things was, back then, with the UserId PR to provide some tool to allow migrating the Document table JSON data when we require to do it. But, right now, we don't have this kind of tool at all, which would/could allow to avoid needing to add extra code in OC to support present and future data. Because taking this logic in consideration, there's nothing that says that eventually we won't require another data change and then we will add extra code to support all the variants of the data that got changed over time which could be more than 2 then in this case.

This is why I was kind of 😢 to see that we need to do these kind of PR's. Maybe, I'm overseeing this, but that's how I feel about part of it.

However cheers, Liquid now seems to work fine.

jtkech:
@Skrypt

Don't worry, look at the changed files, there is not so much code, and we added other things. We now add the `ContentItem` to the shortcode context also when processed from a razor view, we also added a recursivity checking, and the change related to the `Content` property is only 2 or 3 lines of code.

For info when rendering a shortcode template in liquid, it is done in a liquid local scope, when we set a property, we don't overwrite the same of the parent scope, we just hide it. For info, when we read a prop from a local scope, it first try to get it from this scope, then it fallbacks to the parent scope. So just to say that the parent will retrieve its original prop value.

So here the original issue was to be able to also use the liquid content accessor e.g. `Content.Something`, which makes sense as we are writing liquid code. So yes we could have change the content prop name or as you suggested use a `Shortcode` prefix with this syntax e.g. `Shortcode.Content` (not only `Content` as we would have the same "issue").

But, just to say that there is not so much code and that we never overwrite the props of the parent scope.

deanmarcussen:
@Skrypt I'm ok with it. 

It was intentional to want the property as `Content` - we did do that by design (discussion with Seb and I on the implementation), as underlying in shortcodes it is the `Content` property.

So for me feels less like a hack, and more like the right thing to do.

@jtkech for me approved, but giving @Skrypt opportunity to have more thoughts?

Skrypt:
No more comments, I'm fine with it, but we seriously need to start thinking about a way to migrate the document table json data. At least we should have some tool. Right now every time a SQL migration is required in this table we simply add code in OC to handle more and this is what's wrong, no matter the amount of code. It just adds up here and there over time...

jtkech:
@Skrypt @deanmarcussen 

Anyway that's okay for me, so just a last thinking ;)

But here it's a specific case where we are defining a shortcode through liquid, so it would not have hurt me to access the shortcode properties through a dedicated liquid property, e.g. `Shortcode.Args`, `Shortcode.Content`, `Shortcode.Context`

Skrypt:
For me, using same value here is the hack because then this value doesn't define anything specific anymore. Why use a key/value pair then? But, it's ok I understand the intent to universalize the same kind of values directly in the editors. Though, here it's a Shortcode value, so to me it makes sense to make it specific. I don't know, I leave you guys decide 😄 😉

jtkech:
I decided ;)

As it adds the `ContentIem` to the shortcode context and fixes the recursivity

And will be easy to change the shortcode template liquid props acesses in a separate PR if we want

PR #7951, opened 2020-12-12
Fixes Resource Dependencies accumulation and thread safety

jtkech:
Fixes #7712 
Fixes #7939 

On each request, for a given resource definition we may add the dependencies from the require settings if any, but the resource definition is held in a singleton collection, so its dependencies are growing up, and it is not thread safe.

Here we still use any additional dependencies from the require settings, but without mutating the resource definition to prevent its dependencies from growing up, and we no longer mutate a singleton dictionary through a scoped service.

PR #8265, opened 2021-01-08
Introduce ResourcePosition

MichaelPetrinolis:
If a resource has a dependency to '*', it is placed after all other resources are loaded.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Users/Controllers/AccountController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Controllers/AccountController.cs
@@ -506,8 +506,7 @@ namespace OrchardCore.Users.Controllers
             {
                 foreach (var item in state.Value.Errors)
                 {
-                    iix++;
-                    TempData[$"error_{iix}"] = item.ErrorMessage;
+                    TempData[$"error_{iix++}"] = item.ErrorMessage;
                 }
             }
             return RedirectToAction(nameof(Login), new { returnUrl });
--- a/src/OrchardCore/OrchardCore.ResourceManagement.Abstractions/RequireSettings.cs
+++ b/src/OrchardCore/OrchardCore.ResourceManagement.Abstractions/RequireSettings.cs
@@ -22,6 +22,7 @@ namespace OrchardCore.ResourceManagement
         public bool? AppendVersion { get; set; }
         public List<string> Dependencies { get; set; }
         public Action<ResourceDefinition> InlineDefinition { get; set; }
+        public ResourcePosition Position { get; set; }
 
         public Dictionary<string, string> Attributes
         {
@@ -131,7 +132,7 @@ namespace OrchardCore.ResourceManagement
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public RequireSettings UseCondition(string condition)
         {
-            Condition = Condition ?? condition;
+            Condition ??= condition;
             return this;
         }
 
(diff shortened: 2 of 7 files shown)


deanmarcussen:
@MichaelPetrinolis think this makes sense.

so rather than depending on a number of scripts, you can specify that, this resource is rendered last always, even if other scripts depend on it?

In the case of other scripts depending on it, it just forces this one to be registered?

Not having read all your Monaco pr (looks great), just wanted to check the reasoning, and if there will still be an issue with scripts that are unnamed, i.e. general inline foot scripts, as they will still be rendered after this `*` I suspect?
Because they are considered slightly different by the resource manager.

hishamco:
Agree with @deanmarcussen this should respect the resources dependency graph, but in case of TagHelpers for instance, the idea to include them all, while they are separated

MichaelPetrinolis:
@deanmarcussen with this PR the resources expansion takes place in two phases. If a resource has a dependency to `*` or depends to a `star` resource, the expansion of the dependency graph will take place in the second phase. So, in the first phase everything works as previously expect the `star` dependent resources, which will expand in the second phase. So, if other resources depend on a `star` resource, they will still be rendered after the star resource but in the second phase. Also, if you have two resources that you would like to render last, but resource `A` depends on resource `B` you can set resource `A` dependencies [`*`] and resource B dependencies [`*`, `A`] so both will be rendered in the second phase, but resource B will be loaded after resource A.
The tests included in the PR are only for named resources, I suppose the unnamed resources will still work as they used to.

@hishamco sorry i didn't quite understand what you mean with the tag helpers.

The reason for this PR is that monaco uses AMD Loader. see https://stackoverflow.com/questions/55057425/can-only-have-one-anonymous-define-call-per-script-file for a description of the problem.

Using `*` in SetDependencies is a little bit confusing because one would expect that this resource depends and loads all other resources registered instead of what it really does -expanding it in the second phase-. Maybe adding a LoadLast property to distinguish that the resource should be loaded last would be a clearer option.

hishamco:
> @hishamco sorry i didn't quite understand what you mean with the tag helpers.

This is a sample of what I mean:

```
```

Start may be easier in this case, but in case of the resources, the * should respect the resources dependency

MichaelPetrinolis:
Please correct me if i am wrong, what you say is that `*` in the addTagHelper means load all tag helpers from that namespace but `*` in resource registration means Load Last after all other dependencies, so there might be a confusion. 
Do you agree introducing a .SetLoadLast(true) method in resource registration instead of .SetDependencies("*")? 
Would it make it clearer?

hishamco:
> but * in resource registration means Load Last after all other dependencies

`*` here should check each resource dependency, then each dependency will check its dependency and the process continue .. which is hard little bit, 

> In the case of other scripts depending on it, it just forces this one to be registered?

Hope @deanmarcussen ellaborate how the dependency in this case will be resolved

jtkech:
Interesting, would work as module / theme dependencies where the current front end theme depends on all modules implicitly, unless a module depends on this theme, as i remember in that case we break the infinite loop.

> so rather than depending on a number of scripts, you can specify that, this resource is rendered last always, even if other scripts depend on it? In the case of other scripts depending on it, it just forces this one to be registered?

All scripts having a start dependency would need to be at the end in the order they are registered, hmm then i think another script could depend on one of them, but by taking care of the infinite loop of the circular dependency.


> Not having read all your Monaco pr (looks great), just wanted to check the reasoning, and if there will still be an issue with scripts that are unnamed, i.e. general inline foot scripts, as they will still be rendered after this * I suspect?
Because they are considered slightly different by the resource manager.

All scripts having a start dependency would need to be at the end in the order they are registered, but only in their related `Location` e.g. to be at the very end, such a script would need to specify `At="foot"`

deanmarcussen:
> Using * in SetDependencies is a little bit confusing because one would expect that this resource depends and loads all other resources registered instead of what it really does -expanding it in the second phase-

yes it probably is a little confusing. I'm struggling with a name to define this as well.
` .SetLoadLast(true)` sounds better, but also isn't necesarily true, as you say that you can make two scripts load last?

It's essentially positional, right? so similar in a way to shape placement, where we have `:before`, `:10` or `:after`

I half wonder about actually using the positional sorted here, so it might be `SetDependencies(":after")` but I suspect that has an impact on the performance, that we don't need.

That might be useful for other things, except perf and this code is really focusing on two phases of expansion, rather than offering some positional sort.

So could be `SetLoadAfter()` ? or maybe the modifier in `SetDependencies(":after")` - to save having to define extra parameters, for a rarely used feature, but some consistency with positioning features.

Thoughts here rather than specific suggestions... @jtkech may have some naming thoughts as well.
Be good to figure it out, so we can merge monaco.

@hishamco think you've confused this with tag helpers - different thing going on here.

jtkech:
I don't remember well the details, but maybe defining / using a new location e.g. `at="AfterFoot"`, `at="Bottom"` ;)

MichaelPetrinolis:
This is the best/simplest solution, but we should define a new script zone in all themes, and when request a resource that needs to be loaded after all other scripts, we should put it `at="Bottom`.  Should I revert the changes and just leave the circular dependencies detection?

MichaelPetrinolis:
but on the other hand, with the `:after` we could register a resource that will be referenced always last in the section you request it.

sebastienros:
Suggestion: Add a Position property to ResourceDefinition, enum of ByDependency/First/Last, defaults to ByDependency. If Last/First is selected, then it will be re-ordered where it needs to be.

Faster, simpler to understand, and doesn't require hacky values in dependencies.

MichaelPetrinolis:
A brief explanation of the logic (and the tests)

1. No circular dependencies are allowed
2. First resources are registered first, followed by dependency resources and then the last ones
3. If a ByDependency resource depends on a Last resource, it is registered last after the Last resource
4. If a First resource depends on a ByDependency resource, it is also registered First
5. A First Resource cannot depend on a Last resource

deanmarcussen:
@jtkech looks good to me. can you take a look as well?

jtkech:
@MichaelPetrinolis @deanmarcussen 

After a quick look, it seems that in a given shell scope, we update the new `ResourcePosition` in the `ResourceDefinition` (we pass the position by reference), but the resource definition is held in a singleton collection, we already fixed an issue around this #7951. Okay, writing an enum is thread safe, but in a given scope we may not see the right value.

So, as it was done in #7951 for dependencies, i suggest to add to the `RequireSettings` a new `ResourcePosition` property that will be initialized in a given shell scope, and then allowing to mutate it in this same scope.

As i understand it would also allow to not have to use a `Tuple`.

MichaelPetrinolis:
@jtkech the reason not adding a ResourcePosition to RequireSettings in the first place was that position changes only while expanding and i did not know if it was cool to introduce such a read/write property in the abstractions layer. Also preferred tuple instead of adding a private class for simplicity (again is only used in expansion) and the value is stored in an OrderedDictionary which requires casting. The stack is used only for detecting circular dependencies, it could contain only the name of the expanding dependency to achieve the same goal but i am not sure if creating a string and storing it is better than push/pop an object reference (pointer). Please feel free to make changes in this PR directly as you wish :) I'll rename the PR to 'Introduce ResourcePosition' to be more precise.

jtkech:
@MichaelPetrinolis Okay cool, no problem, need to focus on it to be sure, i will do this week end

jtkech:
@MichaelPetrinolis Yes, maybe easy to say but not so simple ;)

I made good progress, i will commit something this night

jtkech:
@MichaelPetrinolis Sorry did not have enough time, will work on it this night ;)

MichaelPetrinolis:
@jtkech thank you 👍

jtkech:
@MichaelPetrinolis 

I updated 2 files, but from the github UI i can't update `RequireSettings` as it is not part of the changed files of this PR, so please pull my changes and commit a minor change of `RequireSettings` so that i will be able to update it.

Also, i created a new `ResourceDictionary` class but from the github UI i can't create a new file for it, so please in the same commit move the new `ResourceDictionary` class from `ResourceManager.cs` to a new file, and clean the `using`

Otherwise, I kept the same logic, I mainly did some refactoring, change where things are done, and tried to limit new allocations. For the circular dependency we have to manage a new stack, but i think it's good to have this checking.

MichaelPetrinolis:
@jtkech done.

jtkech:
@MichaelPetrinolis Cool thanks, i will update it soon

MichaelPetrinolis:
@jtkech  it seems ok. I merged it in monaco editor branch (the reason for this PR) and it works as expected.

jtkech:
@MichaelPetrinolis Cool, thanks for having re-tried it

jtkech (approved):

PR #8269, opened 2021-01-10
Fix inline stylesheet rendering

deanmarcussen:
There was a couple of issues with rendering stylesheets inline from https://github.com/OrchardCMS/OrchardCore/pull/8211

- Didn't work properly through liquid
- Rendered some inline style sheets twice (depending on how they were placed)
- Consistency for `At` supplied for unamed inline styles/scripts


scleaver:
@jtkech or @agriffard  - can you review and merge this if possible?

PR #8280, opened 2021-01-12
Add additional text info to content picker field, so user can distigu…

MikeKry:
Added culture and autoroute (if filled) to content item picker, so user can easily pick content items. Using only display name has become a problem for multilingual sites with page named e.g. "blog" in multiple languages.

image


MikeKry:
@deanmarcussen 

I am not sure, if I should assign somebody to look at this, or what is the exact process with PRs? :)

src/OrchardCore.Modules/OrchardCore.ContentFields/ContentItemDisplayExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+using OrchardCore.ContentManagement.Routing;
+
+namespace OrchardCore.ContentFields
+{
+    public static class ContentItemDisplayExtensions
+    {
+        public static async Task<string> GetContentItemRouteAndCulture(this ContentItem contentItem, IContentManager contentManager, IStringLocalizer S)
+        {
+            var displayText = string.Empty;
+            var additionalParts = new List<string>();
```
deanmarcussen, line 15: this would be a `StringBuilder` from the string builder pool

src/OrchardCore.Modules/OrchardCore.ContentFields/ContentItemDisplayExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+
+            // add culture to display text
+            var cultureAspect = await contentManager.PopulateAspectAsync<CultureAspect>(contentItem);
+            if (cultureAspect.HasCulture)
+            {
+                additionalParts.Add($"{S["Culture"]}: {cultureAspect.Culture}");
+            }
+
+            // add autoroute path to display text
+            var routeAspect = await contentManager.PopulateAspectAsync<RouteHandlerAspect>(contentItem);
```
deanmarcussen, line 25: don't use this aspect get the path. It's not pathbase / tenant aware. this is used for contained routing.
You need the url helper, and `ContentItemMetadata.DisplayRouteValues` to generate a valid path.

deanmarcussen (commented):
Did you open an issue for this @MikeKry ?

Normally we start with issues, offer design guidance, (triage is on Thursday at 10 (PST?)), and then look at PRs.

However, for this one. Lets look at it the issue.

You want more metadata in the content picker?

I like the principle of the idea.

This sounds like an aspect.
We used to have a `DisplayText` property on the `ContentItemMetadata` which allowed you to manipulate the display text's for things like this. Unfortunately it's been obsoleted, and replaced in favour of just `ContentItem.DisplayText`

Maybe we could have a `AdditionalDisplayText` property there.

Here the problem is you've tied it to Culture, and Route. 
What if I want Alias? And what if I don't want Route? 
Or don't want any of them?

The routes on some of my sites are far too long to be displayed in the content picker list, for example.

MikeKry:
@sebastienros 

Ok, thanks for reply. So everything should always start with issue/feature request?  I have created it backwards for now #8374

From your comment I see the issue, that not everyone would like this format of display, I think it can use similiar setup as Autoroute part => liquid syntax so you can define anything you want here, but I am not sure, if it is possible within contentpicker context.

MikeKry:
Updates based on 8374

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/IContentPickerResultProvider.cs
```diff
@@ -7,6 +7,8 @@ namespace OrchardCore.ContentManagement
     public interface IContentPickerResultProvider
     {
         string Name { get; }
+
+        Task<string> GetContentPickerItemDescription(ContentItem contentItem, string pattern, string defaultValue);
```
deanmarcussen, line 11: Don't change this interface, or do anything to do with descriptions here.

deanmarcussen (commented):
Concept looks good

Suggest:
`ContentPickerAspect` and using the aspect system from the content manager to generate the code you have put in the result provider.

To implement, make a `ContentPickerAspect` class w/ `DisplayText` and `Description`

implement a `ContentHandlerBase` and override `GetContentItemAspectAsync` `.ForAspect<ContentPickerAspect` and do the rendering work there.

Then you can safely inject it in either the driver, or the result provider as required.

MikeKry:
@deanmarcussen 

I have finally looked at aspect system and reworked this to aspects. Hopefully it is ok now - tested and still works, removed bunch of unnecessary code in existing implementations.

2 problems I met and not sure they are solved right:
a) in ContentPickerAspect there are content part settings passed as object, because there can't be reference to ContentPickerFieldSettings - did not want to move settings or aspect, as aspect should be referenced from anywhere.
b) in ContentPickerAspect - passing CultureInfo - > I cannot use ContentManager in ContentHandlerBase as it is circular reference, so it is passed into aspect, is it ok?

MikeKry:
@deanmarcussen 
does it give sense to you like that? there are some merge conflicts now but I can update this branch with master if its ok with you

deanmarcussen:
Concept looks good @MikeKry. Would suggest reading the previous pr review

src/OrchardCore.Modules/OrchardCore.ContentFields/Drivers/ContentPickerFieldDisplayDriver.cs
```diff
@@ -19,13 +19,16 @@ public class ContentPickerFieldDisplayDriver : ContentFieldDisplayDriver<Content
     {
         private readonly IContentManager _contentManager;
         private readonly IStringLocalizer S;
+        private readonly IContentPickerResultProvider _contentPickerResultProvider;
 
         public ContentPickerFieldDisplayDriver(
             IContentManager contentManager,
-            IStringLocalizer<ContentPickerFieldDisplayDriver> localizer)
+            IStringLocalizer<ContentPickerFieldDisplayDriver> localizer,
+            IContentPickerResultProvider contentPickerResultProvider)
```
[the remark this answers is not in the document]
MikeKry: removed

src/OrchardCore.Modules/OrchardCore.ContentFields/Drivers/ContentPickerFieldDisplayDriver.cs
```diff
@@ -42,7 +45,7 @@ public override IDisplayResult Display(ContentPickerField field, BuildFieldDispl
 
         public override IDisplayResult Edit(ContentPickerField field, BuildFieldEditorContext context)
         {
-            return Initialize<EditContentPickerFieldViewModel>(GetEditorShapeType(context), async model =>
+            return Initialize(GetEditorShapeType(context), (Func<EditContentPickerFieldViewModel, ValueTask>)(async model =>
```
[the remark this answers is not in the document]
MikeKry: reverted

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/IContentPickerResultProvider.cs
MikeKry: removed title and description from this interface

src/OrchardCore.Modules/OrchardCore.ContentFields/ContentItemDisplayExtensions.cs
MikeKry: moved and reflected in [ContentPickerContentHandler.cs]

MikeKry:
@deanmarcussen 

I looked at your comments when I reworked this and I think I have reflected all your comments in https://github.com/OrchardCMS/OrchardCore/pull/8280/commits/72661068bc33f4788ed78d478b54f91bc543beb6

I have gone through them again and I think everything should be ok, so I commented and resolved them.

MikeKry:
@Skrypt 
Ok, I was out for a while, but I have merged it now and tested - still working ok.

MikeKry:
@Skrypt  
Just a question - does "Needs triage" mean anything for me, or is it just waiting for some group review? I am not very familliar with how this triage etc. works.

Skrypt:
Needs triage means that we need to assign it to a milestone from a Thursday triage meeting. I use also this to make it prioritary over other PRs so that we can review it.

src/OrchardCore.Modules/OrchardCore.Lucene/Services/LuceneContentPickerResultProvider.cs
```diff
@@ -61,7 +61,7 @@ public async Task<IEnumerable<ContentPickerResult>> Search(ContentPickerSearchCo
                     {
                         ContentItemId = doc.GetField("ContentItemId").GetStringValue(),
                         DisplayText = doc.GetField("Content.ContentItem.DisplayText").GetStringValue(),
-                        HasPublished = doc.GetField("Content.ContentItem.Published").GetStringValue() == "true" ? true : false
+                        HasPublished = doc.GetField("Content.ContentItem.Published").GetStringValue() == "true" ? true : false,
```
Skrypt, line 64:
```suggestion
                        HasPublished = doc.GetField("Content.ContentItem.Published").GetStringValue() == "true" ? true : false
```

src/OrchardCore.Modules/OrchardCore.ContentFields/Views/ContentPickerFieldSettings.Edit.cshtml
```diff
@@ -53,6 +61,22 @@
     <span class="hint">@T["The hint text to display for this field on the editor."]</span>
 </div>
 
+<div class="mb-3">
+    <div class="-md-75 w-xl-50 pr-md-3">
+        <label asp-for="TitlePattern">@T["Title Pattern"]</label>
+        <textarea asp-for="TitlePattern" rows="5" class="form-control"></textarea>
+    </div>
+    <span class="hint">@T["The pattern used to render the custom content item title in content picker. DisplayText is used when empty is empty. With Liquid support."]</span>
```
Skrypt, line 69:
```suggestion
    <span class="hint">@T["The pattern used to render the custom content item title in content picker. DisplayText is used when title is empty. With Liquid support."]</span>
```

src/OrchardCore.Modules/OrchardCore.ContentFields/Handlers/ContentPickerContentHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+                var settings = aspect.ContentPickerFieldSettings as ContentPickerFieldSettings;
+
+                if (settings != null) {
+                    aspect.Title = await GetContentPickerItemExtendedDescription(context.ContentItem, aspect.Culture, settings.TitlePattern, context.ContentItem.ToString());
+                    aspect.Description = await GetContentPickerItemExtendedDescription(context.ContentItem, aspect.Culture, settings.DescriptionPattern, string.Empty);
+                }
+            });
+        }
+
+        private async Task<string> GetContentPickerItemExtendedDescription(ContentItem contentItem, CultureInfo culture, string pattern, string defaultValue)
```
Skrypt, line 37:
```suggestion
        private async Task<string> GetContentPickerItemExtendedValue(ContentItem contentItem, CultureInfo culture, string pattern, string defaultValue)
```

src/OrchardCore.Modules/OrchardCore.ContentFields/Handlers/ContentPickerContentHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            var description = defaultValue;
```
Skrypt, line 39:
```suggestion
            var result = defaultValue;
```

src/OrchardCore.Modules/OrchardCore.ContentFields/Handlers/ContentPickerContentHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+            if (!string.IsNullOrEmpty(pattern))
+            {
+                
+                using (CultureScope.Create(culture))
+                {
+                    description = await _templateManager.RenderStringAsync(pattern, NullEncoder.Default, contentItem,
```
Skrypt, line 45:
```suggestion
                    result = await _templateManager.RenderStringAsync(pattern, NullEncoder.Default, contentItem,
```

src/OrchardCore.Modules/OrchardCore.ContentFields/Handlers/ContentPickerContentHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+                        new Dictionary<string, FluidValue>() { [nameof(ContentItem)] = new ObjectValue(contentItem) });
+                }
+            }
+
+            return description;
```
Skrypt, line 50:
```suggestion
            return result;
```

src/OrchardCore.Modules/OrchardCore.ContentFields/Handlers/ContentPickerContentHandler.cs
Skrypt: Here, I don't think you need to do this. This adds string allocation for nothing. You could also use defaultValue and morph its value and return it.

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/IContentPickerResultProvider.cs
```diff
@@ -21,6 +21,7 @@ public class ContentPickerSearchContext
     public class ContentPickerResult
```
Skrypt, line 21: Can you move and extract these classes into separate files? I'm pretty sure they don't belong in the abstraction project.

MikeKry:
@Skrypt  

Implemented suggested changes, 

Merged main into this branch but **I still need to check tomorrow if it did not break anything** - kinda lot changes

MikeKry:
@Skrypt 

Tested, fixed codemirror. It is working fine again with the new changes. Maybe it should be reviewed by @MikeAlhayek, because there were many a lot of conflicts with https://github.com/OrchardCMS/OrchardCore/pull/12103

MikeKry:
Is it possible, that this will make it to version 1.5?

sebastienros:
Sorry for the late reply, and for what's coming ;)

I think it can be much simpler with a single setting that contains the liquid template to use for the description of the content item in this list. Similarly to autoroute paths, with a default value that is just `{{ ContentItem | display_text }}`. Then users can customize each field to display the culture, type, ... without requiring any custom code. You did that part in this PR. I don't think we need to be able to do it by code.

NB: the way you were passing settings and culture to the aspect population made me realize this. Actually each implementation of the handler to populate the aspect should have resolved the culture if they needed it. Same for the settings.

MikeKry:
@sebastienros 

Not sure what you mean, because Autoroute paths generation also pass culture aspect to liquid templates (and imho it can't be generated without manually getting culture aspect). 

It can be seen here: https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs#L406

Also there is no custom code needed from developer, title and description can resolve any liquid expression same as autoroute part.

But it is possible I misunderstood what you mean..? Maybe I would need a little more explanation.

sebastienros:
Correct, you will need to get the culture from the aspect, but then the template is sufficient, something like this:

https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs#L400-L409

MikeKry:
@sebastienros 

So you mean, that new ContentPickerAspect is not needed? That is ok and I can change it, but feels like step-back to state before this commit https://github.com/OrchardCMS/OrchardCore/pull/8280/commits/72661068bc33f4788ed78d478b54f91bc543beb6#diff-c12e6f2791029f70fa686c607691139163bfd007933317dfb86d22c3122e671f

Where @deanmarcussen suggested, that it should rather be an aspect. Are you sure that we want to change it back? I have no problem in doing it, but I admit that I do not understand this concept of aspects and when to use them in OC :).

If so, please suggest me where to place function for resolving liquid expression for description/title. I am not sure whether it should be placed in some handler or service.

MikeKry:
@sebastienros 

What is your decision, should I revert it or no? I would like to finish this as it is open for 1 year already :)

MikeAlhayek:
@MikeKry can you please fix the conflict?

MikeKry:
@MikeAlhayek 

Sure, I pushed merge into my branch. Is there possibility, that it will be merged any time soon? It starts to be very uncomfortable using content picker with growing number of content items.

MikeKry:
I see that release 1.6 is already merged, but this PR did not make it through.. will it ever be merged?

sebastienros:
Tasks to unblock the PR
- remove the aspect
- add a setting that defines the text to display in the list
  - either use an extra checkbox to define an optional setting that will then be used
  - use the setting all the time, and the default value is the liquid text to get the display name of the content item (backward compatibility)

sebastienros:
I would go with a single field right now, not changing how it looks in the UI. We can always add an extra one later on if someone finds a reason to. The view model should not change then, and the view either.

MikeAlhayek:
> Tasks to unblock the PR
> 
>     * remove the aspect
> 
>     * add a setting that defines the text to display in the list
>       
>       * either use an extra checkbox to define an optional setting that will then be used
>       * use the setting all the time, and the default value is the liquid text to get the display name of the content item (backward compatibility)

@MikeKry are you able to do this so it can be unblocked and hopefully merged before 1.7 is released?

MikeKry:
@sebastienros  @MikeAlhayek 

Ok, I have removed aspect and field for description. I have added default value for backwards compatibility. 

image

image

image

src/OrchardCore.Modules/OrchardCore.ContentFields/Views/ContentPickerFieldSettings.Edit.cshtml
```diff
@@ -1,6 +1,16 @@
...
 @using OrchardCore.ContentManagement
 
+<style asp-name="codemirror"></style>
+<script asp-name="codemirror" depends-on="admin" at="Foot"></script>
+<script asp-name="codemirror-mode-javascript" at="Foot"></script>
+<script asp-name="codemirror-addon-display-autorefresh" at="Foot"></script>
+<script asp-name="codemirror-addon-mode-simple" at="Foot"></script>
+<script asp-name="codemirror-addon-mode-multiplex" at="Foot"></script>
+<script asp-name="codemirror-mode-xml" at="Foot"></script>
+
```
MikeAlhayek, line 11: please remove extra space.

src/OrchardCore.Modules/OrchardCore.ContentFields/Settings/ContentPickerFieldSettingsViewModel.cs
```diff
@@ -15,4 +15,5 @@ public class ContentPickerFieldSettingsViewModel
     public bool Multiple { get; set; }
 
     public string[] DisplayedContentTypes { get; set; } = Array.Empty<string>();
+    public string TitlePattern { get; set; } = "{{ Model.ContentItem.DisplayText }}";
```
MikeAlhayek, line 18: please add a new line between line 17-18

src/OrchardCore.Modules/OrchardCore.ContentFields/Settings/ContentPickerFieldSettingsDriver.cs
```diff
@@ -59,12 +65,30 @@ public override async Task<IDisplayResult> UpdateAsync(ContentPartFieldDefinitio
...
         }
 
+        private bool ValidateTextPatterns(UpdatePartFieldEditorContext context, ContentPickerFieldSettingsViewModel model)
+        {
+            bool isValid = true;
+
+            if (!string.IsNullOrEmpty(model.TitlePattern) && !_templateManager.Validate(model.TitlePattern, out var titleErrors))
+            {
+                context.Updater.ModelState.AddModelError(nameof(model.TitlePattern), S["TitlePattern doesn't contain a valid Liquid expression. Details: {0}", string.Join(" ", titleErrors)]);
+                isValid = false;
```
MikeAlhayek, line 86: Please re-label `S["TitlePattern doesn't contain a valid Liquid expression. Details: {0}"` to `S["Title Pattern does not contain a valid Liquid expression. Details: {0}"`

The `return false;`

src/OrchardCore.Modules/OrchardCore.ContentFields/Settings/ContentPickerFieldSettingsDriver.cs
```diff
@@ -59,12 +65,30 @@ public override async Task<IDisplayResult> UpdateAsync(ContentPartFieldDefinitio
...
+            }
+
+            return isValid;
```
MikeAlhayek, line 89: return true instead of using isValid;

src/OrchardCore.Modules/OrchardCore.ContentFields/Services/DefaultContentPickerResultProvider.cs
```diff
@@ -48,15 +59,21 @@ public async Task<IEnumerable<ContentPickerResult>> Search(ContentPickerSearchCo
             var contentItems = await query.Take(50).ListAsync();
 
             var results = new List<ContentPickerResult>();
+            var settings = searchContext.PartFieldDefinition.GetSettings<ContentPickerFieldSettings>();
 
             foreach (var contentItem in contentItems)
             {
-                results.Add(new ContentPickerResult
+                var cultureAspect = await _contentManager.PopulateAspectAsync(contentItem, new CultureAspect());
+                using (CultureScope.Create(cultureAspect.Culture))
```
MikeAlhayek, line 67: use a semicolon `;` at the end of the using line and drop the `{  }` to save indentation. 
```
using CultureScope.Create(cultureAspect.Culture);
```

src/OrchardCore.Modules/OrchardCore.ContentFields/Settings/ContentPickerFieldSettings.cs
```diff
@@ -10,5 +11,10 @@ public class ContentPickerFieldSettings
         public bool DisplayAllContentTypes { get; set; }
         public string[] DisplayedContentTypes { get; set; } = Array.Empty<string>();
         public string[] DisplayedStereotypes { get; set; } = Array.Empty<string>();
+
+        /// <summary>
+        /// The pattern used to build custom title
+        /// </summary>
+        public string TitlePattern { get; set; } = "{{ Model.ContentItem.DisplayText }}";
```
MikeAlhayek, line 18: add attribute [DefaultValue("{{ Model.ContentItem.DisplayText }}")] to the new property since it has a custom default value.

src/OrchardCore.Modules/OrchardCore.ContentFields/Settings/ContentPickerFieldSettingsDriver.cs
```diff
@@ -59,12 +65,30 @@ public override async Task<IDisplayResult> UpdateAsync(ContentPartFieldDefinitio
                         break;
                 }
 
-                context.Builder.WithSettings(settings);
+                var isValid = ValidateTextPatterns(context, model);
+
+                if (isValid)
```
MikeAlhayek, line 70: maybe `if(IsValidateTextPatterns(context, model))` instead?

src/OrchardCore.Modules/OrchardCore.ContentFields/Settings/ContentPickerFieldSettingsDriver.cs
```diff
@@ -59,12 +65,30 @@ public override async Task<IDisplayResult> UpdateAsync(ContentPartFieldDefinitio
...
+                {
+                    context.Builder.WithSettings(settings);
+                }
             }
 
             return Edit(partFieldDefinition);
         }
 
+        private bool ValidateTextPatterns(UpdatePartFieldEditorContext context, ContentPickerFieldSettingsViewModel model)
```
MikeAlhayek, line 79: maybe rename the method do `IsValidTitlePattern`

MikeAlhayek (commented):
@MikeKry can you confirm that HTML is supported in the picker? For example, if I provide the following pattern, will the HTML show up at text or will we see my custom HTML?

```
<a href="/Admin/Contents/ContentItems/{{ Model.ContentItem.ContentItemId }}/Display" target="_blank">{{ Model.ContentItem.DisplayText }}</a>
```

MikeKry:
@MikeAlhayek 

no, html is not rendered. I don't think, it would be good to allow html inside selector, it could break html structure etc.
image


All but one suggestions were applied to code in new commit.

MikeAlhayek:
@MikeAlhayek I think allowing HTML will solve for other scenarios. and if you are providing a custom template contained HTML, you should know what you are doing and the app should allow it.

If you don't mind, please add HTML support in the selector.

MikeKry:
@MikeAlhayek 

Okay, would you like html to be rendered both while selecting and also when items are already selected? 

In example with hyperlinks it can cause redirection when clicking on item to select (but that may be also somehow controlled by html that is inserted)

image

MikeKry:
I have pushed version with both variants showing rendered html

MikeAlhayek:
Humm, I think in the suggestion we only show the Display text, but the selected options should support HTML. Like if the user want to add a link to edit the selected content item or view the selected content item.

MikeKry:
well I think that would be very complex solution.. Stripping html when searching is possible but would lead to nonsense results and using only display text would dump original intention of this pull request.

Main intention why I suggested this change to content picker is to enable quicker and easier searching - using display text only when searching would change nothing in this usecase.

MikeAlhayek:
I understand. I think it would be great if we have a solution for this. See https://github.com/OrchardCMS/OrchardCore/pull/13764 adding HTML will solve that issue as well.

MikeKry:
@MikeAlhayek  

I have reverted html rendering from this feature, as I think that solution suggested in #13764 is way better than allowing custom html.

If we would want to go through with allowing custom html, then I think it would be necessary to create another liquid field with definition for display of selected items. Then it would give sense to use it for buttons etc. but I think it is a bit overkill.

src/OrchardCore.Modules/OrchardCore.ContentFields/Services/DefaultContentPickerResultProvider.cs
MikeAlhayek: @MikeKry this one please

MikeAlhayek:
@sebastienros I think this one is ready, can we triage it?

sebastienros (approved):

PR #8284, opened 2021-01-12
PathBase aware Favicon

jtkech:


sebastienros:
Do you remember why?

jtkech:
@sebastienros 

Just re-tried, without this fix, under a virtual folder it returns a 404 (after clearing the browser cache)

PR #8349, opened 2021-01-19
Upgrade Fluid to 2.0.0-beta-1001

hishamco:


hishamco:
Seems it needs to remove Irony dependency

sebastienros:
Please do not merge until it has been tested with most views, even though all fluid tests pass, there are many more cases in OC that might trigger issues. Found one just by picking the first liquid template I could find in OC.
And also run the functional tests manually, as it will hit multiple pages.

hishamco:
That's why I told you yesterday I need to test few cases .. This may need to be tested by the team if it is done .. after we sure there's no bugs we can merge

sebastienros:
Can you try to fix the build so I can see if something is missing in v2?

hishamco:
Sure, I will try to continue on this PR after an hour, one more thing how we can pass `FilterArgument[]`, I didn't find overload to pass such type. I will struggle to make this pass

sebastienros:
I am working on it. The migration is not easy, and requires some work in Fluid.

hishamco:
> I am working on it. The migration is not easy, and requires some work in Fluid.

Ok, it's in your plate now .. I was busy with the SQL Parser, BTW I have some feedback to Parlot, I will share it ASAP on its repo

hishamco:
Replaced by #8424

PR #8350, opened 2021-01-19
Fixes autoroute trailing slash

jtkech:
`PathBase` traps, it fails if you init it with a value without leading `/` but can have a trailing `/`.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Autoroute/Services/AutorouteEntries.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Autoroute/Services/AutorouteEntries.cs
@@ -32,7 +32,7 @@ namespace OrchardCore.Autoroute.Services
         {
             await EnsureInitializedAsync();
 
-            if (_contentItemIds.TryGetValue(path, out var entry))
+            if (_contentItemIds.TryGetValue(path.TrimEnd('/'), out var entry))
             {
                 return (true, entry);
             }


src/OrchardCore.Modules/OrchardCore.Autoroute/Services/AutorouteEntries.cs
```diff
@@ -32,7 +32,7 @@ public async Task<(bool, AutorouteEntry)> TryGetEntryByPathAsync(string path)
         {
             await EnsureInitializedAsync();
 
-            if (_contentItemIds.TryGetValue(path, out var entry))
+            if (_contentItemIds.TryGetValue(path.TrimEnd('/'), out var entry))
```
deanmarcussen, line 35: @jtkech we might have to fix this at a driver level.

Because now if you use the path `blog/` in autoroute, it will not consider it a duplicate if you try and add another route of `blog` ? Am I wrong?

jtkech: Okay, need to leave but i will check it tomorrow, i thought it was okay because of this

        public AutorouteEntry(string contentItemId, string path, string containedContentItemId = null, string jsonPath = null)
        {
            ContentItemId = contentItemId;

            // Normalize path.
            Path = "/" + path.Trim('/');

            ContainedContentItemId = containedContentItemId;
            JsonPath = jsonPath;
        }

Hmm, and what do we do about case sensitivity, i will check it also

jtkech: @deanmarcussen after a quick look

> Because now if you use the path blog/ in autoroute, it will not consider it a duplicate if you try and add another route of blog ? Am I wrong?

It was already the case, you were already able to enter `blog` and in another item `blog/`, here it's just to make it working with `blog`, and also with `blog/`, if you entered `blog` for a given autoroute.

So yes, this PR fixed one issue, but there was another issue about checking a possible conflict, this because currently we can enter `blog`, `/blog`, `blog/` and `/blog/` and all of them need to be considered as a conflict.

I saw that you query the `AutoroutePartIndex` to check for a possible conflict (both in the handler / driver), here for a given path e.g. `blog` would need to check if `blog`, `/blog`, `blog/` and `/blog/` already exist, not only `blog`.

I thought about just using `_entries.TryGetEntryByPathAsync()`, in place of querying the index to for a possible conflict, with the fix in this PR, and because the path of an autoroute entry is normalized, i tried it it works.

But no luck, autoroute entries only hold autoroutes of published items, not all latest ;)

I will work on it soon in a new PR.

deanmarcussen: Cool, yes, we talked about it in the meeting as well, and are not sure it's a valid idea to actually strip.

Because they are actually two different routes `blog` and `blog/`, so to combine them as one route feels like the wrong thing to do.

Particularly from an seo point of view, where google will analyse them and store them for search as two different routes, which has an impact on search performance.

So we pretty much all agreed that it's not correct to treat them as one route.

It's often something we do as an iisrewrite, to either strip all trailing slashes, or add them. With OC it is always to strip them.
But that sometimes comes down to a matter of taste.

jtkech: @deanmarcussen Okay, i'll move this discussion to #8370

PR #8370, opened 2021-01-20
Fixes Autoroute conflicts

jtkech:


jtkech:
Discussion moved from the already merged #8350 

From @deanmarcussen 

> Cool, yes, we talked about it in the meeting as well, and are not sure it's a valid idea to actually strip.

> Because they are actually two different routes `blog` and `blog/`, so to combine them as one route feels like the wrong thing to do.

> Particularly from an seo point of view, where google will analyse them and store them for search as two different routes, which has an impact on search performance.

> So we pretty much all agreed that it's not correct to treat them as one route.

> It's often something we do as an iisrewrite, to either strip all trailing slashes, or add them. With OC it is always to strip them.
But that sometimes comes down to a matter of taste.

jtkech:
@deanmarcussen 

Yes, serving both with a trailing slash or not is how the default routing works, e.g. all our admin urls, but i understand the SEO point of view for the front end, that may be fixed as you said through rewrite rules.

- But would we still allow to input `blog` and `blog/`, and that would serve 2 different items?

    If so we would need to update `AutorouteEntry` that forces a leading slash but trims the trailing one.

- At least in this PR we still need to consider `blog` and `/blog` as a possible conflict.

- How would we serve `blog/`, as a 404? Or having some default rewrite rules by code.

- And what about the home page url with or without a single slash, that here is both a leading and trailing slash ;)

- Personnaly, i think it's better to still allow `blog` and `blog/`, but not for 2 different items, so as a possible conflict as checked in this PR, and serving the same item as done in #8350

    Then it's up to the user to use rewrite rules e.g. through IIS or by using the aspnetcore rewrite middleware.

    But i'm open to any suggestion ;)

So let me know what has been decided in the meeting, i need to know what we allow to input, what need to be considered as a conflict with `blog`, e.g. `/blog` but not `blog/` and `/blog/`, and what we serve for `blog/`.

jtkech:
@sebastienros @deanmarcussen 

- I was going to merge this one as it is consistent with the already merged #8350 that trims the trailing `/` of the request path, i already did the same fix before for autoroute, notice that on github you go to the same page with a trailing `/`.

- But just saw the video of the last meeting, so okay, if you want i could revert the change of #8350.

    **But what was not clear is do we still allow to input `bog/`?**

    Note: Or `/blog/` that results to the same as the autoroute entry normalizes its path by forcing a leading `/`.

    If so, `blog` and `blog/` would be 2 different items, so we would have to change the autoroute entry to normalizes its path by still forcing a leading `/`, but by not trimming the trailing `/`.

    And then, to check possible conflicts e.g. on `blog`, we would still have to check `blog` and also `/blog`, as added in this PR, but not `blog/` and `blog/` and `/blog/` as also added in this PR.

In any case it is easy to do, but let me know what we want to do

deanmarcussen:
Do you remember who actually raised the issue @jtkech ?

I think we should determine if we are going to consider them the same url, or different, before we go too far on fixing a duplication issue.

I think we should definitely allow **bog** on principle as it's a great url ;)

jtkech:
It was @themattkenney but on gitter

But there are 2 issues, one that could be fixed by reverting my previous PR, and one about what we allow to input and possible conflicts.

E.g. for `blog`, with the current dev, we don't check `/blog` for a possible conflict, so anyway there is something else to fix

> I think we should definitely allow bog on principle as it's a great url ;)

If i understand what you mean, we also should not allow to input `blog/`

Or it would just result to create the `blog` url as it is done currently

If so, as an example, we would have just to revert my previous PR, and then just let this PR as it is, considering that `/blog` and also `blog/` and `/blog/` are possible conflicts of `blog`.

jtkech:
So the option i think is okay for now

- We can still input `blog`, `/blog`, `blog/` and `/blog/`, as it is already the case

- They all generate the same autoroute entry path `/blog`, as it is already the case

- **But we check that `/blog`, `blog/` and `/blog/` are possible conflicts of `blog`, as done in this PR.**

- Then we will just have to revert the change did in #8350 if we want

    `/blog` => the item

    `/blog/` => 404

themattkenney:
@jtkech what was decided on the trailing "/"?

jtkech:
With this PR we just added more possible conflicts while editing

So, currently both `blog` and `blog/` still serve the same page

It may change, but AFAIK nothing has been decided

Personnaly i'm fine with both solutions

If you have some pros and cons let us know ;)

themattkenney:
@jtkech  I am not fully understanding the use case that someone would want blog and blog/ to go to different pages. I would think that would cause confusion for end-users. 

However, if there is a requirement I would recommend we put a setting in the auto-route to allow the trailing "/" to be a separate page or the same page. With the default it to the same page.

jtkech:
@themattkenney Thanks for your answer

> I am not fully understanding the use case that someone would want blog and blog/ to go to different pages. I would think that would cause confusion for end-users.

Yes i agree, but anyway we would not allow to serve 2 different pages, that's why in this PR i added the missing paths to check for possible conflicts.

In fact the option is just to serve the same page with `blog` and `blog/`, as done in a previous PR

Or to return a 404 for `blog/`, so reverting the previous PR, but anyway not a different page

PR #8376, opened 2021-01-20
Default admin settings

agriffard:
Do not enable `DisplayDarkMode`, `DisplayMenuFilter `and `DisplayNewMenu `admin settings by default.

### DIFF
--- a/src/OrchardCore/OrchardCore.Admin.Abstractions/Models/AdminSettings.cs
+++ b/src/OrchardCore/OrchardCore.Admin.Abstractions/Models/AdminSettings.cs
@@ -2,7 +2,7 @@ namespace OrchardCore.Admin.Models
 {
     public class AdminSettings
     {
-        public bool DisplayDarkMode { get; set; }
+        public bool DisplayDarkMode { get; set; } = true;
 
         public bool DisplayMenuFilter { get; set; }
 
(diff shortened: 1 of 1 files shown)


Skrypt:
What was the previous reason to enable it by default? Because no one will know there is a setting for it "I think". Should we do the same with the darkmode then?

agriffard:
> Should we do the same with the darkmode then?

My preference would be indeed to not enable the Dark mode by default.

agriffard:
I made `DisplayDarkMode` also not enabled by default.

deanmarcussen:
Can you explain why?

agriffard:
> Can you explain why?

because @Skrypt thumbed up on my comment explaining that my preference is to not enable the settings by default, even DarkMode.

agriffard:
I am always surprised when I open a new site admin for the first time and it is dark (because it is based on my os settings). 
I'd let the choice to the admin to enable it (but I understand why @Skrypt did it because else, people might not know that the feature exists).

Skrypt:
Well it was more because that's how it is supposed to be designed. It's supposed to be affected by your current OS mode. Now, I added the ability to be able to toggle it and save it as a preference. But generally, the dark mode is a OS mode thing that browsers should react to automatically.

PR #8382, opened 2021-01-21
Move Autoroute RemovePart to content handler

deanmarcussen:
Follow on from here https://github.com/OrchardCMS/OrchardCore/pull/8204

Moving the removal of an autoroute part (when it is no longer defined on the content type definition) to the contenthandler base and tracking it for the index provider.

Makes the code a little trickier in the index provider bit, as now the `var part = contentItem.As<AutoroutePart>()` maybe null as it has already been removed. But tested and seems good.

Think it is the better place to remove this.

/cc @jtkech 

We should do the same for the localized content index, but it needs a little though on how to track the path = null, as that index doesn't have the same concept I suspect.


jtkech (approved):

jtkech:
> We should do the same for the localized content index, but it needs a little though on how to track the path = null, as that index doesn't have the same concept I suspect.

Yes, maybe by setting `LocalizationSet` or `Culture`, or both strings to null, maybe good to add it to this PR, unless it get merged before ;)

PR #8384, opened 2021-01-21
Move tags to own viewmodel

deanmarcussen:
Related to https://github.com/OrchardCMS/OrchardCore/issues/8368


urbanit:
@deanmarcussen everything seems fine except when creating a new tag inside blogpost - ``` Microsoft.Data.SqlClient.SqlException (0x80131904): The incoming request has too many parameters. The server supports a maximum of 2100 parameters. Reduce the number of parameters and resend the request.```

Probably it's not on changes you have made but we propagate the problem...

deanmarcussen:
@urbanit thanks for testing. I will merge it soon.

I assume you can't create a new tag in the Taxonomy editor either (but maybe not, it depends on how many content items are being updated at a time)

Probably hitting this https://github.com/sebastienros/yessql/issues/339

Can you add a comment there, when you find out more?

Good to have people using these new releases so we see where the issues are.

urbanit:
You are right, no Tag creation in Taxonomy Editor as well. Probably it's a problem with Autoroute Part (this is the last query on profiler)

PR #8391, opened 2021-01-21
Remove properties that cannot be exported from site settings

deanmarcussen:
Removes the `Properties` and `Identifier` value from the site settings export screen.

All properties should be exported through a seperate step.

If we're missing any steps the registration is simple for the step (no views or anything, we have a shared helper for them now).

/cc @Skrypt

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Settings/Views/Items/SiteSettingsDeploymentStep.Fields.Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Settings/Views/Items/SiteSettingsDeploymentStep.Fields.Edit.cshtml
@@ -3,6 +3,7 @@
 @using Newtonsoft.Json
 @{
     var settings = Model.Settings;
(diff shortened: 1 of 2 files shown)


src/OrchardCore.Modules/OrchardCore.Settings/Views/Items/SiteSettingsDeploymentStep.Fields.Edit.cshtml
```diff
@@ -3,6 +3,7 @@
 @using Newtonsoft.Json
 @{
     var settings = Model.Settings;
+    var ignoredPropeties = new HashSet<string> { "Identifier", "Properties" };
```
agriffard, line 6: ignoredPropeties => ignoredProperties

deanmarcussen: and that's why I shouldn't write pr's while trying to talk to y'all at triage

Skrypt (approved):

Issue #8324, opened 2021-01-15
Disable the CDN by default


sebastienros:
For better adoption, the CDN usages needs to be disabled by default. It should be opt-in in the admin.
And CDN sources should use the provider that has the most coverage, especially with coverage in China which has the most restrictions for now.

https://github.com/OrchardCMS/OrchardCore/pull/7599
https://github.com/OrchardCMS/OrchardCore/issues/7721

hyzx86:
> And CDN sources should use the provider that has the most coverage, especially with coverage in China which has the most restrictions for now.

Unfortunately, there are very few.😂
Such CDN servers can only be found from Chinese providers.

It is a good idea to disable it by default and then enable it manually by the site admin.

agriffard:
See also #8315 (Doc to explain how to disable the cdn programmatically) and #8154 (Attempt to use CloudFlare)

sebastienros:
In parallel, could someone from China make a module that updates the cdn urls with ones that work? Though I wouldn't want to have this module in this repository since most contributors can't test it, and there would be no way for us to ensure that these are not malicious versions. However if the hashes matched then that would be great.

sebastienros:
@agriffard I saw the PR, and if we fix this current issue then the PR should be the opposite, how to enable it by default.

PR #8398, opened 2021-01-22
Disable UseCdn by default

agriffard:
Related to #8324
Fixes https://github.com/OrchardCMS/OrchardCore/issues/8324


deanmarcussen:
I think it would be better just to set the `SiteSettings.UseCdn = false;` (and the same in `ResourceManagementOptions`)

agriffard:
> I think it would be better just to set the SiteSettings.UseCdn = false; (and the same in ResourceManagementOptions)

Ok, done.

deanmarcussen:
Cool, will you do the docs as well?

Just something simple like
```
CDNs are disabled by default by the Resource Manager.

This is to allow access to resources when an internet connection is not available, i.e. internal networks, or China.

It is recommended to enable the CDN setting after setup ... etc etc.
```

agriffard:
Section added to Setup documentation.

deanmarcussen (approved):

Issue #8410, opened 2021-01-23
can't load recipe with more than 2100 items per step


lampersky:
### Describe the bug
**Import Deployment Packege** and **JSON Import** can't hadle more than **2100** items per step when using **MS SQL server**.

`Version: 1.0.0-rc2-15572`

### To Reproduce
1. Setup **Blank Site** with **SQL Server** database,
2. Create new ContentType **ExampleType**,
3. Add Field of type TextField named **Name**,
4. Try to import attached recipe [reproRecipe.zip](https://github.com/OrchardCMS/OrchardCore/files/5859249/reproRecipe.zip),
5. Wait a bit for error page.

### Expected behavior
Such big file should be imported without any problems, maybe by chunking into smaller pieces.

### Screenshots
![too-many-parameters](https://user-images.githubusercontent.com/6403130/105561963-a8f29300-5d18-11eb-8eec-6ebf37d0373e.png)

### Exception Details

```
Microsoft.Data.SqlClient.SqlException (0x80131904): The incoming request has too many parameters. The server supports a maximum of 2100 parameters. Reduce the number of parameters and resend the request.
   at Microsoft.Data.SqlClient.SqlCommand.<>c.<ExecuteDbDataReaderAsync>b__188_0(Task`1 result)
   at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
--- End of stack trace from previous location ---
   at Dapper.SqlMapper.QueryAsync[T](IDbConnection cnn, Type effectiveType, CommandDefinition command) in /_/Dapper/SqlMapper.Async.cs:line 418
   at YesSql.Store.ProduceAsync[T](WorkerQueryKey key, Func`2 work, Object[] args)
   at YesSql.Services.DefaultQuery.Query`1.ListImpl()
   at OrchardCore.ContentManagement.DefaultContentManager.ImportAsync(IEnumerable`1 contentItems) in /home/runner/work/OrchardCore/OrchardCore/src/OrchardCore/OrchardCore.ContentManagement/DefaultContentManager.cs:line 453
[...]

PR #8413, opened 2021-01-23
Upgrade YesSql

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/8410

All databases CI run https://github.com/OrchardCMS/OrchardCore/actions/runs/505512669

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Workflows/Indexes/WorkflowIndexProvider.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Workflows/Indexes/WorkflowIndexProvider.cs
@@ -10,7 +10,7 @@ namespace OrchardCore.Workflows.Indexes
         public int DocumentId { get; set; }
         public string WorkflowTypeId { get; set; }
         public string WorkflowId { get; set; }
-        public int WorkflowStatus { get; set; }
+        public WorkflowStatus WorkflowStatus { get; set; }
         public DateTime CreatedUtc { get; set; }
     }
 
@@ -35,7 +35,7 @@ namespace OrchardCore.Workflows.Indexes
                         WorkflowTypeId = workflow.WorkflowTypeId,
                         WorkflowId = workflow.WorkflowId,
                         CreatedUtc = workflow.CreatedUtc,
-                        WorkflowStatus = (int)workflow.Status
+                        WorkflowStatus = workflow.Status
                     }
                 );
 
--- a/src/OrchardCore.Modules/OrchardCore.Workflows/Indexes/WorkflowIndexProvider.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Workflows/Indexes/WorkflowIndexProvider.cs
@@ -10,7 +10,7 @@ namespace OrchardCore.Workflows.Indexes
         public int DocumentId { get; set; }
         public string WorkflowTypeId { get; set; }
         public string WorkflowId { get; set; }
-        public WorkflowStatus WorkflowStatus { get; set; }
+        public int WorkflowStatus { get; set; }
         public DateTime CreatedUtc { get; set; }
     }
 
@@ -35,7 +35,7 @@ namespace OrchardCore.Workflows.Indexes
                         WorkflowTypeId = workflow.WorkflowTypeId,
                         WorkflowId = workflow.WorkflowId,
                         CreatedUtc = workflow.CreatedUtc,
-                        WorkflowStatus = workflow.Status
+                        WorkflowStatus = (int)workflow.Status
                     }
                 );
 
--- a/src/OrchardCore.Build/Dependencies.props
+++ b/src/OrchardCore.Build/Dependencies.props
@@ -42,12 +42,12 @@
     <PackageManagement Include="xunit.analyzers" Version="0.10.0" />
     <PackageManagement Include="xunit" Version="2.4.1" />
     <PackageManagement Include="xunit.runner.visualstudio" Version="2.4.3" />
-    <PackageManagement Include="YesSql.Abstractions" Version="2.0.0-beta-1738" />
-    <PackageManagement Include="YesSql.Core" Version="2.0.0-beta-1738" />
-    <PackageManagement Include="YesSql.Provider.MySql" Version="1.0.0-beta-1738" />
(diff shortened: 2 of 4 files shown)


src/OrchardCore.Modules/OrchardCore.Workflows/Indexes/WorkflowIndexProvider.cs
```diff
@@ -10,7 +10,7 @@ public class WorkflowIndex : MapIndex
         public int DocumentId { get; set; }
         public string WorkflowTypeId { get; set; }
         public string WorkflowId { get; set; }
-        public int WorkflowStatus { get; set; }
+        public WorkflowStatus WorkflowStatus { get; set; }
```
jtkech, line 13: I already tried it after the previous update of yessql, where i think the enum type was already supported

But it was not working as the related column of the index table is still a string

Note: I tried to define here a string instead, and doing a toString() below, it was working, but i think it is more natural to hold  here an enum with an int. Maybe worth to tell it to @sebastienros as it was working before, by defining here an enum, and with column of string type.

src/OrchardCore.Modules/OrchardCore.Workflows/Indexes/WorkflowIndexProvider.cs
```diff
@@ -35,7 +35,7 @@ public override void Describe(DescribeContext<Workflow> context)
                         WorkflowTypeId = workflow.WorkflowTypeId,
                         WorkflowId = workflow.WorkflowId,
                         CreatedUtc = workflow.CreatedUtc,
-                        WorkflowStatus = (int)workflow.Status
```
jtkech, line 38: Idem as above

src/OrchardCore.Modules/OrchardCore.Workflows/Indexes/WorkflowIndexProvider.cs
deanmarcussen: Hmm strange that it is a string in the index definition, that just feels wrong ;)

But also worked when I tried filtering by status. (with SQLite). I will check it with SQL Server.

I assume you were getting exceptions when you tried? No exceptions so far.

But also more type mapping has occured as well

jtkech: Need to leave i will check it again this night on sql server, if it works on your side

deanmarcussen: Seems fine on sql. I leave you to check it tonight.

For these indexes, it should really be an int. Maybe worth updating on your document id pr? Perhaps also lengths for the typeid's as well. seems pointless to be 256, but I leave that idea to you to decide

sfmskywalker (approved):

src/OrchardCore.Modules/OrchardCore.Workflows/Indexes/WorkflowIndexProvider.cs
jtkech: @deanmarcussen 

Okay cool it works on my side, maybe enum was not yet supported when i tried it

jtkech (approved):

sebastienros:
Enums will fail on indexes with postgres in this version. The next nuget version is fixing it.

src/OrchardCore.Modules/OrchardCore.Workflows/Indexes/WorkflowIndexProvider.cs
jtkech: @deanmarcussen 

> For these indexes, it should really be an int. Maybe worth updating on your document id pr?

Yes but e.g. with sqlite we can't alter a column in a migration step

> Perhaps also lengths for the typeid's as well. seems pointless to be 256, but I leave that idea to you to decide

Yes, i thought about this, but we can't alter a column length with sqlite in a migration step, and based on my new findings it seems to be easier that all new indexes can also be created for existing databases with the old lengths, and knowing that the number of rows in index pages only depends on the actual columns values sizes.

But yes, we could also change the columns sizes at least for new installations, but it open the possibility to define in the future a new index that can be created on a new installation, but would fail on a migrtion step.

But not against the idea ;)

sebastienros:
I updated this PR with the latest version and also ran the functional tests.

deanmarcussen:
Hmm, still getting properties > 2100 errors. Will investigate

sebastienros:
Updated with 2098

jtkech:
@deanmarcussen Just for info about the `WorkflowSatus` enum

So for now it still work in the dev branch by defining it as an int in the index provider, and casting to an int the enum of the WF. But in the `DocumentId` PR we add an index to the `WorkflowIndex` table including the `WorkflowSatus` (defined as a string column), and the functional tests on `PostgreSQL` fails.

So in the  `DocumentId` PR, for now i defined it as a string in the index provider, and apply a `ToString()` on the enum of the WF. So, as a reminder, if this PR gets merged first, i will have to revert this change and test if it still fails or not.

sebastienros:
@jtkech enums should be mapped to int. I think it won't work with strings now.
Also, for safety the enums should be initialized with a specific number, or each value, such that if the order changes, the value won't.

jtkech:
@sebastienros 

Okay but the index table is created with a `WorkflowStatus` column of type string

And we can't alter the column type e.g. with sqlite in a migration step

Doing a `ToString()` in the mapping seems to work, the PostgreSQL func test pass

And i will try it with concrete WF instances

deanmarcussen:
OK, looks good, found a bug in the `Import` routine in the content manager as well. (was adding to the Take, when it shouldn't, so with many queries rose to 2000, for the `IsIn` which jumped over the 2100 parameters limit, as `IsIn` is used twice.

@jtkech @sebastienros for infos.

I reverted the Workflow changes to this pr. 
- Not a major issue
- Needs a bit more thought and if useful can be done on the `DocumentId` pr I think @jtkech 
- The problem of it being stored as a string, isn't great, but it isn't major.

Merging this when it passes the functional tests, as it is fixing the bigger issue ;)

P.S. yes default number on enums is useful, often `Finished=10, Fault = 20` etc al, is useful, just to allow room to place a new one in the middle. Too late for that now though I imagine ;)

Issue #7852, opened 2020-12-04
Bug: Exporting a Deployment Plan that includes the Admin settings step produces wrong json


serverside-is:
![image](https://user-images.githubusercontent.com/3681658/101177599-e81d3580-363f-11eb-9d1e-9a2b70fd9efc.png)

Adding the `Admin Settings` step and the `Deployment Plans`  to a `Deployment Plan`  (as in the above picture), and downloading the json file, produces something like this:

```
[...]
          "Steps": [
            {
              "Type": "SiteSettingsPropertyDeploymentStep`1",
              "Step": {
                "Id": "6582efaa9f7e4246a4722320fa19ee2c",
                "Name": "AdminSettings_SiteSettingsPropertyDeploymentStep"
              }
            },
[...]
```
When we import this package, we get:

```
InvalidOperationException: No changes have been made. The following types of deployment plans cannot be added: SiteSettingsPropertyDeploymentStep`1. Please ensure that the related features are enabled to add these types of deployment plans.
OrchardCore.Deployment.Recipes.DeploymentPlansRecipeStep.ExecuteAsync(RecipeExecutionContext context) in DeploymentPlansRecipeStep.cs, line 74
```
This seems to work just fine:

```
[...]
          "Steps": [
            {
              "Type": "AdminSettings_SiteSettingsPropertyDeploymentStep",
              "Step": {
                "Id": "6582efaa9f7e4246a4722320fa19ee2c",
                "Name": "AdminSettings_SiteSettingsPropertyDeploymentStep"
              }
            },
[...]
```

To produce the second result, I did this modification to DeploymentPlanDeploymentSource.cs (OrchardCore.Deployment module):

```
-- Type = step.GetType().Name,
++ Type = (step.GetType().GenericTypeArguments.Count() == 0)?step.GetType().Name:step.Name,
```

Although this is working for me, I'm not creating a PR because I'm not sure if this has other implications.

deanmarcussen:
Yes, your code will probably work, but later implications...

What the `Source needs to do is a similar action to the `RecipeStep` 
https://github.com/OrchardCMS/OrchardCore/blob/dedf32dbb60573f1cf48addc0a94f126f42f3010/src/OrchardCore.Modules/OrchardCore.Deployment/Recipes/DeploymentPlansRecipeStep.cs#L53

where it will need to find the factory that matches the `Name` and asks the factory for it's type.

[...]

PR #8419, opened 2021-01-23
Fix Step Type derivation for SiteSettings<T> Steps

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/7852


agriffard:
Build failed:
`OrchardCore.Deployment/Deployment/DeploymentPlanDeploymentSource.cs(34,60): error CS1061: 'IServiceProvider' does not contain a definition for 'GetServices' and no accessible extension method 'GetServices' accepting a first argument of type 'IServiceProvider' could be found (are you missing a using directive or an assembly reference?) `

src/OrchardCore.Modules/OrchardCore.Deployment/Deployment/DeploymentPlanDeploymentSource.cs
```diff
@@ -1,16 +1,23 @@
...
 
 namespace OrchardCore.Deployment.Deployment
 {
     public class DeploymentPlanDeploymentSource : IDeploymentSource
     {
         private readonly IDeploymentPlanService _deploymentPlanService;
+        private readonly IServiceProvider _serviceProvider;
 
-        public DeploymentPlanDeploymentSource(IDeploymentPlanService deploymentPlanService)
+        public DeploymentPlanDeploymentSource(
```
hishamco, line 15: Is there prevent from injecting `IEnumerable<IDeploymentStepFactory>>`?

deanmarcussen: again they are transient so yes they are preventing from being injected by their nature.

Skrypt: IDeploymentStepFactory is a singleton.

Skrypt: And services.AddTransient<IDeploymentSource, DeploymentPlanDeploymentSource>();
So it will always have a different value.

jtkech: The main rule is that a singleton can't inject a scoped. A transient can inject anything but it is not recommanded that it injects a scoped, because the transient may be injected by a singleton (but it is then detected at runtime).

So, normally a transient `IDeploymentSource` can inject `IDeploymentStepFactory` singletons. What is wrong currently is that a transient `IDeploymentSource` injects a scoped `IDeploymentPlanService`, but that's okay as long as this transient is not injected by a sigleton.

deanmarcussen: So I agree @jtkech 

I've reversed the resolutions, so that `IDeploymentPlanService` is resolved. Obviously not a major, as the transient isn't grabbed by a singleton, but it just felt better to fix.

jtkech: Okay cool

jtkech: @deanmarcussen 

When i said it's "wrong", i only meant that we have to take care of it

**So, if we think that this transient can't be injected by a singleton** (directly or indirectly), only by a scoped.

- This transient could still inject the scoped service, as we do with our tag helpers that can be considered as transients, see my comment https://github.com/OrchardCMS/OrchardCore/pull/8421#discussion_r566553905

    But it may be still useful to resolve the scoped service lazily by using the injected `IServiceProvider`, but here only if we want to resolve it on demand, only if we need it, but not to fix a DI lifetime issue.

    Note: As you know, injecting a service provider can also be used to solve a circular dependency

**If we think that this transient can be injected by a singleton** (directly or indirectly)

- We should resolve the scoped service lazily, but by using the right service provider, not by using the injected `IServiceProvider` that would be the one used when resolving the parent singleton.

    So, to fix a potential DI issue, we would have to retrieve the right service provider by using e.g. `RequestServices` which in fact, in our case, returns the service provider of the current shell scope.

So as you want, if we are in the first case we could just inject all services, or maybe just keep the code as you already did, maybe just remove or update the related comment. But in that case, why not registering these `IDeploymentSource` transients as scoped services instead. But if we are in the second case, using the injected service provider doesn't solve the DI concern. At least for now it seems that we are in the first case, otherwise we would already had DI issues

deanmarcussen: Yeah I think we're probably on option 1 here, I'll have a look over the weekend. But probably revert to how it was.

`IDeploymentSoure` as scoped. I suspect this might make more sense. Will look further, but perhaps a different pr.

jtkech: Yes, no problem, anyway if a singleton reference e.g. a scoped through a transient, it is detected at runtime.

deanmarcussen: Back to as before. 
`IDeploymentSource` could go to scoped, but it is not hurting anything currently. Bigger things to worry about ;)

PR #8424, opened 2021-01-25
Upgrade to Fluid 2.0

sebastienros:
Tested TheBlog, TheAgency and ComingSoon themes.

And while doing the migration, realized some tag could have used other constructs, like identifiers instead of strings sometimes. Or instead of accepting a list or arguments, named or not, have the first be "something" and the rest be named arguments ... well, can't be perfect on the first try ;)


src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Tags/FluidTagHelper.cs
```diff
@@ -0,0 +1,118 @@
...
+            // e.g., {% helper "input", for: "Text", class: "form-control" %}
+
+            identifier = identifier ?? (await arguments[0].Expression.EvaluateAsync(context)).ToStringValue();
+
+            // These mapping will assign an argument name to the first element in the filter arguments,
+            // such that the tag helper can be matched based on the expected attribute names.
+            if (DefaultArgumentsMapping.TryGetValue(identifier, out var mapping))
+            {
+                arguments = new List<FilterArgument>(arguments);
+                arguments[0] = new FilterArgument(mapping, arguments[0].Expression);
```
lahma, line 51: Should this be `.Add`?

lahma: Ah it's initialized, with arguments, probably it has always > 0 items and cannot use the original list

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewParser.cs
```diff
@@ -0,0 +1,80 @@
...
+            RegisterParserTag("shape_remove_item", Primary.And(Primary), ShapeRemoveItemTag.WriteToAsync);
+            RegisterParserTag("shape_add_properties", Primary.And(ArgumentsList), ShapeAddPropertyTag.WriteToAsync);
+            RegisterParserTag("shape_remove_property", Primary.And(Primary), ShapeRemovePropertyTag.WriteToAsync);
+            RegisterParserTag("shape_pager", Primary.And(ArgumentsList), ShapePagerTag.WriteToAsync);
+
+            RegisterParserTag("httpcontext_add_items", ArgumentsList, HttpContextAddItemTag.WriteToAsync);
+            RegisterParserTag("httpcontext_remove_items", Primary, HttpContextRemoveItemTag.WriteToAsync);
+
+            RegisterParserTag("helper", ArgumentsList, FluidTagHelper.WriteArgumentsTagHelperAsync);
+            RegisterParserTag("shape", ArgumentsList, async (list, writer, encoder, context) => await FluidTagHelper.WriteToAsync("shape", list, null, writer, encoder, context));
```
lahma, line 57: does langversion allow using static lambdas? would make sure no accidental closure allocations happen

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Tags/ClearAlternatesTag.cs
```diff
@@ -3,13 +3,12 @@
...
 using Fluid.Ast;
-using Fluid.Tags;
 
 namespace OrchardCore.DisplayManagement.Liquid.Tags
 {
-    public class ClearAlternatesTag : ExpressionTag
+    public class ClearAlternatesTag
     {
-        public override async ValueTask<Completion> WriteToAsync(TextWriter writer, TextEncoder encoder, TemplateContext context, Expression expression)
+        public static async ValueTask<Completion> WriteToAsync(Expression expression, TextWriter writer, TextEncoder encoder, TemplateContext context)
```
lahma, line 11: Some of these tags have really simple body and might benefit from the `return Awaited()` trick to remove async state machines, beneficial when tags are evaluated a lot and common case is that task is already resolved

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Tags/RenderBodyTag.cs
```diff
@@ -4,14 +4,13 @@
...
 namespace OrchardCore.DisplayManagement.Liquid.Tags
 {
-    public class RenderBodyTag : SimpleTag
+    public class RenderBodyTag
     {
-        public override async ValueTask<Completion> WriteToAsync(TextWriter writer, TextEncoder encoder, TemplateContext context)
+        public static async ValueTask<Completion> WriteToAsync(TextWriter writer, TextEncoder encoder, TemplateContext context)
         {
             if (!context.AmbientValues.TryGetValue("ThemeLayout", out dynamic layout))
             {
```
lahma, line 16: static methods could be inline candidates when there's no throw clause (would require ThrowHelper), I think this is too large though to be considered, but generally tThrowHelper is a good thing

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Tags/ShapePagerTag.cs
```diff
@@ -1,75 +1,72 @@
...
 using OrchardCore.Mvc.Utilities;
 
 namespace OrchardCore.DisplayManagement.Liquid.Tags
 {
-    public class ShapePagerTag : ExpressionArgumentsTag
+    public class ShapePagerTag
     {
         private static readonly HashSet<string> _properties = new HashSet<string>
         {
             "Id", "PreviousText", "NextText", "PreviousClass", "NextClass", "TagName", "ItemTagName"
```
lahma, line 19: Would it make sense to also have the expected "raw" before conversion using `ToPascalCaseUnderscore` value here to match without need to always evaluate name with that function inside the foreach?

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Tags/ShapePagerTag.cs
```diff
@@ -1,75 +1,72 @@
...
                             foreach (var value in values)
                             {
                                 objectValue.ItemClasses.Add(value);
                             }
                         }
-                        else if (classes.Type == FluidValues.Array)
+                        else if (itemClasses.Type == FluidValues.Array)
                         {
-                            foreach (var value in classes.Enumerate())
+                            foreach (var value in itemClasses.Enumerate())
```
lahma, line 60: Might make sense to open up some of the helpers I introduced as internal, like `ToStringList()` or similar, could use the return value automatically and get correct size list to use

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Cache/CacheTag.cs
```diff
@@ -119,13 +117,13 @@ public override async ValueTask<Completion> WriteToAsync(TextWriter writer, Text
...
 
-        private async Task<string> EvaluateStatementsAsync(TextEncoder encoder, TemplateContext context)
+        private static async Task<string> EvaluateStatementsAsync(IReadOnlyList<Statement> statements, TextEncoder encoder, TemplateContext context)
         {
             using (var sb = StringBuilderPool.GetInstance())
             {
                 using (var content = new StringWriter(sb.Builder))
                 {
-                    foreach (var statement in Statements)
+                    foreach (var statement in statements)
```
lahma, line 126: for loop doesn't allocate the boxed enumerator that is needed to enumerate the collection interface, micro optimization

Skrypt:
Hmm, started analysing the code, got distracted. Looks like a lot of breaking changes that could potentially happen. Will try to make time to test it.

sebastienros:
@Skrypt ideally you just need to try this branch on your site and check they still work

I don't think there should be any breaking changes, if not for {%- -%} that are necessary to remove the trailing whitespace if you want.
The other known breaking change is that you can do `{{ assign a = 1 + 2 }}`, it has to be `{{ assign a = 1 | plus: 2 }}` which is the Liquid spec.

Skrypt:
Yeah, exactly the idea I had, testing this on my website locally. Should be a good test.

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewParser.cs
```diff
@@ -0,0 +1,80 @@
...
+            RegisterParserTag("shape_tab", Primary.And(Primary), ShapeTabTag.WriteToAsync);
+            RegisterParserTag("shape_remove_item", Primary.And(Primary), ShapeRemoveItemTag.WriteToAsync);
+            RegisterParserTag("shape_add_properties", Primary.And(ArgumentsList), ShapeAddPropertyTag.WriteToAsync);
+            RegisterParserTag("shape_remove_property", Primary.And(Primary), ShapeRemovePropertyTag.WriteToAsync);
+            RegisterParserTag("shape_pager", Primary.And(ArgumentsList), ShapePagerTag.WriteToAsync);
+
+            RegisterParserTag("httpcontext_add_items", ArgumentsList, HttpContextAddItemTag.WriteToAsync);
+            RegisterParserTag("httpcontext_remove_items", Primary, HttpContextRemoveItemTag.WriteToAsync);
+
+            RegisterParserTag("helper", ArgumentsList, FluidTagHelper.WriteArgumentsTagHelperAsync);
```
deanmarcussen, line 56: So i imagine to move away from the tag helper bridge, and allow other modules to register tags, this class would need to be constructed from DI, and instanced, so that registration could happen, through an `IOptions` arrangement?

It would be a much nicer syntax, not to be using the helper

sebastienros: Note you can already do it in this branch, since registering a tag is only about calling `_liquidViewParser.RegisterTag` (tag as in liquid tag). With this branch the tenant has its own parser instance that can be configured.

But the main issue I have with Liquid Tag -> Tag Helpers is this bridge. I'd prefer the `WriteAsync` method to just be invoking the correct code in the first place, instead of dynamically looking for a tag helper, instantiate it, map data structures, and then get back the results to write it again ... That should be doable one tag at a time, and would improve perf a lot.

sebastienros: BTW I am also taking not as I look at traces on how to improve the way we pass all these services in each new template context. (IUrlHelper, DisplayHelper, ...). And I have an idea on how to fix it. And also remove the static accesses to TemplateContext such that each tenant can have their own config. Statics are bad :/

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewTemplate.cs
```diff
@@ -23,86 +22,25 @@
...
-using OrchardCore.DynamicCache.Liquid;
 using OrchardCore.Environment.Shell.Scope;
 using OrchardCore.Liquid;
 
 namespace OrchardCore.DisplayManagement.Liquid
 {
-    public class LiquidViewTemplate : BaseFluidTemplate<LiquidViewTemplate>
+    public class LiquidViewTemplate
     {
+        public static LiquidViewParser Parser = new LiquidViewParser();
```
deanmarcussen, line 34: I'm looking for this `_liquidViewParser.RegisterTag` but this static was all I found ;)

sebastienros: LiquidTemplate.Parser, yeah still static, need to remove LiquidTemplate altogether for this change. Will do with the other things I mentioned should be improved. But it can be done now, before it couldn't... at least

jtkech:
@sebastienros @Skrypt @deanmarcussen 

I tried it at least with TheBlogTheme, looks good

- About tag helpers, yes we use them intensively, and i agree that for each of our custom tag helper, we could have a liquid tag in place of using the bridge, even we will have to maintain each tag helper and its related liquid tag. But i think it is still useful to keep the bridge, so that it still allows to use in liquid any other mvc tag helper.

    I asked me how tag helpers behave but under razor views, i think the perf impact is lower because views are pre-compiled, as i remember they have been fully integrated in the razor language project.

- About perf, **on my local machine** it's around 800 rqs for the about page, but less than 600 rqs for the home, so i started to check how it was behaving 1.5 year ago, for the about page it was quite the same around 800, but around 700 rqs for the home page, i don't remember when it was around 1000 rqs, at least for the about page. So it seems, at least on my machine, that it is quite the same for the about page since a long time, **but it is now mainly slower for the home**.

I will continue to go back in time to see when it started to happen.

jtkech:
About perfs, just to say that we now have many more cool features, taxonomies, shortcodes, sanitizing, image resizing tokens, and so on,  so it's normal that perfs has relatively "decreased".

About the helper tag, i started to do some traces about **cpu resources**, it seems that most of the resources are not consumed to instanciate a given tag helper, but to process the tag helper itself (see the image below), knowing that this tag helper can be e.g. a shape tag helper. But yes, tomorrow need to **focus on allocations** ;)

I also quickly saw some other costs (still only on cpu times, not on allocations yet), but need to be tested again, around ContentItemConverter ReadJson(), ContentTypeDefinition GetSettings(), ContentManager Get(), HtmlHelper Partial(), ShapeFactory CreateShape(), MemoryCache TryGetValue(), ImageSharp, AngleSharp, to be continued.

Performance

jtkech:
@sebastienros Just for infos as i just saw the last meeting video and noticed 2 or 3 little things.

- About `LiquidTemplateContextExtensions` where we resolve services and call all `ILiquidTemplateEventHandler`.

    **We already do this once per scope** because of this line https://github.com/OrchardCMS/OrchardCore/blob/4c71d093a728d726bea24e0024ad9b58a9749aa8/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewTemplate.cs#L316

    Yes, we could just pass `Services`, which is already a property of the `LiquidTemplateContext`, but not of the fluid `TemplateContext` from which it derives. So a filter could just use the `Services` ambient value, or the above context property, or maybe just `ShellScope.Services` but i know you don't like it ;)

    That said, `UrlHelper` allows a filter to not have to resolve `IUrlHelperFactory` to build an `UrlHelper` based on the current view context, `ThemeLayout` to not have to resolve and use `ILayoutAccessor`.

    Yes, doing it lazily would be still useful, but maybe some as `UrlHelper` are used most of the time, and resolved services are cached on aspnetcore side, so we are seeing them here, maybe just because this is where they are first resolved.

    Then the model and `ViewLocalizer` (contextualized with the view context) are taken into account per template

- About the `ArrayPool`, as i remember what i saw in the diff screen, you were creating an `array`, but you were not using it, you were still adding `buffer.ToArray()` to the `_fragments`, but i may be wrong ;)

Skrypt:
From a first test by merging this on top of my website I get an issue.

`{{ Model.Content | shape_render }}` doesn't work even if the shape is there.

sebastienros:
Thanks, I still have more work to do, and will also try the shape_render filter.

sebastienros:
All tests pass. Please test and send some feedback on the code.

src/OrchardCore.Modules/OrchardCore.Liquid/Startup.cs
```diff
@@ -19,29 +21,36 @@ namespace OrchardCore.Liquid
...
+                // Convert JToken to FluidValue
+                options.ValueConverters.Add(x => x is JObject o ? new ObjectValue(o) : null);
+                options.ValueConverters.Add(x => x is JValue v ? v.Value : null);
+                options.ValueConverters.Add(x => x is DateTime d ? new ObjectValue(d) : null);
+
+                options.Filters.AddFilter("json", JsonFilter.Json);
+                options.Filters.AddFilter("jsonparse", JsonParseFilter.JsonParse);
+            })
+            .AddLiquidFilter<TimeZoneFilter>("local")
+            .AddLiquidFilter<SlugifyFilter>("slugify")
```
deanmarcussen, line 47: ok, so as we discussed, I've moved all the filters that use DI services to classes, which are instaniated by the `LiquidFilterResolver` as required, with services from the constructor.

I've left all the filters that don't use DI, like the above `JsonParseFilter` as statics.
As that seemed like a good idea ;)

src/OrchardCore/OrchardCore.Liquid.Abstractions/LiquidFilterResolver.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace OrchardCore.Liquid
+{
+    public class LiquidFilterResolver<TLiquidFilter> where TLiquidFilter : class, ILiquidFilter
+    {
+        public ValueTask<FluidValue> ResolveThenProcessAsync(FluidValue input, FilterArguments arguments, TemplateContext context)
+        {
+            var services = ((LiquidTemplateContext)context).Services;
+            var filter = services.GetRequiredService<TLiquidFilter>();
+
+            return filter.ProcessAsync(input, arguments, context);
```
deanmarcussen, line 16: Just resolves filter by type from DI container. then runs it.

src/OrchardCore/OrchardCore.Liquid.Abstractions/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace OrchardCore.Liquid
+{
+    public static class ServiceCollectionExtensions
+    {
+        public static IServiceCollection AddLiquidFilter<TLiquidFilter>(this IServiceCollection services, string name) where TLiquidFilter : class, ILiquidFilter
+        {
+            // services.Configure<TemplateOptions>(o => o.Filters.AddFilter(name, new LiquidFilterResolver(typeof(T)).ResolveThenProcessAsync));
+            services.Configure<TemplateOptions>(o => o.Filters.AddFilter(name, new LiquidFilterResolver<TLiquidFilter>().ResolveThenProcessAsync));
+            services.AddTransient<TLiquidFilter>();
```
deanmarcussen, line 12: I've added them as `Transient`. They could equally be `Scoped` I think.

Need to decide on what is best there.

Went with `Transient` as @jtkech has confirmed that all the `TagHelpers` are always transient, so trying to keep some consistency. However `TagHelpers` need to be transient because they do bindings, whereas liquid filters, don't bind. So maybe scoped is advantageous.

Thoughts @jtkech ?

test/OrchardCore.Benchmarks/ShapeFactoryBenchmark.cs
```diff
@@ -45,17 +45,19 @@ static ShapeFactoryBenchmark()
             _templateContext.AmbientValues["DisplayHelper"] = new DisplayHelper(null, shapeFactory, null);
         }
 
+        // TODO this benchmark is meaningless as the benchmark noops as the input is not a shape.
```
deanmarcussen, line 48: usless benchmark, it nooped before running.

test/OrchardCore.Benchmarks/FluidShapeRenderBenchmark.cs
```diff
@@ -0,0 +1,161 @@
...
+        //   [Host]     : .NET Core 3.1.8 (CoreCLR 4.700.20.41105, CoreFX 4.700.20.41903), X64 RyuJIT
+        //   DefaultJob : .NET Core 3.1.8 (CoreCLR 4.700.20.41105, CoreFX 4.700.20.41903), X64 RyuJIT
+
+
+        // |                       Method |     Mean |     Error |    StdDev | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
+        // |----------------------------- |---------:|----------:|----------:|------:|--------:|-------:|------:|------:|----------:|
+        // |   OriginalShapeRenderDynamic | 3.798 us | 0.0663 us | 0.1108 us |  1.00 |    0.00 | 2.2888 |     - |     - |   9.37 KB |
+        // | ShapeRenderWithAmbientValues | 3.156 us | 0.0278 us | 0.0232 us |  0.82 |    0.03 | 2.2240 |     - |     - |   9.09 KB |
+        // |            ShapeRenderStatic | 3.180 us | 0.0253 us | 0.0224 us |  0.83 |    0.03 | 2.1935 |     - |     - |   8.96 KB |
+        // |      ShapeRenderWithResolver | 3.164 us | 0.0206 us | 0.0161 us |  0.83 |    0.03 | 2.1973 |     - |     - |   8.98 KB |
```
deanmarcussen, line 51: so resolving adds a tiny bit of allocation (for the instance), but constructor is marginally faster than `serviceprovider.GetService()`

Changed these benchmarks from what you saw yesterday to simulate creating a template context as well. (as we would have done each liquid render).

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/RequestLiquidTemplateEventHandler.cs
```diff
@@ -1,15 +1,13 @@
...
 using System.Threading.Tasks;
 using Fluid;
-using Fluid.Values;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.DependencyInjection;
 using OrchardCore.Liquid;
 
 namespace OrchardCore.DisplayManagement.Liquid
 {
+    // TODO this code remains for backwards compatability. It can be removed in a futre release.
```
deanmarcussen, line 10: I think we can probably drop this in favour of just injecting.
all it's doing now is this (and was used in our code in one place - the `SwitchCultureUrlFilter` which I've removed.

```
          if (_httpContext != null)
            {
                context.SetValue("Request", _httpContext.Request);
                context.SetValue("HttpContext", _httpContext);
            }
```

deanmarcussen (commented):
Made liquid filters constructable - where they used DI.

Marginally faster, on a micro bench, but really about making them syntactically nicer, and therefore easier to understand and implement your own filters.

Done in one commit (with a merge conflict), so easy to drop if it performs way worse in the real world.

src/OrchardCore/OrchardCore.Liquid.Abstractions/ServiceCollectionExtensions.cs
jtkech: @deanmarcussen 

As liquid filters, i think it is better to register them as scoped services so that the services are injected once per scope, unless for some if they need to hold something that would need to be different per instance, but i don't think so, e.g. as the view context for a given tag helper (if it has this property), and as you said because of they bound data.

jtkech:
@sebastienros 

As i have tried it works fine, looks good!

I commented out the properties of `LiquidContentAccessor` that are now useless, i added them through #8259 to fix #8241 that will need to be fixed differently, hmm maybe good to re-open the related issue.

And in `TemplateShortcodeProvider` i added the missing shortcode `Content` parameter that we still need to pass to the liquid template manager parameters, check if i did it correctly

            parameters["Content"] = new StringValue(model.Content);

Just for info i noticed that the member access strategy of `LiquidUserAccessor` is commented out, maybe still in progress

src/OrchardCore.Modules/OrchardCore.Shortcodes/Services/TemplateShortcodeProvider.cs
```diff
@@ -47,27 +54,27 @@ public async ValueTask<string> EvaluateAsync(string identifier, Arguments argume
...
-                    {
-                        contentAccessor.Content = model.Content ?? "";
-                        scope.SetValue("Content", contentAccessor);
-                    }
-                    else
-                    {
-                        scope.SetValue("Content", model.Content ?? "");
-                    }
+            parameters["Args"] = new ObjectValue(model.Args);
+            parameters["Content"] = new StringValue(model.Content);
```
jtkech, line 74: I added this line as we still need to pass the Content property

src/OrchardCore/OrchardCore.Liquid.Abstractions/LiquidContentAccessor.cs
```diff
@@ -5,8 +5,10 @@ namespace OrchardCore.Liquid
     /// </summary>
     public class LiquidContentAccessor
     {
-        public string Content { get; set; }
+        // TODO: Fix 'Content' property conflict differently, see #8259
 
-        public override string ToString() => Content;
+        //public string Content { get; set; }
+
+        //public override string ToString() => Content;
```
jtkech, line 12: I commented these props that are now useless

src/OrchardCore.Modules/OrchardCore.Users/Startup.cs
```diff
@@ -205,13 +208,44 @@ public class LiquidStartup : StartupBase
...
+                    //    nameof(User.UserId) => user.UserId,
+                    //    nameof(User.UserName),
+                    //    nameof(User.NormalizedUserName),
+                    //    nameof(User.Email),
+                    //    nameof(User.NormalizedEmail),
+                    //    nameof(User.EmailConfirmed),
+                    //    nameof(User.IsEnabled),
+                    //    nameof(User.RoleNames),
+                    //    nameof(User.Properties)
+                    //_ => NilValue.Instance
```
jtkech, line 239: Still in progress i assume, as they are commented out

src/OrchardCore/OrchardCore.Liquid.Abstractions/ServiceCollectionExtensions.cs
deanmarcussen: @jtkech cool. I debated with myself about this.

On the one hand, liquid template manager is scoped. On the other hand, it maybe that during a request 3 different templates are rendering. (more on the admin tbf, but also with liquid parts or the liquid filter, where inside a liquid page, you might render another liquid template in the same request). 

So scoped is fine, so long as no one tries to hold scoped values in them - but where we have done recursion checking, we used a scoped helper anyway, so it's abstracted out. 

All good to go with scoped.

jtkech: @deanmarcussen 

Okay, so in `AddLiquidFilter()` we can use `AddScoped<>()`

But in `AddTagHelpers()` we still need `AddTransient<>()`

I will do the change

deanmarcussen: Cool. I just tweaked it to remove my commented out code.

jtkech: I hesitated to do it myself ;)

jtkech:
@deanmarcussen

Just for infos

Forgot to say that for the scoped liquid filters, we can't just inject services that need to be contextualized from the viewContext, we need to also contextualize them, will check if that's okay.

jtkech:
@deanmarcussen @sebastienros 

> I'm not sure we can change this to IShape safely

> The DefaultHtmlDisplay handles IHtmlContent as well (and turns objects that are neither into .ToString()

> Recently the ShapeRenderFilter was changed from only IShape to any object, because it wouldn't render shape.Metadata.ChildContent, i.e. Wrappers

I took a look at the `ThemeLayout` that has been updated by @sebastienros, the pre-rendered zones are now wrapped in a `PositionWrapper` that is an `IShape` and can be instanciated by passing an `IHtmlContent` or a `string`

And i updated the `ShapeRenderFilter` to check for an `IShape`, `IHtmlContent` or a string and e.g. in the last case to return a `StringHtmlContent`.

Maybe we also need to update the following in `RazorPage.cs` and `Page.cs`, i will check it

        public Task<IHtmlContent> DisplayAsync(dynamic shape)
        {
            EnsureDisplayHelper();
            return _displayHelper.ShapeExecuteAsync(shape);
        }

sebastienros:
TODO: CoerceHtmlString is called on ZoneOnDemand because some zod == null don't work. It's harmless but it should be fixed.

sebastienros:
@jtkech I found what you meant with ChildContent, someone I won't name who lives in Chicago seems to use DisplayAsync() with many things that are not shapes. Even IHtmlContent which can be rendered directly. I added some code to handle these case though we should remove the usages IMO.

sebastienros:
All tests are passing. We should get someone to test it on their site, fix any bug found, and merge it.

Skrypt:
Ok @deanmarcussen asked me to test this. Will do it right now.

deanmarcussen:
Updated `LiquidFilterDelegateResolver` to send the `LiquidTemplateContext` casted to the `ILiquidFilter`. 

Just for convenience

src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
```diff
@@ -80,6 +80,35 @@ public IShapeFactory Factory
...
+                else if (shape is string str)
+                {
+                    return Task.FromResult<IHtmlContent>(new HtmlString(HtmlEncoder.Encode(str)));
+                }
+                else
+                {
+                    throw new ArgumentException("DisplayAsync requires an instance of IShape");
+                }
+            }
+        }
```
jtkech, line 105: At some point i did something like this but reverted it as implemented by `ShapeExecuteAsync()` => `DefaultHtmlDisplay.ExecuteAsync()` where i fixed the use case of `ZoneOnDemand` not considered as null if passed as an `IShape`. But yes, here it is done immediately and returns something if it is a string.

But why not keeping the same logic as in `DefaultHtmlDisplay.CoerceHtmlString()`, e.g. using a `StringHtmlContent` and not throwing if neither an `IHtmlContent` nor a string, here the copy of `CoerceHtmlString()`.

        private static IHtmlContent CoerceHtmlString(object value)
        {
            // So here just need to check if null as ZoneOnDemand is an IShape
            if (value == null || value is ZoneOnDemand)
            {
                return HtmlString.Empty;
            }

            if (value is IHtmlContent result)
            {
                return result;
            }

            // Convert to a string and HTML-encode it <= Remove / Correct the comment saying it will be encoded.
            return new StringHtmlContent(value.ToString());
        }

sebastienros: I'd like that we find where it's not correctly used and fix it. Next step for instance would be to use strongly typed models in the forms for instance. And finally remove DisplayAsync(dynamic). Or at least have an "object" overload to handle explicitly "anything", and not "magic"

deanmarcussen:
That pager shape causes so much pain. Maybe it is time it became a PagerShapeViewModel

sebastienros:
@deanmarcussen which would also contain shape view models for its elements

src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
jtkech: @sebastienros

Okay, what i also suggested here (also in `Page.cs`), regarding what we do in other places, is something like this

        public Task<IHtmlContent> DisplayAsync(dynamic shape)
        {
            if (shape is IShape s)
            {
                EnsureDisplayHelper();
                return _displayHelper.ShapeExecuteAsync(s);
            }

            if (shape == null)
            {
                return Task.FromResult<IHtmlContent>(HtmlString.Empty);
            }

            if (shape is IHtmlContent hc)
            {
                return Task.FromResult(hc);
            }

            // Not necessary as managed below, but maybe faster
            // if (shape is string str)
            // {
            //    return Task.FromResult<IHtmlContent>(new StringHtmlContent(str));
            // }

            return Task.FromResult<IHtmlContent>(new StringHtmlContent(shape.ToString()));
        }

jtkech: At least maybe use `new StringHtmlContent(str)`
in place of `new HtmlString(HtmlEncoder.Encode(str)`

src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
```diff
@@ -80,6 +80,35 @@ public IShapeFactory Factory
...
+            }
+            else
+            {
+                if (shape is IHtmlContent hc)
+                {
+                    return Task.FromResult(hc);
+                }
+                else if (shape is string str)
+                {
+                    return Task.FromResult<IHtmlContent>(new HtmlString(HtmlEncoder.Encode(str)));
```
jtkech, line 98: See my suggestion in the other comment or at least here use new `StringHtmlContent(str)` in place of encoding here

And maybe replace the `else` and `else if` blocks by only `if` blocks as they always returns something ;)

Also not sure about throwing below, regarding what we do elsewhere, returning an empty html string

sebastienros: Ok for the suggestions, but I want to throw here. I/we need to add a comment to mention this should go away eventually, and that's why we throw. We can make it obsolete once we know we have removed the other usages of this method.

jtkech: Okay cool then

sebastienros:
Are the failures in the functional tests bugs in cypress?

Skrypt:
I've had the same kind of issue with Cypress but the issue was totally different than what was logged. So, I think the best thing is to try these functional tests locally to see what is happening exactly.

deanmarcussen:
> @deanmarcussen which would also contain shape view models for its elements

Exactly. At which point it would become usable by consumers. Whereas now you literally have to dig through the `PagerShape` code on github, just to know what the parameters are.

But for laters.

Re: Cypress. 
Nothing else is failing on cypress, so it might be worth looking in the artifacts/logs produced, to see if there's anything. But I can't find them anymore, the artifacts view seems to have disappeared from github.
Possibly a repository setting has changed, or I am blind.

@jptissot any ideas?

jtkech:
In fact i just followed the commit history and reverted the change from which the func tests began to fails

So, i commented out this in `LiquidViewTemplate.cs`

                var localClock = context.Services.GetRequiredService<ILocalClock>();

                var localTimeZone = await localClock.GetLocalTimeZoneAsync();

                var systemTimeZoneId = TZConvert.IanaToWindows(localTimeZone.TimeZoneId);

                var now = await localClock.LocalNowAsync;

                context.TimeZone = TimeZoneInfo.FindSystemTimeZoneById(systemTimeZoneId);
                
                context.Now = () => now;

I will try to understand, i'm reading this https://github.com/mattjohnsonpint/TimeZoneConverter/issues/84#issuecomment-751828277

sebastienros:
It needs to be `TzConvert.TryGetTimeZoneInfo`, which will work everywhere.

jtkech:
Yes, `TZConvert.GetTimeZoneInfo()` works but maybe better to use `TryGetTimeZoneInfo()`, but what we would do if it returns `false` ;) I let you update it as you want

Issue #480, opened 2016-12-28
Html in menu item


agriffard:
Am I allowed to enter text with HTML in a Link menu item?
Ex: In the Name of a link menu item, enter `<i class="fa fa-twitter"></i>`
The text is encoded so it is not well displayed.

Do we have to change the view MenuItemLink-LinkMenuItem.cshtml ?
`<a href="@Url.Content(linkMenuItemPart.Url)">@Html.Raw(linkMenuItemPart.Name)</a>`
Or do we have to create another kind of menu item that allows HTML ?

sebastienros:
Another kind of menu item. Html menu item. The name is only used on the menu editor, and the content of the menu item is rendered in Raw,

agriffard:
Todo: Create an Html Menu item

agriffard:
#480

Issue #7418, opened 2020-10-28
Menu's should use the DisplayText property


jptissot:
The ContentMenuItem and LinkMenuItem should use the DisplayText property instead of both parts having a "Name" property. Or maybe a CommonMenuItemPart to share the name and common properties.

Skrypt:
Maybe just like the TitlePart it should simply fill the DisplayText from the Name.

hishamco:
@Skrypt I think I already did something similar in another PR, I will check ..

jptissot:
@Skrypt Good idea, simple PR :)

hishamco:
@jptissot I already did that if I'm not wrong in another PR, I don't remember the issue, but I will check all my PRs or add a new one if it's needed

hishamco:
I found my PR the issue was on renaming the display name, when I investigate it's obvious that the `DisplayName` not set, I will check if I can do the fix for this first, or do it in my other PR, then check if this issue still exists

I will assign this to me while it's related

hishamco:
May be related to #6129

deanmarcussen:
For infos.

The `DisplayText` property didn't exist when the Menu module was written, hence it wasn't used.

It's not completely trivial.
as @Skrypt says the `DisplayText` property could be migrated to
will also need a migration to migrate existing menus to update their `DisplayText` so the templates continue working. 
Migration should be ok, I think, small number of menu content items per site.

jptissot:
Yes, I was thinking about migrations as well @deanmarcussen I have been playing with the menu shapes and have found some other small issues. For example, I was unable to attach a dynamic part to the LinkMenuItem for common  menu fields. I will document this further when I have time.

hishamco:
I will take care about the other issue and keep this for `1.0.x`

sebastienros:
This has to be a breaking change since templates will need to be updated. But that's an easy thing to document.

Issue #905, opened 2017-08-12
Add _target property to menu item


Hazzamanic:
I think it is a common enough scenario that people want menu items to open in a new tab to be included in the core menu item link.

agriffard:
Would be simple if we had an Html menu item in which you can add any code you want? see #480

PR #8438, opened 2021-01-26
Add html menu type

flew2bits:
These changes add an HTML Menu Item option to the available menu types. I've tested it on both an existing site as well as a new site and the migrations seem to work correctly.

This addresses issue #480.


flew2bits:
This does not, however, address issue #905, since the anchor tag is generated by the template, so there's no way to set the target attribute. That would require an additional field.

deanmarcussen:
This looks good @flew2bits 

I'm not sure if @agriffard had been intending that the link and html were generated together, allowing therefore any kind of html in the Menu Item.

Perhaps he will add a comment.

Two things that would probably be useful here.
We have an issue to just use `DisplayText` instead of a `Name` property see here -> https://github.com/OrchardCMS/OrchardCore/issues/7418

For a new menu item, it might make sense just to do that with `DisplayText` rather than a name, to save any migrations later for it.

The other thing is that as it is html, it should probably by sanitized, by the sanitizer, during save, as is done for the html body part.

More work, as it will then a setting as well, but hopefully you get the idea.

flew2bits:
Thanks for the feedback!

> I'm not sure if agriffard had been intending that the link and html were generated together, allowing therefore any kind of html in the Menu Item.

Based on the discussion in 905, I would guess that this is the case, but I based this PR on the original comment on 480. It's not hard to change, of course.

> For a new menu item, it might make sense just to do that with DisplayText rather than a name, to save any migrations later for it.

I'm not really sure what would be required for this, but I can look into it. Although, it looks to be a complicated issue and could maybe be dealt with later?

> The other thing is that as it is html, it should probably by sanitized, by the sanitizer, during save, as is done for the html body part.

I'll take a look at how to make this work.

flew2bits:
Regarding the sanitizing, do you think that it should also provide settings like the HtmlBodyPart to allow disabling sanitization?

flew2bits:
Based on what I'm reading in #7418, should I keep the Name property in the part and just populate the ContentItem DisplayText value from that? I'm not sure how I would go about updating the DisplayText, otherwise.

src/OrchardCore.Modules/OrchardCore.Menu/Drivers/HtmlMenuItemPartDisplayDriver.cs
```diff
@@ -0,0 +1,48 @@
...
+                model.Url = part.Url;
+                model.Html = part.Html;
+                model.MenuItemPart = part;
+            });
+        }
+
+        public override async Task<IDisplayResult> UpdateAsync(HtmlMenuItemPart part, IUpdateModel updater)
+        {
+            await updater.TryUpdateModelAsync(part, Prefix, x => x.Name, x => x.Url, x => x.Html);
+
```
deanmarcussen, line 44: Here you can bind to the viewmodel, and set the `DisplayText
```
var model = new HtmlMenuItemPartEditViewModel();
if (await update.TryUpdateModelAsync(model, Prefix))
{
    part.ContentItem.DisplayText = model.Name;
    part.Url = model.Url;
    part.Html = model.Html;
}

Does that help make sense?

So because you just have the Name property in the ViewModel, no need to have on the part.

It will save us migrating it later. Sing out if more questions

flew2bits: I've made the updates you've suggested. The HTML Sanitizer is configurable on the part.

flew2bits:
One other thing that probably needs to be addressed is updating the themes to include a view for the new menu item, but that should be trivial.

deanmarcussen:
@agriffard did you have any comments on this regards the url?

I like it as is, I think the `_target="blank"` is more a template thing. Or a different issue.

I might try and get the other menus migrated to use display text before we merge it, so it is all consistent together.

@flew2bits can you merge dev and resolve the merge conflict?

flew2bits:
I've merged and fixed the conflict.

deanmarcussen:
@agriffard looks good to me, I let reviews for my other related pr https://github.com/OrchardCMS/OrchardCore/pull/8664 happen, before merging this as they are related.

There will be a merge conflict on the migrations, so this should go first, and I can fix the other.

PR #8458, opened 2021-01-29
Fix Do not display setup recipes in admin

NogginBox:
Typo in this line means assignment is used rather than equality.

( r.IsSetupRecipe = false)

Always equates to false.

So after this recipe list is always empty,


agriffard:
I don't even know why I didn't use `!r.IsSetupRecipe`

NogginBox:
I thought you were trying to make it clear that you were checking for false. I know some people who think that '!' is too small and sometimes missed.

jtkech (approved):

PR #8459, opened 2021-01-29
compare, not assign

infofromca:


agriffard:
Already done in #8458.

Issue #8466, opened 2021-01-30
Filtering Content Items causes error 500 page intermittently - "pending requests working on this transaction"


mariojsnunes:
Hello,

When my app is deployed on an Azure Linux App Service, with Azure Sql, I've been getting this error intermittently when filtering Content Items on the backoffice.
I've see it happen when I filter by Content Type or Publish Status, or Sort.
Since it's not happening in localhost, my theory atm is that it's because the Azure machine is slower.

> Microsoft.Data.SqlClient.SqlException (0x80131904): The transaction operation cannot be performed because there are pending requests working on this transaction.
>    at Microsoft.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
>    at Microsoft.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
>    at Microsoft.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
>    at Microsoft.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
>    at Microsoft.Data.SqlClient.TdsParser.Run(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj)
>    at Microsoft.Data.SqlClient.TdsParser.TdsExecuteTransactionManagerRequest(Byte[] buffer, TransactionManagerRequestType request, String transactionName, TransactionManagerIsolationLevel isoLevel, Int32 timeout, SqlInternalTransaction transaction, TdsParserStateObject stateObj, Boolean isDelegateControlRequest)
>    at Microsoft.Data.SqlClient.SqlInternalConnectionTds.ExecuteTransactionYukon(TransactionRequest transactionRequest, String transactionName, IsolationLevel iso, SqlInternalTransaction internalTransaction, Boolean isDelegateControlRequest)
>    at Microsoft.Data.SqlClient.SqlInternalTransaction.Commit()
>    at Microsoft.Data.SqlClient.SqlTransaction.Commit()
[...]

PR #4698, opened 2019-11-01
Fixes an SqlQuery triggered by a content handler.

jtkech:
Fixes #4669

Issue #4669, opened 2019-10-29
SQLite Error 6: 'database table is locked: ContentItemIndex


protoncm:
Hello Orchard Core Team

Currently I using  `dev` branch (c7401d3 ) of Orchard Core. i have encounter on this error when i tried to save my content item.

here is the error.
```
Microsoft.Data.Sqlite.SqliteException
  HResult=0x80004005
  Message=SQLite Error 6: 'database table is locked: ContentItemIndex'.
  Source=Microsoft.Data.Sqlite
  StackTrace:
   at Microsoft.Data.Sqlite.SqliteException.ThrowExceptionForRC(Int32 rc, sqlite3 db)
   at Microsoft.Data.Sqlite.SqliteDataReader.NextResult()
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader(CommandBehavior behavior)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReaderAsync(CommandBehavior behavior, CancellationToken cancellationToken)
   at Microsoft.Data.Sqlite.SqliteCommand.<ExecuteDbDataReaderAsync>d__60.MoveNext()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Dapper.SqlMapper.<QueryAsync>d__33`1.MoveNext() in C:\projects\dapper\Dapper\SqlMapper.Async.cs:line 468
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at OrchardCore.Queries.Sql.SqlQuerySource.<ExecuteQueryAsync>d__7.MoveNext() in LOCAL_PATH\OrchardCore.Modules\OrchardCore.Queries\Sql\SqlQuerySource.cs:line 82
```

### What i have done before. 

I create my own content type call _LiquidPage_ with _TitlePart_ and _LiquidPart_. 
[...]

PR #8468, opened 2021-01-30
Added missing transaction commit

mariojsnunes:
Try fix to #8466
Another possible option would be to remove the transaction as it's only a Query. But I'm not sure about it...

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Queries/Sql/SqlQuerySource.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Queries/Sql/SqlQuerySource.cs
@@ -72,6 +72,7 @@ namespace OrchardCore.Queries.Sql
                     using (var transaction = connection.BeginTransaction(_session.Store.Configuration.IsolationLevel))
                     {
                         documentIds = await connection.QueryAsync<int>(rawQuery, parameters, transaction);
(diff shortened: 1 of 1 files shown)


jtkech:
> Another possible option would be to remove the transaction as it's only a Query. But I'm not sure about it...

We added it through #4698 to fix a db locking issue #4669

Not sure we need a transaction commit as we are only reading and the transaction is disposed at the end of the using block, but if we need it we could use `await transaction.CommitAsync()`

One thing i noticed in the trace of your original issue, is that `CommitTransactionAsync()` is called twice.

Need more thinkings, i will let you know

mariojsnunes:
I've updated the PR if you think it will do no harm, I could test if it solves.
Please tell me if I can help in any way.

jtkech:
@mariojsnunes 

> I've updated the PR if you think it will do no harm

Thanks for the async call, and also the `await using` ;)

> I could test if it solves.

Yes please

As commented in your issue, can you check if it was working with an older OrchardCore version, e.g. one month ago.

mariojsnunes:
Since it was found not to be the cause of the issue, should I close this?

jtkech:
@mariojsnunes Okay cool, so i close it

PR #8480, opened 2021-02-02
Cypress 6.4.0

agriffard:
https://www.npmjs.com/package/cypress/v/6.4.0


jptissot:
You should also upgrade the docker image in the all database ci file.

jptissot:
Looks good. All DB tests still failing but for a different reason.

Issue #6524, opened 2020-06-27
ShapesViewModel proposal


deanmarcussen:
With more and more dynamic shapes coming into Orchard Core (all the tab/card grouping shapes, and also shapes like the `ContentCard` it would be potentially useful to have a `ShapesViewModel` that could be used as a hard type abstract class to create specific targetted view models, i.e. a `TabViewModel`.

The problem with the `ShapeViewModel` is it doesn't support a list of positioned `Items`, so right now we are using `Shape` because many of these tab/card/content card shapes require a list of positioned items as well.

So I was not intending that a `ShapesViewModel` was actually an `IShape`, just an abstract that could be used to build shapes with, and the mixin would still turn it into an `IShape`, by mixing in the `ShapeViewModel` which has all the required `IShape` properties.

The idea being you could then use the `IShapeFactory` to create a shape, then add items to it.
```
var shape = ShapeFactory.CreateAsync<TabViewModel>("Tab", m =>
{
    m.TabId = tabId
});

foreach(var item in thetabs)
{
    shape.Add(item);
}
```

for example

sebastienros:
Create a new ShapeViewModel that implements IShapeContainer and has Items. You can inherit from it.
Or just inherit from Shape

PR #8481, opened 2021-02-02
Refactor ContentZone - Tabs, Card, Column containers

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6524

Current usage builds 3 different shapes, even when displaying on the front end, and there are no groups.

Here we do a quick check first, to see if there are any groups, and if there are none, just render the zone directly. No extra shapes.

Also
- removes uses a dynamic, and moves to hard typed models
- uses `ToLookup` instead of `GroupBy` (better usage)
- uses less dynamics for injected properties, i.e. `DisplayAsync` can be resolved directly to its interface instead of being dynamic (this needs evaluation, as we might do it for the other C# Shape Attributes, or not)

Will put some comments in the code changes directly.

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement/Zones/ZoneShapes.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/Zones/ZoneShapes.cs
@@ -33,9 +33,24 @@ namespace OrchardCore.DisplayManagement.Zones
 
             var shapes = ((IEnumerable<dynamic>)Shape);
 
-            // TODO evaluate these for null before perfoming all the groupings.
+            // Evaluate shapes for grouping metadata.
+            var isGrouped = shapes.Any(x => x is IShape s && 
+                (!String.IsNullOrEmpty(s.Metadata.Tab) || 
+                !String.IsNullOrEmpty(s.Metadata.Card) || 
+                !String.IsNullOrEmpty(s.Metadata.Column)));
 
-            string identifier = Shape.Identifier;
+            // When there is no grouping metadata on any shapes just render the Zone.
+            if (!isGrouped)
+            {
+                foreach (var item in shapes)
+                {
+                    htmlContentBuilder.AppendHtml(await DisplayAsync.ShapeExecuteAsync(item));
+                }
+
+                return htmlContentBuilder;
+            }
+
+            string identifier = Shape.Identifier ?? String.Empty;
 
             var groupings = shapes.GroupBy(x =>
             {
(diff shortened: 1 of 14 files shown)


src/OrchardCore/OrchardCore.DisplayManagement/Zones/ZoneShapes.cs
```diff
@@ -13,43 +14,70 @@ namespace OrchardCore.DisplayManagement.Zones
...
+
         [Shape]
-        public async Task<IHtmlContent> Zone(dynamic DisplayAsync, dynamic Shape)
+        public async Task<IHtmlContent> Zone(IDisplayHelper DisplayAsync, IEnumerable<object> Shape)
         {
             var htmlContentBuilder = new HtmlContentBuilder();
             foreach (var item in Shape)
             {
-                htmlContentBuilder.AppendHtml(await DisplayAsync(item));
+                htmlContentBuilder.AppendHtml(await DisplayAsync.ShapeExecuteAsync(item));
```
deanmarcussen, line 26: Here we can inject and use `IDisplayHelper` instead of treating it as dynamic, same with the `Shape`

No a required change, but in my mind, makes it easier to actually see what you're working with, plus less dynamic accessing

src/OrchardCore/OrchardCore.DisplayManagement/Zones/ZoneShapes.cs
```diff
@@ -13,43 +14,70 @@ namespace OrchardCore.DisplayManagement.Zones
...
                 {
-                    key = "Content";
+                    htmlContentBuilder.AppendHtml(await DisplayAsync.ShapeExecuteAsync(item));
                 }
 
-                // Remove any positioning modifier.
-                var modifierIndex = key.IndexOf(';');
-                if (modifierIndex != -1)
+                return htmlContentBuilder;
+            }
```
deanmarcussen, line 55: Noop grouping here and just render the shapes instead of creating all the group shapes.

src/OrchardCore/OrchardCore.DisplayManagement/Zones/ZoneShapes.cs
```diff
@@ -58,189 +86,187 @@ public async Task<IHtmlContent> ContentZone(dynamic DisplayAsync, dynamic Shape,
...
 
                     return new PositionalGrouping(null);
-                }, FlatPositionComparer.Instance).ToList();
+                }, FlatPositionComparer.Instance).ToArray();
+
+                var container = (GroupingsViewModel)await ShapeFactory.CreateAsync<GroupingsViewModel>("TabContainer", m =>
+                {
+                    m.Identifier = identifier;
+                    m.Groupings = orderedGroupings;
+                });
```
deanmarcussen, line 112: Proper model instead of `Shape`, model inherits from `Shape`

deanmarcussen (commented):
a few different things in this (all same area at least), so needs to be reviewed

deanmarcussen:
Agreed in triage, for future pr.

- move all shape attributes away from dynamic
- layers move zone holding away from dynamic.

PR #8486, opened 2021-02-02
#8176 preview deployment plan

### LINKED ISSUE #8176: Preview deployment plan
**Is your feature request related to a problem? Please describe.**
Some times, I only want to preview what will be generated in the deployment plan. For now, it have to download the plan and review it in local. Would it be possible to preview it online?

**Describe the solution you'd like**
Provid a choice in the popup Availible Targets dialog of a deployment plan. Then the new choice will navigate to a new window and review the deployment plan content in it.

HengzheLi:
Fixes #8176.

* Add a Preview deployment target

Below are few screenshots:
image
image
image

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Deployment/Services/InMemoryFileBuilder.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Deployment/Services/InMemoryFileBuilder.cs
@@ -1,0 +1,19 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace OrchardCore.Deployment.Services
+{
+    public class InMemoryFileBuilder : IFileBuilder
+    {
+        public byte[] PlanInMemory { get; private set; }
+        public async Task SetFileAsync(string subpath, Stream stream)
+        {
+            //The passed in stream is a MemoryStream which means it is a seekable stream. So the Length always available
+            PlanInMemory = new byte[stream.Length];
+            await stream.ReadAsync(PlanInMemory, 0, (int)stream.Length);
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


HengzheLi:
I want to preview the deployment plan in a new browser tab. But it seems that the plan export request is post using javascript not href. And I can't find the posting javascript yet.

src/OrchardCore.Modules/OrchardCore.Deployment/Services/InMemoryFileBuilder.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace OrchardCore.Deployment.Services
+{
+    public class InMemoryFileBuilder : IFileBuilder
+    {
+        public byte[] PlanInMemory { get; private set; }
+        public async Task SetFileAsync(string subpath, Stream stream)
+        {
+            //The passed in stream is a MemoryStream which means it is a seekable stream. So the Length always available
+            PlanInMemory = new byte[stream.Length];
+            await stream.ReadAsync(PlanInMemory, 0, (int)stream.Length);
```
deanmarcussen, line 16: I think you could just hold the stream here, rather than reading, and then rereading it when you convert to utf8

Probably just return `File(inMemoryFileBuilder.Steam, "application/json")` is enough, and let the controller take reading of streaming it.

HengzheLi: Tried by saving a reference to the stream into a Stream variable in InMemroyFileBuilder. But a error saying that "Cannot access a closed Stream" when return the result using controller's File().

jptissot:
Please undo all unrelated changes.

HengzheLi:
> Please undo all unrelated changes.

Removed the changes. They were included by incorrectly syncing the master branch.

PR #8487, opened 2021-02-02
Add shortcode name validation

### LINKED ISSUE #8399: Can create shortcode with space or hyphen in name but these are not rendered.
It is possible to create a shortcode with a name like ```my shortcode``` or ```lovely-shortcode``` but when you use them in content they are not rendered and show the shortcode in the final content

### To Reproduce
1. Create short code with name ```my shortcode``` and content ```<span>shortcode content</span>```
2. Creat a content item and use the shortcode in the html body ```[my shortcode]```

### Expected behavior
Either:
- When you view or preview the content item the shortcode is replaced with the shortcode's content.
- When tying to create the shortcode you will be told that the name is not valid.

NogginBox:
Fixes #8399 by validating that a shortcode name is correct.

Also:
- Changed the order and grouping of validation so that name and content are both validated at the same time. Before it would only validate the template content if the name was valid.
- Added inline validation message for template content to create and edit view
- Prevent rename from deleting existing short codes (if you try to rename to an existing shortcode name)
- Always redirect on successful edit

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Shortcodes/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Shortcodes/Controllers/AdminController.cs
@@ -36,8 +36,6 @@ namespace OrchardCore.Shortcodes.Controllers
         private readonly IHtmlLocalizer H;
         private readonly dynamic New;
 
-        private const string NameValidationRegex = "^[a-zA-Z$_][a-zA-Z0-9$_]*$";
-
         public AdminController(
             IAuthorizationService authorizationService,
             ShortcodeTemplatesManager shortcodeTemplatesManager,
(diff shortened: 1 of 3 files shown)


src/OrchardCore.Modules/OrchardCore.Shortcodes/Controllers/AdminController.cs
```diff
@@ -34,6 +35,8 @@ public class AdminController : Controller
         private readonly IHtmlLocalizer H;
         private readonly dynamic New;
 
+        private const string NameValidationRegex = "[a-zA-Z$_][a-zA-Z0-9$_]*";
```
deanmarcussen, line 38: Can we do this without regex? 

Preference is char matching (both performance, and the issue of unconstrained regex)

I would have to look at shortcodes, but probably you can just lean on the Shortcodes/Parlot `ReadIdentifier` and compare the sequence.

NogginBox: I've realised my regex needed start and end anchors, so it should have been "^[a-zA-Z$_][a-zA-Z0-9$_]*$".

I've had a go at using parlot, please see my latest push to see if it's okay?

deanmarcussen: Yeah looks much better to me. can you remove the regex?

NogginBox: Doh! Yes, I should not have left that there. Taken it out.

deanmarcussen (approved):

Issue #8472, opened 2021-01-31
Add Remove ContentLocalizationPart handler / index when removed from type definition


deanmarcussen:
Apply the equivalent of https://github.com/OrchardCMS/OrchardCore/pull/8382

to the `ContentLocalizationPart` to remove its entries from indexes, when the part is removed from a content type definition.

jtkech:
Then yes idem for `ContentLocalizationPart`, maybe update it also for alias where content definition is used in the map delegate, this is also the case for all sql content fields index, but thare are fields

I noticed that for autoroute, if you remove the part a record is added in the index when you re-publish the item, that's okay, but if you publish it again, the record is removed.

So if another running instance (not re-started) is only requested after the 2nd publish, it will not be in sync

So this record would need to be kept in the index, as we do for a removed item

I will fix it in #8442 

For now #8442 is a only a first pass, i will try these use cases tomorrow

deanmarcussen:
> Then yes idem for `ContentLocalizationPart`, maybe update it also for alias where content definition is used in the map delegate, this is also the case for all sql content fields index, but thare are fields

Agreed. harder to do for fields, but possible.

> I noticed that for autoroute, if you remove the part a record is added in the index when you re-publish the item, that's okay, but if you publish it again, the record is removed.

I thought we decided to do that, but set the Path to null, so it is not picked up? But maybe I missed something there. Tomorrow ;)

jtkech:
As a reminder, there is also the `PublishLaterPartIndexProvider` that also removes the related part in the map delegate

jtkech:
> I thought we decided to do that, but set the Path to null, so it is not picked up? But maybe I missed something there. Tomorrow ;)

Yes,  it is read as a null path so the entry is removed, but when the item is published again the related index row is removed, so if another instance has not yet updated its entries (done on demand) it will not be in sync as it will not find this row.

Update: I'm closed to find a solution, needs some last tweaks

PR #8491, opened 2021-02-03
Fixes #8442: Add some index map .When()

### LINKED ISSUE #8442: Taxonomy content list filter issues
### Describe the bug
I was testing the taxonomy content list filter and I found some issues with it's logic that meant to all matching content items were returned. This depended on the status of the content item.

### To Reproduce
1. Using the blog theme activate the taxonomy filter feature and set it to work for "tags"
2. Go to Content Definition and list the blog post types.
2. Create a blog post and tag it with one of the tags - I used the "exploration" tag and publish it (Test Post 1)
3. Create a blog post and tag it with the same tag as above eg. "exploration" and save it as draft (Test Post 2)
4. Create a blog post and tag it with the same tag as above eg. "exploration" and publish it, now edit it and save it as draft (Test Post 3)
5. Now use the tags filter to filter on the tag above eg. "exploration" (Show filter defaults to latest) and you will notice that of the 3 posts you created above only the published one (Test Post 1) is in the results - all 3 test posts should have been int he results.
6. Keep the tag filter the same and change the "show" filter to Published... Test Post 1 and Test Post 3 are in the results which is fine.
7. Keep the tag filter the same and change the "show" filter to Unpublished... there are no results but Test Post 2 should have shown.
8. Keep the tag filter the same and change the "show" filter to All Versions... you will notice that of the 3 posts you created above only the published one (Test Post 1) is in the results - all 3 test posts should have been int he results.

### Expected behavior
Here is the expected behavior for the filter combinations and the 3 test posts above:

1. Tag filter "exploration" and Show filter "Latest" - would expect all 3 test posts should show
2. Tag filter "exploration" and Show filter "Owned by me" - all 3 test posts should show
2. Tag filter "exploration" and Show filter "Published" - Test Post 1 and Test Post 3 should show (this filter works).
3. Tag filter "exploration" and Show filter "Unpublished" - Test Post 2 should be show
4. Tag filter "exploration" and Show filter "All versions" - all 3 test posts should show (not sure how All versions differs from latest - doesn't really make sense)

### Screenshots
taxonomy-filter

jtkech:
Fixes #8442

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Alias/Indexes/AliasPartIndex.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Alias/Indexes/AliasPartIndex.cs
@@ -32,6 +32,7 @@ namespace OrchardCore.Alias.Indexes
         public override void Describe(DescribeContext<ContentItem> context)
         {
             context.For<AliasPartIndex>()
+                .When(contentItem => (contentItem.IsPublished() || contentItem.Latest) && contentItem.Has<AliasPart>())
                 .Map(contentItem =>
                 {
                     var aliasPart = contentItem.As<AliasPart>();
--- a/src/OrchardCore.Modules/OrchardCore.Alias/Indexes/AliasPartIndex.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Alias/Indexes/AliasPartIndex.cs
@@ -36,8 +36,7 @@ namespace OrchardCore.Alias.Indexes
                 .Map(contentItem =>
                 {
                     var aliasPart = contentItem.As<AliasPart>();
-
-                    if (aliasPart == null)
+                    if (aliasPart == null || String.IsNullOrEmpty(aliasPart.Alias))
                     {
                         return null;
                     }
@@ -66,20 +65,13 @@ namespace OrchardCore.Alias.Indexes
                         return null;
                     }
 
-                    var alias = aliasPart.Alias;
-
-                    if (!String.IsNullOrEmpty(alias) && (contentItem.Published || contentItem.Latest))
+                    return new AliasPartIndex
                     {
-                        return new AliasPartIndex
-                        {
-                            Alias = alias.ToLowerInvariant(),
-                            ContentItemId = contentItem.ContentItemId,
-                            Latest = contentItem.Latest,
-                            Published = contentItem.Published
-                        };
-                    }
-
-                    return null;
+                        Alias = aliasPart.Alias.ToLowerInvariant(),
+                        ContentItemId = contentItem.ContentItemId,
+                        Latest = contentItem.Latest,
+                        Published = contentItem.Published
+                    };
                 });
         }
     }
--- a/src/OrchardCore.Modules/OrchardCore.Autoroute/Indexes/AutoroutePartIndex.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Autoroute/Indexes/AutoroutePartIndex.cs
@@ -107,13 +107,16 @@ namespace OrchardCore.ContentManagement.Records
         public void Describe(DescribeContext<ContentItem> context)
         {
             context.For<AutoroutePartIndex>()
+                .When(contentItem => contentItem.Has<AutoroutePart>())
                 .Map(async contentItem =>
                 {
                     var part = contentItem.As<AutoroutePart>();
+                    if (part == null)
+                    {
+                        return null;
+                    }
 
-                    // When the part has been removed from the type definition it will return null here but we still need to process the index.
-                    var partRemoved = _partRemoved.Contains(contentItem);
-                    if (!partRemoved && part == null)
+                    if (String.IsNullOrEmpty(part.Path))
                     {
                         return null;
                     }
@@ -125,11 +128,7 @@ namespace OrchardCore.ContentManagement.Records
                         return null;
                     }
 
-                    if (!partRemoved && String.IsNullOrEmpty(part.Path))
-                    {
-                        return null;
-                    }
-
+                    var partRemoved = _partRemoved.Contains(contentItem);
                     var results = new List<AutoroutePartIndex>
                     {
                         // If the part is disabled, or removed, a record is still added but with a null path.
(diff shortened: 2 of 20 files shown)


src/OrchardCore.Modules/OrchardCore.Alias/Indexes/AliasPartIndex.cs
```diff
@@ -32,11 +32,11 @@ public AliasPartIndexProvider(IServiceProvider serviceProvider)
         public override void Describe(DescribeContext<ContentItem> context)
         {
             context.For<AliasPartIndex>()
+                .When(contentItem => (contentItem.IsPublished() || contentItem.Latest) && contentItem.Has<AliasPart>())
```
deanmarcussen, line 35: So the problem with using `.When` here is twofold.
`.When` stops the deletes happening from the Index (that's its purpose).

so if you use it to do `contentItem.IsPublished()` then when a content item is not published (but was previously), it's previous record will not be deleted from the Index, so will remain, and will still have the `Published` value set to true.

The Alias should also be covered by the `Remove()` concept we introduced (seperate issue. didn't think about that one as well)

src/OrchardCore.Modules/OrchardCore.Autoroute/Indexes/AutoroutePartIndex.cs
```diff
@@ -107,13 +107,11 @@ public override async Task UpdatedAsync(UpdateContentContext context)
         public void Describe(DescribeContext<ContentItem> context)
         {
             context.For<AutoroutePartIndex>()
+                .When(contentItem => contentItem.Has<AutoroutePart>())
```
deanmarcussen, line 110: The second problem with using `When` here is it will stop the `.Map` being called, because the part no longer has the `AutoroutePart` because it has been removed previously by the handler, so will stop the delete of the item.

But in this case I think we could check in `When` to see if `.Has<AutoroutePart>() || _part.Removed.Contains()`, and return true, which would allow the `.Map` to continue, and do what it needs to, i.e issue the necessary deletes.

deanmarcussen (commented):
Agree with moving the field indexes to just published, or latest - I saw this previously, but did not want to change it as was working on something else.

src/OrchardCore.Modules/OrchardCore.Alias/Indexes/AliasPartIndex.cs
jtkech: Yes, it's only a first pass, i will try thes use cases tomorrow

> The Alias should also be covered by the Remove() concept we introduced (seperate issue. didn't think about that one as well)

Yes i saw it, see my comment in #8472

src/OrchardCore.Modules/OrchardCore.Autoroute/Indexes/AutoroutePartIndex.cs
jtkech: But it is in the item you enlisted in the `_partRemoved`

It's a first pass, and i saw another problem, i will see tomorrow

I thought about using `_part.Removed` but i was thinking of not adding a bad closure, but that's the same in the map delegate, so maybe

I also need to check more in depth what return `Has<>` and `As<>`

src/OrchardCore.Modules/OrchardCore.Taxonomies/Indexing/TaxonomyIndex.cs
```diff
@@ -35,21 +35,17 @@ public TaxonomyIndexProvider(IServiceProvider serviceProvider)
         public override void Describe(DescribeContext<ContentItem> context)
         {
             context.For<TaxonomyIndex>()
+                .When(contentItem => contentItem.IsPublished() || contentItem.Latest)
                 .Map(contentItem =>
                 {
-                    if (!contentItem.IsPublished())
```
deanmarcussen, line 40: Cool for first pass.

I think on taxonomies we should add `Published` and `Latest` to the index.

because right now, I have a lot of SQL Queries, that do not `JOIN` on the `ContentItemIndex` because they did not need to. And I figure if I have them, other people do too, especially from the various issues where people have asked about how to query taxonomies with SQL Queries. 
So good to be able to keep to one query, no join. and easy to add `Published` and `Latest` with defaults for both (true, false) - I think I added this to `Alias` some time ago.

src/OrchardCore.Modules/OrchardCore.Autoroute/Indexes/AutoroutePartIndex.cs
jtkech: Okay, just saw that you remove the part in the handler section of the index provider ;)

And i was planning to do it in the map delegate, as alias does for now

Also, the record added when removing the part needs to be kept in the index even if we publish again the item, otherwise another running instance may be not in sync, will see this night / tomorrow

src/OrchardCore.Modules/OrchardCore.Taxonomies/Indexing/TaxonomyIndex.cs
jtkech: > I think on taxonomies we should add Published and Latest to the index.

This is what it now does, with the above `.When` that i added

deanmarcussen: Haha, go to bed. 
What I mean is we should track the `Published` and `Latest` values, in the Index itself

so the index is now
```
    public class TaxonomyIndex : MapIndex
    {
        public string TaxonomyContentItemId { get; set; }
        public string ContentItemId { get; set; }
        public string ContentType { get; set; }
        public string ContentPart { get; set; }
        public string ContentField { get; set; }
        public string TermContentItemId { get; set; }
    }
```

and we should make it

```
    public class TaxonomyIndex : MapIndex
    {
        public string TaxonomyContentItemId { get; set; }
        public string ContentItemId { get; set; }
        public string ContentType { get; set; }
        public string ContentPart { get; set; }
        public string ContentField { get; set; }
        public string TermContentItemId { get; set; }
        public bool Published { get; set; }
        public bool Latest { get; set; }
    }
```

jtkech:
I will restart it from scratch based on the last dev where #8245 and #8501 have been merged

PR #8504, opened 2021-02-04
Bring back database integration testing

deanmarcussen:
An experiment to see if we can't have a bit of database testing that runs all the time, instead of manually - when nuget containers aren't broken ;) 

The experiment is to see how much time it will actually take on the CI hours and if it works!


deanmarcussen:
Non trivial. Another day

PR #8506, opened 2021-02-04
Content list search does not submit on Safari, iOS or PadOS browsers

### LINKED ISSUE #8505: Content search does not submit on Safari MacOS or iOS
### Describe the bug
On Safari for MacOS, iOS (any browser) or PadOS the search on the content list does not submit and so the user can not search. Because there is no search button, users can click it with mouse either.

### To Reproduce
Steps to reproduce the behavior in Safari browser:
1. Go to content list
2. Enter text in the search field
3. Press enter
4. Nothing happens

scleaver:
Fixes #8505 

Safari, iOS or PadOS browsers requires a submit input that does not have `display: none` applied. There are two submit inputs in the forms already with `display: none` applied via a `d-none` class. Solution is to use the `sr-only` which hides the submit buttons without using `display: none`. This might be better for accessibility too but not sure since there is two submit buttons. I think submit.Filter input is used by the text field technically so could probably get away with just applying the `sr-only` class to it - I will await feedback.

As part of this also made two extra enhancements:
 
1. added a label with `sr-only` to the search field for accessibility.
2. added the search input mode for better keyboard experience on devices.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsAdminList.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsAdminList.cshtml
@@ -3,8 +3,8 @@
 <script asp-name="bootstrap-select" depends-on="admin" at="Foot"></script>
 
 <!-- Hidden submit button do not remove -->
-<input type="submit" name="submit.Filter" id="submitFilter" class="d-none" />
-<input type="submit" name="submit.BulkAction" class="d-none" />
+<input type="submit" name="submit.Filter" id="submitFilter" class="sr-only" />
+<input type="submit" name="submit.BulkAction" class="sr-only" />
 @Html.HiddenFor(o => o.Options.BulkAction)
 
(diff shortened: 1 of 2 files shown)


Skrypt (commented):
I approve the changes. Apply it everywhere we have a search bar.
Also consider adding a "hidden" attribute on these inputs which should also work now.

```html
<input type="submit" name="submit.Filter" id="submitFilter" hidden class="sr-only" />
<input type="submit" name="submit.BulkAction" hidden class="sr-only" />
```

or

```html
<input type="submit" name="submit.Filter" id="submitFilter" hidden />
<input type="submit" name="submit.BulkAction" hidden />
```

scleaver:
> I approve the changes. Apply it everywhere we have a search bar.
> Also consider adding a "hidden" attribute on these inputs which should also work now.

@Skrypt Unfortunately a hidden does not work. I tested it hoping it would :(

https://stackoverflow.com/questions/52236168/cant-submit-form-using-enter-in-safari-browser

PR #8512, opened 2021-02-04
Add redirect to returnUrl after user edit

arnaudroystatcan:
Add code to redirect to the returnUrl after a successful edit on an user profile.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Users/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Controllers/AdminController.cs
@@ -379,6 +379,11 @@ namespace OrchardCore.Users.Controllers
             }
             else
             {
+                if (!String.IsNullOrEmpty(returnUrl))
+                {
+                    return LocalRedirect(returnUrl);
+                }
+
                 return RedirectToAction(nameof(Index));
             }
         }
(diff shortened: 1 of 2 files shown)


deanmarcussen:
I'm not sure i get this. From the discussion you opened a while ago, it was about the cancel button.

There is already code when editing the own user to redirect to a return url? Otherwise they have come from the userslist screen and should always go back there.

I think you previous commit, with the changes to the user save buttons looks more like it fixes what you previously described, it would just need to check if there is a returnurl value in the view data, and use `Index` if there is not.

Unless you are trying to fix a different issue than you previously asked about?

arnaudroystatcan:
In our case, we are using the platform to run hackathons so we have a contentType Team. Users can be assigned to those teams. We decided to add a dropdown of the team members in the ContentItems screen (refer to picture) that would redirect to that user's profile. Once any edit is saved the admin would be redirected back to the contentItems view.

My previous commit with the Cancel button was just me trying to see if we could also use the returnUrl when the edit was canceled but returnUrl being optionnal I thought it was a bad idea.

image

deanmarcussen:
ok, so different use case, than what we had already discussed? Was it you that I had the discussion with, sorry it was a week ago?

Seems good then.

I'm also ok with the `Cancel` `returnUrl` if it was checked that it is not `null would work fine. 
Often our `returnUrl` are optional, and checked in this way. Something like this is fine `var returnUrl = ViewData["returnUrl"]?.ToString();`

arnaudroystatcan:
The change I made is handled in the controller (Edit post) after a user clicks save, but the cancel is handled in the view. 
`<div class="form-group">
    <button class="btn btn-primary save" type="submit">@T["Save"]</button>
    <a class="btn btn-secondary cancel" role="button" asp-route-action="Index">@T["Cancel"]</a>
</div>
`
I'm not too sure how to go about this, i'm fairly new to razor. I think we would need to check inline if Viewdata["returnUrl"] is null then go for href or asp-route-action="Index".

sebastienros (approved):

PR #8515, opened 2021-02-05
Index mapping

### LINKED ISSUE #8472: Add Remove ContentLocalizationPart handler / index when removed from type definition

jtkech:
Fixes #8472 

@deanmarcussen about the little "problem" for Autoroutes, here the `.When()` was useful, we use it so that the mapping still run if the part was just removed, otherwise only if the item has an autoroute part. This still allows to add a record when the part was removed, but prevent to remove this record if the item is published again. Otherwise another instance may be not in sync if it updates its autoroute entries (from the index table) after the item was published again.

I used the same for the localization index as it is also used for syncing localization entries. But not e.g. for the alias index where i used the same `.When()` but here when the part was removed the index row is just removed (as the part = null).

Note: For the removed part i had to enlist the contentItemId, this to add / update / keep / remove records as needed.

Still in progress, will see for remaining indexes, and what we can do for content field indexes.


deanmarcussen:
Looks good @jtkech, makes sense, and I like the use of `When`

jtkech:
@deanmarcussen 

> Maybe to consider the fields seperately (for removing), up to you if you have a plan ;)

Yes, a beginning of a plan, but anyway in a separate PR ;)

Maybe good to have `HasField<>`, `AsField<>` and `RemoveField<>` extensions

I marked this one as ready but i didn't merge it yet in case you want to look at it again

src/OrchardCore.Modules/OrchardCore.Lists/Indexes/ContainedPartIndex.cs
```diff
@@ -18,17 +18,23 @@ public override void Describe(DescribeContext<ContentItem> context)
                 .When(contentItem => contentItem.Has<ContainedPart>())
                 .Map(contentItem =>
                 {
+                    // Remove index records of soft deleted items.
+                    if (!contentItem.Published && !contentItem.Latest)
```
deanmarcussen, line 22: I'm not sure about this one

This means when restoring the content item it is lost from the list, and often lost from the admin controller view, because by default they are not listable.

It should never be lost from the list it is contained in.

Currently I do this restoration as required manually with sql.
This would mean having to then create an index record for the restored item as well.

Now the audit trail module will probably do this automatically (i.e. save it), as the item would still have the `ContainedPart` 

But if there is ever any audit trail, restore content item, from an action in a list, we'd probably need a query on the index (which would now be empty).

So I think no advantage to removing it, but many disadvantages.

src/OrchardCore.Modules/OrchardCore.Layers/Indexes/LayerMetadataIndex.cs
```diff
@@ -17,16 +17,22 @@ public override void Describe(DescribeContext<ContentItem> context)
                 .When(contentItem => contentItem.Has<LayerMetadata>())
                 .Map(contentItem =>
                 {
+                    // Remove index records of soft deleted items.
+                    if (!contentItem.Published && !contentItem.Latest)
+                    {
+                        return null;
+                    }
```
deanmarcussen, line 24: as below, we rely on this to have a record. crucial. or items disappear.

src/OrchardCore/OrchardCore.ContentLocalization.Abstractions/Records/LocalizedContentItemIndex.cs
```diff
@@ -48,29 +79,28 @@ public override Task RemovedAsync(RemoveContentContext context)
...
         {
             context.For<LocalizedContentItemIndex>()
+                .When(contentItem => contentItem.Has<LocalizationPart>() || _partRemoved.Contains(contentItem.ContentItemId))
                 .Map(contentItem =>
                 {
-                    var part = contentItem.As<LocalizationPart>();
-
-                    if (part == null)
+                    // If the content item was removed, a record is still added.
+                    if (!contentItem.Published && !contentItem.Latest && !_itemRemoved.Contains(contentItem))
```
deanmarcussen, line 86: seems ok to do this here, as when returning to published, or republishing, the `LocalizationPart` would still contain the value, so index recreated on save.

but also maybe not, if ever a query was wanted for deleted items as part of a localization set?

deanmarcussen (commented):
thanks @jtkech 

fine with the fields, concerned about the list / contained index.

src/OrchardCore/OrchardCore.ContentLocalization.Abstractions/Records/LocalizedContentItemIndex.cs
jtkech: Here we do the same as for `AutoroutePartIndex` as we refresh the entries from the index table and by using a cursor, so as for autoroutes we need to add a record when deleting an item (or part).

src/OrchardCore.Modules/OrchardCore.Lists/Indexes/ContainedPartIndex.cs
jtkech: Yes, i was not sure about contained items, and just because thinking about audit trail

But if audit trail restore a container to re-publish / re-draft it, i thought it would retrieve the ids of the contained item and then restore them knowing their ids, this to re-publish / re-save them

src/OrchardCore.Modules/OrchardCore.Layers/Indexes/LayerMetadataIndex.cs
jtkech: I think you mean as above as they are contained items, widgets in a layer, right?

If so same question

src/OrchardCore.Modules/OrchardCore.Lists/Indexes/ContainedPartIndex.cs
jtkech: Okay, just checked, this is the contained item that knows its parent ;)

So i will revert this for the `containedPartIndex`

src/OrchardCore.Modules/OrchardCore.Layers/Indexes/LayerMetadataIndex.cs
jtkech: Okay, idem i will revert this one

src/OrchardCore/OrchardCore.ContentLocalization.Abstractions/Records/LocalizedContentItemIndex.cs
jtkech: For the part remove, here we set Culture to null, as we set the Path to null for autoroutes

> but also maybe not, if ever a query was wanted for deleted items as part of a localization set?

That would be ok because here (as for autoroutes) we keep the record when it was deleted, and i take care that it is kept in the table, so that other running instances don't miss it

Need to leave, will see this night / tomorrow

jtkech:
@deanmarcussen 

I reverted the check of soft deleted items in `LayerMetadataIndex` and `ContainedPartIndex` as they are contained items, but i added this check in the `.When()`, so removed items are still indexed, but at least the mapping is not executed.

For `LocalizedContentItemIndex` it was the same from a previous PR, originally the `Latest` column was not there and all non published were removed from the index, now at least we keep one version in the index when removed (no active version left).

So, if you are okay with this, we could merge this PR

One idea on the fly, why not having an `AliasEntries` service, and maybe manage the index as we do for autoroutes

src/OrchardCore.Modules/OrchardCore.Layers/Indexes/LayerMetadataIndex.cs
```diff
@@ -14,19 +14,20 @@ public class LayerMetadataIndexProvider : IndexProvider<ContentItem>
         public override void Describe(DescribeContext<ContentItem> context)
         {
             context.For<LayerMetadataIndex>()
-                .When(contentItem => contentItem.Has<LayerMetadata>())
+                // Keep index records of soft deleted items as they are contained items.
+                .When(contentItem => contentItem.Has<LayerMetadata>() && (contentItem.Published || contentItem.Latest))
```
deanmarcussen, line 18: Hmm, ok, so I like the `.When(contentItem => contentItem.Has<LayerMetadata>()` but I don't think we're gaining anything with `&& (contentItem.Published || contentItem.Latest)` - it's a micro optimization I think, which makes the api more complicated (and potentially prone to strange behaviour)

Say for example you have a handler that catches and changes something about this, when the document is unpublished, then it doesn't get saved to the index.

I'm not dead against it, but it seems like a micro optimization, that just makes the api more confusing?

jtkech: @deanmarcussen 

Yes i agree about optimization but i only thought about what i should do, keep the index when removing an item (as it is a contained item) that was already indexed as an active(s) version(s), but still not indexing all versions, e.g. when we publish / draft / publish / draft, and so on.

Just re-tried as it was before, indeed we enlist all versions, there is not so many items in a layer, but more blog posts. Is it what we want? If so i'm fine with it.

Then i re-tried as it is in this PR, we index only active versions, but when we delete an item, and i was surprised, the index is not kept in the table, so yes, at least moving the check from the mapping to the `When()` is useless.

But this is because **when an item is filtered by the `When()`, it is still removed from the index**, this is what i thought initially (i may have tested it) but, as i remember, for another index table you said to me the opposite ;) So we removed a filter in the `.When()` as we thought that it would not remove the index row as it is suposed to do.

But not sure at 100%, i will redo tests tonight, i may be wrong. Then i will check if we can still keep the last actives versions when removing an item version (e.g. unpublish a not latest one), but without indexing all the history of all versions, or just let it as it was before for the 2 related index tables ;)

deanmarcussen: > But this is because when an item is filtered by the When(), it is still removed from the index,

This shouldn't happen. `When` should only be an indicator to YesSql that it should call `Map`. When it returns true, it should call `Map` which will force the delete and when it returns null, will not add the creates. But remember it's only being called on the content item being saved, not reindexing the whole index table.

So I would expect that say, for example, publishing an item, results in one call to reset the indexes on the old item (i.e. delete and recreate), and one call for the new item. Because only those two items are being saved.

jtkech:
@deanmarcussen 

> But remember it's only being called on the content item being saved, not reindexing the whole index table.

Yes, i know

> When it returns true, it should call Map

Yes, as checked in `MapNew()`

> which will force the delete

No, not at this point in `MapNew()`, if the map return null here it does nothing, as when the filter return false. This is before in `MapDeleted()` that the index may be removed if the filter still returns true (see below).

Okay i understand, when updating an entity, before `MapNew()` on the new entity, we do `MapDeleted()` on the old object, it also checks the filter, but in our case the old content doesn't have `Published = Latest = false`, that's why here the filter `contentItem.Published || contentItem.Latest` still returns true, and the index row is still removed.

            await MapDeleted(oldDoc, oldObj, collection);

            await MapNew(oldDoc, entity, collection);

Based on the above, i could get it working as i wanted and described by using here also an `_itemRemoved`, then yes, when an item is removed, you can only restore manually the last active version kept in the index. I saw that to restore an item, to be showed in the admin you also need to update the `ContentItemIndex` table by setting Latest = true, right?

But yes, makes sense for a contained item to keep all versions in the index, at least until we have the audit trail feature, so finally i reverted the changes to the contained part and the layer metadata indexes, so that it is as before ;)

deanmarcussen:
> I saw that to restore an item, to be showed in the admin you also need to update the ContentItemIndex table by setting Latest = true, right?

Yes, it's not uncomplicated to actually restore an item. Have to reset the content item index table, and have to reset the `Published` `Latest` values in the `Document.Content` itself. Or the index finds it, but it still doesn't show because it reads the values from the `Document.Content`...

But one does, what customers need, when they click the wrong button ;)

Cool, me I'm happy with these, looks really good.

> One idea on the fly, why not having an AliasEntries service, and maybe manage the index as we do for autoroutes

Yeah I've thought the same thing a couple of times. Much better for performance, than the two queries we do now, one to get the alias, and another for the content item. Would not hurt anything I think! Other than making it a bit more complicated... One day I need to work on contained aliases as well, but that's another issue.

The other one that has occured to me is a taxonomy cache, much like we do with the Layers cache. As you pointed out, the blog recipe has grown since last year when it was performance tested. And now has two taxonomies, categories, and tags. It's super normal to have a few taxonomies that help build side navs, or tag categories, so the idea of caching them doesn't seem like a bad one. We could potentially have a way to exclude particular ones (through an options, just as a backup, in case people are using 1000's taxonomies). But so much more usual to have 2-10. Equally I think I included options on the `TaxonomyShapes` / `TermShapes` to cache, like the menus do. But maybe not turned on for the blog recipe. I should check that to see. Also maybe not useful, if it is retrieved in the driver. Will have a look, but just for more ideas ;)

deanmarcussen (approved):

PR #8521, opened 2021-02-05
Fix modals z-index with Trumbowyg.

### LINKED ISSUE #5540: Media: Delete does not work from Insert popup
When you want to insert a Media in a Wysiwyg editor, a popup opens. 
Deleting a media does not work in this popup.

Skrypt:
Fixes #5540


hishamco:
BOOOM!! 🚀 🚀 🚀

Skrypt:
Yeah, general rule to keep in mind. Using always higher values on z-index will only get us to always add higher numbers.
If some component uses 99999999 as a z-index generally it's bad. I try to keep inside the 1000 to 1100 range.
Here the issue is the z-index of the Trumbowyg which was way too high for nothing. I tested it in different places and it seems to work. Let's see if it breaks somewhere else we never know. 😄

hishamco:
I think we should need to use mixins or docs them somewhere to avoid such conflict issues

Skrypt:
I remember it mostly because it comes from Bootstrap which uses this range of z-index.

PR #8272, opened 2021-01-10
Add shortcode resolver to HTMLField

MikeKry:
Added shortcode resolver to htmlfield, based on htmbody part + inspiration from markdown field.

But @deanmarcussen  I have noticed that htmlbodypart has no branch to process "SanitizeHtml" prop, but markdownpart/field does. Should htmlbodypart/htmlfield also include processing of sanitizehtml? 

Related to #8241

PR #8536, opened 2021-02-08
Correct GraphQL name for HtmlField

MikeKry:
Fix name of htmlfield graphql type


MikeKry:
@deanmarcussen 

just noticed that I have put wrong name into HtmlField graphql field (does not affect functionality, but it shows errors and fragment names with incorrect name) 

originally in https://github.com/OrchardCMS/OrchardCore/pull/8272

deanmarcussen (approved):

PR #8548, opened 2021-02-08
#7518 background tasks, nextStartTime is assumed as local time

### LINKED ISSUE #7518: Background tasks, time zones should be converted
image
image

China UTC Plus 8, 8 a.m ： 0 8 * * *
now ： 0 0 * * *

lampersky:
fixes #7518

### DIFF
--- a/src/OrchardCore/OrchardCore/BackgroundTasks/BackgroundTaskScheduler.cs
+++ b/src/OrchardCore/OrchardCore/BackgroundTasks/BackgroundTaskScheduler.cs
@@ -1,17 +1,20 @@
 using System;
 using NCrontab;
+using OrchardCore.Modules;
 
 namespace OrchardCore.BackgroundTasks
 {
     public class BackgroundTaskScheduler
     {
-        public BackgroundTaskScheduler(string tenant, string name, DateTime referenceTime)
+        public BackgroundTaskScheduler(string tenant, string name, DateTime referenceTime, IClock clock, string timeZoneId)
         {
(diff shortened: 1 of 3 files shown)


Skrypt (approved):

sebastienros:
I don't think it's correct. `ILocalClock` does the job for you. In this PR the TimeZoneId setting is shortcutting the ITimeZoneSelector. I think you should just resolve `ILocalClock` instead of `IClock` and then access the `LocalNowAsync` value. It will resolve the correct timezone and convert the value for you, and based on any custom logic a module might have set.

lampersky:
Thank you, will fix it.

I thought ILocalClock is the environment local time (where the app is
hosted and run), not site time (based on what user choose in site setting).

Skrypt:
ILocalClock does the job for you. Forgot about this one.

sebastienros:
@Skrypt I didn't remember either, but the code looked too complicated so I checked what the DateTimeShapes were doing.

jtkech:
@Skrypt @sebastienros 

So, we would need to also update the `TimerEvent` that does

        private bool IsExpired()
        {
            if (StartedUtc == null)
            {
                StartedUtc = StartAtUtc ?? _clock.UtcNow;
            }

            var schedule = CrontabSchedule.Parse(CronExpression);
            var whenUtc = schedule.GetNextOccurrence(StartedUtc.Value);

            return _clock.UtcNow >= whenUtc;
        }

If so, at least open a new issue.

PR #8549, opened 2021-02-08
Fix WorkflowBlockingActivitiesIndex table indices name length for PostgreSQL

### LINKED ISSUE #8545: PostgreSQL database issue with latest
### Describe the bug
Error occurs on Postgresql when enabling our Hackathon feature. Works fine with SQLite but with postgresql, the site completely crashes. 

### To Reproduce
Create a tenant with a long enough name to go over the 64 character limit for table names on postgresql

### Expected behavior
Migrations to execute without crashing the site.

### Screenshots
```bash
fail: OrchardCore.Data.Migration.DataMigrationManager[0]
      => SpanId:332116e92b5d7a4f, TraceId:d4cacbf54ed6af418a87fa58e897948c, ParentId:0000000000000000 => ConnectionId:0HM6C8V30N54S => RequestPath:/tj5ogvh/Admin/Features RequestId:0HM6C8V30N54S:00000016
      Error while running migration version 0 for 'OrchardCore.Workflows'.
      System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
       ---> Npgsql.PostgresException (0x80004005): 42P07: relation "tj5ogvh_IDX_WorkflowBlockingActivitiesIndex_DocumentId_Activity" already exists
         at Npgsql.NpgsqlConnector.<ReadMessage>g__ReadMessageLong|194_0(NpgsqlConnector connector, Boolean async, DataRowLoadingMode dataRowLoadingMode, Boolean readingNotifications, Boolean isReadingPrependedMessage)
         at Npgsql.NpgsqlDataReader.NextResult(Boolean async, Boolean isConsuming, CancellationToken cancellationToken)
         at Npgsql.NpgsqlDataReader.NextResult()
         at Npgsql.NpgsqlCommand.ExecuteReader(CommandBehavior behavior, Boolean async, CancellationToken cancellationToken)
         at Npgsql.NpgsqlCommand.ExecuteReader(CommandBehavior behavior, Boolean async, CancellationToken cancellationToken)
         at Npgsql.NpgsqlCommand.ExecuteNonQuery(Boolean async, CancellationToken cancellationToken)
         at Npgsql.NpgsqlCommand.ExecuteNonQuery()
         at Dapper.SqlMapper.ExecuteCommand(IDbConnection cnn, CommandDefinition& command, Action`2 paramReader) in /_/Dapper/SqlMapper.cs:line 2813
         at Dapper.SqlMapper.ExecuteImpl(IDbConnection cnn, CommandDefinition& command) in /_/Dapper/SqlMapper.cs:line 572
         at YesSql.Sql.SchemaBuilder.Execute(IEnumerable`1 statements)
         at YesSql.Sql.SchemaBuilder.AlterTable(String name, Action`1 table)
         at YesSql.Sql.SchemaBuilder.AlterIndexTable(Type indexType, Action`1 table, String collection)
         at YesSql.Sql.SchemaBuilderExtensions.AlterIndexTable[T](ISchemaBuilder builder, Action`1 table, String collection)
         at OrchardCore.Workflows.Migrations.Create()
        Exception data:
          Severity: ERROR
          SqlState: 42P07
          MessageText: relation "tj5ogvh_IDX_WorkflowBlockingActivitiesIndex_DocumentId_Activity" already exists
          File: index.c
          Line: 884
          Routine: index_create
         --- End of inner exception stack trace ---
         at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
         at OrchardCore.Data.Migration.DataMigrationManager.UpdateAsync(String featureId)
fail: OrchardCore.Data.Migration.DataMigrationManager[0]
      => SpanId:332116e92b5d7a4f, TraceId:d4cacbf54ed6af418a87fa58e897948c, ParentId:0000000000000000 => ConnectionId:0HM6C8V30N54S => RequestPath:/tj5ogvh/Admin/Features RequestId:0HM6C8V30N54S:00000016
      Error while running migration version 0 for 'StatCan.OrchardCore.Hackathon'.
      System.ObjectDisposedException: The CancellationTokenSource has been disposed.
         at System.Threading.CancellationTokenSource.ThrowObjectDisposedException()
         at System.Threading.CancellationTokenSource.CancelAfter(Int32 millisecondsDelay)
         at System.Threading.CancellationTokenSource.CancelAfter(TimeSpan delay)
         at Npgsql.Util.ResettableCancellationTokenSource.Stop()
         at Npgsql.NpgsqlReadBuffer.<Ensure>g__EnsureLong|40_0(NpgsqlReadBuffer buffer, Int32 count, Boolean async, Boolean readingNotifications)
         at Npgsql.NpgsqlConnector.<ReadMessage>g__ReadMessageLong|194_0(NpgsqlConnector connector, Boolean async, DataRowLoadingMode dataRowLoadingMode, Boolean readingNotifications, Boolean isReadingPrependedMessage)
         at Npgsql.NpgsqlDataReader.NextResult(Boolean async, Boolean isConsuming, CancellationToken cancellationToken)
         at Npgsql.NpgsqlDataReader.Consume(Boolean async)
         at Npgsql.NpgsqlDataReader.Close(Boolean connectionClosing, Boolean async, Boolean isDisposing)
         at Npgsql.NpgsqlConnector.CloseOngoingOperations(Boolean async, CancellationToken cancellationToken)
         at Npgsql.NpgsqlTransaction.Dispose(Boolean disposing)
         at YesSql.Session.ReleaseTransaction()
         at YesSql.Session.Cancel()
         at YesSql.Services.DefaultQuery.Query`1.FirstOrDefaultImpl()
         at OrchardCore.Data.Documents.DocumentStore.GetOrCreateMutableAsync[T](Func`1 factoryAsync)
         at OrchardCore.Documents.DocumentManager`1.GetOrCreateMutableAsync(Func`1 factoryAsync)
         at OrchardCore.Queries.Services.QueryManager.SaveQueryAsync(String name, Query query)
         at OrchardCore.Queries.Recipes.QueryStep.ExecuteAsync(RecipeExecutionContext context)
         at OrchardCore.Recipes.Services.RecipeExecutor.<>c__DisplayClass7_0.<<ExecuteStepAsync>b__0>d.MoveNext()
      --- End of stack trace from previous location ---
         at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell)
         at OrchardCore.Recipes.Services.RecipeExecutor.ExecuteStepAsync(RecipeExecutionContext recipeStep)
         at OrchardCore.Recipes.Services.RecipeExecutor.ExecuteAsync(String executionId, RecipeDescriptor recipeDescriptor, Object environment, CancellationToken cancellationToken)
         at OrchardCore.Recipes.Services.RecipeExecutor.ExecuteAsync(String executionId, RecipeDescriptor recipeDescriptor, Object environment, CancellationToken cancellationToken)
         at OrchardCore.Recipes.Services.RecipeExecutor.ExecuteAsync(String executionId, RecipeDescriptor recipeDescriptor, Object environment, CancellationToken cancellationToken)
         at OrchardCore.Recipes.Services.RecipeMigrator.ExecuteAsync(String recipeFileName, IDataMigration migration)
         at StatCan.OrchardCore.Hackathon.HackathonMigrations.CreateAsync() in C:\repos\statcan\StatCan.OrchardCore\src\Modules\StatCan.OrchardCore.Hackathon\HackathonMigrations.cs:line 31
         at OrchardCore.Data.Migration.DataMigrationManager.UpdateAsync(String featureId)
Unhandled exception. Npgsql.NpgsqlException (0x80004005): Received backend message ReadyForQuery while expecting CommandCompleteMessage. Please file a bug.
   at Npgsql.NpgsqlConnector.ExecuteInternalCommand(Byte[] data, Boolean async, CancellationToken cancellationToken)
   at Npgsql.NpgsqlConnector.Rollback(Boolean async, CancellationToken cancellationToken)
   at Npgsql.NpgsqlTransaction.Rollback()
   at Npgsql.NpgsqlTransaction.Dispose(Boolean disposing)
   at YesSql.Session.ReleaseTransaction()
   at YesSql.Session.Cancel()
   at YesSql.Session.FlushAsync()
   at YesSql.Services.DefaultQuery.Query`1.ListImpl()
   at StatCan.OrchardCore.Hackathon.Indexes.IndexMigrations.CreateHackathonUsersIndex() in C:\repos\statcan\StatCan.OrchardCore\src\Modules\StatCan.OrchardCore.Hackathon\Indexes\Migrations.cs:line 80
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__140_1(Object state)
   at System.Threading.QueueUserWorkItemCallback.<>c.<.cctor>b__6_0(QueueUserWorkItemCallback quwi)
   at System.Threading.ExecutionContext.RunForThreadPoolUnsafe[TState](ExecutionContext executionContext, Action`1 callback, TState& state)
   at System.Threading.QueueUserWorkItemCallback.Execute()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
```

I will add to this issue when I find more information. Just investigating right now.

Skrypt:
Fixes #8545

## TODO 

Consider adding documentation on index name convention.
Right now, index names are redundant. We use the word "index" when they start with "IDX".
Maybe worth to at least name them, from now, consistently by using a convention.

Here in this PR for example I'm using

| Prefix        | Table Name (simplified) | Column Name | Column Name | 
| ------------- |:-------------:| -----:|-----:|
IDX | WFBA | DocumentId | ActivityId |
IDX | WFBA | DocumentId | ActivityName | 

| Name | Previous Name |
| ------------- |:-------------|
IDX_WFBA_DocumentId_ActivityId | IDX_WorkflowBlockingActivitiesIndex_DocumentId_ActivityId |
IDX_WFBA_DocumentId_ActivityName | IDX_WorkflowBlockingActivitiesIndex_DocumentId_ActivityName |

## Functional Tests - all Databases

Before: https://github.com/OrchardCMS/OrchardCore/actions/runs/549629776
After: https://github.com/OrchardCMS/OrchardCore/actions/runs/549605287

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Workflows/Migrations.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Workflows/Migrations.cs
@@ -68,7 +68,7 @@ namespace OrchardCore.Workflows
             );
 
             SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
-                .CreateIndex("IDX_WFBA_DocumentId_ActivityId",
+                .CreateIndex("IDX_WFBlockingActivities_DocumentId_ActivityId",
                     "DocumentId",
                     "ActivityId",
                     "WorkflowTypeId",
@@ -76,7 +76,7 @@ namespace OrchardCore.Workflows
             );
 
             SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
-                .CreateIndex("IDX_WFBA_DocumentId_ActivityName",
+                .CreateIndex("IDX_WFBlockingActivities_DocumentId_ActivityName",
                     "DocumentId",
                     "ActivityName",
                     "WorkflowTypeId",
@@ -129,7 +129,7 @@ namespace OrchardCore.Workflows
             );
 
             SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
-                .CreateIndex("IDX_WFBA_DocumentId_ActivityId",
+                .CreateIndex("IDX_WFBlockingActivities_DocumentId_ActivityId",
                     "DocumentId",
                     "ActivityId",
                     "WorkflowTypeId",
@@ -137,7 +137,7 @@ namespace OrchardCore.Workflows
             );
 
             SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
-                .CreateIndex("IDX_WFBA_DocumentId_ActivityName",
+                .CreateIndex("IDX_WFBlockingActivities_DocumentId_ActivityName",
                     "DocumentId",
                     "ActivityName",
                     "WorkflowTypeId",
--- a/src/OrchardCore.Modules/OrchardCore.Workflows/Migrations.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Workflows/Migrations.cs
@@ -84,7 +84,7 @@ namespace OrchardCore.Workflows
             );
 
             // Shortcut other migration steps on new content definition schemas.
-            return 4;
+            return 3;
         }
 
         // This code can be removed in a later version.
@@ -129,38 +129,6 @@ namespace OrchardCore.Workflows
             );
 
             SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
-                .CreateIndex("IDX_WorkflowBlockingActivitiesIndex_DocumentId_ActivityId",
-                    "DocumentId",
-                    "ActivityId",
-                    "WorkflowTypeId",
-                    "WorkflowId")
-            );
-
-            SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
-                .CreateIndex("IDX_WorkflowBlockingActivitiesIndex_DocumentId_ActivityName",
-                    "DocumentId",
-                    "ActivityName",
-                    "WorkflowTypeId",
-                    "WorkflowCorrelationId")
-            );
-
-            return 3;
-        }
-
-        // This code can be removed in a later version.
-        public int UpdateFrom3()
-        {
-            //For PostgreSQL
-            SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
-                .DropIndex("IDX_WorkflowBlockingActivitiesIndex_DocumentId_Activity"));
-
-            SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
-                .DropIndex("IDX_WorkflowBlockingActivitiesIndex_DocumentId_ActivityId"));
-
-            SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
-                .DropIndex("IDX_WorkflowBlockingActivitiesIndex_DocumentId_ActivityName"));
-
-            SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
                 .CreateIndex("IDX_WFBA_DocumentId_ActivityId",
                     "DocumentId",
                     "ActivityId",
@@ -176,7 +144,7 @@ namespace OrchardCore.Workflows
                     "WorkflowCorrelationId")
             );
 
-            return 4;
+            return 3;
         }
     }
 }
--- a/src/OrchardCore.Modules/OrchardCore.Workflows/Migrations.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Workflows/Migrations.cs
@@ -84,7 +84,7 @@ namespace OrchardCore.Workflows
             );
 
             // Shortcut other migration steps on new content definition schemas.
-            return 3;
+            return 4;
         }
 
         // This code can be removed in a later version.
--- a/src/OrchardCore.Modules/OrchardCore.Workflows/Migrations.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Workflows/Migrations.cs
@@ -68,7 +68,7 @@ namespace OrchardCore.Workflows
             );
 
             SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
-                .CreateIndex("IDX_WorkflowBlockingActivitiesIndex_DocumentId_ActivityId",
+                .CreateIndex("IDX_WFBA_DocumentId_ActivityId",
                     "DocumentId",
                     "ActivityId",
                     "WorkflowTypeId",
@@ -76,7 +76,7 @@ namespace OrchardCore.Workflows
             );
 
             SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
-                .CreateIndex("IDX_WorkflowBlockingActivitiesIndex_DocumentId_ActivityName",
+                .CreateIndex("IDX_WFBA_DocumentId_ActivityName",
                     "DocumentId",
                     "ActivityName",
                     "WorkflowTypeId",
@@ -145,6 +145,38 @@ namespace OrchardCore.Workflows
             );
 
             return 3;
+        }
+
+        // This code can be removed in a later version.
+        public int UpdateFrom3()
+        {
+            //For PostgreSQL
+            SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
+                .DropIndex("IDX_WorkflowBlockingActivitiesIndex_DocumentId_Activity"));
+
+            SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
+                .DropIndex("IDX_WorkflowBlockingActivitiesIndex_DocumentId_ActivityId"));
+
+            SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
+                .DropIndex("IDX_WorkflowBlockingActivitiesIndex_DocumentId_ActivityName"));
+
+            SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
+                .CreateIndex("IDX_WFBA_DocumentId_ActivityId",
+                    "DocumentId",
+                    "ActivityId",
+                    "WorkflowTypeId",
+                    "WorkflowId")
+            );
+
+            SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
+                .CreateIndex("IDX_WFBA_DocumentId_ActivityName",
+                    "DocumentId",
+                    "ActivityName",
+                    "WorkflowTypeId",
+                    "WorkflowCorrelationId")
+            );
+
+            return 4;
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/OrchardCore.Modules/OrchardCore.Workflows/Migrations.cs
```diff
@@ -146,5 +146,37 @@ public int UpdateFrom2()
 
             return 3;
         }
+
+        // This code can be removed in a later version.
+        public int UpdateFrom3()
+        {
+            //For PostgreSQL
+            SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
+                .DropIndex("IDX_WorkflowBlockingActivitiesIndex_DocumentId_Activity"));
```
Skrypt, line 155: @jptissot Can you confirm that it creates this index on your PostgreSQL database on the WorkflowBlockingActivities table.

jtkech:
@Skrypt I will approve

But the problem is that dropping indexes don't work for all providers

> IF EXISTS
> Applies to: SQL Server ( SQL Server 2016 (13.x) through current version).

One option is to not support older SQL versions, in that case, because we aded all `DocumentId` indexes, we could add a migration step to all modules to remove all the redundant indexes.

@Skrypt for now i think it is better to just update the names without creating an `UpdateFrom3()`, as i think that if it failed on a `PostgreSQL` installation it remained on version 2, and quickly merge this PR ;)

There is another index to rename: `IDX_ContentPickerFieldIndex_DocumentId_SelectedContentItemId`, so idem for this one, just rename it without a new migration step in the related migrations class.

Yes, i agree with a new naming convention for new indexes, to remove the `Index` part and to use acronyms, maybe not full acronyms, e.g `WFBlockingActivities` in place of `WFBA`.

So, for now, to be more consistent with the existing names in existing migrations, i would suggest e.g.

    IDX_WFBlockingActivities_DocumentId_ActivityId
    IDX_WFBlockingActivities_DocumentId_ActivityName
    IDX_ContentPickerField_DocumentId_SelectedItemId

---

No luck, i did a func test on all databases, but i missed it after adding the last index related to `ActivityName`. On google they say it is 63 chars and `IDX_WorkflowBlockingActivitiesIndex_DocumentId_ActivityName` = 59 chars, maybe it depends on the version, would be interesting to know the real limitation, based on your test seems to be 55 chars. At least we need to be aware of this limitation and to confirm the real one, and yes good to have a new convention.

sfmskywalker (approved):

deanmarcussen (commented):
@Skrypt agree with @jtkech here.

We don't need the migration to rename, just change the names as you have done with the migration.

Skrypt:
I don't think it makes a big difference if I add an UpdateFrom3() method here since these are marked to be removed later on.
I just makes sure that the migration will get executed, for people that uses < SQL Server 2016 I think it's too bad simply.
We either need to fix the code on YesSQL side later on or tell these people to upgrade to at least 2016.

We can clearly see in the stack trace that under PostgreSQL it tried to create the index twice which is something that we can at least fix with a migration. Else, why we have this DropIndex method? I understand your comments but here I'm trying to fix the migration issue like it should be fixed normally, if the migration failed previously it left trails which I need to remove.

Skrypt:
```SQL
/*
    Create unique index 
*/

-- Old block of code
IF  EXISTS (SELECT * FROM sys.indexes WHERE NAME = N'IDX_TOY_BRANDS_NAME') 
DROP INDEX [IDX_TOY_BRANDS_NAME] ON [TOY].[BRANDS]
GO

-- New block of code
DROP INDEX IF EXISTS [IDX_TOY_BRANDS_NAME] ON [TOY].[BRANDS]
GO

-- Add new index
CREATE UNIQUE INDEX [IDX_TOY_BRANDS_NAME] ON [TOY].[BRANDS] (NAME);  
GO
```

https://www.mssqltips.com/sqlservertip/4402/new-drop-if-exists-syntax-in-sql-server-2016/

Skrypt:
Ok, made the changes requested. Though I kept the naming because I think that tenant names might make these index names get longer. The smaller these index names are the better "I think".

jtkech:
@Skrypt 

If we would have kept the UpgradeFrom3, at least we would have to clear the UpgradeFrom2, otherwise on PostgreSQL it would still fail for a migration from 2 to 3, the path for them being to add, then remove, then re-add these indexes, so because it is a recent change

> Though I kept the naming because I think that tenant names might make these index names get longer

As i saw on sql server, the tenant prefix is used for the table name but not for the indexes on this table, otherwise there would be many other index names to reduce.

jtkech:
@Skrypt 

Okay to use the shortest names ;)

What about also reducing `IDX_ContentPickerFieldIndex_DocumentId_SelectedContentItemId` in this PR?

Skrypt:
> As i saw on sql server, the tenant prefix is used for the table name but not for the indexes on this table, otherwise there would be many other index names to reduce.

Ok so basically these index names should never use the table name in them. They are associated with the table already.

jtkech:
@Skrypt 

Wow, good catch, so just e.g. `IDX_DocumentId_ActivityId`, even better !!!

And for the remainig one just `IDX_DocumentId_SelectedContentItemId `

jptissot:
> Edit : Ok we keep the full table names because of SQLite.

Is there a way to have an exception for SQLite in Yessql instead of making all db's use a very long index name ?

jtkech:
> Edit : Ok we keep the full table names because of SQLite.

No luck :(

Skrypt:
No here see a SQLite database : 

image

The Indices are not related with a table. This is why we are using the table name in the index names.

Skrypt:
> Is there a way to have an exception for SQLite in Yessql instead of making all db's use a very long index name ?

Maybe but that would be on YesSQL side. We pass a custom index name when we create it so we could pass the table name as a parameter to that method.

jtkech:
@Skrypt 

Okay, so as i understand, for sqlite the index name doesn't need to include the table name, but it has to be unique in the database, so how you reduced the names here is okay, right?

Skrypt:
Yes it doesn't need to use the table name but it's an identifier to make a visual relation between the table and it's indices (for SQLite). So, the names are fine but what @jptissot is suggesting would be the appropriate thing to do. Simply don't use the table name for every db providers except SQLite. Then, it requires a change in the current CreateIndex method to use the table name and append it to the index name or not.

```C#
            SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
                .CreateIndex("IDX_WFBA_DocumentId_ActivityName",
                    "DocumentId",
                    "ActivityName",
                    "WorkflowTypeId",
                    "WorkflowCorrelationId")
            );
```

```C#
            SchemaBuilder.AlterIndexTable<WorkflowBlockingActivitiesIndex>(table => table
                .CreateIndex("IDX_WFBA_DocumentId_ActivityName",
                    "DocumentId",
                    "ActivityName",
                    "WorkflowTypeId",
                    "WorkflowCorrelationId", nameof(WorkflowBlockingActivitiesIndex)) // table name as optional param.
            );
```

If you look at this SchemaBuilder we already kinda pass the table name based on the WorkflowBlockingActivitesIndex model. Then for SQLite we would need to just add this table name.

Skrypt:
Or maybe a better idea would be to have this as a static class in Orchard Core with a static method that would return the proper index name based on the current db provider.

Skrypt:
```C#
public static class IndexNameProvider
{
   private string Prefix = "IDX"; // make this configurable eventually ...
   
   public static method GetIndexName(string[] columnNames, string tableName)
   {
       indexName = Prefix + "_" + tableName;
       
       foreach(string column in columnNames)
       {
           indexName += "_" + column;
       }
   }
}
`

jtkech (approved):

PR #8550, opened 2021-02-09
Add net5.0 and netcoreapp3.1 target framework multi-targetting

dafergu2:
This PR enables the built nuget packages to target both .NET 5.0 and .NET Core 3.1.

It does this by updating the `src/OrchardCore.Build/Dependencies.AspNetCore.props` to target multiple frameworks instead of a single one.

Significant changes include:

1) Removing the `<TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework>` from all projects.  Due to the `Directory.Build.props` files in the source structure, the `Dependencies.AspNetCore.props` file is always imported in project builds.  This makes the TF node declaration in each project redundant and possibly increases maintenance effort.

    This was done in stages, the first being commenting them out, the second being removing the comment in its own commit.  This allows an easy revert to have the project nodes without removing the other changes in the .props file.  This could be useful for testing if needed.

1) Several code changes needed to be made to support the .NET 5.0 SDK tool changes.  In particular, code analysis complained of using `throw ex` in favor `throw`, thus the change in cfb396994d1404b0091f742ce1810636adc631b6.  In addition, the compiler complained when using the `QueryString` struct potentially without initializing it.  A simple fix was to using a string with the same logic instead.

1)  When running dotnet pack, projects in the `OrchardCore.Modules` folder failed with a NU5129 error indicating the <AssemblyName>.props file for the project was missing from the `build\` folder.  Using a msbuild binary log (`dotnet pack --no-build /bl`) and MSBuild Structure Log Viewer (in my case, the Blazor version at https://github.com/laurenprinn/MSBuildStructuredLog due to being on a mac...awesome job on this Lauren!) I was able to track it down to the fact that the props file was being placed in the TFM specific folders.  Multi-targetting needs the .props file to be in the non-TFM parent folder, so the fix was to simply remove the `$(TargetFramework)` path segment in the different files (see f6fed10).

# Known Issues and follow-up
* The `Microsoft.AspNetCore.Authentication.AzureAD.UI` package when updated to 5.0.1 indicated that some api's in use had been deprecated.  Rather than fix the warnings and test that package, the warnings were ignored via pragma statements and need to be addressed in the future.
*  During testing, random build failures occurred when building the Templates due to a template.config directory not being able to be deleted.  This could be due to a timing issue, but I haven't been able to track down why this happened.  Repeating the build generally solves the problem.
*  Comprehensive integration testing has not been performed locally.  Small tests will be started shortly.

### DIFF
--- a/src/OrchardCore.Build/Dependencies.AspNetCore.props
+++ b/src/OrchardCore.Build/Dependencies.AspNetCore.props
@@ -9,6 +9,7 @@
     <When Condition=" '$(TargetFramework)' == 'net5.0' ">
       <PropertyGroup>
         <AspNetCoreVersion>5.0.3</AspNetCoreVersion>
+        <MicrosoftExtensionsVersion>5.0.1</MicrosoftExtensionsVersion>
       </PropertyGroup>
     </When>
     <Otherwise>
@@ -19,10 +20,12 @@
       -->
       <PropertyGroup>
         <AspNetCoreVersion>3.1.12</AspNetCoreVersion>
+        <MicrosoftExtensionsVersion>3.1.12</MicrosoftExtensionsVersion>
       </PropertyGroup>
     </Otherwise>
   </Choose>
 
+  <!-- Add all of the AspNetCore packages -->
   <ItemGroup>
     <PackageManagement Include="Microsoft.AspNetCore.Authentication.AzureAD.UI" Version="$(AspNetCoreVersion)" />
     <PackageManagement Include="Microsoft.AspNetCore.Authentication.Facebook" Version="$(AspNetCoreVersion)" />
@@ -39,15 +42,10 @@
     <PackageManagement Include="Microsoft.Extensions.Http.Polly" Version="$(AspNetCoreVersion)" />
   </ItemGroup>
 
-  <!-- For .NET 5.0 packages that do not have a version number that matches $(AspNetCoreVersion) available -->
-  <ItemGroup Condition=" '$(TargetFramework)' == 'net5.0' ">
-
-    <!-- Fixes error NU1102: Unable to find package Microsoft.Extensions.Http.Polly with version (>= 5.0.3), Found 50 version(s) in NuGet [ Nearest version: 5.0.1 ]-->
-    <PackageManagement Remove="Microsoft.Extensions.Http.Polly" />
-    <PackageManagement Include="Microsoft.Extensions.Http.Polly" Version="5.0.1" />
-
-    <!-- Fixes error NU1102: Unable to find package Microsoft.Extensions.Caching.StackExchangeRedis with version (>= 5.0.3), Found 45 version(s) in NuGet [ Nearest version: 5.0.1 ] -->
-    <PackageManagement Remove="Microsoft.Extensions.Caching.StackExchangeRedis" />
-    <PackageManagement Include="Microsoft.Extensions.Caching.StackExchangeRedis" Version="5.0.1" />
+  <!-- Add all Microsoft.Extensions packages -->
+  <ItemGroup>
+    <PackageManagement Include="Microsoft.Extensions.Caching.StackExchangeRedis" Version="$(MicrosoftExtensionsVersion)" />
+    <PackageManagement Include="Microsoft.Extensions.Http.Polly" Version="$(MicrosoftExtensionsVersion)" />
   </ItemGroup>
+
 </Project>
--- a/src/OrchardCore.Build/Dependencies.AspNetCore.props
+++ b/src/OrchardCore.Build/Dependencies.AspNetCore.props
@@ -1,10 +1,32 @@
 <Project>
   <PropertyGroup>
-    <AspNetCoreVersion>3.1.12</AspNetCoreVersion>
-    <AspNetCoreTargetFramework>netcoreapp3.1</AspNetCoreTargetFramework>
+    <AspNetCoreTargetFrameworks>netcoreapp3.1;net5.0</AspNetCoreTargetFrameworks>
+    <TargetFrameworks>$(AspNetCoreTargetFrameworks)</TargetFrameworks>
   </PropertyGroup>
+  <Choose>
+    <When Condition=" '$(TargetFramework)' == 'net5.0' ">
+      <PropertyGroup>
+        <AspNetCoreVersion>5.0.1</AspNetCoreVersion>
+      </PropertyGroup>
+    </When>
+    <Otherwise>
+      <!--
+      Fallback when not targetting .net 5.0.
+      At this time, that would mean .net core 3.1.  If there are more target frameworks in the future,
+      this logic should change slightly.
+      -->
+      <PropertyGroup>
+        <AspNetCoreVersion>3.1.12</AspNetCoreVersion>
+      </PropertyGroup>
+    </Otherwise>
+  </Choose>
+
   <ItemGroup>
-    <PackageManagement Include="Microsoft.AspNetCore.Authentication.AzureAD.UI" Version="$(AspNetCoreVersion)" />
+    <!-- Needs to stay at older package due to items being obsoleted -->
+    <PackageManagement Include="Microsoft.AspNetCore.Authentication.AzureAD.UI" Version="3.1.12" />
+  </ItemGroup>
+
+  <ItemGroup>
     <PackageManagement Include="Microsoft.AspNetCore.Authentication.Facebook" Version="$(AspNetCoreVersion)" />
     <PackageManagement Include="Microsoft.AspNetCore.Authentication.Google" Version="$(AspNetCoreVersion)" />
     <PackageManagement Include="Microsoft.AspNetCore.Authentication.MicrosoftAccount" Version="$(AspNetCoreVersion)" />
--- a/src/OrchardCore.Build/Dependencies.AspNetCore.props
+++ b/src/OrchardCore.Build/Dependencies.AspNetCore.props
@@ -38,8 +38,6 @@
     <PackageManagement Include="Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation" Version="$(AspNetCoreVersion)" />
     <PackageManagement Include="Microsoft.AspNetCore.Mvc.Testing" Version="$(AspNetCoreVersion)" />
     <PackageManagement Include="Microsoft.AspNetCore.Owin" Version="$(AspNetCoreVersion)" />
-    <PackageManagement Include="Microsoft.Extensions.Caching.StackExchangeRedis" Version="$(AspNetCoreVersion)" />
-    <PackageManagement Include="Microsoft.Extensions.Http.Polly" Version="$(AspNetCoreVersion)" />
   </ItemGroup>
 
   <!-- Add all Microsoft.Extensions packages -->
--- a/src/OrchardCore.Build/Dependencies.AspNetCore.props
+++ b/src/OrchardCore.Build/Dependencies.AspNetCore.props
@@ -1,12 +1,14 @@
 <Project>
   <PropertyGroup>
+    <!-- The default TFM for web app and console projects is net5.0 -->
+    <AspNetCoreTargetFramework>net5.0</AspNetCoreTargetFramework>
     <AspNetCoreTargetFrameworks>netcoreapp3.1;net5.0</AspNetCoreTargetFrameworks>
     <TargetFrameworks>$(AspNetCoreTargetFrameworks)</TargetFrameworks>
   </PropertyGroup>
   <Choose>
     <When Condition=" '$(TargetFramework)' == 'net5.0' ">
       <PropertyGroup>
-        <AspNetCoreVersion>5.0.1</AspNetCoreVersion>
+        <AspNetCoreVersion>5.0.3</AspNetCoreVersion>
       </PropertyGroup>
     </When>
     <Otherwise>
@@ -22,11 +24,7 @@
   </Choose>
 
   <ItemGroup>
-    <!-- Needs to stay at older package due to items being obsoleted -->
-    <PackageManagement Include="Microsoft.AspNetCore.Authentication.AzureAD.UI" Version="3.1.12" />
-  </ItemGroup>
-
-  <ItemGroup>
+    <PackageManagement Include="Microsoft.AspNetCore.Authentication.AzureAD.UI" Version="$(AspNetCoreVersion)" />
     <PackageManagement Include="Microsoft.AspNetCore.Authentication.Facebook" Version="$(AspNetCoreVersion)" />
     <PackageManagement Include="Microsoft.AspNetCore.Authentication.Google" Version="$(AspNetCoreVersion)" />
     <PackageManagement Include="Microsoft.AspNetCore.Authentication.MicrosoftAccount" Version="$(AspNetCoreVersion)" />
@@ -39,5 +37,23 @@
     <PackageManagement Include="Microsoft.AspNetCore.Owin" Version="$(AspNetCoreVersion)" />
     <PackageManagement Include="Microsoft.Extensions.Caching.StackExchangeRedis" Version="$(AspNetCoreVersion)" />
     <PackageManagement Include="Microsoft.Extensions.Http.Polly" Version="$(AspNetCoreVersion)" />
+  </ItemGroup>
+
+  <!-- For .NET 5.0 packages that do not have a version number that matches $(AspNetCoreVersion) available -->
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net5.0' ">
+    <!-- Fixes the following obsolete warnings used across multiple files in OrchardCore.Microsoft.Authentication
+      warning CS0618: 'AzureADOptions' is obsolete: 'This is obsolete and will be removed in a future version. Use Microsoft.Identity.Web instead. See https://aka.ms/ms-identity-web
+      warning CS0618: 'AzureADDefaults' is obsolete: 'This is obsolete and will be removed in a future version. Use Microsoft.Identity.Web instead. See https://aka.ms/ms-identity-web.
+    -->
+    <PackageManagement Remove="Microsoft.AspNetCore.Authentication.AzureAD.UI" />
+    <PackageManagement Include="Microsoft.AspNetCore.Authentication.AzureAD.UI" Version="3.1.12" />
+
+    <!-- Fixes error NU1102: Unable to find package Microsoft.Extensions.Http.Polly with version (>= 5.0.3), Found 50 version(s) in NuGet [ Nearest version: 5.0.1 ]-->
+    <PackageManagement Remove="Microsoft.Extensions.Http.Polly" />
+    <PackageManagement Include="Microsoft.Extensions.Http.Polly" Version="5.0.1" />
+
+    <!-- Fixes error NU1102: Unable to find package Microsoft.Extensions.Caching.StackExchangeRedis with version (>= 5.0.3), Found 45 version(s) in NuGet [ Nearest version: 5.0.1 ] -->
+    <PackageManagement Remove="Microsoft.Extensions.Caching.StackExchangeRedis" />
+    <PackageManagement Include="Microsoft.Extensions.Caching.StackExchangeRedis" Version="5.0.1" />
   </ItemGroup>
 </Project>
--- a/src/OrchardCore.Build/Dependencies.AspNetCore.props
+++ b/src/OrchardCore.Build/Dependencies.AspNetCore.props
@@ -41,12 +41,6 @@
 
   <!-- For .NET 5.0 packages that do not have a version number that matches $(AspNetCoreVersion) available -->
   <ItemGroup Condition=" '$(TargetFramework)' == 'net5.0' ">
-    <!-- Fixes the following obsolete warnings used across multiple files in OrchardCore.Microsoft.Authentication
-      warning CS0618: 'AzureADOptions' is obsolete: 'This is obsolete and will be removed in a future version. Use Microsoft.Identity.Web instead. See https://aka.ms/ms-identity-web
-      warning CS0618: 'AzureADDefaults' is obsolete: 'This is obsolete and will be removed in a future version. Use Microsoft.Identity.Web instead. See https://aka.ms/ms-identity-web.
-    -->
-    <PackageManagement Remove="Microsoft.AspNetCore.Authentication.AzureAD.UI" />
-    <PackageManagement Include="Microsoft.AspNetCore.Authentication.AzureAD.UI" Version="3.1.12" />
 
     <!-- Fixes error NU1102: Unable to find package Microsoft.Extensions.Http.Polly with version (>= 5.0.3), Found 50 version(s) in NuGet [ Nearest version: 5.0.1 ]-->
     <PackageManagement Remove="Microsoft.Extensions.Http.Polly" />
--- a/src/OrchardCore.Build/Dependencies.AspNetCore.props
+++ b/src/OrchardCore.Build/Dependencies.AspNetCore.props
@@ -15,7 +15,7 @@
     <Otherwise>
       <!--
       Fallback when not targetting .net 5.0.
-      At this time, that would mean .net core 3.1.  If there are more target frameworks in the future,
+      At this time, that would mean .net core 3.1. If there are more target frameworks in the future,
       this logic should change slightly.
       -->
       <PropertyGroup>
--- a/src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
+++ b/src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
@@ -6,7 +6,7 @@
   <Import Project="..\OrchardCore\OrchardCore.Application.Cms.Core.Targets\OrchardCore.Application.Cms.Core.Targets.props" />
 
   <PropertyGroup>
-    <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework>
+    <!-- <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework> -->
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <WarningsNotAsErrors>612,618</WarningsNotAsErrors>
     <AspNetCoreHostingModel>InProcess</AspNetCoreHostingModel>
--- a/src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
+++ b/src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
@@ -6,7 +6,6 @@
   <Import Project="..\OrchardCore\OrchardCore.Application.Cms.Core.Targets\OrchardCore.Application.Cms.Core.Targets.props" />
 
   <PropertyGroup>
-    <!-- <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework> -->
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <WarningsNotAsErrors>612,618</WarningsNotAsErrors>
     <AspNetCoreHostingModel>InProcess</AspNetCoreHostingModel>
--- a/src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
+++ b/src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
@@ -6,6 +6,7 @@
   <Import Project="..\OrchardCore\OrchardCore.Application.Cms.Core.Targets\OrchardCore.Application.Cms.Core.Targets.props" />
 
   <PropertyGroup>
+    <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <WarningsNotAsErrors>612,618</WarningsNotAsErrors>
     <AspNetCoreHostingModel>InProcess</AspNetCoreHostingModel>
--- a/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs
@@ -319,11 +319,10 @@ namespace OrchardCore.Tenants.Controllers
                 pathString = pathString.Add('/' + shellSettings.RequestUrlPrefix);
             }
 
-            QueryString queryString;
-
+            var queryString = String.Empty;
             if (!String.IsNullOrEmpty(token))
             {
-                queryString = QueryString.Create("token", token);
+                queryString = QueryString.Create("token", token).ToString();
             }
 
             return $"{Request.Scheme}://{hostString + pathString + queryString}";
--- a/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs
@@ -319,7 +319,7 @@ namespace OrchardCore.Tenants.Controllers
                 pathString = pathString.Add('/' + shellSettings.RequestUrlPrefix);
             }
 
-            QueryString queryString;
+            QueryString queryString = null;
             if (!String.IsNullOrEmpty(token))
             {
                 queryString = QueryString.Create("token", token);
--- a/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs
@@ -319,10 +319,10 @@ namespace OrchardCore.Tenants.Controllers
                 pathString = pathString.Add('/' + shellSettings.RequestUrlPrefix);
             }
 
-            var queryString = String.Empty;
+            QueryString queryString;
             if (!String.IsNullOrEmpty(token))
             {
-                queryString = QueryString.Create("token", token).ToString();
+                queryString = QueryString.Create("token", token);
             }
 
             return $"{Request.Scheme}://{hostString + pathString + queryString}";
--- a/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs
@@ -319,7 +319,7 @@ namespace OrchardCore.Tenants.Controllers
                 pathString = pathString.Add('/' + shellSettings.RequestUrlPrefix);
             }
 
-            QueryString queryString = null;
+            QueryString queryString = QueryString.Empty;
             if (!String.IsNullOrEmpty(token))
             {
                 queryString = QueryString.Create("token", token);
--- a/src/OrchardCore/OrchardCore.Application.Cms.Core.Targets/OrchardCore.Application.Cms.Core.Targets.csproj
+++ b/src/OrchardCore/OrchardCore.Application.Cms.Core.Targets/OrchardCore.Application.Cms.Core.Targets.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework>
+    <!-- <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework> -->
      <!-- NuGet properties-->
     <Title>Target for Modular OrchardCore CMS Application</Title>
     <Description>$(OCCMSDescription)
--- a/src/OrchardCore/OrchardCore.Application.Cms.Core.Targets/OrchardCore.Application.Cms.Core.Targets.csproj
+++ b/src/OrchardCore/OrchardCore.Application.Cms.Core.Targets/OrchardCore.Application.Cms.Core.Targets.csproj
@@ -1,11 +1,10 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <!-- <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework> -->
      <!-- NuGet properties-->
     <Title>Target for Modular OrchardCore CMS Application</Title>
     <Description>$(OCCMSDescription)
-    
+
     Converts the application into a modular OrchardCore CMS application with TheAdmin theme but without any front-end Themes.</Description>
     <PackageTags>$(PackageTags) OrchardCoreCMS CMS</PackageTags>
   </PropertyGroup>
@@ -24,7 +23,7 @@
     When a package is not directly referenced, e.g only through the reference of this bundle package,
     the files in its build folder are not evaluated on building if this folder is marked as private.
     This can be defined by using the 'PrivateAssets' attribute.
-    
+
     Here, only project references are used but when packing the bundle they become package references,
     and with the same 'PrivateAssets' attribute.
   -->
@@ -32,8 +31,8 @@
   <ItemGroup>
     <ProjectReference Include="..\OrchardCore.Application.Targets\OrchardCore.Application.Targets.csproj" PrivateAssets="none" />
 
-    <!-- 
-      When adding a reference to this list, please keep it ordered alphabetically, and set PrivateAssets="none" 
+    <!--
+      When adding a reference to this list, please keep it ordered alphabetically, and set PrivateAssets="none"
     -->
     <ProjectReference Include="..\..\OrchardCore.Modules\OrchardCore.Admin\OrchardCore.Admin.csproj" PrivateAssets="none" />
     <ProjectReference Include="..\..\OrchardCore.Modules\OrchardCore.AdminDashboard\OrchardCore.AdminDashboard.csproj" PrivateAssets="none" />
--- a/src/OrchardCore/OrchardCore.Application.Cms.Core.Targets/OrchardCore.Application.Cms.Core.Targets.csproj
+++ b/src/OrchardCore/OrchardCore.Application.Cms.Core.Targets/OrchardCore.Application.Cms.Core.Targets.csproj
@@ -9,7 +9,20 @@
     <PackageTags>$(PackageTags) OrchardCoreCMS CMS</PackageTags>
   </PropertyGroup>
 
-  <!-- Add ".props" and ".targets" files in the package specific to the "Cms.Core" bundle -->
+  <!--
+    Add ".props" and ".targets" files in the package specific to the "Cms.Core" nuget package
+
+    When the OrchardCore.Application.Cms.Core.Targets package is referenced, the .props and .targets
+    files present in the build folder will also be referenced by the generated .props and .targets
+    files in the obj folder of the referencing project.
+
+    See: https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package#include-msbuild-props-and-targets-in-a-package
+    * The files in the build folder will be applied to any TFM.
+    * The files in the build\$(TargetFramework) folder will be applied to only referencing projects targeting that TFM
+    * The files in the buildMultiTargeting folder will only be applied when the referencing project's $(TargetFramework)
+      is empty
+  -->
+
   <ItemGroup>
     <None Include="OrchardCore.Application.Cms.Core.Targets.props" Pack="true">
       <PackagePath>build\OrchardCore.Application.Cms.Core.Targets.props</PackagePath>
--- a/src/OrchardCore/OrchardCore.Application.Cms.Core.Targets/OrchardCore.Application.Cms.Core.Targets.csproj
+++ b/src/OrchardCore/OrchardCore.Application.Cms.Core.Targets/OrchardCore.Application.Cms.Core.Targets.csproj
@@ -12,10 +12,10 @@
   <!-- Add ".props" and ".targets" files in the package specific to the "Cms.Core" bundle -->
   <ItemGroup>
     <None Include="OrchardCore.Application.Cms.Core.Targets.props" Pack="true">
-      <PackagePath>build\$(TargetFramework)\OrchardCore.Application.Cms.Core.Targets.props</PackagePath>
+      <PackagePath>build\OrchardCore.Application.Cms.Core.Targets.props</PackagePath>
     </None>
     <None Include="OrchardCore.Application.Cms.Core.Targets.targets" Pack="true">
-      <PackagePath>build\$(TargetFramework)\OrchardCore.Application.Cms.Core.Targets.targets</PackagePath>
+      <PackagePath>build\OrchardCore.Application.Cms.Core.Targets.targets</PackagePath>
     </None>
   </ItemGroup>
 
--- a/src/OrchardCore/OrchardCore.Configuration.KeyVault/OrchardCore.Configuration.KeyVault.csproj
+++ b/src/OrchardCore/OrchardCore.Configuration.KeyVault/OrchardCore.Configuration.KeyVault.csproj
@@ -1,9 +1,5 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
-  <PropertyGroup>
-    <!-- <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework> -->
-  </PropertyGroup>
-
   <ItemGroup>
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
   </ItemGroup>
--- a/src/OrchardCore/OrchardCore.Configuration.KeyVault/OrchardCore.Configuration.KeyVault.csproj
+++ b/src/OrchardCore/OrchardCore.Configuration.KeyVault/OrchardCore.Configuration.KeyVault.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework>
+    <!-- <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework> -->
   </PropertyGroup>
 
   <ItemGroup>
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,7 +6,7 @@
 			"type": "coreclr",
 			"request": "launch",
 			"preLaunchTask": "Web App: Dotnet Build Debug",
-			"program": "${workspaceRoot}/src/OrchardCore.Cms.Web/bin/Debug/netcoreapp3.1/OrchardCore.Cms.Web.dll",
+			"program": "${workspaceRoot}/src/OrchardCore.Cms.Web/bin/Debug/net5.0/OrchardCore.Cms.Web.dll",
 			"args": [],
 			"cwd": "${workspaceRoot}/src/OrchardCore.Cms.Web",
 			"env": {
@@ -20,7 +20,7 @@
 			"request": "launch",
 			"internalConsoleOptions": "openOnSessionStart",
 			"preLaunchTask": "Web App: Dotnet Build Debug",
-			"program": "${workspaceRoot}/src/OrchardCore.Cms.Web/bin/Debug/netcoreapp3.1/OrchardCore.Cms.Web.dll",
+			"program": "${workspaceRoot}/src/OrchardCore.Cms.Web/bin/Debug/net5.0/OrchardCore.Cms.Web.dll",
 			"args": [],
 			"cwd": "${workspaceRoot}/src/OrchardCore.Cms.Web",
 			"env": {
@@ -47,7 +47,7 @@
 			"type": "coreclr",
 			"request": "launch",
 			"preLaunchTask": "Web App: Dotnet Build Release",
-			"program": "${workspaceRoot}/src/OrchardCore.Cms.Web/bin/Release/netcoreapp3.1/OrchardCore.Cms.Web.dll",
+			"program": "${workspaceRoot}/src/OrchardCore.Cms.Web/bin/Release/net5.0/OrchardCore.Cms.Web.dll",
 			"args": [],
 			"cwd": "${workspaceRoot}/src/OrchardCore.Cms.Web",
       "env": {
--- a/src/OrchardCore.Modules/OrchardCore.Admin/OrchardCore.Admin.csproj
+++ b/src/OrchardCore.Modules/OrchardCore.Admin/OrchardCore.Admin.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk.Razor">
 
   <PropertyGroup>
-    <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework>
+    <!-- <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework> -->
     <AddRazorSupportForMvc>true</AddRazorSupportForMvc>
 
     <!-- NuGet properties-->
--- a/src/OrchardCore.Modules/OrchardCore.Admin/OrchardCore.Admin.csproj
+++ b/src/OrchardCore.Modules/OrchardCore.Admin/OrchardCore.Admin.csproj
@@ -1,7 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk.Razor">
 
   <PropertyGroup>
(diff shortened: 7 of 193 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=8550) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=8550) before we can accept your contribution.<br/><br/>:x: dafergu2 [sign now](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=8550)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OrchardCMS/OrchardCore?pullRequest=8550) it.</sub>

src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
```diff
@@ -6,7 +6,6 @@
   <Import Project="..\OrchardCore\OrchardCore.Application.Cms.Core.Targets\OrchardCore.Application.Cms.Core.Targets.props" />
 
   <PropertyGroup>
-    <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework>
```
Skrypt, line 9: Here maybe it's because it needed to be renamed to AspNetCoreTargetFrameworks ?

Skrypt: <TargetFrameworks>$(AspNetCoreTargetFrameworks)</TargetFramework>

Skrypt, line 9: Have you tried with : 

`<TargetFrameworks>$(AspNetCoreTargetFrameworks)</TargetFrameworks>`

dafergu2: @Skrypt On line 3, the project imports `..\OrchardCore.Build\Dependencies.props`, which in turn imports `Dependencies.AspNetCore.props`.  Due to that, the `<TargetFrameworks>` property is already set which makes this redundant.

deanmarcussen:
Nice. Good work thank you.

We should update the tests to build and run against both frameworks (probably another pr).

dafergu2:
> Nice. Good work thank you.
> 
> We should update the tests to build and run against both frameworks (probably another pr).

@deanmarcussen Thanks!  What would we need to do to update the tests?  I believe `dotnet test` automatically runs the tests for both frameworks.  This is from the PR gate test build:

image

Did you mean the functional tests?

If you're good with this, should I move the PR out of draft?  I haven't had time to do a good manual test of the different packages locally but it seems like most everything is working based on the tests.  It was late last night when I finished staring at build logs.

Also, is there a standard for how the team wants to do framework dependent code?  Typically I've seen this done via preprocessor code, but I've also seen it done with separate implementations that is conditionally compiled (potentially excluded via the project file).  Since there is a heavy use of custom MSBuild tasks, it wouldn't be far from the mark to have target files that are custom per framework if needed.  Even in some cases splitting classes up via partial classes to allow for target framework/platform specific code.  It might be good to document that approach.

src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
Skrypt: Ok, though if I want to compile this assembly/module only "outside of OC build context" ; I then need to keep it in that folder so that it gets compiled. Ok, I don't mind. I'm pretty sure there was a reason that we kept it redundant but we will see. Keep it removed for now, no big issue there.

Skrypt:
Yes move it out of draft. This one is a keeper. Functional test in a second PR.

src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
dafergu2: Yes, I thought of that.  That's the reason that I split out the commits for removing it by first commenting it in the initial commit that made the TFM change, and then in a second commit removing the comments (and some inconsquential whitespace that my editor made due to the standard setting of removing trailing whitespace).  That way we can revert the second commit easily and then do a find and replace without removing the changes in the `Dependencies.AspNetCore.props` file.

I think a lot of the projects will not build properly outside of the OC build context due to the `<PackageManagement>` tasks set up to control the package versions and common dependencies at a central location.  It took me a bit to wade through the build dependency chain and understand what was going on.

dafergu2:
Moving out of draft.

dafergu2:
My initial tests of running the `OrchardCore.Cms.Web` project targeting .NET 5 seems to result in a much faster initial load and site creation.  That could be my imagination, though, as I've not been able to stopwatch it.

src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
jtkech: @dafergu2 

@Skrypt is right, and our web project file is also an example of the needed properties of a web project file

Personally i would have kept `$(AspNetCoreTargetFramework)`, but replaced by `$(AspNetCoreTargetFrameworks)`, in all project files, and kept the definition of `AspNetCoreTargetFramework` and add `AspNetCoreTargetFrameworks` in the global `Dependencies.AspNetCore.props`, so that each individual project can choose

I already tried multi targetting with OrchardCore, but a while ago so, for the `PackagePath` of msbuild scripts, i don't remember if i kept the `$(TargetFramework)` as all project are inner built 2 times but with the right TFM each time, or maybe if i used a `buildMultiTargetting` folder as aspnetcore does.

Need to focus on it again, then i will review it

dafergu2: @jtkech I had originally done that, but as I as was understanding the flow of the build tasks, it seemed to add more to confusion to have those extra prop settings.  An easy to miss typo with that also caused a significant amount of time to be consumed tracking it down when I was testing.  Each project can still override the `<TargetFramework>` property, it would just need to be added as `<TargetFramework>net5.0</TargetFramework>` or similar.

When multi-targeting, the `OrchardCore.Cms.Web` project will be built for both.  If you try to run it, it indicates that the framework needs to be specified with --framework.  Should there be a default setting or should the user specify?  I haven't tried building and running on an environment that has only the 3.1 framework or 5.0 framework installed, so I'm unsure of the exact behavior at that point.

Anecdotally, confusion on the $(AspNetCoreTargetFramework) MSBuild var was expressed by a coworker on my current project which we are implementing using Orchard Core.

I'm happy to add back in the properties if that's desired, though.  I wanted to walk through some of my thought process to show the change was made after some deliberation and not just at a whim. :)  All is good, just let me know and I'll make any changes desired.

With regards to multi-targeting and packaging, this was the larger challenge as it involved figuring out how the `Directory.Build.<targets|props>` files were working for the Modules and Themes packages.  The `PackagePath` was causing the assembly props file to be copied from the `src\OrchardCore\OrchardCore.Module.Targets\Package.Build.props`  to `build\netcoreapp3.1\<Assembly>.props` and `build\net5.0\<Assembly>.props` respectively for each TFM.  This was causing the nuget validation to issue a NU5129 error in which it complained that the `build\` folder (without a TFM subsegment) had a .props file but not one named in the convention of the <Assembly>.props naming convention.  By removing the TFM segment in the <PackagePath> property, the validation was fixed.  This makes sense because the nuget package is now packaging together targets for multiple frameworks in a single package rather than separate packages for each TFM.

I didn't find a good explanation of the `build/` folder and the other handful of related folders, though, so if there is a resource that explains which to use when, please let me know.  By moving the .props file to the higher level, it seems to have fixed the problem and the resulting packages seem correct on the surface.

Thanks so much for your quick feedback!  :)

src/OrchardCore.Build/Dependencies.AspNetCore.props
```diff
@@ -1,10 +1,32 @@
...
   <PropertyGroup>
-    <AspNetCoreVersion>3.1.11</AspNetCoreVersion>
-    <AspNetCoreTargetFramework>netcoreapp3.1</AspNetCoreTargetFramework>
+    <AspNetCoreTargetFrameworks>netcoreapp3.1;net5.0</AspNetCoreTargetFrameworks>
+    <TargetFrameworks>$(AspNetCoreTargetFrameworks)</TargetFrameworks>
   </PropertyGroup>
+  <Choose>
+    <When Condition=" '$(TargetFramework)' == 'net5.0' ">
+      <PropertyGroup>
+        <AspNetCoreVersion>5.0.1</AspNetCoreVersion>
```
sebastienros, line 9:
```suggestion
        <AspNetCoreVersion>5.0.3</AspNetCoreVersion>
```

src/OrchardCore.Build/Dependencies.AspNetCore.props
```diff
@@ -1,10 +1,32 @@
...
+      </PropertyGroup>
+    </When>
+    <Otherwise>
+      <!--
+      Fallback when not targetting .net 5.0.
+      At this time, that would mean .net core 3.1.  If there are more target frameworks in the future,
+      this logic should change slightly.
+      -->
+      <PropertyGroup>
+        <AspNetCoreVersion>3.1.11</AspNetCoreVersion>
```
sebastienros, line 19:
```suggestion
        <AspNetCoreVersion>3.1.12</AspNetCoreVersion>
```

src/OrchardCore.Build/Dependencies.AspNetCore.props
```diff
@@ -1,10 +1,32 @@
...
+      </PropertyGroup>
+    </Otherwise>
+  </Choose>
+
+  <ItemGroup>
+    <!-- Needs to stay at older package due to items being obsoleted -->
+    <PackageManagement Include="Microsoft.AspNetCore.Authentication.AzureAD.UI" Version="3.1.11" />
```
sebastienros, line 26: Can you give more details? We need to track this, and eventually bump the version.

src/OrchardCore.Build/Dependencies.AspNetCore.props
dafergu2: @sebastienros Thanks, I'll check that out.  I had specifically set it to 5.0.2 before and had some issues with packages not being found.  I'll try out 5.0.3. :)

src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
sebastienros: The Cms.Web project should have a single target. In that case net5.0. Same for the templates. Then if a user wants to target netcoreapp3.1 they can create their own project.

src/OrchardCore.Build/Dependencies.AspNetCore.props
sebastienros: I saw the comment on the PR description now. We still need to find a solution. Even if that means `#if` statements in the code.

dafergu2: @sebastienros The following warnings occur when using the latest package:

```
Build succeeded.

/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/Configuration/AzureADOptionsConfiguration.cs(18,32): warning CS0618: 'AzureADOptions' is obsolete: 'This is obsolete and will be removed in a future version. Use Microsoft.Identity.Web instead. See https://aka.ms/ms-identity-web.' [/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/OrchardCore.Microsoft.Authentication.csproj]
/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/Configuration/AzureADOptionsConfiguration.cs(53,44): warning CS0618: 'AzureADOptions' is obsolete: 'This is obsolete and will be removed in a future version. Use Microsoft.Identity.Web instead. See https://aka.ms/ms-identity-web.' [/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/OrchardCore.Microsoft.Authentication.csproj]
/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/Configuration/AzureADOptionsConfiguration.cs(74,31): warning CS0618: 'AzureADOptions' is obsolete: 'This is obsolete and will be removed in a future version. Use Microsoft.Identity.Web instead. See https://aka.ms/ms-identity-web.' [/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/OrchardCore.Microsoft.Authentication.csproj]
/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/Configuration/OpenIdConnectOptionsConfiguration.cs(12,42): warning CS0618: 'AzureADOptions' is obsolete: 'This is obsolete and will be removed in a future version. Use Microsoft.Identity.Web instead. See https://aka.ms/ms-identity-web.' [/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/OrchardCore.Microsoft.Authentication.csproj]
/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/Configuration/OpenIdConnectOptionsConfiguration.cs(15,66): warning CS0618: 'AzureADOptions' is obsolete: 'This is obsolete and will be removed in a future version. Use Microsoft.Identity.Web instead. See https://aka.ms/ms-identity-web.' [/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/OrchardCore.Microsoft.Authentication.csproj]
/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/Configuration/AzureADOptionsConfiguration.cs(40,31): warning CS0618: 'AzureADDefaults' is obsolete: 'This is obsolete and will be removed in a future version. Use Microsoft.Identity.Web instead. See https://aka.ms/ms-identity-web.' [/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/OrchardCore.Microsoft.Authentication.csproj]
/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/Configuration/AzureADOptionsConfiguration.cs(46,31): warning CS0618: 'AzureADDefaults' is obsolete: 'This is obsolete and will be removed in a future version. Use Microsoft.Identity.Web instead. See https://aka.ms/ms-identity-web.' [/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/OrchardCore.Microsoft.Authentication.csproj]
/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/Configuration/CookieOptionsConfiguration.cs(25,61): warning CS0618: 'AzureADDefaults' is obsolete: 'This is obsolete and will be removed in a future version. Use Microsoft.Identity.Web instead. See https://aka.ms/ms-identity-web.' [/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/OrchardCore.Microsoft.Authentication.csproj]
/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/Configuration/CookieOptionsConfiguration.cs(26,63): warning CS0618: 'AzureADDefaults' is obsolete: 'This is obsolete and will be removed in a future version. Use Microsoft.Identity.Web instead. See https://aka.ms/ms-identity-web.' [/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/OrchardCore.Microsoft.Authentication.csproj]
/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/Configuration/OpenIdConnectOptionsConfiguration.cs(23,25): warning CS0618: 'AzureADDefaults' is obsolete: 'This is obsolete and will be removed in a future version. Use Microsoft.Identity.Web instead. See https://aka.ms/ms-identity-web.' [/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/OrchardCore.Microsoft.Authentication.csproj]
/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/Configuration/OpenIdConnectOptionsConfiguration.cs(28,54): warning CS0618: 'AzureADDefaults' is obsolete: 'This is obsolete and will be removed in a future version. Use Microsoft.Identity.Web instead. See https://aka.ms/ms-identity-web.' [/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/OrchardCore.Microsoft.Authentication.csproj]
/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/Startup.cs(60,63): warning CS0618: 'AzureADOptions' is obsolete: 'This is obsolete and will be removed in a future version. Use Microsoft.Identity.Web instead. See https://aka.ms/ms-identity-web.' [/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/OrchardCore.Microsoft.Authentication.csproj]
/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/Configuration/AzureADOptionsConfiguration.cs(55,38): warning CS0618: 'AzureADDefaults' is obsolete: 'This is obsolete and will be removed in a future version. Use Microsoft.Identity.Web instead. See https://aka.ms/ms-identity-web.' [/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/OrchardCore.Microsoft.Authentication.csproj]
/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/Configuration/AzureADOptionsConfiguration.cs(78,38): warning CS0618: 'AzureADDefaults' is obsolete: 'This is obsolete and will be removed in a future version. Use Microsoft.Identity.Web instead. See https://aka.ms/ms-identity-web.' [/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/OrchardCore.Microsoft.Authentication.csproj]
/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/Configuration/AzureADOptionsConfiguration.cs(83,40): warning CS0618: 'AzureADDefaults' is obsolete: 'This is obsolete and will be removed in a future version. Use Microsoft.Identity.Web instead. See https://aka.ms/ms-identity-web.' [/Users/df/development/icagile/dafergu2/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/OrchardCore.Microsoft.Authentication.csproj]
    15 Warning(s)
```
I figured the safest route would be to simply use the older version rather than try and include an update to the Azure AD code for this PR.  Would it be reasonable to use the older version in this PR and create an issue to make sure this gets upgraded soon?  Or, would it be reasonable to use the newer version and disable the warnings for now?

src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
dafergu2: @sebastienros Okay, I'll make that change.  Same thing for the `OrchardCore.Mvc.Web.csproj` and the other web app or console apps in the tests folder.

src/OrchardCore.Build/Dependencies.AspNetCore.props
sebastienros: Instead please just add an ignore warnings pragma in the AzureADOptionsConfiguration file

dafergu2: @sebastienros I've added pragma warning disable/restores in the impacted files in the latest commit

src/OrchardCore.Build/Dependencies.AspNetCore.props
dafergu2: Done

src/OrchardCore.Build/Dependencies.AspNetCore.props
dafergu2: Done

src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
dafergu2: The Cms.Web and the Mvc.Web projects will now default to the `$(AspNetCoreTargetFramework)` prop that is defined in the `OrchardCore.Build\Dependencies.AspNetCore.props` file.  That prop is not set based on the multi-target passes made by MSBuild, but is defined as a string constant.

You may want to look at commit msg for 363324f07e5b8b13f1d672c4decb46015aab32e9 for a little detail about why the Cms.Web is conditionally setting the `<TargetFramework>` prop.  This is specifically because the `OrchardCore.Tests.csproj` references the Cms.Web project.  We want the `OrchardCore.Tests.csproj` to be fully multi-targeted since that will make `dotnet test` run the test suite for all of the targets in `<TargetFrameworks>` instead of having to manually do it for each target.  If you look at the change in the `OrchardCore.Tests.csproj`, you'll notice that the project reference item is specifying a `SetTargetFramework` metadata attribute.  This will force the correct TFM to be used when referencing from the `OrchardCore.Tests.csproj` file on both the netcoreapp3.1 and net5.0 targets, but allows the Cms.Web project to set it when it is empty and not referenced by the `OrchardCore.Tests.csproj`.

dafergu2: The reference by the `OrchardCore.Tests.csproj` is causing the `netcoreapp3.1` TFM target to be built for `Cms.Web` to be built, not the pattern described above, unfortunately.

The functional tests are failing due to them expecting the `netcoreapp3.1` binary of `OrchardCore.Mvc.Web` to be present.  I've updated the cypress et. al js scripts to  use the `net5.0`.  If both are needed, we can reference the `OrchardCore.Mvc.Web.csproj` in the functional test project and specify the target framework in the reference to force it to build.

dafergu2:
@sebastienros @jtkech @Skrypt I've updated the PR.  I wasn't sure if you the reviewer or author usually resolves conversation/comments, so I've left them open.  Let me know if there are any other desired changes.

src/OrchardCore.Build/Dependencies.AspNetCore.props
```diff
@@ -18,4 +38,16 @@
     <PackageManagement Include="Microsoft.Extensions.Caching.StackExchangeRedis" Version="$(AspNetCoreVersion)" />
     <PackageManagement Include="Microsoft.Extensions.Http.Polly" Version="$(AspNetCoreVersion)" />
   </ItemGroup>
+
+  <!-- For .NET 5.0 packages that do not have a version number that matches $(AspNetCoreVersion) available -->
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net5.0' ">
```
sebastienros, line 43: Create a `<MicrosoftExtensionsVersion>5.0.1</MicrosoftExtensionsVersion>` tag like AspNetCoreVersion instead.

dafergu2: Great call, I'll do that.

dafergu2: @sebastienros Done

src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
dafergu2: So to follow up again, after more testing (and with lots of cleaning out of obj/bin/temp folders between tests) and solution rebuilds, I believe the defaulting of Cms.Web and Mvc.Web doesn't need to the conditional, so I reworked those settings to be a normal setting.

sebastienros (approved):

sebastienros:
@jtkech do you approve this PR ?

jtkech:
@dafergu2 @sebastienros 

Sorry, didn't have enough time, also need to retrieve what i did when trying multi targetting, as i remember for the targets / props script files, aspnetcore puts the shared ones under a specific folder, and under each TFM folder there are script files that point to the shared ones, this in place of putting the shared ones under the parent folder, that's the same but it allows in a future version to point to different scripts, but maybe not worth in our case.

i will review it asap

src/OrchardCore/OrchardCore.Application.Cms.Core.Targets/OrchardCore.Application.Cms.Core.Targets.csproj
```diff
@@ -1,39 +1,38 @@
...
+
     Converts the application into a modular OrchardCore CMS application with TheAdmin theme but without any front-end Themes.</Description>
     <PackageTags>$(PackageTags) OrchardCoreCMS CMS</PackageTags>
   </PropertyGroup>
 
   <!-- Add ".props" and ".targets" files in the package specific to the "Cms.Core" bundle -->
   <ItemGroup>
     <None Include="OrchardCore.Application.Cms.Core.Targets.props" Pack="true">
-      <PackagePath>build\$(TargetFramework)\OrchardCore.Application.Cms.Core.Targets.props</PackagePath>
+      <PackagePath>build\OrchardCore.Application.Cms.Core.Targets.props</PackagePath>
```
jtkech, line 15: Here and in all other places below, can you try this nuspec convention

buildMultitargeting | (4.0) .props and .targets in the buildMultitargeting folder, for cross-framework targeting
-- | --

     <PackagePath>buildMultitargeting\OrchardCore.Application.Cms.Core.Targets.props</PackagePath>

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs
```diff
@@ -315,11 +315,10 @@ private string GetEncodedUrl(ShellSettings shellSettings, string token)
...
-            QueryString queryString;
-
+            var queryString = String.Empty;
             if (!String.IsNullOrEmpty(token))
             {
-                queryString = QueryString.Create("token", token);
+                queryString = QueryString.Create("token", token).ToString();
             }
 
             return $"{Request.Scheme}://{hostString + pathString + queryString}";
```
jtkech, line 324: I will revert this change as it was the right usage, building all uri components, an `HostString`, `PathString` and a `QueryString`, and adding them to build a string, so that their `ToString()`, that does the right encoding for each uri component, are implicitly called.

Yes, here it would be the same as the `ToString()` of the `QueryString` returns an empty string if applied on a null `QueryString`, but i prefer to keep the pattern of adding uri components.

 https://github.com/dotnet/aspnetcore/blob/c925f99cddac0df90ed0bc4a07ecda6b054a0b02/src/Http/Http.Abstractions/src/QueryString.cs#L53-L68

jtkech:
@dafergu2 

Good job! I did a little review and i will approve it.

But first can you try to use this nuspec convention?

buildMultitargeting | (4.0) .props and .targets in the buildMultitargeting folder, for cross-framework targeting
-- | --

So e.g. the following, and in other related places.

     <PackagePath>buildMultitargeting\OrchardCore.Application.Cms.Core.Targets.props</PackagePath>

src/OrchardCore/OrchardCore.Application.Cms.Core.Targets/OrchardCore.Application.Cms.Core.Targets.csproj
dafergu2: Sure thing.

src/OrchardCore/OrchardCore.Configuration.KeyVault/OrchardCore.Configuration.KeyVault.csproj
```diff
@@ -1,9 +1,5 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
-  <PropertyGroup>
-    <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework>
-  </PropertyGroup>
-
   <ItemGroup>
```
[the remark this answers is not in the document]
dafergu2: No, the `src/OrchardCore/Directory.Build.props` will automatically be picked up by msbuild.  It contains a project import for `src/OrchardCore.Build/OrchardCore.Commons.props` which contains all of the nuget properties.

src/OrchardCore.Build/Dependencies.AspNetCore.props
```diff
@@ -18,4 +41,11 @@
     <PackageManagement Include="Microsoft.Extensions.Caching.StackExchangeRedis" Version="$(AspNetCoreVersion)" />
     <PackageManagement Include="Microsoft.Extensions.Http.Polly" Version="$(AspNetCoreVersion)" />
   </ItemGroup>
+
+  <!-- Add all Microsoft.Extensions packages -->
+  <ItemGroup>
+    <PackageManagement Include="Microsoft.Extensions.Caching.StackExchangeRedis" Version="$(MicrosoftExtensionsVersion)" />
+    <PackageManagement Include="Microsoft.Extensions.Http.Polly" Version="$(MicrosoftExtensionsVersion)" />
```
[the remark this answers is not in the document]
dafergu2: Ah yes!  I believe I fell victim to the copy not cut when moving the code.  Thanks, I'll make that change.

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs
dafergu2: > @dafergu2
> 
> Good job! I did a little review and i will approve it.
> 
> But first can you try to use this nuspec convention?
> 
> buildMultitargeting	(4.0) .props and .targets in the buildMultitargeting folder, for cross-framework targeting
> So e.g. the following, and in other related places.
> 
> ```
>  <PackagePath>buildMultitargeting\OrchardCore.Application.Cms.Core.Targets.props</PackagePath>
> ```
> 
> https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files
> 
> **Update:** And according to the choice we will have to update the comment in the `Package.Build.props` file.

Thanks!  I'm working on testing the nupsec change now.  One question on process, there were a lot of comment change updates that seem to update style in separate commits.  Is that due to an extension/editorconfig to autocorrect?  If so, is there, can you point me to it so that I can make sure the styling fits expected guidelines?  A follow up question is, do you want me to collapse all of the related commits into a single commit to keep the history a little cleaner?  For example, all of the ones that update comments into a single commit, the ones for reverting the ApiController change back to using the struct all the way through in one commit?  Some projects want that to be done to make history neat, others don't, so just wanted to check before I made the changes.

dafergu2:
> @dafergu2
> 
> Good job! I did a little review and i will approve it.
> 
> But first can you try to use this nuspec convention?
> 
> buildMultitargeting	(4.0) .props and .targets in the buildMultitargeting folder, for cross-framework targeting
> So e.g. the following, and in other related places.
> 
> ```
>  <PackagePath>buildMultitargeting\OrchardCore.Application.Cms.Core.Targets.props</PackagePath>
> ```
> 
> https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files
> 
> **Update:** And according to the choice we will have to update the comment in the `Package.Build.props` file.

@jtkech Thanks! I'm working on testing the nupsec change now. One question on process, there were a lot of comment change updates that seem to update style in separate commits. Is that due to an extension/editorconfig to autocorrect? If so, is there, can you point me to it so that I can make sure the styling fits expected guidelines? A follow up question is, do you want me to collapse all of the related commits into a single commit to keep the history a little cleaner? For example, all of the ones that update comments into a single commit, the ones for reverting the ApiController change back to using the struct all the way through in one commit? Some projects want that to be done to make history neat, others don't, so just wanted to check before I made the changes.

sebastienros:
When the PR is merged it will be squashed, so no need to clean the history.

jtkech:
@dafergu2 

> @jtkech Thanks! I'm working on testing the nupsec change now

Cool, let me know your results

> One question on process, there were a lot of comment change updates that seem to update style in separate commits. Is that due to an extension/editorconfig to autocorrect? If so, is there, can you point me to it so that I can make sure the styling fits expected guidelines?

We have an `.editorconfig` at the root of the solution, but there is nothing about comments. I did the changes that you saw myself through the github UI, it was mainly to remove an extra space after a dot `.`, and adding a dot at the end of some lines, hmm and also adding some single quotes to `'SomeName'`

dafergu2:
@jtkech So, after testing with changing the folder name in the nuget packages for the .props and .targets files, I don't think that change is wanted.

Looking at https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package#include-msbuild-props-and-targets-in-a-package, we see the following.

> MSBuild .props and .targets files for cross-framework targeting can be placed in the \buildMultiTargeting folder. During package installation, NuGet adds the corresponding <Import> elements to the project file with the condition, that the target framework is not set (the MSBuild property $(TargetFramework) must be empty).

During my testing, the desired import did not make it into the `obj\MyModule.csproj.nuget.g.targets` or the `obj\MyModule.csproj.nuget.g.props` when using the `buildMultiTargeting` subfolder.

In most situations, the `$(TargetFramework)` var is going to be set when building a project based on one of the templates.  The `build` and `buildMultiTargeting` folders are simply making the referencing project have some props and targets imported in the generated nuget .targets and .props file in the `obj` folder.

Thinking out loud, I could see `$(TargetFramework)` not being set if someone were building a custom module that they themselves published as a nuget package.  The consumer of the package wouldn't need the extra .props and .targets file at that point, would they?

If we leave the .props and .targets files in the `build` folder but not under a specific TFM subfolder, then the imports happen as desired regardless of whether the consumer is netcoreapp3.1 or net5.0.  Wouldn't we want the props and targets file included regardless of TFM?  Right now, unless I missed something, all of those properties and tasks in those files are TFM agnostic, so I think that is the correct.

In the same doc as above, the build folder is described as:

> When NuGet installs a package with \build files, it adds MSBuild <Import> elements in the project file pointing to the .targets and .props files. (.props is added at the top of the project file; .targets is added at the bottom.) A separate conditional MSBuild <Import> element is added for each target framework.

It feels like the only reason to use `buildMultiTargeting` is to set props and targets only during the initial phase of the multitarget build and not during the subsequent repetitions that produce the actual assemblies for each TFM.  And only then when you know your package is consumer is going to be a multitarget package as well.  I haven't been able to find a really good explanation of the use case where the `buildMultiTargeting` would be preferred over just the `build` folder.

agriffard:
Please fix the build errors since the new Rules module and projects have been merged:

```
C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\..\src\OrchardCore.Modules\OrchardCore.Rules\OrchardCore.Rules.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [D:\a\OrchardCore\OrchardCore\test\OrchardCore.Tests\OrchardCore.Tests.csproj]
C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\..\src\OrchardCore\OrchardCore.Rules.Abstractions\OrchardCore.Rules.Abstractions.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [D:\a\OrchardCore\OrchardCore\test\OrchardCore.Tests\OrchardCore.Tests.csproj]
C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\..\OrchardCore\OrchardCore.Rules.Abstractions\OrchardCore.Rules.Abstractions.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [D:\a\OrchardCore\OrchardCore\src\OrchardCore.Modules\OrchardCore.Layers\OrchardCore.Layers.csproj]
C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\..\OrchardCore.Modules\OrchardCore.Rules\OrchardCore.Rules.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [D:\a\OrchardCore\OrchardCore\src\OrchardCore\OrchardCore.Application.Cms.Targets\OrchardCore.Application.Cms.Targets.csproj]
C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\OrchardCore.Rules.Abstractions\OrchardCore.Rules.Abstractions.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [D:\a\OrchardCore\OrchardCore\src\OrchardCore\OrchardCore.Application.Cms.Targets\OrchardCore.Application.Cms.Targets.csproj]
C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\..\..\src\OrchardCore.Modules\OrchardCore.Rules\OrchardCore.Rules.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [D:\a\OrchardCore\OrchardCore\test\OrchardCore.Tests.Pages\OrchardCore.Application.Pages\OrchardCore.Application.Pages.csproj]
C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\..\..\src\OrchardCore\OrchardCore.Rules.Abstractions\OrchardCore.Rules.Abstractions.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [D:\a\OrchardCore\OrchardCore\test\OrchardCore.Tests.Pages\OrchardCore.Application.Pages\OrchardCore.Application.Pages.csproj]
C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\..\OrchardCore.Modules\OrchardCore.Rules\OrchardCore.Rules.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [D:\a\OrchardCore\OrchardCore\src\OrchardCore\OrchardCore.Application.Cms.Core.Targets\OrchardCore.Application.Cms.Core.Targets.csproj]
C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\OrchardCore.Rules.Abstractions\OrchardCore.Rules.Abstractions.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [D:\a\OrchardCore\OrchardCore\src\OrchardCore\OrchardCore.Application.Cms.Core.Targets\OrchardCore.Application.Cms.Core.Targets.csproj]
```

dafergu2:
@jtkech Based on your feedback above, I'm adding and clarifying some comments in the .csproj/.props/.targets files.  I'll push that up shortly.  Feel free to reject if you feel they are not needed.

dafergu2:
> Please fix the build errors since the new Rules module and projects have been merged:
> 
> ```
> C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\..\src\OrchardCore.Modules\OrchardCore.Rules\OrchardCore.Rules.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [D:\a\OrchardCore\OrchardCore\test\OrchardCore.Tests\OrchardCore.Tests.csproj]
> C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\..\src\OrchardCore\OrchardCore.Rules.Abstractions\OrchardCore.Rules.Abstractions.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [D:\a\OrchardCore\OrchardCore\test\OrchardCore.Tests\OrchardCore.Tests.csproj]
> C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\..\OrchardCore\OrchardCore.Rules.Abstractions\OrchardCore.Rules.Abstractions.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [D:\a\OrchardCore\OrchardCore\src\OrchardCore.Modules\OrchardCore.Layers\OrchardCore.Layers.csproj]
> C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\..\OrchardCore.Modules\OrchardCore.Rules\OrchardCore.Rules.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [D:\a\OrchardCore\OrchardCore\src\OrchardCore\OrchardCore.Application.Cms.Targets\OrchardCore.Application.Cms.Targets.csproj]
> C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\OrchardCore.Rules.Abstractions\OrchardCore.Rules.Abstractions.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [D:\a\OrchardCore\OrchardCore\src\OrchardCore\OrchardCore.Application.Cms.Targets\OrchardCore.Application.Cms.Targets.csproj]
> C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\..\..\src\OrchardCore.Modules\OrchardCore.Rules\OrchardCore.Rules.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [D:\a\OrchardCore\OrchardCore\test\OrchardCore.Tests.Pages\OrchardCore.Application.Pages\OrchardCore.Application.Pages.csproj]
> C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\..\..\src\OrchardCore\OrchardCore.Rules.Abstractions\OrchardCore.Rules.Abstractions.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [D:\a\OrchardCore\OrchardCore\test\OrchardCore.Tests.Pages\OrchardCore.Application.Pages\OrchardCore.Application.Pages.csproj]
> C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\..\OrchardCore.Modules\OrchardCore.Rules\OrchardCore.Rules.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [D:\a\OrchardCore\OrchardCore\src\OrchardCore\OrchardCore.Application.Cms.Core.Targets\OrchardCore.Application.Cms.Core.Targets.csproj]
> C:\Users\runneradmin\AppData\Local\Microsoft\dotnet\sdk\5.0.102\Microsoft.Common.CurrentVersion.targets(1662,5): error : Project '..\OrchardCore.Rules.Abstractions\OrchardCore.Rules.Abstractions.csproj' targets 'net5.0'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v3.1'. [D:\a\OrchardCore\OrchardCore\src\OrchardCore\OrchardCore.Application.Cms.Core.Targets\OrchardCore.Application.Cms.Core.Targets.csproj]
> ```

Fixed and will be pushing shortly.  Thanks!

jtkech:
@dafergu2  your PR has been merged, cool !!!

> MSBuild .props and .targets files for cross-framework targeting can be placed in the \buildMultiTargeting folder. During package installation, NuGet adds the corresponding elements to the project file with the condition, that the target framework is not set (the MSBuild property $(TargetFramework) must be empty).

Good to know, thanks, so we would need to have them in each TFM folder, and also in the multi TFM one

But when TargetFrameworks is used, it is not intended that TargetFramework is also non empty.

Just tried to build the web app under Visual Studio, there are 3 builds, 2 for 5.0 and one for 3.1, we don't see the 2nd build of 5.0 as it has nothing to do, but we see 3 executions of the related targets file that copy translation files.

Seems to be because both TargetFramework and TargetFrameworks are both defined in a shared props file.

dafergu2:
> @dafergu2 your PR has been merged, cool !!!
> 
> > MSBuild .props and .targets files for cross-framework targeting can be placed in the \buildMultiTargeting folder. During package installation, NuGet adds the corresponding elements to the project file with the condition, that the target framework is not set (the MSBuild property $(TargetFramework) must be empty).
> 
> Good to know, thanks, so we would need to have them in each TFM folder, and also in the multi TFM one
> 
> But when TargetFrameworks is used, it is not intended that TargetFramework is also non empty.
> 
> Just tried to build the web app under Visual Studio, there are 3 builds, 2 for 5.0 and one for 3.1, we don't see the 2nd build of 5.0 as it has nothing to do, but we see 3 executions of the related targets file that copy translation files.
> 
> Seems to be because both TargetFramework and TargetFrameworks are both defined in a shared props file.

@jtkech I did a search on `<TargetFramework>` and it does not seem to be set in any shared build file.  It is likely that the build for the web app ran three times because the web app is defaulting to `net5.0` when the `$(TargetFramework)` is empty.  This was added so that when people clone the repo, they can simply do a `dotnet run` without specifying the target framework to run under.

image

When the mult-target builds happen, I believe it will run once with a `$(TargetFramework)` of '', and then once with each target framework set.  So, in the shared aspnet core .props file, we default the `$(AspNetCoreTargetFramework)` value to `net5.0.  Thus on the initial build of `$(TargetFramework == ''`, the web project will seem to have a TFM of `net5.0`.

```
    <!-- The default TFM for web app and console projects is net5.0 -->
    <AspNetCoreTargetFramework>net5.0</AspNetCoreTargetFramework>
```

That gets set in the OrchardCore.Cms.Web.csproj here (again, specifically to default to net5.0 so when running the web app or projects, you don't have to specify a TFM.

```
  <PropertyGroup>
    <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework>
    ...
  </PropertyGroup>
```

We could put in a conditional before that to track when the csproj is run with `$(TargetFramework0 == ''` and prevent other tasks from happening when it is built in that manner.  That way it does the multi-target build but the other tasks could execute only if the `$(TargetFramework) == ''` initially.  Like:

```
  <PropertyGroup>
    <IsAnyTFMBuild>false</IsAnyTFMBuild>
    <IsAnyTFMBuild Condition=" '$(TargetFramework)' == '' ">true</IsAnyTFMBuild>
    <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework>
    ...
  </PropertyGroup>
```
and then in `OrchardCore.Application.Cms.Core.Targets.targets` something like:

```
  <Target Name="CopyPackageTranslationFiles" AfterTargets="Build" Condition="!$(IsAnyTFMBuild)" >
    <Message Text="Copying translation files: $(MSBuildProjectName)" Importance="high" />
    ...
  </Target>
```

Another option I haven't tested, but we may be able to also use, is setting `<TargetFrameworks></TargetFrameworks>` to only build the web (and other projects) with `net5.0` if that made more sense.

```
  <PropertyGroup>
    <TargetFrameworks></TargetFrameworks>
    <TargetFramework>$(AspNetCoreTargetFramework)</TargetFramework>
    ...
  </PropertyGroup>
```

This disadvantage here is that the web and console apps will now only build with the default TFM of `net5.0` so CI/CD and unit testing may miss so TFM disparity issues with either compiling or runtime tests.

I think this is only going to happen when people run the web and console apps within a clone from the OrchardCore code and not others who use the nuget packages.

Thoughts?

sebastienros:
We can add custom build steps for 3.1 on the rest of the solution in the CI. If that helps make the dev box easier.

jtkech:
@dafergu2 

For me TargetFramework OR TargetFrameworks should be set in project files, not in the shared props file, and in a given project file only one of them should be set, this by using the properties defined in the shared props file.

Easy to do for the mvc and the web apps, for all other library projects faster to put it in the `Directory.Build.props` of their parent folder, I think this is why @sebastienros needed to specify the net5.0 framework in #8673 

Then maybe we could have kept the build/TFM package paths, if when building with a list of TargetFrameworks we let TargetFramework empty, and let it set by the system on each inner build, and / or use the buildMultiTargetting/

But not sure, and that's okay as it is if it works.

i will check it this week end, based on my results, maybe i will ask you to do another PR ;)

dafergu2:
@jtkech Yup, we could definitely segregate for the parent folder structure.  We would need to import the aspnetcore props after that though, so the dep chain would likely need to be changed slightly on where that shared aspnetcore props file is imported.

I do like the fact, though, that the web projects build for both frameworks specifically to cover `netcoreapp3.1` and `net5.0` build.  That way compile errors would be can be caught between TFMs.

The only real hiccup with only building `net5.0` was that the test project referenced the web project.  If we want tests to run for both, then that should be multi-targeted.  Or a multiple CI/CD builds get run that specifies the TFM to use for the test project by passing it into the build commands.

I'm good with any mods you want to make, honestly.  I've learned a lot about msbuild these past few weeks.  I'm always game to keep improving or doing things slightly different that fits in line with the team/process better.

Thanks for all the help @jtkech and @sebastienros as well as willingness to pull in suggested changes.  It is very much appreciated and refreshing.

jtkech:
No problem, thanks to you too ;)

> I do like the fact, though, that the web projects build for both frameworks specifically to cover netcoreapp3.1 and net5.0 build. That way compile errors would be can be caught between TFMs.

No problem if TargetFramework and TargetFrameworks (but not both) are only set in project files (or in the Directory.Build.props of their parent folder), e.g. if we will want to build the web app for both TFMs, we will just have to replace in its project file `TargetFramework = $(AspNetCoreTargetFramework)` with `TargetFrameworks = $(AspNetCoreTargetFrameworks)`

I will check that this w-end

jtkech:
Will try to also prevent these warnings while building multi TFMs

146>OrchardCore.Cms.Web -> C:\Sources\OrchardCore\src\OrchardCore.Cms.Web\bin\Debug\netcoreapp3.1\OrchardCore.Cms.Web.dll
146>OrchardCore.Cms.Web -> C:\Sources\OrchardCore\src\OrchardCore.Cms.Web\bin\Debug\net5.0\OrchardCore.Cms.Web.dll
146>Copying translation files: OrchardCore.Cms.Web
146>Copying translation files: OrchardCore.Cms.Web
146>C:\Sources\OrchardCore\src\OrchardCore\OrchardCore.Application.Cms.Core.Targets\OrchardCore.Application.Cms.Core.Targets.targets(26,5): **warning MSB3026: Could not copy** "C:\Users\JeanThierry\.nuget\packages\orchardcore.translations.zh-tw\1.0.0-rc2-108003\content\Localization\zh-TW\OrchardCore.Deployment.po" to "Localization\zh-TW\OrchardCore.Deployment.po". Beginning retry 1 in 1000ms. The process cannot access the file 'Localization\zh-TW\OrchardCore.Deployment.po' **because it is being used by another process.**

sebastienros:
I did a build on the fluid2 branch which got dev merged, and for me it was quite fast when doing a `dotnet build -c release` and didn't get translation files copied multiple times.

jtkech:
@sebastienros But from the cmd line there are still other little problems

- **Under Visual Studio**: When both TFM and TFMs are defined,  it starts a multi targeting outer build, that does 2 inner builds, so the target copying the translation files is executed 3 times, in fact when executed by the outer build they are not really copied for another reason (see at the end), so there are copied 2 times by the 2 inner builds.

    Here the conflict may happen when the files are first copied but concurrently by the 2 inner builds, after that's okay because we check if they have changed before copying them.

- **With the dotnet cmd line**: When both TFM and TFMs are defined, and without specifying a `--framework`, it does an outer build but that does only one inner build on the TFM, not 2 builds on all TFMs, then yes the target copying the files is executed once, but by the outer build, so, as said above, they are not really copied.

- The translation files are not really copied when the target is executed by the outer build, this is because a translation package enlist its files to be copied through a target in a `buildTransitive/` folder (so even if it is referenced transitively), but  the targets under `buildTransitive/` folders are not executed by an outer build.

So, in a given project, directly or through an import, we should define a TFM OR multi TFMs but not both, so that the right TFM(s) are always built, even if from the cmd line, and the target copying the translation files should be executed through a singular build, or through an inner build but only one, i'm working on this.

dafergu2:
> @sebastienros But from the cmd line there are still other little problems
> 
> * **Under Visual Studio**: When both TFM and TFMs are defined,  it starts a multi targeting outer build, that does 2 inner builds, so the target copying the translation files is executed 3 times, in fact when executed by the outer build they are not really copied for another reason (see at the end), so there are copied 2 times by the 2 inner builds.
>   Here the conflict may happen when the files are first copied but concurrently by the 2 inner builds, after that's okay because we check if they have changed before copying them.
> * **With the dotnet cmd line**: When both TFM and TFMs are defined, and without specifying a `--framework`, it does an outer build but that does only one inner build on the TFM, not 2 builds on all TFMs, then yes the target copying the files is executed once, but by the outer build, so, as said above, they are not really copied.
> * The translation files are not really copied when the target is executed by the outer build, this is because a translation package enlist its files to be copied through a target in a `buildTransitive/` folder (so even if it is referenced transitively), but  the targets under `buildTransitive/` folders are not executed by an outer build.
> 
> So, in a given project, directly or through an import, we should define a TFM OR multi TFMs but not both, so that the right TFM(s) are always built, even if from the cmd line, and the target copying the translation files should be executed through a singular build, or through an inner build but only one, i'm working on this.

I think the biggest problem with that is going to be the tests that reference the Cms.Web project.  Would you still want to run the tests for both TFMs?  If so, then the Cms.Web project would need to be built for both targets as well.

@jtkech I can work on modifying things if you want?  Wouldn't mind have a specific issue to track the exact requirements so that I could make sure I covered it all.

deanmarcussen:
@dafergu2 before I forget do you need to open an issue to resolve the obsolete warnings from OpenID / AzureAd so we can remember to get it migrated?

jtkech:
@dafergu2

Thanks, that's okay, only last tweaks to do.

The main thing i did is to be able to only have a TFM or TFMs but not both. Doing this now dotnet builds both TFMs when a project define it, which was not the case before (unless the dependencies), but was the case under VStudio.

Note: To not change all library projects i used the Directory.Build.props of their parent folders.

That said, there are opened issues about building concurrently, particularly when a multi targetable app references another app (so multi targetable), and when we do a dotnet build at the solution level, as we do through the CI. E.g. the Benchmarks project references the Tests project that references the Cms.Web project, we are crazy ;)

So we know build the solution for each TFM separately, but the dependencies are still built in parallel and only one restore happens, so we don't lose so much time compared to build the solution for both TFMs in parralel.

But doing this, now all projects need to be multi targetable. If, for some projects, we will want to support only one TFM, we will be able to define a TFM for a given project, and remove its TFMs, but then through the CI we will have to do separate dotnet build, from different folders, this in place of doing a global dotnet build at the solution level.

Idem, we may want to just run the unit tests on net5.0, will see

Just need to do some last tweaks

dafergu2:
> @dafergu2 before I forget do you need to open an issue to resolve the obsolete warnings from OpenID / AzureAd so we can remember to get it migrated?

@deanmarcussen It would probably be a good idea to add in an issue.  I didn't have plans to make those updates as I'm not familiar with that code at the moment.

PR #8552, opened 2021-02-09
Using ILocalClock instead of IClock (BackgroundTaskScheduler)

lampersky:
Using ILocalClock instead of IClock.

fixes https://github.com/OrchardCMS/OrchardCore/issues/7518

### DIFF
--- a/src/OrchardCore/OrchardCore/BackgroundTasks/BackgroundTaskScheduler.cs
+++ b/src/OrchardCore/OrchardCore/BackgroundTasks/BackgroundTaskScheduler.cs
@@ -6,15 +6,14 @@ namespace OrchardCore.BackgroundTasks
 {
     public class BackgroundTaskScheduler
     {
-        public BackgroundTaskScheduler(string tenant, string name, DateTime referenceTime, IClock clock, string timeZoneId)
+        public BackgroundTaskScheduler(string tenant, string name, DateTime referenceTime, ILocalClock localClock)
         {
             Name = name;
             Tenant = tenant;
             ReferenceTime = referenceTime;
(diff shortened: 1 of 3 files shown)


jtkech:
But `ConvertToLocalAsync()` is called synchronously, at least better to use `.GetResult().GetAwaiter()` in place of `.Result`, but better to call it asynchronously, when i will have time i will see if i can create and use a `CanRunAsync()`, and do a PR

        public bool CanRun()
        {
            ...
            var referenceTimeLocal = LocalClock.ConvertToLocalAsync(ReferenceTime).Result.DateTime;

Note: We may need to do the same in the WF `TimerEvent` activity

lampersky:
I've chooseen this way, because I wanted to avoid refactor all CanRun
references.

Those methods should also be changed to async when CanRunAsync is added.

GetSchedulersToRun -> GetSchedulersToRunAsync,
GetShellsToRun -> GetShellsToRunAsync

jtkech:
@lampersky 

Okay, i understand, no problem ;) I will check it when i will have time

For info, when we can't call asynchronously `.GetResult().GetAwaiter()` is better than `.Result`, e.g. to dispatch errors on exceptions

jtkech:
@lampersky 

No worries but i reverted your PR, but to still use your previous one. This because `ILocalClock` is a scoped service that can't be held by the scheduler components that are hold by the background service that is an app level singleton. Your previous PR uses `IClock` that doesn't take into account our time zone selectors, but at least it is an app level singleton, see #8564.

Will see later how we can use `IClock`, and maybe asyncronously ;)

PR #8563, opened 2021-02-10
date_utc Liquid filter 

### LINKED ISSUE #8562: Liquid : get utc now datetime
As reported here https://github.com/OrchardCMS/OrchardCore/discussions/8126
We can't get the UtcNow datetime from a liquid template.
We could get it by using the TimeZone filters to get the DateTime with it's offset and then remove the offset on the datetime.
Though, I think we should have a better way to handle this.

We should either be able to configure Fluid to return the UtcNow datetime when doing {{ 'now' }} and/or have a liquid filter that could return a Utc Datetime from a datetime that has a timezone. {{ datetime | date_utc }}.

Skrypt:
Fixes #8562 

WIP


Skrypt:
Not required as it is already part of Fluid.

Issue #8325, opened 2021-01-15
Admin UI : gulp-sass not working with Node 15


Skrypt:
Also, gulp-sass seems to be no longer up to date with latest SASS development.
We need to make the gulp pipeline work with at least latest Node 15 version.
There might be some changes required to make SASS compilation work with Node 15.
Need to investigate further and submit a working PR.

/cc @jptissot

deanmarcussen:
Recommended on the gulp-sass repository is to upgrade to gulp-dart-sass as node-sass is deprecated.

https://www.npmjs.com/package/gulp-dart-sass

Quickly tried it as latest node-sass doesn't compile on the latest nodejs anymore.

Worked fine to get some js out, just had to comment out a few things in the gulpfile related to the blog theme which also has a gulp-sass dependency.

Skrypt:
I tried it quickly last week but did not wanted to waste too much time on this as I was working on something else.
What you report is exactly what I found. Need to make a proper PR and see if there are some caveats on moving to Node 15 for other modules like GraphiQL or else.

PR #8570, opened 2021-02-10
Update from node-sass to dart-sass

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/8325

Because the pipeline won't work for me anymore on node 15...

replaces `gulp-sass` with it's newer `gulp-dart-sass` 

Would be great if someone on windows could try this branch and make sure it builds ok for them too.

Mostly we end up with formatting changes on the non minified files, but the dart compiler found an issue with the `switch.sass` file name, so fixed that as well.

/cc @Skrypt @jptissot @agriffard


src/OrchardCore.Themes/TheAdmin/Assets/scss/modules/admin/_switch.scss
```diff
@@ -20,7 +20,7 @@ $custom-control-indicator-disabled-bg: #e9ecef;
 $custom-control-description-disabled-color: #868e96;
 $white: white;
 $theme-colors: (
-    "$success": $success,
+    "success": $success,
```
deanmarcussen, line 23: @Skrypt this was the change I had to make to get it to compile.

when I look at the css we were generating, there were instances of
`.alert-$success` in `TheAdmin.css`

Those gone now

Skrypt: These switch are not maintained anymore. They are missing their javascript, I think it should probably be removed simply.

src/OrchardCore.Themes/TheComingSoonTheme/wwwroot/gulpfile.js
```diff
@@ -10,7 +10,7 @@ const header = require("gulp-header");
 const merge = require("merge-stream");
 const plumber = require("gulp-plumber");
 const rename = require("gulp-rename");
-const sass = require("gulp-sass");
+const sass = require("gulp-dart-sass");
```
Skrypt, line 13: These might get overwritten when @agriffard will update the theme.

src/OrchardCore.Themes/TheComingSoonTheme/wwwroot/package.json
```diff
@@ -42,7 +42,7 @@
         "gulp-header": "2.0.9",
         "gulp-plumber": "^1.2.1",
         "gulp-rename": "2.0.0",
-        "gulp-sass": "4.1.0",
+        "gulp-dart-sass": "1.0.2",
```
Skrypt, line 45: Same

Skrypt:
Still works with node 14.15.4

Skrypt:
Dark mode alert-success needs some adjustments. Strange one.

src/OrchardCore.Themes/TheComingSoonTheme/wwwroot/gulpfile.js
deanmarcussen: Yeah was thinking might just pr the source to upgrade it as well. To save that issue.

deanmarcussen:
ok reverted the theme changes, and removed building the themes from the primary gulpfile.

They really shouldn't have been there anyway, they're not part of the `assets.json` arrangement, and because they are external, are more dificult to control and manage.

To script the building of those for @agriffard I think they should just be part of the npm scripts(i.e. `npm run build-themes`), not part of the task runner.

Skrypt (approved):

PR #8580, opened 2021-02-11
Only one query in ContentAliasLiquidTemplateEventHandler

jtkech:


deanmarcussen:
Perhaps we should call `Loaded` - which `GetAsync` would do.

Shame we can't take advantage of the cache, unless there is a way to query it for an alias. Could be an option.

jtkech:
Oops, forgot this one, did it too quickly, will fix it

sebastienros (approved):

Issue #8423, opened 2021-01-25
Unauthorized ApiController request should not return 200 status code


HengzheLi:
### Describe the bug
I used PostMan to try ApiController. The requesting path is /api/content/[contentitemid] with an Authorization hearder whose value is only "Bearer"(no "token"). 
The response is empty and the status code is 200. 

### Reproduce steps
1. Create a fresh tenant using agency recipe.
2. Use postman to send a below request
```
"request": {
"auth": {
	"type": "bearer",
	"bearer": [
		{
			"key": "token",
			"value": "",
			"type": "string"
		}
	]
},
"method": "GET",
"header": [],
"url": {
	"raw": "https://localhost:5001/api/content/[replace to your content item id]",
	"protocol": "https",
	"host": [
		"localhost"
	],
	"port": "5001",
	"path": [
		"api",
		"content",
		"[replace to your content item id]"
	]
}
```

### Expected behavior
Unauthenticated request should get 401 return code.

I debugged the code and located to below code line in class OrchardCore.Mvc.Utilities.ControllerExtensions
```
controller.User.Identity.IsAuthenticated ? (ActionResult)controller.Forbid(authenticationSchemes) : (ActionResult)controller.Challenge(authenticationSchemes);
```
Seems like the code assumes that the request is sent by a login user. But actually, controller.User is null in this case as I didn't login when I sent the request. 

BTW, I take a look at the code of the class ApiAuthenticationHandler which is an authentication handler for schema "Api". It only check if there is a "Bearer" word in SchemaMap. 
```
protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
        {
            if (!_authenticationOptions.Value.SchemeMap.ContainsKey(Options.ApiAuthenticationScheme))
            {
                return AuthenticateResult.NoResult();
            }

            return await Context.AuthenticateAsync(Options.ApiAuthenticationScheme);
        }

        protected override Task HandleChallengeAsync(AuthenticationProperties properties)
        {
            if (!_authenticationOptions.Value.SchemeMap.ContainsKey(Options.ApiAuthenticationScheme))
            {
                return Task.CompletedTask;
            }

            return Context.ChallengeAsync(Options.ApiAuthenticationScheme);
        }
```
Seems like this behavior can't defend an unthenticated user. So what does this authencation handler use for? In my understand, shouldn't it validate if there is a valid token here?

sebastienros:
Need to be reproduced. There might still be a cookie sent. 

[...]

PR #8591, opened 2021-02-13
Fix status codes for api scheme

deanmarcussen:
Follow on from https://github.com/OrchardCMS/OrchardCore/pull/7055

@kevinchalet so `Forbid("Api")` didn't work to get status codes, instead of a 401 page back, because the `Diagnostics` module is using `app.UseStatusCodePagesWithReExecute("/Error/{0}");` and redirecting the status code result back to the Orchard Core error pages.

Might be other ways to fix this, but using this I get the correct responses from normal controllers, i.e. an error page, and a good header from an api request using the `Api` scheme, which has a `WWW-Authenticate` : `Bearer` and the 401.

Turning off the `Diagnostics` module works as well, but then you don't get the good error pages from non api controllers.

There is another issue on https://github.com/OrchardCMS/OrchardCore/issues/8423 but not the same, so I will comment there.


deanmarcussen:
I would have preferred to do this in the `Diagnostics` module, but as far as I can tell, there is no http feature, which can return the type of scheme specified by `Forbid("Api")`.

kevinchalet:
In theory, we could use MVC's built-in `[SkipStatusCodePages]` for that, but it's impacted by a bug that makes it run too late to be useful: https://github.com/dotnet/aspnetcore/issues/10317.

What you have should work fine, but if you're interested in opting for a more global and non-auth-specific option, you could copy MVC's filter and fix it to run before the authentication handlers 😃

deanmarcussen:
Hmm, so that would mean yet another attribute that has to be applied to api controllers... like the ignore antiforgery one that no one ever knows they need to apply.

I must say, I'm confused that applying the `ApiController` attribute doesn't produce this behaviour by default. Seems odd.

I think I'd rather go with this, and consider applying the mvc one if/when it is fixed

Issue #2148, opened 2018-07-27
Suggested addition: HTML META SEO


pgy866:
Suggested addition: HTML META
Similar to the Autoroute (OrchardCore.Autoroute) custom URL path feature,
For SEO, a module optimized for each page title may be a widget
At the same time, global website settings can be set to set the priority page settings globally.

Is there already a plan for this, I have not found it?
Is there a need for this module?
![_20180727223056](https://user-images.githubusercontent.com/28694676/43329271-8b4f6aa6-91f2-11e8-8e39-1f994b635659.png)
![_20180727231149](https://user-images.githubusercontent.com/28694676/43329273-8bce1478-91f2-11e8-9ded-18acf83203c2.png)

pgy866:
https://github.com/IDeliverable/IDeliverable.Seo
Find this in ORCHARD 1, orchard CORE does not know if it will be developed as a core

PR #8594, opened 2021-02-13
MetaTags module

agriffard:
Provides a way to specify Description and Keywords meta tags:

image

Related to #2148


hishamco:
This could be SEO related, it could be a feature of SEO, then we could add plenty of SEO services

agriffard:
Should the part be included by default on some content types in the recipes?

jptissot:
@agriffard What about localized Meta tags ? Is this going to be handled by dynamic localization ?

Skrypt:
I think that if you will localize a content item then it will have it's own Meta tab. Then the localized content item will contains it's own meta tag data.

hishamco:
>  Is this going to be handled by dynamic localization ?

I think this should be handled by dynamic localization, while it's changeable

agriffard:
Closing. A bit too simple compared to Seth and EtchUK SEO modules.

hishamco:
@agriffard if you don't mind we can add this to OCC.Modules repo

PR #4567, opened 2019-10-18
AutoSetup feature for Setup module

MatthijsKrempel:
In scenario's where you want to share your OrchardCore website with your friends, be it via code, docker or kubernetes, it is very frustrating to let them go through the setup proces.

All that is a thing of the past with this new AutoSetup feature!

You can now add a recipe that publishes your OrchardCore configuration on Startup!

AutoSetup will look for a valid OrchardCore configuration in appsettings or environment variables and provision your site. This way you wont have to instruct your coworkers or your clients.

PR #2824, opened 2018-12-08
Adding configuration support

sebastienros:
Fixes #2038
Fixes #1456
Fixes #825
Fixes #2051

PR #8596, opened 2021-02-14
Feature/auto setup v2

rserj:
The Second proposal of AutoSetup based on
https://github.com/OrchardCMS/OrchardCore/pull/4567
Some Key Notes
- Implemented as Separate Module which has to be referenced to Web project and Enabled like this
   `services.AddOrchardCms().AddSetupFeatures("OrchardCore.AutoSetup");`
- Implemented Auto Setup as Middleware
- Ability To start Setup by navigating to specified URL (AutoSetupOptions.TriggerSetupUrl ) (Might be useful for Kubernetes Init containers)
- Does not conflict with a Setup module, if AutoSetup configuration wasn't provided Setup screen will be displayed


Skrypt:
This feature always missing to be able to create multiple tenants. You can create one main tenant but can't add tenants inside that default tenant. Maybe it shouldn't, I don't know, but somehow if you want to automate the creation of OC tenants it would probably require to run recipes too on them afterward.

rserj:
@Skrypt I noticed there is a CreateTenantTask, Can the Setup recipe create a Workflow Task and setup another tenant(s) for now?

- Another option will be creating a SetupTenant Recipe Step, but it can be problematic since we have to wait until the current recipe has finished its execution, maybe somehow relying on ISetupEventHandler or Defered tasks/IRecipeStepQueue (in O1).
- In the future, we might generate Admin credentials AccessToken-Key/Session-Key so we can use the CLI tool to log in as Admin and manage tenants

rserj:
Added support for Multitenant setup

I do not know why the build is failed:
>AutoSetupMiddleware.cs(165,17): error CS0117: 'SetupContext' does not contain a definition for 'AdminEmail' 

I didn't touch SetupContext and it has a definition for AdminEmail and other properties

src/OrchardCore.Cms.Web/appsettings.json
```diff
@@ -79,6 +79,35 @@
...
+    //    "SiteTimeZone": "Europe/Amsterdam",
+    //    "AdminUsername": "Admin",
+    //    "AdminEmail": "info@orchardproject.net",
+    //    "AdminPassword": "OrchardCoreRules1!",
+    //    "DatabaseProvider": "Sqlite",
+    //    "DatabaseConnectionString": "",
+    //    "DatabaseTablePrefix": "",
+    //    "RecipeName": "Agency"
+    //  },
+    //  "SubTenants": [
```
deanmarcussen, line 96: I think this could just be one array, where you find the `Default` shell by it's name.

src/OrchardCore.Cms.Web/appsettings.json
```diff
@@ -79,6 +79,35 @@
...
+    //    {
+    //      "SiteName": "AutoSetup Sub Tenant",
+    //      "SiteTimeZone": "Europe/Amsterdam",
+    //      "AdminUsername": "tenantadmin",
+    //      "AdminEmail": "tenant@orchardproject.net",
+    //      "AdminPassword": "OrchardCoreRules1!",
```
deanmarcussen, line 102: I don't like the idea of putting the admin password in appsettings. 
It should be by environment variable, and here for demo purposes, at least a comment to that effect

deanmarcussen (commented):
A quick look as I see the build failed.

The reason for the build failure is this pr was merged https://github.com/OrchardCMS/OrchardCore/pull/7885

so the setup context has changed to use a properties dictionary.

It's interesting as a module, because then you might expect that you can continue to use it to setup tenants, once the default shell is running. 

But you wouldn't be able to do that I suspect?

src/OrchardCore.Cms.Web/appsettings.json
rserj: Right, I'm using Env. Variables too, I can remove example from appsettings.json and put them to launchSettings.json?

src/OrchardCore.Modules/OrchardCore.AutoSetup/Options/AutoSetupOptions.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// Gets or sets the Root tenant setup options.
+        /// </summary>
+        public BaseTenantSetupOptions RootTenant { get; set; }
+
+        /// <summary>
+        /// Gets or sets the Sub-tenants.
+        /// </summary>
+        public List<TenantSetupOptions> SubTenants { get; set; } = new List<TenantSetupOptions>();
+
+        public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
```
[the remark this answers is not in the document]
rserj: Thanks for your review
>Why you have made this an IValidatableObject ?

I think the main reason having Localized error messages

>The site will still fail to start / bind correctly here, when trying to configure the options, if a bad value is supplied.

The site should start fine even with incorrect values, the validation will fail in AutoSetupStartupFilter and Logged

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
```diff
@@ -0,0 +1,181 @@
...
+        /// <summary>
+        /// The middleware auto-setup invoke.
+        /// </summary>
+        /// <param name="httpContext">
+        /// The http context.
+        /// </param>
+        /// <returns>
+        /// The <see cref="Task"/>.
+        /// </returns>
+        public async Task InvokeAsync(HttpContext httpContext)
```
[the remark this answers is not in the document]
rserj: In this case Middleware run in DI's Shell Scope meaning I can not resolve scoped (request) services, I can not resolve ISetupService service

src/OrchardCore.Modules/OrchardCore.AutoSetup/Options/AutoSetupOptions.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// <summary>
+    /// The auto setup options.
+    /// </summary>
+    public class AutoSetupOptions : IValidatableObject
+    {
+        /// <summary>
+        /// Gets or sets the Url which will trigger AutoSetup.
+        /// Leave it Empty if you want to Trigger Setup on any request
+        /// </summary>
+        public string TriggerSetupUrl { get; set; }
```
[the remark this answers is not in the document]
rserj: Double setup "SetupSetup" confusing me, but I'm ok with it if others don't mind

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
```diff
@@ -0,0 +1,181 @@
...
+        /// </summary>
+        /// <param name="options"> The tenant setup options. </param>
+        /// <param name="setupService"> The setup service. </param>
+        /// <param name="shellSettings"> The tenant shell settings. </param>
+        /// <returns> The <see cref="SetupContext"/>. to setup the site </returns>
+        private static SetupContext GetSetupContext(BaseTenantSetupOptions options, ISetupService setupService, ShellSettings shellSettings)
+        {
+            var recipes = setupService.GetSetupRecipesAsync()
+                .GetAwaiter()
+                .GetResult();
```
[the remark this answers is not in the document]
rserj: thanks for noticing, fixed

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupStartupFilter.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        /// <summary>
+        /// The configure a middleware pipeline
+        /// </summary>
+        /// <param name="next"> The next middleware in the execution pipeline. </param>
+        /// <returns> The <see cref="Action"/>. </returns>
+        public Action<IApplicationBuilder> Configure(Action<IApplicationBuilder> next)
+        {
+            var scopedServices = ShellScope.Services;
```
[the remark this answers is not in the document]
rserj: I'm not sure, I saw it is being used in the other parts, @jtkech can you suggest, please?

src/OrchardCore.Modules/OrchardCore.AutoSetup/Options/AutoSetupOptions.cs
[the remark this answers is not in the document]
deanmarcussen: Hmm interesting, we don't actually localize logging.

> The site should start fine even with incorrect values,

so super easy to get startup exceptions with bad json. but ok, so the validation is about confirming that all the values have been completed.

src/OrchardCore.Modules/OrchardCore.AutoSetup/Options/AutoSetupOptions.cs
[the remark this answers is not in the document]
deanmarcussen: Typo:  suggest - `AutoSetupPath`

a. Trigger is weird
b. It's more a path than a url.

src/OrchardCore.Cms.Web/appsettings.json
rserj: I also thought about this, then decided to separate it explicitly since 
- Not everyone knows the Default naming convention
- The Default Tenant does not need `RequestUrlHost` and `RequestUrlPrefix` params

rserj:
>It's interesting as a module, because then you might expect that you can continue to use it to setup tenants, once the default >shell is running.

>But you wouldn't be able to do that I suspect?

Right, since you enabling it as a Setup feature `AddSetupFeatures("OrchardCore.AutoSetup")`, it will be turned off after the Default tenant gets installed.
I tried to create a Recipe step for creating a new Tenant, in this case, It might be a recursion problem. 
E.g if you added the NewTenant step into Default's Agency recipe and then try to use the same Agency recipe to install the new Sub-Tenant(s). In this case, we can fix recursion by checking Site Names OR do not allow use same recipes for Default and Sub tenants.

src/OrchardCore.Modules/OrchardCore.AutoSetup/Options/AutoSetupOptions.cs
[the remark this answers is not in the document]
rserj: Fixed

src/OrchardCore.Cms.Web/appsettings.json
rserj: Added Warning comment into appsettings.json

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupStartupFilter.cs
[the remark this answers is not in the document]
jtkech: @rserj @deanmarcussen 

Good question ;)

Yes it is used in some places, but because it is less DI firendly we could say it is better to use the DI when it is possible, we use it e.g. for `ISiteService` that is a tenant singleton for some reasons but that needs to be executed in a tenant / shell scope context, this to resolve scoped services.

Here it's the same kind of thing, yes `IStartupFilter` are executed under a shell scope context while building the tenant pipeline, but they are resolved through the tenant / shell container, not through a shell scope, we could have done it but we opted to keep for a tenant the pattern of `IStartupFilter` as for a regular "app" (here the tenant as an app with its own shell container).

            // Create a nested pipeline to configure the tenant middleware pipeline
            var startupFilters = appBuilder.ApplicationServices.GetService<IEnumerable<IStartupFilter>>();


So here, if you inject `IServiceProvider` it would be related to the tenant / shell container, not the one of the current shell scope, so you would not be able to resolve scoped services. So yes, using `ShellScope.Services;` is a solution, the other solution is to inject `IHttpContextAccessor` to get the `RequestServices` that we take care to be the services of the shell scope if there is a current one ;)

Both are using an `AsyncLocal` object to be retrieved in a given async execution flow, the current `HttpContext` or current `ShellScope`. Personnaly i would use `ShellScope`, but `IHttpContextAccessor` is more aspnetcore friendly.

Hmm, maybe another solution through a module startup `Configure()` that passes a scoped service provider, and with a lower `ConfigureOrder` to get called earlier. Hmm, i could get an auto setup working by only using the config sources stack + a few lines of code https://github.com/OrchardCMS/OrchardCore/pull/4567#issuecomment-544059139, but i don't remember what was missing.

i would need to take a deeper look on your PR, i will do soon when i will have time.

src/OrchardCore.Cms.Web/appsettings.json
sebastienros: I think it should use the default sections, including for other tenants.
I also think it should only initialize the tenant the first time it's requested, and not on the first request of the default one.

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
```diff
@@ -0,0 +1,182 @@
...
+            {
+                foreach (var tenant in _setupOptions.Value.SubTenants)
+                {
+                    if(!await SetupTenant(setupService, tenant, CreateTenantSettings(tenant)))
+                    {
+                        break;
+                    }
+                }
+
+                httpContext.Response.Redirect("/");
```
sebastienros, line 89: ~/ so it also works with the prefix of the default tenant

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
```diff
@@ -0,0 +1,182 @@
...
+            {
+                _logger.LogInformation($"AutoSetup successfully provisioned the site {setupOptions.SiteName}");
+
+                return true;
+            }
+
+            var stringBuilder = new StringBuilder();
+            foreach (var error in setupContext.Errors)
+            {
+                stringBuilder.Append($"{error.Key} : '{error.Value}'");
```
sebastienros, line 120: AppendLine

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
```diff
@@ -0,0 +1,182 @@
...
+            }
+
+            _logger.LogError($"AutoSetup failed installing the site '{setupOptions.SiteName}' with errors: {stringBuilder}");
```
sebastienros, line 123:
```suggestion
            _logger.LogError("AutoSetup failed installing the site '{SiteName}' with errors: {Errors}", setupOptions.SiteName, stingBuilder);
```

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
```diff
@@ -0,0 +1,182 @@
...
+
+            shellSettings.Name = setupOptions.SiteName;
+            shellSettings.RequestUrlHost = setupOptions.RequestUrlHost;
+            shellSettings.RequestUrlPrefix = setupOptions.RequestUrlPrefix;
+            shellSettings.State = TenantState.Uninitialized;
+
+            shellSettings["ConnectionString"] = setupOptions.DatabaseConnectionString;
+            shellSettings["TablePrefix"] = setupOptions.DatabaseTablePrefix;
+            shellSettings["DatabaseProvider"] = setupOptions.DatabaseProvider;
+            shellSettings["Secret"] = Guid.NewGuid().ToString();
```
sebastienros, line 144: What is that?

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupStartupFilter.cs
```diff
@@ -0,0 +1,88 @@
...
+
+                    if (validationErrors.Any())
+                    {
+                        var stringBuilder = new StringBuilder();
+                        foreach (var error in validationErrors)
+                        {
+                            stringBuilder.Append(error.ErrorMessage);
+                        }
+
+                        _logger.LogError("AutoSetup did not start, configuration has following errors: {errors}", stringBuilder.ToString());
```
sebastienros, line 64:
```suggestion
                        _logger.LogError("AutoSetup did not start, configuration has following errors: {Errors}", stringBuilder.ToString());
```

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupStartupFilter.cs
```diff
@@ -0,0 +1,88 @@
...
+using Microsoft.Extensions.Options;
+using OrchardCore.AutoSetup.Options;
+using OrchardCore.Environment.Shell.Scope;
+
+namespace OrchardCore.AutoSetup
+{
+    /// <summary>
+    /// The filter which registers <see cref="AutoSetupMiddleware"/> to setup the site.
+    /// </summary>
+    public class AutoSetupStartupFilter : IStartupFilter
```
sebastienros, line 18: Is it necessary since you could register the middleware from Configure in the same Startup class that registers this filter?

sebastienros:
I'd like to see if we can pass the arguments from the json file directly on the command line (since they are read the same way as from the ENV).

rserj:
> I'd like to see if we can pass the arguments from the json file directly on the command line (since they are read the same way as from the ENV).

I tested it via env variables set in the launchSettings.json file in the Web project and the Docker container.
Env variables look like this

 Default Tenant
 "OrchardCore__OrchardCore_AutoSetup__RootTenant__SiteName": "AutoSetup Example"
 "OrchardCore__OrchardCore_AutoSetup__RootTenant__RecipeName": "Agency"
 "OrchardCore__OrchardCore_AutoSetup__RootTenant__SiteTimeZone": "America/California"
 "OrchardCore__OrchardCore_AutoSetup__RootTenant__AdminUsername": "admin"
 "OrchardCore__OrchardCore_AutoSetup__RootTenant__AdminEmail": "my.admin@example.com"
 "OrchardCore__OrchardCore_AutoSetup__RootTenant__AdminPassword": "p@$Sw0rd"
 "OrchardCore__OrchardCore_AutoSetup__RootTenant__DatabaseProvider": "Sqlite"
 "OrchardCore__OrchardCore_AutoSetup__TriggerSetupUrl": ""
 
Single Sub Tenants
"OrchardCore__OrchardCore_AutoSetup__SubTenants__0__SiteName" : "Tenant Site"
"OrchardCore__OrchardCore_AutoSetup__SubTenants__0__RecipeName": "Agency"
"OrchardCore__OrchardCore_AutoSetup__SubTenants__0__SiteTimeZone": "America/California"
"OrchardCore__OrchardCore_AutoSetup__SubTenants__0__AdminUsername": "tenant_admin"
"OrchardCore__OrchardCore_AutoSetup__SubTenants__0__AdminEmail" : "tenant.admin@example.com"
"OrchardCore__OrchardCore_AutoSetup__SubTenants__0__AdminPassword": "p@$Sw0rd"
"OrchardCore__OrchardCore_AutoSetup__SubTenants__0__DatabaseTablePrefix": "tenant"
"OrchardCore__OrchardCore_AutoSetup__SubTenants__0__DatabaseProvider": "Sqlite"
"OrchardCore__OrchardCore_AutoSetup__SubTenants__0__DatabaseConnectionString": ""
"OrchardCore__OrchardCore_AutoSetup__SubTenants__0__RequestUrlHost": ""
"OrchardCore__OrchardCore_AutoSetup__SubTenants__0__RequestUrlPrefix": "tenant"

the 0 - Is an array index

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupStartupFilter.cs
rserj: Thanks for review
I thought about this, I think I could do this by using middleware in the Startup. I guess the main difference is that ASP.NET Core uses IStartupFilters in order to build the application pipeline, by using IStartupFilter I can guarantee that AutosetupMiddleware will be the first in the pipeline. Please correct me if I'm wrong.

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
rserj: I have to create a tenant shell before I setup one. (I do not have to do it for a Default tenant since shell gets created on the first start.)
I saw it in the `OrchardCore.Tenants.ApiController.Create` Action. I think for better reusability we should create an abstraction to manage tenants, but not in this PR

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
rserj: Fixed

src/OrchardCore.Cms.Web/appsettings.json
rserj: @sebastienros sorry it is not clear to me, can you please rephrase your concern and/or give me an example?

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
rserj: Fixed

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
rserj: Fixed

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
rserj: >shellSettings["Secret"] = Guid.NewGuid().ToString();

I do not know, it is from OrchardCore.Tenants.ApiController.Create

Commit d918b13c
Adding configuration support (#2824)

jtkech: Yes, i think that @sebastienros was only talking about the `Secret`, it is used to create an encrypted token that is appended to the url to be able to setup a tenant that was created and still in the `Uninitialized` state, e.g. through the admin tenant page, the secret is generated when we create a tenant, and then re-used and encrypted for security to provide the admin link to setup the tenant.

So here that's okay to generate a secret, as in the api controller, that you will not use as you create the tenant and then setup it yourself, bypassing any setup controller, so you don't have to generate a protected url using a dataProtector.

So the secret is used between the creation and the setup of a given tenant, so here for me that's okay, unless we would want a security token for the url that will trigger an auto setup = create + setup tenant(s) ;)

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        /// <summary>
+        /// Get setup context from the configuration.
+        /// </summary>
+        /// <param name="options"> The tenant setup options. </param>
+        /// <param name="setupService"> The setup service. </param>
+        /// <param name="shellSettings"> The tenant shell settings. </param>
+        /// <returns> The <see cref="SetupContext"/>. to setup the site </returns>
+        private static async Task<SetupContext> GetSetupContext(BaseTenantSetupOptions options, ISetupService setupService, ShellSettings shellSettings)
```
jtkech, line 157: `GetSetupContextAsync`

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupStartupFilter.cs
jtkech: Yes, your startup filter will run first (app level ones still run first) as we call the startup filters of a given tenant (using its own app builder) before building its pipeline (so calling the modules startup `Configure()`).

But for testing you can try to use a startup class, the one related to your auto setup feature, and override the `ConfigureOrder` property with a negative value, e.g. -100

But just to test it because somewhere i like the idea of using a startup filter, the only thing is that the startup `Configure()` provides the shell scope service provider, as commented in another comment, but for me not an issue by retrieving it with any other way

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// <summary>
+        /// Setup tenant.
+        /// </summary>
+        /// <param name="setupService"> The setup service. </param>
+        /// <param name="setupOptions"> The tenant setup options. </param>
+        /// <param name="shellSettings"> The tenant shell settings. </param>
+        /// <returns>
+        /// The <see cref="Task"/>.
+        /// </returns>
+        public async Task<bool> SetupTenant(ISetupService setupService, BaseTenantSetupOptions setupOptions, ShellSettings shellSettings)
```
jtkech, line 104: `SetupTenantAsync()`

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// <param name="httpContext">
+        /// The http context.
+        /// </param>
+        /// <returns>
+        /// The <see cref="Task"/>.
+        /// </returns>
+        public async Task InvokeAsync(HttpContext httpContext)
+        {
+            var currentShellSettings = ShellScope.Context.Settings;
+            var setupService = httpContext.RequestServices.GetRequiredService<ISetupService>();
```
jtkech, line 76: Just to say that here we use the `AsyncLocal ShellScope` to retrieve the settings, but not to resolve the service ;)

For me that's okay to use it or not for both, so just for infos the other way would be

           var currentShellSettings = httpContext.RequestServices.GetRequiredService<ShellSettings>();

jtkech:
@rserj

Just did a little review, will try to complete it asap, maybe this night.

src/OrchardCore.Modules/OrchardCore.AutoSetup/Manifest.cs
```diff
@@ -0,0 +1,11 @@
+using OrchardCore.Modules.Manifest;
+
+[assembly: Module(
+    Name = "Auto Setup",
+    Author = ManifestConstants.OrchardCoreTeam,
+    Website = ManifestConstants.OrchardCoreWebsite,
+    Version = ManifestConstants.OrchardCoreVersion,
+    Description = "The auto setup module allows automatically install the application on the first start",
```
Skrypt, line 8:
```suggestion
    Description = "The auto setup module allows to automatically install the application / tenants",
```

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
rserj: Reverting it back, when I use redirection to ~/ it redirects me to https://localhost:5001/~/ 404

rserj:
@jtkech @Skrypt  Thanks for your reviews, I've applied your suggestions of renaming and description change.
I also got rid of the AutoSetupFilter, and change the resolution of the ShellSettings

rserj:
AutoSetupDemo
short demo

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupStartupFilter.cs
rserj: Thanks for the reviews, I replaced AutoSetupStartupFilter  as a middleware in Statup.cs

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
jtkech: @rserj 

Yes, `~/` works when using Redirect (returning a RedirectResult) from a controller, because when executing the result the url helper.Content() is called and it uses the pathBase.

The pathBase contains an eventual virtual folder, and at this point it already contains the tenant prefix, the default tenant may have one prefix, so try the following

    httpContext.Response.Redirect(httpContext.Request.PathBase)

jtkech:
@rserj

Yes, `~/` works when using Redirect (returning a RedirectResult) from a controller, because when executing the result the url helper.Content() is called and it uses the pathBase.

The pathBase contains an eventual virtual folder, and at this point it already contains the tenant prefix, the default tenant may have one prefix, so try the following

    httpContext.Response.Redirect(httpContext.Request.PathBase)

jtkech:
@rserj 

See my above comment, and here some other thoughts, not sure it would be better but i let you try the following suggestions if you want, you will be able to test them faster than me ;)

Yes the `Default` tenant is a special one as it has a fixed name, it doesn't need an url prefix but it may have one, and it always exists, e.g. with a minimum shell descriptor before a setup, otherwise it is a tenant like the other ones.

- So first, in the auto setup configuration, in place of using `RootTenant` and `SubTenants` i would have used only one array named `Tenants`, also including the `Default` tenant (hmm, it would have to include its shell settings Name).


- Then why not using the configuration stack that already exists to fully configure all tenants, by specifying for each tenant a section with its name and containing all the needed values. Here you will need to specify a `State` property equals to `Unitialized` so that the sections are recognized as tenant sections. Doing this, the shell settings manager will harverst all tenants, and automatically create them and put them in an `Uninitialized` state. See https://github.com/OrchardCMS/OrchardCore/pull/4567#issuecomment-544059139

    The auto setup config would only contain `AutoSetupPath`, and if the feature is enabled, for a given tenant you will be able to create a scope on this tenant, bind through the shell configuration the provided values to a setup context, and then do an auto setup on it.


I understand what you did, so not sure the above is better, the idea being to re-use what already exists to pre-configure tenants, so maybe worth to try it.

rserj:
> * So first, in the auto setup configuration, in place of using `RootTenant` and `SubTenants` i would have used only one array named `Tenants`, also including the `Default` tenant (hmm, it would have to include its shell settings Name).

Thanks for your review, I've combined `RootTenant` and `SubTenants` into one `Tenants` collection, but instead of 'ShellName' and 'State' attributes I added IsDefault flag

src/OrchardCore.Cms.Web/appsettings.json
sebastienros: Not sure which part you didn't understand. What I mean is that last time I saw the code, the tenants were all initialized when the application is starting. It would be better if a tenant was auto-initialized on the first request to it. So if you open the default tenant, it's initialized, then if you open tenant1 it get initialized.

sebastienros:
Something in the documentation to make it discoverable, with an example of settings and also mention how to use it with environment variables.

sebastienros:
Also update the csproj since we merged the net5.0 support by default.

src/OrchardCore.Cms.Web/appsettings.json
rserj: I've implemented on-demand initialization please take a look

rserj:
@sebastienros since your last review I added a couple of changes as you asked
- Implemented On-Demand Tenant Setup
- merged dev, fixed project to .net core 5 multi-targeting
- Added documentation

src/OrchardCore.Modules/OrchardCore.AutoSetup/Options/TenantSetupOptions.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        /// <summary>
+        /// Gets or sets the tenant request url prefix.
+        /// </summary>
+        public string RequestUrlPrefix { get; set; }
+
+        /// <summary>
+        /// Gets the Flag which indicates a Default/Root shell/tenant.
+        /// </summary>
+        public bool IsDefault => string.Equals(ShellName, DefaultShellName, StringComparison.InvariantCultureIgnoreCase);
```
jtkech, line 84: using OrchardCore.Environment.Shell;

         string.Equals(ShellName, ShellHelper.DefaultShellName, StringComparison.OrdinalIgnoreCase);

src/OrchardCore.Modules/OrchardCore.AutoSetup/Options/TenantSetupOptions.cs
```diff
@@ -0,0 +1,147 @@
...
+
+            if (String.IsNullOrEmpty(ShellName) || !Regex.IsMatch(ShellName, @"^\w+$"))
+            {
+                yield return new ValidationResult(T["Invalid shell name. Can not be empty and must contain characters only and no spaces.", "Shell name"], new[] { nameof(ShellName) });
+            }
+
+            if (!IsDefault && string.IsNullOrWhiteSpace(RequestUrlPrefix) && string.IsNullOrWhiteSpace(RequestUrlHost))
+            {
+                yield return new ValidationResult(T["For no Default Tenant RequestUrlPrefix or RequestUrlHost should be provided", "Tenant Url"], new[] { "TenantUrl" });
+            }
```
jtkech, line 104: I didn't check all properties, only the `RequestUrlPrefix`. So here maybe good to add the following check as it is done in the Tenants controller and api controller, but anyway not in the WF CreateTenantTask, so as you want ;)

Note: Validations would need to be consistent everywhere, maybe in a separate PR.


            if (!String.IsNullOrWhiteSpace(shellSettings.RequestUrlPrefix))
            {
                if (shellSettings.RequestUrlPrefix.Contains('/'))
                {
                    ModelState.AddModelError(nameof(CreateApiViewModel.RequestUrlPrefix), S["The url prefix can not contain more than one segment."]);
                }
            }

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
```diff
@@ -0,0 +1,212 @@
...
+                            {
+                                if (setupTenant != tenant)
+                                {
+                                    await CreateTenantSettingsAsync(tenant);
+                                }
+                            }
+                        }
+                    }
+
+                    var redirectUrl = "/";
```
jtkech, line 103: Did you try my suggestion? The pathBase contains an eventual virtual folder, and at this point it already contains the tenant prefix (the default tenant may have one prefix).

    httpContext.Response.Redirect(httpContext.Request.PathBase)

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
```diff
@@ -0,0 +1,212 @@
...
+                    {
+                        redirectUrl = setupTenant.RequestUrlHost;
+                    }
+                    else if (!string.IsNullOrEmpty(setupTenant.RequestUrlPrefix))
+                    {
+                        redirectUrl = $"/{setupTenant.RequestUrlPrefix}";
+                    }
+
+                    httpContext.Response.Redirect(redirectUrl);
+                    await httpContext.Response.CompleteAsync();
```
jtkech, line 114: `CompleteAsync()`: Is it necessary?

src/OrchardCore.Modules/OrchardCore.AutoSetup/Startup.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// <param name="serviceProvider">
+        /// The "Shell Scope" service provider.
+        /// </param>
+        public override void Configure(IApplicationBuilder app, IEndpointRouteBuilder routes, IServiceProvider serviceProvider)
+        {
+            var currentShellSettings = serviceProvider.GetRequiredService<ShellSettings>();
+            var logger = serviceProvider.GetRequiredService<ILogger<Startup>>();
+
+
+            if (currentShellSettings.State == Environment.Shell.Models.TenantState.Uninitialized)
```
jtkech, line 73: Add an using so that you can just use `TenantState.Uninitialized`

src/OrchardCore.Modules/OrchardCore.AutoSetup/Startup.cs
```diff
@@ -0,0 +1,110 @@
...
+                        }
+                        else
+                        {
+                            app.Map(options.AutoSetupPath, appBuilder => appBuilder.UseMiddleware<AutoSetupMiddleware>());
+                        }
+                    }
+                }
+            }
+
+            base.Configure(app, routes, serviceProvider);
```
jtkech, line 107: You can remove this line

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
```diff
@@ -0,0 +1,212 @@
...
+        /// <param name="shellHost"></param>
+        /// <param name="logger"> The logger. </param>
+        public AutoSetupMiddleware(
+            RequestDelegate next,
+            IShellSettingsManager shellSettingsManager,
+            IOptions<AutoSetupOptions> setupOptions,
+            IShellHost shellHost,
+            ILogger<AutoSetupMiddleware> logger)
+        {
+            _logger = logger;
```
jtkech, line 60: Maybe put the logger at the end to keep the same order

jtkech:
@rserj

One of my concerns was to use or not pre-configured tenants, we can already do that by defining tenant sections, which are recognized as such if they contains a `"State": "Uninitialized"` field. In that case, when you start the app all pre-configured tenants are created automatically by the shell settings manager. And you can have global values (just under the OrchardCore section) that will be shared by all tenants.

But also makes sense to have a separate section providing data that will only be used if the auto setup feaure is used, and that will override any pre-configured value, as done through the Tenants api controller. Note: But not for all properties in the Tenants mvc controller, e.g. if the database provider is already pre-configured, it is not shown on the setup screen.

So, i'm not against and I will approve, but first i just did another little review.

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
rserj: >Did you try my suggestion?

Yes, as you mentioned below context.Request.PathBase can be empty in this case I have to redirect to /.  
Plus since we have `AutoSetupOptions.AutoSetupPath` option the Request Url can be different from Tenant Url in this case I have to redirect to Tenant's Url

src/OrchardCore.Modules/OrchardCore.AutoSetup/Options/TenantSetupOptions.cs
```diff
@@ -0,0 +1,147 @@
...
+
+    /// <summary>
+    /// The tenant setup options.
+    /// </summary>
+    public class TenantSetupOptions
+    {
+        /// <summary>
+        /// The default/root shell name.
+        /// </summary>
+        private const string DefaultShellName = "Default";
```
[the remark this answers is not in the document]
rserj: fixed

src/OrchardCore.Modules/OrchardCore.AutoSetup/Options/TenantSetupOptions.cs
rserj: fixed

src/OrchardCore.Modules/OrchardCore.AutoSetup/Startup.cs
rserj: fixed

src/OrchardCore.Modules/OrchardCore.AutoSetup/Startup.cs
rserj: fixed

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
rserj: fixed

rserj:
> One of my concerns was to use or not pre-configured tenants, we can already do that by defining tenant sections, which are recognized as such if they contains a "State": "Uninitialized" field

Do You mean defining the tenants section in App_Data tenants.json file? Cause Autosetup feature will make sure and create other Tenants only if the Default tenant has successfully installed.

src/OrchardCore.Modules/OrchardCore.AutoSetup/Startup.cs
```diff
@@ -0,0 +1,109 @@
...
+
+                    if (validationErrors.Any())
+                    {
+                        var stringBuilder = new StringBuilder();
+                        foreach (var error in validationErrors)
+                        {
+                            stringBuilder.Append(error.ErrorMessage);
+                        }
+
+                        logger.LogError("AutoSetup did not start, configuration has following errors: {errors}", stringBuilder.ToString());
```
deanmarcussen, line 92: Here, you are using localized strings for the validation, but then recording the validation errors in the log file, so it doesn't make sense that they are localized. Because log files don't contain localization.

If it was going to the front end, it would make sense for them to be localized, but when they are being appended to the log file, then it makes less sense. Because none of the ASP.Net errors, are localized, nor is the actual error message, of `AutoSetup did not start`.

So having `AutoSetup did not start : {errors in another language}` doesn't quite add up to me?

src/OrchardCore.Modules/OrchardCore.AutoSetup/Startup.cs
```diff
@@ -0,0 +1,109 @@
...
+        /// </param>
+        public override void Configure(IApplicationBuilder app, IEndpointRouteBuilder routes, IServiceProvider serviceProvider)
+        {
+            var currentShellSettings = serviceProvider.GetRequiredService<ShellSettings>();
+            var logger = serviceProvider.GetRequiredService<ILogger<Startup>>();
+
+
+            if (currentShellSettings.State == TenantState.Uninitialized)
+            {
+                var optionsAccessor = serviceProvider.GetRequiredService<IOptions<AutoSetupOptions>>();
```
deanmarcussen, line 76:
```suggestion
                var options = serviceProvider.GetRequiredService<IOptions<AutoSetupOptions>>().Value;
```

src/OrchardCore.Modules/OrchardCore.AutoSetup/Startup.cs
rserj: thanks for your review, I agree with you, I've removed localization

src/OrchardCore.Modules/OrchardCore.AutoSetup/Startup.cs
rserj: Fixed

jtkech:
@rserj 

> Do You mean defining the tenants section in App_Data tenants.json file? Cause this Autosetup feature will make sure and create other Tenants only if the Default tenant has successfully installed.

No, in any regular config source as the `appsettings.json` at the root of the app. `tenants.json` is a kind of config source but that is more intended to be mutated when the setup is completed, it holds some required settings (not all) and has an higher precedence so that a pre-configured value can be overridden (or not), same logic for any other value with the specific `appsettings.json` under a tenant folder, and that has an higher precedence than e.g. the one at the app root.

A Default tenant always exists (even it is not pre-configured) at least in an Uninitialized state, and we can create and setup another tenant even if the default one is not setup, we currently do this in our unit tests.

But that's okay, as said makes sense to have a separate and dedicated config section for all auto setup data, and then, because other tenants are not intended to be pre-configured (so not auto created by the shell settings manager), we need to use the default tenant to create them. I'm okay with this, but i think you got the idea ;)

Note: But what happens if you create a tenant already pre-configured, hmm i think it will just override its settings

> Yes, as you mentioned below context.Request.PathBase can be empty in this case I have to redirect to /

So use `$"{context.Request.PathBase}/"`, maybe check if the `PathBase.Value` is empty before adding a trailing slash. Note: Using the `PathBase` **ensures that we use the right url encoding**, this is what it does through its implicit `ToString()`.

**I'm trying it**, before talking about url host and prefix, will see later, if you have a virtual folder, can be simulated by doing the following `Map()` in the main startup, curently the redirect doesn't work, if i use the `PathBase` it works

        public void Configure(IApplicationBuilder app, IHostEnvironment env)
        {
            app.Map("/virtual", builder =>
            {
                if (env.IsDevelopment())
                {
                    builder.UseDeveloperExceptionPage();
                }

                builder.UseStaticFiles();

                builder.UseOrchardCore();
            });
        }

> Plus since we have AutoSetupOptions.AutoSetupPath option the Request Url can be different from Tenant Url in this case I have to redirect to Tenant's Url

Okay, you do a `Map(options.AutoSetupPath)`, in that case the `PathBase` will also contain the setup path, but if you do

    app.MapWhen(ctx => ctx.Request.Path.StartsWithSegments(options.AutoSetupPath), appBuilder =>
        appBuilder.UseMiddleware<AutoSetupMiddleware>());

The request `PathBase` will still contain a virtual folder (if one), the url prefix (if one), but the setup path will be in the request `Path`, so here also you can redirect by using the `PathBase`, i tried it it works

Note: We have a `StartsWithNormalizedSegments()` extension that deals better with trailing slashes.

    app.MapWhen(ctx => ctx.Request.Path.StartsWithNormalizedSegments(options.AutoSetupPath), appBuilder =>
        appBuilder.UseMiddleware<AutoSetupMiddleware>());

See in the `RunningShellTable` how we deal with the host and port (or not) and prefix (or not), see also the `Index.cshtml` of `OC.Tenants`, see `GetDisplayUrl()` for display only, and `GetEncodedUrl()` that does the url encoding for the link, this through the implicit `ToString()` of each uri component. Hmm, in this `Index.cshtml` we add the port of the current request indifferently, i think that if a host is defined (here as a string) it should contain the right port if needed, i will fix it.

Note: If the host is defined (not empty), we still need to take into account the original `PathBase` that may contain a virtual folder, this before we append to it the tenant prefix, we have an httpContext feature for this.

    var pathString = httpContext.Features.Get<ShellContextFeature>().OriginalPathBase;

Then about the prefix, regarding the code in the running shell table, if a host is defined for a tenant, this tenant can have a prefix or not, i tried it with a fake domain by using the `localtest.me` site, it works with or without defining a prefix for this tenant, so in the `Index.cshtml` that's okay, we can always add the url prefix segment if not empty.

So, aplying the same, here the code that i tried and that works if we use the above `MapWhen()`, e.g. it works with a virtual folder and the url is well encoded before updating the location header.


                    var pathBase = httpContext.Request.PathBase;
                    if (!pathBase.HasValue)
                    {
                        pathBase = "/";
                    }

                    var hostString = httpContext.Request.Host;
                    if (!String.IsNullOrEmpty(setupTenant.RequestUrlHost))
                    {
                        hostString = new HostString(setupTenant.RequestUrlHost);
                    }

                    var pathString = httpContext.Features.Get<ShellContextFeature>().OriginalPathBase;
                    if (!String.IsNullOrEmpty(shellSettings.RequestUrlPrefix))
                    {
                        pathString = pathString.Add('/' + shellSettings.RequestUrlPrefix);
                    }

                    httpContext.Response.Redirect($"{httpContext.Request.Scheme}://{hostString + pathString}");

                    return;

Also, i had some `ThrowResponseAlreadyStartedException`, in fact when you redirect you just need to bypass the pipeline by doing a return, if we still have this exception we may have to check if the response `HasStarted`, will see.


But but but, if you define an url host for a given tenant (not the default one), because you pre-created it, the host is already in the ambient context (and also the url prefix), anyway you will not be able to trigger its setup if you don't use the url host that you defined for it. For the default tenant, hmm we fallback to it for any host, even if not yet defined, so if your running instance is already listening for this host, seems to be okay. If so, we would not need the above code, even for the default tenant, and the final code would be

                    var pathBase = httpContext.Request.PathBase;
                    if (!pathBase.HasValue)
                    {
                        pathBase = "/";
                    }

                    httpContext.Response.Redirect(pathBase);

                    return;

Sorry for the long story, more line to describe what i was meaning, than the final code ;)

---

What do you think about checking that the url prefix doesn't contain more than one segment, as done in the `OC.Tenants` api and mvc controllers.

Finally, do a `Remove and Sort usings` in your files, remove unused ones, order alphabetically, but `System.*` first.

jtkech:
@rserj 

First, i don't want to bother you with my suggestions and i don't want to block your PR ;)

For now i only did minor changes, trying it revealed some little issues in other places that i'm fixing in separate PRs. As said above, sometimes the browser pre-sends some requests and then it just shows up that the tenant is initializing, normal and it doesn't prevent the setup to complete, so not a big issue but better to see the redirection operate.

- For now, seems to be okay when not using an `AutoSetupPath`, unless if i intentionaly hit f5 multiple times, so one suggestion would be to not have the `AutoSetupPath` option, need more time to check it.

    Note: Here the singularity is that we are mutating things but through an http GET method.

Also related to multiple requests, sometimes i had yesSql / sqlite exceptions, here also it doesn't prevent the setup already started to complete. For now we only mark that the tenant is initializing, but not immediately and without any locking system. I already thought about this, now maybe worth to do something.

- So, the other suggestion would be to prevent multiple setups on the same tenant, maybe a global solution as now there are multiple ways to setup a tenant (Api, Worflow), but still not sure if it is worth doing it, maybe a compromise.

So, need a little more time to complete my tests and commit some suggestions, then you will be free to revert / update what you don't agree with.

jtkech:
@rserj

Sorry, did not have so much time, but just fixed the race conditions, i will commit my suggestions tomorrow

rserj:
@jtkech thanks for reviewing PR, I reworked the redirection logic by using `MapWhen `and `PathBase` I tested it with `AutoSetupPath`, and without it, it works for me.

jtkech (approved):

Issue #8372, opened 2021-01-20
Improve layer rules


deanmarcussen:
Discussed here https://www.youtube.com/watch?v=9eCYnzvIyl0

Improve the layer rules filters to include more than just a javascript interpreted rule, but also in a similar way to deployment steps, offer code based rulebuilders, with properties, that can be selected to build a rule.

I.e. 
- Always, would be a `True` rulebuilder. With no properties.
- IsHomePage could be build via a `Url` rulebuilder, with a properties of `Url` : `Equal`, `NotEqual` etc.
- Or `IsHomePage` could a rule in itself, with no properties, and the rule evaluation could be hardcoded.

A `RuleBuilder` might to too much, but not hard to do with a little vue app / view component.

The idea is simply to have more c# based rules, while still offering js as an option.

Sample from the azure edge cdn rule builders for ideas.

<img width="693" alt="Screenshot 2021-01-20 at 10 36 04" src="https://user-images.githubusercontent.com/13782679/105163419-cc94be00-5b0b-11eb-8262-d74ce6806dc2.png">

Issue #8470, opened 2021-01-30
ConcurrencyException on Multi-Tenant


jeffolmstead:
This might be most relevant to @jtkech - I upgraded to the latest code bits and now my multi-tenant (primary website) on my dev box throws this exception:

OrchardCore.Data.Documents.DocumentStoreCommitException: The 'SiteSettings' could not be persisted and cached as it has been changed by another process.
 ---> YesSql.ConcurrencyException: The document could not be updated as it has been changed by another process.
   at YesSql.Commands.UpdateDocumentCommand.ExecuteAsync(DbConnection connection, DbTransaction transaction, ISqlDialect dialect, ILogger logger)
   at YesSql.Session.FlushAsync()
   at YesSql.Session.CommitAsync()
   at YesSql.Session.CommitAsync()
   at OrchardCore.Data.Documents.DocumentStore.CommitAsync()
   --- End of inner exception stack trace ---
   at OrchardCore.Data.Documents.DocumentStore.<>c__11`1.<UpdateAsync>b__11_1(Exception exception)
   at OrchardCore.Data.Documents.DocumentStore.CommitAsync()
   at OrchardCore.Environment.Shell.Scope.ShellScope.BeforeDisposeAsync()
   at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell)
   at OrchardCore.Environment.Shell.Scope.ShellScope.ActivateShellInternalAsync()
   at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell)
   at OrchardCore.Modules.ModularTenantContainerMiddleware.Invoke(HttpContext httpContext)
   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware.Invoke(HttpContext context)

The lower level tenants work fine. It is a dev box so it isn't a huge deal, likely can blow away the main site and rebuild it without too much issue. However, if there is something important to try and figure out here, I will gladly help. I know Concurrency has been a topic in recent pull requests. I thought about removing each tenant (e.g. shell) to see if one of them could somehow be causing the issue. Look forward to any thoughts you have / how I might best diagnose this.

jtkech:
We had the same kind of issue that has been fixed, the comments in #7449 may help us

I noticed that your trace has `YesSql.Session.CommitAsync()` twice

I would ask to check for a non awaited task in your code, but it was working before your upgrade

[...]

PR #8597, opened 2021-02-14
Rules Module

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/8372

Replaces javascript layer rules with a more ui friendly, and faster condition evaluator.

Demo towards the end of this meeting https://www.youtube.com/watch?v=3K_q3tY_UwQ

UI  now draggable between conditions groups.
Screenshot 2021-02-13 at 17 56 08

Feel like I've done enough there on the ui, if someone else wants to improve it with badges / tags etc al, then they are welcome to :)

This is its own module, `Rules` which is designed to allow other modules, i.e. `Layers` to use the UI to build Rules.
It still needs a controller in the `Layers` module, as `Layers know how their rules are stored.

It's done this way, so that other modules can potentially make use of `Rules` as well. For example, I'd quite like to have menu rules - we have the facility to cache differently for anonymous / authenticated / roles / urls, so mostly translates across quite well, and I will make use of this Rules module, in my own modules, which require rules.

Needs a bit of a look to make sure I've handled the `ValueTask` evaluators correctly.

The abstractions include `IConditionEvaluator` 's which evaluate conditions and `IConditionComparers` which provide methods like `Equals`, `StartsWith` etc.

Both are extensible.

`IConditionEvaluator` is extensible with display drivers, as conditions require ui, and resolved through an `IConditionResolver` from scoped DI

`IConditionComparers` are extensible through `IOptions` - they only require localizing, and are low impact dependencies, so the `IOptions<ConditionOperatorOptions>` is used to hold singleton instances of the `Comparer` - rather than having to resolve them all from DI every time. 

They are basically static singletons, but done with `IOptions`, so that they can be localized, and described for translations.


src/OrchardCore.Modules/OrchardCore.Layers/Migrations.cs
```diff
@@ -29,5 +56,29 @@ public int UpdateFrom1()
...
+
+                layer.Rule = String.Empty;
+                #pragma warning restore 0618
+            }
+
+            await _layerService.UpdateAsync(layers);
+
+            var feature = _extensionManager.GetFeatures().FirstOrDefault(f => f.Id == "OrchardCore.Rules");
+            
+            await _shellFeaturesManager.EnableFeaturesAsync(new[] { feature }, force: true);
```
deanmarcussen, line 79: @jtkech your thoughts here if you have some time?

Despite taking a dependency in the `Manifest` for old installations, I have still had to activate the new `OrchardCore.Rules` feature manually here.

That surprised me?

And this may not be the best way to force activation of the new feature

agriffard:
Tried it.
Note: With an existing instance, the `Never` layer has not been 'migrated': the layer rule as empty so there was an error on the front. 
I set it to `false` in the admin and it worked.

deanmarcussen:
Thanks @agriffard 

How did you make the `Never` layer? via a recipe? 

Because the original ui had the `Rule` as `Required` so shouldn't have let you make the `Never` empty - you should have had to return `false`.

I'll see if I can manage that better in the migration however.

agriffard:
I created the `Never` layer manually, and I am almost sure that I gave a `false` value.

src/OrchardCore.Modules/OrchardCore.Layers/Migrations.cs
jtkech: @deanmarcussen Yes, a dependency is automatically enabled when you enable the dependent through a recipe or through the admin, but not when an already enable feature has a new dependency. Not usual that a module upgrade adds a dependency, we already did that but for missing deps related to features commonly enabled through a recipe.

> And this may not be the best way to force activation of the new feature

Looks good, as i remember here you could just enable `OC.Layers`  that woud enable its dependencies and then the missing one. Hmm, without magic string you could also use `ITypeFeatureProvider` to get the layer feature.

    var layerFeature = _typeFeatureProvider.GetFeatureForDependency(migration.GetType());
    _shellFeaturesManager.EnableFeaturesAsync(new[] { layerFeature }, force: true);

src/OrchardCore.Modules/OrchardCore.Rules/Services/RuleService.cs
```diff
@@ -0,0 +1,53 @@
...
+                => await task;
+            
+            foreach(var childCondition in rule.Conditions)
+            {
+                var evaluator = _conditionResolver.GetConditionEvaluator(childCondition);
+
+                var task = evaluator.EvaluateAsync(childCondition);
+                if (!task.IsCompletedSuccessfully)
+                {
+                    if (!Awaited(task).Result)
```
jtkech, line 30: I understand the pattern but because you can't return directly if true, here you are  forced to run it synchronously, is it intended, even if most of the time it completes immediately? Here it's like a `Getawaiter().GetResult()`

src/OrchardCore.Modules/OrchardCore.Layers/Services/LayerFilter.cs
```diff
@@ -113,13 +113,14 @@ public async Task OnResultExecutionAsync(ResultExecutingContext context, ResultE
                     bool display;
                     if (!layersCache.TryGetValue(layer.Name, out display))
                     {
-                        if (String.IsNullOrEmpty(layer.Rule))
+                        var task = _ruleService.EvaluateAsync(layer.LayerRule);
+                        if (!task.IsCompletedSuccessfully)
                         {
-                            display = false;
+                            display = AwaitedRule(task).Result;
```
jtkech, line 119: Normally this kind of pattern is used to not have to use an await and return a Task / ValueTask directly, but here we already are in an async task, here the call is synchronous as with `Getawaiter().GetResult()`.

src/OrchardCore.Modules/OrchardCore.Rules/Drivers/ContentTypeConditionEvaluatorDriver.cs
```diff
@@ -0,0 +1,60 @@
...
+                (context.Shape.Properties.TryGetValue(nameof(ContentTypeSettings.Stereotype), out var stereotype) && 
+                stereotype == null))
+            {
+                _contentTypes.Add(contentItem.ContentType);
+            }
+
+            return Task.FromResult<IDisplayResult>(null);
+        }
+
+        ValueTask<bool> IConditionEvaluator.EvaluateAsync(Condition condition)
```
jtkech, line 43: Is it necessary to implement the interface explicitly? As i remember i removed some in Orchard that was not necessary, maybe i also needed to update the services registrations, but not sure

deanmarcussen:
> I created the `Never` layer manually, and I am almost sure that I gave a `false` value.

ok found the issue there, where it didn't correctly migrate unknown conditions, to scripts. Fixed

src/OrchardCore.Modules/OrchardCore.Rules/Services/RuleService.cs
deanmarcussen: Reverted @jtkech - I don't well understand that pattern, so was hoping for better optimization.

src/OrchardCore.Modules/OrchardCore.Rules/Drivers/ContentTypeConditionEvaluatorDriver.cs
deanmarcussen: Yes, in this case, because the driver inherits from `ContentDisplayDriver` but is also an `IConditionEvaluator`

implementing interface explicitly, in a similar way in which we do the `IContentPartDisplayDriver` - to do the dispatch to a `TPart`, or in this case a `TCondition`.

Maybe there is a better pattern for dispatch? 
but I've learnt that one since O1 as well ;)

src/OrchardCore.Modules/OrchardCore.Layers/Migrations.cs
deanmarcussen: Done. nice cheers - works better. I like that

src/OrchardCore.Modules/OrchardCore.Rules/Drivers/ContentTypeConditionEvaluatorDriver.cs
jtkech: Okay no problem, and maybe better for readability, and it is required when a class implement 2 methods with the same signature but for 2 different interfaces, but not the case here, so just to say that here we could also use

        public ValueTask<bool> EvaluateAsync(Condition condition)
            => EvaluateAsync(condition as ContentTypeCondition);

deanmarcussen: Ok, cool. was there a disadvantage to it? 
I quite liked it for the readability... but don't mind if there's a preference the other way.

Just noticed the `Evaluate` can just be private, so changing that.

jtkech: @deanmarcussen 

No disadvantage, was just curious if there was any other reason, yes about readability

deanmarcussen: @jtkech looked again, and yes it was a bit odd on that driver :) cheers for spotting.

I've kept it on `ConditionEvaluator<T>` as it looks more obvious to me when I read it.

Any other thing you saw? will probably merge it tomorrow if not.

jtkech: Nothing else, i didn't have time to review it more, but i'm confident ;)

Skrypt:
Small issue with existing websites : 

image

deanmarcussen:
Can you check if you have any failed migrations @Skrypt ? and look at your migration history for layers

There is a migration in place to create this layer rule, but if it has failed to run, then ...

Skrypt:
```
2021-02-22 16:59:07.7049|Default||95e920cc-43bb5fe59cdfb2c5.||OrchardCore.Data.Migration.DataMigrationManager|ERROR|Error while running migration version 2 for 'OrchardCore.Layers'. YesSql.ConcurrencyException: The document could not be updated as it has been changed by another process.
   at YesSql.Commands.UpdateDocumentCommand.ExecuteAsync(DbConnection connection, DbTransaction transaction, ISqlDialect dialect, ILogger logger)
   at YesSql.Session.FlushAsync()
   at YesSql.Services.DefaultQuery.Query`1.FirstOrDefaultImpl()
   at OrchardCore.Environment.Shell.Data.Descriptors.ShellDescriptorManager.GetShellDescriptorAsync() in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Infrastructure\Shell\ShellDescriptorManager.cs:line 52
   at OrchardCore.Environment.Shell.Data.Descriptors.ShellDescriptorManager.UpdateShellDescriptorAsync(Int32 priorSerialNumber, IEnumerable`1 enabledFeatures, IEnumerable`1 parameters) in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Infrastructure\Shell\ShellDescriptorManager.cs:line 74
   at OrchardCore.Environment.Shell.ShellDescriptorFeaturesManager.UpdateFeaturesAsync(ShellDescriptor shellDescriptor, IEnumerable`1 featuresToDisable, IEnumerable`1 featuresToEnable, Boolean force) in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore\Shell\ShellDescriptorFeaturesManager.cs:line 84
   at OrchardCore.Environment.Shell.ShellFeaturesManagerExtensions.EnableFeaturesAsync(IShellFeaturesManager shellFeaturesManager, IEnumerable`1 features, Boolean force) in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Abstractions\Shell\Extensions\ShellFeaturesManagerExtensions.cs:line 18
   at OrchardCore.Layers.Migrations.UpdateFrom2Async() in E:\Repositories\affaires-extra-sc\src\OrchardCore.Modules\OrchardCore.Layers\Migrations.cs:line 78
   at OrchardCore.Data.Migration.DataMigrationManager.UpdateAsync(String featureId) in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Data\Migration\DataMigrationManager.cs:line 224    at YesSql.Commands.UpdateDocumentCommand.ExecuteAsync(DbConnection connection, DbTransaction transaction, ISqlDialect dialect, ILogger logger)
   at YesSql.Session.FlushAsync()
   at YesSql.Services.DefaultQuery.Query`1.FirstOrDefaultImpl()
   at OrchardCore.Environment.Shell.Data.Descriptors.ShellDescriptorManager.GetShellDescriptorAsync() in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Infrastructure\Shell\ShellDescriptorManager.cs:line 52
   at OrchardCore.Environment.Shell.Data.Descriptors.ShellDescriptorManager.UpdateShellDescriptorAsync(Int32 priorSerialNumber, IEnumerable`1 enabledFeatures, IEnumerable`1 parameters) in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Infrastructure\Shell\ShellDescriptorManager.cs:line 74
   at OrchardCore.Environment.Shell.ShellDescriptorFeaturesManager.UpdateFeaturesAsync(ShellDescriptor shellDescriptor, IEnumerable`1 featuresToDisable, IEnumerable`1 featuresToEnable, Boolean force) in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore\Shell\ShellDescriptorFeaturesManager.cs:line 84
   at OrchardCore.Environment.Shell.ShellFeaturesManagerExtensions.EnableFeaturesAsync(IShellFeaturesManager shellFeaturesManager, IEnumerable`1 features, Boolean force) in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Abstractions\Shell\Extensions\ShellFeaturesManagerExtensions.cs:line 18
   at OrchardCore.Layers.Migrations.UpdateFrom2Async() in E:\Repositories\affaires-extra-sc\src\OrchardCore.Modules\OrchardCore.Layers\Migrations.cs:line 78
   at OrchardCore.Data.Migration.DataMigrationManager.UpdateAsync(String featureId) in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Data\Migration\DataMigrationManager.cs:line 224
```

Skrypt:
image

image

Fails on line 78. I'm guessing that since line 74 is noop the records are not yet updated.

Skrypt:
Ok this is a caching issue.

```
2021-02-23 10:49:59.9782|Default||fb2792bf-4d4499ea160702ae.||Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware|ERROR|An unhandled exception has occurred while executing the request. OrchardCore.Data.Documents.DocumentStoreCommitException: The 'LayersDocument' could not be persisted and cached as it has been changed by another process.
 ---> YesSql.ConcurrencyException: The document could not be updated as it has been changed by another process.
   at YesSql.Commands.UpdateDocumentCommand.ExecuteAsync(DbConnection connection, DbTransaction transaction, ISqlDialect dialect, ILogger logger)
   at YesSql.Session.FlushAsync()
   at YesSql.Session.CommitAsync()
   at YesSql.Session.CommitAsync()
   at OrchardCore.Data.Documents.DocumentStore.CommitAsync() in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Data\Documents\DocumentStore.cs:line 141
   --- End of inner exception stack trace ---
   at OrchardCore.Data.Documents.DocumentStore.<>c__11`1.<UpdateAsync>b__11_1(Exception exception) in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Data\Documents\DocumentStore.cs:line 96
   at OrchardCore.Data.Documents.DocumentStore.CommitAsync() in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Data\Documents\DocumentStore.cs:line 160
   at OrchardCore.Environment.Shell.Scope.ShellScope.BeforeDisposeAsync() in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Abstractions\Shell\Scope\ShellScope.cs:line 345
   at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell) in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Abstractions\Shell\Scope\ShellScope.cs:line 241
   at OrchardCore.Environment.Shell.Scope.ShellScope.ActivateShellInternalAsync() in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Abstractions\Shell\Scope\ShellScope.cs:line 288
   at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell) in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Abstractions\Shell\Scope\ShellScope.cs:line 235
   at OrchardCore.Modules.ModularTenantContainerMiddleware.Invoke(HttpContext httpContext) in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore\Modules\ModularTenantContainerMiddleware.cs:line 59
   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware.Invoke(HttpContext context)    at OrchardCore.Data.Documents.DocumentStore.<>c__11`1.<UpdateAsync>b__11_1(Exception exception) in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Data\Documents\DocumentStore.cs:line 96
   at OrchardCore.Data.Documents.DocumentStore.CommitAsync() in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Data\Documents\DocumentStore.cs:line 160
   at OrchardCore.Environment.Shell.Scope.ShellScope.BeforeDisposeAsync() in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Abstractions\Shell\Scope\ShellScope.cs:line 345
   at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell) in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Abstractions\Shell\Scope\ShellScope.cs:line 241
   at OrchardCore.Environment.Shell.Scope.ShellScope.ActivateShellInternalAsync() in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Abstractions\Shell\Scope\ShellScope.cs:line 288
   at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell) in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore.Abstractions\Shell\Scope\ShellScope.cs:line 235
   at OrchardCore.Modules.ModularTenantContainerMiddleware.Invoke(HttpContext httpContext) in E:\Repositories\affaires-extra-sc\src\OrchardCore\OrchardCore\Modules\ModularTenantContainerMiddleware.cs:line 59
   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware.Invoke(HttpContext context)
```

deanmarcussen:
@jtkech I talked through this with @Skrypt and we got further ahead with it, but then moved onto another error with the `MediaTokenSettingsUpdater` giving the same style of concurrency exception - though why it is trying to activate I'm not sure, as the feature should be on.

I can't repro locally with SQL server which is not helping.

I see a possible issue in the `ILayerService` but see what you think

Here in the migration (and in all the controllers that do layer updates)
https://github.com/OrchardCMS/OrchardCore/blob/99016246405930de39cbc3d929433ad32b1a1d83/src/OrchardCore.Modules/OrchardCore.Layers/Migrations.cs#L61

Then in the service
https://github.com/OrchardCMS/OrchardCore/blob/99016246405930de39cbc3d929433ad32b1a1d83/src/OrchardCore.Modules/OrchardCore.Layers/Services/LayerService.cs#L60-L64

It is reloading, setting layers on the existing, but saving the sent in document.

So I'm not sure what is coming from the YesSql Identity Map, and what is / or isn't. 

Or if two calls to `LoadAsync` are fetching from the Identity Map, or requerying.

It seemed like when we changed that line to `_documentManager.UpdateAsync(layers)` we got past the migration error with @Skrypt but as I said we then moved onto another with `MediaTokenSettingsUpdater`

@Skrypt can you post the log for that one here as well if you still have it?

Skrypt:
image

jtkech:
@Skrypt @deanmarcussen 

Just tried, i can't repro with sqlite

> It is reloading, setting layers on the existing, but saving the sent in document.

That's okay, it is scoped cached

@deanmarcussen did you try it with @Skrypt or with another app?

In the above trace i see 2 consecutive `YesSql.Session.CommitAsync()`

I already saw this in another issue, as i remember it was a non awaited task, but not sure

Can't right now, i will work on it this night, maybe with @Skrypt if he is awake

In the meantime, can you try to comment out the line that enable the feature => Save shell descriptor that does an earlier `CommmitAsync()`

Also, @Skrypt do you have a custom migration e.g. a `Task<int> UpdateFromN()` but without the `Async` suffix?

deanmarcussen:
@jtkech ok, cool wasn't sure they were scoped cache.

I tried it with @Skrypt on his site, and saw it fail, but can't repro on sqlite or SQL server locally.

But maybe yes, it is another unawaited task, or the `UpdateFromN()` are two possibilities, it reads more like that, when we read the trace.

Skrypt:
We found the issue.

jtkech:
Same issue as #8470 

Could be fixed by https://github.com/sebastienros/yessql/issues/343 opened 21 days ago ;)

deanmarcussen:
Great! 

Ok so we need to fix that in YesSql.

Are we seeing it now because we have just started using the Concurrency Checking feature @jtkech ?

It just seems so long ago that the `Version` property went in.

Skrypt:
Maybe there is nothing to fix too. The issue is that my Document table Version column should have been migrated properly by having affected a value to my contentItems. I know that if I have for example multiple versions of the same content item it should have incremented the Version on each of the content items from oldest to latest version. So, I've started working on a SQL Query here to update this in SQL Server "at least". Else, it would require a complete migration that loops on all Content Items that have a Version = NULL in the database and then set the proper Version to each of it's ContentItemVersion's...

Else, for all the other records we have in the Document table they can be reset to Version = 0 unless I forgot something.

deanmarcussen:
Yup the sql looks like the right thing to do. Should sort any potential future issues

jtkech:
@deanmarcussen 

> Are we seeing it now because we have just started using the Concurrency Checking feature @jtkech ?

Yes, it was already used but only for `User`, now it is also used for all shared documents as Layers, Templates, SiteSettings, and so on, but it is still not used e.g. for `ContentItem` whose `Version` is not checked and incremented, so always equal to 1, unless on old database where they can be null.

@Skrypt

That's okay, the Version column can have any non null value, it is only used to check that when we load a row for update with a given `Version`, it still has the same `Version` in the database when we update it, meaning that no concurrent process already updated it, if so we increment the `Version` and update the row.

So, it is only a **row** `Version` for concurrency checking, it is not related to the `ContentItemVersionId`, e.g. we have different rows per `ContentItemVersionId` that may have the same value in the `Version` column, anyway for now we don't use concurrency checking for content items, so currently the `Version` is not incremented and will be always equal to 1 (after the first save), so on existing database we would have to set all null ones to 1, but that's okay to set them to 0.

I think that's the best to do is to update YesSql, when a `Version` is null it is mapped to 0 in the related YesSql document, so when updating a document whose `Version` is 0, we would just have to say that it is okay if the current `Version` of the related row in the database is equal to `0 OR is NULL`.

Skrypt:
Ok, I agree for a fix on YesSQL side if possible. By the mean time the workaround is to execute the SQL Query.

PR #8602, opened 2021-02-15
Fix Lucene boost values on queries.

### LINKED ISSUE #8600: [Lucene] Applying a boost doesn't seem to have any effect
I've tried two ways:
1. First is from this https://github.com/OrchardCMS/OrchardCore/issues/4825#issuecomment-554665911.

```
{
  "from": 0,
  "query": {
    "bool": {
      "should": [
        {
          "wildcard": {
            "TitlePart": "test*",
            "boost": 2
          }
        },
        {
          "wildcard": {
            "Article.Body": "test*"
          }
        }
      ]
    }
  },
  "size": 10
}
```
It returns the same order as when there is no boost property. Some articles without "test" in the title appear first.

2. I was also trying to use with the simpleQueryString.
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html
But it doensn't work, as if I didn't include the "TitlePart" field.

```
{
  "from": 0,
  "query": {
    "bool": {
      "must": [
        {
          "simple_query_string": {
            "analyze_wildcard": true,
            "fields": [
              "TitlePart^2",
              "Article.Body"
            ],
            "query": "(test) (test*)"
          }
        }
      ]
    }
  },
  "size": 10
}
```

With:
```
"fields": [
    "TitlePart^2",
    "Article.Body"
  ]
```
Returns only results of Article.Body

With:
```
"fields": [
    "TitlePart^2"
  ]
```
Returns no results.

With:
```
"fields": [
    "TitlePart"
  ]
```
Returns results.

Am I doing something wrong?
If it's a bug or not implemented I can help.

Skrypt:
WIP

- [x] Fix SimpleQueryStringQueryProvider field weights.
- [x] Fix RegexpQueryProvider field weights.
- [x] Fix Boolean "Should" query field weights when using multiple Wildcard Queries.
- [ ] Add Unit tests

Fixes #8600

### DIFF
--- a/src/OrchardCore/OrchardCore.Lucene.Core/QueryProviders/SimpleQueryStringQueryProvider.cs
+++ b/src/OrchardCore/OrchardCore.Lucene.Core/QueryProviders/SimpleQueryStringQueryProvider.cs
@@ -24,7 +24,7 @@ namespace OrchardCore.Lucene.QueryProviders
 
             foreach(var field in fields)
             {
-                if(field.Contains('^') && Single.TryParse(field.Split("^").Last(), out weight))
+                if(field.Contains('^', StringComparison.Ordinal) && Single.TryParse(field.Split("^").Last(), out weight))
                 {
                     weights.Remove(field);
                     weights.Add(field.Split("^").First(), weight);
--- a/src/OrchardCore/OrchardCore.Lucene.Core/QueryProviders/SimpleQueryStringQueryProvider.cs
+++ b/src/OrchardCore/OrchardCore.Lucene.Core/QueryProviders/SimpleQueryStringQueryProvider.cs
@@ -1,4 +1,6 @@
+using System;
 using System.Linq;
+using System.Globalization;
 using Lucene.Net.QueryParsers.Simple;
 using Lucene.Net.Search;
 using Newtonsoft.Json.Linq;
@@ -17,7 +19,18 @@ namespace OrchardCore.Lucene.QueryProviders
             var queryString = query["query"]?.Value<string>();
             var fields = query["fields"]?.Values<string>() ?? new string[0];
             var defaultOperator = query["default_operator"]?.Value<string>().ToLowerInvariant() ?? "or";
-            var weights = fields.ToDictionary(field => field, field => 1.0f);
+            float weight = 1.0f;
+            var weights = fields.ToDictionary(field => field, field => weight);
+
+            foreach(var field in fields)
+            {
+                if(field.Contains("^") && Single.TryParse(field.Split("^").Last(), out weight))
+                {
+                    weights.Remove(field);
+                    weights.Add(field.Split("^").First(), weight);
+                }
+            }
+
             var queryParser = new SimpleQueryParser(context.DefaultAnalyzer, weights);
             switch (defaultOperator)
             {
--- a/src/OrchardCore/OrchardCore.Lucene.Core/QueryProviders/SimpleQueryStringQueryProvider.cs
+++ b/src/OrchardCore/OrchardCore.Lucene.Core/QueryProviders/SimpleQueryStringQueryProvider.cs
@@ -24,7 +24,7 @@ namespace OrchardCore.Lucene.QueryProviders
 
             foreach(var field in fields)
             {
-                if(field.Contains("^") && Single.TryParse(field.Split("^").Last(), out weight))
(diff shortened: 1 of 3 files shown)


src/OrchardCore/OrchardCore.Lucene.Core/QueryProviders/SimpleQueryStringQueryProvider.cs
```diff
@@ -17,7 +19,18 @@ public Query CreateQuery(ILuceneQueryService builder, LuceneQueryContext context
             var queryString = query["query"]?.Value<string>();
             var fields = query["fields"]?.Values<string>() ?? new string[0];
             var defaultOperator = query["default_operator"]?.Value<string>().ToLowerInvariant() ?? "or";
-            var weights = fields.ToDictionary(field => field, field => 1.0f);
+            float weight = 1.0f;
+            var weights = fields.ToDictionary(field => field, field => weight);
+
+            foreach(var field in fields)
+            {
+                if(field.Contains("^") && Single.TryParse(field.Split("^").Last(), out weight))
```
deanmarcussen, line 27: Could probably use `StringComparison.Ordinal` in the `Contains`.

but looks good to me

src/OrchardCore/OrchardCore.Lucene.Core/QueryProviders/SimpleQueryStringQueryProvider.cs
```diff
@@ -17,7 +19,18 @@ public Query CreateQuery(ILuceneQueryService builder, LuceneQueryContext context
             var queryString = query["query"]?.Value<string>();
             var fields = query["fields"]?.Values<string>() ?? new string[0];
             var defaultOperator = query["default_operator"]?.Value<string>().ToLowerInvariant() ?? "or";
-            var weights = fields.ToDictionary(field => field, field => 1.0f);
+            var weight = 1.0f;
+            var weights = fields.ToDictionary(field => field, field => weight);
+
+            foreach(var field in fields)
+            {
+                if(field.Contains("^", StringComparison.Ordinal) && Single.TryParse(field.Split("^").Last(), out weight))
```
jtkech, line 27: For infos, just before going to sleep, by default it doesn't use any comparisonType 

        public bool Contains(string value)
        {
            if (value == null)
                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.value);

            return SpanHelpers.IndexOf(
                ref _firstChar,
                Length,
                ref value._firstChar,
                value.Length) >= 0;
        }

Here the best is to pass a character `field.Contains('^')`

        public bool Contains(char value) => SpanHelpers.Contains(ref _firstChar, value, Length);

Skrypt: I don't think it was needed because a '^' char might be the same in every culture. But here, I can't assert this fact because I'm not omniscient 😉

Skrypt:
These Lucene Queries needs unit test badly but I will make a second PR.

Skrypt:
Unit Tests on a future PR. Started coding something but not functional yet.

src/OrchardCore/OrchardCore.Lucene.Core/QueryProviders/SimpleQueryStringQueryProvider.cs
```diff
@@ -17,7 +19,18 @@ public Query CreateQuery(ILuceneQueryService builder, LuceneQueryContext context
             var queryString = query["query"]?.Value<string>();
             var fields = query["fields"]?.Values<string>() ?? new string[0];
             var defaultOperator = query["default_operator"]?.Value<string>().ToLowerInvariant() ?? "or";
-            var weights = fields.ToDictionary(field => field, field => 1.0f);
+            var weight = 1.0f;
+            var weights = fields.ToDictionary(field => field, field => weight);
+
+            foreach(var field in fields)
+            {
+                if(field.Contains('^', StringComparison.Ordinal) && Single.TryParse(field.Split("^").Last(), out weight))
```
jtkech, line 27: @Skrypt 

Yes, for readability maybe better to specify `StringComparison.Ordinal` as this is what we want, even here it executes a few more lines of code (but no significant impact)

So just to know, did you re-add it because you saw some perf impact? Normally `Contains()` doesn't use any comparison type by default.

Note: Didn't check the version of `Split('^')` accepting a char ;)

PR #8603, opened 2021-02-16
Dashboard Enhancement - Support Drag, Drop and Resize 

### LINKED ISSUE #8601: Dashboard Enhancement - Support Drag, Drop and Resize
Admin Dashboard Usability - Support Drag, Drop and Resize

As per https://github.com/OrchardCMS/OrchardCore/pull/7943#issuecomment-774180173

- "snap grid" 
- Have a button to go in edit/view mode, and the edit mode being protected by a permission. In edit mode we can add new widgets, move, edit, resize and delete them.
- All widgets can always be moved or resized with `ManageAdminDashboard` Permission 
- Show "undo" when there is a change.

ns8482e:
Added support for drag drop sort and resize
Fixes #8601

dashboard-3


hishamco:
Great work Niraj, can we have a demo in the standup?

agriffard:
image

Here is how it looks like when I enable the feature.
The bottom is over the border probably because of the `list-group-flush` class in the list created from the recipe.

Could we hide the date and status in View mode?

When I click on 'Manage dashboard':

image

Add some space between buttons.
I didn't understand what was doing the 'View' button until I click on it. It is more like a 'Go back to dashboard'.

When I create or edit a widget and specify Height = 2 or Width = 2, it is not taken into account.

image

Add space between` successfully updated.` and `Undo`

agriffard:
If possible: Do not display Card header if title is empty.

image

ns8482e:
> If possible: Do not display Card header if title is empty.

Good catch - seems an issue with `TitlePart`, will fix it.

src/OrchardCore.Modules/OrchardCore.AdminDashboard/DashboardManageAttribute.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc.Filters;
+namespace OrchardCore.AdminDashboard
+{
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
+    public class DashboardManageAttribute : Attribute, IAsyncResourceFilter
```
deanmarcussen, line 8: I would suggest just using an `httpContext.Feature` rather than needing another filter here.

src/OrchardCore.Modules/OrchardCore.AdminDashboard/Drivers/DashboardContentDisplayDriver.cs
```diff
@@ -0,0 +1,119 @@
...
+            _authorizationService = authorizationService;
+            _contentManager = contentManager;
+
+        }
+
+        public override async Task<IDisplayResult> DisplayAsync(ContentItem model, BuildDisplayContext context)
+        {
+            var httpContext = _httpContextAccessor.HttpContext;
+            var isManageRequest = httpContext.Items.ContainsKey(typeof(DashboardManageAttribute));
+            var hasPublished = await _contentManager.HasPublishedVersionAsync(model);
```
deanmarcussen, line 36: Here I think you can just return `Task.FromResult<IDisplayResult>(null)` if the feature is not set.

src/OrchardCore.Modules/OrchardCore.AdminDashboard/Drivers/DashboardContentDisplayDriver.cs
```diff
@@ -0,0 +1,119 @@
...
+            {
+                return Task.FromResult(isManageRequest);
+            });
+            results.Add(dragHandle);
+
+            var editButton = Dynamic("Dashboard_EditButton", m =>
+            {
+                m.ContentItem = model;
+            }).Location("ActionsMenu:after")
+            .RenderWhen(async () =>
```
deanmarcussen, line 60: Probably you don't need to use `RenderWhen` here at all.

If you just return null on `!isManageRequest`

src/OrchardCore.Modules/OrchardCore.AdminDashboard/Drivers/DashboardContentDisplayDriver.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            var deleteButton = Dynamic("Dashboard_DeleteButton", m =>
+            {
+                m.ContentItem = model;
+            }).Location("ActionsMenu:after")
+            .RenderWhen(async () =>
+            {
+                
+                return await Task.FromResult(isManageRequest && hasDeletePermission);
+            });
```
deanmarcussen, line 74: Here you can just return the Task, rather than awaiting a result.
i.e.
`.RenderWhen(() => Task.FromResult(hasDeletePermission)`

src/OrchardCore.Modules/OrchardCore.AdminDashboard/Drivers/DashboardContentDisplayDriver.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            var shapeTag = Shape("DashboardWidget__ContentsTags", new ContentItemViewModel(model))
+            .Location("Detail", "Tags:10")
+            .RenderWhen( () =>
+            {
+                return Task.FromResult(isManageRequest);
+            });
+            results.Add(shapeTag);
+
+            return await Task.FromResult<IDisplayResult>(Combine(results.ToArray()));
```
deanmarcussen, line 116:
```suggestion
            return await Combine(results.ToArray());
```

src/OrchardCore.Modules/OrchardCore.Html/Drivers/HtmlBodyPartDisplayDriver.cs
```diff
@@ -43,7 +43,6 @@ public override IDisplayResult Display(HtmlBodyPart HtmlBodyPart, BuildPartDispl
             return Initialize<HtmlBodyPartViewModel>(GetDisplayShapeType(context), m => BuildViewModelAsync(m, HtmlBodyPart, context))
                 .Location("Detail", "Content:5")
                 .Location("Summary", "Content:10")
-                .Location("DetailAdmin", "Content:10") // For dashboard widgets;
```
deanmarcussen, line 46: Intentional to remove these?

Why?

src/OrchardCore.Modules/OrchardCore.Markdown/Drivers/MarkdownBodyPartDisplayDriver.cs
```diff
@@ -47,7 +47,6 @@ public override IDisplayResult Display(MarkdownBodyPart markdownBodyPart, BuildP
             return Initialize<MarkdownBodyPartViewModel>(GetDisplayShapeType(context), m => BuildViewModel(m, markdownBodyPart, context))
                 .Location("Detail", "Content:10")
                 .Location("Summary", "Content:10")
-                .Location("DetailAdmin", "Content:10") // For dashboard widgets
```
deanmarcussen, line 50: as above

deanmarcussen (commented):
For info @ns8482e 

The `.RenderWhen` is a small optimization to prevent too much work being done when items aren't being placed. (Like when rendering `Detail` so stopping work that `SummaryAdmin` shapes have to do, (mostly for auth) that will never be placed).

So not so necesary here, as you can just get out quicker if it's not dashboard widget related.

src/OrchardCore.Modules/OrchardCore.Html/Drivers/HtmlBodyPartDisplayDriver.cs
ns8482e: Reusing existing `Detail` `DisplayType` and it's placement, so no need of  defining `DetailAdmin` Display type and it's placement.

src/OrchardCore.Modules/OrchardCore.Markdown/Drivers/MarkdownBodyPartDisplayDriver.cs
ns8482e: Reusing existing Detail DisplayType and it's placement, so no need of defining DetailAdmin Display type and it's placement.

deanmarcussen: I think the intention here was actually to have a different display type, so it could be customised for the admin if required.

src/OrchardCore.Modules/OrchardCore.Html/Drivers/HtmlBodyPartDisplayDriver.cs
sebastienros: It can't be Detail. Detail is for front-end.

sebastienros:
I don't see the reason to display the metadata when not in edit mode. Most of the time this dashboard will be in display mode, so this should be optimized for this mode.

agriffard:
A `dropdown-item` class is missing on some buttons like Unpublish.

src/OrchardCore.Modules/OrchardCore.Html/Drivers/HtmlBodyPartDisplayDriver.cs
ns8482e: UI content is same for `Detail` and `DetailAdmin`, why create new `DisplayType`?  Creating `DetailAdmin` display type means we need to define `Location` in drivers for every parts available in `OrchardCore` and duplicate the templates with same code

src/OrchardCore.Modules/OrchardCore.AdminDashboard/DashboardManageAttribute.cs
ns8482e: Updated code to use Feature

src/OrchardCore.Modules/OrchardCore.AdminDashboard/Drivers/DashboardContentDisplayDriver.cs
ns8482e: Done

src/OrchardCore.Modules/OrchardCore.AdminDashboard/Drivers/DashboardContentDisplayDriver.cs
ns8482e: Removed `RenderWhen`

src/OrchardCore.Modules/OrchardCore.AdminDashboard/Drivers/DashboardContentDisplayDriver.cs
ns8482e: Removed `RenderWhen`

src/OrchardCore.Modules/OrchardCore.Html/Drivers/HtmlBodyPartDisplayDriver.cs
ns8482e: Revert back to use `DetailAdmin`

src/OrchardCore.Modules/OrchardCore.Markdown/Drivers/MarkdownBodyPartDisplayDriver.cs
ns8482e: Revert back to use `DetailAdmin`

ns8482e:
> I don't see the reason to display the metadata when not in edit mode. Most of the time this dashboard will be in display mode, so this should be optimized for this mode.

Removed metadata from View

ns8482e:
> A `dropdown-item` class is missing on some buttons like Unpublish.

fixed

agriffard:
Can you please enahnce the drag and drop area?
The yellow triangle is difficult to understand.
Can you display a zone with a dashed border?

ns8482e:
> Can you please enahnce the drag and drop area?
> The yellow triangle is difficult to understand.
> Can you display a zone with a dashed border?

- That could be because of dense flow, I removed it for now, may be we could add it later in settings.
- Removed triangle drop cursor and replaced with border.

src/OrchardCore.Modules/OrchardCore.AdminDashboard/Drivers/DashboardContentDisplayDriver.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            var results = new List<IDisplayResult>();
+
+            var hasPublished = await _contentManager.HasPublishedVersionAsync(model);
+            var hasDraft = model.HasDraft();
+            var hasEditPermission = await _authorizationService.AuthorizeAsync(httpContext.User, CommonPermissions.EditContent, model);
+            var hasDeletePermission = await _authorizationService.AuthorizeAsync(httpContext.User, CommonPermissions.DeleteContent, model);
+            var hasPublishPermission = await _authorizationService.AuthorizeAsync(httpContext.User, CommonPermissions.PublishContent, model);
+
+            var dragHandle = Dynamic("Dashboard_DragHandle", m =>
```
deanmarcussen, line 50:
```suggestion
            var dragHandle = Initialize<ContentItemViewModel>("Dashboard_DragHandle", m =>
```

Great thanks for moving it to a feature, and cleaning up the async.

Suggestion here is that we are trying to move away from dynamic where possible, and of all the shape factory creations, this is the better one for allocations. It will create a strongly typed `ContentItemViewModel` for each, which means you can use that as the model in the views, rather than dynamic.

I'm not seeing any reason for them to be dynamic, unless I am wrong?

So here, and all the other shapes from this driver, look like they can be initialized.

Can you also have a look through and clean up all the extra line breaks too please ?

src/OrchardCore.Modules/OrchardCore.AdminDashboard/Drivers/DashboardContentDisplayDriver.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            if (hasDraft && hasEditPermission)
+            {
+                var discardDraftButton = Dynamic("Dashboard_DiscardDraftButton", m =>
+                {
+                    m.ContentItem = model;
+                }).Location("ActionsMenu:after");
+                results.Add(discardDraftButton);
+            }
+            var shapeTag = Shape("DashboardWidget_DetailAdmin__ContentsTags", new ContentItemViewModel(model))
```
deanmarcussen, line 102: Here to `Initialize<ContentItemViewModel>` is better

ns8482e: done

src/OrchardCore.Modules/OrchardCore.AdminDashboard/Drivers/DashboardContentDisplayDriver.cs
ns8482e: done

src/OrchardCore.Modules/OrchardCore.Html/Drivers/HtmlBodyPartDisplayDriver.cs
ns8482e: @dalenewman  @sebastienros  with `DetailAdmin` - display type

Currently adding a field or part (e.g. media field) in dashboard widget doesn't render them by default and requires to manually define placement for each field or part used

sebastienros: I think we realized that with @agriffard and the solution is to add a default placement for DetailAdmin like we do for Detail.

sebastienros: Apparently like the one you are removing here ;)

sebastienros:
Some other comments from the meeting that happened yesterday and is not yet live:
- Would be great to see the "snap" effect on the grid when we resize
- Hide the position/width/height inputs behind a type setting such that by default they are not displayed when we edit a widget. The visual editor is sufficient now.
- Rename the "View" button into "Close", and "New" into "Add". Could be "Close Editor", "Leave Editor" something more obvious. And also "Add Widget" if there is space. I found "New" and "View" not obvious in their intent.

sebastienros:
We need to start thinking about which widgets we should provide in some recipes (not the scope of this PR). For instance the Blog recipe could enable it, and provide a widget displaying the latest posts, archive stats, tags stats (tag cloud?). This could be done with dynamic types and liquid templates, and would be a great sandbox for devs to see what can be done with queries and liquid. A good set of ideas will probably be found in WordPress.
Alternatively each module could provide at least one widget type.

ns8482e:
> Some other comments from the meeting that happened yesterday and is not yet live:
> 
> * Would be great to see the "snap" effect on the grid when we resize

Done

> * Hide the position/width/height inputs behind a type setting such that by default they are not displayed when we edit a widget. The visual editor is sufficient now.

Does it mean we don't need `DashboardPartDisplayDriver` and attach `DashboardPart` using weld - the magical way and not needed to add `DashboardPart` manually?

> * Rename the "View" button into "Close", and "New" into "Add". Could be "Close Editor", "Leave Editor" something more obvious. And also "Add Widget" if there is space. I found "New" and "View" not obvious in their intent.

Done

> * The grid should always be 6 of width, and not switch to 7 when the menu is collapsed

Currently it's responsive - adding fix width, will also apply to smaller screen and become non-responsive. We can add configurable columns/row settings in future PR.

giannik:
#8769

PR #8605, opened 2021-02-16
Provide admin user id and other properties to recipes.

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/7603

Provides a `Feature` allowing setup to pass the admin user id to recipe migrations.

And then uses similar to make sure that all things that execute recipes, get the basic environment information they need.

This is almost an event handler, but it isn't quite, because it has to run before the recipeexecutor, and shouldn't be run when running nested recipes.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Recipes/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Recipes/Controllers/AdminController.cs
@@ -26,36 +26,33 @@ namespace OrchardCore.Recipes.Controllers
         private readonly IExtensionManager _extensionManager;
         private readonly IAuthorizationService _authorizationService;
         private readonly IEnumerable<IRecipeHarvester> _recipeHarvesters;
-        private readonly INotifier _notifier;
         private readonly IRecipeExecutor _recipeExecutor;
-        private readonly ISiteService _siteService;
         private readonly IEnumerable<IRecipeEnvironmentProvider> _environmentProviders;
-        private readonly ILogger _logger;
+        private readonly INotifier _notifier;
         private readonly IHtmlLocalizer H;
+        private readonly ILogger _logger;
 
         public AdminController(
             IShellHost shellHost,
             ShellSettings shellSettings,
-            ISiteService siteService,
             IExtensionManager extensionManager,
-            IHtmlLocalizer<AdminController> localizer,
             IAuthorizationService authorizationService,
             IEnumerable<IRecipeHarvester> recipeHarvesters,
             IRecipeExecutor recipeExecutor,
-            INotifier notifier,
             IEnumerable<IRecipeEnvironmentProvider> environmentProviders,
+            INotifier notifier,
+            IHtmlLocalizer<AdminController> localizer,
             ILogger<AdminController> logger)
         {
             _shellHost = shellHost;
             _shellSettings = shellSettings;
-            _siteService = siteService;
-            _recipeExecutor = recipeExecutor;
             _extensionManager = extensionManager;
             _authorizationService = authorizationService;
             _recipeHarvesters = recipeHarvesters;
+            _recipeExecutor = recipeExecutor;
+            _environmentProviders = environmentProviders;
             _notifier = notifier;
             H = localizer;
-            _environmentProviders = environmentProviders;
             _logger = logger;
         }
 
@@ -107,8 +104,6 @@ namespace OrchardCore.Recipes.Controllers
                 _notifier.Error(H["Recipe was not found."]);
                 return RedirectToAction("Index");
             }
-
-            var site = await _siteService.GetSiteSettingsAsync();
 
             var environment = new Dictionary<string, object>();
             await _environmentProviders.InvokeAsync((provider, env) => provider.SetEnvironmentAsync(env), environment, _logger);
--- a/src/OrchardCore.Modules/OrchardCore.Recipes/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Recipes/Controllers/AdminController.cs
@@ -106,7 +106,7 @@ namespace OrchardCore.Recipes.Controllers
             }
 
             var environment = new Dictionary<string, object>();
-            await _environmentProviders.InvokeAsync((provider, env) => provider.SetEnvironmentAsync(env), environment, _logger);
+            await _environmentProviders.InvokeAsync((provider, env) => provider.PopulateEnvironmentAsync(env), environment, _logger);
 
             var executionId = Guid.NewGuid().ToString("n");
 
--- a/src/OrchardCore.Modules/OrchardCore.Recipes/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Recipes/Controllers/AdminController.cs
@@ -1,16 +1,17 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Security.Claims;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Localization;
+using Microsoft.Extensions.Logging;
 using OrchardCore.DisplayManagement.Notify;
 using OrchardCore.Environment.Extensions;
 using OrchardCore.Environment.Shell;
 using OrchardCore.Environment.Shell.Models;
+using OrchardCore.Modules;
 using OrchardCore.Recipes.Services;
 using OrchardCore.Recipes.ViewModels;
 using OrchardCore.Security;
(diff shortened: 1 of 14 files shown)


src/OrchardCore.Modules/OrchardCore.Recipes/Controllers/AdminController.cs
```diff
@@ -102,6 +109,10 @@ public async Task<ActionResult> Execute(string basePath, string fileName)
             }
 
             var site = await _siteService.GetSiteSettingsAsync();
```
jtkech, line 111: We don't need this line anymore

src/OrchardCore.Modules/OrchardCore.Recipes/Services/RecipeDeploymentTargetHandler.cs
```diff
@@ -32,6 +42,9 @@ public async Task ImportFromFileAsync(IFileProvider fileProvider)
                 RecipeFileInfo = fileProvider.GetFileInfo("Recipe.json")
             };
 
+            var environment = new Dictionary<string, object>();
+            await _environmentProviders.InvokeAsync((provider, env) => provider.SetEnvironmentAsync(env), environment, _logger);
+
             await _recipeExecutor.ExecuteAsync(executionId, recipeDescriptor, new Dictionary<string, object>(), CancellationToken.None);
```
jtkech, line 48: You need to pass the `environment` variable

src/OrchardCore.Modules/OrchardCore.Recipes/Controllers/AdminController.cs
```diff
@@ -110,13 +121,7 @@ public async Task<ActionResult> Execute(string basePath, string fileName)
 
             try
             {
-                await _recipeExecutor.ExecuteAsync(executionId, recipe, new Dictionary<string, object>
-                {
-                    { "SiteName", site.SiteName },
-                    { "AdminUsername", User.Identity.Name },
-                    { "AdminUserId", User.FindFirstValue(ClaimTypes.NameIdentifier) }
```
jtkech, line 117: So now the admin user name / id will be populated by the related provider from the site settings super user, not from the current user that did the action, so it may be different but maybe this is before that is was wrong.

src/OrchardCore.Modules/OrchardCore.Settings/Services/RecipeEnvironmentSiteNameProvider.cs
```diff
@@ -0,0 +1,30 @@
...
+        private readonly ISiteService _siteService;
+
+        public RecipeEnvironmentSiteNameProvider(ISiteService siteService)
+        {
+            _siteService = siteService;
+        }
+
+        public async Task SetEnvironmentAsync(IDictionary<string, object> environment)
+        {
+            // When these have already been set by another provider, do not reset them.
```
jtkech, line 19: Hmm, just not sure if a module depending on OC.Settings want to override the value but have its provider registered later, but yes weird to override the site name

src/OrchardCore.Modules/OrchardCore.Users/Services/RecipeEnvironmentSuperUserProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+        public RecipeEnvironmentSuperUserProvider(ISiteService siteService, IUserService userService)
+        {
+            _siteService = siteService;
+            _userService = userService;
+        }
+
+        public async Task SetEnvironmentAsync(IDictionary<string, object> environment)
+        {
+            // When these have already been set by another provider, do not reset them.
+            if (!environment.ContainsKey("AdminUserId") && !environment.ContainsKey("AdminUsername"))
```
jtkech, line 23: Hmm, just not sure if a module depending on OC.Settings want to override the value but have its provider registered later, but yes weird to override the admin user name / id

src/OrchardCore/OrchardCore.Recipes.Abstractions/Services/IRecipeEnvironmentProvider.cs
```diff
@@ -0,0 +1,10 @@
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace OrchardCore.Recipes.Services
+{
+    public interface IRecipeEnvironmentProvider
+    {
+        Task SetEnvironmentAsync(IDictionary<string, object> environment);
```
jtkech, line 8: Maybe `PopulateEnvironmentAsync` or `MergeEnvironmentAsync`, but not sure, too tired ;)

src/OrchardCore.Modules/OrchardCore.Recipes/Controllers/AdminController.cs
jtkech: I did the change

src/OrchardCore.Modules/OrchardCore.Recipes/Services/RecipeDeploymentTargetHandler.cs
jtkech: I did the change

src/OrchardCore/OrchardCore.Recipes.Abstractions/Services/IRecipeEnvironmentProvider.cs
jtkech: I did the change with `PopulateEnvironmentAsync()`

Revert it if you don't like it, i'm also fine with `SetEnvironmentAsync()` ;)

src/OrchardCore.Modules/OrchardCore.Users/Services/RecipeEnvironmentSuperUserProvider.cs
deanmarcussen: Yes, a little odd, but also because site settings returns a default value, when it has not yet been configured, i.e. My Orchard Application`

However the setup recipe environment provider actually runs later anyway, so this check is not necesary, to resolve the issue, (I added it before discovering the registration was later for the setup environment).
So I left the check in, in case registration orders change one day, without us noticing, for safety. 

But not needed so happy to remove it, or leave it there, just in case.

Or I could just put an order property on all of them, so it's more obvious what order they would be executed in?

src/OrchardCore/OrchardCore.Recipes.Abstractions/Services/IRecipeEnvironmentProvider.cs
deanmarcussen: I like `PopulateEnvironmentAsync` thanks.

src/OrchardCore.Modules/OrchardCore.Recipes/Controllers/AdminController.cs
deanmarcussen: > before that is was wrong.

I think probably it was wrong before. The `AdminUserId` should refer to the `AdminUserId`.

If someone was wanting to use `CurrentUserId` we should (or they should) implement a `CurrentUserId` property, and then use that in their recipes?

Often it will be the same user, but it just didn't seem right to me as it was.

src/OrchardCore.Modules/OrchardCore.Recipes/Services/RecipeDeploymentTargetHandler.cs
deanmarcussen: ty :)

src/OrchardCore.Modules/OrchardCore.Recipes/Controllers/AdminController.cs
jtkech: Okay cool then

src/OrchardCore.Modules/OrchardCore.Users/Services/RecipeEnvironmentSuperUserProvider.cs
jtkech: @deanmarcussen 

Are you talking about `RecipeEnvironmentFeatureProvider` that run later and that needs an higher precedence, if so you would need to remove this check, but if it may run first you would need to keep the checking.

That said the order may only depend on an alphabetic order if the related features don't depend on each other, so yes better to have an `Order` property to ensure some important ordering, but keeping the DI registration order for the same `Order` values. An higher precedence meaning running last with an higher `Order`.

deanmarcussen: yeah that's what I'm meaning.

Cool. I agree, I'll remove this check. 

And place an `Order` property on the `IRecipeEnvironmentProvider` - makes it much more obvious what order things are happening in.

> An higher precedence meaning running last with an higher `Order`

Agree

deanmarcussen: Done

jtkech (approved):

PR #8607, opened 2021-02-16
Bootstrap 4.6

agriffard:
Rebuild assets with Bootstrap 4.6: Admin and TheTheme.


deanmarcussen:
Have you tried it in the admin to see what is broken?

agriffard:
I tried the admin, Nothing seems to be broken.
There was no breaking changes in the 4.6 version that could have an impact on what we use.

deanmarcussen (approved):
since the other pr already merged, we should get this in ;)

PR #8610, opened 2021-02-16
Fixes #8608 Adding Widget without layer breaks site

### LINKED ISSUE #8608: Adding Widget without layer breaks site
### Adding Widget without layer breaks site
There is no restriction in admin from creating a widget not assigned to any layer.
Adding such widget breaks the site :
Capture d’écran 2021-02-16 151630

### To Reproduce
1. Setup a new CMS instance with "Blank" template
2. Go to Content > Content Definition > Content Types and create a widget (eg. Html widget)
3. Go to Design > Settings > Zones and add a Zone (eg. Footer)
4. Go to Design > Widgets and click "Add Widget"
(note the message "Please create a layer before adding Widgets" but it's still possible to create one)
5. Fill the form and Publish your widget
6. Open your site : its broken

### Expected behavior
- Don't allow to publish widget without layer
- Don't throw if layer is missing

TFleury:
Fixes #8608 Adding Widget without layer breaks site
- Don't allow to publish widget without layer
- Don't throw if layer is null

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Layers/Drivers/LayerMetadataWelder.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Layers/Drivers/LayerMetadataWelder.cs
@@ -8,6 +8,7 @@ using OrchardCore.DisplayManagement.Views;
 using OrchardCore.Layers.Models;
 using OrchardCore.Layers.Services;
 using OrchardCore.Layers.ViewModels;
+using OrchardCore.Mvc.ModelBinding;
 
 namespace OrchardCore.Layers.Drivers
 {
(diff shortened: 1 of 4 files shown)


Skrypt:
What about Layers that are already published if we remove a Layer. Are they still trying to render? Maybe worth taking a look at that one too.

TFleury:
> What about Widgets that are already published if we remove a Layer. Are they still trying to render? Maybe worth taking a look at that one too.

At least it won't break your site 🤣

I'll take a look.

TFleury:
It seems to be handled : https://github.com/OrchardCMS/OrchardCore/blob/d599c7f348de5802b525b0676cf4bd0f1af99aee/src/OrchardCore.Modules/OrchardCore.Layers/Controllers/AdminController.cs#L224-L233

deanmarcussen (approved):
Technically you can still lose track of items in a layer, by removing the layer with the settings driver - there is no check there, it could be similar to what's in the admin controller.

But seperate issue

PR #8614, opened 2021-02-16
Code optimization on #8600

Skrypt:
Just some optimization for @sebastienros 

screencapture-localhost-5001-Admin-Lucene-Query-2021-02-16-16_27_17

Here the content item with Title 4 has an HtmlBodyPart with "Something" content.
All good.

the perf gain is not noticeable with that amount of records but that's probably better than using a .Contains().

And just to make sure here is another test : 

image


src/OrchardCore/OrchardCore.Lucene.Core/QueryProviders/SimpleQueryStringQueryProvider.cs
```diff
@@ -24,10 +23,12 @@ public Query CreateQuery(ILuceneQueryService builder, LuceneQueryContext context
 
             foreach(var field in fields)
             {
-                if(field.Contains('^', StringComparison.Ordinal) && Single.TryParse(field.Split("^").Last(), out weight))
+                var fieldWeightArray = field.Split("^");
+
+                if(field.Contains('^', StringComparison.Ordinal) && Single.TryParse(fieldWeightArray.Last(), out weight))
```
Skrypt, line 28: I assume that here I could only do a if fieldWeightArray.Length > 1 too but I feel like what I did is safer.

src/OrchardCore/OrchardCore.Lucene.Core/QueryProviders/SimpleQueryStringQueryProvider.cs
```diff
@@ -24,10 +23,12 @@ public Query CreateQuery(ILuceneQueryService builder, LuceneQueryContext context
 
             foreach(var field in fields)
             {
-                if(field.Contains('^', StringComparison.Ordinal) && Single.TryParse(field.Split("^").Last(), out weight))
+                var fieldWeightArray = field.Split("^");
```
sebastienros, line 26:
```suggestion
                var fieldWeightArray = field.Split("^", 2);
```

src/OrchardCore/OrchardCore.Lucene.Core/QueryProviders/SimpleQueryStringQueryProvider.cs
sebastienros: yes, use Length > 1 and `fieldWeightArray[1]`

Skrypt:
There it is @jtkech

Skrypt:
If anyone finds something else please tell me before someone merges this.

hishamco:
> If anyone finds something else please tell me before someone merges this.

I may have a look Jasmin, one more thing, is there a unit test(s) to validate this to avoid unexpected behaviors?

sebastienros (approved):

Skrypt:
No, there is no unit test or functional test ; but I created an issue about it. So it's up for grab.
I started trying to create a unit tests and analyzing how this could work.

src/OrchardCore/OrchardCore.Lucene.Core/QueryProviders/SimpleQueryStringQueryProvider.cs
```diff
@@ -24,10 +23,12 @@ public Query CreateQuery(ILuceneQueryService builder, LuceneQueryContext context
 
             foreach(var field in fields)
             {
-                if(field.Contains('^', StringComparison.Ordinal) && Single.TryParse(field.Split("^").Last(), out weight))
+                var fieldWeightArray = field.Split('^', 2);
```
jtkech, line 26: @Skrypt

Perfect 👍

PR #8615, opened 2021-02-17
ILocalClock in BackgroundTaskScheduler

### LINKED ISSUE #7518: Background tasks, time zones should be converted

Skrypt:
Fixes #7518 

WIP

### DIFF
--- a/src/OrchardCore/OrchardCore/BackgroundTasks/BackgroundTaskScheduler.cs
+++ b/src/OrchardCore/OrchardCore/BackgroundTasks/BackgroundTaskScheduler.cs
@@ -24,16 +24,14 @@ namespace OrchardCore.BackgroundTasks
         public DateTime ReferenceTime { get; set; }
         public BackgroundTaskSettings Settings { get; set; }
         public BackgroundTaskState State { get; set; }
-        public ITimeZone TimeZone { get; set; }
-
         public bool Released { get; set; }
         public bool Updated { get; set; }
 
         public bool CanRun()
         {
-            var referenceTimeLocal = _localClock.ConvertToLocalAsync(ReferenceTime).GetAwaiter().GetResult().DateTime;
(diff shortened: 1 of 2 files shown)


Skrypt:
Ok nevermind, I just got the logs issue by restarting the whole thing.

Skrypt:
So while debugging this I found that it's failing here : 

https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.Settings/Services/DefaultTimeZoneSelector.cs#L25

It cannot retrieve the appropriate TimeZone settings from the database or cache.

jtkech:
@Skrypt 

Yes this is exactly the issue i had and fixed, `ISiteService` is a singleton for some reasons but e.g. `GetSiteSettingsAsync()` needs to be executed in a scope context as it uses e.g. `ISession`. And the bg task schedulers are singleton components hold by the bg service, and `CanRun()` is not executed in a scope context, we do that to not have to create a scope context just to check if a shell has something to do.

Another approach would be to always create scopes periodically to check if the schedulers can run, but we plan to have the option to release the resources of a tenant when it has no activity (no http request), but if we create a scope periodically, it will be always active. On another side how to make it inactive if it has e.g. a timer bg task that needs to run every minutes, maybe another option to say if a tenant can be freed, meaning its bg tasks can be stopped until it is http requested again.

Skrypt:
I don't think that it would be appropriate to not execute a background task meant to be executed every minutes. Meaning that if a background task would mutate the data for the next time it's going to be displayed on a Http Request this would fail to achieve it's goal. 

Then, here, if we want to free a tenant because it's not been requested since a while, we would need to have background tasks that are not using tenant contexts. These background tasks are started per tenant context and use their tenant context to know where to apply the data mutations so... yeah ... 😄 

It's not impossible to achieve but it needs more work.

Issue #8611, opened 2021-02-16
Add DocumentId to AdminDashboard Index


deanmarcussen:
Because it snuck in without one @jtkech

jtkech:
@deanmarcussen 

Through another issue, i saw that a full non clustered index, so including `DocumentId`, may also prevent some dead lock on sqlite, so, even if there will never be a lot of dashboard widgets, yes good to add here a light index (few columns).

I agree

deanmarcussen:
Yes, interesting about the deadlock - saw that issue as well.

I assume it is because without an index, it is doing a table scan just for the join, which locks the table, whereas when it is index scanning, no need to lock. (guessing here, but interesting).

If I have time I will add it today. There is another issue that just came up on gitter as well, so will wait for that pr.

jtkech:
> I assume it is because without an index, it is doing a table scan just for the join, which locks the table, whereas when it is index scanning, no need to lock

I assume too, but i think so ;)

> so will wait for that pr.

No problem

PR #8624, opened 2021-02-17
Add DocumentId index to dashboard widgets

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/8611

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.AdminDashboard/Migrations.cs
+++ b/src/OrchardCore.Modules/OrchardCore.AdminDashboard/Migrations.cs
@@ -26,7 +26,7 @@ namespace OrchardCore.AdminDashboard
             );
 
             SchemaBuilder.AlterIndexTable<DashboardPartIndex>(table => table
-                .CreateIndex($"IDX_{nameof(DashboardPartIndex)}_{nameof(DashboardPartIndex.Position)}",
+                .CreateIndex("IDX_DashboardPart_DocumentId",
                     "DocumentId",
                     nameof(DashboardPartIndex.Position))
             );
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.AdminDashboard/Migrations.cs
```diff
@@ -37,7 +43,20 @@ public async Task<int> UpdateFrom1Async()
...
-            return 2;
+            // Shortcut other migration steps on new content definition schemas.
+            return 3;
         }
+
+        // This code can be removed in a later version.
+        public int UpdateFrom2()
+        {
+            SchemaBuilder.AlterIndexTable<DashboardPartIndex>(table => table
+                .CreateIndex($"IDX_{nameof(DashboardPartIndex)}_{nameof(DashboardPartIndex.Position)}",
```
jtkech, line 54: An index table has at least one `IDX_SomeIndex_DocumentId` index that may contain all columns, even if it is only one more column as here, so i would have named it `IDX_DashboardPartIndex_DocumentId`, as for layer

            SchemaBuilder.AlterIndexTable<LayerMetadataIndex>(table => table
                .CreateIndex("IDX_LayerMetadataIndex_DocumentId", "DocumentId", "Zone")
            );

Note: As suggested by @Skrypt, maybe in the future we could also reduce the names by removing `IDX_` and / or `Index`

Skrypt: I would keep the prefix "IDX_" and remove the "Index" word since it is redundant.

deanmarcussen: Combination of both suggestions I hope?

jtkech (approved):

PR #8629, opened 2021-02-18
Added error checking for decrypting facebook and twitter settings

flew2bits:
In the situation where a data protection key gets deleted or is otherwise unavailable, both the Twitter and Facebook settings screens only show a "Save" button. This update allows the settings to be edited, showing an error that some settings could not be decrypted.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Facebook/Drivers/FacebookSettingsDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Facebook/Drivers/FacebookSettingsDisplayDriver.cs
@@ -57,7 +57,8 @@ namespace OrchardCore.Facebook.Drivers
                     try
                     {
                         model.AppSecret = protector.Unprotect(settings.AppSecret);
-                    } catch (CryptographicException)
+                    }
+                    catch (CryptographicException)
                     {
                         model.AppSecret = string.Empty;
                         model.CouldNotDecryptSettings = true;
--- a/src/OrchardCore.Modules/OrchardCore.Facebook/Drivers/FacebookSettingsDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Facebook/Drivers/FacebookSettingsDisplayDriver.cs
@@ -61,7 +61,7 @@ namespace OrchardCore.Facebook.Drivers
                     catch (CryptographicException)
                     {
                         model.AppSecret = string.Empty;
-                        model.CouldNotDecryptSettings = true;
+                        model.HasDecryptionError = true;
                     }
                 }
             }).Location("Content:0").OnGroup(FacebookConstants.Features.Core);
--- a/src/OrchardCore.Modules/OrchardCore.Facebook/Drivers/FacebookSettingsDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Facebook/Drivers/FacebookSettingsDisplayDriver.cs
@@ -1,3 +1,4 @@
+using System.Security.Cryptography;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.DataProtection;
@@ -53,7 +54,14 @@ namespace OrchardCore.Facebook.Drivers
                 model.SdkJs = settings.SdkJs;
                 if (!string.IsNullOrWhiteSpace(settings.AppSecret))
                 {
-                    model.AppSecret = protector.Unprotect(settings.AppSecret);
+                    try
+                    {
+                        model.AppSecret = protector.Unprotect(settings.AppSecret);
+                    } catch (CryptographicException)
+                    {
+                        model.AppSecret = string.Empty;
+                        model.CouldNotDecryptSettings = true;
+                    }
                 }
             }).Location("Content:0").OnGroup(FacebookConstants.Features.Core);
         }
--- a/src/OrchardCore.Modules/OrchardCore.Facebook/Drivers/FacebookSettingsDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Facebook/Drivers/FacebookSettingsDisplayDriver.cs
@@ -3,6 +3,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.DataProtection;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
 using OrchardCore.DisplayManagement.Entities;
 using OrchardCore.DisplayManagement.Handlers;
 using OrchardCore.DisplayManagement.Views;
@@ -20,19 +21,23 @@ namespace OrchardCore.Facebook.Drivers
         private readonly IHttpContextAccessor _httpContextAccessor;
         private readonly IShellHost _shellHost;
         private readonly ShellSettings _shellSettings;
+        private readonly ILogger _logger;
 
         public FacebookSettingsDisplayDriver(
             IAuthorizationService authorizationService,
             IDataProtectionProvider dataProtectionProvider,
             IHttpContextAccessor httpContextAccessor,
             IShellHost shellHost,
-            ShellSettings shellSettings)
+            ShellSettings shellSettings,
+            ILogger<FacebookSettingsDisplayDriver> logger
+            )
         {
             _authorizationService = authorizationService;
             _dataProtectionProvider = dataProtectionProvider;
             _httpContextAccessor = httpContextAccessor;
             _shellHost = shellHost;
             _shellSettings = shellSettings;
+            _logger = logger;
         }
 
         public override async Task<IDisplayResult> EditAsync(FacebookSettings settings, BuildEditorContext context)
@@ -60,6 +65,7 @@ namespace OrchardCore.Facebook.Drivers
                     }
                     catch (CryptographicException)
                     {
+                        _logger.LogError("The app secret could not be decrypted. It may have been encrypted using a different key.");
                         model.AppSecret = string.Empty;
                         model.HasDecryptionError = true;
                     }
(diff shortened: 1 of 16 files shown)


src/OrchardCore.Modules/OrchardCore.Facebook/Drivers/FacebookSettingsDisplayDriver.cs
```diff
@@ -53,7 +54,14 @@ public override async Task<IDisplayResult> EditAsync(FacebookSettings settings,
                 model.SdkJs = settings.SdkJs;
                 if (!string.IsNullOrWhiteSpace(settings.AppSecret))
                 {
-                    model.AppSecret = protector.Unprotect(settings.AppSecret);
+                    try
+                    {
+                        model.AppSecret = protector.Unprotect(settings.AppSecret);
+                    } catch (CryptographicException)
```
deanmarcussen, line 60: newline

src/OrchardCore.Modules/OrchardCore.Facebook/Views/FacebookSettings.Edit.cshtml
```diff
@@ -6,6 +6,11 @@
     <a class="seedoc" href="@(OrchardCore.Admin.Constants.DocsUrl)reference/modules/Facebook/#core-components" target="_blank">@T["See documentation"]</a>
 </p>
 
+@if (Model.CouldNotDecryptSettings)
+{
+    <p class="alert alert-danger">@T["An error occurred while decrypting the Secret Key, so it has been left blank. Anything using these settings will not currently work."]</p>
```
deanmarcussen, line 11:
```suggestion
    <p class="alert alert-danger">@T["An error occurred while decrypting a setting, so the feature has been disabled. Please enter and save to enable this feature."]</p>
```

src/OrchardCore.Modules/OrchardCore.Twitter/Drivers/TwitterSettingsDisplayDriver.cs
```diff
@@ -41,30 +42,43 @@ public override async Task<IDisplayResult> EditAsync(TwitterSettings settings, B
...
-                }
-            }).Location("Content:5").OnGroup(TwitterConstants.Features.Twitter);
+                    model.APIKey = settings.ConsumerKey;
+                    if (!string.IsNullOrWhiteSpace(settings.ConsumerSecret))
+                    {
+                        try
+                        {
+                            var protector = _dataProtectionProvider.CreateProtector(TwitterConstants.Features.Twitter);
+                            model.APISecretKey = protector.Unprotect(settings.ConsumerSecret);
+                        } catch (CryptographicException)
```
deanmarcussen, line 54: newline

src/OrchardCore.Modules/OrchardCore.Twitter/Drivers/TwitterSettingsDisplayDriver.cs
```diff
@@ -41,30 +42,43 @@ public override async Task<IDisplayResult> EditAsync(TwitterSettings settings, B
...
+                        model.APISecretKey = string.Empty;
+                    }
+                    model.AccessToken = settings.AccessToken;
+                    if (!string.IsNullOrWhiteSpace(settings.AccessTokenSecret))
+                    {
+                        try
+                        {
+                            var protector = _dataProtectionProvider.CreateProtector(TwitterConstants.Features.Twitter);
+                            model.AccessTokenSecret = protector.Unprotect(settings.AccessTokenSecret);
+                        } catch (CryptographicException)
```
deanmarcussen, line 71: newline

src/OrchardCore.Modules/OrchardCore.Twitter/Views/TwitterSettings.Edit.cshtml
```diff
@@ -4,6 +4,11 @@
 <p class="alert alert-warning">@T["The current tenant will be reloaded when the settings are saved."]
     <a class="seedoc" href="@(OrchardCore.Admin.Constants.DocsUrl)reference/modules/Twitter/#twitter-integration" target="_blank">@T["See documentation"]</a></p>
 
+@if (Model.CouldNotDecryptSettings)
+{
+    <p class="alert alert-danger">@T["An error occurred while decrypting some settings, so they have been left blank. Anything using these settings will not currently work."]</p>
```
deanmarcussen, line 9:
```suggestion
    <p class="alert alert-danger">@T["An error occurred while decrypting a setting, so the feature has been disabled. Please enter and save to enable this feature."]</p>
```

deanmarcussen (commented):
Looks good.

I am just suggesting the text changes as a way to keep the text the same for both across all the translations - makes it easier to translate.

src/OrchardCore.Modules/OrchardCore.Facebook/Views/FacebookSettings.Edit.cshtml
flew2bits: Should it say that the feature is disabled, since that's an actual function of Orchard, or should it maybe say "...the feature is non-functional" or something similar instead?

deanmarcussen: Well generally the `Options` catches the exceptions, and just stops working, so is is close enough to a feature being disabled, i.e. it doesn't work. 
But you're right it's not technically a feature disabled. Non-functional is a bit too techy (for me) for the ui. But if you can come up with a better way of saying it please do.

Or just keep it shorter and don't say so much (it was just really long, and a lot of information)
`An error occurred while decrypting a setting. Please apply and save.`

Normally something is logged by the options class, so for more info users should be looking in logs - not that anyone ever remembers to check there ;)

src/OrchardCore.Modules/OrchardCore.Facebook/Drivers/FacebookSettingsDisplayDriver.cs
flew2bits: fixed

src/OrchardCore.Modules/OrchardCore.Facebook/Views/FacebookSettings.Edit.cshtml
flew2bits: I updated the text to the shorter message.

src/OrchardCore.Modules/OrchardCore.Twitter/Views/TwitterSettings.Edit.cshtml
flew2bits: Shortened the error message.

src/OrchardCore.Modules/OrchardCore.Twitter/Drivers/TwitterSettingsDisplayDriver.cs
flew2bits: fixed

src/OrchardCore.Modules/OrchardCore.Twitter/Drivers/TwitterSettingsDisplayDriver.cs
flew2bits: fixed

flew2bits:
I suspect that this issue may appear in other settings pages. I'm going to go through and check out other features like GitHub, Google, etc. There may also be additional places in Twitter and Facebook that need error checking as well.

flew2bits:
I added checks for decryption errors to the Google, Github, and Microsoft Authentication modules, as well.

src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/ViewModels/MicrosoftAccountSettingsViewModel.cs
```diff
@@ -14,5 +14,7 @@ public class MicrosoftAccountSettingsViewModel
         public string CallbackPath { get; set; }
 
         public bool SaveTokens { get; set; }
+
+        public bool CouldNotDecryptSettings { get; set; } = true;
```
deanmarcussen, line 18:
```suggestion
        public bool HasDecryptionError { get; set; }
```

Not sure why this one was set true by default?
Since you did it to all of them (great), I had to think about the appropriate name, for the now new pattern. 

Suggest `HasDecryptionError` might be better.

I think we should log the error at the time as well, as is done in the `Options` configuration classes - it may be that the error was on startup 3 weeks ago, but the user is only viewing now.

src/OrchardCore.Modules/OrchardCore.Twitter/ViewModels/TwitterSettingsViewModel.cs
```diff
@@ -15,5 +15,7 @@ public class TwitterSettingsViewModel
 
         [Required(AllowEmptyStrings = false, ErrorMessage = "Access token secret is required")]
         public string AccessTokenSecret { get; set; }
+
+        public bool CouldNotDecryptSettings { get; set; } = false;
```
deanmarcussen, line 19:
```suggestion
        public bool CouldNotDecryptSettings { get; set; }
```

And so no need to set false, when that is the default value of a bool

src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/ViewModels/MicrosoftAccountSettingsViewModel.cs
flew2bits: I honestly don't know why I defaulted this one to true, other than I got distracted by something. I've pushed updates that address all of these issues.

flew2bits:
The latest push removes the default values for the decryption error attribute; the attribute has been renamed to `HasDecryptionError`; and I've added logging for all of the decryption errors.

deanmarcussen (approved):
Thanks for tweaking it, and doing the other ones as well.

Issue #6318, opened 2020-06-01
Store tokens in Document table still is too much


jersiovic:
Currently the tokens and authorizations issued are stores at Content table.
As I mentioned in the past, this is too much https://github.com/OrchardCMS/OrchardCore/issues/233

In fact, in the issue mentioned it was solved with a background task that removes outdated tokens.
However I still think this can cause problems with a big site with a lot of grow which consumes the ids counter quickly. At least if you are going to have an app with many users, and all of them obtaining tokens. 
This, together with the fact that maybe your application store some types of documents that will be grow quickly it could lead too soon to be near fo the limits of an int.

So, my proposal is first to send a PR that  moves authorizations and tokens to another document table, using YesSQL Collection.
I've done it for our proyect in a private module, so I can do the same in a PR.

jersiovic:
Well, I will provide that PR if this other PR at YesSQL #241 is approved cause I need the feature added to YesSql for fixing this at Orchard.

In fact if that is done I see feasible in a future PR to add a feature that allows to make configurable the collections where entities of a module are stored

jersiovic:
@kevinchalet is it completly necessary to store all the tokens on db?
In my roadmap I wanted to work on a filter that allows to the server revoke any user issued for an user that has a token creation date before a date.
In that way I don't need to store access tokens in db to revoke them, And  I only need to store revokations for the days can last based on the days you give to its expiration date.
With the new approach for OpenIddict 3.0 you are forcing the server to access db on each request. With that you are killing one of the advantages of using tokens to check identity that is we don't need access db for that.
Doing what I propose you can store in cache revoked tokens while arrive their expiration date. And that cache is queried by the filter to decide if a token is valid.
It is not expected you have too much revocations so its acceptable they are on memory.

kevinchalet:
> @kevinchalet is it completly necessary to store all the tokens on db?

Nothing is 100% definitive. I can and may change the design based on feedback collected during beta1 😄 

> With the new approach for OpenIddict 3.0 you are forcing the server to access db on each request. With that you are killing one of the advantages of using tokens to check identity that is we don't need access db for that.

The **OpenIddict validation handler won't do any DB request by default**. Developers who want to check the token status for each request will have to enable it using `options.EnableTokenEntryValidation()`. A similar option named `options.EnableAuthorizationEntryValidation()` exists for authorization validation.

Also, when I mean token entry, it's only metadata in most cases: only reference tokens are stored with their payload in the database.

[...]

Issue #8422, opened 2021-01-25
Every modules can set a table name prefix (Collection)


chdq0306:
Every modules can set a "colection",like this
![image](https://user-images.githubusercontent.com/4231580/105658363-4a3c3d80-5f01-11eb-8701-42f8f127f268.png)

deanmarcussen:
Yes, it would be good to have an implementionat / or support for YesSql.Collections.

Maybe one way to handle it (and the transition, because we need to think of existing sites), is to register an `IOptions` class, which can provide `Collection` names.

It could be empty by default, i.e. no collection, and when a collection is registered use the collection name, for the module / feature.

The only dificulty is they would have to be configured before the site is setup (or manually rename the tables)

chdq0306:
How to use the Tokens?Where set? And other modules ？

chdq0306:
I have got it by myself with yessql collection;
OrchardCore is not only a cms or little saas ,but a powered saas now!

chdq0306:
已经解决了，谢谢大家，学到了很多东西。虽然不是完全按照OpenID Tokens的例子搞的，但是也学习了。

PR #8635, opened 2021-02-19
Add support for collections to OpenID Tokens

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/6318

And related to https://github.com/OrchardCMS/OrchardCore/issues/8422

So as @sebastienros quoted, 'Can't be simpler'

Sadly, it's never quite as simple as that. 

To be non breaking all the index migrations to create the index tables have to be run again as well.

So while it's doable, it's a messy migration, and not very configurable.
- Initialize new collection table.
- Indexes have to be created again. 
- Documents retrieved
- Stored in new table
- Removed from old table
- Old index tables dropped
- Hope the site still starts ;)

And what if you also want to store applications in that table as well (but we decided it was tokens in one, and applications in another). 
How do you say I want applications and tokens in that table? and content in this table? and users in another?

So counter proposal. 

An options class to register support for collections, which need to be registered before the site is setup, i.e. migrations run, (or you have to rename your index tables manually, which is not a big chore, but easier I think to push onto users who wish to use the feature).

The idea being that using this we can slowly begin adding support for collections to other features at will, and allow more of a transition.

To say you want tokens in a collection use you call `services.WithCollection<OpenIdToken>("Tokens");` 

Tested works well. 

An experimental proposal.

Might be other (more breaking) ways to achieve it, so all suggestions welcome.

/cc @kevinchalet @jersiovic @sebastienros 

Screenshot 2021-02-19 at 12 55 02


kevinchalet:
Looks very nice! 👍🏻 

Since we're gonna break things, should we also do that for the other collections? Authorizations would very likely benefit from that. For consistency, we may also want to do that for applications and scopes too.

deanmarcussen:
Well the idea was to not break things too much, by making it optin :)

If it gets the nod, then I'd do the rest of `OpenId` as part of this pr.

And start towards the other things that would benefit (we really want users in a different table, just to make management easier) in another pr.

deanmarcussen:
Discussed at the meeting.

Preference is to do a migration.

- Collection for every Model that we think may contain many many items.
  - Users
  - OpenId Tokens
  - Authorizations
  - Audit trail

Content Items may be difficult as it is quite breaking on any existing queries, whereas Users and OpenId are less likely to have custom queries.

Will probably approach OpenId first, in an effort to not do massive / lengthy migrations all at the same time (though no one can guarantee that users will update quickly, so ...)

The other thing to be careful of here, is that we don't start hitting the MySql / Postgres naming lengths of 64 characters, for tables / indexes etc. This becomes a little challenging as one cannot guarantee the length users will choice for their table prefix, so we do our best.

@kevinchalet would you see `Authorizations` and `Tokens` going into the same `Document` table, e.g.

- `OpenId_Document`
- `OpenId_OpenIdTokenIndex` (the naming convention is now is `Type` based so...)

or two different document tables
- `Tokens_Document`
- `Tokens_OpenIdTokenIndex`
- `Auths_Document`
- `Auths_OpenIdAuthorizationIndex` 

I probably see `Application` and `Scope` remaining in the normal document table

Skrypt:
Ok, for the length limitation, I think we should start thinking about a Helper method that will return the appropriate name. This way, it will be consistent for every database providers.

src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Migrations/OpenIdMigrations.cs
```diff
@@ -53,23 +66,26 @@ public int Create()
                 .Column<string>(nameof(OpenIdAuthorizationIndex.Status))
                 .Column<string>(nameof(OpenIdAuthorizationIndex.Subject))
                 .Column<string>(nameof(OpenIdAuthorizationIndex.Type))
-                .Column<DateTime>(nameof(OpenIdAuthorizationIndex.CreationDate)));
+                .Column<DateTime>(nameof(OpenIdAuthorizationIndex.CreationDate)),
+                collection: OpenIdAuthorizationCollection);
 
             SchemaBuilder.AlterIndexTable<OpenIdAuthorizationIndex>(table => table
-                .CreateIndex("IDX_OpenIdAuthorizationIndex_DocumentId_ApplicationId",
+                .CreateIndex("IDX_COL_OpenIdAuth_DocumentId_ApplicationId",
```
deanmarcussen, line 73: @jtkech @Skrypt 

So I invent a new convention here (always subject to change)
where I use
`IDX_COL` to indicate Index and Collection, and then an abbreviated name of the index class (to get some reduction in indexes, and avoid mysql / postrgres issues).

Needs to be different to the existing index names, because both must co-exist for a time during the migration.

deanmarcussen:
We end up with a full migration of both `Authorizations` and `Tokens` into a new `Collection` called `OpenId_Document`, and associated indexes.

All existing documents moved from the original table to the other and old indexes deleted, new indexes create etc.

Is it a little weird that the index names are `OpenId_OpenIdTokenIndex` ?

maybe, but they are type based now, so can't change that easily.

Screenshot 2021-03-11 at 14 40 07

deanmarcussen:
I decided to only do `Authorizations` and `Tokens` because they feel like the large ones, that you want to be able to manage seperately.

Open to `Applications` and `Scopes` moving in as well, if you prefer @kevinchalet ?

kevinchalet:
IMHO, each entity type should have its own table, but I guess it's not exactly how things are expected to work in YesSql, so a single table for all the OpenID entities is likely a good compromise 😄 

(with SQL databases now supporting native JSON columns and indexes on JSON values, I'm not so sure the "one table to rule them all" approach behind YesSql will really age well over time 😅)

sebastienros:
> with SQL databases now supporting native JSON columns

Maybe, just try to run a query on a table with such a column. I think you still need to define specific indexes so it materializes it for you.

kevinchalet:
> Maybe, just try to run a query on a table with such a column.

Just tested with SQL Server:

```sql
SELECT * FROM OpenIddictApplications
CROSS APPLY OPENJSON(RedirectUris)
WHERE value = 'urn:postman'
```

I haven't tested with MySQL/MariaDB, but IIRC, the syntax should be even simpler:

```sql
SELECT * FROM OpenIddictApplications
WHERE JSON_CONTAINS(RedirectUris, 'urn:postman');
```

Regarding indexes, both regular (using computed columns if necessary) and full text search indexes are supported:
  - https://techcommunity.microsoft.com/t5/sql-server/indexing-json-documents-in-sql-server-2016/ba-p/384456
  - https://techcommunity.microsoft.com/t5/sql-server/indexing-json-arrays-using-full-text-search-indexes/ba-p/384568

kevinchalet:
BTW, wasn't this PR merged a bit prematurely?

sebastienros:
We decided to merge it. Was there something to wait for?

About the SQL, maybe sqlite also supports it. I feel like I remember something about it.
But performance is horrible in the tests I did when using things like JSON_CONTAINS or JSON_VALUE? Are you using it somewhere? Need to study more ...

kevinchalet:
> Was there something to wait for?

We haven't decided yet whether we should do something similar for the application and scope entities (see https://github.com/OrchardCMS/OrchardCore/pull/8635#issuecomment-796787642)

> Are you using it somewhere?

I know an OpenIddict user successfully uses these functions in custom Dapper stores. I've heard the performance is good (as long as things are indexed) but since I haven't implemented that personally, I can't say how good or how bad it really is. I'll definitely give it a try, tho 😄

sebastienros:
We talked about it during a meeting with @deanmarcussen and decided to go with a single openid document table. Only based on the expected cardinality of the other entities.

That said, I am definitely not opposed to suggesting to have a document table per type. Keeping in mind that joins can only be done in a single collection, without requiring custom SQL statements.

JSON: I often do some queris on json in sql server, and it's slow .... but maybe it's because I am dumb, and I there is a custom type I should use here: https://github.com/dotnet/crank/blob/f4e7c22ed736dfc31c0f374e3e7fd7889f4b21c8/src/Microsoft.Crank.Controller/JobSerializer.cs#L54

deanmarcussen:
I don’t mind doing it applications and scopes but there seemed to be less need for that, as they are more likely to contain only a few documents each. 

Where this was about grouping the heavier usage items into their own table?

If you feel strongly about it, let me know n I’ll add it (easily done), but it felt like to me that doing that combines document sets where you might want to completely clear the auth and tokens but not touch apps. For example. Obviously they all have the clr type so could be filtered but that was my logic

kevinchalet:
Well, I'm a "consistency always wins" guy and it honestly seems weird/confusing/unexpected to have an `OpenId` table with half the OpenID entities 😄

If we keep them separate, you can be sure folks will end up posting threads to ask why they can't find the client definitions (or the scopes) in the `OpenId` table.

@sebastienros have you tried persisted computed columns? I guess that would be similar to what you have with YesSql, where indexed things are basically stored twice.

deanmarcussen:
I can see the consistency point.

let me have a look tomorrow, it is not a big change.

note to self: need to check the collection initialization is registered in all startups.

sebastienros:
I didn't realize we were missing some other entities. I agree we should either have on named table for tokens, or eveyrthing in an "OpenIDDocument" table. Thanks for the feedback.

src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Migrations/OpenIdMigrations.cs
jtkech: @deanmarcussen 

`IDX_COL_OpenIdAuth_DocumentId_ApplicationId` LGTM for a new and shorter name

Hmm, but now we know that we always include `DocumentId`, so another idea would be

`IDX_COL_OpenIdAuthorization_ApplicationId`

Hmm, maybe with a shorter standardized prefix for indexes + collection

`IDXC_OpenIdAuthorization_ApplicationId`

But anyway seems to be too late ;)

PR #8637, opened 2021-02-19
Add fieldTypeName to Query Schema for GraphQL.

### LINKED ISSUE #8631: SQL Query Schema with custom db table (collision with content type)
When you want to use a custom table in a SQL Query with a schema to expose your query in the GraphQL endpoint it can collide with an existing Content Type.

### To Reproduce
1 - Create a "Test" content type without any specific part or field.
2 - Create a "Test" table in your database with a "Name" column as a varchar.
3 - Create a "Test" SQL Query with an object schema
```json
{
    "type": "object",
    "properties": {  
        "Name" : {
            "type" : "string",
            "description" : "This is the name of the test."
        }
    }
}
```

```SQL
SELECT Name FROM Test
```

image

4 - Try to execute a GraphQL query on that Test query.

### Expected behavior
It should add both the Content Type and Query in the GraphQL endpoint without making them collide. Right now it will take in priority the Content Type over the Query leaving the Query unusable unless you rename it to something different than any of your content type names.

### Screenshots

image

Working solution for now is to rename the Query : 

image

Skrypt:
Fixes #8631 (partially)

This allows to be able to configure a custom Field Type Name for the GraphQL endpoint when creating a Query Schema.
Now we can use "fieldTypeName" in the schema like this : 

SQL Query example :

```json
{
    "type": "object",
    "fieldTypeName": "TestSqlQuery",
    "properties": {  
        "Name" : {
            "type" : "string",
            "description" : "This is the 'Name' colum of the test table."
        }
    }
}
```

orchard

This work with both Lucene, SQL and returning content items or not from the Query.


Skrypt:
Do not merge yet. More work for Lucene Queries.

sebastienros:
Not a fan of this solution. I'd rather see:
- nothing done, if not just a warning when two same named are found
- a number added automatically when the query is added
- a setting in the query to define a custom name

Otherwise this is breaking, not sure it solves all the issues. What if a sql and a lucene query have the same name? Or could these queries be in a specific container, like `queries.recentBlogPost`, which might be equivalent to `queriesRecentBlogPosts` you can say ;) But my preferred option is a custom name for graphql, or just do nothing and a warning to let users know this name can be an issue if it is the same as a content type. Or just prevent it in validation.

Skrypt:
I personally would prefer seeing these not colliding at all and have a naming convention for the sake of having this working without a glitch. We could totally use a container for these Queries by using a prefix instead. Though the naming convention of GraphQL types is using camelCase. Should we use a . or _ as a separator then?

Skrypt:
Well I did not work on this since I don't want to waste too much time on this. I think it should be a breaking change but let's say, not for 1.0 but for later on. I want to work on this if we decide that we make some cleanup on the names else, this is just some side work on validation which should be done on a different PR.

Skrypt:
I went with the option that was more natural to me. Allow to add a custom name for these GraphQL field types. Now, this doesn't fix the issue if we don't add some validation on the Queries when we create them. It should validate that these Queries have either a different name than existing content types or that the schema `fieldTypeName` has a different name.

Though, this is not fixing the issue because existing Queries will not automatically be fixed and nothing prevents a user to add a Content Type that would have the same name than a Query later on. So, this needs to be validated on both sides which to me adds some overhead that would not be necessary if these Queries we're using their own container.

So, here first fix to be backward compatible is to add the ability to at least rename these Queries fieldType name.

Skrypt:
I should also say that we could also add in these Schema "DeprecationReason" strings on the root Schema element and on each of it's fields. This way, someone could define by himself when a Schema or just some of it's fields are deprecated. The GraphiQL endpoint still allows you to see these deprecated FieldTypes.

Skrypt:
@sebastienros I will invest time on a different PR to make these Queries inside a container. I think, we could also do the same with Content Types eventually. I will introduce a DeprecatedReason settings on the Schema on this other PR.

Skrypt:
@carlwoodhouse While you are here review this one too 😄

carlwoodhouse:
> @carlwoodhouse While you are here review this one too 😄

Haha ! I will look at this one when I’m at a computer later :)

Skrypt:
@sebastienros merging or closing?

sebastienros:
I like this solution. It's non-breaking, right?

carlwoodhouse:
Not had chance to look through it all but I like it conceptually too

Issue #5536, opened 2020-02-12
Edit Content Part, Add New Field, Kills GraphQL


kef7:
Adding "OpenInNewWindow" Bool field to LinkMenuItemPart from Orchard Core Admin breaks GraphQL. (I believe the content part definition for LinkMenuItemPart came from The Agency theme.)

- Accessing /api/graphql gives 400 Bad Request

- GraphQL admin area shows "No Schema" message; can't execute queries

- IIS error in Windows Event Log: 
Category: Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware
EventId: 1
RequestId: 80000d14-0400-f900-b63f-84710c7967bb
RequestPath: /api/graphql
SpanId: |c69cdb0c-44c6865f99a2049c.
TraceId: c69cdb0c-44c6865f99a2049c
ParentId: 

An unhandled exception has occurred while executing the request.

Exception: 
System.ArgumentOutOfRangeException: A field with the name: linkMenuItem is already registered for GraphType: LinkMenuItem (Parameter 'Name')
   at GraphQL.Types.ComplexGraphType`1.AddField(FieldType fieldType)
   at OrchardCore.ContentManagement.GraphQL.Queries.Types.DynamicContentTypeBuilder.Build(FieldType contentQuery, ContentTypeDefinition contentTypeDefinition, ContentItemType contentItemType) in C:\projects\orchardcore\src\OrchardCore\OrchardCore.ContentManagement.GraphQL\Queries\Types\DynamicContentTypeBuilder.cs:line 65
   at OrchardCore.ContentManagement.GraphQL.Queries.ContentTypeQuery.BuildAsync(ISchema schema) in C:\projects\orchardcore\src\OrchardCore\OrchardCore.ContentManagement.GraphQL\Queries\ContentTypeQuery.cs:line 63
   at OrchardCore.Apis.GraphQL.Services.SchemaService.GetSchemaAsync() in C:\projects\orchardcore\src\OrchardCore.Modules\OrchardCore.Apis.GraphQL\Services\SchemaService.cs:line 56
   at OrchardCore.Apis.GraphQL.GraphQLMiddleware.ExecuteAsync(HttpContext context, ISchemaFactory schemaService) in C:\projects\orchardcore\src\OrchardCore.Modules\OrchardCore.Apis.GraphQL\GraphQLMiddleware.cs:line 79
[...]

PR #8669, opened 2021-02-25
Fix GraphQL dynamic field conflicts

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/5536

Caught me again today, so time to have a crack at it.

@carlwoodhouse not especially my area, but best I can tell it fixes it.

Can now add dynamic fields, to anything that has been preregistered by another builder.

### DIFF
--- a/src/OrchardCore/OrchardCore.ContentManagement.GraphQL/Queries/Types/DynamicContentTypeBuilder.cs
+++ b/src/OrchardCore/OrchardCore.ContentManagement.GraphQL/Queries/Types/DynamicContentTypeBuilder.cs
@@ -37,13 +37,16 @@ namespace OrchardCore.ContentManagement.GraphQL.Queries.Types
             {
                 var partName = part.Name;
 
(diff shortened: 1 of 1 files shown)


carlwoodhouse:
LGTM - might be an idea to supplement with some tests though ?

carlwoodhouse (commented):
LGTM - might be an idea to supplement with some tests though

Skrypt:
I don't think unit tests are worth it if we move to Hot Chocolate later on.

carlwoodhouse:
Maybe not ... but you’d still want to test that scenario; regardless of the implementation

carlwoodhouse:
I’m happy for it to go as it is btw; as in the current state it’s broken anyway !

Skrypt:
I agree that it needs unit test but let's merge it and create an issue to create a unit test.

Skrypt (approved):

carlwoodhouse (approved):

PR #8673, opened 2021-02-25
Fixing CI with net5.0

sebastienros:


Skrypt:
Should be all good.

jptissot (approved):

Skrypt (approved):
One step at a time LGTM

jptissot:
I wonder if we should modify the cypress test runner to be able to target both runtimes ? Maybe needs to be looked at.

agriffard:
Linux build is still in error though:

```
Successfully tagged ***/orchardcore-cms-linux:dev
Using default tag: latest
The push refers to repository [docker.io/***/orchardcore-cms-linux]
tag does not exist: ***/orchardcore-cms-linux:latest
Error: Process completed with exit code 1.
```
https://github.com/OrchardCMS/OrchardCore/runs/1982490625?check_suite_focus=true

Skrypt:
Here the "latest" means Orchard Core version RC2.
I'm not sure why it even tries to update it since we should be only updating the "dev" one.

Skrypt:
See here

image

RC2 and Latest have the same size.

sebastienros:
that's a good point, `latest` probably means rc2, not dev, if there is a `dev` tag. So it should just not be used.

PR #8695, opened 2021-02-28
Multi-targeting tweaks

jtkech:
Currently, building the web app under Visual Studio does a multi-targeting build, and executes the target that copies the translation files multiple times, that may happen concurently. Through the dotnet cmd line, it only does a singular TFM build, and the copy of translation files happens once but in an outer build where there are not really copied.

- Here we prevent the target that copies the translation files to be executed concurrently.

- For a given project, we define a singular TFM or a multi TFMs but not both, so that a build under Visual Studio or through the dotnet cmd line behaves the same, and the copy of translation files always happens, and only once.

- Here, for the web app we define a multi TFMs, not a singular TFM, because the web app is referenced by the tests project defining a multi TFMs. Anyway, under Visual Studio, even with a singular TFM for the web app, all the project dependencies defining a multi TFMs will be built on each TFM.

    **Update**: Idem for the mvc web app on which we may want to run func tests in all TFMs.

---

To be able to only have a singular TFM in a given project, without a multi TFMs, i removed the definition of a multi TFMs from the shared props file, i only kept its shared value. I would have preferred to put back these definitions in each project, even if they use shared values, e.g. at some point only the OpenId was using a multi TFMs, but there are too many modules / core projects, so i put them in the `Directory.Build.props` of their parent folder.


Skrypt:
These changes are fixing my own project which started failing to build after merging latest dev. Applying some of this PR changes on my project fixed it.

agriffard:
Can we find a place in the docs to explain the multi-targeting?

jtkech:
It is related to the aspnetcore doc describing how an app can specify a `TargetFamework` (Singular), or `TargetFameworks` (Plural) in a project file. Maybe just say somewhere in our doc that we support `net5.0` and `netcoreapp3.1` Frameworks

And a link to the aspnetcore doc https://docs.microsoft.com/en-us/dotnet/standard/frameworks

Skrypt (approved):

Skrypt:
@jtkech merge it sooner than later. This one fixes the issues that latest dev cannot be built from Visual Studio. I kind of wasted 2 hours last night on this so it's probably better to merge what we already have here to not make everyone waste too much time on this.

PR #8702, opened 2021-03-01
Try to fix docker linux push

jtkech:


Skrypt:
Merge it. It's an empiric one.

.github/workflows/release_ci.yml
```diff
@@ -69,7 +69,7 @@ jobs:
         rm -rf ./src/OrchardCore.Cms.Web/App_Data_Tests
         dotnet publish -c Release -o ./.build/release --no-build --no-restore --framework net5.0
         docker build -f Dockerfile-CI -t orchardproject/orchardcore-cms-linux:dev .
-        docker push orchardproject/orchardcore-cms-linux
+        docker push orchardproject/orchardcore-cms-linux:dev
```
Skrypt, line 72: Actually the issue might be that this will try always only update the `dev` tagged image while we would need to be able to eventually push an image for the `latest` tag.

Skrypt: The issue is maybe directly in the Docker hub website settings for our images.

jtkech:
@Skrypt Yes it's just empiric one for testing ;)

.github/workflows/release_ci.yml
jptissot: I am getting the same issue here: https://github.com/StatCan/StatCan.OrchardCore/runs/2007163352?check_suite_focus=true#step:7:113

jtkech:
@jptissot 

That's okay now by adding the :tag when pushing to docker

See in the commits page, it is green, it was red before but for another reason

It was failing on building, maybe a temporary problem, so i re-run the jobs

PR #8706, opened 2021-03-01
Cypress 6.6.0

agriffard:
https://github.com/cypress-io/cypress/releases/tag/v6.6.0
https://www.npmjs.com/package/cypress/v/6.6.0

### DIFF
--- a/.github/workflows/functional_all_db.yml
+++ b/.github/workflows/functional_all_db.yml
@@ -10,7 +10,7 @@ jobs:
     name: Functional Tests - MVC
     runs-on: ubuntu-latest
     container:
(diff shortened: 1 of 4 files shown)


jptissot:
Don't forget https://github.com/OrchardCMS/OrchardCore/blob/dev/.github/workflows/functional_all_db.yml#L13 

I think it's safe to skip some of these updates unless something big is added.

Skrypt:
Don't merge anything in dev please I'm doing tests on the Docker push pr.

Skrypt:
Ok it's fine now you can merge what you want. The build might not work yet though so you probably better merge these PR after we fixed the CI builds.

PR #8708, opened 2021-03-01
.NET SDK in GitHub Actions

agriffard:
Use SDK versions 3.1.406 and 5.0.103.

### DIFF
--- a/.github/workflows/functional_all_db.yml
+++ b/.github/workflows/functional_all_db.yml
@@ -20,7 +20,7 @@ jobs:
     - name: Setup .NET Core 5.0
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 5.0.102
+        dotnet-version: 5.0.103
     - name: Fix nuget restore issue
       run: |
           apt update && apt --only-upgrade install ca-certificates -y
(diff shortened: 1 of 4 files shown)


sebastienros:
We can use something like `dotnet-version: 5.0.*` already.
Being deterministic has it's advantages though, since there is less risk to be breaking. But a floating on the minor part I assume is fine.

agriffard:
SDK 5.0.200 has been released: https://github.com/dotnet/core/blob/main/release-notes/5.0/5.0.3/5.0.200-sdk.md

No changes of version for the Nuget packages, still 5.0.3. 
Ex: https://www.nuget.org/packages/Microsoft.AspNetCore.Authentication.AzureAD.UI

agriffard:
It managed to get the latest version:

```
Run actions/setup-dotnet@v1
  with:
    dotnet-version: 5.0.*
  env:
    DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
    DOTNET_CLI_TELEMETRY_OPTOUT: true
    DOTNET_ROOT: /home/runner/.dotnet
/home/runner/work/_actions/actions/setup-dotnet/v1/externals/install-dotnet.sh --version 5.0.200
```

PR #8717, opened 2021-03-02
At the solution level, build each TFM separately

jtkech:
There are some opened issues on the dotnet / msbuild repos about multi-targeting build, particularly when building at the solution level a clean solution, and when an application is referenced by another application as we do for unit tests.

We prevented to execute the target copying translation files concurrently when building a multi-targets application, but not when building concurrently applications using the same target, or referencing another app using this target.

Here, at the solution level we build each TFM separately, each step still building project dependencies in parallel.

Note: So i needed to define multi TFMs for all app projects.

### DIFF
--- a/.github/workflows/pr_ci.yml
+++ b/.github/workflows/pr_ci.yml
@@ -30,7 +30,8 @@ jobs:
         dotnet-version: 5.0.102
     - name: Build
       run: |
-        dotnet build --configuration Release
+        dotnet build --configuration Release --framework net5.0
+        dotnet build --configuration Release --framework netcoreapp3.1
     - name: Unit Tests
       run: |
         dotnet test -c Release --no-restore --no-build ./test/OrchardCore.Tests/OrchardCore.Tests.csproj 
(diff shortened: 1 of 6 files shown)


agriffard:
Now, the code generation templates defaults the Framework to `netcoreapp3.1`.

The `--framework` parameter with a f minus is still not working, even after adding the .template.config files.
We need to use: `dotnet new occms --Framework net5.0`

Tell us if we have to change the docs:
https://docs.orchardcore.net/en/dev/docs/getting-started/templates/#generate-an-orchard-cms-web-application

jtkech:
Oops sorry, i thought i was a lowercase as specified here https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-new

I let you try it and then yes update the doc accordingly

> Now, the code generation templates defaults the Framework to netcoreapp3.1.

To prevent race conditions, we now build the SOLUTION for each TFM separately, so all projects need to support both TFMs, including the template package. Before we were building the template package only for net5.0, so maybe that's why the default was this one.

Issue #8731, opened 2021-03-03
Spurious notification when editing current user about not being able to disable self


deanmarcussen:
Editing the current user now always sends a notification that you cannot disable the current user.

This is because the field is now disabled, so doesn't bind, so is always false.

Fix is simple, just need to check it first in the driver to see if it is currently allowed to disable the user being edited.

PR #8732, opened 2021-03-03
Fix for incorrect notification when editing user

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/8731

Regression from https://github.com/OrchardCMS/OrchardCore/pull/8117


Skrypt (approved):

Skrypt:
Using disabled form elements is not consistent from one browser to an other. I think we may just use a read-only class instead.

PR #8750, opened 2021-03-04
Media cache : Azure Blob Storage issue with 0 byte files

Skrypt:
Fixes partially #8715

In the MediaFileStoreResolverMiddleware, if the request is cancelled it should throw a 404 response and not move to any next middleware (ImageSharp). This is basically adding a new line in the log everytime we request an image that doesn't exist to get resized. This adds a line in the logs because ImageSharp can't get the file. (A bot could fill the logs by requesting a wrong image filename).

Do not throw when we have a request cancellation in the DefaultMediaFileStoreCacheFileProvider. (Same thing, here, it is filling the logs for nothing).

To repro the issue, just stop a page request manually when you are loading a page with a lot of images. It will log a lot of different things. This is cleaning most of them. Now, it doesn't fix the entire #8715 issue though, but it fixes issues that I came across while reviewing that code.

Still need to see if I can get 0 byte files from cancelled requests. Technically, we remove any files that gets a cancellationToken but for some reason these files did not get removed maybe because of the `if (File.Exists(cachePath)) ... then remove file` in the `catch`. I did not get any 0 byte files last night so the issue is intermittent.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Media/Services/MediaFileStoreResolverMiddleware.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Services/MediaFileStoreResolverMiddleware.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Net;
 using System.Collections.Concurrent;
 using System.Threading;
 using System.Threading.Tasks;
--- a/src/OrchardCore.Modules/OrchardCore.Media/Services/MediaFileStoreResolverMiddleware.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Services/MediaFileStoreResolverMiddleware.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Net;
 using System.Collections.Concurrent;
 using System.Threading;
 using System.Threading.Tasks;
@@ -107,7 +108,16 @@ namespace OrchardCore.Media.Services
             }, LazyThreadSafetyMode.ExecutionAndPublication)).Value;
 
             // Always call next, this middleware always passes.
-            await _next(context);
+            if (await _mediaFileStoreCache.IsCachedAsync(subPathValue))
+            { 
+                await _next(context);
+            }
+
+            if (!context.Response.HasStarted)
+            {
+                context.Response.StatusCode = (int)HttpStatusCode.NotFound;
+            }
+
             return;
         }
     }
--- a/src/OrchardCore.Modules/OrchardCore.Media/Services/MediaFileStoreResolverMiddleware.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Services/MediaFileStoreResolverMiddleware.cs
@@ -107,8 +107,8 @@ namespace OrchardCore.Media.Services
                 }
             }, LazyThreadSafetyMode.ExecutionAndPublication)).Value;
 
+            // Always call next, this middleware always passes.
             await _next(context);
-
             return;
         }
     }
--- a/src/OrchardCore.Modules/OrchardCore.Media/Services/MediaFileStoreResolverMiddleware.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Services/MediaFileStoreResolverMiddleware.cs
@@ -109,13 +109,8 @@ namespace OrchardCore.Media.Services
 
             // Always call next, this middleware always passes.
             if (await _mediaFileStoreCache.IsCachedAsync(subPathValue))
-            { 
-                await _next(context);
-            }
-
-            if (!context.Response.HasStarted)
             {
-                context.Response.StatusCode = (int)HttpStatusCode.NotFound;
+                await _next(context);
             }
 
             return;
--- a/src/OrchardCore.Modules/OrchardCore.Media/Services/MediaFileStoreResolverMiddleware.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Services/MediaFileStoreResolverMiddleware.cs
@@ -107,11 +107,7 @@ namespace OrchardCore.Media.Services
                 }
             }, LazyThreadSafetyMode.ExecutionAndPublication)).Value;
 
-            // Always call next, this middleware always passes.
-            if (await _mediaFileStoreCache.IsCachedAsync(subPathValue))
-            {
-                await _next(context);
-            }
+            await _next(context);
 
             return;
         }
--- a/src/OrchardCore/OrchardCore.Media.Core/DefaultMediaFileStoreCacheFileProvider.cs
+++ b/src/OrchardCore/OrchardCore.Media.Core/DefaultMediaFileStoreCacheFileProvider.cs
@@ -71,7 +71,6 @@ namespace OrchardCore.Media.Core
             }
             catch (Exception ex)
             {
-                _logger.LogError(ex, "Error saving file {Path}", cachePath);
                 if (File.Exists(cachePath))
                 {
                     try
@@ -84,7 +83,12 @@ namespace OrchardCore.Media.Core
                         throw;
                     }
                 }
-                throw;
+
+                if(!cancellationToken.IsCancellationRequested)
+                {
+                    _logger.LogError(ex, "Error saving file {Path}", cachePath);
+                    throw;
+                }
             }
         }
 
--- a/src/OrchardCore/OrchardCore.Media.Core/DefaultMediaFileStoreCacheFileProvider.cs
+++ b/src/OrchardCore/OrchardCore.Media.Core/DefaultMediaFileStoreCacheFileProvider.cs
@@ -67,7 +67,8 @@ namespace OrchardCore.Media.Core
                 using (var fileStream = File.Create(cachePath))
                 {
                     await stream.CopyToAsync(fileStream, StreamCopyBufferSize);
-
+                    await stream.FlushAsync();
+                    
                     if (fileStream.Length == 0)
                     {
                         throw new Exception($"Error retrieving file (length equals 0 byte) : {cachePath}");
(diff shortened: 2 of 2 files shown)


src/OrchardCore/OrchardCore.Media.Core/DefaultMediaFileStoreCacheFileProvider.cs
```diff
@@ -84,7 +83,12 @@ public async Task SetCacheAsync(Stream stream, IFileStoreEntry fileStoreEntry, C
                         throw;
                     }
                 }
-                throw;
+
+                if(!cancellationToken.IsCancellationRequested)
```
agriffard, line 87: Format, pls.

src/OrchardCore.Modules/OrchardCore.Media/Services/MediaFileStoreResolverMiddleware.cs
```diff
@@ -107,7 +108,11 @@ public async Task Invoke(HttpContext context)
             }, LazyThreadSafetyMode.ExecutionAndPublication)).Value;
 
             // Always call next, this middleware always passes.
-            await _next(context);
+            if (await _mediaFileStoreCache.IsCachedAsync(subPathValue))
```
deanmarcussen, line 111: I don't think we should do this, but fix the issue in ImageSharp for over logging.

Here the request would stall, because it doesn't call `_next`

src/OrchardCore/OrchardCore.Media.Core/DefaultMediaFileStoreCacheFileProvider.cs
```diff
@@ -67,6 +67,11 @@ public async Task SetCacheAsync(Stream stream, IFileStoreEntry fileStoreEntry, C
                 using (var fileStream = File.Create(cachePath))
                 {
                     await stream.CopyToAsync(fileStream, StreamCopyBufferSize, cancellationToken);
+
+                    if (fileStream.Length == 0)
+                    {
+                        throw new Exception($"Error retrieving file (length equals 0 byte) : {cachePath}");
```
deanmarcussen, line 73: This one cool, but also we talked about not giving the stream copy the request cancellationToken. 

So changing to `await stream.CopyToAsync(fileStream, StreamCopyBufferSize, default);` or the appropriate overload.

Let's see the logs you can get from your server with this in it though, so we can see the stack trace.

jtkech: > This one cool, but also we talked about not giving the stream copy the request cancellationToken.

Or, as we did at some point with @Skrypt, check in the following catch `!cancellationToken.IsCancellationRequested` before logging and throwing

                if (!cancellationToken.IsCancellationRequested)
                {
                    _logger.LogError(ex, "Error saving file {Path}", cachePath);
                    throw;
                }

Doing this was reducing logs, but when trying it with @Skrypt we didn't get any 0 length file, so maybe @Skrypt was able to repro and added the length check, but maybe worth to still keep and then check the cancelation token

src/OrchardCore.Modules/OrchardCore.Media/Services/MediaFileStoreResolverMiddleware.cs
jtkech: At some point with @Skrypt we were returning a 404 through the Response, after checking that the response has not yet started.

> Here the request would stall, because it doesn't call _next

For me it make sense to do this, it would just skip the following middleware including ISharp, returning back to the "parent" / calling ones.

deanmarcussen: Hmm, but we want the following middleware to be called, it maybe `ImageSharp` or it maybe `StaticFileMiddleware`, but we want them to be called, so that they can process the not found (and fix ImageSharp not to over log the not found).

Here we're just making a repeat of the `IsCachedAsync` that is already called earlier in the middleware (and managed with the concurrent dictionary)

So it depends what issue @Skrypt is trying to fix with this, whether it's a zero length issue, or another ?

Just need to understand the purpose behind it.

src/OrchardCore/OrchardCore.Media.Core/DefaultMediaFileStoreCacheFileProvider.cs
Skrypt: Well we want to keep throwing an exception and there could be an error saving the file without receiving a cancellationToken. We can't rely on the cancellationToken. So, I removed it and yes I need to remove the cancellationToken from the CopyToAsync because here we download the file in the cache. There is no reason to cancel the download of that file.

src/OrchardCore.Modules/OrchardCore.Media/Services/MediaFileStoreResolverMiddleware.cs
Skrypt: Ok I think we all agree this is a patch in the Middleware because of ImageSharp logging level throwing an exception when it doesn't find the file we want it to process. Here, I kept it because, this cause a security issue where someone could fill the logs of a website by requesting a url that would try to resize a file that doesn't exist. We may add a comment here in the mean time that this should be removed. We can't rely on the fact that this will be fixed or not in ImageSharp.

src/OrchardCore/OrchardCore.Media.Core/DefaultMediaFileStoreCacheFileProvider.cs
Skrypt: Done I removed the cancellationToken. And now it will only catch the issue where there's a 0 byte fileStream. We just want to track it for now.

src/OrchardCore.Modules/OrchardCore.Media/Services/MediaFileStoreResolverMiddleware.cs
jtkech: @deanmarcussen 

I think it's okay as we know here that it is a media file, otherwise we call the next middleware above line 60.

Unless maybe for a "media" file that is not intended to be processed by IS, and, as i remember, to be served by a subsequent custom static file provider that anyway will not find it, in fact the `DefaultMediaFileStoreCacheFileProvider` itself that is an `IMediaFileProvider` being also an `IStaticFileProvider` to serve "media" files coming from a blob and physically cached locally, but not intended to be processed by IS , but i may be wrong, as i remember.

> Here we're just making a repeat of the IsCachedAsync that is already called earlier in the middleware (and managed with the concurrent dictionary)

But the file may have been deleted right after, and when the file has been cached we don't reach again this code path.

> I'm so annoying I always want to understand why!

me too ;)

@Skrypt 

> Because, right now, we can't rely on the fact that this will be fixed or not in ImageSharp.

Yes, for now better to do something

src/OrchardCore/OrchardCore.Media.Core/DefaultMediaFileStoreCacheFileProvider.cs
jtkech: @Skrypt Okay, no problem, will see

One question, did you be able to repro the 0 byte issue and then reach the line where you check the length?

That would explain that the file would not be deleted (in the catch)

Then about the token it will depend if we want to stop or not the download on a request aborted, if we want to stop the download, good to check again in the catch block if the token has been canceled to not throw.

Skrypt: Yeah, we are trying things. I need to move onto something else for the remaining of the week. I might have time to test this back next week. For nice, I keep it on ice. I think we have what we need there. The 0 byte issue is hard to repro ; I need to spam my own website to make it queue up tasks a lot before reaching it.

deanmarcussen: Yep, this bit looks good for trying things, or at least seeing if we can repro it at all.

src/OrchardCore.Modules/OrchardCore.Media/Services/MediaFileStoreResolverMiddleware.cs
deanmarcussen: Ok, so if we're trying things, i.e. some extra logging, I'd rather not make other changes at the same time.

James has already agreed to the suggestion to reduce the logging, so it's just a matter of doing that being done.

I don't want to introduce differences here, between media used with blob storage vs media used without blob storage.

If we feel there's a need to address a potential logging issue, we can either do that in ImageSharp, or it would need to be a different piece of middleware that handles the issue for both normal storage and blob storage (much prefer just to handle it in ImageSharp)

jtkech: Okay, not enough focused on this PR, so just some feedbacks related to what a given comment was talking about at a given time. That said, not following the pipeline is not incompatible with the 0 length checking that is done before and anyway will throw. But okay, for testing maybe better to just add the 0 lentgh checking first.

> I don't want to introduce differences here, between media used with blob storage vs media used without blob storage.

Okay, for a local storage we would need another middleware just to check if a given media file exists, funny ;)

Skrypt:
Ok waiting on ImageSharp.Web update which I will add here and remove the lines in the Middleware.

src/OrchardCore/OrchardCore.Media.Core/DefaultMediaFileStoreCacheFileProvider.cs
```diff
@@ -66,7 +66,12 @@ public async Task SetCacheAsync(Stream stream, IFileStoreEntry fileStoreEntry, C
                 }
                 using (var fileStream = File.Create(cachePath))
                 {
-                    await stream.CopyToAsync(fileStream, StreamCopyBufferSize, cancellationToken);
+                    await stream.CopyToAsync(fileStream, StreamCopyBufferSize);
+
```
deanmarcussen, line 70: can you add a `await stream.FlushAsync()` here as well?

Because we realised the dispose would not be the async

PR #8758, opened 2021-03-05
Fix gitignore for project template files

deanmarcussen:
This removes the `.template.config` files from git, so that the new `.gitignore` change will actually ignore them.

They are created locally on each build from the `.template.config.src`, but the values change, i.e. get templated, based on the framework selected.

So having them in git means building locally keeps changing the files, to the current framework etc.

Also reported on gitter and I saw it a couple of days ago, but can't repro anymore, might have been causing `Error MSB4044 The "ReplaceFileText" task was not given a value for the required parameter "ReplacementText". OrchardCore.ProjectTemplates` to occur.

/cc @agriffard might be worth pulling this and validating that templates can still be built, and that I've haven't broken it in some other way, that I can't see on my local machine.


jtkech:
Yes, maybe some files have been committed before updating .gitignore in a first step

Issue #8763, opened 2021-03-05
support for SQL window functions


lampersky:
It would be good to have support for SQL window functions.

I know window functions are not the part of **SQL92 syntax**, but they are quite useful and available in all modern database engines. They are part of [SQL:2003](https://en.wikipedia.org/wiki/SQL:2003) standard.

I've extended **SQL Parser** and **SQL Grammar** to support **OVER** clause with **PARTITION BY** and **ORDER BY** so now you should be able to perform queries like:

```
SELECT
    COUNT(1) OVER () as col1,
    COUNT(1) OVER (PARTITION BY Type) as col2,
    COUNT(1) OVER (PARTITION BY Type ORDER BY Type) as col3,
    COUNT(1) OVER (PARTITION BY Type, Id ORDER BY Type, Id DESC) as col4
FROM Document;
```
_This is just a fake query, which doesn't make much sense._

Once my other [PR](https://github.com/OrchardCMS/OrchardCore/pull/8557) is merged you will also be able to use other window functions (argumentless) like:
- ROW_NUMBER(),
- RANK(),
- DENSE_RANK(),
- and more...

https://github.com/lampersky/OrchardCore/commit/dd293e56de25656f51b6c949686d73b5e0b245b5

I've tested window functions with orchard core and those databases:
- **PostgreSQL 13.2** (Debian 13.2-1.pgdg100+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit,
- **MySQL 8.0.23** (MySQL Community Server - GPL), Linux (x86_64),
- **SQL Server 12.0.2000.8** and **SQL Server 15.0.4102.2**,
- **SQLite 3.33.0**.

![mssql](https://user-images.githubusercontent.com/6403130/110154877-16e4ac80-7de5-11eb-9b9f-df027baefa39.png)
![mysql](https://user-images.githubusercontent.com/6403130/110154900-1ba96080-7de5-11eb-89b5-f747cdcc40fa.png)
![postgres](https://user-images.githubusercontent.com/6403130/110154903-1cda8d80-7de5-11eb-86bf-f020d50b3a2c.png)
![sqlite](https://user-images.githubusercontent.com/6403130/110154904-1cda8d80-7de5-11eb-8876-82efc5ec0b20.png)

Skrypt:
Maybe then what we need is to start a new parser that would be following the SQL:2003 standards.

PR #8764, opened 2021-03-05
window functions

lampersky:
support for SQL window functions

fixes https://github.com/OrchardCMS/OrchardCore/issues/8763


Skrypt:
If this parser should be SQL92 compliant then we need a new parser and allow to use one or the other. Now, maybe we could also inherit the SQL92 parser on the new one. But definitely we should not mix these if we say it should be SQL92 compliant it should be.

sebastienros (approved):

Issue #8753, opened 2021-03-04
Test migration from RC2 to 1.0


sebastienros:
And document steps in the release notes.

deanmarcussen:
[yessql.db.zip](https://github.com/OrchardCMS/OrchardCore/files/6096904/yessql.db.zip)

As a reference point here is an RC2 database with the blog and workflows setup.

username `admin` password `Password01_`

Now to see how the upgrade goes ;)

deanmarcussen:
Issues to resolve so far

- Fails to validate cookie when logged in, until `UserId` migration has run. Solution. Log Out first.
- `Layers` migration errors
- `UserPickerField` migration 

`while running migration version 5 for 'OrchardCore.Users'. Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 1: 'no such table: UserPickerFieldIndex'.`

deanmarcussen:
All fixed currently.

Keeping this open to retest prior to any further migration changes.

sebastienros:
TODO: Add this discussion https://github.com/OrchardCMS/OrchardCore/discussions/9119 to the RC2 release notes

agriffard:
Version 1.0.0 released.

PR #8776, opened 2021-03-07
Fix migrations from RC2 to latest dev

deanmarcussen:
Related to https://github.com/OrchardCMS/OrchardCore/issues/8753

More notes to come need to test something else, that might be problematic.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Startup.cs
@@ -239,7 +239,7 @@ namespace OrchardCore.ContentFields
         }
     }
 
-    [RequireFeatures("OrchardCore.Users", "OrchardCore.ContentFields.Indexing.SQL")]
+    [Feature("OrchardCore.ContentFields.Indexing.SQL.UserPicker")]
     public class UserPickerSqlIndexingStartup : StartupBase
     {
         public override void ConfigureServices(IServiceCollection services)
--- a/src/OrchardCore.Modules/OrchardCore.Layers/Migrations.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Layers/Migrations.cs
@@ -1,8 +1,10 @@
 using System;
 using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
 using OrchardCore.Data.Migration;
 using OrchardCore.Environment.Extensions;
 using OrchardCore.Environment.Shell;
+using OrchardCore.Environment.Shell.Scope;
 using OrchardCore.Layers.Indexes;
 using OrchardCore.Layers.Services;
 using OrchardCore.Rules;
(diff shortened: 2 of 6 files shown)


src/OrchardCore.Modules/OrchardCore.ContentFields/Startup.cs
```diff
@@ -239,7 +239,7 @@ public override void Configure(IApplicationBuilder builder, IEndpointRouteBuilde
         }
     }
 
-    [RequireFeatures("OrchardCore.Users", "OrchardCore.ContentFields.Indexing.SQL")]
+    [Feature("OrchardCore.ContentFields.Indexing.SQL.UserPicker")]
```
deanmarcussen, line 242: This has to move to a feature, or the migrationdependencies are not ordered correctly.

However moving it to a feature has no impact on the existing migrations, as they are recorded by class type not feature type.

src/OrchardCore.Modules/OrchardCore.Layers/Migrations.cs
```diff
@@ -64,18 +60,25 @@ public async Task<int> UpdateFrom2Async()
...
+
+            // Registered as a deferred task so the migration can complete before the shell reactivates the module causing migrations to run circularly and fail.                 
+            ShellScope.RegisterBeforeDispose((scope) =>
+            {
+                var typeFeatureProvider = scope.ServiceProvider.GetRequiredService<ITypeFeatureProvider>();
+                var layerFeature = typeFeatureProvider.GetFeatureForDependency(GetType());
+
+                // Reenable the layer feature so the new dependency of OrchardCore.Rules is activated.
+                var shellFeaturesManager = scope.ServiceProvider.GetRequiredService<IShellFeaturesManager>();
+                return shellFeaturesManager.EnableFeaturesAsync(new[] { layerFeature }, force: true);
```
deanmarcussen, line 80: @jtkech this was the slightly problematic one.

On an RC2 site `UpdateFrom1()` has not run.
This is `UpdateFrom2()`.

When `UpdateFrom2()` starts running, it activates the feature, which starts a feedback loop, back into the data migrations, and tries to execute `UpdateFrom1()` again. Which fails, because the indexes have already been created.

Here, I register the feature reactivation as deferred, so that all migrations have run, and the migration record comitted to the db, so the next shell that starts, is reading from the latest migration record, and doesn't try to rerun the migration.

Quickly done, might be better ways to achieve it.

deanmarcussen (commented):
General notes.

You need to be logged out before running the upgrade from RC2 to dev, or the cookie validator may try to revalidate an old cookie with the int `UserId`. This will throw exceptions from the `UserStore` about the user id indexes not being available, because the migration has not yet run. cookie validation happening before hand.

The solution is reasonably simple. Logout first.

@agriffard I'll try and get this written up for some release notes.

src/OrchardCore.Modules/OrchardCore.Layers/Migrations.cs
jtkech: @deanmarcussen Okay, LGTM

For some syncing reasons `UpdateShellDescriptorAsync()` does a `CommitAsync()`, so before the end of migrations, so anyway i prefer to defer it and if possible in its own scope. So, maybe better to use `ShellScope.AddDeferredTask()`, it will do the same but in its own scope.

deanmarcussen: `AddDeferredTask` done. Probably what I was looking for anyway.

jtkech: Does it work? I assume yes ;)

Forgot to say that maybe better to define before `var type = GetType()` and use it in the delegate, this to prevent a closure holding a reference on the migrations scoped service, that will be used later in the context of another scope.

We could also move up both line 75 and 76, hmm but you would need to re-inject `ITypeFeatureProvider`

            var layerFeature = _typeFeatureProvider.GetFeatureForDependency(GetType());

            ShellScope.AddDeferredTask((scope) =>
            {
                // Reenable the layer feature so the new dependency of OrchardCore.Rules is activated.
                var shellFeaturesManager = scope.ServiceProvider.GetRequiredService<IShellFeaturesManager>();
                return shellFeaturesManager.EnableFeaturesAsync(new[] { layerFeature }, force: true);
            });

PR #8782, opened 2021-03-07
use AddHostedService

infofromca:
fix #8781


src/OrchardCore/OrchardCore/Modules/Extensions/OrchardCoreBuilderExtensions.cs
```diff
@@ -93,7 +93,7 @@ public static OrchardCoreBuilder WithFeatures(this OrchardCoreBuilder builder, p
         /// </summary>
         public static OrchardCoreBuilder AddBackgroundService(this OrchardCoreBuilder builder)
         {
-            builder.ApplicationServices.AddSingleton<IHostedService, ModularBackgroundService>();
+            builder.ApplicationServices.AddHostedService<ModularBackgroundService>();
```
jtkech, line 96: We have another one under `OrchardCore/Shell/ServiceCollectionExtensions.cs` line 35 that we could also change ;)

jtkech (approved):

Issue #8785, opened 2021-03-08
Enabling AdminDashboard feature with other SQL modifying features causes SQL mods to fail 


lazcool:
If the AdminDashboard feature is enabled, along with other features that modify the SQL database as part of their migration, then the SQL db isn't modified.

The easiest way to reproduce this, is to add the AdminDashboard feature to the start-up recipe.

deanmarcussen:
I would image there must be a exception going into the logs here.

Can you check the related logs @lazcool ?

What I have seen happening with these recipe migrations, that create and save content items, is that other features, which also created content item index, need to be prioritized correctly, so that the indexes they are creating, have been created, byt the time a `FlushAsync` is called (causing the db, to call into index providers, which are registered by said feature).

This is where the `Feature` attribute on migrations is important, to get them going to the feature they are linked to. and/or the correct Module / Feature depenencies.

A log will show this though

andycyberian:
I've stripped out our custom modules from our project and have tried running a fresh setup inlcuding the Admin Dashboard in the recipe, these are the Features being added below:

`    "steps": [
        {
            "name": "feature",
            "disable": [],
            "enable": [
                // SaaS
                "OrchardCore.HomeRoute",
                "OrchardCore.Admin",
                "OrchardCore.Diagnostics",
                "OrchardCore.DynamicCache",
                "OrchardCore.Features",
                "OrchardCore.Localization",
                "OrchardCore.Navigation",
                "OrchardCore.Recipes",
                "OrchardCore.Resources",
                "OrchardCore.Roles",
                "OrchardCore.Settings",
                "OrchardCore.Themes",
                "OrchardCore.Users",
                // Content Management
                "OrchardCore.Alias",
                "OrchardCore.Autoroute",
                "OrchardCore.Html",
                "OrchardCore.Https",
                "OrchardCore.ContentFields",
                "OrchardCore.ContentPreview",
                "OrchardCore.Contents",
                "OrchardCore.ContentTypes",
                "OrchardCore.CustomSettings",
                "OrchardCore.Deployment",
                "OrchardCore.Deployment.Remote",
                "OrchardCore.Feeds",
                "OrchardCore.Flows",
                "OrchardCore.Indexing",
                "OrchardCore.Layers",
                "OrchardCore.Lucene",
                "OrchardCore.Lists",
                "OrchardCore.Liquid",
                "OrchardCore.Markdown",
                "OrchardCore.Media",
[...]

PR #8786, opened 2021-03-08
Migration steps convention

jtkech:


deanmarcussen:
Is this one related to https://github.com/OrchardCMS/OrchardCore/issues/8785 @jtkech - not sure if you would have seen it yet?

jtkech:
Yes i saw #8785 so just tried it but was not able to repro, i will comment it in the issue

So here just to use the migration steps conventions

Issue #8093, opened 2020-12-19
Admin: Display buttons at the top in sticky zone on Create/Edit


giannik:
Hey guys,
Looking at the new UI features in the admin I would like to make a suggestion.
I think when viewing a content item in edit mode or dispaly mode (displaytype =DetailAdmin) the actions zone (with buttons for save, publish etc) should be at the top in a sticky zone just under the main Nav bar and always be visible regardless of scrolling.
@Skrypt @agriffard .

agriffard:
See https://github.com/OrchardCMS/OrchardCore/pull/8067#issuecomment-748202190

@giannik Please keep your titles concised and descriptive ;-)
You can also link to existing issue/PR with their #number.

hishamco:
@giannik we tried the suggestion in edit template and will be demo on Tuesday if all agrees we can generalize this in the create and edit pages that have a long content

hishamco:
Related to #8067

giannik:
Great, Did'nt see it was already in the agenda.

agriffard:
You can keep it open. 
@hishamco 's PR is a first try on the Edit Template that really needs it.

sebastienros:
duplicate of https://github.com/OrchardCMS/OrchardCore/issues/2255

PR #8797, opened 2021-03-09
Sticky Action buttons on Create/Edit content

### LINKED ISSUE #2255: Fix the location of "Publish" "Save Draft"
In template editing, content editing, etc.
Fix the position of buttons such as "Publish" "Save Draft",
Will it be better?

agriffard:
On Create/Edit content, put the actions buttons at the top in a sticky toolbar.

media

Fixes #2255

Related to #8093


Skrypt:
Have you tested in the Menu and in the Widgets to make sure it works fine?

Skrypt:
orchard

Here is a reminder for myself on what Github does with their buttons.
Might add more screens for getting ideas.

deanmarcussen:
And for more views here is an example from devops...

Where the TopBar is a breadcrumb, rather than a title, and the title is the title.

Now I know we don't have a breadcrumb (yet), but that makes sense to me that the top bar is reserved for navigational purposes (one short term solution could be to render the title in both (as a psuedo breadcrumb in the lighter / smaller font), and later implement an actual breadcrumb. A crumb would be useful)

Screenshot 2021-03-10 at 08 27 28

agriffard:
Title and Actions buttons in the same toolbar:

image

/cc @deanmarcussen

agriffard:
Templates: Title and Actions buttons in the same toolbar:

image

hishamco:
What's the status of this?

agriffard:
> What's the status of this?

I implemented this sticky toolbar (with title on the left and actions on the right) in the Create/Edit Content and Create/Edit Template pages.

Are there any other admin pages where it could be useful?

hishamco:
AFAIK the templates pages, but this could be helpful in queries, where a lot of code can scroll the page

Issue #7868, opened 2020-12-05
Admin Users


agriffard:
Add filters in admin users page like:
- Only enabled.
- Only not email confirmed.

hishamco:
Filter by: 

- Activity (Enable, Disable, All)
 
- Email Confirmation (Confirmed, Not Confirmed, All)

- Role

I can work on this if you are not started yet

agriffard:
@hishamco Do you still want to work on this feature?

hishamco:
Sure if no one pick it up ;)

PR #8800, opened 2021-03-09
Custom user filters

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/7868

Extends the users admin list, in a similar way to the contents list, to provide filtering, and extensible filters (i.e. add your own, based on your own user profile data / indexes you've created).

TODO: Some clean up of old / never implemented filters


src/OrchardCore/OrchardCore.Users.Core/OrchardCore.Users.Core.csproj
```diff
@@ -15,6 +15,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\OrchardCore.DisplayManagement.Abstractions\OrchardCore.DisplayManagement.Abstractions.csproj" />
```
deanmarcussen, line 18: Need this reference to have `IUpdateModel` 

Need `IUsersAdminListFilter` in the Core project because `User` is in the core project.

Wonder about moving `User` to the abstractions project, to avoid this `DisplayManagement.Abstractions` reference.

src/OrchardCore.Modules/OrchardCore.Users/Controllers/AdminController.cs
```diff
@@ -154,46 +133,71 @@ public async Task<ActionResult> Index(UserIndexOptions options, PagerParameters
...
-                Users = userEntries,
-                Options = options,
-                Pager = pagerShape
-            };
-
-            model.Options.UserFilters = new List<SelectListItem>() {
-                new SelectListItem() { Text = S["All"], Value = nameof(UsersFilter.All) },
+                new SelectListItem() { Text = S["Enabled Users"], Value = nameof(UsersFilter.Enabled) },
+                new SelectListItem() { Text = S["Disabled Users"], Value = nameof(UsersFilter.Disabled) },
+                new SelectListItem() { Text = S["All Users"], Value = nameof(UsersFilter.All) },
```
agriffard, line 140: Suggestion: Move `All Users` at the top of the list.

deanmarcussen:
Moved to including `Authenticated` in the roles index (as the equivalent of `No roles`).

No migration for this, updated when editing user (don't want to do yet another user migration for it)

UI now shows `Authenticated (no roles)` as the text for this filter.

Leaving commented filters (never implemented) in the code base, maybe someone will take it as a hint to see about considering implementations for them

PR #8810, opened 2021-03-10
Add referer property on Liquid Request object

### LINKED ISSUE #8808: In workflows redirect task need  to redirect to the referer url
I am triggering a workflow (using a post incoming request event) from a page and when it completes 
I wish to return to the page it was originally called using the redirect task. 
So there should be a liquid helper on the Request object to return the referer inside the redirect task as : {{Request.Referer}}

giannik:
fixes #8808

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/RequestLiquidTemplateEventHandler.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/RequestLiquidTemplateEventHandler.cs
@@ -40,6 +40,7 @@ namespace OrchardCore.DisplayManagement.Liquid
                     case "Scheme": return new StringValue(request.Scheme);
                     case "Method": return new StringValue(request.Method);
                     case "Route": return new ObjectValue(new RouteValueDictionaryWrapper(request.RouteValues));
+                    case "Referer": return new StringValue(request.GetTypedHeaders().Referer.PathAndQuery);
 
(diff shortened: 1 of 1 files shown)


giannik:
I am having some issues using this new property from redirect task becuase of the urlencoder thats used in redirecttask.
It only works if i pass in null  as the encoder  when calling : 
```
  var location = await _expressionEvaluator.EvaluateAsync(Location, workflowContext, _urlEncoder);
```
 
On the other hand if i use in liquid event handler the following  : 

```
case "Referer": return  new  HtmlContentValue(new HtmlString(request.GetTypedHeaders().Referer.PathAndQuery)) ;
``` 

I get the exception : 

```
 System.InvalidCastException: Unable to cast object of type 'System.Text.Encodings.Web.DefaultUrlEncoder' to type 'System.Text.Encodings.Web.HtmlEncoder'.
```
so i am wondering where the fix should be ?


@sebastienros

sebastienros:
The PR needs to be updated since we merged the Fluid2 branch

sebastienros:
Better create a new PR. Question though, can't it already be accessed using the `Headers` property that is declared a few lines above? If that's the case there we shouldn't add it, otherwise it's opening the door to adding every single known ones.

giannik:
Hey , I just tried with the new fluid 2.0 branch and this works :  {{Request.Headers.Referer}}
So I'll close this branch.

PR #8815, opened 2021-03-10
Bump OpenIddict to 3.0.2

kevinchalet:
https://github.com/openiddict/openiddict-core/releases/tag/3.0.2


kevinchalet:
I'm not sure I understand what's the point of listing the versions (specially patch versions). Can't you just point folks to the .props file? Or at least, just indicate the major versions to reduce the burden.

agriffard:
This page is not just for the .NET dependencies.
The main goal was to list the different licenses.
There are also the client libraries, some packages and libraries for Test, ...

In my opinion, it seems a good habit to keep it maintained when a version is updated but you can do as you want.

kevinchalet:
I'm not opposed to it - it's just that it makes the workflow more complicated - you can't just click on "edit the .props" file on GitHub to send a PR in a few clicks and consider it done - for what seems to me a very little gain 🤷🏻

PR #8816, opened 2021-03-10
Admin UI : fix filterable list-group borders

### LINKED ISSUE #7960: Filters: Top border disappear on search

Skrypt:
Fixes #7960

There is no CSS selector for `:not([style*="display: none"]):first-child` so we need to do it with javascript. We can't use the `:first-child` sub-selector at least.


src/OrchardCore.Modules/OrchardCore.ContentTypes/Views/Admin/ListParts.cshtml
```diff
@@ -69,5 +79,4 @@
             }
         });
     });
-    //]]>
-</script>
+</script>
```
agriffard, line 82: newline

Skrypt: done

agriffard (approved):

Skrypt:
CI failing to do Unit Tests : 

Error Message:
   System.IO.IOException : The process cannot access the file 'C:\Users\runneradmin\AppData\Local\Temp\dfid3wjr.fzf' because it is being used by another process.

jtkech:
@Skrypt 

I already saw this issue that happens intermitently, seems to be related to the sqlite file that we can't delete as the sqlite file locking is not (at least not immediately) released. I already tried to find a solution but we would need to hae a deeper access to sqlite adapters, i think we could just use a try / catch when deleting the file

I will do a PR soon, for now you can just re-run the CI jobs

PR #8821, opened 2021-03-11
VS Test SDK 16.9.1

agriffard:
https://github.com/microsoft/vstest/releases/tag/v16.9.1


agriffard:
Was there a reason to not update it?

hishamco:
Can we have a BOT to upgrade the packages automatically?

sebastienros:
What is this package adding? If only support for .NET test types, then we don't use it

agriffard:
Used to run the tests on VS.

PR #8828, opened 2021-03-11
Added missing namespace `OrchardCore.Taxonomies.Helper` to `TaxonomyOrchardHelperExtensions.cs`

Diman96:
because it does not allow to reference global namespace in Razor views.


deanmarcussen:
The intended usage of these extensions is
`@Orchard. GetTaxonomyTermAsync`

where Orchard is the `IOrchardHelper`

What problem are you having that is making you want to do this?

xperiandri:
@deanmarcussen, looks like someone forgot to add namespace to the file and the extensions class appeared in `global` namespace.
Those `GetTaxonomyTermAsync` cannot be imported as extension method anyhow in Razor. And an explicit call also cannot be done.

deanmarcussen:
This is all by design (all the `IOrchardHelper` extension methods are designed this way - and they all work fine in Razor)

can you share some code where it is not working?

Diman96:
@deanmarcussen  
`GetTaxonomyTermAsync` not found.
`<PackageReference Include="OrchardCore.Taxonomies" Version="$(OrcahrdCoreVersion)" />`

image

deanmarcussen:
It's because you're passing a `dynamic`

You can't pass dynamic to extension methods, you have to cast them first.

`@await Orchard.GetTaxonomyTermAsync((string)Model.ContentItem.Content.... )`

PR #8837, opened 2021-03-12
Remaining OpenId classes to OpenID Collection Table

deanmarcussen:
It'll be easy they said ;)

Bug in YesSql needs fixing first.

Reduce indexes don't work in collections.

Everything else works though...

/cc @kevinchalet for any more requests ;)

### DIFF
--- a/src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Indexes/OpenIdApplicationIndex.cs
+++ b/src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Indexes/OpenIdApplicationIndex.cs
@@ -29,7 +29,12 @@ namespace OrchardCore.OpenId.YesSql.Indexes
     }
 
     public class OpenIdApplicationIndexProvider : IndexProvider<OpenIdApplication>
-    {
+    {        
+        private const string OpenIdCollection = OpenIdApplication.OpenIdCollection;
+
+        public OpenIdApplicationIndexProvider()
+            => CollectionName = OpenIdCollection;
+            
         public override void Describe(DescribeContext<OpenIdApplication> context)
         {
             context.For<OpenIdApplicationIndex>()
--- a/src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Migrations/OpenIdMigrations.cs
+++ b/src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Migrations/OpenIdMigrations.cs
@@ -61,7 +61,8 @@ namespace OrchardCore.OpenId.YesSql.Migrations
 
             SchemaBuilder.CreateReduceIndexTable<OpenIdAppByRoleNameIndex>(table => table
                 .Column<string>(nameof(OpenIdAppByRoleNameIndex.RoleName))
-                .Column<int>(nameof(OpenIdAppByRoleNameIndex.Count)));
+                .Column<int>(nameof(OpenIdAppByRoleNameIndex.Count)),
+                collection: OpenIdApplicationCollection);
 
             SchemaBuilder.AlterIndexTable<OpenIdAppByRoleNameIndex>(table => table
                 .CreateIndex("IDX_COL_OpenIdAppByRoleName_RoleName",
@@ -504,11 +505,16 @@ namespace OrchardCore.OpenId.YesSql.Migrations
                 _session.Delete(scope);
             }
 
-            // This can be safely dropped here as the index provider now only writes to the new collection table.
+            // Flush the saved documents so that the old reduced indexes will be calculated
+            // and commited to the transaction before they are then dropped.
+            await _session.FlushAsync();
+
+            // These can be safely dropped after flushing.
             SchemaBuilder.DropMapIndexTable<OpenIdApplicationIndex>();
             SchemaBuilder.DropReduceIndexTable<OpenIdAppByLogoutUriIndex>();
             SchemaBuilder.DropReduceIndexTable<OpenIdAppByRedirectUriIndex>();
             SchemaBuilder.DropReduceIndexTable<OpenIdAppByRoleNameIndex>();
+
             SchemaBuilder.DropMapIndexTable<OpenIdScopeIndex>();   
             SchemaBuilder.DropReduceIndexTable<OpenIdScopeByResourceIndex>(); 
 
--- a/src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Migrations/OpenIdMigrations.cs
+++ b/src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Migrations/OpenIdMigrations.cs
@@ -12,6 +12,8 @@ namespace OrchardCore.OpenId.YesSql.Migrations
     {
         private const string OpenIdTokenCollection = OpenIdToken.OpenIdCollection;
         private const string OpenIdAuthorizationCollection = OpenIdAuthorization.OpenIdCollection;
+        private const string OpenIdApplicationCollection = OpenIdApplication.OpenIdCollection;
+        private const string OpenIdScopeCollection = OpenIdScope.OpenIdCollection;
 
         private readonly ISession _session;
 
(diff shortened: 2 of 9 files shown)


src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Indexes/OpenIdApplicationIndex.cs
```diff
@@ -29,7 +29,12 @@ public class OpenIdAppByRoleNameIndex : ReduceIndex
     }
 
     public class OpenIdApplicationIndexProvider : IndexProvider<OpenIdApplication>
-    {
+    {
```
kevinchalet, line 32: nit: extra spaces 😄

kevinchalet (approved):

src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Migrations/OpenIdMigrations.cs
```diff
@@ -391,5 +405,120 @@ public async Task<int> UpdateFrom6Async()
...
+        public async Task<int> UpdateFrom7Async()
+        {
+            // Create all index tables with the new collection value.  
+            SchemaBuilder.CreateMapIndexTable<OpenIdApplicationIndex>(table => table
+                .Column<string>(nameof(OpenIdApplicationIndex.ApplicationId), column => column.WithLength(48))
+                .Column<string>(nameof(OpenIdApplicationIndex.ClientId), column => column.Unique()),
+                collection: OpenIdApplicationCollection);
+
+            SchemaBuilder.AlterIndexTable<OpenIdApplicationIndex>(table => table
+                .CreateIndex("IDX_COL_OpenIdApp_DocumentId",
```
jtkech, line 418: I like what you did in all other places by adding `_COL` and just removing `Index`, so keeping the table name

Then when the table name is too long we can apply some acronym rules, i.e.  `WorkflowBlockingActivities => WFBlockingActivities`, or here but just as an example `OpenIdApplication => OIDApplication`

So here would be `IDX_COL_OpenIdApplication_DocumentId`, but not so important if too late ;)

jtkech:
@deanmarcussen 

Okay, i like the `IDX_COL` prefix

Then keep the table name as long as it is possible, but yes without the `Index` part

If too long, better to first remove `_DocumentId` before abrevating the table name

(if you want here, for `OpenId` you could remove all `_DocumentId`)

YesSql doesn't append `_DocumentId` for its automatically created index `IDX_FK_AutoroutePartIndex`

So here e.g.
`IDX_COL_OpenIdApp_DocumentId` =>
`IDX_COL_OpenIdApplication_DocumentId` OR
`IDX_COL_OpenIdApplication`

And for this one that is already updated and merged
`IDX_COL_OpenIdAuth_DocumentId_ApplicationId` =>
`IDX_COL_OpenIdAuthorization_ApplicationId`

deanmarcussen:
> @deanmarcussen
> 
> Okay, i like the `IDX_COL` prefix
> 
> Then keep the table name as long as it is possible, but yes without the `Index` part

Agreed. Will update the others as well, and do some checking on other dbs. WIll probably enable some features here for testing, and push them at the functional db tests, to be sure.

Was seeing yesterday more MySql length issues, particularly with reduce indexes, because of the length, i. `OpenId_OpenIdAppByRoleNameIndex_OpenId_Document` which is now 48 characters, so ok, but with table prefixes as well :(

YesSql bits to finish fixing first so no rush ;)

jtkech:
@deanmarcussen 

> Was seeing yesterday more MySql length issues, particularly with reduce indexes, because of the length, i. OpenId_OpenIdAppByRoleNameIndex_OpenId_Document which is now 48 characters, so ok, but with table prefixes as well :(

I assume here `OpenId` is not a tenant table prefix but the collection name prefix, right?

Jus tried with the last dev to enable `OpenId`, it was failing, so i moved the collection configuration as you did here, but it was still producing `IDX_FK_OpenIdAppByRoleNameIndex_Document` without the collection name, maybe related to the fix you did on YesSql, but i thought it was included in the last package that we now reference in the last dev.

Would need to try your branch, hmm but here you don't reference the last version of YesSql, i would need it, right?

Anyway, if so we need to have short collection name, for example here `OID`, and this acronym could be also used (if needed) for the table name included in index names e.g. `OIDByRoleName`, and for reduce also for the table name itself.

Just thought about Postgres that needs unique index names, not per table but for the whole table, what happens with multi tenants, will the index names also include the tenant table prefix? If so we would need to limit these prefixes.

deanmarcussen:
> Jus tried with the last dev to enable OpenId, it was failing,

That concerns me. Should be working with dev. Will check. Which features did you enable?

Latest dev should only have collections for Tokens, and Authorizations, so Applications not yet affected.

> maybe related to the fix you did on YesSql, but i thought it was included in the last package that we now reference in the last dev.

The fix for this pr is not in the latest YesSql release. Still to come (with the other fix).

> Just thought about Postgres that needs unique index names, not per table but for the whole table, what happens with multi tenants, will the index names also include the tenant table prefix?

I believe this is already happening (coded in in YesSql). Because each functional test uses a generated table prefix (and it took a bit to get that small enough) - but didn't you fix the table prefix javascript at one point?

But yes, really we need to be well under the 64/63 limit, to allow for table prefixes (and I would have to open a database to be 100% sure what it is doing under the hood)

jtkech:
@deanmarcussen 

Okay, i merged #8857 

Yes, it logs error on the OpenId background task as mentioned in #8856 

So for now, nothing more that i can check on my side, we have to wait for the related updates on YesSql side

Then, maybe still time to find the best compromise for shorter table / index / collection / prefix names ;)

Let me know if you will need any help ;)

agriffard:
Conflicts to solve.

jtkech:
@deanmarcussen 

Any new on the related update of yesSql you are waiting?

Would be good to fix the OpenId bag task exception

deanmarcussen:
Released last night @jtkech .

Updated on a seperate pr - to resolve the background tasks.

Will update this one (maybe today if I have time) but leave it for you to review, and decide if we are happy on the index names

jtkech:
I like the new index names in `OpenIdMigrations` !!!

I like also the `OpenId` collection name, unless you think we could still have too long names

Hmm, a shorter name as `OID` would be only 3 char less ;)

deanmarcussen:
@jtkech ok was just writing a note for you :)

I tried and is ok with tenants/mysql. Couple more things to test there.

I think best to stick with `OpenId` for the collection name. Otherwise it's been 10 days now, and another migration for the current users.

Couple more things to test, but if you want to change any of the index names, just push something, or suggest a change.

Hope to finish testing tomorrow, and get it finished off, so it's all in sync

jtkech:
> I think best to stick with OpenId for the collection name

No problem

jtkech:
LGTM, i will approve but let you finish your tests and remove the dontmerge label

src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Indexes/OpenIdApplicationIndex.cs
jtkech: Done ;)

jtkech:
@deanmarcussen 

One question, when you move data to a new collection, are the new index tables (we create just before) populated?

If not, we could use a defferred task (before we could not while activating the shell, now we can)

Note: At the end of the last step we drop indexes, i assume we don't care about old SQL versions

deanmarcussen:
@jtkech testing notes...  just going to keep updating this as I tweak things, so you know why as well

MySql table prefix of `MyTable`

```
---> MySqlConnector.MySqlException (0x80004005): Identifier name 'MyTable_FK_OpenId_OpenIdAppByLogoutUriIndex_OpenId_Document_DocumentId' is too long
```

deanmarcussen:
@jtkech 

> One question, when you move data to a new collection, are the new index tables (we create just before) populated?

Yes, the tables are created as part of the transaction, and then populated during the save. (the old indexes are cleaned at the same time, which isn't really necesary, as they are later dropped, but no way to avoid it - would throw exceptions if they were dropped first.

> Note: At the end of the last step we drop indexes, i assume we don't care about old SQL versions

That's what I assumed here as well. So the version number gets reset, because it is a new document in the new table.

This remains blocked by YesSql for the moment. need to have a chat with @sebastienros see if there isn't a way to solve this name / length issue once and for all.

jtkech:
@deanmarcussen 

> this is probabl YesSql's default indexes now causing the issue

Okay

    CreateForeignKey("FK_" + bridgeTableName + "_DocumentId", bridgeTableName, new[] { "DocumentId" }, documentTable, new[] { "Id" });
    MyTable_FK_OpenId_OpenIdAppByLogoutUriIndex_OpenId_Document => 59 chars

    TablePrefix?               MyTable_
    "FK_" +                    FK_
    bridgeTableName +          OpenId_OpenIdAppByLogoutUriIndex_OpenId_Document
    "_DocumentId",             _DocumentId

Maybe remove from the bridgeTableName the `Index` part => 5 chars less
Maybe use only once the collection name => 6 chars less

Hmm, if always a bridge to the `Document` table of the related collection, maybe use a fixed and shorter suffix

    MyTable_FK_OpenId_OpenIdAppByLogoutUri_BRDG => 43 chars
    MyTable_FK_OID_OpenIdAppByLogoutUri_BRDG => 40 chars

Then this one

    MyTable_FK_OpenId_OpenIdAppByRedirectUri_BRDG => 45 chars
    MyTable_FK_OID_OpenIdAppByRedirectUri_BRDG => 42 chars

jtkech:
@deanmarcussen 

> Yes, the tables are created as part of the transaction, and then populated during the save.

Good to know, i though it was a problem

Note: About dropping indexes and the versions we don't care, i was talking about SQL versions before SQL 2016

deanmarcussen:
@jtkech 

> > this is probabl YesSql's default indexes now causing the issue

Sick of it always being a problem.

Am looking to see it can hash the key name if it’s over 63/64 chars for the dialect.

something like ‘FK_hash’

only for indexes so won’t solve table name lengths, but should resolve this

deanmarcussen:
@jtkech 

ok, I mark this ready, and let you have a look, for final thoughts.

Tested
- MySql with a table prefix
- Postgres with a table prefix

Thoughts: 

Now we have the hash we could use a longer index name. If you are bothered about consistency and conventions for index names.


> Note: At the end of the last step we drop indexes, i assume we don't care about old SQL versions

Just have to live with it I think. Could move them into a DeferredTask, because it would push the exception till later, and allow the migration to complete. But really we need people to be running 2016+ It's not just about dropping indexes there, but also the index key lengths

jtkech (approved):

Issue #8838, opened 2021-03-12
Regression from Dashboard Widgets


deanmarcussen:
### To Reproduce

Go the edit the Blog, a `ListPart`

The `DetailAdmin` display location added for Dashboard Widgets causes both the `TitlePart` and `HtmlBodyPart` to be rendered in the `Content` zone.

### Expected behavior

I didn't follow completely the choice to use `DetailAdmin` for the dashboard widgets.

Feels like it conflicts with existing `DetailAdmin` views (of which there is only really the `ListPart`)

Easy to fix the `TitlePart` by giving it a `DetailAdmin` view and removing the existing rendering from the `Content.DetailAdmin` shape.

Less easy to fix the `HtmlBodyPart` showing as well. Without resorting to a different display type.

Which would probably be my preference. `DisplayAdmin` for the `ListPart` as it **was** the only one making use of it.

The point is when using the `ListPart` with more parts on it's parents, the list display gets ridiculous, where it shouldn't be showing these extra details, just the list of items that are contained in it (unless user specifically put them there with placement).

`DisplayAdmin` makes sense to me, as the controller it's viewed through is actually called `Display`

### Screenshots

<img width="1126" alt="Screenshot 2021-03-12 at 10 49 20" src="https://user-images.githubusercontent.com/13782679/110932807-df658b00-8323-11eb-80d0-f4388c7fffdd.png">

hishamco:
I just notice the same bug, hope it could be fixed soon

sebastienros:
> @sebastienros will want to do it differently :)

You bet! I am glad you realized your PR would be controversial, like it was bothering you somehow but couldn't find a better way without going deeper in the reflection ;)

I don't buy the argument of Display controller -> `DisplayAdmin` ... or we would change the front-end display type to `Display` instead of `Details`.
But that's not the issue here. You are creating a new display type such that no other part can be rendered than the ones that you want to be in this screen. And because you decided that `ListPart` was fine, we got only `ListPart` ... let's call this display type `ListPartAdmin` ;) Why wouldn't `TitlePart` be by default on a `DetailAdmin` page? Or a body part? Quite opinionated.
What if someone wants to display something else in the admin, that doesn't have a ListPart? It will show empty.

How do we handle that in OC usually? Placement or custom templates. The only dynamic solution being placement in this case. 
_Ideally_ we would configure the Blog to hide its `Title` and `HtmlBodyPart` from the `DetailAdmin` display type.

[...]

PR #4184, opened 2019-09-07
Only show fields in Summary and Detail display types

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/3838 

and helps with https://github.com/OrchardCMS/OrchardCore/pull/4154

So a proposal only, but I go with the easy option : Always hide fields in `DetailAdmin` display type.

This stops them showing in the List view which is currently the only view making use of `DetailAdmin`

If someone wants a field to show in `DetailAdmin` or the List, then `placement.json` in any custom module will participate in placement for TheAdmin theme and the fields can be shown. (for me sometime I want them, they're useful as a header in a list container, but more often than not I don't want them, and it just looks weird when they show up by default).

Example placement that a custom module could use to override the driver setting and place a field. 

```
"TextField": [
    {
      "displayType": "DetailAdmin",
      "place": "Content"
    }
]
```

PR #4154, opened 2019-09-04
Add media field to blog & blog post

hishamco:
Fixes #1444

PR #8839, opened 2021-03-12
Placement for dashboard widgets

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/8838

@sebastienros will want to do it differently :) so no rush to merge it yet.


sebastienros:
Where did my comments go?

deanmarcussen:
Did you mean these comments ? https://github.com/OrchardCMS/OrchardCore/issues/8838#issuecomment-797817828

I'm cutting and paste them, because I need to reply...

You bet! I am glad you realized your PR would be controversial, like it was bothering you somehow but couldn't find a better way without going deeper in the reflection ;)

I don't buy the argument of Display controller -> DisplayAdmin ... or we would change the front-end display type to Display instead of Details.
But that's not the issue here. You are creating a new display type such that no other part can be rendered than the ones that you want to be in this screen. And because you decided that ListPart was fine, we got only ListPart ... let's call this display type ListPartAdmin ;) Why wouldn't TitlePart be by default on a DetailAdmin page? Or a body part? Quite opinionated.
What if someone wants to display something else in the admin, that doesn't have a ListPart? It will show empty.

How do we handle that in OC usually? Placement or custom templates. The only dynamic solution being placement in this case.
Ideally we would configure the Blog to hide its Title and HtmlBodyPart from the DetailAdmin display type.

I thought we could share the DetailAdmin with the dashboard too, so we don't create too many display types that we have to provide defaults in each display driver. Maybe we need one for the dashboard widgets (Dashboard?). But I don't think this removes the issue of how we decide what is on DetailAdmin.

We can already handle the placement for Blog with the recipe or theme. That would be the best solution for now IMO. But only for Blog. What about other types?
A nice solution, if we don't want to use @TFleury 's new admin placement feature, would be to provide checkboxes with common display types, maybe in the current list of Parts/Fields from the Content Type Definitions screen, and there we could check/uncheck where this part should be hidden (displaying is harder to manage, because placement, and also what's the default values if nothing is set?). This would be a custom placement info provider. And we could also use it as part of the Blog recipe to actually hide Title and HtmlBodyPart from appearing there. Some global settings would let us provide which display types we want to show in the list (DetailAdmin, SummaryAdmin). Probably would be for admin placement only. Could also contain just "string" placement rules ("-", "/Header", ...) but it would be close to the placement.json editor. We definitely need some better UI there.

So for now I would just provide a placement file with the Blog recipe to hide the parts from this view. And I want to repeat that if we decide to add a new display type for dashboard, the issue remains, though might be postponed.

sebastienros:
Yes, why don't I see them?

sebastienros:
Ah they are on the issue :/

deanmarcussen:
So done rather quickly, because sometimes it's better to get the wrong pr started, to at least get a discussion.

> I don't buy the argument of Display controller -> DisplayAdmin
Agree.

Couple of other prs that were related
https://github.com/OrchardCMS/OrchardCore/pull/4154#issuecomment-529038520
https://github.com/OrchardCMS/OrchardCore/pull/4184

So you argued against using placement to solve this issue with `Lists` when fields were added.

And I very much agreed with the argument, so we moved fields to not display everywhere.

I don't think placement is the right way to solve it, obviously I can solve it for the Blog Theme, but am more interested in solving it globally, so it doesn't prove to be a pain for existing sites, and future sites.

I do see a future where fields are being rendered in the dashboard widgets, -> Field => DetailAdmin. (@agriffard is probably minutes away from making a pr to do that ;)

> Maybe we need one for the dashboard widgets (Dashboard?). But I don't think this removes the issue of how we decide what is on DetailAdmin.

I think this would be the better choice for `DashboardWidgets`

we (internally) often create custom display types, to fit specific needs, so while we don't want to have a million of them in OC, it seems like the dashboard widgets are quite a specific use case, which would suit a custom display type.

There is (for me) a connection between `DisplayAdmin` and the `Detail` controller action. But maybe that's just because I use lists a lot...

Back in O1 we had a list controller, and maybe that's the other thing to consider here, is that maybe lists are a bit special, and shouldn't be `DetailAdmin` at all, but a specific controller. 

Given that we've struggled a few times to resolve things showing up where they shouldn't.

But its own controller introduces its own problems.

Skrypt:
I've had issues posting reviews on PR's too this week-end. It was litterally freezing after clicking the send button.

deanmarcussen:
Notes from conversation with @sebastienros 

- Everything is about placement
- Driver placement is about convention
- Maybe we need a new convention for the `ListPart` to say that it should only display it's contained items
- Or maybe we need a new convention for the dashboard widgets.

For me there is still a connection between the `AdminController.Display` action, which is used to display the list, and its items and the convention of placement that should be applied to the outcome.

Mostly this connection is about expected outcomes. i.e. when using `AdminController.Display` there is an expectation of an outcome, that a special display operation is underway, and the conventions for this display should be applied, i.e. the drivers should be able to know where the display call is coming from, or at least it's intention.

deanmarcussen:
That being said, and following on from our discussion of conventions.

The `DashboardWidgets` also follow a convention, with a stereotype? 

If I'm not mistaken?

Maybe the stereotype could become involved with the display type of `DetailAdmin`. 
So we would not need necesarily a new display type to differentiate, but a convention that says `DetailAdmin:DashboardWidget` 

(would have to look to see if we don't already have this functionality already and just need to add it)

sebastienros:
Maybe adding something like `Location("Content", "1", stereotype)` to define a position that is only for a specific stereotype. Unless this concept is not related to the current object model, and it's the concern of a different component.

deanmarcussen:
> Maybe adding something like Location("Content", "1", stereotype) to define a position that is only for a specific stereotype. Unless this concept is not related to the current object model, and it's the concern of a different component.

Ok I went back and forth a bit on this, and ended up with specific placement for the `DashboardWidget` stereotype.

I've added it in `placement.json` files where it was used, rather than customising something special in the driver, because it is not related to the current object model, to do so, would just be replicating code that the `IPlacementNodeFilters` already do and are better equipped to support.

The thread of my thought is that `DashboardWidgets` are what require parts to be displayed in `DetailAdmin`, so it should be the placement specifier.

Rather than the alternate of Lists hiding parts/fields they don't want to see. Because that would mean much confusion and need to write custom placement.

deanmarcussen:
Approved in Triage.

TODO on another pr : implement stereotype placement seperate from content type placement

PR #8831, opened 2021-03-11
Arrows for collapsible panels

agriffard:
Use the same kind of arrows (Right if collapsed, Down if expanded) for all collapsible panels in admin.

image

PR #8841, opened 2021-03-12
Fix arrows for collapsible panels in RTL languages

hishamco:
Extend #8831 to make the arrows show properly in RTL languages

RTL Arrows

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Flows/Views/ContentCard-BagPart.Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Flows/Views/ContentCard-BagPart.Edit.cshtml
@@ -1,3 +1,4 @@
+@using OrchardCore.Localization
 @inject OrchardCore.ContentManagement.Metadata.IContentDefinitionManager ContentDefinitionManager
 
 @{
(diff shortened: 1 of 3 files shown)


hishamco:
BTW @agriffard , seems there's no need to use DOM as I said in your PR

agriffard:
Do it everywhere it is needed, ex: FlowPart.Edit (search for `angle-right`).

The problem I encounter is when you add a new bag or form element, this is always LTR.

hishamco:
There's no `angle-right` anymore ;)

> The problem I encounter is when you add a new bag or form element, this is always LTR.

I didn't deal with bag, if the case is to add a new item in the list, I will try to check it now

hishamco:
When I'm trying to add a new item the arrow goes down, after publishing it shows the left arrow which is correct in case of RTL languages

agriffard:
> When I'm trying to add a new item the arrow goes down

But when you collapse it, it will be a left arrow.

hishamco:
> But when you collapse it, it will be a left arrow.

That's right for RTL but not for LTR. Are you talking about LTR?

agriffard:
> There's no angle-right anymore

Do you the change also here:

https://github.com/OrchardCMS/OrchardCore/blob/fdec2c6a9a985cd6ceee7dce2e60a3347872aacd/src/OrchardCore.Modules/OrchardCore.Flows/Views/ContentCard-FlowPart.Edit.cshtml#L21

hishamco:
Seems there 3 places, I think I made a typo while searching, I will update the PR

agriffard:
In RTL, the dot `.` moved at the begining (ex: `.Services - Provides a collection behavior for your content item`) is a bit weird.

hishamco:
This is because the text is English, I will try to attach a new screenshot

hishamco:
Services

PR #8847, opened 2021-03-12
Darkmode should work with multi tenants in the same browser instance

### LINKED ISSUE #8846: Darkmode doesn't work with muli tenants in the same browser
The issue with the current implementation is whenever you have multiple tenants, and you start toggling between the darkmode say 'Light' all the other tenants will use `Light` even you set it previously to `Dark`, this is because the cookie and local storage shared between them

hishamco:
Fixes #8846


hishamco:
The simple fix to make the dark mode not conflict with multi tenants even in the same browser by prefix tenant name in cookie and local storage key

src/OrchardCore.Themes/TheAdmin/Views/Layout.cshtml
```diff
@@ -1,11 +1,13 @@
-@inject DarkModeService DarkModeService;
-@inject IOptions<AdminOptions> AdminOptions;
+@using OrchardCore.Environment.Shell
```
Skrypt, line 1: DarkModeService.TenantName so that we can expose this service in liquid later on.

hishamco: I didn't think about a scenario to use it later on in liquid, do you have one?

Skrypt: So I've just looked and we have sitename exposed in Liquid.
https://docs.orchardcore.net/en/dev/docs/reference/modules/Liquid/#site

Should we also expose the tenant name then?

Skrypt: > I didn't think about a scenario to use it later on in liquid, do you have one?

I did not either when I did the feature. But I figured out afterward that we can create Admin Templates (Liquid) and we can override the current admin Layout.cshtml file from the current Admin theme by using an Admin Template named "Layout". So, yes we should be able to use the Darkmode in Liquid too.

Skrypt: And it would clean up that view of that specific `@using OrchardCore.Environment.Shell` which I think should be done in the Service instead.

Skrypt (changes requested):
Apply the changes from my previous comment about the CurrentTenant

src/OrchardCore.Themes/TheAdmin/Views/Layout.cshtml
```diff
@@ -1,11 +1,13 @@
...
+@inject DarkModeService DarkModeService
+@inject IOptions<AdminOptions> AdminOptions
 @{
     var adminSettings = Site.As<AdminSettings>();
     var darkMode = await DarkModeService.IsDarkModeAsync();
+    var tenant = Context.Features.Get<ShellContextFeature>().ShellContext.Settings.Name;
```
jtkech, line 7: The `ShellContextFeature` is more intended to be used in special cases, e.g. when buiding the tenant pipeline, to retrieve some infos from the ambient request when no more in a given shell scope.

Here better to just resolve / inject `ShellSettings`, so from the actual shell container

Skrypt: Not in the view. In the service please 😄

jtkech: Yes you're right, looks better to do it in the service itself

    public string CurrentTenant => _shellSettings.Name;

Skrypt (approved):

PR #8849, opened 2021-03-12
Fixing WidgetsList rendering

sebastienros:


sebastienros:
Since it passes the tests and it's breaking a feature I am merging it. Please review the code still in case something is wrong. There is a breaking change in the API (other interface) but it's a sub interface of IShape, so any code should still work, or with minor updates. Also less `dynamic`.

PR #8857, opened 2021-03-13
Move OpenID collection initialization to Core Startup

deanmarcussen:
/cc @jtkech 

Yes, doesn't work right, if you don't enable token validation currently.

So I move it to where the migrations are registered.

If you are checking, and this is good, please merge, or I merge it tomorrow - the other pr will not be ready.


jtkech:
Okay, i will try it this night

PR #8863, opened 2021-03-14
Use new .AddAsync() in contentitem tag helper

giannik:
because now shapes are added async


sebastienros:
Might also be better to remove the `dynamic` usage here

jtkech:
@sebastienros 

Here we need the dynamic usage as we access dynamic props e.g. 

                    dynamic content = context.Shape;
                    string handle = content.Handle;

Then here we could just use the following where `context.Shape` is an `IShape`, and without having to add an `using`

                    await context.Shape.AddAsync(displayShape, "");

jtkech (approved):

sebastienros:
I use Get property everywhere now instead of dynamic.

jtkech:
Okay cool

Do we need to change it here? maybe better in a separate PR to also update other ones

sebastienros:
Another is fine. I just do it when I see one. The GetProperty handles when there is no key, and can also take a default value. There is also a Try version.

PR #8867, opened 2021-03-15
Remove dynamic usages from IShapeTableProvider

giannik:
related #8863 
includes shapes : 

- content
- menu
- taxonomy term
- navigation

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Shapes.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Shapes.cs
@@ -2,6 +2,7 @@ using System;
 using Microsoft.Extensions.DependencyInjection;
 using OrchardCore.ContentManagement;
 using OrchardCore.ContentManagement.Display;
+using OrchardCore.DisplayManagement;
 using OrchardCore.DisplayManagement.Descriptors;
 using OrchardCore.DisplayManagement.ModelBinding;
 
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Shapes.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Shapes.cs
@@ -14,8 +14,8 @@ namespace OrchardCore.Contents
             builder.Describe("Content")
                 .OnDisplaying(displaying =>
                 {
-                    dynamic shape = displaying.Shape;
-                    ContentItem contentItem = shape.ContentItem;
+                    var shape = displaying.Shape;
+                    var contentItem = shape.GetProperty<ContentItem>("ContentItem");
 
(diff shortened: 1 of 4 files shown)


deanmarcussen:
If you're feeling super keen the `TermShapes` could be done also :)

giannik:
sure Im on a roll now.

Virus-free.
www.avast.com

sebastienros:
LGTM

sebastienros:
Interestingly this is already checked:

```
        public static bool operator ==(ZoneOnDemand a, object b)
        {
            // if ZoneOnDemand is compared to null it must return true
            return b == null || ReferenceEquals(b, Nil.Instance);
        }
```

And there is a unit test:

```
        [Fact]
        public void ZoneOnDemandIsEqualToNull()
        {
            var zoneOnDemand = CreateZoneOnDemand("SomeZone");

            Assert.True(zoneOnDemand == null);
            Assert.False(zoneOnDemand != null);
            Assert.True(zoneOnDemand == Nil.Instance);
            Assert.False(zoneOnDemand != Nil.Instance);
        }
```

But there is a test to ensure that if it's casted to `Shape` then it's not. That might be the issue.

jtkech:
Yes, i did these unit tests

As i remember the operator overload was only called if as a ZOD or as dynamic, but not e.g. if as a Shape

So at some point, e.g. i thought about defining other overloads, elsewhere, e.g. for IShape / Shape that would check if is a ZOD, but didn't want to complicate things while the dev branch was working

But now interesting to find a good solution, i will work on it soon

sebastienros:
I don't think adding operator overloads in Shape is good, because it makes it know about ZOD.
I'd go with fixing IsSectionDefined(zod) by not using dynamic and checking for ZOD directly (`is ZoneOnDemand zod && zod != null`) or adding a custom method to `Zones` to check for it, like `Layout.Zones.Has/Exists/IsEmpty("Content")`. Then looking for everywhere we do the check and replace it.

jtkech:
Yes, just took a quick look, one problem is that ZOD can be retrieved as a `Shape`, `IShape`, also as an `object`

And e.g. we can't define a static operator overload on an interface

So yes better to check if it is a ZOD where it needs to be done, as i did at some point e.g. in `ThemeLayout.cs`

Then yes, looks good to have an extension for `Zones`

I will do a PR to fix it everywhere it needs to be done

Just need some time, i think tomorrow.

jtkech:
@giannik 

Oops sorry, my last comments were not directly related to your PR, i thought i was under the #8871 thread.

jtkech (approved):

PR #8868, opened 2021-03-15
Fix layers content type rule to not include stereotypes

deanmarcussen:
Fixes logic

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Rules/Drivers/ContentTypeConditionEvaluatorDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Rules/Drivers/ContentTypeConditionEvaluatorDriver.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using OrchardCore.ContentManagement;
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.Rules/Drivers/ContentTypeConditionEvaluatorDriver.cs
```diff
@@ -31,10 +32,10 @@ public ContentTypeConditionEvaluatorDriver(IConditionOperatorResolver operatorRe
         public override Task<IDisplayResult> DisplayAsync(ContentItem contentItem, BuildDisplayContext context)
         {
             // Do not include Widgets or any display type other than detail.
-            if (context.DisplayType == "Detail" && context.Shape.TryGetProperty(nameof(ContentTypeSettings.Stereotype), out string stereotype))
+            if (String.Equals("Detail", context.DisplayType, StringComparison.OrdinalIgnoreCase) && !context.Shape.TryGetProperty(nameof(ContentTypeSettings.Stereotype), out string stereotype))
```
jtkech, line 35: Here that's fine to still use `context.DisplayType == "Detail"`, here `String.Equals()` will do the same wih one more method call, this to check the comparison type, and to do the same if it is `OrdinalIgnoreCase`

sebastienros:
Is it the Rules module that defined when a widget is visible, and not the Layers?

src/OrchardCore.Modules/OrchardCore.Rules/Drivers/ContentTypeConditionEvaluatorDriver.cs
jtkech: And if `out string stereotype` is not used, you coul just use the `_` discard char

deanmarcussen: Good idea @jtkech Done.

deanmarcussen:
> Is it the Rules module that defined when a widget is visible, and not the Layers?

Yes, I kept most of the rules, in rules (this particular one is a bit more layers though).

Mostly because the existing javascript layer methods get used in a lot of places, workflows etc, so while orientated at layers, are a bit more universal than just layers. The url one for example, I have used all over the place with workflows.

Open to moving them back to layers, or as a feature somewhere?

PR #8871, opened 2021-03-15
Fix IsSectionDefined

NogginBox:
Changes to Zone indexer broke the IsSectionDefined method.

Zone must dynamic for null check to work in case it is a ZoneOnDemand. The null check does not work if it is an IShape.

### DIFF
--- a/src/OrchardCore.Themes/TheTheme/Views/Layout.cshtml
+++ b/src/OrchardCore.Themes/TheTheme/Views/Layout.cshtml
@@ -33,7 +33,7 @@
         @await RenderSectionAsync("Messages", required: false)
         @await RenderBodyAsync()
     </main>
-    @if (ThemeLayout.Zones["Footer"] != null) {
+    @if (IsSectionDefined("Footer")) {
     <footer>
         <div class="container">
             @await RenderSectionAsync("Footer", required: false)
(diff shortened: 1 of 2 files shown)


Skrypt:
Can't we use a `Nullable<IShape>` ?

jtkech:
@NogginBox Good catch

But yes, as @Skrypt said, one goal is to not re-introduce dynamic usages

I will work on it this night

sebastienros:
I think it should be fine to say that ZOD compares to null when it's empty. And if we want to check if a ZOD is already assigned in the ZoneHolding dictionary, then use the dictionary key lookup.

jtkech:
Okay cool, need to finish some other things, then i will focus on it

NogginBox:
@jtkech Looks like the fix for this is beyond the scope of this PR. Do you want me to close this and copy the details into an issue for you to do your PR against?

jtkech:
@NogginBox no problem, no rush, just started #8880, i will let you know ;)

jtkech:
@NogginBox 

Okay, #8880 has been merged, so i close this one

Thanks again for having pointed out the related issue

NogginBox:
@jtkech thanks for fixing.

My theme went crazy when IsSectionDefined stopped working.

PR #8872, opened 2021-03-15
Fixes LiquidUserAccessor

### LINKED ISSUE #8862: The liquid user filters are not resolved
Try using the liquid filter  :  {{ User | user_id }}
It returns null.

Seems like this line below does not return a claimsrpincipal an exits

https://github.com/OrchardCMS/OrchardCore/blob/47da5ec7c1a984f3ec545dee4c04d3b9b7f5e5c6/src/OrchardCore.Modules/OrchardCore.Users/Liquid/UserFilters.cs#L26

Similarlly it is not possible to access the properties of the user because this line has been commented out.

https://github.com/OrchardCMS/OrchardCore/blob/47da5ec7c1a984f3ec545dee4c04d3b9b7f5e5c6/src/OrchardCore.Modules/OrchardCore.Users/Startup.cs#L241

jtkech:
Fixes #8862 

- So before we were setting the `User` object in a given liquid scope, allowing to use on it an user filter

      {{ User | user_id }}

- Now we globally use a `LiquidUserAccessor` that would allow to access its properties on demand. The problem is that if we use `User` alone without specifying a property, it returns null

    So, here i updated the `LiquidUserAccessor` to return the user itself if we use a specific prop name, for now i used `Current` just to check that the following works

      {{ User.Current | user_id }}

- Then, i removed the other prop names (they were commented out), so we still return null because, as i understand, we are intended to use the related user liquid filters instead, but i may be wrong.

**Would be good to have an accessor allowing to return the object itself if no property is specified**

But maybe this already exists ;)


sebastienros:
For db user:

```
                o.Scope.SetValue("User", new ObjectValue(new LiquidUserAccessor()));

                o.MemberAccessStrategy.Register<User, FluidValue>((obj, name, context) =>
                {
                    return name switch
                    {
                        nameof(User.UserId) => obj.UserId,
                        nameof(User.UserName) => obj.UserName,
                        //nameof(User.NormalizedUserName),
                        //nameof(User.Email),
                        //nameof(User.NormalizedEmail),
                        //nameof(User.EmailConfirmed),
                        //nameof(User.IsEnabled),
                        //nameof(User.RoleNames),
                        //nameof(User.Properties)
                    _ => return NilValue.Instance
                    };
                });
```

For User claims:

```
public static ValueTask<FluidValue> UserId(FluidValue input, FilterArguments args, TemplateContext ctx)
        {
            if (input is LiquidUserAccessor)
            {
                var httpContextAccessor = liquidTemplateContext.Services.GetRequiredService<IHttpContextAccessor>();
                var user = httpContextAccessor.HttpContext.User;
            }

...
```

jtkech:
Okay i will take look this night

Didn't thought about dealing with the `LiquidUserAccessor` in the filter itsef ;)

Hmm, but need to be done in all liquid user filters

deanmarcussen:
Slightly edited @jtkech 

The thing we forgot about is their are two types of user object. The `User` that acts as an accessor to the `ClaimsPrincipal` - for filter access only really, so yes, all the user filters might need the change.

And the other was the db user object, which can be retrieve via some of the user filers, i.e. `users_by_id`

jtkech:
Okay i got it

sebastienros (approved):

jptissot:
@sebastienros Will [these](https://docs.orchardcore.net/en/dev/docs/reference/modules/Liquid/#user) user properties still be available on the User object ?

sebastienros:
Probably not. So we should intercept `Identity` to return a `ClaimsIdentity`. How did it work before?

jtkech:
@jptissot @sebastienros 

Okay, before the User (ClaimsPrincipal) was added in the liquid scope, now it's a `LiquidUserAccessor` that for now is only used throuh the liquid user filters.

We just need to also register a `MemberAccessStrategy` for `LiquidUserAccessor` to access these remaining properties, currently it's only done for the User (database).

I will do it tonight

jtkech:
@jptissot @sebastienros Just for info, done through #8925

PR #8880, opened 2021-03-16
ZoneOnDemand tweaks

jtkech:
@sebastienros 

- Okay good, now `IShape` has an `AddAsync()`, so we don't need anymore to check if it is a ZOD before calling it.

- I thought about getting rid of ZOD, but in many places where we get a zone to add a shape, we would need a null check and if null create a ZOD to be able to add a shape on it, so seems to be still useful to auto create a zone on demand.

- So here, for now i just added some `is ZoneOnDemand` checks where it seems they were missing.

- Then just created an `IShape.IsNullOrEmpty()` extension that i will start to use.

- For info in `RenderSectionAsync()` of `RazorPage` and `Page`, before it was

            var zone = ThemeLayout[name]; <= dynamic

            if (required && zone != null && zone is Shape && zone.Items.Count == 0)
            {
                throw new InvalidOperationException("Zone not found: " + name);
            }

    Maybe `zone != null` was added to not throw precisely if it is a ZOD (here as dynamic) so with no items, and `zone is Shape` for pre-rendered zones, that now are `PositionWrapper` that fail when accessing the `Items` property. Anyway looks good to have removed all these checks.

    But as it is now, it always fails if required and if zone (here an `IShape`) is not null     

            var zone = ThemeLayout.Zones[name]; <= IShape

            if (required && zone != null)
            {
                throw new InvalidOperationException("Zone not found: " + name);
            }
  
    So at least we would need to check `if (required && zone == null)`, but because zone is now an `IShape` that is at least a ZOD, for now i used the following. Note: If we still want to not fail on a ZOD, this check would become useless, otherwise, that's why we should not pre-render a ZOD in `ThemeLayout.cs` to be able to check it here.

            if (required && zone is ZoneOnDemand)
            {
                throw new InvalidOperationException("Zone not found: " + name);
            }

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
@@ -297,9 +297,7 @@ namespace OrchardCore.DisplayManagement.Razor
                 throw new ArgumentNullException(nameof(name));
             }
 
-            var zone = ThemeLayout.Zones[name];
-
-            return !zone.IsNullOrEmpty();
+            return ThemeLayout.Zones.IsEmpty(name);
         }
 
         /// <summary>
--- a/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
@@ -299,7 +299,7 @@ namespace OrchardCore.DisplayManagement.Razor
 
             var zone = ThemeLayout.Zones[name];
 
-            return zone != null;
+            return !(zone is ZoneOnDemand);
         }
 
         /// <summary>
@@ -367,7 +367,7 @@ namespace OrchardCore.DisplayManagement.Razor
 
             var zone = ThemeLayout.Zones[name];
 
-            if (required && zone != null)
+            if (required && zone is ZoneOnDemand)
             {
                 throw new InvalidOperationException("Zone not found: " + name);
             }
--- a/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
@@ -297,7 +297,7 @@ namespace OrchardCore.DisplayManagement.Razor
                 throw new ArgumentNullException(nameof(name));
             }
 
-            return !ThemeLayout.Zones.IsEmpty(name);
+            return ThemeLayout.Zones.IsNotEmpty(name);
         }
 
         /// <summary>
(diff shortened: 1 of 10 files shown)


src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
```diff
@@ -299,7 +299,7 @@ public new bool IsSectionDefined(string name)
 
             var zone = ThemeLayout.Zones[name];
 
-            return zone != null;
+            return !(zone is ZoneOnDemand);
```
sebastienros, line 302: why not use `return zone.IsNullOrEmpty()` ?

sebastienros (commented):
I'd still add a method on `Zones` to check if one is empty. That would hide the ZOD usage from outside.

src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
jtkech: @sebastienros 

Yes, i just created this extension and i plan to use it ;

jtkech: See in my first comment ;)

> Then just created an IShape.IsNullOrEmpty() extension that i will start to use.

jtkech:
> I'd still add a method on Zones to check if one is empty. That would hide the ZOD usage from outside.

Okay, i will look at it

sebastienros:
Looks good, however .... I am thinking that we usually want to check the opposite, like `IsNotEmpty()` ... and you can see that as you are adding `!` everywhere.

jtkech:
> Looks good, however .... I am thinking that we usually want to check the opposite, like IsNotEmpty() ... and you can see that as you are adding ! everywhere.

Yes, good idea !

jtkech:
Need a little more time to check if i didn't miss something

Need to check if it is okay in `ThemeLayout.cs` to set a `PositionWrapper`, it is an `IShape` and not a ZOD, hmm i think here we render the `Content` zone a 2nd time, i will try it.

                    // Check if the zone hasn't been processed and is not empty.
                    if (zone.Value is IShape shape && !shape.IsNullOrEmpty())
                    {
                        ThemeLayout.Zones[zone.Key] = new PositionWrapper(await DisplayAsync(shape), "");
                    }

sebastienros (approved):
Is there any docs to update?

jtkech:
I will check about the doc

Yes, we are calling twice `ThemeLayout.Zones[...] = new PositionWrapper(await DisplayAsync(...), "")` for the `Content` zone, this is not really 2 rendering as the 2nd time the shape is already a `PositionWrapper` so an `IHtmlContent`, but there are some extra calls to `ShapeExecuteAsync()` and `_htmlDisplay.ExecuteAsync()`.

Hmm, in `ThemeLayout` would need to also check if the `IShape` is also an `IHtmlContent`, and / or checking it e.g. in `RazorPage` to prevent these extra calls

I'm checking it

PR #8892, opened 2021-03-17
Fix npm build

### LINKED ISSUE #8891: npm install failing
npm install right now fails to install OrchardCore.Markdown module.

Skrypt:
Fixes #8891 

- Update EasyMDE to 2.14.0
- Update to lockfileVersion 2 (backward compatible)
- Use gulp-dart-sass instead of gulp-sass for TheComingSoonTheme, TheBlogTheme, TheAgencyTheme. (not waiting on these themes to get updated)

Lots of other issues reported by npm install but most of them are in the bootstrap themes which just needs to get updated.


agriffard:
TheBlogTheme and TheComingSoonTheme package.json are coming from StartBootstrap so if we make changes in them, they could be overwritten when they will be updated to bs 4.6.0 if we don't remember that we modified them.

Otherwise, LGTM.

Skrypt:
Yes, I agree, but if we wait on their author to fix their npm install we will wait for a while. Right now, in the state they are, if you do npm install on them, it fails on installing `gulp-sass`. So, the npm install is broken and thus makes the global one fail too. If we need to update these themes later on, I hope the author will notice that they are broken too.

Issue #8887, opened 2021-03-17
ArgumentOutOfRangeException: Specified argument was out of the range of valid values - 1.0.0-rc2/VS Enterprise newer versions


jason8496as:
### Describe the bug

We have been using **Orchard Core CMS 1.0.0-rc2** to develop a brand new website.

Recently we discovered an interesting issue that only happens when upgrading the Visual Studio version and doesn't let us visualize some of the site pages.

We are not pretty sure if it has something to do with the IDE itself or maybe with any of the individual components that it install when upgrading.

The admin portal can be accessed and all of the content is shown within it, but when trying to open some site pages we get the following error.

> An unhandled exception occurred while processing the request.  
> 
> ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'index')
> Microsoft.AspNetCore.Mvc.ViewFeatures.Buffers.ViewBufferTextWriter.Write(char[] buffer, int index, int count)

### To Reproduce
1. Set up a new **.Net Core 3.1 Web App** with all Orchard Core CMS requirements.
2. Update Visual Studio Enterprise to one of the latest version (e.g. 16.9.2 version)
3. Try to run the web project

### Expected behavior
The website is working as expected and every single page can be accessed using older versions of Visual Studio, for example, version 16.8.3.

### Screenshots
![image](https://user-images.githubusercontent.com/73246192/111510179-2502c900-8713-11eb-9c7c-0f8c625d8559.png)

jtkech:
I'm currently working on the `PagerShapes`

And I got the exact same exception

Repro

- Setup with the blog recipe
- Create more than 10 blog posts
- Go the blog home page that list the first 10 blog posts
- Click on `Older posts` => you get the exception

jason8496as:
@antonysb27 please feel free to add any further details you might consider important.

jtkech:
No problem, I'm trying it, i will let you know if i find anything

Skrypt:
Issue is related with latest merged changes in shapes with Fluid 2.0
I noticed this one last week.

```C#
            // specific cross-requests route data can be passed to the shape directly (e.g., OrchardCore.Users)
            var shapeRoute = shape.GetProperty("RouteData");
```

Issue was there if I remember correctly.

jtkech:
For my use case, i tried it before Fluid 2.0, same issue, i'm working on it

jtkech:
I found a fix at least for my use case

[...]

PR #8894, opened 2021-03-18
PagerShapes: Use IShapeFactory and cleanup

jtkech:
Maybe related to #8887

In my case to repro

- Setup with the blog recipe
- Create more than 10 blog posts
- Go to the blog home page that list the first 10 blog posts
- Click on Older posts => you get the following exception

Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware|ERROR|An unhandled exception has occurred while executing the request. System.**ArgumentOutOfRangeException**: Specified argument was out of the range of valid values. (Parameter 'index')
   at Microsoft.AspNetCore.Mvc.ViewFeatures.Buffers.ViewBufferTextWriter.Write(Char[] buffer, Int32 index, Int32 count)
   at System.IO.TextWriter.Write(ReadOnlySpan`1 buffer)
   at System.IO.TextWriterExtensions.WritePartialString(TextWriter writer, String value, Int32 offset, Int32 count)
   at System.Text.Encodings.Web.TextEncoder.Encode(TextWriter output, String value, Int32 startIndex, Int32 characterCount)
   at System.Text.Encodings.Web.TextEncoder.Encode(TextWriter output, String value)
   at Microsoft.AspNetCore.Mvc.ViewFeatures.**StringHtmlContent.WriteTo**(TextWriter writer, HtmlEncoder encoder)
   at Microsoft.AspNetCore.Html.**HtmlContentBuilder.WriteTo**(TextWriter writer, HtmlEncoder encoder)
   at Microsoft.AspNetCore.Mvc.Rendering.**TagBuilder.WriteTo**(TagBuilder tagBuilder, TextWriter writer, HtmlEncoder encoder, TagRenderMode tagRenderMode)

So in `PagerShapes` it fails when adding an `StringHtmlContent` in the inner `HtmlContentBuilder` of a `TagBuilder`

It fails under liquid, i also tried it with the dev branch before Fluid 2.0, it also fails so not related.

Replacing the `StringHtmlContent` by a copy here named `HtmlContentString` fixed the issue

The only difference is that the original `StringHtmlContent` does in its `WriteTo()`

    encoder.Encode(writer, _input);

With our new `HtmlContentString` here we do

    writer.Write(encoder.Encode(_value))

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Navigation/PagerShapes.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Navigation/PagerShapes.cs
@@ -129,7 +129,7 @@ namespace OrchardCore.Navigation
                 });
         }
 
-        private string EncodeAlternateElement(string alternateElement)
+        private static string EncodeAlternateElement(string alternateElement)
         {
             return alternateElement.Replace("-", "__").Replace('.', '_');
         }
@@ -145,7 +145,7 @@ namespace OrchardCore.Navigation
         }
 
         [Shape]
-        public async Task<IHtmlContent> Pager_Links(Shape shape, DisplayContext displayContext, dynamic New, IHtmlHelper Html, DisplayContext DisplayContext,
+        public async Task<IHtmlContent> Pager_Links(Shape shape, DisplayContext displayContext, IShapeFactory shapeFactory, IHtmlHelper Html,
             string PagerId,
             int Page,
             int PageSize,
@@ -156,12 +156,7 @@ namespace OrchardCore.Navigation
             object NextText,
             object LastText,
             object GapText,
-            bool ShowNext,
-            string ItemTagName,
-            IDictionary<string, string> ItemAttributes
-            // parameter omitted to workaround an issue where a NullRef is thrown
-            // when an anonymous object is bound to an object shape parameter
-            /*object RouteValues*/)
+            bool ShowNext)
         {
             var noFollow = shape.Attributes.ContainsKey("rel") && shape.Attributes["rel"] == "no-follow";
             var currentPage = Page;
@@ -189,7 +184,7 @@ namespace OrchardCore.Navigation
             var lastText = LastText ?? S[">>"];
             var gapText = GapText ?? S["..."];
 
-            var httpContextAccessor = DisplayContext.ServiceProvider.GetService<IHttpContextAccessor>();
+            var httpContextAccessor = displayContext.ServiceProvider.GetService<IHttpContextAccessor>();
             var httpContext = httpContextAccessor.HttpContext;
 
             var routeData = new RouteValueDictionary(Html.ViewContext.RouteData.Values);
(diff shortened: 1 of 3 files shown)


jtkech:
@sebastienros 

In fluid the same change has been done for the `StringValue`on Feb 6, 2021
https://github.com/sebastienros/fluid/blob/ad942870db05bf7f57b23e4d77fa628ec780a891/Fluid/Values/StringValue.cs#L149-L156

                // perf: Don't use this overload
                // encoder.Encode(writer, _value);

                // Use a transient string instead of calling
                // encoder.Encode(TextWriter) since it would
                // call writer.Write on each char if the string
                // has even a single char to encode
                writer.Write(encoder.Encode(_value));

So maybe worth to replace all our `StringHtmlContent` usages with a new custom `HtmlContentString` as described in my previous comment, doing a `writer.Write(encoder.Encode(_value))` in place of a `encoder.Encode(writer, _value)`.

At least it fixed the pager issue under liquid

Then yes, would be better to fully understand why it was failing

I will look in the source code of writers / encoders, what does `.WritePartialString()` as seen in the above trace.

jtkech:
So just to confirm that it is related to encoding

If for the pager i don't use any special char, e.g. `Older Posts` in place of `Older Posts →`, it doesn't fail

Can't find the source code of `System.IO.TextWriterExtensions.WritePartialString()`, looks like a new thing

So i think i will merge this one for the cleanup part of the pager, as it uses `IShapeFactory` in place of the dynamic `New`, and as it already fixes an issue when used under liquid

Then i will open a separate PR suggesting to replace all the `StringHtmlContent` by a new custom `HtmlContentString` using `writer.Write(encoder.Encode(_value))` in place of a `encoder.Encode(writer, _value)` as anyway it is beter to do that.

sebastienros:
System.Text.Encodings.Web was updated with the latest runtime release, for a security fix. There might be a bug in it, but I would bet for a bug in our own buffers. The startIndex out of range exception and the fact that we see buffer overflows in other places is the same bug: the offset is bigger than the expected string, and either the array is too small and there is an exception, or the array is big enough and it show some stale content from a previously pooled buffer.

So probably that some pooled buffer is not correctly initialized, or size is miscalculated. The only strange thing is that it's triggered by the Encoder update. I will check the changes.

src/OrchardCore.Modules/OrchardCore.Navigation/PagerShapes.cs
```diff
@@ -550,7 +572,16 @@ private IHtmlContent CoerceHtmlString(object value)
                 return result;
             }
 
-            return new StringHtmlContent(value.ToString());
+            return new HtmlContentString(value.ToString());
+        }
+
+        private class HtmlContentString : IHtmlContent
```
sebastienros, line 578: Why a new class since we already have it?

sebastienros: Or is it that in the meantime you created it ;)

madelson:
@sebastienros we aren't using Orchard but are seeing the same error after updating the encodings library. I'm curious, did you ever trace this down or report to MSFT?

sebastienros:
@madelson yes, it's well understood and reported (PS: I work for MSFT). Can you comment on https://github.com/dotnet/aspnetcore/issues/31299 and explain how you use it to repro the issue? Because OC was the only app reporting the issue, and there was a mitigation ready, this didn't get fix in a 5.0 service release.

sebastienros:
@madelson though if your issue is not related to an "empty" call to Encode(), then you probably have an issue that the Encoder now rents the buffers, so you should not keep a reference to them. This is a change in behavior from the .NET side.

PR #8907, opened 2021-03-19
StringHtmlContent to HtmlContentString

jtkech:
Suggest to replace all `StringHtmlContent` by a new `HtmlContentString` that uses `writer.Write(encoder.Encode(_value))`, in place of `encoder.Encode(writer, _value)` that calls `writer.Write()` on each char even if only one need to be encoded.

Not sure it optimizes something because currently i can't compare things on my too old machine

But at least it fixed an issue while using the pager under liquid, see #8894 

Same exception as reported in #8887 

@sebastienros just for info, regardless this PR,, when i do some apache tests i noticed that now the rps are better for the blog home page than for the about page, maybe due to my machine but i don't think so.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Navigation/PagerShapes.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Navigation/PagerShapes.cs
@@ -1,19 +1,17 @@
 using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
-using System.Text.Encodings.Web;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Html;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Rendering;
-using Microsoft.AspNetCore.Mvc.ViewFeatures;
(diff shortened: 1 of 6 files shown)


jtkech:
@sebastienros 

What you think about this one (see my above comment)?

And about my apache tests (regardless this PR) that now are slower on the about page

sebastienros (approved):

sebastienros:
I will check the perf, but that very surprising, the home page does more sql requests.

jtkech:
Okay cool, maybe my old machine, just to be sure ;)

And what about using a new `HtmlContentString`, as suggested here, in place of `StringHtmlContent` so that the encoder doesn't write each char if only one need to be encoded, as you did for the fluid `StringValue`

At least it fixed a pager issue under liquid

hishamco:
> And what about using a new HtmlContentString, as suggested here, in place of StringHtmlContent so that the encoder doesn't write each char if only one need to be encoded, as you did for the fluid StringValue

I want to point out to what Seb did in Fludi, but you write it before few seconds ;)

jtkech:
@sebastienros 

Oops sorry, just saw that you approved it ;)

PR #8908, opened 2021-03-19
Fix ChildContent rendering in liquid

TFleury:
Map Shape's Metadata property and render IHtmlContent from ShapeRenderFilter

See #8021 and #8024

Sample usage with Form_Wrapper shape :
```liquid
{% assign formPart = Model.ContentItem.Content.FormPart %}
<form action="{{ formPart.Action }}" method="{{ formPart.Method }}" enctype="{{ formPart.EncType }}" class="form-content {{ formContentTypeClassName }}">
    {% if formPart.EnableAntiForgeryToken %}
        {% antiforgerytoken %}
    {% endif %}

  {{ Model.Metadata.ChildContent | shape_render }}
</form>
```

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ShapeRenderFilter.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ShapeRenderFilter.cs
@@ -22,7 +22,8 @@ namespace OrchardCore.DisplayManagement.Liquid.Filters
                 return new HtmlContentValue(await task);
             }
 
-            if (input.ToObjectValue() is IShape shape)
+            var obj = input.ToObjectValue();
+            if (obj is IShape shape)
             {
                 var task = _displayHelper.ShapeExecuteAsync(shape);
                 if (!task.IsCompletedSuccessfully)
@@ -31,6 +32,10 @@ namespace OrchardCore.DisplayManagement.Liquid.Filters
                 }
 
                 return new ValueTask<FluidValue>(new HtmlContentValue(task.Result));
+            }
+            else if (obj is IHtmlContent html)
+            {
+                return new ValueTask<FluidValue>(new HtmlContentValue(html));
             }
 
             return new ValueTask<FluidValue>(new HtmlContentValue(HtmlString.Empty));
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewTemplate.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewTemplate.cs
@@ -117,6 +117,8 @@ namespace OrchardCore.DisplayManagement.Liquid
                         return shape.Attributes;
                     case "Items":
                         return shape.Items;
+                    case "Metadata":
+                        return shape.Metadata;
                     default:
                         if (shape.Properties.TryGetValue(n, out var result))
                         {


src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ShapeRenderFilter.cs
```diff
@@ -32,6 +33,10 @@ static async ValueTask<FluidValue> Awaited(Task<IHtmlContent> task)
 
                 return new ValueTask<FluidValue>(new HtmlContentValue(task.Result));
             }
+            else if (obj is IHtmlContent html)
```
Skrypt, line 36: Are you sure about this? Should it not be (if obj is IShape && obj is IHtmlContent) at the same time?

TFleury: Don't think so. basicaly `Shape` is not `IHtmlContent`

It should be equivalent to DisplayAsync ?https://github.com/OrchardCMS/OrchardCore/blob/37459b27a7437d398998e521a650a65172db6b3e/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs#L82-L100

Maybe we should also handle strings.

sebastienros, line 36: I want to understand what you are writing that should result in that behavior. It doesn't hurt, but it's kind of wrong to let people use `shape_render` on anything and it just works. Because tomorrow it's a string, which will fail, then an integer ... The dev in me just wants to throw an exception to let them know that a shape should be used for `shape_render`.

But if you show some sample liquid that someone could write and end up to be an `IHtmlContent` passed to `shape_render` I might be more open.

sebastienros: This example is actually only kept for backward compatibility. We will remove `DisplayAsync(dynamic)` in a later version, to keep the strongly types one.

TFleury: The `Form_Wrapper` (on top of this PR discussion) is an exemple of rendering `IHtmlContent`.
The implementation behind is `Microsoft.AspNetCore.Mvc.ViewFeatures.Buffers.ViewBuffer`.

I agree with you and it don't make sense to use `shape_render` on something that is not an `IShape`, but I didn't found an existing suitable filter for `IHtmlContent`. Maybe we need a new dedicated filter ?

sebastienros: {{ youHtmlContent }}

sebastienros: Now I am having doubts ;)

TFleury: > {{ youHtmlContent }}

It renders the type name

sebastienros: We need a converter then. The same kind we do for JObject, to convert an `IHtmlContent` to a `FluidValue`. I am approving this PR but we need to start another that removes this piece of code and provides a way to render an IHtmlContent.

sebastienros (approved):

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewTemplate.cs
```diff
@@ -117,6 +117,8 @@ public ShapeAccessor() : base(_getter)
                         return shape.Attributes;
                     case "Items":
                         return shape.Items;
+                    case "Metadata":
```
jtkech, line 120: Yes, all props defined by the `IShape` interface

For the switch i would say to use the new c# pattern, but not so good for the `default` if not a simple expression, but maybe worth to do it ;)

                return n switch
                {
                    nameof(IShape.Id) => shape.Id,
                    nameof(IShape.TagName) => shape.TagName,
                    ...
                    _ => (Func<object>)(() =>
                    {
                        if (shape.Properties.TryGetValue(n, out var result))
                        {
                            return result;
                        }
                        ...
                    })

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ShapeRenderFilter.cs
jtkech: One solution would be to make `Shape.Metadata.ChildContent` both an `IHtmlContent` and an `IShape` as the `PositionWrapper` is. But yes, `Shape.Metadata.ChildContent` is not an `IShape`, so looks better to have a filter / converter.

Note: From memory we already have a fluid value that is an `IHtmlContent` => `HtmlContentValue`

jtkech:
Maybe worth to wait a little, i will work on it this night by adding a converter as suggested in #8918

jtkech:
@TFleury 

Sorry, with the new github access rights, i was not sure to be able to commit directly to your repo, so I did another PR #8924 that adds the missing shape `Metadata` property access, and that also includes the fluid `IHtmlContent` converter.

jtkech:
So i close it, sorry again ;)

PR #8933, opened 2021-03-20
DetailAdmin Titles Tweaks

jtkech:
Not to be merged, just to show the difference

So because in the blog page i saw that the title is rendered twice with the same size (Title segment and part) while in the DetailAdmin display type.

DetailAdmin1

Here naively, but just to show the difference, i updated the Title segment text, and for the Title part i just added a view template for this display type, as already done specifically for the dashboard module.

DetailAdmin2


deanmarcussen:
See also @jtkech 

https://github.com/OrchardCMS/OrchardCore/pull/8839

need to add a stereotype default binding

jtkech:
Okay, i read it

In the meantime, here why not just change the text of the page title to not be the same as the blog title

Issue #8283, opened 2021-01-12
Remove Liquid helpers


sebastienros:
Replace all first-level liquid helper properties like Queries, Content, to actual filter to be more like Liquid

`Queries.RecentBlogPosts | query` => `"RecentBlogPost" | query`
`Content["4asdasd"]` => `"4adasdasd" | content_item`

jtkech:
I added the bug label as the current workaround to prevent a conflict between the liquid `Content` accessor and the shortcode `Content` property, will be removed when upgrading to fluid 2.0.

So we will need to do it, at least for the liquid `Content` accessor.

PR #8937, opened 2021-03-21
Shortcode / Liquid Content conflict

jtkech:
This is a known conflict

I know that we could replace the `Queries`, `Site`, `User` and `Content` accessors to liquid filters as suggested in #8283

But in the meantine, maybe worth to still provide a workaround for the `Content` property

Knowing that it is only related to shortcodes templates that are defined by using liquid syntax.

Here we don't modify the `LiquidContentAccessor` marker class, we just define a new local derived one

Allowing to both use the shortcode content and the liquid content acessor.

    Hello {{ Content }} : {{ Content.ContentItemId['4pnytr3heyd9rshtsb7w96z1nk'].DisplayText }}

    [hello]world[/hello] => Hello world : About

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Shortcodes/Services/TemplateShortcodeProvider.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Shortcodes/Services/TemplateShortcodeProvider.cs
@@ -14,9 +14,9 @@ namespace OrchardCore.Shortcodes.Services
         private readonly ShortcodeTemplatesManager _shortcodeTemplatesManager;
         private readonly ILiquidTemplateManager _liquidTemplateManager;
         private readonly HtmlEncoder _htmlEncoder;
-        private readonly HashSet<string> _identifiers = new HashSet<string>();
 
         private ShortcodeTemplatesDocument _shortcodeTemplatesDocument;
+        private readonly HashSet<string> _identifiers = new HashSet<string>();
 
         public TemplateShortcodeProvider(
             ShortcodeTemplatesManager shortcodeTemplatesManager,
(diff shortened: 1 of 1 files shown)


jtkech:
@deanmarcussen 

What you think about this one, much more simple than the previous workaround ;)

jtkech:
I will merge it, the fix is simple and `LiquidContentAccessor` is defined in `OC.Liquid.Abstractions`

And it allows to still use the following, before creating the 5 new related liquid filters in a separate PR, knowing that, if we want to follow the same convention, we may need to do the same for `Queries`, `SiteSettings` and `User` accessors.

    Content.ContentItemId[...]
    Content.ContentItemVersionId[...]
    Content.Latest[...]
    Content.Alias[...]
    Content.Slug[...]

MikeKry:
@jtkech  (maybe also @deanmarcussen  ?)

Two questions:
1) is it possible to mark PRs as breaking change with some label, so it would be easily visible that we have to check for some change?
2) this actually broke all my shortcodes using this:
image

I wonder if this is equivalent? Did not get it from code..

image

jtkech:
@MikeKry 

The liquid `Content` accessor fro shortcode templates was no more supported since the update to Fluid 2.0

Because it is planned to replace these accessors with full liquid filters

In the meantime i did this PR to re-introduce it for backwards compatibility

The syntax may have still changed, but from memory it's since a while

Yes, as i remember, try `Content.Alias[...]`

MikeKry:
@jtkech 

ok, thanks for explanation. 

btw. I looked up Fluid 2.0 PR but it also has no tag for breaking change or something, that could help me determine if upgrade is "safe" or not :(

deanmarcussen:
@MikeKry yes we should make more effort to mark these as breaking.

I would suggest it is worth keeping an eye on gitter, as it was announced and discussed in some detail there.

jptissot:
Maybe a pinned issue like Kevin did with the OpenId update would be useful.

deanmarcussen:
@jtkech @MikeKry 

Have we actually broken something here?

If the syntax has changed @jtkech why? and why not have the old syntax as well.

And is it just for the shortcode accessor that we have broken something, or the general `Content` accessor as well?

we could probably live with the shortcode accessor syntax changing, it was not in RC2, so dev is always at the risk of breaking things, but be good to know what exactly the issue is here.

jtkech:
@deanmarcussen @MikeKry 

About the syntax, just checked it for Alias, it seems to be the same since the beginning https://github.com/OrchardCMS/OrchardCore/commit/e3f1c65084ce443bce91665f7b34ff075b36df7a#diff-de1a41e30851e4888a3c9236bc7935be195e5bddeaee203869ddd498ffd242af

Then, just after Fluid 2.0  the liquid Content accessor from shortcode templates was no more supported

Here we just re-introduce it but by just defining a local internal wrapper that inherits from `LiquidContentAccessor`

Okay, no problem, in fact i just tried e.g. `Content.slug["about"]` but not `Content.Latest["slug:about"]`, but just saw that now we also have the following

                o.MemberAccessStrategy.Register<LiquidContentAccessor, LiquidPropertyAccessor>("Latest", (obj, context) =>
                {
                    var liquidTemplateContext = (LiquidTemplateContext)context;

                    return new LiquidPropertyAccessor(liquidTemplateContext, (name, context) =>
                    {
                        return GetContentByHandleAsync(context, name, true);
                    });
                });

So just tried also `Content.Latest["slug:about"]` through a shortcode template, and it still works

So no problem ;)

PR #8968, opened 2021-03-24
Auto enabling Feature dependencies

### LINKED ISSUE #8942: Test upgrade from RC2 with Shortcodes
Reported on gitter

When upgrading from RC2 the shape type `ShortcodesModal` is not found.

I had thought this was related to the renaming, but I suspect it is because the html editors now depend on shortcodes.

But when upgrading (i.e. on site startup) the site does not recheck it's dependencies.

I suspect we just need a migration, similar to what was done for layers / rules to enable the feature.

Need to check it for the content fields, html body part, and markdown body part probably.

So it might be a couple of migrations, just to be sure that each module activates the shortcodes feature.

Will check.

jtkech:
Fixes #8942 

@deanmarcussen about the 2 points we discussed in some issues threads about features dependencies, here we fix the first one about auto-enabling the missing features after some changes in the hard coded feature dependencies.

This at the source through the `ShellDescriptor` that is already auto-composed e.g. with the `_alwaysEnabledFeatures` and other `configuredFeatures`.

I tried it by first removing from the `Layers` the `Rules` dependency, then by disabling the `Rules` feature through the Admin. Then i re-add the hard coded `Rules` dependency and restarted the site, the `Rules` feature was auto-enabled and you can see it as an enabled feature in the Admin.

You don't see it in the `ShellDescriptor` in the database until it is saved again by enabling / disabling any feature, but this is not a problem as it is internally enabled, even to keep in sync running instances in a distributed env, each instance will build a `ShellDescriptor` taking into account the hard coded dependencies.

---

Would be good to try it for the migrations from rc2 without enabling new dependencies through migration steps.

Next step: Fix the migration execution order issue with the `RequireFeatures` attribute, maybe a separate PR.

Maybe i will remove the ShellStateManager / Updater / Coordinator, less database accesses, and just call the `IFeatureEventHandler` maybe in `ShellDescriptorFeaturesManager`, maybe in a separate PR.

### DIFF
--- a/src/OrchardCore/OrchardCore.Infrastructure/Shell/ShellDescriptorManager.cs
+++ b/src/OrchardCore/OrchardCore.Infrastructure/Shell/ShellDescriptorManager.cs
@@ -6,6 +6,7 @@ using Dapper;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using OrchardCore.Environment.Extensions;
 using OrchardCore.Environment.Shell.Configuration;
 using OrchardCore.Environment.Shell.Descriptor;
 using OrchardCore.Environment.Shell.Descriptor.Models;
@@ -23,6 +24,7 @@ namespace OrchardCore.Environment.Shell.Data.Descriptors
         private readonly IShellConfiguration _shellConfiguration;
         private readonly IEnumerable<ShellFeature> _alwaysEnabledFeatures;
         private readonly IEnumerable<IShellDescriptorManagerEventHandler> _shellDescriptorManagerEventHandlers;
+        private readonly IExtensionManager _extensionManager;
         private readonly ISession _session;
         private readonly ILogger _logger;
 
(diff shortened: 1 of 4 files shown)


deanmarcussen:
@jtkech I think you're onto something here.

I will try this branch as for both Layers and Shortcodes.

Here is a reference sqlite database configured on RC2 with a bunch of features enabled. 

https://github.com/OrchardCMS/OrchardCore/issues/8753

deanmarcussen:
RC2 `ShellDescriptor`

```
{
   "SerialNumber":6,
   "Features":[
      {
         "Id":"OrchardCore.ShortCodes" <-- name changed to `OrchardCore.Shortcodes
      }
}
```

Yep. Works.

I removed the migration in `Layers` that would enable the `Rules` feature. It enabled itself based on the dependency graph.

Same with the `Shortcodes` feature. In RC2 it was enabled as `OrchardCore.ShortCodes`. And it reenabled itself as `OrchardCore.Shortcodes`

jtkech:
@deanmarcussen Cool !!!

Maybe some last tweaks to do, i will check it this night

Yes, maybe a separate PR for the `RequiresFeatures`

PR #8994, opened 2021-03-27
Fix OpenIdApplicationStep to include ResponseType rst: permissions

ropstah:
… in [ApplicationController.cs](https://github.com/OrchardCMS/OrchardCore/blob/620d7f3e4ee1178f015fcf2b2939c056abec51cd/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/ApplicationController.cs#L203)

Without this, the `rst:` prefixed permissions are only added to the OpenIdApplication instance upon first "save" via Admin UI.


kevinchalet (approved):
Thanks for your PR! 🎉

PR #9001, opened 2021-03-28
Webpack 5

### LINKED ISSUE #9000: Fix webpack dependency
There is a security warning on a dependency from webpack. We might need to upgrade. In the graphql module.
- > 4.46.0
It's completely removed from 5.x also.

agriffard:
Update to Webpack 5, React 17.

https://www.npmjs.com/package/graphiql-explorer/v/0.6.3

Fixes #9000


agriffard:
itsover9000

jptissot:
I would not merge this as graphiql is still using webpack 4. Make sure to test everything ;)

sebastienros:
@Skrypt can you comment?

sebastienros:
It's a 1.0 issue

Skrypt:
It is risky but as we can see @agriffard has already updated these packages before.
If we compare with the GraphiQL Explorer repository packages.json file.

https://github.com/OneGraph/graphiql-explorer/blob/master/package.json

So, we can proceed if the GraphiQL explorer still works but it is risky still. We don't run any functional tests like they do and it could potentially break some features of the GraphiQL Explorer.

Question is, why do we need to update it to Webpack 5.0 if it's spec is to build it with 4.0?
Do we want to only have one Webpack version for all our assets absolutely?
Also, we should never need to build this GraphiQL Explorer at first, just use it.

Skrypt:
So the GraphiQL explorer is imported in a TypeScript file. Also dependent on it's bindings and TypeScript version and else. I mean, if @agriffard can confirm that it still works perfectly then let's merge it and cross fingers / close eyes. 😄

Skrypt:
Here is something maybe worth reading : 

https://github.com/OneGraph/graphiql-with-extensions

Skrypt:
Ok I tested this after doing a `webpack` on the module folder and it works. Though I noticed that some other PR just changed back the lockfileVersion to version 1. Added a commit to set them back to lockfileVersion 2.

Skrypt (approved):

Issue #8059, opened 2020-12-17
Admin: Key bindings to trigger button actions


agriffard:
In Create/Edit admin pages, provide some keyboard shortcuts to trigger the button actions.

Ex: Alt+S to save.

Handle the Draft/Publish and the 'and continue' cases.

/cc @hishamco

sebastienros:
If we do it please take a look at the common shortcuts in other web apps. At least WP, or apps that have got some accessibility work.

PR #9002, opened 2021-03-28
Add keyboard shortcuts for action buttons

hishamco:
Address #8059

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.PublishButton.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Views/Content.PublishButton.cshtml
@@ -18,3 +18,26 @@ else
         </div>
     </div>
 }
+
+<script at="Foot">
+    $(function () {
+        const altKeyCode = 18;
+        const sKeyCode = 83;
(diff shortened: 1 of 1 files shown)


hishamco:
Planning to create a `ShortcutTagHelper` to make it easier if it's possible

hishamco:
Seems TagHelper will make things complicated, @agriffard is this looks good to you, so I can continue to apply the shortcuts whenever its possible

deanmarcussen:
The idea looks good.

I don't understand wy you would make it a taghelper?
Two suggestions
- Before going too far you need to have a good check against reserved keyboard shortcuts in the monaco editor and codemirror editors, to make sure that nothing we do will conflict, and document what shortcut keys are actually free and not going to conflict. (even if this is only in the admin js to begin with, we would need to have it written down, or noone will know)
- Use data-attributes and a general script in the admin js, to select shortcuts based on those data attributes.

hishamco:
> I don't understand wy you would make it a taghelper?

I was thinking about TagHelpers and more precisely `TagHelperComponent`, at the end seems the easiest way is to use data attributes as you mentioned

PR #9005, opened 2021-03-29
Make DocumentManager a Tenant Singleton

jtkech:


src/OrchardCore.Cms.Web/Properties/launchSettings.json
```diff
@@ -12,7 +12,7 @@
       "commandName": "IISExpress",
       "launchBrowser": true,
       "environmentVariables": {
-        "ASPNETCORE_ENVIRONMENT": "Development"
+        "ASPNETCORE_ENVIRONMENT": "Production"
```
Skrypt, line 15: Why?

jtkech: Only testing code, need to be reverted

It prevents razor runtime compilation while in dev mode under VStudio

src/OrchardCore/OrchardCore.Infrastructure/Documents/DocumentManager.cs
```diff
@@ -40,13 +38,19 @@ public class DocumentManager<TDocument> : IDocumentManager<TDocument> where TDoc
             }
         }
 
+        public Type DocumentStoreServiceType { get; set; }
```
hishamco, line 41: Could be readonly property?

jtkech: Just tried but we can't as there are derived types that assign it

Skrypt:
hmm did you not change these recently?

jtkech:
No, DocumentManager is a scoped service since a while

It may be worth to make it a singleton to not have to resolve it on each request, but still need to be executed in a scoped context when first getting data from the database, and when updating data in the db, but now the lazy resolution of scoped services as `IDocumentStore` => `ISession` is done under the hood

Particularly for AutorouteEntries that is a singleton

@sebastienros asked me to do it

jtkech:
Hi @deanmarcussen 

Just for infos i did a little refactoring on `Sitemaps`, first because `SitemapEntries` depends on `SitemapManager` that depends on `SitemapEntries`, can be easily fixed by resolving it lazily, but also because of another reason.

The shared `SitemapRouteDocument` is built from the also shared `SitemapDocument`, and when getting the produced document we only check its identifier which doesn't ensure that it has been built from the last update of the source, so even if they both have an identifier, we don't check the consistency of the whole.

So here i merged the 2 services in only one to fix the circular dependency, and now the produced sitemap routes entries are no more in a shared document but cached localy by the service, and we check if the Identifier of the source document (that is still shared so we can use its identifier) has changed, if so we re-build these local entries. And, even if while doing this the source document is updated again, the identifier of the source will change again, and so on.

---

Also see my changes in `SitemapIndexTypeUpdateHandler` in the loop looks like you were using return in place of breaking the loop, and calling `UpdateSitemapAsync()` inside the parent loop in place of doing it after, as i saw it is done in `ContentTypesSitemapSourceUpdateHandler`. Also, in all these handlers i thought about checking if something was changed before calling `UpdateSitemapAsync()` but i let it as is.

Issue #8985, opened 2021-03-26
Remove Enable Rules feature from Layers migration


deanmarcussen:
@jtkech think it makes sense to remove the deferred task from the layers migration now as a result of https://github.com/OrchardCMS/OrchardCore/pull/8968

Can't think of a good reason to keep it ?

jtkech:
@deanmarcussen 

Yes, you can get rid of it ;)

PR #9028, opened 2021-03-31
Remove enable rules feature from layers migration

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/8985


Skrypt (approved):
Closing eyes and approving 😉

jtkech (approved):
I also want to fix the ordering issue when the `RequireFeatures` attribute is used ;)

I will try to do it soon but recently i saw other things that i think it's time and good to do first

PR #9036, opened 2021-04-02
Rule-based validation of form elements (Lombiq Technologies: OCORE-51)

### LINKED ISSUE #3980: Client-side validation of the form elements
In Orchard 1.x, you have the ability to add validation rules to form elements when building the form.
For example, you can have the required, minimum and max length attributes, use a custom validation message and decide to show a custom validation message instead. We can also add a regular expression here too.

In Orchard Core, you cannot have validation for the input fields in this way when using the `OrchardCore.Forms` module. One solution is to use a `ViewModel` with attributes that come from the `System.ComponentModel.DataAnnotations` namespace and when the form has been submitted, you can check the `ModelStateDictionary` for errors using the `IsValid` property.

.NET Core uses the [jQuery Unobtrusive Validation](https://github.com/aspnet/jquery-validation-unobtrusive) script to prevent submissions until the form is valid, ergo, [it's the recommended way to use client-side validation](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-2.2#client-side-validation). We could also provide a way for users to turn on/off the client-side validation for every input elements, that has validation enabled to prevent an unnecessary form post to a server and have a better user experience when working with forms.

SpikeXy:
Continuation of https://github.com/Lombiq/OrchardCore/pull/4.

Fixes #3980.

A recipe to see a demo of everything in this feature: https://gist.github.com/SpikeXy/8d9dc39fa3cde80076e5f9883c3e43f0 Note that you'll have to regenerate the Workflow URL and update that for the two forms.


agriffard:
Do you want to make a demo during the next meeting?

SpikeXy:
Piedone will demo it.

src/OrchardCore.Modules/OrchardCore.Forms/Entities/Scripting/ValidateRuleMethod.cs
```diff
@@ -0,0 +1,242 @@
...
+                    return str.Equals(compare, StringComparison.InvariantCultureIgnoreCase);
+                })
+            };
+            _isAfter = new GlobalMethod
+            {
+                Name = "isAfter",
+                Method = serviceProvider => (Func<string, string, bool>)((str, compare) =>
+                {
+                    var originResult = DateTime.TryParse(str, CultureInfo.InvariantCulture, DateTimeStyles.None, out var originDate);
+                    var compareResult = DateTime.TryParse(compare, CultureInfo.InvariantCulture, DateTimeStyles.None, out var compareDate);
```
Piedone, line 55: Such `DateTimes`s should be parsed with the request culture.

src/OrchardCore.Modules/OrchardCore.Forms/Entities/Scripting/ValidateRuleMethod.cs
```diff
@@ -0,0 +1,242 @@
...
+using System.Globalization;
+using System.Text;
+using System.Text.RegularExpressions;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using OrchardCore.Scripting;
+
+namespace OrchardCore.Forms.Entities.Scripting
+{
+    public class ValidateRuleMethod : IGlobalMethodProvider
```
Piedone, line 12: This could directly execute the JS library or just execute the server-side validation directly in C# and not as `GlobalMethod`. If we use JS scripting then a validator could be to use free JS code too.

src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/ValidationRulePart.Fields.Edit.cshtml
```diff
@@ -0,0 +1,40 @@
...
+        <option value="isLength">@T["IsLength"]</option>
+        <option value="isNumeric">@T["IsNumeric"]</option>
+        <option value="matches">@T["Matches"]</option>
+
+    </select>
+    <span class="hint">@T["The Validation type."]</span>
+</div>
+
+<div class="form-group" >
+    <label asp-for="Option">@T["Validation Options"]</label>
```
Piedone, line 30: It's not clear from the UI how Validation Options should be used. At least automatically populating it with an example for the selected rule (or adding a placeholder text to the input field for it) would help a lot.

Piedone (changes requested):
Added the feedback from the meeting.

Also, for the others, here's a recipe to see a demo of everything in this feature: https://gist.github.com/SpikeXy/8d9dc39fa3cde80076e5f9883c3e43f0 Note that you'll have to regenerate the Workflow URL and update that for the two forms.

SpikeXy:
It is ok now.

src/OrchardCore.Modules/OrchardCore.Forms/Workflows/Activities/ValidateRuleFormTask.cs
```diff
@@ -78,20 +72,20 @@ public override ActivityExecutionResult Execute(WorkflowExecutionContext workflo
...
-                    if (option.Contains("\\", StringComparison.Ordinal))
-                    {
-                        option = option.Replace("\\", "|-BackslashPlaceholder-|");
-                    }
-                    var formItemValue = httpContext.Request.Form[item.elementId.ToString()];
-                    var engine = _scriptingManager.GetScriptingEngine("js");
-                    var scope = engine.CreateScope(_scriptingManager.GlobalMethodProviders.SelectMany(x => x.GetMethods()), ShellScope.Services, null, null);
-                    var jsScriptStr = $"js: {type}('{formItemValue}','{option}')";
-                    if (!Convert.ToBoolean(engine.Evaluate(scope, jsScriptStr)))
+                    string type = item.type.ToString();
```
Piedone, line 75:
```suggestion
                    var type = item.type.ToString();
```

The below ones too.

src/OrchardCore.Modules/OrchardCore.Forms/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,124 @@
...
+using System.Text;
+using System.Text.RegularExpressions;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace OrchardCore.Forms.Extensions
+{
+    public static class StringExtensions
+    {
+        public static bool ValidateInputByRule(this string type, string input, string option)
```
Piedone, line 13: A couple of small issues here:
image

src/OrchardCore.Modules/OrchardCore.Forms/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,124 @@
...
+                    Single.TryParse(obj["min"]?.ToString(), out min);
+                    if (original >= min && (max == 0 || original <= max)) return true;
+                    return false;
+                case "isInt":
+                    return ValidateIs<int>(input);
+                case "isJSON":
+                    if (String.IsNullOrWhiteSpace(input)) return false;
+                    var value = input.Trim();
+                    if ((value.StartsWith("{") && value.EndsWith("}")) || //For object
+                        (value.StartsWith("[") && value.EndsWith("]"))) //For array
```
Piedone, line 57:
```suggestion
                    if ((value.StartsWith("{") && value.EndsWith("}")) || // For object.
                        (value.StartsWith("[") && value.EndsWith("]"))) // For array.
```

src/OrchardCore.Modules/OrchardCore.Forms/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,124 @@
...
+            {
+                return false;
+            }
+        }
+
+        public static bool CompareDatetime(string input, string option, string symbol)
+        {
+            var originResult = DateTime.TryParse(input, DateTimeFormatInfo.CurrentInfo, DateTimeStyles.None, out var originDate);
+            var compareResult = DateTime.TryParse(option, DateTimeFormatInfo.CurrentInfo, DateTimeStyles.None, out var compareDate);
+            if(symbol == ">") return originResult && compareResult && originDate > compareDate;
```
Piedone, line 120:
```suggestion
            if (symbol == ">") return originResult && compareResult && originDate > compareDate;
```

src/OrchardCore.Modules/OrchardCore.Forms/Assets/js/Validation/validationRuleHelper.js
```diff
@@ -75,4 +75,38 @@ function validationElementFunction(type, originElementId, option, validationMess
...
+    var selectedValidationOptionItem = e.options[e.options.selectedIndex];
+    var isShowOption = selectedValidationOptionItem.getAttribute('is-show-option');
+    var isShowValidationMessage = selectedValidationOptionItem.getAttribute('is-show-validationMessage');
+    var validationMessage = $(e.closest('.edit-item-parts')).find('.validation-message');
+    var validationOption = $(e.closest('.edit-item-parts')).find('.validation-option');
+
+    if (isShowOption === "0") {
+        validationOption.css('display', 'none');
+    } else {
+        validationOption.css('display', 'block');
```
Piedone, line 88:
```suggestion
    }
    else {
        validationOption.css('display', 'block');
```

Below too.

Piedone (changes requested):

SpikeXy:
it's ok.

src/OrchardCore.Modules/OrchardCore.Forms/Models/ValidationRuleModel.cs
```diff
@@ -0,0 +1,9 @@
+namespace OrchardCore.Forms.Models
+{
+    public class ValidationRuleModel
```
Piedone, line 3: I'd just use `ValidationRule` (it's already a model).

src/OrchardCore.Modules/OrchardCore.Forms/Assets/js/Validation/validationRuleHelper.js
Piedone: This is missing.

SpikeXy:
Sorry, what is missing?

src/OrchardCore.Modules/OrchardCore.Forms/Assets/js/Validation/validationRuleHelper.js
Piedone: <- This is missing :).

SpikeXy:
is this ok?

Piedone:
LGTM then! @agriffard would you also do your review?

agriffard:
> LGTM then! @agriffard would you also do your review?

No, I won't do any review.

Piedone:
Are you OK to merge this then, or is there anybody with a particular interest in Forms?

src/OrchardCore.Modules/OrchardCore.Forms/OrchardCore.Forms.csproj
```diff
@@ -11,10 +11,24 @@
     <PackageTags>$(PackageTags) OrchardCoreCMS</PackageTags>
   </PropertyGroup>
 
+  <ItemGroup>
```
sebastienros, line 14: Not necessary.

src/OrchardCore.Modules/OrchardCore.Forms/Views/Form.Wrapper.cshtml
```diff
@@ -7,16 +7,32 @@
...
     @if (formPart.EnableAntiForgeryToken)
     {
         @Html.AntiForgeryToken()
     }
 
     @await DisplayAsync(Model.Metadata.ChildContent)
 </form>
+
+<script at="Foot">
+    (function ($) {
```
sebastienros, line 25: It should not depend on jQuery.
Also this should be in a different file.

src/OrchardCore.Modules/OrchardCore.Forms/Workflows/Activities/ValidateRuleFormTask.cs
```diff
@@ -0,0 +1,97 @@
...
+            var httpContext = _httpContextAccessor.HttpContext;
+            var validateRule = true;
+            var rules = new List<ValidationRule>();
+            foreach (var item in httpContext.Request.Form)
+            {
+                if (item.Key.Equals("validationRules", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    if (!String.IsNullOrEmpty(item.Value[0]))
+                    {
+                        rules = JsonConvert.DeserializeObject<List<ValidationRule>>(item.Value[0]);
```
sebastienros, line 63: This fails the whole reason of the server-side validation. Anyone can hack the payload with custom validation that will invalidate any validation.

sebastienros:
There is the big security issue that is part of the core design. Another issue is that that would have been nice to create something that can be extended, by creating rules that define the client-side and server-side code. Some ideas might be
- have a dedicated controller that does the validation in ajax for the client, so the code is only in c#, and the Task would also use the same code
- or have some providers for both client-side and server-side equivalent of the validation, including the html editor (standard shape composition)

Another thing to look into is how Regex should always have a timeout to prevent DOS attacks with forged payloads. This is not the case currently, but this is minor compared to the security issue I pointed in the comments.

Piedone:
Sebastien also elaborate on these during the meeting: https://www.youtube.com/watch?v=TQLN7fQwN5I&t=2317s

src/OrchardCore.Modules/OrchardCore.Forms/Controllers/ValidationRuleController.cs
```diff
@@ -0,0 +1,87 @@
...
+            ValidationRuleHelpers validationRuleHelpers,
+            IStringLocalizer<ValidationApiController> stringLocalizer)
+        {
+            _authorizationService = authorizationService;
+            _validationRuleHelpers = validationRuleHelpers;
+            S = stringLocalizer;
+        }
+        [HttpGet]
+        [Route("ValidateFormByRule")]
+        public async Task<IActionResult> ValidateFormByRule(string contentItemId, string formName, string formValue)
```
sebastienros, line 35: Isn't it the opposite of what was advertised in the first place, and the form is validated server-side all the time? Or is it just getting a field of the form and not the whole form.

src/OrchardCore.Modules/OrchardCore.Forms/Helpers/ValidationRuleHelper.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            var contentTypeDefinition = _contentDefinitionManager.GetTypeDefinition(contentItem.ContentType);
+            if (contentTypeDefinition == null) return null;
+
+            foreach (var contentTypePartDefinition in contentTypeDefinition.Parts)
+            {
+                var partName = contentTypePartDefinition.Name;
+                var partTypeName = contentTypePartDefinition.PartDefinition.Name;
+                //var contentType = contentTypePartDefinition.ContentTypeDefinition.Name;
+                var partActivator = _contentPartFactory.GetTypeActivator(partName);
```
sebastienros, line 84: Can't `partName` be a named part in this case? Why not partTypeName?

src/OrchardCore.Modules/OrchardCore.Forms/OrchardCore.Forms.csproj
```diff
@@ -27,6 +43,7 @@
     <ProjectReference Include="..\..\OrchardCore\OrchardCore.Module.Targets\OrchardCore.Module.Targets.csproj" />
     <ProjectReference Include="..\..\OrchardCore\OrchardCore.ResourceManagement\OrchardCore.ResourceManagement.csproj" />
     <ProjectReference Include="..\..\OrchardCore\OrchardCore.Workflows.Abstractions\OrchardCore.Workflows.Abstractions.csproj" />
+    <ProjectReference Include="..\OrchardCore.Flows\OrchardCore.Flows.csproj" />
```
sebastienros, line 46: Requires a justification. Modules don't reference modules, it shows there is a wrong dependency, or that there is a need for a separate project.

src/OrchardCore.Modules/OrchardCore.Forms/Controllers/ValidationRuleController.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            if (String.IsNullOrWhiteSpace(contentItemId) || String.IsNullOrWhiteSpace(formName))
+            {
+                return BadRequest(S["contentItemId, formName are required parameters"]);
+            }
+
+            //if (!await _authorizationService.AuthorizeAsync(User, Permissions.ValidationRule))
```
sebastienros, line 42: No need for a permission to validate something, and the Permission class should then be removed too.

src/OrchardCore.Modules/OrchardCore.Forms/Providers/IsByteLengthProvider.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace OrchardCore.Forms.Providers
+{
+    public class IsByteLengthProvider : IValidationRuleProvider
+    {
+        public int Index => 6;
+        public string DisplayName => "Is ByteLength";
+        public string Name => "IsByteLength";
+        public bool IsShowOption => true;
+        public string OptionPlaceHolder => "{&quot;min&quot; : 0, &quot;max&quot; : 20}";
```
sebastienros, line 15: If it contains Html this should be an HtmlString, to handle encoding correctly, and set expectations.

src/OrchardCore.Modules/OrchardCore.Forms/Providers/IsByteLengthProvider.cs
```diff
@@ -0,0 +1,22 @@
...
+using OrchardCore.Forms.Helpers;
+using OrchardCore.Forms.Services;
+using OrchardCore.Forms.Services.Models;
+
+namespace OrchardCore.Forms.Providers
+{
+    public class IsByteLengthProvider : IValidationRuleProvider
+    {
+        public int Index => 6;
+        public string DisplayName => "Is ByteLength";
```
sebastienros, line 12: This might be a `LocalizedString`

src/OrchardCore.Modules/OrchardCore.Forms/Services/IValidationRuleProvider.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using OrchardCore.Forms.Services.Models;
+
+namespace OrchardCore.Forms.Services
+{
+    public interface IValidationRuleProvider
```
sebastienros, line 9: It looks like it should use the Options pattern, and configuration (startup) would register these options. That would make localized strings easier to define as you can inject the T in a Startup class, and also prevent DI usage.

src/OrchardCore.Modules/OrchardCore.Forms/Services/ValidationRuleServiceExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace OrchardCore.Forms.Services
+{
+    public static class ValidationRuleServiceExtensions
+    {
+        /// <summary>
+        /// Adds validation rule services.
+        /// </summary>
+        public static IServiceCollection AddValidationRules(this IServiceCollection services)
+        {
+            services.AddScoped<IValidationRuleService, ValidationRuleService>();
```
sebastienros, line 16: It might be useful to create a custom content item metadata for things that can be validated by returning their value in specific logical types, and each validator could handle these special logical types (datetime, numeric, text, boolean, json, html, any). Then a form widget could implement this metadata. However I haven't looked an how validators know which property then need to validate for a specific widget, so I might be mislead.

src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/ValidationRulePart.Fields.Edit.cshtml
```diff
@@ -0,0 +1,54 @@
+@using OrchardCore.Forms.ViewModels;
+
+@model ValidationRulePartEditViewModel
+@{
+    dynamic dyModel = Model;
```
sebastienros, line 5: There must be a better way to do it. Model is probably a shape here, hence `IShape` which has `Metadata`.

src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/ValidationRulePart.Fields.Edit.cshtml
```diff
@@ -0,0 +1,54 @@
...
+</div>
+
+<script asp-src="~/OrchardCore.Forms/Scripts/validationRuleHelper.min.js" debug-src="~/OrchardCore.Forms/Scripts/validationRuleHelper.js" asp-name="validationRuleHelper" depends-on="jQuery" at="Foot"></script>
+
+<script at="Foot">
+    (function ($) {
+        $(function () {
+            if ('@typeName' != '') {
+                var selector = 'select[name="@typeName"]'
+                validationTypeChange($(selector)[0], '@Model.Option', '@Model.ErrorMessage')
```
sebastienros, line 50: to prevent javascript injection, create dummy hidden divs with data attributes that contain the messages.

src/OrchardCore.Modules/OrchardCore.Forms/Assets/js/Validation/validationRuleHelper.js
```diff
@@ -0,0 +1,122 @@
...
+}
+
+function validationElementFunction(contentItemId, originElementId) {
+    var isIE = checkIE();
+    var elementId = '#' + originElementId;
+    var timestamp = new Date().getTime();
+    var validationElementId = originElementId + timestamp;
+    var submitBtn = $(elementId).closest('form').find(':submit')[0];
+
+    $(elementId).bind('keyup change', function (e) {
```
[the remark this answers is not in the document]
SpikeXy: yes, isn't that necessary?

src/OrchardCore.Modules/OrchardCore.Forms/Controllers/ValidationRuleController.cs
SpikeXy: both of form filed and  form will be validated by this function, every form field will request that api for validation before client form submit.

src/OrchardCore.Modules/OrchardCore.Forms/Helpers/ValidationRuleHelper.cs
SpikeXy: sorry, this has not completed yet.

src/OrchardCore.Modules/OrchardCore.Forms/wwwroot/Scripts/validationRuleHelper.js
```diff
@@ -0,0 +1,122 @@
...
+}
+
+
+function validationEvent(contentItemId, validationElementId, elementId, e) {
+    var validationElement = document.getElementById(validationElementId);
+    var elementName = $(elementId).attr('name');
+    var elementValue = $(elementId).val();
+
+    $.ajax({
+        url: validationRuleApi + "?contentItemId=" + contentItemId + "&formName=" + elementName + "&formValue=" + elementValue
```
sebastienros, line 21: POST (in body), and encoded

src/OrchardCore.Modules/OrchardCore.Forms/OrchardCore.Forms.csproj
```diff
@@ -11,10 +11,26 @@
...
 
+  <ItemGroup>
+    <None Include="wwwroot\Scripts\validationRuleHelper.js">
+      <CopyToPublishDirectory>Never</CopyToPublishDirectory>
+    </None>
+    <None Include="wwwroot\Scripts\validationRuleHelper.min.js">
+      <CopyToPublishDirectory>Never</CopyToPublishDirectory>
+    </None>
+  </ItemGroup>
+
```
jtkech, line 33: Yes, remove all these extra item groups, we have msbuild scripts that embeds resources including those under wwwroot

src/OrchardCore.Modules/OrchardCore.Forms/wwwroot/Scripts/validationRuleHelper.js
```diff
@@ -0,0 +1,122 @@
+
+'use strict';
+var validationSet = new Set();
+var validationRuleApi = "/api/validationApi/ValidateFormByRule";
+
```
jtkech, line 5: Need to be tenant aware (e.g. including a tenant url prefix), and to support a possible virtual folder

E.g. by using the `Url` helper in a razor file and a `data-*` attribute whose value can be retrieved by the script.

SpikeXy: ok

src/OrchardCore.Modules/OrchardCore.Forms/OrchardCore.Forms.csproj
SpikeXy: ok

src/OrchardCore.Modules/OrchardCore.Forms/OrchardCore.Forms.csproj
SpikeXy: ok,i removed it.

SpikeXy:
Now,this commit works fine,need i make a video demo?

JosNun:
Looks like there's been some great progress here towards dynamic forms. I've been poking at it a bit, and in `ValidateRuleFormTask.cs`, `form.ContentItemId` doesn't seem to exist on the form. It's possible that I've set something up wrong, and am missing a piece, but it doesn't look like the client is sending the `ContentItemId` in the FormData?

Also, what's the status of this? There seems to have been a good bit of progress. Just curious if there's any blockers before this gets merged.

Piedone:
Is this something you'd like to revisit @SpikeXy or should we close?

SpikeXy:
This commit can be closed , I'll pull out a separate module if anyone uses it.

[YuSheng.OrchardCore.Form.Validation](https://github.com/SpikeXy/YuSheng.OrchardCore.Form.Validation)

Piedone:
OK then.

PR #9037, opened 2021-04-02
Fix fluid operators parsing

sebastienros:


agriffard:
Please also update the version in docs/resources/libraries/Readme.md

hishamco:
> Please also update the version in docs/resources/libraries/Readme.md

We always forgot to update this, it would be nice if we can find a way to update them automatically

sebastienros:
Done

Skrypt (approved):

PR #8266, opened 2021-01-08
Monaco editor

MichaelPetrinolis:
Add monaco-editor (local and cdn) to resources, and use it as an editor for text and html fields. It depends on #8265 to load monaco loader and overcome the issue that only one anonymous define call per script file can be defined.

Related to #2345

PR #9041, opened 2021-04-02
Remove monaco dev build files

MichaelPetrinolis:
remove dev build js files from Resources. 
Should we also remove the js map files and/or rename the monaco folders (remove the min folder under monaco)? 
Map files are ~10MB uncompressed.


Skrypt:
Unless these map files are generated by our own gulpfile.js they are unecessary. The problem is that if any changes are made then the .map files will not be updated. We should not update these assets so keeping the .map files is unecessary.

MichaelPetrinolis:
@Skrypt these are the map files distributed with monaco editor and copied from node_modules through assets.json pipeline. But I agree they are unnecessary.

src/OrchardCore.Modules/OrchardCore.Resources/Assets.json
```diff
@@ -320,14 +313,6 @@
     "inputs": [
       "node_modules/monaco-editor/min/**/*"
```
Skrypt, line 314: Instead you should try using the dev.js file and instead of copying it just let it get minified by the gulp pipeline. So we will get ocmonaco.js and ocmonaco.min.js files.

src/OrchardCore.Modules/OrchardCore.Resources/ResourceManagementOptionsConfiguration.cs
```diff
@@ -428,14 +428,14 @@ ResourceManifest BuildManifest()
 
             manifest
                 .DefineScript("monaco-loader")
-                .SetUrl("~/OrchardCore.Resources/Scripts/monaco/min/vs/loader.js", "~/OrchardCore.Resources/Scripts/monaco/dev/vs/loader.js")
+                .SetUrl("~/OrchardCore.Resources/Scripts/monaco/vs/loader.js")
```
Skrypt, line 431: Here once you get the minified version you can add it back

src/OrchardCore.Modules/OrchardCore.Resources/ResourceManagementOptionsConfiguration.cs
```diff
@@ -428,14 +428,14 @@ ResourceManifest BuildManifest()
...
                 .SetPosition(ResourcePosition.Last)
                 .SetVersion(monacoEditorVersion);
 
             manifest
                 .DefineScript("monaco")
                 .SetAttribute("data-tenant-prefix", _tenantPrefix)
-                .SetUrl("~/OrchardCore.Resources/Scripts/monaco/min.js", "~/OrchardCore.Resources/Scripts/monaco/dev.js")
+                .SetUrl("~/OrchardCore.Resources/Scripts/monaco/ocmonaco.js")
```
Skrypt, line 438: Here once you get the minified version you can add it back.

MichaelPetrinolis:
@Skrypt Monaco has its own AMD loader to handle its resources. I think that if we create the *.min.js versions it will not work, as it will keep requesting the *.js files and not the *.min.js. Also @sebastienros suggested to remove source files in https://github.com/OrchardCMS/OrchardCore/pull/8266#issuecomment-812217858

Skrypt (approved):

PR #9046, opened 2021-04-02
Fixes #8785 #8967 #9064 Use DeferredTask in ContentStep

### LINKED ISSUE #8785: Enabling AdminDashboard feature with other SQL modifying features causes SQL mods to fail

### LINKED ISSUE #8967: Recipe enabling a module throws `SqliteException: SQLite Error 1: 'no such table: WorkflowTypeIndex'.`
### Describe the bug

Getting this exception when executing a recipe that enables the `OrchardCore.Workflows.Http` module.
Orchard version `1.0.0-rc2-15907`

### To Reproduce
1. Run our codebase: https://github.com/StatCan/StatCan.OrchardCore
2. Setup the site with the `HackathonTheme` setup recipe using SQLite
3. Go to the admin site and execute the `Contact form example` recipe from the Recipes admin page
4. Exception occurs. All requests fail.
5. Restart the site
6. Exception is gone but recipe did not fully complete
7. Run the `Contact form example` recipe a second time
8. Executes properly

### Expected behavior
No exception is thrown when the recipe enables the `OrchardCore.Workflows.Http` module.

Here is the recipe: https://github.com/StatCan/StatCan.OrchardCore/blob/056ccbf797d5e9ab79883e602e9bac149dfb4e99/src/Modules/StatCan.OrchardCore.VueForms/Recipes/ContactForm.recipe.json#L17

Setup recipe: https://github.com/StatCan/StatCan.OrchardCore/blob/5249e65dd61eeb1406b8db356773827d6a7071ef/src/Themes/HackathonTheme/Recipes/hackathon.setup.recipe.json#L11

/cc @jtkech

### LINKED ISSUE #9064: WorkflowTypeIndex not available for subsequent recipe steps?
### Describe the bug

I was using build 15800 for a while, and during that time, I built a web app that I could completely recreate on a fresh local tenant using a recipe via Blank recipe + Package Import. It would turn on a set of features, define content types, define workflows, and then create content items with forms that submit to the workflows.

I then updated to build 16039. Now, when I create a Blank tenant and run the recipe to recreate the app locally (for regression testing purposes), I find that the "Package Import" process fails when attempting to create the content items. The error I get is...

`SQLite Error 1: 'no such table: WorkflowTypeIndex'.`

This never happened previously, so I'm wondering if something changed to where the indexes' creation has been delayed or something? I would have thought the index would be created when the feature is turned on, which would have been completed at the end of the Feature recipe step, so the subsequent `content` step would be executing with the index already existing. Is there something I'm missing?

The recipe steps I have are...

1. Feature
    -  "Application.Default",
    - [CustomFormWidgets]
    - "OrchardCore.Liquid",
    - "OrchardCore.Scripting",
    - "OrchardCore.Workflows",
    - "OrchardCore.Workflows.Http",
    - [CustomLiquidFilters]
    - "OrchardCore.Microsoft.Authentication.AzureAD",
    - "OrchardCore.Resources",
    - "OrchardCore.Email",
    - "OrchardCore.Users",
    - "OrchardCore.Users.Registration",
    - [CustomAuthentication],
    - [CustomWorkflowActivities],
    - "OrchardCore.Settings",
    - "OrchardCore.Admin",
    - "OrchardCore.Templates",
    - "OrchardCore.Contents",
    - "OrchardCore.ContentTypes",
    - "OrchardCore.Alias",
    - "OrchardCore.Autoroute",
    - "OrchardCore.Shortcodes",
    - "OrchardCore.ContentFields",
    - "OrchardCore.ContentPreview",
    - "OrchardCore.Deployment",
    - "OrchardCore.Contents.Deployment.AddToDeploymentPlan",
    - "OrchardCore.Contents.FileContentDefinition",
    - "OrchardCore.CustomSettings",
    - "OrchardCore.Deployment.Remote",
    - "OrchardCore.Diagnostics",
    - "OrchardCore.DynamicCache",
    - "OrchardCore.Widgets",
    - "OrchardCore.Features",
    - "OrchardCore.Feeds",
    - "OrchardCore.Flows",
    - "OrchardCore.Forms",
    - "OrchardCore.HomeRoute",
    - "OrchardCore.Html",
    - "OrchardCore.Indexing",
    - "OrchardCore.Layers",
    - "OrchardCore.Lists",
    - "OrchardCore.Lucene",
    - "OrchardCore.Markdown",
    - "OrchardCore.Media",
    - "OrchardCore.Title",
    - "OrchardCore.Menu",
    - "OrchardCore.Navigation",
    - "OrchardCore.Recipes",
    - "OrchardCore.Roles",
    - "OrchardCore.Shortcodes.Templates",
    - "OrchardCore.Taxonomies",
    - "OrchardCore.Taxonomies.ContentsAdminList",
    - "OrchardCore.Themes",
    - "TheAdmin",
    - [CustomAdminTheme],
    - [CustomFrontendTheme]
2. WorkflowType
3. Templates
4. ContentDefinition
5. Themes
6. CustomStepToRefreshWorkflowHttpRequestTokens
7. content
    - main menu
    - side menu
    - form1
    - form2

The recipe executes successfully if I just remove the last `content` step, but it fails if the `content` step is included. Aside from the forms having some custom shortcodes in an HtmlBodyPart, they are typical run-of-the-mill forms without any custom functionality. This leads me to think the problem lies somewhere else, especially since it all worked fine before updating to the latest dev build.

I also wondered if maybe the recipe deployment plan made from a 15800 build was having issues being imported into a 16039 build? But the content/format looked largely the same.

### To Reproduce
1. Have similar recipe steps executed on a Blank fresh tenant with a SQLite database? Not sure how best to reproduce it.

### Expected behavior
Able to successfully import content.

jtkech:
Maybe related to #8785 and #8967 

**Update**: Fixes #8785 fixes #8967 Fixes #9064

Hi @deanmarcussen 

So here we adjust the dependencies order of the shell blueprint for the type / feature pairs having required features defined through the `RequireFeatures` attribute. E.g. the types of the `OC.Contents` feature that require the `OC.Workflows` feature will be after the types of `OC.Workflows`, e.g. for service registrations (but not only)

Some maybe it fixes the 2 related issues because in the stack trace of both issues we have the following

      IContentHandler thrown from OrchardCore.Contents.Workflows.Handlers.ContentsHandler by SqliteException
      Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 1: 'no such table: WorkflowTypeStartActivitiesIndex'.

But even if the `OC.Contents.Workflows.Handlers` is registered after the workflows feature services, it doesn't prevent a given migration of any other feature to use a content step before the execution of the worflows migration.

So maybe this PR only fix a part of the issue but it is still good to have this ordering adjustment, e.g. if a migration is registered through a startup class type that uses the `RequireFeatures` attribute, this migration will run after the ones of the required features

And this can be used as a workaround in some cases, e.g. if you have a migration that uses the content step (through a migration recipe) and you know that the WF will be enable in the same migration recipe, the workaround would be to register the migration by using a startup type requiring the WF feature, with this PR it will run after the WF migration.

I will work on these remaining things and see if it has to be part of this PR or a separate one.


jtkech:
@deanmarcussen Just for infos

About #8785 and #8967 i think i found how to fix them. As said originally migrations were used for database migrations and content definitions, not e.g. for creating contents. So i tried something that i thought about before, use a deferred task for the recipe content step and it was working.

It also fixes the recipe steps ordering issue if in a given recipe you define a content step before a type definition step, and a given content step that imports a set of content items is done in its own shell scope.

Maybe there are other recipe steps that would need to be deferred after migrations, will see.

So now i'm not sure if we keep the adjustement of the dependencies ordering based on the `RequireFeatures`, maybe better to not do it in term of perfs, but this is only done when  creating a shell container, so i will think about it.

**Update**: But then the created items was not lucene indexed because they are indexed also in a deferred task and for now a deferred task can't trigger a deferred task. So the unit tests were failing, I was avoiding this but normally a deferred task should not trigger a deferred task, so here it is now allowed, and now the created items are also lucene indexed.

jtkech:
So, i reverted the `RequireFeatures` dependencies re-ordering because, even if it may be good, it doesn't fix the mentioned issues, so i kept it on my side to be re-used if we really need it to fix other issues. So here

- I did some little cleanups and refactorings, e.g. i removed the `CompiledFeatureEntry` and `NonCompiledFeatureEntry` classes that we created when we were using dynamic compilation.

- Here to fix the mentioned issues the main change (e.g. in a recipe migration) is to execute the `ContentStep` in a deferred task so that all migrations have completed, and to allow a `ShellScope` deferred task to trigger another deferred task so that e.g. content items created in a deferred task are still Lucene indexed in another deferred task.

- And as said i reverted the changed made in `CompositionStrategy` for required dependencies ordering.

deanmarcussen:
@jtkech so a deferred task for the content step. Interesting.

What happens if the step throws an exception? (like a validation exception for example)

I've not tracked what happens with exceptions during a deferred task, but suspect it may no longer be caught?

jtkech:
@deanmarcussen 

We don't use a deferred task if the shell is already activated, meaning not through a migration, otherwise it is executed after all migrations in a new shell scope but where errors are logged and exceptions are thrown as in any other scope.

deanmarcussen:
Cool, ok.

So in the case of a content recipe step error from a migration the migration will complete and increment, and save despite the failure? 
because it’s a new scope, and the session was committed in the prior scope?

is ok with me, just clarifying behavior, so we know what issues to expect ;)

jtkech:
Yes, the migrations complete, only the content items import would fail

jtkech:
In summary we prevent content item creations during definition alterations and database table migrations ;)

PR #9059, opened 2021-04-03
Upgrading to YesSql 3

sebastienros:


sebastienros:
I want to be able to explain the differences and the resulting breaking changes before we merge it.

sebastienros:
image

PR #9062, opened 2021-04-04
Fixes #9039 view buffer OutOfRangeException

### LINKED ISSUE #9039: Specified argument was out of the range of valid values. (Parameter 'index')
i saw some ticket about this issue , but i don't know if main same or not.

After update to new version, i get this error in showing blog and my module views too.

```
 
ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'index')
Microsoft.AspNetCore.Mvc.ViewFeatures.Buffers.ViewBufferTextWriter.Write(char[] buffer, int index, int count)
 
```
repo:
it works with: 
1- create new blog
2- add new blog post
3-create new .liquid them page "Content-Blog.liquid"
```

 <div class="container">
        <div class="row">
			{{ Model.Content | shape_render }}
		</div>
  </div>
```
it works fine with new blog and post but old blog doesn't work.
It looks it has issue in "{{ Model.Content | shape_render }}". If i remove this line, It shows my page.

jtkech:
Fixes #9039

@sebastienros 

When rendering a liquid view we write directly to the underlying razor page output. So, when we faced the aspnetcore (encoding / buffer) issue, i first tried to render it through our custom view buffer before writing to the page output.

But our view buffer had another issue (because of rented chars), so as a temporary workaround i was rendering the liquid template in an intermediate string, which is not good in term of allocations.

In the meantime you fixed our view buffer writer, so we re-used it in the liquid template manager and some other places as the `FluidTagHelper` and `ZoneTag`. But still not when rendering a Liquid view where we now write again directly to the underlying razor page output.

**So, because in aspnetcore there are still some components that encode directly to the writer in place of using an intermediate string, mainly the `HtmlContentBuilder` and the `TagBuilder`, we still have some use cases where the aspnetcore view buffer writer throws an `ArgumentOutOfRangeException`.**

I could repro #9039 and other ones e.g. when a liquid view renders a razor view using e.g. a `TagBuilder`.

So, to fix these issues one by one, i needed to add in many places the following, in `ResourcesTag`, `ScriptTag`, `StyleTag`, also in the `ShapeRenderFilter` and / or in `HtmlContentValue`, and so on.

            var content = new ViewBufferTextWriterContent();
            buffer.WriteTo(content, (HtmlEncoder)encoder);
            content.WriteTo(writer, (HtmlEncoder)encoder);

We only face this issue under Liquid because aspnetcore uses intermediate writers / buffers and / or check if it is a view buffer writer to behave differently. **So to have a more global solution in this PR, in place of doing the above in many places, here i just did it when rendering a Liquid view page.**

            var content = new ViewBufferTextWriterContent();
            await context.EnterScopeAsync(page.ViewContext, (object)page.Model);
            await template.FluidTemplate.RenderAsync(content, htmlEncoder, context);
            content.WriteTo(page.Output, htmlEncoder);

For me it makes sense to use our view buffer writer as a relay before writing to the page output, and only once per liquid page in place of having to use it in many other places, and maybe more in the future, and because it seems to fix all the issues i could repro before. **Let me know if you think it's okay in term of perfs / allocations**.

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ShapeStringifyFilter.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ShapeStringifyFilter.cs
@@ -1,16 +1,11 @@
 using System;
-using System.Collections.Generic;
 using System.IO;
-using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 using Fluid;
 using Fluid.Values;
 using Microsoft.AspNetCore.Html;
-using Microsoft.AspNetCore.Mvc.Localization;
 using Microsoft.AspNetCore.Razor.TagHelpers;
-using Microsoft.Extensions.DependencyInjection;
 using OrchardCore.Liquid;
-using OrchardCore.Mvc.Utilities;
 
 namespace OrchardCore.DisplayManagement.Liquid.Filters
 {
@@ -27,7 +22,9 @@ namespace OrchardCore.DisplayManagement.Liquid.Filters
         {
             static async ValueTask<FluidValue> Awaited(Task<IHtmlContent> task)
             {
-                return new HtmlContentValue(await task);
+                using var writer = new StringWriter();
+                (await task).WriteTo(writer, NullHtmlEncoder.Default);
+                return new StringValue(writer.ToString(), false);
             }
 
             if (input.ToObjectValue() is IShape shape)
@@ -38,17 +35,12 @@ namespace OrchardCore.DisplayManagement.Liquid.Filters
                     return Awaited(task);
                 }
 
-                StringValue value;
-                using (var writer = new StringWriter())
-                {
-                    task.Result.WriteTo(writer, NullHtmlEncoder.Default);
-                    value = new StringValue(writer.ToString(), false);
-                }
-
-                return new ValueTask<FluidValue>(value);
+                using var writer = new StringWriter();
+                task.Result.WriteTo(writer, NullHtmlEncoder.Default);
+                return new ValueTask<FluidValue>(new StringValue(writer.ToString(), false));
             }
 
-            return new ValueTask<FluidValue>(NilValue.Instance);
+            return new ValueTask<FluidValue>(StringValue.Empty);
         }
     }
 }
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ShapeStringifyFilter.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ShapeStringifyFilter.cs
@@ -1,4 +1,3 @@
-using System;
 using System.IO;
 using System.Threading.Tasks;
 using Fluid;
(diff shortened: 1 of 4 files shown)


src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ShapeStringifyFilter.cs
```diff
@@ -27,7 +21,9 @@ public ValueTask<FluidValue> ProcessAsync(FluidValue input, FilterArguments argu
         {
             static async ValueTask<FluidValue> Awaited(Task<IHtmlContent> task)
             {
-                return new HtmlContentValue(await task);
+                using var writer = new StringWriter();
```
sebastienros, line 24: Not good

sebastienros: If the goal to invert the call with the encoder, this is not necessary anymore. It doesn't call it char by char like before

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ShapeStringifyFilter.cs
```diff
@@ -38,17 +34,12 @@ static async ValueTask<FluidValue> Awaited(Task<IHtmlContent> task)
...
 
-                StringValue value;
-                using (var writer = new StringWriter())
-                {
-                    task.Result.WriteTo(writer, NullHtmlEncoder.Default);
-                    value = new StringValue(writer.ToString(), false);
-                }
-
-                return new ValueTask<FluidValue>(value);
+                using var writer = new StringWriter();
```
sebastienros, line 37: If it is not written to the page output directly it doesn't need to be buffered anymore since you are assume a global buffer for liquid now. Or is it not just liquid here?

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ShapeStringifyFilter.cs
jtkech: It is the `shape_stringify` filter, not the `shape_render`, it is not intended to be written to the buffer, but e.g. to assign a local liquid var. It was already using a `StringWriter` below when the task completed immediately

Here in the `Awaited()` method it was using an `HtmlContentValue` because i assume a bad copy / paste from the `ShapeRenderFilter`

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/ShapeStringifyFilter.cs
jtkech: It is the `shape_stringify` filter, not the `shape_render`, it is not intended to be written to the buffer, but e.g. to assign a local liquid var. It was already using a `StringWriter` and just did here a little refactoring

And above in the `Awaited()` method it was using an `HtmlContentValue` because i assume a bad copy / paste from the `ShapeRenderFilter`, so i updated it to do the same

jtkech: It is rarely used compared to the shape render filter

sebastienros:
Ok to move the `ViewBufferTextWriterContent` centrally, but I don't see it being removed from where it's used right now.

jtkech:
I think it is okay, the `ViewBufferTextWriterContent` is only used when there is no writer parameter or when we need to produce an `IHtmlContent` that need to be rendered later in another context.

- In the `ZoneTag` to produce an `IHtmlContent` that we add to a given Layout Zone.

- In the `FluidTagHelper` to produce an `IHtmlContent` for the child content of a given tag helper.

- Finally in the `LiquidTemplateManager.RenderHtmlContentAsync()` that returns an `IHtmlContent`.

Note: When rendering a given web page, the rendering of our Liquid page is triggered multiple times, for all the liquid view template files that are involved in the page rendering.

sebastienros (approved):

PR #9068, opened 2021-04-06
Fix layer rules import step

deanmarcussen:
Fixes an issue importing more complex types, than the defaults.

Discussion is here https://github.com/OrchardCMS/OrchardCore/discussions/9011#discussioncomment-573776

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Layers/Deployment/AllLayersDeploymentSource.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Layers/Deployment/AllLayersDeploymentSource.cs
@@ -1,5 +1,6 @@
 using System.Linq;
 using System.Threading.Tasks;
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using OrchardCore.Deployment;
 using OrchardCore.Entities;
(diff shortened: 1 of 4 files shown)


Skrypt:
@intellimedhu Please confirm it fixes your issue.

intellimedhu:
Using the JSON ([here](https://github.com/OrchardCMS/OrchardCore/discussions/9011#discussioncomment-573776)), I get an exception:

https://github.com/OrchardCMS/OrchardCore/blob/0fe03ee8892afcdac23c4dddba47fb1f6cacf564/src/OrchardCore.Modules/OrchardCore.Layers/Recipes/LayerStep.cs#L91


image

intellimedhu:
Works fine when removing this object:

```
"Operation": {
    "CaseSensitive": false
}
```

Should ignore "Operation"?

deanmarcussen:
ok, thanks for checking. (thought I'd tested that, but obviously not). Probably need to do a type lookup from the factories. Will look later

deanmarcussen:
@intellimedhu have had to include type handling for the operations. 

You'll need to adapt your json to include the types for the operations (or recreate from the deployment plans, which is now including it automatically)

PR #9075, opened 2021-04-06
Performance optimizations

sebastienros:


src/OrchardCore/OrchardCore.Contents.Core/ContentTypePermissionsHelper.cs
```diff
@@ -44,6 +44,8 @@ public class ContentTypePermissionsHelper
             { CommonPermissions.ListContent.Name, ListContent }
         };
 
+        public static Dictionary<ValueTuple<string, string>, Permission> PermissionsByType = new Dictionary<ValueTuple<string, string>, Permission>();
```
deanmarcussen, line 47: Maybe should be an immutable dictionary?

sebastienros: That would help to not have to clone the current one when a new dynamic permission is added. However there is a finite number of permissions added, and an unbounded number of lookups. It happens that immutable dictionaries are much slower on lookups, so when I can I prefer the cloning technique. No locks, no concurrent dictionary, no immutable one. It depends on each usage.

deanmarcussen (approved):

PR #9081, opened 2021-04-06
Use PoParser.Core

hishamco:
The aim is to use a new parser for portable objects that uses Parlot behind the scene to make it faster


test/OrchardCore.Tests/Localization/PortableObjectParserTests.cs
```diff
@@ -100,17 +100,17 @@ public void ParseOnlyTrimsLeadingAndTrailingQuotes()
...
         }
 
-        [Fact]
-        public void ParseHandleUnclosedQuote()
-        {
-            // msgctxt "
-            // msgid "Foo \"{0}\""
-            // msgstr "Foo \"{0}\""
+        //[Fact]
+        //public void ParseHandleUnclosedQuote()
```
hishamco, line 104: Handle unclosed quote is special implementation I see in OC, we can customize tha `PoParser` to support it out of the box

hishamco:
Benchmarks may added later to measure the PERF

test/OrchardCore.Tests/Localization/PortableObjectParserTests.cs
hishamco: IMHO this case should be ignore coz it break the syntax, but we could handle it by custom parser implementation

hishamco:
FYI the new parser need some improvement in terms of PERF, for now it expose a rich set of metadata about the parsed PO file as `SyntaxNode`s

hishamco:
@Skrypt you could have a look if you are interested ;)

sebastienros:
How faster is it? Not that it makes sense for parsing PO files to be faster, but that's the only reason invoked.

hishamco:
I don't think it's fast as native parsing, by the new library still need some improvement in terms of PERF, coz I'm still have many question about Parlot - may be I did something not recommended-

The thing that is good about this library is exposing the entire syntax tree, so the developer can easily work with the nodes themselves something similar to Irony, but agian the PERF is crucial

One more thing is related to pasring and Parlot, could you please have a quick look to [SqlParser](https://github.com/hishamco/SqlParser) which I already started in #8215 few months ago, still there are some work to finalize few things remaining in SELECT, but I need your recommendation in case something goes wrong

Thanks

jptissot:

hishamco:
I will close this for now

PR #9083, opened 2021-04-07
Remove liquid type from recipes

deanmarcussen:
Due to concerns about sanitization it was decided to remove the Liquid Content Type, and Liquid Widget Type from the sample recipes.

This because liquid is not sanitizable, as you can do too many things with it.

The liquid part still exists, but the type needs to be created at the discretion of an administrator.

On this pr also, I have cleaned up the recipes, removing a duplicate Image Widget, and duplicate MenuItemPart that had managed to sneak into the agency theme.

So what looks like lots of changes, is mostly cleaning them up.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentTypes/Permissions.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentTypes/Permissions.cs
@@ -7,7 +7,7 @@ namespace OrchardCore.ContentTypes
     public class Permissions : IPermissionProvider
     {
         public static readonly Permission ViewContentTypes = new Permission("ViewContentTypes", "View content types.");
-        public static readonly Permission EditContentTypes = new Permission("EditContentTypes", "Edit content types.");
+        public static readonly Permission EditContentTypes = new Permission("EditContentTypes", "Edit content types.", isSecurityCritical: true);
 
         public Task<IEnumerable<Permission>> GetPermissionsAsync()
         {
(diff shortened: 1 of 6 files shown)


Skrypt:
These ends up to be created most of the time but for these themes I guess they are not required. Not sure what you meant by sanitization?

sebastienros:
The `EditContentTypes` permission is not marked as security critical. Should we change this also, or add a custom permission to any part that could be abused? Easiest is the former.

sebastienros:
Related suggestion https://github.com/OrchardCMS/OrchardCore/issues/9111

deanmarcussen:
Funny I very nearly moved all the widgets from agency and blog into a shared reusable recipe file to sort some consistency out. Might just do that then here then.

and yes probably mark edit types as security critical is easiest.

@Skrypt the idea being we can’t prevent script injection with liquid, so at users discretion to create the type

Skrypt:
Ok for a recipe and any added security for these.

deanmarcussen:
> The `EditContentTypes` permission is not marked as security critical. Should we change this also, or add a custom permission to any part that could be abused? Easiest is the former.

I'm just reviewing this suggestion. 
Not sure we should actually do this.
'security critical' currently means be wary of assigning this permission, as it can be used to elevate other permissions. i.e. Assign Roles, or Import / Export, can be used to upgrade a role, to a higher set of permissions etc.

Here we are trying to say be careful who you give this content type too, as you can use it to add javascript to a page.
So possibly a different thing, i.e. we would need to add the same kind of warning to `Templates` / `Shortcode Templates`, as they are places that this could be done as well.

It could also be another type of warning, 'security vulnerable' perhaps, as a lower level of warning, than 'security critical'

Or also it could just be a note on the hint for the `LiquidPart`

deanmarcussen:
Notes from discussion with @sebastienros 

- Use security critical as an indicator on more features
- Content Type editing
- Template editing
- Shortcode template editing

sebastienros (approved):

PR #9106, opened 2021-04-09
Added remote deployment permissions

### LINKED ISSUE #9101: Remote deployment roles missing
Roles for remote deployment can't be set, they aren't in admin at all, probably because they are not registered in startup.cs? OrchardCore.Modules/OrchardCore.Deployment.Remote/Permissions.cs

Because of this, only admin can deploy to another environment. Is there a reason for that, or can I fix it and create PR?

MikeKry:
Added missing permissions setup

image

Fixes #9101


deanmarcussen (approved):
For future reference @MikeKry if you 'fix' an issue, if you put `Fixes` and the link to the issue, github will link and close it on merge.

(we tend to use Related, when there maybe multiple things on the issue, and don't want to close it)

hishamco (approved):

MikeKry:
@deanmarcussen 
Ok, I did not know about that as I do not use github, only for OC :) Thanks for info, I will try to remember that

PR #9108, opened 2021-04-09
Elasticsearch / OpenSearch feature :sparkles::boom:

### LINKED ISSUE #4316: Implement ElasticSearch module
Same as Lucene

Sen-Gupta:
Fixes #4316

### Install ElasticSearch or OpenSearch with Docker compose

OpenSearch Docker Compose file : 

[opensearch.txt](https://github.com/OrchardCMS/OrchardCore/files/7174902/opensearch.txt)

ElasticSearch Docker Compose file : 

[elasticsearch.txt](https://github.com/OrchardCMS/OrchardCore/files/7197469/elasticsearch.txt)

- Copy these files in a folder named Docker somewhere safe.
- Rename these files extension as .yml instead of .txt.
- Open up a Terminal or Command Shell in this folder.
- Execute `docker-compose -f opensearch.yml -p opensearch up` to deploy OpenSearch containers.
- Wait for the containers to be fully created.
- Stop the docker containers with CTRL+C in the command shell. This will stop the containers from running.
- Execute `docker-compose -f elasticsearch.yml -p elasticsearch up` to deploy ElasticSearch containers.
- Always execute only one or the other because they use the same external 9200 port.

Advice : don't remove these files from their folder if you want to remove all their containers at once later on in Docker desktop.

You should get this result in Docker Desktop app : 

133547386-5af33f19-e5a1-426e-8f11-4739bcadc691

### Set up ElasticSearch or OpenSearch in Orchard Core

- Add Elastic Connection in the shell configuration (OrchardCore.Cms.Web appsettings.json file)

```JSON
"OrchardCore_Elastic": {
    "Url": "http://localhost:9200"
}
```

- Start an Orchard Core instance with VS Code debugger
- Go to Orchard Core features, Enable ElasticSearch.

## Implementation details

**_Analyzed and Stored Properties are not very meaningful in context of ElasticSearch._**

### Analyzed 

Analyzed is default for strings in Elastic Search.
By default all string fields are stored twice in elastic as "analyzed" and stored as "text" field type of elastic and again stored as is as in a "keyword" field type of elastic.

So we will have a field called ContentItemId(Text) analyzed and another called ContentItemId.Keyword(as is as) to match on exact values using TermQuery for fields like ContentItemId or emails (Elastic Stores text fields in 2 fields analyzed vs not analyzed, a field ContentItemId.Keyword is created automatically)

ElasticSearch documentation: 
https://www.elastic.co/blog/strings-are-dead-long-live-strings
https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-index-search-time.html

### Stored 

Stored is really an overhead and only required if we are processing thousands of large documents.
By default Elastic will store the entire document into a field called **__source_** and retrieves them when asked them from Index itself.

ElasticSearch documentation: 
https://www.elastic.co/guide/en/elasticsearch/reference/7.12/search-fields.html

### DSL Query Syntax

It is suggested to always use MatchQuery instead TermQuery for text fields in Elastic, where fully confident use (.Keyword) fields for exact match with TermQuery. (e.g. matching id, or fields like email, phone number, hostname)
 
ElasticSearch documentation: 
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-term-query.html

## TODO
- [ ] Fix Taxonomy Field indexing. (Added "Inherited" in IgnoredFields; may need a better fix). :bug:
- [ ] Refactor Search form to support both Lucene and Elastic. :construction:
- [ ] Restructure project names using OrchardCore.Search.X :recycle: :truck:
- [ ] Add Advance Elastic Configuration. (Cluster connection strings) - Optional (Can be done in Kibana or OpenSearch Dashboards)
- [x] Add Advance Support for multilingual search. (Equivalent of Lucene Analyzers) :sparkles:
- [x] Refactor UI to hide Analyzed/Stored in case of Elastic as Indexing Options for Fields. :recycle: :lipstick:
- [ ] Refactor ContentIndexSettings as IContentIndexSettings : See https://github.com/OrchardCMS/OrchardCore/pull/10515
- [ ] Module documentation. :memo:


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=9108) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=9108) before we can accept your contribution.<br/><br/>:x: Sen-Gupta [sign now](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=9108)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OrchardCMS/OrchardCore?pullRequest=9108) it.</sub>

Sen-Gupta:
Please tag as Do not merge!

Skrypt:
Massive PR. Awesome work. Need to review 😉

jtkech:
@Sen-Gupta 

Thanks for this work, it may fix one of the remaining concerns when running in a distributed environment

src/OrchardCore.Cms.Web/appsettings.json
```diff
@@ -80,6 +80,9 @@
     //  "ChangePasswordUrl": "ChangePassword",
     //  "ExternalLoginsUrl": "ExternalLogins"
     //},
+    "OrchardCore_Elastic": {
```
sebastienros, line 83: To be removed before merge

src/OrchardCore.Modules/OrchardCore.Search.Elastic/Startup.cs
```diff
@@ -0,0 +1,181 @@
...
+            services.AddShapeAttributes<SearchShapes>();
+        }
+
+        public override void Configure(IApplicationBuilder app, IEndpointRouteBuilder routes, IServiceProvider serviceProvider)
+        {
+            routes.MapAreaControllerRoute(
+                name: "Elastic.Search",
+                areaName: "OrchardCore.Search.Elastic",
+                pattern: "Search",
+                defaults: new { controller = "Search", action = "Search" }
```
sebastienros, line 99: This is conflicting with the Lucene module, does it mean we shouldn't have both enabled at the same time?

src/OrchardCore.Cms.Web/appsettings.json
sebastienros: But needs to be documented somewhere.

sebastienros: Also is there a notion of tenant, or store, such that it could automatically use the tenant name to create an isolated for each tenant, on the same ES deployment?

Skrypt:
My first test of building an index gets me an error.

image

Sen-Gupta:
@Skrypt , please provide me the ContentTypes and Contents that you included for indexing. I'm able to do build and run MatchAll query perfectly. It seem it has something to do with ContentTypes or the fields included!

src/OrchardCore.Cms.Web/appsettings.json
Sen-Gupta: @sebastienros  Notion of Tenants Yes: https://github.com/OrchardCMS/OrchardCore/pull/9108/commits/3f946a4bb62ca0f46616d2a6e9f91f1a7c9c483d

Skrypt:
I used TheBlogTheme recipe. I'm indexing Blog, BlogPost and Article content types. Seems like the issue is related with the FullText custom field.

Sen-Gupta:
> I used TheBlogTheme recipe. I'm indexing Blog, BlogPost and Article content types. Seems like the issue is related with the FullText custom field.

Trying to reproduce! 

A colleague of mine, also complained about it, i believe he was using the same theme, I'm on agency.

Should be a quick fix. The FullText Field is anyway not required by Elastic, it has it's own way to support FullText Search.

Skrypt:
Right now I'm trying to test this but the ElasticSearch docker container takes half of my PC ram (around 7 gb). So debugging this is not the best experience.

Skrypt:
Ok I confirm that this works if I'm removing the FullText config on the Article Content Type and only allow this Content Type to be indexed in the index.

src/OrchardCore.Modules/OrchardCore.Search.Elastic/OrchardCore.Search.Elastic.csproj
```diff
@@ -0,0 +1,46 @@
...
+    <ProjectReference Include="..\..\OrchardCore\OrchardCore.Search.Elastic.Core\OrchardCore.Search.Elastic.Core.csproj" />
+    <ProjectReference Include="..\..\OrchardCore\OrchardCore.Module.Targets\OrchardCore.Module.Targets.csproj" />
+    <ProjectReference Include="..\..\OrchardCore\OrchardCore.Navigation.Core\OrchardCore.Navigation.Core.csproj" />
+    <ProjectReference Include="..\..\OrchardCore\OrchardCore.Recipes.Abstractions\OrchardCore.Recipes.Abstractions.csproj" />
+    <ProjectReference Include="..\..\OrchardCore\OrchardCore.ResourceManagement\OrchardCore.ResourceManagement.csproj" />
+    <ProjectReference Include="..\..\OrchardCore\OrchardCore.Search.Abstractions\OrchardCore.Search.Abstractions.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="NEST" Version="7.12.0" />
```
Skrypt, line 43: This can be moved to the OrchardCore.Build Dependencies.props file.
This is why the compiler returns a warning about using a non managed dependency.

src/OrchardCore.Modules/OrchardCore.Search.Elastic/Services/ElasticIndexManager.cs
```diff
@@ -0,0 +1,228 @@
...
+                            }
+                            else
+                            {
+                                elasticDocument.Set(entry.Name, ((DateTime)(entry.Value)).ToUniversalTime());
+                            }
+                        }
+                        //else
+                        //{
+                        //    elasticDocument.Set(entry.Name, null);
+                        //}
```
Skrypt, line 169: Can you confirm that we can use an Exists query to reach the same goal if we remove these?

src/OrchardCore.Modules/OrchardCore.Search.Elastic/OrchardCore.Search.Elastic.csproj
Sen-Gupta: Moved!

src/OrchardCore.Modules/OrchardCore.Search.Elastic/Services/ElasticIndexManager.cs
Sen-Gupta: Let me first do a RCA and figure out the best solution!

src/OrchardCore.Modules/OrchardCore.Search.Elastic/OrchardCore.Search.Elastic.csproj
Sen-Gupta: @Skrypt do I need to add it anywhere else, I still see the warnings!

src/OrchardCore.Modules/OrchardCore.Search.Elastic/Services/ElasticIndexManager.cs
Sen-Gupta: > @Sen-Gupta
> 
> Thanks for this work, it may fix one of the remaining concerns when running in a distributed environment

You guys have done some amazing work. I'm  just helping myself!

src/OrchardCore.Modules/OrchardCore.Search.Elastic/Services/ElasticIndexManager.cs
```diff
@@ -0,0 +1,228 @@
...
+                        break;
+
+                    case DocumentIndex.Types.Text:
+                        if (entry.Value != null && !String.IsNullOrEmpty(Convert.ToString(entry.Value)))
+                        {
+                            elasticDocument.Set(entry.Name, Convert.ToString(entry.Value));
```
Skrypt, line 198: This needs to also use the DocumentIndexOptions to define if the field is Analyzed or Stored.

src/OrchardCore.Modules/OrchardCore.Search.Elastic/Services/ElasticIndexManager.cs
Skrypt: The issue is that if we don't index a value for a Field or Part then it won't find any document when we do a wildcard query.

```json
{
  "query": {
    "wildcard": {
      "Article.TitlePart": {
        "value": "*"
      }
    }
  }
}
```

So ElasticSearch documentation recommends using a default null value for these.

https://www.elastic.co/guide/en/elasticsearch/reference/current/null-value.html

src/OrchardCore.Modules/OrchardCore.Search.Elastic/Startup.cs
Skrypt: I think we should move the Search feature out of the Lucene module instead and make it using an abstraction on the Search service.

Sen-Gupta: @Skrypt @sebastienros  Perhaps evaluate, if we need to follow something like this

1) OrchardCore.Search (All abstractions related to Search
2) OrchardCore.Indexing (Abstractions for common indexing configurations)
3) OrchardCore.Search.Queries (May need to do away with OrchardCore.Search??)
4) OrchardCore.Search.Components (All Search based components/like drop downs)
5) OrchardCore.Search.Lucene
6) OrchardCore.Search.Elastic
7) OrchardCore.Search.Azure

src/OrchardCore.Modules/OrchardCore.Search.Elastic/OrchardCore.Search.Elastic.csproj
Sen-Gupta: @Skrypt, please validate and Resolve this conversation!

src/OrchardCore.Cms.Web/appsettings.json
Sen-Gupta: @sebastienros please resolve this, if it answers the question!

Sen-Gupta:
> @Sen-Gupta
> 
> Thanks for this work, it may fix one of the remaining concerns when running in a distributed environment

You guys have done some amazing work! I'm just helping myself! ;-)

Skrypt:
https://github.com/OrchardCMS/OrchardCore/pull/9117

I created this draft PR for the analyzers. There is some work remaining to do on it though.

src/OrchardCore.Modules/OrchardCore.Search.Elastic/Services/ElasticIndexManager.cs
Sen-Gupta: @Skrypt 

All text fields are analyzed by default in Elastic Search!
(https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-index-search-time.html)

Stored is really an overhead and only required if we are processing thousands of large documents.
By default Elastic will store the entire document into a field called _source and retrieves them when asked them from Index itself.
https://www.elastic.co/guide/en/elasticsearch/reference/7.12/search-fields.html

I just kept both of these for later implementations as noted in my first comment with PR. (Advanced Scenarios) ;-)

I think, we should 
1) first solid the basic functionalities, merge with Dev (More people can provide feedback)
2) Define an architecture leading to abstraction for (Indexing, Querying, Search based components) [This is mandatory so that developers can swap search providers without needed to code)
3) Build advanced features per provider on the go.

Skrypt: Ok, I looked at these documentation pages and they strongly suggest not adding a Stored custom field because it is already in the _source field. This makes sense. Now, we need to see if we shouldn't do the same with the Lucene implementation so that it be consistent then. Though, from the documentation they say that they allow still to do it so if it's an option we might want to keep it still.

Sen-Gupta: @Skrypt , We must store with with Lucene as there is no way, to build a document when retrieved from Lucene.
Unfortunately, I had to work with Lucene, Solr and Elastic and Azure Search last 10 years at some point of time!

Sen-Gupta: @Skrypt for some reason, while debugging, i found that while added fields to Lucene everything was turned to "String" in OrcharcdCore.Lucene. I'm not sure if that changed, (I remember debugging OrchardCore.Lucene around 4 months back).

I might be wrong though! ;-)

Skrypt: Where do you see this? Because If I'm looking at LuceneIndexManager we use Int32Field, DoubleField and StringField.

Sen-Gupta:
@Skrypt things changed 2 days back. The Elastic Licensing model could be deterrent.  

This is the way forward: (Elastic Fork, maintained by community and supported by AWS)
https://aws.amazon.com/blogs/opensource/introducing-opensearch/

Skrypt:
Ok but in the mean time current version of ElasticSearch should be sufficient to be able to develop this untill we get a working version of OpenSearch. I've not found any Docker image yet for OpenSearch.

src/OrchardCore.Modules/OrchardCore.Search.Elastic/Services/ElasticIndexManager.cs
Sen-Gupta: @Skrypt , I see your point! I have observed though if field has no data, it is set to null and document will not have this field at all!

Sen-Gupta:
@Skrypt Have a look at the Mappings from Elastic: (Auto Mappings)

{
  "mappings": {
    "_doc": {
      "properties": {
        "AliasPart": {
          "type": "text",
          "fields": {
            "keyword": {
              "type": "keyword",
              "ignore_above": 256
            }
          }
        },
        "Article": {
          "properties": {
            "AutoroutePart": {
              "type": "text",
              "fields": {
                "keyword": {
                  "type": "keyword",
                  "ignore_above": 256
                }
              }
            },
            "HtmlBodyPart": {
              "type": "text",
              "fields": {
                "keyword": {
                  "type": "keyword",
                  "ignore_above": 256
                }
              }
            },
            "TitlePart": {
              "type": "text",
              "fields": {
                "keyword": {
                  "type": "keyword",
                  "ignore_above": 256
                }
              }
            }
          }
        },
        "AutoroutePart": {
          "type": "text",
          "fields": {
            "keyword": {
              "type": "keyword",
              "ignore_above": 256
            }
          }
        },
        "Blog": {
          "properties": {
            "AutoroutePart": {
              "type": "text",
              "fields": {
                "keyword": {
                  "type": "keyword",
                  "ignore_above": 256
                }
              }
            },
            "HtmlBodyPart": {
              "type": "text",
              "fields": {
                "keyword": {
                  "type": "keyword",
                  "ignore_above": 256
                }
              }
            },
            "TitlePart": {
              "type": "text",
              "fields": {
                "keyword": {
                  "type": "keyword",
                  "ignore_above": 256
                }
              }
            }
          }
        },
        "Content": {
          "properties": {
            "BodyAspect": {
              "properties": {
                "Body": {
                  "type": "text",
                  "fields": {
                    "keyword": {
                      "type": "keyword",
                      "ignore_above": 256
                    }
                  }
                }
              }
            },
            "ContentItem": {
              "properties": {
                "Author": {
                  "type": "text",
                  "fields": {
                    "keyword": {
                      "type": "keyword",
                      "ignore_above": 256
                    }
                  }
                },
                "ContentItemId": {
                  "type": "text",
                  "fields": {
                    "keyword": {
                      "type": "keyword",
                      "ignore_above": 256
                    }
                  }
                },
                "ContentItemVersionId": {
                  "type": "text",
                  "fields": {
                    "keyword": {
                      "type": "keyword",
                      "ignore_above": 256
                    }
                  }
                },
                "ContentType": {
                  "type": "text",
                  "fields": {
                    "keyword": {
                      "type": "keyword",
                      "ignore_above": 256
                    }
                  }
                },
                "CreatedUtc": {
                  "type": "date"
                },
                "DisplayText": {
                  "type": "text",
                  "fields": {
                    "keyword": {
                      "type": "keyword",
                      "ignore_above": 256
                    }
                  }
                },
                "Latest": {
                  "type": "boolean"
                },
                "ModifiedUtc": {
                  "type": "date"
                },
                "Owner": {
                  "type": "text",
                  "fields": {
                    "keyword": {
                      "type": "keyword",
                      "ignore_above": 256
                    }
                  }
                },
                "Published": {
                  "type": "boolean"
                },
                "PublishedUtc": {
                  "type": "date"
                }
              }
            }
          }
        },
        "ContentItemId": {
          "type": "text",
          "fields": {
            "keyword": {
              "type": "keyword",
              "ignore_above": 256
            }
          }
        },
        "HtmlBodyPart": {
          "type": "text",
          "fields": {
            "keyword": {
              "type": "keyword",
              "ignore_above": 256
            }
          }
        },
        "Id": {
          "type": "text",
          "fields": {
            "keyword": {
              "type": "keyword",
              "ignore_above": 256
            }
          }
        },
        "Menu": {
          "properties": {
            "AliasPart": {
              "type": "text",
              "fields": {
                "keyword": {
                  "type": "keyword",
                  "ignore_above": 256
                }
              }
            },
            "TitlePart": {
              "type": "text",
              "fields": {
                "keyword": {
                  "type": "keyword",
                  "ignore_above": 256
                }
              }
            }
          }
        },
        "Taxonomy": {
          "properties": {
            "AliasPart": {
              "type": "text",
              "fields": {
                "keyword": {
                  "type": "keyword",
                  "ignore_above": 256
                }
              }
            },
            "AutoroutePart": {
              "type": "text",
              "fields": {
                "keyword": {
                  "type": "keyword",
                  "ignore_above": 256
                }
              }
            },
            "TitlePart": {
              "type": "text",
              "fields": {
                "keyword": {
                  "type": "keyword",
                  "ignore_above": 256
                }
              }
            }
          }
        },
        "TitlePart": {
          "type": "text",
          "fields": {
            "keyword": {
              "type": "keyword",
              "ignore_above": 256
            }
          }
        }
      }
    }
  }
}

Sen-Gupta:
@Skrypt @sebastienros 
The query module can support both.
image

Skrypt:
You mean at the same time right?

Sen-Gupta:
> You mean at the same time right?

Yes!

Known issue, since we have indexing setting (Field (stored and analyzed) in both modules, we see it twice if both Lucene and Elastics are enabled.

@Skrypt, I need help on following:

> Shall we keep indexing setting centralized or stand-alone for each module?
> On selection of a field as (Analyzed/stored), I need to send a call to Elastic to put the mapping.
> The null_values, can be configurable too.

deanmarcussen:
I don't understand why you want to have both Lucene and Elastic at the same time?

What's the use case?

>Known issue, since we have indexing setting (Field (stored and analyzed) in both modules, we see it twice if both Lucene and Elastics are enabled.

Exactly the problem!

For media we swap services, so it's either local file system, or blob, so be good to understand why you are trying / wanting both at once

Sen-Gupta:
@deanmarcussen Personally, i think, it should be one at a time. Could you please check if this is okay with all stakeholder?

Though there is a possibility to run both side by side. That's the beauty of provider model.

Skrypt:
@deanmarcussen I don't think it hurts having both enabled. You might want to use some custom code with the Lucene Provider while you can't really extend ElasticSearch. That's one use case. Of course the common use case will be to use one or the other but why should we prevent using both?

giannik:
I have been playing around with elastic module. Works great!
But seems to have issue with taxonomy fields.
When I rebuild an elastic index with  content types that have taxonomy fields I get the following error : 

Not sure what the elastic behaviour should be.

image

sebastienros:
I also think there is no issue with having them both enabled at the same time. It might be useful for smoothly migrating from Lucene to ES.

For the settings, we need a header to split them in the field settings.

Sen-Gupta:
> ue with taxonomy fields.



> I have been playing around with elastic module. Works great!
> But seems to have issue with taxonomy fields.
> When I rebuild an elastic index with content types that have taxonomy fields I get the following error :
> 
> Not sure what the elastic behaviour should be.
> 
> image

Fixed!

src/OrchardCore.Modules/OrchardCore.Search.Elastic/Services/ElasticIndexManager.cs
```diff
@@ -148,11 +148,17 @@ public async Task<ElasticTopDocs> SearchAsync(string indexName, string query)
             entries.Add("ContentItemId", documentIndex.ContentItemId);
             entries.Add("Id", documentIndex.ContentItemId);
 
+          
+
             foreach (var entry in documentIndex.Entries)
             {
-                if (entry.Name.Contains("Content.ContentItem.FullText") || entry.Name.Contains("Analyzed") || entry.Name.Contains("Sanitize") || entry.Name.Contains("Normalized"))
+                if (entries.ContainsKey(entry.Name)
+                    || entry.Name.Contains("Content.ContentItem.FullText")
```
Skrypt, line 156: Use defined constant from IndexingConstants.cs so that it be easier to manage.

https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore/OrchardCore.Contents.Core/Indexing/IndexingConstants.cs#L19

MikeKry:
Any news on this?

Skrypt:
Still need some tweaks here and there and time is an issue for me to add time on this as it fall low in my priorities right now. Though I tried it and it works.

Piedone:
Would you need to finish this @Skrypt or can @Sen-Gupta do it?

Skrypt:
@Piedone This generally working but I'd prefer adding what is missing while testing it. Of course @Sen-Gupta can also do the work.

Sen-Gupta:
Hello All,

@Skrypt @Piedone 
Perhaps evaluate, if we need to follow something like this

OrchardCore.Search (All abstractions related to Search
OrchardCore.Indexing (Abstractions for common indexing configurations)
OrchardCore.Search.Queries (May need to do away with OrchardCore.Search??)
OrchardCore.Search.Components (All Search based components/like drop downs)
OrchardCore.Search.Lucene
OrchardCore.Search.Elastic
OrchardCore.Search.Azure

If not then then.......

**Bare Minimum,**

If It is decided to keep it standalone (Either Lucene or Elastic), then at-least, we will have to have another component that says, Elastic Search Dropdown Field(We have one to drive drop down from Lucene search)

I don't like the idea of duplicating component instead we should have an interface for the data source for the search based drop down.


Other than this, there is one more known improvement to be made. (I can do it this weekend)

Please suggest!

Piedone:
Thanks for the replies.

I think such a division looks good @Sen-Gupta though I'd change the following:

- Add what would go to OrchardCore.Search.Components to OrchardCore.Search instead.
- I don't think we need an OrchardCore.Search.Azure project (I don't see why OrchardCore.Search.Elastic wouldn't work with Elastic in Azure too).

In any case, if possible, don't duplicate OrchardCore.Lucene.

Sen-Gupta:
> Thanks for the replies.
> 
> I think such a division looks good @Sen-Gupta though I'd change the following:
> 
> * Add what would go to OrchardCore.Search.Components to OrchardCore.Search instead.
> * I don't think we need an OrchardCore.Search.Azure project (I don't see why OrchardCore.Search.Elastic wouldn't work with Elastic in Azure too).
> 
> In any case, if possible, don't duplicate OrchardCore.Lucene.

Sen-Gupta:
@Piedone @Skrypt @sebastienros 


The MVP is ready to be merged!

Analyzed and Stored Properties are not very meaningful in context of ElasticSearch.

**Analyzed**

Analyzed is default for strings in Elastic Search.
By default all string fields are stored twice in elastic as "analyzed" and stored as "text" field type of elastic and again stored as is as in a "keyword" field type of elastic.

So we will have a field called ContentItemId(Text) analyzed and another called ContentItemId.Keyword(as is as) to match on exact values using TermQuery for fields like ContentItemId or emails (Elastic Stores text fields in 2 fields analyzed vs not analyzed, a field ContentItemId.Keyword is created automatically)

More details: 
https://www.elastic.co/blog/strings-are-dead-long-live-strings
https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-index-search-time.html

**Stored**

Stored is really an overhead and only required if we are processing thousands of large documents.
By default Elastic will store the entire document into a field called _source and retrieves them when asked them from Index itself.
https://www.elastic.co/guide/en/elasticsearch/reference/7.12/search-fields.html


**DSL Query Syntax**

It is suggested to alwasy use MatchQuery instead TermQuery for text fields in Elastic, where fully confident use (.Keyword fields for same)
More details here: 
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-term-query.html

**Pending**
- [ ] Refactor Search Based Fields to support both Lucene and Elastic
- [ ] Add Advance Elastic Elastic Configuration
- [ ] Add Advance Support for multilingual search

Sen-Gupta:
Hello All, (@Skrypt @jtkech @sebastienros)

I see there is an conflict in the the PR. (I am to do a refresh of the PR)

It has been waiting for quite some time, the remaining work as identified above is not critical for experimental and get feedback going.

Let me know actions points from my end to get it going for preview?

Regards
Sen

Piedone:
@jtkech would you check this out as well, please?

Skrypt:
I won't have time to catch up on in short time @jtkech but let me know if you are diving in.

src/OrchardCore.Modules/OrchardCore.Search.Elastic/Controllers/SearchController.cs
```diff
@@ -0,0 +1,220 @@
...
+            {
+                _logger.LogInformation("Couldn't execute Elastic search. No Elastic settings was defined.");
+                return BadRequest("Elastic Search is not configured.");
+            }
+
+            var elasticIndexSettings = await _elasticIndexSettingsService.GetSettingsAsync(searchSettings.SearchIndex);
+
+            if (elasticIndexSettings == null)
+            {
+                _logger.LogInformation($"Couldn't execute search. No Elastic index settings was defined for ({searchSettings.SearchIndex}) index.");
```
sebastienros, line 109: Wrong usage of logging. Don't use interpolated strings

```
 _logger.LogInformation("Couldn't execute search. No Elastic index settings was defined for ({SearchIndex}) index.", searchSettings.SearchIndex);
```

src/OrchardCore.Modules/OrchardCore.Search.Elastic/Services/ElasticIndexManager.cs
```diff
@@ -0,0 +1,246 @@
...
+                            }
+                        }
+                        //else
+                        //{
+                        //    elasticDocument.Set(entry.Name, null);
+                        //}
+                        break;
+
+                    case DocumentIndex.Types.Integer:
+                        if (entry.Value != null && Int32.TryParse(entry.Value.ToString(), out var value))
```
sebastienros, line 189: @Skrypt had to change it to int64

sebastienros:
Needs human validation

Skrypt:
https://opensearch.org/downloads.html

It needs to be tested against OpenSearch also.

Skrypt:
Okay, I'm removing the "don't merge" label as I think this could be merged in dev as it is functional but will require some more work to get it coded as we want it. More work for other PR's.

src/OrchardCore.Modules/OrchardCore.Search.Elastic/Services/ElasticIndexManager.cs
```diff
@@ -0,0 +1,246 @@
...
+            _elasticClient = elasticClient;
+        }
+
+        public async Task<bool> CreateIndexAsync(string indexName)
+        {
+            //Get Index name scoped by ShellName
+            if (!await Exists(indexName))
+            {
+                var response = await _elasticClient.Indices.CreateAsync(indexName);
+                return response.Acknowledged;
```
Skrypt, line 56: This could use some Exception catching here as it currently silently fails when connecting to OpenSearch for example.

Skrypt:
Ok, OpenSearch works fine.

Skrypt:
OpenSearch Docker Compose file : 

[docker-compose.txt](https://github.com/OrchardCMS/OrchardCore/files/7174675/docker-compose.txt)

ElasticSearch Docker Compose file : 

[docker-compose.txt](https://github.com/OrchardCMS/OrchardCore/files/7174677/docker-compose.txt)

src/OrchardCore/OrchardCore.Queries.Abstractions/ElasticQueryResults.cs
```diff
@@ -0,0 +1,10 @@
+using System.Collections.Generic;
+
+namespace OrchardCore.Queries
+{
+    public class ElasticQueryResults : IQueryResults
```
Skrypt, line 5: I'm not sure this is necessary as maybe we should rename the LuceneQueryResults class to SearchEngineQueryResults and reuse it.

Skrypt: Done

Sen-Gupta:
@Skrypt Do you think? Instead of SearchEngineQueryResults, SearchQueryResults looks more generic and intuitive?

Skrypt:
I finally kept it as you can see but I moved it in OrchardCore.Lucene.Abstractions where it belongs. The interface was not used like it should standing there in the same library than it's classes. Now, I use it as a contract for other modules to implement a IQueryResults.

Skrypt:
I'm now testing this with my own website which has a lot of Lucene Queries. Trying to migrate them to ElasticSearch Queries but there is some issues in the code that we need to fix. For example, the query I'm testing seems to work in the admin UI but not on the frontend.

Skrypt:
I'm testing this with my own website. I copied these modules in my own Nuget packages solution after fixing the IQueryResult issue. Maybe I should fix the dependency in this PR first.

Sen-Gupta:
> I'm testing this with my own website. I copied these modules in my own Nuget packages solution after fixing the IQueryResult issue. Maybe I should fix the dependency in this PR first. Let me try to create repro data and everything.

Thanks!

Just a note: Ideally we should be passing a dictionary of Dictionary<string, object> to Elastic. 

Elastic is smart enough to infer the types and create fields types accordingly, which could be handy in running queries later such as range, date, lte, gte and much more.

I will revisit the CreateElasticDocument method.

Skrypt:
Here for a reference on the error message : 

```
"Invalid NEST response built from a successful (200) low level call on POST: /default_search/_bulk\r\n# Invalid Bulk items:\r\n  operation[0]: index returned 400 _index: default_search _type: _doc _id: k2XsfHwBtNVfrcVKSdzi _version: 0 error: Type: illegal_argument_exception Reason: \"mapper [Content.ContentItem.DisplayText] cannot be changed from type [text] to [ObjectMapper]\"\r\n# Server indicated deprecations:\r\n- 299 Elasticsearch-7.14.1-66b55ebfa59c92c15db3f69a335d500018b3331e \"Elasticsearch built-in security features are not enabled. Without authentication, your cluster could be accessible to anyone. See https://www.elastic.co/guide/en/elasticsearch/reference/7.14/security-minimal-setup.html to enable security.\"\r\n# Audit trail of this API call:\r\n - [1] HealthyResponse: Node: http://localhost:9200/ Took: 00:00:00.0483953\r\n# Request:\r\n<Request stream not captured or already read to completion by serializer. Set DisableDirectStreaming() on ConnectionSettings to force it to be set on the response.>\r\n# Response:\r\n<Response stream not captured or already read to completion by serializer. Set DisableDirectStreaming() on ConnectionSettings to force it to be set on the response.>\r\n"
```

Sen-Gupta:
> Here for a reference on the error message :
> 
> ```
> "Invalid NEST response built from a successful (200) low level call on POST: /default_search/_bulk\r\n# Invalid Bulk items:\r\n  operation[0]: index returned 400 _index: default_search _type: _doc _id: k2XsfHwBtNVfrcVKSdzi _version: 0 error: Type: illegal_argument_exception Reason: \"mapper [Content.ContentItem.DisplayText] cannot be changed from type [text] to [ObjectMapper]\"\r\n# Server indicated deprecations:\r\n- 299 Elasticsearch-7.14.1-66b55ebfa59c92c15db3f69a335d500018b3331e \"Elasticsearch built-in security features are not enabled. Without authentication, your cluster could be accessible to anyone. See https://www.elastic.co/guide/en/elasticsearch/reference/7.14/security-minimal-setup.html to enable security.\"\r\n# Audit trail of this API call:\r\n - [1] HealthyResponse: Node: http://localhost:9200/ Took: 00:00:00.0483953\r\n# Request:\r\n<Request stream not captured or already read to completion by serializer. Set DisableDirectStreaming() on ConnectionSettings to force it to be set on the response.>\r\n# Response:\r\n<Response stream not captured or already read to completion by serializer. Set DisableDirectStreaming() on ConnectionSettings to force it to be set on the response.>\r\n"
> ```
> 
> illegal_argument_exception Reason: "mapper [Content.ContentItem.DisplayText] cannot be changed from type [text] to [ObjectMapper]

Skrypt:
I found the issue. It's a custom IndexingHandler that I use with Lucene which handles null values on a numeric field.
I will try to see if passing a null value will end up working in that case. Because, with Lucene it doesn't.

Sen-Gupta:
Apologies, I closed the ticket by mistake!

All we need is in the Elastic Index Manager: 

1) Check and ensure we do not added duplicate field in a document
2) And datatype of each field remains same (DispalyText - string and DisplayText - Object) will not be accepted by Elastic. (Elastic will infer field datatype the first time it encounter the field and expects the same type thereafter. 

The issue we see is exactly this.

In the screenshot below (Old code) you may also wants to place a Log Error to flag duplicate field in a document just before continue to flag the issues.


image

Skrypt:
Ok now, I made everything to index except Taxonomies but we knew that already. I think the issue is that I had commented line 160 locally because I don't have these IndexingConstant with a Nuget package solution. I moved them in the ElasticIndexManager and everything started to work again. Now, the issue we had with Lucene is the fact that we can't do an "exists" query to find null values on a specific field. So, we ended up using null_value by indexing a "NULL" string that we know internally to use. Here, Lucene doesn't care if you index a different type of value in a document because we don't have a Schema like in ElasticSearch. The issue is how we index with Lucene. Need to see if we could implement an "exists" query.

Though ElasticSearch seems to allow using a null_value per field, that way it would allow indexing a "NULL" value as a string on a field that has a float type for example.

https://www.elastic.co/guide/en/elasticsearch/reference/7.15/null-value.html

I'm looking to understand how this could be set globally for every field we index.

Skrypt:
As for the first field type indexing first win thing it makes sense.

Sen-Gupta:
@Skrypt passing Mapping is not required as long as we ensure the above two conditions! Let Elastic do it's magic!

Skrypt:
Fixed the settings. We have now separate sets of settings per indexing provider.

image

infofromca:
parameterizedQuery on line of ElasticQuerySource.cs has not been used

infofromca:
after Install ElasticSearch or OpenSearch with Docker compose, it is ok now.
for strong code, we still need 
if (var elasticSearchResult !=null) under line 68 of ElasticQuerySource.cs

Piedone:
Could you get back to this, @Sen-Gupta?

Skrypt:
@Piedone This is more like a POC than a complete functional Pull Request. So, it works on the backend but everything we use on the frontend (method helpers) needs to be refactored correctly.

Sen-Gupta:
@Skrypt @Piedone I believe, we should look at holistically about refactoring our current search to be provider based model based on Interfaces and rename projects as suggested at very beginning of this thread!  

It may delay a little but it goes a long way!!

Skrypt:
@Sen-Gupta This is where I'm at : https://github.com/OrchardCMS/OrchardCore/pull/10515
To support multiple indexing provider settings in Orchard Core. That's the only part the ElasticSearch module is dependent on in Orchard Core so far. Refactoring the search module as provider-based can be done in this PR in parallel without issues.

src/OrchardCore.Modules/OrchardCore.Search.Elastic/Views/ElasticContentIndexSettings.Edit.cshtml
```diff
@@ -0,0 +1,12 @@
+@model OrchardCore.Search.Elastic.Settings.ElasticContentIndexSettingsViewModel
+
+<h4>Elastic Search</h4>
```
Skrypt, line 3: Here also is the visual distinction.

Skrypt:
Moved to Orchard Core repository owned branch skrypt/elasticsearch

PR #9116, opened 2021-04-10
Implementing script and style blocks

### LINKED ISSUE #8970: {% script ... %} within a Liquid widget is not rendered at all
We have a page where there is a content (a Paragraph widget) and a script (within a Liquid widget). The Liquid widget starts with a `{% script %}` tag in order to load a js library that is only needed in this particular page (there is no need to include it in all pages of the site):

image

We have this `vendor-uaparser` declared with `DefineScript` in `ResourceManagementOptionsConfiguration`:

```
// https://cdnjs.com/libraries/UAParser.js
_manifest
    .DefineScript("vendor-uaparser")
    .SetUrl("~/myTheme/vendor/ua-parser/0.7.23/ua-parser.min.js", "~/myTheme/vendor/ua-parser/0.7.23/ua-parser.js")
    .SetCdn("https://cdnjs.cloudflare.com/ajax/libs/UAParser.js/0.7.23/ua-parser.min.js", "https://cdnjs.cloudflare.com/ajax/libs/UAParser.js/0.7.23/ua-parser.js")
    .SetCdnIntegrity("sha512-5gjoQXsBQsm4QreUg5isPzOlXKMTjsWYv6ijaT691hNfUKK2rdpL7oMFzp7Px65avCEx2HIuxgR0U0bMAJIANA==", "sha512-/8yVMB5bgQfbg4OrhpH4DvIFYw+m7q2nJPhJ1J2vILYn6mdJiA4600SmeQE/SfNwreOZEceAsoey4lN30RGxpg==")
    .SetVersion("0.7.23");
```

When this page is rendered, the `ua-parser.js` script is not included at all in the html being generated:

image

Please note the `<!-- Script Begins--><!--Script Ends-->` placeholders with nothing in between.

We have also tried using:
`{% script name:"vendor-uaparser", src:"https://cdnjs.cloudflare.com/ajax/libs/UAParser.js/0.7.23/ua-parser.min.js" %}`
instead, with the same results (the script tag is not rendered).

We are using OC 1.0.0-rc2-16014.

This worked in version 1.0.0-rc2-15878.

sebastienros:
Also refactoring internals to write directly to the writer instead of buffer into IHtmlContent.

Fixes #8970

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Resources/Liquid/StyleBlock.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Resources/Liquid/StyleBlock.cs
@@ -14,6 +14,7 @@ namespace OrchardCore.Resources.Liquid
 {
     public class StyleBlock
     {
+        private static readonly char[] Separators = new[] {',', ' '};
         public static async ValueTask<Completion> WriteToAsync(List<FilterArgument> argumentsList, IReadOnlyList<Statement> statements, TextWriter writer, TextEncoder encoder, TemplateContext context)
         {
             var services = ((LiquidTemplateContext)context).Services;
@@ -89,10 +90,8 @@ namespace OrchardCore.Resources.Liquid
                 // This allows additions to the pre registered style dependencies.
                 if (!String.IsNullOrEmpty(dependsOn))
                 {
-                    setting.SetDependencies(dependsOn.Split(new[] { ',', ' ' }, StringSplitOptions.RemoveEmptyEntries));
+                    setting.SetDependencies(dependsOn.Split(Separators, StringSplitOptions.RemoveEmptyEntries));
                 }
-
-                // TODO: implement styleblock
 
                 var content = "";
 
--- a/src/OrchardCore.Modules/OrchardCore.Resources/Liquid/StyleBlock.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Resources/Liquid/StyleBlock.cs
@@ -1,0 +1,173 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text.Encodings.Web;
+using System.Threading.Tasks;
+using Fluid;
+using Fluid.Ast;
+using Microsoft.AspNetCore.Mvc.Rendering;
+using Microsoft.Extensions.DependencyInjection;
+using OrchardCore.Liquid;
+using OrchardCore.ResourceManagement;
+
+namespace OrchardCore.Resources.Liquid
+{
+    public class StyleBlock
+    {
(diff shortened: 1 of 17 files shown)


sebastienros:
Fixes #8970

agriffard:
> Fixes #8970

Needs to be in the first comment.

deanmarcussen:
Is it possible to get this implemented as well?

https://github.com/OrchardCMS/OrchardCore/blob/4e1f974eede50f374a400317771040d1f1ba8dd5/src/OrchardCore.Modules/OrchardCore.Resources/Liquid/ScriptTag.cs#L199

That was previously working via the tag helper bridge, and is in common use as `{% script %}` rather than `{% scriptblock %}`

same for style

(or we deprecate it, and make some notes in the docs)

deanmarcussen:
Related https://github.com/OrchardCMS/OrchardCore/issues/8970

sebastienros:
I thought about being able to use the same tag for blocks, but in terms of performance it's very bad because it has to process the whole file until it finds our there is not {% endscript %} to know there it's a self-closing one. And that for all {% script %} tags in the page.

deanmarcussen:
Yes, I see the issue. Theoretically you could determine if it should be self closing, i.e. should seek, from two of the arguments, `src` and `name`. 

But complicated, so perhaps doc updates are better.

sebastienros:
The docs already refer to scriptblock and styleblock, this is not new.

deanmarcussen:
Yes, ok, I was just looking to see if the docs reference the inline rendering through the old tag helper bridge and I can’t find it. So undocumented in the first place ;)

src/OrchardCore.Modules/OrchardCore.Resources/Liquid/StyleBlock.cs
```diff
@@ -0,0 +1,173 @@
...
+
+                if (!String.IsNullOrEmpty(version))
+                {
+                    setting.UseVersion(version);
+                }
+
+                // This allows additions to the pre registered style dependencies.
+                if (!String.IsNullOrEmpty(dependsOn))
+                {
+                    setting.SetDependencies(dependsOn.Split(new[] { ',', ' ' }, StringSplitOptions.RemoveEmptyEntries));
```
jtkech, line 92: This couple of separators is used in many places, maybe time to define a static char array but in a separate PR ;)

src/OrchardCore.Modules/OrchardCore.Resources/Liquid/StyleBlock.cs
```diff
@@ -0,0 +1,173 @@
...
+                }
+
+                // TODO: implement styleblock
```
jtkech, line 95: As far as i can see, all LGTM

So just remove this comment, and the related comments in the `ScriptTag` and `StyleTag`

**Update**: After one hour of reviewing

jtkech (approved):

Issue #2345, opened 2018-09-05
Monaco editor


sebastienros:
Looks way more polished with better editor experience.

sebastienros:
Note that Monaco doesn't work correctly on mobile.

sebastienros:
We can't get rid of codemirror but we could have an extra editor based on it assuming people don't need mobile view.

agriffard:
Anyone wants to contribute to a Monaco Editor for Html field and part?

MichaelPetrinolis:
Hello, i created #8266 to use Monaco in HTML and Text fields. I added monaco in the Recources project, it works with cdn and local scripts. To load monaco last and avoid conflicts with other scripts, i added a special dependency '*' that places the scripts with it last when loading (#8265). A property LoadLast would be better, but i prefered to use the existing properties.

MichaelPetrinolis:
@hishamco thanx, is nice to have monaco in the CMS, especially in the resources, eg. we could use the diffEditor in the audit trail module to display differences and many other use cases. Don't know what to make a gif about, it is just another editor for TextField and HtmlField, with settings to define a json configuration for creating the editor.

sebastienros:
- These are custom editors that we can pick in the field/part settings?
- For other places like json editors, it would be great to be able to leverage the json schemas. At least if defined when the resource is imported, like it's done for codemirror.
- There is  a Liquid extension in VsCode, would it be possible that it works there too?
- What about the Workflow JavaScript/Json editors?

sebastienros:
And finally, how does it look like in mobile mode? I tried it recently and it was fine, but how about in OC.

MichaelPetrinolis:
@sebastienros,

- This PR provides Monaco editor as option for TextField And HTML Field
- It seems that we can configure JSON language support and provide schemas (https://github.com/Microsoft/monaco-json). I could test it with Monaco editor options in this PR. Now for JSON configuration, I used code mirror.
- Unfortunately, VSCode extensions do not work for Monaco. But with Monarch we could configure a custom language and provide IntelliSense for Liquid (https://microsoft.github.io/monaco-editor/monarch.html)
- We could use Monaco for Workflow JavaScript/Json editors and provide JS hints and JSON schema validation and IntelliSense, but personally I have not tested it.
- If we enable automaticLayout option, works nice. To look better in mobile, one could disable minimap and/or line numbers, but this is configurable.

MichaelPetrinolis:
[...]

PR #9121, opened 2021-04-10
Monaco editor Preview

agriffard:
Handle Preview for Monaco editor.

https://github.com/OrchardCMS/OrchardCore/issues/2345#issuecomment-817142735

/cc @sebastienros


sebastienros:
Looks good. Weird that in one case there is both the class and the javascript event handler.

Side note, have you tried two editors in the same page ?

agriffard:
> have you tried two editors in the same page ?

Tested with 1 HtmlPart and 2 HtmlFields.

sebastienros (approved):

Issue #8416, opened 2021-01-23
Ability to change the logo and ico on the Admin panel without using custom theme.


orchardc:
### Is your feature request related to a problem? Please describe.
Currently, it is not possible to change the Admin panel's logo and ico without a custom theme (set the basetheme to Admin).

### Describe the solution you'd like
Ability to change the logo and ico on the Admin panel without a custom theme.

### Describe alternatives you've considered
A new feature is required.

Skrypt:
This should be a tenant setting then.

deanmarcussen:
I'm not sure another setting is a good idea.

Every setting we add, forces users who have custom admin themes, to also accept that setting.

What if I have already set all the tenants logo to my logo, and don't want them to be able to change it?

And yet suddenly there is a setting that allows them too?

With every new setting we just get further away, from the principle of providing a theming engine which allows users to customize themes as they require.

And then next week someone will want a setting for admin font sizes.
and colors...

Now other options that don't require settings, that might work, for this issue.
- move the relevant bits to a shape, in the `Admin` module. 
- Then to override you can use either the admin templates, or simply place a shape in your own module, take a dependency on the Admin module, and it will override it. No admin theme required.
- We've already done this for the `UserMenu` and it has worked well.

now for css, perhaps we should consider admin layers. much like we have admin templates. would be a simple adaption to the existing layers module.

Skrypt:
I totally agree with that. The only setting that should be kept is the Darkmode one. Because it's a browser/OS feature which can be implemented or not by Admin theme. The others are things that we added to the Admin Theme which are our customizations. For example the admin menu search is a customization that could be removable from a custom admin template.

So basically, we have these settings for them to be changeable on runtime. Which a Liquid template is meant for.

Skrypt:
Also please note that you are supposed to be able to change this image by using only CSS.
See https://stackoverflow.com/questions/28832129/replace-image-src-location-using-css

orchardc:
Hi Skrypt, yes I have tried but the css approach won't apply to the **Admin pane**l wihout using a custom Theme.

Skrypt:
You should be able from the admin to use a Layout Admin Template which should override the current one in your Admin theme. This by enabling the Admin Templates theme of course.

![image](https://user-images.githubusercontent.com/3228637/105613876-586c4b80-5d93-11eb-9e0a-db259c4249db.png)

And then using a custom .css file in this Layout template.
[...]

PR #9149, opened 2021-04-12
Move admin branding to a distinct shape

TFleury:
Just to allow admin re-branding without overriding whole Layout.


src/OrchardCore.Themes/TheAdmin/Views/Layout.cshtml
```diff
@@ -32,8 +31,7 @@
     <div class="ta-wrapper">
         <div class="ta-navbar-top navbar-expand-lg bg-primary">
             <div class="d-flex w-100">
-                <a class="ta-navbar-brand" href="@Url.Content("~/" + AdminOptions.Value.AdminUrlPrefix)"><img src=@Url.Content("~/TheAdmin/logo.png") alt="@Site.SiteName" /><span>@Site.SiteName</span></a>
-
+                <shape type="Branding" />
```
deanmarcussen, line 34:
```suggestion
                @await DisplayAsync(await New.Branding())
```

deanmarcussen (commented):
@TFleury looks good. can you see this issue https://github.com/OrchardCMS/OrchardCore/issues/8416

and do the same with the ico.

I would also suggest putting the shape in the Admin module, rather than the theme. 
It makes it more overridable, i.e. can be done by another module, rather than having to have an admin theme (obviously can be done through admin templates as well)

TFleury:
> I would also suggest putting the shape in the Admin module, rather than the theme.
It makes it more overridable, i.e. can be done by another module, rather than having to have an admin theme (obviously can be done through admin templates as well)

Not sure we want to override this shape from any module, and it could conflict with front-end shape, or maybe we should rename it to 'AdminBranding'.
AdminTemplates and admin theme override seems to me good enough options to customize branding.

> do the same with the ico

It seems more complicated, it's not possible to register a resource (a link resource in this case) from a shape directly rendered in layout. We need to add the shape in a Zone to make it work.

Another way could be to have an "AdminWidgets" feature, or just to have admin layers. What do you think about this option ?

TFleury:
I found a workaround to allow registration of the icon link, but it's quite dirty :

At the top of the layout :
```razor
@{
    var adminSettings = Site.As<AdminSettings>();
    var darkMode = await DarkModeService.IsDarkModeAsync();
    var brandingHtml = await DisplayAsync(await New.Branding());
}
```
And then :
```razor
<div class="ta-navbar-top navbar-expand-lg bg-primary">
            <div class="d-flex w-100">
                @brandingHtml
                <div class="flex-grow-1">
                    @if (adminSettings.DisplayTitlesInTopbar)
                    {
                        @await RenderSectionAsync("Title", required: false)
                    }
                </div>
```

deanmarcussen:
> rename it to 'AdminBranding'.

sounds good

> At the top of the layout :

also looks good, doesn't feel dirty to me, just how these things work, i.e. call it before the tag helper to render resources.

> Another way could be to have an "AdminWidgets" feature, or just to have admin layers. 

There's a branch here where I made admin layers https://github.com/OrchardCMS/OrchardCore/tree/deanmarcussen/adminlayers. 

I'd literally forgotten about it, so thank you for reminding me, maybe deserves a demo. 

The issue that I ran into (I think) was that I didn't like was that the standard themes widget templates were not discovered, because of course it was in the admin, which started making it seem a bit weird.

 (and then I forgot I wrote it!)

agriffard:
As a reminder, I also tried to provide some admin settings for this purpose but it has been decided not to merge it because we could do ite with Custom Admin Theme and Custom Theme Settings: #5920

src/OrchardCore.Themes/TheAdmin/Views/Layout.cshtml
```diff
@@ -1,8 +1,10 @@
 @inject DarkModeService DarkModeService
-@inject IOptions<AdminOptions> AdminOptions
 @{
     var adminSettings = Site.As<AdminSettings>();
     var darkMode = await DarkModeService.IsDarkModeAsync();
+
+    // Branding is pre-rendred to allow resource injection
```
deanmarcussen, line 6:
```suggestion
    // Branding is pre-rendered to allow resource injection
```

deanmarcussen (commented):
We looked at yesterdays meeting, and it all made sense.

Any chance of some docs, saying the shape is available to override?

deanmarcussen (approved):

PR #9157, opened 2021-04-13
Search title

agriffard:
In TheBlogTheme, you cannot use `{{ Model.ContentItem.DisplayText }}` in the title.

### DIFF
--- a/src/OrchardCore.Themes/TheBlogTheme/Views/Shared/Search.liquid
+++ b/src/OrchardCore.Themes/TheBlogTheme/Views/Shared/Search.liquid
@@ -7,7 +7,7 @@
             <div class="row">
                 <div class="col-lg-8 col-md-10 mx-auto">
                     <div class="site-heading">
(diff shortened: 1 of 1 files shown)


agriffard:
image

jtkech (approved):

agriffard:
Is there an equivalent of `RenderTitleSegments()` in liquid to add the title in the head?

jtkech:
There is the `page_title` liquid tag

            RegisterParserTag("page_title", ArgumentsList, RenderTitleSegmentsTag.WriteToAsync);

agriffard:
How `{% page_title "Search" %}` and ` | t` can be used at the same time?

jtkech:
Hmm, as i remember at some point we could use a filter on an argument, so maybe

    {% page_title "Search" | t %}

If it doesn't work, try by assigning a var first `{% assign search = "Search" | t  %}`

PR #9168, opened 2021-04-14
Create a liquid widget guide

deanmarcussen:
I felt bad that we had to drop liquid widgets, so I wrote a guide on how to create one.

(with a warning)

### DIFF
--- a/src/docs/guides/create-liquid-widget/README.md
+++ b/src/docs/guides/create-liquid-widget/README.md
@@ -63,7 +63,7 @@ To customize this template
 
 This template will only render the `LiquidPart`, and will override the default widget template.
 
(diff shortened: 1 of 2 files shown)


src/docs/guides/create-liquid-widget/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+- Create a template with the following content
+
+``` liquid
+{{ Model.Content.LiquidPart | shape_render }}
+```
+
+This template will only render the `LiquidPart`, and will override the default widget template.
+
+You could also choice to render the widget sizing classes.
```
sebastienros, line 66:
```suggestion
You could also choose to render the widget sizing classes.
```

PR #9177, opened 2021-04-15
Fix drop widget to empty zone

### LINKED ISSUE #9176: Drop widget to empty zone
In the Widgets admin page, it is not possible to drag and drop a widget into a zone that does not have any widgets.

agriffard:
media

Fixes #9176

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Layers/Views/Admin/Index.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Layers/Views/Admin/Index.cshtml
@@ -45,10 +45,10 @@
                             @zone
                         </div>
                         <div class="layer-zone card-body" data-zone="@zone">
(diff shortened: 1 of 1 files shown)


agriffard:
It might need an always visible icon to indicate that you can drag and drop a widget, and a drop zone to display where you are dropping the widget.

deanmarcussen:
You can inject a selector shape into the content summaries, as we do when browsing the blog litems, and the are sortable.

https://github.com/OrchardCMS/OrchardCore/blob/851502f3bd05e036895455d0e6e00d2f946e61ac/src/OrchardCore.Modules/OrchardCore.Lists/Views/ListPart.DetailAdmin.cshtml#L94

Assuming that's what you are looking for ;)

agriffard:
Merging it as it fixes at least the corresponding issue.

PR #9182, opened 2021-04-16
Fix lucene index creation/update at startup

sebastienros:


src/OrchardCore.Modules/OrchardCore.Lucene/Services/LuceneIndexingService.cs
```diff
@@ -104,7 +104,7 @@ public async Task ProcessContentItemsAsync(string indexName = default)
                 // Create a scope for the content manager
                 var shellScope = await _shellHost.GetScopeAsync(_shellSettings);
 
-                await shellScope.UsingServiceScopeAsync(async scope =>
+                await shellScope.UsingAsync(async scope =>
```
Skrypt, line 107: Have you tested this to make sure it doesn't bite it's own tail?

src/OrchardCore.Modules/OrchardCore.Lucene/Services/LuceneIndexInitializerService.cs
```diff
@@ -11,42 +10,32 @@ namespace OrchardCore.Lucene
...
-            _luceneIndexSettingsService = luceneIndexSettingsService;
-            _luceneIndexingService = luceneIndexingService;
             _shellSettings = shellSettings;
-            _logger = logger;
         }
 
         public Task ActivatedAsync()
         {
-            if (_shellSettings.State != TenantState.Uninitialized)
+            if (_shellSettings.State == TenantState.Running)
```
Skrypt, line 21: Positivism :)

src/OrchardCore.Modules/OrchardCore.Lucene/Services/LuceneIndexInitializerService.cs
```diff
@@ -11,42 +10,32 @@ namespace OrchardCore.Lucene
...
             {
                 ShellScope.AddDeferredTask(async scope =>
                 {
-                    try
-                    {
-                        var luceneIndexSettings = await _luceneIndexSettingsService.GetSettingsAsync();
+                    var luceneIndexSettingsService = scope.ServiceProvider.GetRequiredService<LuceneIndexSettingsService>();
+                    var luceneIndexingService = scope.ServiceProvider.GetRequiredService<LuceneIndexingService>();
+                    var indexManager = scope.ServiceProvider.GetRequiredService<LuceneIndexManager>();
```
Skrypt, line 27: Makes sense using the current scope and resolve it's services. +1

src/OrchardCore.Modules/OrchardCore.Lucene/Services/LuceneIndexInitializerService.cs
```diff
@@ -1,6 +1,5 @@
-using System;
 using System.Threading.Tasks;
-using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.DependencyInjection;
```
Skrypt, line 2: Tried to avoid it but ... works too.

src/OrchardCore.Modules/OrchardCore.Lucene/Services/LuceneIndexingService.cs
Skrypt: Ok it works so I'll find out why by myself 😉

Skrypt: Kind of the same though. Here with UsingServiceScopeAsync we set _serviceScopeOnly to true which will not process in the ActiveShellInternalAsync method. But passing a bool activateShell = false will not call the ActivateShellInternalAsync method at all. So it will be faster. Good one.

Skrypt: Ah! And also it would have prevented other DeferredTasks to be triggered I guess. Ok. all good.

jtkech: @Skrypt 

Yes we needed to re-use `UsingAsync()` but with `activateShell: false` but i will check why

Both are not activating the shell, so the only difference is that `UsingServiceScopeAsync` will also not call deferred tasks and terminating event handlers, so maybe something to check here

The problem is that i was not able to repro the issue that @sebastienros had and that was only happening when the index needs to be created.

src/OrchardCore.Modules/OrchardCore.Lucene/Services/LuceneIndexInitializerService.cs
jtkech: @Skrypt 

Yes, because otherwise we resolve services not from the right scope service provider

Didn't see it when reviewing, while mounting my SSD ;)

src/OrchardCore.Modules/OrchardCore.Lucene/Services/LuceneIndexingService.cs
Skrypt: Why do we still have this UsingServiceScopeAsync method then? Need to understand why it's necessary. I've seen it's used just in few places. I'm guessing it makes sense to be used internally but maybe not as public.

src/OrchardCore.Modules/OrchardCore.Lucene/Services/LuceneIndexInitializerService.cs
jtkech: @Skrypt 

Here i suggested to @sebastienros to use `!= TenantState.Uninitialized` meaning not during a setup

But `== TenantState.Running` is okay, knowing that the other states are `Disabled` or `Invalid`

src/OrchardCore.Modules/OrchardCore.Lucene/Services/LuceneIndexingService.cs
jtkech: Because it's a "light" one that doesn't call any event handlers including activating / activated handlers, but also terminating / terminated handlers, so just to resolve and use some services that we know don't use them.

And also that doesn't manage deferred tasks (the ones that could be enlisted in this scope), e.g. if we know that we don't need them here

So e.g. when you use a minimum shell descriptor, or just to resolve the shell descriptor when we first grab it before first building the shell container

It was created before i added the `activateShell` option of the regular one

src/OrchardCore.Modules/OrchardCore.Lucene/Services/LuceneIndexInitializerService.cs
Skrypt: Yeah I thought that since these are Singletons that there would be only one instance of them though. But actually there is one of them per scope.

src/OrchardCore.Modules/OrchardCore.Lucene/Services/LuceneIndexingService.cs
Skrypt: Don't fix what ain't broken they say. All good.

jtkech: @Skrypt 

Okay, so the issue that @sebastienros had was that on the first request the layout Content zone was not rendered, after resolving the services from the right scope he was still having the problem but only when the Search was not existing and need to be built.

The problem is that i can't repro when re-using `UsingServiceScopeAsync()` so for me it seems to be a timing issue that has been fixed by using `UsingAsync()` that does a little more works. So maybe on the first rendering the Lucene index was not fully created on Lucene side and then available.

Maybe i can't repro the issue he had because of my new SSD ;)

Skrypt: I'm surprised because you can search on a page while the index is rebuilding.
The issue looks more like the index did not get created on first request or it did not reach yet the foreach loop that indexes the content items. Though, this will happen only on a page that uses a Lucene Query. Which means the request returned the proper results because the index did not start indexing content items yet. It's not an issue since this can happen any time that you do a rebuild of an entire index. If you do a page request at the same time it is more likely that your Lucene Query will just return no result at all.

jtkech: > The issue looks more like the index did not get created on first request or it did not reach yet the foreach loop that indexes the content items

The deferred task that we use here is executed at the end of the scope used to activate the shell so for sure before any rendering that will happen in a further scope

>Though, if I was using the incorrect scope it can have looked on the wrong place to recreate the indexes.

Always better to resolve services from the current scope, even it is a tenant singleton as the shell may have been rebuild e.g. after a migration recipe using a features step.

But we still had the same problem even after having changed the services resolutions, only fixed by using `UsingAsync()`.

But as you said no problem, will see ;)

Skrypt:
I just tested this with 3 tenants. One host (SaaS recipe) and 2 child tenants (Blog recipe).
The indices don't get created if we are doing a first request on the host tenant. It does get created when we do a first request on each of the 2 child tenants. Which means these are not pre-warmed.

One thing I'm seeing is that if an index takes too much time to get indexed on Azure it could potentially reach a request timeout.
Though, it should resume indexing the remaining if you enable the Lucene background task feature.

PR #9186, opened 2021-04-16
Enable auto setup by default

sebastienros:


deanmarcussen:
Docs to tweak ;)

I'm ok with it.
I had put some code in my startups to only run this in development, probably just out of paranoia, that someone might put some values towards prod... 

```
            var builder = services.AddOrchardCms();
            if (_env.IsDevelopment())
            {
                builder.AddSetupFeatures("OrchardCore.AutoSetup");
            }
```

but probably just paranoia on my part.

sebastienros:
I decided to move it to the Startup.cs file such that it's not "mandatory". Otherwise it would not be easy to remove it.

src/docs/reference/modules/AutoSetup/README.md
```diff
@@ -121,3 +121,7 @@ Do not forget to enable the Autosetup feature in Web project's Startup file:
             .AddSetupFeatures("OrchardCore.AutoSetup");
     }
 ```
+
+This feature is enabled by default in the default project included in the source code, but
+is not with the application templates to prevent any unexpected behavior when a custom projects
```
jtkech, line 126: `when a custom project` without `s`

PR #9192, opened 2021-04-16
Don't allocate for page output writing in LiquidViewTemplate

lahma:
We had a little performance tuning session with @sebastienros  and tried to figure out why so many string allocations were necessary for liquid page output. With his help I was able to narrow it down to inefficient output strategy.

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewTemplate.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewTemplate.cs
@@ -56,6 +56,11 @@ namespace OrchardCore.DisplayManagement.Liquid
                 var content = new ViewBufferTextWriterContent();
                 await context.EnterScopeAsync(page.ViewContext, (object)page.Model);
                 await template.FluidTemplate.RenderAsync(content, htmlEncoder, context);
+                
+                // Use ViewBufferTextWriter.Write(object) from ASP.NET directly since it will use a special code path
+                // for IHtmlContent. This prevent the TextWriter methods from copying the content from our buffer 
+                // if we did content.WriteTo(page.Output)
+                
                 page.Output.Write(content);
             }
             finally
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewTemplate.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewTemplate.cs
@@ -56,7 +56,7 @@ namespace OrchardCore.DisplayManagement.Liquid
                 var content = new ViewBufferTextWriterContent();
                 await context.EnterScopeAsync(page.ViewContext, (object)page.Model);
                 await template.FluidTemplate.RenderAsync(content, htmlEncoder, context);
-                content.WriteTo(page.Output, htmlEncoder);
+                page.Output.Write(content);
             }
             finally
             {


jtkech:
Yes i thought about using `page.Output.Write(content)` but was not sure if we could assume that the writer is an mvc `ViewBufferContentWriter` taking into account that we pass to it an `IHtmlContent` object.

Effectively at runtime we see that it is a `ViewBufferContentWriter` but in the code it is passed as a `TextWriter`.

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewTemplate.cs
```diff
@@ -56,7 +56,7 @@ internal static async Task RenderAsync(RazorPage<dynamic> page)
                 var content = new ViewBufferTextWriterContent();
                 await context.EnterScopeAsync(page.ViewContext, (object)page.Model);
                 await template.FluidTemplate.RenderAsync(content, htmlEncoder, context);
-                content.WriteTo(page.Output, htmlEncoder);
+                page.Output.Write(content);
```
sebastienros, line 59: When calling `content.WriteTo(page.Output)` our implementation goes over each chunk of the internal string builders and calls Write(ReadOnlySpan), which is not implemented in aspnet core https://github.com/dotnet/aspnetcore/blob/main/src/Mvc/Mvc.ViewFeatures/src/Buffers/ViewBufferTextWriter.cs

So the TextWriter implementation is used, and it rents an array and then calls Write(char[]), which in `ViewBufferTextWriter` does a string conversion.

By doing `page.Otuput.Write(content)` it's now invoking `ViewBufferTextWriter.Write(IHtmlContent)` which doesn't allocate anything. And then when the `IHtmlContent` is used, it should write directly to the response buffer (haven't verified this theory).

jtkech: > which is not implemented in aspnet core

The base implementation of `TextWriter` does this

https://github.com/dotnet/runtime/blob/2810ae5ff692108291692dec5ac36ba698af887b/src/libraries/System.Private.CoreLib/src/System/IO/TextWriter.cs#L183-L196

        public virtual void Write(ReadOnlySpan<char> buffer)
        {
            char[] array = ArrayPool<char>.Shared.Rent(buffer.Length);

            try
            {
                buffer.CopyTo(new Span<char>(array));
                Write(array, 0, buffer.Length);
            }
            finally
            {
                ArrayPool<char>.Shared.Return(array);
            }
        }

sebastienros: That's my second phrase, exactly.

jtkech: Okay cool then

jtkech:
What i meant is that maybe doing this we would call the `writer(object? value)` of the `TextWriter`

We would need to cast it to a `ViewBufferContentWriter` but we can't as it is an internal class

This is why i didn't do it, but not sure at all ;)

Textwriter

sebastienros:
yes it's the Write from the TextWriter, but the actual implementation is never a TextWriter, unless it doesn't define the overload for it. In this case we know it's ASP.NET's buffer, and we know what the Write(object) does so we want to optimize for this case. I am adding a comment.

jtkech:
Okay, just saw that their `ViewBufferTextWriter ` also overload `Write(object value)`

PR #9202, opened 2021-04-17
Add pooling to JSON.NET serialization

lahma:
JSON.NET by default doesn't use pooling as it's meant to be pluggable. Here's a proof of concept of using pooling in YesSql serialization and deserialization. Easier to do on Orchard side as there's performance test suite that @sebastienros can run against the PR - maybe something that YesSql could provide OOTB, if dependencies are OK there.

The gist is the `PoolingJsonContentSerializer` for YesSql which uses `PoolingJsonSerializer` which again can be used then elsewhere too. Moved pooling related items to Abstractions\Pooling and visible via `InternalsVisibleTo` not to open API too much. Shown also in `DefaultDynamicCacheService` which was earlier doing some allocations due to direct `JsonConvert` usage.

Also set `LineInfoHandling.Ignore` for loading data, this caused significant amount of allocations. This will probably lose the ability to get specific line info if JSON in database is invalid, but hopefully that's not a common scenario?


src/OrchardCore/OrchardCore.Abstractions/Pooling/StringBuilderPool.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        private StringBuilderPool(ObjectPool<StringBuilderPool> pool)
+        {
+            Debug.Assert(pool != null);
+            _pool = pool;
+        }
+
+        public int Length => Builder.Length;
+
+        internal static ObjectPool<StringBuilderPool> CreatePool(int size = 1000)
```
sebastienros, line 33: Maybe taken from another context, but 1000 is too big for this one. There is a single content item deserialized/serialized at a time per request. So unless we expect 1000 concurrent requests ... RPS doesn't got up after 8 right now. Si I would say 32 max. Need to check the size of the buffer too for content items. I believe the agency's home page should be bigger than 1KB.

sebastienros: Agency homepage is 10KB, so this pool would not help.

sebastienros: Updated to 32/32KB to run the benchmarks

sebastienros: This is with a single connection before/after
```
| Max GC Heap Size (MB)         | 303           |    ▁▁▁▁▃▁▂▃▂▃▄▅▅▆▇█▂▃▄▄▅▆▂▂▃▂▂▂▂▂▁▁▁▁▃▂▂▂ |
| Max Number of Gen 0 GCs / sec | 4.00          |     ▂▄ ▂█  ▂       ▂     ▄  ▂▂▂▂▂▂▂▂▂ ▂   |
| Max Number of Gen 1 GCs / sec | 1.00          |      █  █  █             █  █             |
| Max Number of Gen 2 GCs / sec | 1.00          |         █                █                |
| Max Time in GC (%)            | 3.00          |      ▃▃ ███                               |
| Max Gen 0 Size (B)            | 9,735,824     |                                 █         |
| Max Gen 1 Size (B)            | 17,358,440    |     ▃  ▄▃▃▃█████████████████              |
| Max Gen 2 Size (B)            | 35,886,016    |      ▂▂▂▃▃▃▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄█████████████ |
| Max LOH Size (B)              | 614,392       |     ▃▃▃██████████████████████████████████ |
| Max Allocation Rate (B/sec)   | 299,604,616   |    ▁▂▄ ▃█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▂▂▂▂▂▂▂▂▂▁   |

| Max GC Heap Size (MB)         | 247           |   ▁▁▁▁▁▂▃▂▃▄▅▆▇█▂▃▄▂▃▁▂▃▁▂▃▂▂▂▁▄▄▃▃▃▃▃▃▃ |
| Max Number of Gen 0 GCs / sec | 3.00          |    ▅▃ █▅ ▃      ▃  ▃ ▃  ▃  ▃▃▃▃ ▃▃▃▃▃    |
| Max Number of Gen 1 GCs / sec | 1.00          |    █  █  █         █                     |
| Max Number of Gen 2 GCs / sec | 1.00          |       █            █                     |
| Max Time in GC (%)            | 9.00          |    ▄  █▅▅                                |
| Max Gen 0 Size (B)            | 288           |    █████████████████████████████████████ |
| Max Gen 1 Size (B)            | 13,694,688    |        ▆▆▇▇▇▇▇▇▇███               ▁▁▁▁▁▁ |
| Max Gen 2 Size (B)            | 35,550,184    |    ▂▂▂▃▃▃▅▅▅▅▅▅▅▅▅▅█████████████████████ |
| Max LOH Size (B)              | 614,392       |    ▃▃▃██████████████████████████████████ |
| Max Allocation Rate (B/sec)   | 229,831,464   |   ▁▅▃ █▇▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▃▃▃▃▃▃▃▃▃▁   |
```

lahma: I was much lower originally https://github.com/OrchardCMS/OrchardCore/commit/53d3e22ff4a0d404ba7170e4f3815761e8594ebd

lahma: I think I'll do a commit for switching to zstring and we'll see how it behaves

lahma:
I pushed a version that switches all `StringBuilderPool` usage to` ZString`/new` ZStringWriter`, *except for ViewBufferTextWriterContent*. The `ZStringWriter` would also serve the other PR where pooled version would make sense.

Would be interesting to get numbers from this and if it affects much as not touching `ViewBufferTextWriterContent` . You updated the now removed pool which is not used by `ViewBufferTextWriterContent`, it has the one I didn't want to move yet (`OrchardCore.DisplayManagement.StringBuilderPool`).

lahma:
And there are still places where `JsonConvert.SerializeObject`/`JsonConvert.DeserializeObject` are used instead of the pooling version. I guess this mostly is now for seeing how YesSql behaves with hopefully improved reading by utilizing ArrayPool<char>.

src/OrchardCore/OrchardCore.Abstractions/Pooling/ZStringWriter.cs
```diff
@@ -0,0 +1,287 @@
...
+            {
+                throw new ObjectDisposedException(nameof(_sb));
+            }
+
+            _sb.Append(buffer.AsSpan(index, count));
+        }
+
+        public override void Write(ReadOnlySpan<char> buffer)
+        {
+            if (GetType() != typeof(StringWriter))
```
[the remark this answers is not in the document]
lahma: good catch, removed all such checks as class is sealed

TFleury:
Sorry but I see internal sealed classes and a dependency on ZString in an abstraction library.

IMO, everything you put in `OrchardCore.Abstractions.Pooling` should go to `OrchardCore.Infrastructure.Pooling`.

lahma:
If that's the correct place I can move in next iteration where I'm going to replace some StringBuilderPool usage.

TFleury:
I'm not 100% sure what's the best place, but quite sure it should not go into `OrchardCore.Abstractions` library. Doing so it adds ZString dependency to quite all OC libraries.

I also think we should avoid `InternalsVisibleTo` usage, maybe using an abstraction layer.

lahma:
InternalsVisible was mostly about not exposing outside of OC. ZString can be quite fundamental, it's efficient string handling after all. Newtonsoft.Json is also there,  that's the reason I put it there.  I'm very open to ideas how to improve and I can fix via PR.

sebastienros:
It's always better to explicitly depend on what you use, and not assume it's coming from a different dependency.
Also Thierry is currently trying to fix those unnecessary dependency graphs. I'd say that if a module requires JSON.NET or ZString it should reference it. Even if it's coming from a different dependency already, "by chance".

lahma:
The `ZStringWriter` was just accepted to `ZString` library. I can do the PR to reduce dependencies when there's a new release.

PR #9203, opened 2021-04-17
Use pooled StringBuilder instances with StringWriter

lahma:
`StringWriter`s were always using their own `StringBuilder`, now using the pool. `ScriptTagHelper` still uses new'ed builders as it cannot see `StringBuilderPool`, maybe future improvement. @sebastienros can you please run benchmarks for this.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Handlers/FullTextAspectContentHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Handlers/FullTextAspectContentHandler.cs
@@ -64,7 +64,7 @@ namespace OrchardCore.Contents.Handlers
                     if (bodyAspect != null && bodyAspect.Body != null)
                     {
                         using var stringBuilderPool = StringBuilderPool.GetInstance();
-                        await using var sw = new StringWriter(stringBuilderPool.Builder);
+                        using var sw = new StringWriter(stringBuilderPool.Builder);
                         // Don't encode the body
                         bodyAspect.Body.WriteTo(sw, NullHtmlEncoder.Default);
                         fullTextAspect.Segments.Add(sw.ToString());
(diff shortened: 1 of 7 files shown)


src/OrchardCore.Modules/OrchardCore.Contents/Handlers/FullTextAspectContentHandler.cs
```diff
@@ -62,12 +63,11 @@ public override Task GetContentItemAspectAsync(ContentItemAspectContext context)
...
                     if (bodyAspect != null && bodyAspect.Body != null)
                     {
-                        using (var sw = new StringWriter())
-                        {
-                            // Don't encode the body
-                            bodyAspect.Body.WriteTo(sw, NullHtmlEncoder.Default);
-                            fullTextAspect.Segments.Add(sw.ToString());
-                        }
+                        using var stringBuilderPool = StringBuilderPool.GetInstance();
+                        await using var sw = new StringWriter(stringBuilderPool.Builder);
```
jtkech, line 67: Does `StringWriter` implement `DisposeAsync()` ?

sebastienros: It does through TextWriter, though it seems to be useless code inside around Dispose(), I think it's better to call Dispose().

lahma: wouldn't compile otherwise, I will revert back to simple `using`

jtkech: > wouldn't compile otherwise

Oops ;)

lahma:
Here using `ZString` and `ZStringWriter` would make sense as gives nicer API, trying it out on #9202

sebastienros:
ZString can be used later

lahma:
I created a PR for ZString to add ZStringWriter there (can get rid of custom one), when that lands as a version I can revisit usage and remove most of the StringBuilder usage I guess.

PR #9204, opened 2021-04-17
Reduce string allocations in MediaTokenService

lahma:
`MediaTokenService` used `String.Concat` which internally allocates string builder as there always seems to be more than one parameter to calculate for. Also a separate key was derived from using `TokenCacheKeyPrefix + commandValues`. Now using pooled string builder to build key with prefix and hashing the part of it that does not contain the prefix.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenService.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenService.cs
@@ -5,6 +5,7 @@ using System.Text;
 using Microsoft.AspNetCore.WebUtilities;
 using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Options;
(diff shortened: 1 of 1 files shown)


sebastienros:
String.Concat uses a stack allocated char[256] by default and a pooled array if it overflows it (see ValueStringBuilder). I don't think it's necessary here to use the StringBuilderPool.

lahma:
Might make sense to have `ValueStringBuilder` look-alike available and stack alloc, sizes usually should be lower than 256 in general case. Would `ZString` be acceptable dependency?

sebastienros:
I think it's fine, though I wish they had a shared source variation of the package so we don't add extra dlls for single classes.

sebastienros:
Do you want to reword this one once ZString is part of the dependencies?

sebastienros:
I am merging it and other PRs using ZString can be created

PR #9206, opened 2021-04-17
Optimize some Fluid access patterns

lahma:
* instead of looping 8 converters, use 3 that handle required cases
* directly convert `ViewLocalizer` to `ObjectValue` instead of looping converters
* create fast case for `ZoneTag` evaluating literals

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Liquid/OrchardCore.Liquid.csproj
+++ b/src/OrchardCore.Modules/OrchardCore.Liquid/OrchardCore.Liquid.csproj
@@ -8,7 +8,6 @@
 
     The liquid module enables content items to have liquid syntax.</Description>
     <PackageTags>$(PackageTags) OrchardCoreCMS</PackageTags>
-    <LangVersion>latest</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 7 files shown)


agriffard:
blog-test.js failed on Linux:
```
 1) Blog Tests
       Displays the home page of the blog recipe:
     AssertionError: Timed out retrying after 4000ms: Expected to find element: `.subheading`, but never found it.
```

Relaunching build to see if it is still happening.

lahma:
This is pretty scary, why only on linux...

agriffard:
> This is pretty scary, why only on linux...

It is just that Functional Tests are only run on Linux.

deanmarcussen:
I had to revert langversion latest recently because when I did it project wide we ran into razor compilation errors on net core 3.

probably ok here, but it’s just static that you used it for I think? So not really necessary and maybe better not to bump just one project, or it gets confusing as to which ones can n can’t have it

lahma:
I like to use the `static` modifier for lambdas for documentation purposes, it shows that it's "being used right", otherwise there could be context capturing which can cause closure instances to be instantiated to hold the state. But I'm fine changing it too, just checked that the project wasn't using Razor targets which seemed to be the reason to disallow usage of latest features.

lahma:
Reverted back to c# 8 and found the probable cause, `LiteralValue.ToString()` is different from `LiteralValue.ToStringValue()` ... Wasn't the first time for me..

PR #9208, opened 2021-04-17
Ensure ViewBufferTextWriterContent is properly disposed and releasing pooled instances

lahma:
`ViewBufferTextWriterContent` is `IDisposable` and relies on somewhat luck to get disposed (`WriteTo`) needs to be called. The recent PR took the `IHtmlContent` to buffer but there's nobody to dispose such instances which have not had `WriteTo` called. Removing `IHtmlContent` as the life-cycle doesn't correspond to usage.

Changing the usage patterns to have `using` and materializing the produced string when needed.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Liquid/Services/LiquidTemplateManager.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Liquid/Services/LiquidTemplateManager.cs
@@ -21,8 +21,8 @@ namespace OrchardCore.Liquid.Services
         private readonly IServiceProvider _serviceProvider;
 
         public LiquidTemplateManager(
-            IMemoryCache memoryCache, 
-            LiquidViewParser liquidViewParser, 
+            IMemoryCache memoryCache,
+            LiquidViewParser liquidViewParser,
             IOptions<TemplateOptions> templateOptions,
             IServiceProvider serviceProvider
             )
@@ -72,11 +72,9 @@ namespace OrchardCore.Liquid.Services
                 }
             }
 
-            var htmlContentWriter = new ViewBufferTextWriterContent();
-
+            using var htmlContentWriter = new ViewBufferTextWriterContent();
             await result.RenderAsync(htmlContentWriter, encoder, context, model);
-
-            return htmlContentWriter;
+            return new HtmlString(htmlContentWriter.ToString());
         }
 
         public Task RenderAsync(string source, TextWriter writer, TextEncoder encoder, object model = null, IEnumerable<KeyValuePair<string, FluidValue>> properties = null)
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewTemplate.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewTemplate.cs
@@ -53,15 +53,14 @@ namespace OrchardCore.DisplayManagement.Liquid
 
             try
             {
-                var content = new ViewBufferTextWriterContent();
+                using var content = new ViewBufferTextWriterContent();
                 await context.EnterScopeAsync(page.ViewContext, (object)page.Model);
                 await template.FluidTemplate.RenderAsync(content, htmlEncoder, context);
-                
-                // Use ViewBufferTextWriter.Write(object) from ASP.NET directly since it will use a special code path
-                // for IHtmlContent. This prevent the TextWriter methods from copying the content from our buffer 
-                // if we did content.WriteTo(page.Output)
-                
-                page.Output.Write(content);
+
+                // Use ViewBufferTextWriter.WriteAsync(string value) from ASP.NET directly since so it will write content from
+                // pooled string builder and allow us to return it to pool by having the using directive above
+
+                await page.Output.WriteAsync(content.ToString());
             }
             finally
             {
@@ -265,7 +264,7 @@ namespace OrchardCore.DisplayManagement.Liquid
                     context.TimeZone = timeZoneInfo;
                 }
 
-                // Configure Fluid with the local date and time 
+                // Configure Fluid with the local date and time
                 var now = await localClock.LocalNowAsync;
 
                 context.Now = () => now;
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Tags/FluidTagHelper.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Tags/FluidTagHelper.cs
@@ -88,12 +88,14 @@ namespace OrchardCore.DisplayManagement.Liquid.Tags
 
             if (content != null)
             {
+                var output = content.ToString();
                 tagHelperOutput = new TagHelperOutput(
                     identifier,
-                    outputAttributes, (_, e) => Task.FromResult(new DefaultTagHelperContent().AppendHtml(content))
+                    outputAttributes, (_, e) => Task.FromResult(new DefaultTagHelperContent().AppendHtml(output))
                 );
 
-                tagHelperOutput.Content.AppendHtml(content);
+                tagHelperOutput.Content.AppendHtml(output);
+                content.Dispose();
             }
             else
             {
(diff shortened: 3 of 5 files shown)


src/OrchardCore/OrchardCore.DisplayManagement.Liquid/LiquidViewTemplate.cs
```diff
@@ -53,15 +53,14 @@ internal static async Task RenderAsync(RazorPage<dynamic> page)
...
-                // Use ViewBufferTextWriter.Write(object) from ASP.NET directly since it will use a special code path
-                // for IHtmlContent. This prevent the TextWriter methods from copying the content from our buffer 
-                // if we did content.WriteTo(page.Output)
-                
-                page.Output.Write(content);
+
+                // Use ViewBufferTextWriter.WriteAsync(string value) from ASP.NET directly since so it will write content from
+                // pooled string builder and allow us to return it to pool by having the using directive above
+
+                await page.Output.WriteAsync(content.ToString());
```
sebastienros, line 63: We intentionally pass the `IHtmlContent` instance such that it calls `WriteTo`, which will release the resources.

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Tags/FluidTagHelper.cs
```diff
@@ -88,12 +88,14 @@ public static async ValueTask<Completion> WriteToAsync(string identifier, List<F
 
             if (content != null)
             {
+                var output = content.ToString();
                 tagHelperOutput = new TagHelperOutput(
                     identifier,
-                    outputAttributes, (_, e) => Task.FromResult(new DefaultTagHelperContent().AppendHtml(content))
+                    outputAttributes, (_, e) => Task.FromResult(new DefaultTagHelperContent().AppendHtml(output))
```
sebastienros, line 94: I think the issue here is that the same buffer is used twice, I wonder why? @jtkech

sebastienros: Maybe I made a mistake when I refactored it too

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Tags/ZoneTag.cs
```diff
@@ -50,7 +50,7 @@ public static async ValueTask<Completion> WriteToAsync(List<FilterArgument> argu
 
                 if (zone is Shape shape)
                 {
-                    await shape.AddAsync(content, position);
+                    await shape.AddAsync(content.ToString(), position);
```
sebastienros, line 53: passe as an `IHtmlContent` on purpose and it will be released when `WriteTo` is called

src/OrchardCore/OrchardCore.DisplayManagement.Liquid/ViewBufferTextWriterContent.cs
```diff
@@ -315,6 +343,27 @@ public override Task WriteLineAsync(StringBuilder value, CancellationToken cance
...
+            using var pool = StringBuilderPool.GetInstance();
+            using var sw = new StringWriter(pool.Builder);
+            WriteTo(sw, NullHtmlEncoder.Default);
+            return sw.ToString();
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void CheckDisposed()
+        {
+            if (_disposed)
```
sebastienros, line 357: More than `_disposed` it should actually check for `_released` or just throw an exception in the `WriteTo` method if it's called twice, or already released.

lahma:
I think this PR doesn't address anything meaningful so I'm closing it. Probably should check if there are string builder taken from pool that are not returned.

src/OrchardCore.Modules/OrchardCore.Liquid/Services/LiquidTemplateManager.cs
```diff
@@ -72,11 +72,9 @@ public async Task<IHtmlContent> RenderHtmlContentAsync(string source, TextEncode
                 }
             }
 
-            var htmlContentWriter = new ViewBufferTextWriterContent();
-
+            using var htmlContentWriter = new ViewBufferTextWriterContent();
             await result.RenderAsync(htmlContentWriter, encoder, context, model);
-
-            return htmlContentWriter;
+            return new HtmlString(htmlContentWriter.ToString());
```
jtkech, line 77: Sorry, as usual i didn't check it in details as i am working on other things ;)

But even if it is just a feeling, i prefer to point it out when i'm not sure

Here looks like we allocate a string in place of returning our `ViewBufferTextWriterContent` and let it write to the final output later, this through its `WriteTo()` as it is an `IHtmlContent`, and by the mvc `ViewBufferTextWriter`.

sebastienros: @jtkech the PR is closed now, but check this instead https://github.com/OrchardCMS/OrchardCore/issues/9209

jtkech: Okay

lahma:
I think there's still conceptually a problem here. What if the `ViewBufferTextWriterContent` is added to output queue but request is aborted for some reason, say an application error or client closes connection. There's no way to clean up acquired resources as there is no try-catch semantics. Eventually string builders have been exhausted and pool won't return anymore.

lahma:
I wonder if using `context.Response.RegisterForDispose(content);` would make sense, that would ensure the disposal I guess?

jtkech:
@lahma 

Just for infos we always execute in what we call a `ShellScope` and we can have multiple ones in a given request scope, also we may execute in a context where there is no RequestContext, e.g. in a Background Task

We already have our `RegisterBeforeDispose()` that we use to commit the session asynchronously before a given shell scope is disposed, so maybe a good candidate. Hmm, but yes if something fails it may still not be executed.

                    ShellScope.RegisterBeforeDispose(scope =>
                    {
                        return scope.ServiceProvider
                            .GetRequiredService<IDocumentStore>()
                            .CommitAsync();
                    });

Yes, would be good to defer the disposing because our buffer can't be rendered twice and i had to fix an issue where it was rendered twice.

sebastienros:
It the process fails, the builder is not returned to the pool. That is totally ok. The GC will collect it, the pool will create a new instance. This is that simple, and I don't any issues with it. This is an "exceptional" case.

lahma:
The pool holds references to instances and thus GC is not allowed for them? If pool has only "lost instances" the inner array becomes full of references. Then only thing happening is switching the first item which is always effective as `Allocate` takes it and it becomes null (Dispose returning it as first item), but the pool will effectively become single item pool, always allocating if the single item is not available (and doing linear search for array which is always full).

Or am I missing something...

```chsarp
        internal T Allocate()
        {
            T inst = _firstItem;
            if (inst == null || inst != Interlocked.CompareExchange(ref _firstItem, null, inst))
            {
                inst = AllocateSlow();
            }
            return inst;
        }

        private T AllocateSlow()
        {
            var items = _items;

            for (int i = 0; i < items.Length; i++)
            {
                T inst = items[i].Value;
                if (inst != null)
                {
                    if (inst == Interlocked.CompareExchange(ref items[i].Value, null, inst))
                    {
                        return inst;
                    }
                }
            }
            return CreateInstance();
        }

        internal void Free(T obj)
        {
            if (_firstItem == null)
            {
                _firstItem = obj;
            }
            else
            {
                FreeSlow(obj);
            }
        }

        private void FreeSlow(T obj)
        {
            var items = _items;
            for (int i = 0; i < items.Length; i++)
            {
                if (items[i].Value == null)
                {
                    items[i].Value = obj;
                    break;
                }
            }
        }
    }
```

sebastienros:
As I read it from a phone, it's removing the reference when it's provided.

sebastienros:
Proof is the code in the Release method that noops when the capacity has changed.

lahma:
Ah you're totally correct., maybe I should sleep more - returning more items to pool than taking from it would have been my case but not what's happening. The `StringBuilderPool` is the correct solution for `ViewBufferTextWriterContent ` but using `ArrayPool<T>` is not (like in` ZString`), `ArrayPool<T>` needs to be returned (try-finally). Sorry for the confusion.

PR #9239, opened 2021-04-20
Update DataMigration.cs comments

deanmarcussen:


agriffard:
It seems you cannot use `<` or `>` because it is in a `<code>` markup.

deanmarcussen:
Meh, I'll have to read the docs ;)

agriffard:
Isn't it supposed to be : `&lt;int&gt;` ?

deanmarcussen:
Yes, ty @agriffard had no idea you could use that in comments, but of course it is just xml. Cheers

PR #9246, opened 2021-04-21
Remove OrchardCore.Data.Abstractions dependency on YesSql

### LINKED ISSUE #9245: Remove OrchardCore.Application.Targets dependency on YesSql
Using the modular application scenario, you must be able to choose your data access layer, or even not use anyone.

Actually there is a dependency on YesSql and by cascade on Dapper.

Dependency graph :
`OrchardCore.Application.Targets` => `OrchardCore` => `OrchardCore.Infrastructure.Abstractions` => `OrchardCore.Data.Abstractions` => `YesSql.Core`

I think `OrchardCore.Data.Abstractions`, as an abstraction layer, should not rely on a specific implementation.

Related discussion [here](https://github.com/OrchardCMS/OrchardCore/discussions/7716#discussioncomment-140165)

TFleury:
Closes #9245

Makes dependency graph cleaner.

The main goal of the PR is to remove YesSql dependency in modular application scenario.


src/OrchardCore.Modules/OrchardCore.Lists/OrchardCore.Lists.csproj
```diff
@@ -25,6 +25,7 @@
     <ProjectReference Include="..\..\OrchardCore\OrchardCore.Contents.Core\OrchardCore.Contents.Core.csproj" />
     <ProjectReference Include="..\..\OrchardCore\OrchardCore.Contents.TagHelpers\OrchardCore.Contents.TagHelpers.csproj" />
     <ProjectReference Include="..\..\OrchardCore\OrchardCore.ContentTypes.Abstractions\OrchardCore.ContentTypes.Abstractions.csproj" />
+    <ProjectReference Include="..\..\OrchardCore\OrchardCore.Data.YesSql\OrchardCore.Data.YesSql.csproj" />
```
hishamco, line 28: > I think OrchardCore.Data.Abstractions, as an abstraction layer, should not rely on a specific implementation.

Also OrchardCore modules should not rely on a specific implementation

TFleury: They should not, but many of them do.

The PR removes the dependency of `OrchardCore.Data.Abstractions` on `YesSql.Core`.

Because of that, actually many modules in OC has an indirect dependency on `YesSql`, even if not needed.

DataMigration which is a core feature directly relies on YesSql because of `ISchemaBuilder` property which is a `YesSql` interface.

This PR does not fix all dependency problems over `YesSql`, it's just a first step to isolate and identify which module relies on `YesSql`. There is much more work to create all necessary abstraction layers to properly decouple modules from infrastructure.
But I think it's acceptable to have `YesSql` dependency for everything that is related to content management, but not for modules that could be used in a different context.

sebastienros:
Haven't looked at the changes yet. I actually tried it 2 weeks ago, and AFAIR there were some things from YesSql.Core that some modules were using, so I stopped in the middle as the changes were becoming too big for my intent. Glad you are doing it.

sebastienros:
Oh, and the fact that it made me add a Dapper dependency, I think that was the thing.

sebastienros:
To abstract from yessql in modules that are not part of the CMS, we might need to add some "Store" abstractions, like Identity has. And implement a store for YesSql in a child module. But that might mean a lot more modules :/

src/OrchardCore.Modules/OrchardCore.Lists/OrchardCore.Lists.csproj
hishamco: IMHO some of the YesSQL primitive was to be moved into `OrchardCore.Data.Abstractions`, so almost the modules will rely on it not in YesSQL

> The PR removes the dependency of OrchardCore.Data.Abstractions on YesSql.Core

This is a very good step

hishamco:
Index/Reduce pattern is commonly used in many modules which is a YesSQL thing ;(

hishamco:
I think it's too late to abstract everything from YesSQL, may be it's good thing to minimize the dependency as much as we can then we can think about this in `2.0.0` because there will a lot of breaking changes

TFleury:
I think it's just a first step, then we will be able to decouple modules one by one.

sebastienros:
Can you solve the conflict?

TFleury:
@sebastienros I think this PR is ready.
I will make other PRs to address other dependency issues when possible.

PR #9247, opened 2021-04-21
Introduce MSBuild tasks programability

### LINKED ISSUE #4991: Create programmable MSBuild Tasks
@jtkech I'm thinking to introduce programmable MSBuild tasks that will be useful for various reasons, such as:

- Copy Translations files

- Cleanup files

- or anything you may think of

I knew we can do that in `.csproj` files, be here we can have a benefit of test ability, readability, productivity, flexibility .. etc

hishamco:
Fixes #4991, this PR is just to show how MSBuild tasks programability will simplify a lot of the stuff that we made in .csproj files


hishamco:
This just the beginning if we agreed on the principle I simplify all the MSBuild tasks which is almost created by @jtkech, I'm sure the  old task sometimes hard to understand and I'm sure it's hard to write and debug

hishamco:
I'm not sure why the path in `AssemblyFile` is not go back into the `OrchardCore.Build`?

jptissot:

hishamco:
I don't think I may revise this PR, it will offers many things for us in terms of creating tasks programatically

hishamco:
Closed in favor of #12364

PR #9251, opened 2021-04-21
Remove some async await

jtkech:


test/OrchardCore.Tests/ResourceManagement/ResourceManagerTests.cs
```diff
@@ -798,13 +798,13 @@ public async Task RenderLocalStyle()
...
         #region Helpers
-        private async Task<IDocument> ParseHtmlAsync(IHtmlContent content)
+        private Task<IDocument> ParseHtmlAsync(IHtmlContent content)
         {
             using (var writer = new StringWriter())
             {
                 content.WriteTo(writer, HtmlEncoder.Default);
 
-                return await browsingContext.OpenAsync(res => res.Content(writer.ToString()).Address(basePath));
+                return browsingContext.OpenAsync(res => res.Content(writer.ToString()).Address(basePath));
```
lahma, line 807: this escapes the using and writer can be disposed before await is done

lahma: ah but the writer is not needed after this, so is fine 👍🏻

lahma: or depends when the callback is called, so can be a bomb,  confusing 🤯

sebastienros: Dont' remove it. There are more risks than benefits.

jtkech: @lahma I think you're right

Just didn't see the using, i'll revert this one, thanks

sebastienros (approved):
The recommendation from the aspnet team is to leave async/await, or that it doesn't really matter much for perf, but on the opposite the stack trace doesn't show the whole stack when it's removed. So next time don't bother changing it.

lahma:
Is this runtime version dependent? Full framework has penalty at least and added stack probably slows things on hot paths?

PR #9259, opened 2021-04-22
Added Total support for Lucene API queries

hyzx86:
sample:
`/api/queries/LQueryName?parameters={from:0,size:20}&withTotal=true`

Result:
```
{
  items:[...],
  count:1231
}
```

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Queries/Controllers/ApiController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Queries/Controllers/ApiController.cs
@@ -27,7 +27,9 @@ namespace OrchardCore.Queries.Controllers
         [Route("{name}")]
         public async Task<IActionResult> Query(
             string name,
-            string parameters)
+            string parameters,
+            bool? withTotal
+            )
         {
             var query = await _queryManager.GetQueryAsync(name);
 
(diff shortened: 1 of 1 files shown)


hyzx86:
image

If you don't pass the `withTotal` parameter, the array will be returned directly

Skrypt:
Good one, need to see if we can do the same in Liquid helper methods and else. The optional param should not break anything, I like it.

src/OrchardCore.Modules/OrchardCore.Queries/Controllers/ApiController.cs
```diff
@@ -47,8 +49,24 @@ IQueryManager queryManager
                 JsonConvert.DeserializeObject<Dictionary<string, object>>(parameters)
                 : new Dictionary<string, object>();
 
-            var result = await _queryManager.ExecuteQueryAsync(query, queryParameters);
-            return new ObjectResult(result.Items);
+            try
+            {
+                var result = await _queryManager.ExecuteQueryAsync(query, queryParameters);
+                if (withTotal ?? false && result is LuceneQueryResults)
```
sebastienros, line 55: I believe we should just return the `result` as it in all cases, or have an interface that is common to all results from queries to return the value from this controller.

The fact we see `LuceneQueryResults` means that there is an unnecessary dependency on the Lucene module.

Issue #9266, opened 2021-04-23
Font awesome script in TheAgency


agriffard:
https://github.com/OrchardCMS/OrchardCore/blob/c141b02d43e1b96e9c04516d06c3153675f996e0/src/OrchardCore.Themes/TheAgencyTheme/Views/Layout.liquid#L13

The Liquid filter `script` is not rendered in the Head by default as it was used to be so the font-awesome icons don't appear anymore in TheAgency theme.

![image](https://user-images.githubusercontent.com/703248/115894009-17023f80-a459-11eb-983f-d3cd5410882a.png)

Is it now required to specify the `at` attribute?
`{% script name:"font-awesome", version:"5", at:"Foot" %}`

agriffard:
https://docs.orchardcore.net/en/dev/docs/reference/modules/Resources/#specify-location

> If the location is not specified, or specified as Inline, the script will be inserted wherever it is placed (inline).

`at:"Inline"` does not work.

PR #9267, opened 2021-04-23
Script Liquid

agriffard:
Related to #9266


jtkech (approved):

Skrypt:
I'm not sure why you need to push it at the footer because in this page example they set it in the header : 

https://fontawesome.com/how-to-use/javascript-api/setup/configuration

agriffard:
I took example of TheAdmin but I can change.

deanmarcussen:
Is this the same issue as https://github.com/OrchardCMS/OrchardCore/issues/9271

If you can confirm @agriffard , I will look into it

agriffard:
> Is this the same issue as #9271
> 
> If you can confirm @agriffard , I will look into it

There is no `at`, so yes, I guess.

Skrypt (approved):

PR #9268, opened 2021-04-23
Update AngleSharp and HtmlSanitizer packages

agriffard:
https://github.com/AngleSharp/AngleSharp/releases/tag/v0.15.0
https://github.com/mganss/HtmlSanitizer/releases/tag/v5.0.404

### DIFF
--- a/src/OrchardCore.Build/Dependencies.props
+++ b/src/OrchardCore.Build/Dependencies.props
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageManagement Include="AngleSharp" Version="0.14.0" />
+    <PackageManagement Include="AngleSharp" Version="0.15.0" />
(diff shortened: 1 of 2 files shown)


agriffard:
Should we wait for a stable `6.x` version before merging it?
https://www.nuget.org/packages/HtmlSanitizer/6.0.409-beta

lahma:
I'm not sure how long that will take to be released, upside for waiting is if we get both libraries to version that have optimizations it would be easier to check perf before and after PR where area was HTML sanitization. I don't mind, do what you feel is the best option 🙂

agriffard:
Ok, so @lahma is not just improving OC, he's also improving the libraries used by OC. 

Next step, it will be the .NET Framework itself.

lahma:
OC started to be quite lean allocation-wise so I had to go hunting in the dependencies 😉

PR #9277, opened 2021-04-26
Email should handle file and stream attachments

hishamco:
While working on other library related to Emails, I realize there's a missing feature in sending attachments


deanmarcussen:
This introduces a large security issue where the email service can now access and send any file from the file system.

Please refer to the original attachment pr for the notes I made there explaining that this was a bad idea.
https://github.com/OrchardCMS/OrchardCore/pull/7380

hishamco:
> This introduces a large security issue where the email service can now access and send any file from the file system.

This is simliar to upload file if we use it in OC, but as Email APIs this should be supported like what I had ever seen in `System.Net.Mail.SmtpClient` as well as 'MimiKit.SmtpClient`

I think we could exclude some files extension like the work done in Media

jptissot:

PR #9293, opened 2021-04-27
Update code generation templates for zsh on MacOS

deanmarcussen:
The newer zsh shell is picky about * in a cmd and needs to be wrapped in quotes.

Think a few users have complained about this on gitter.

can someone try the new command on windows and check if it works with the quotes (it does with bash the old shell). If it does I'll just update the cmd to be used for both. @agriffard if you have time?

`dotnet new -i "OrchardCore.ProjectTemplates::1.0.0-rc2-*"`

and

`dotnet new -i "OrchardCore.ProjectTemplates::1.0.0-rc2-*" --nuget-source https://nuget.cloudsmith.io/orchardcore/preview/v3/index.json `

### DIFF
--- a/src/docs/getting-started/templates/README.md
+++ b/src/docs/getting-started/templates/README.md
@@ -12,10 +12,22 @@ Once the .NET Core SDK has been installed, type the following command to install
 dotnet new -i OrchardCore.ProjectTemplates::1.0.0-rc2-*
 ```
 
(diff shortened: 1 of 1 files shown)


agriffard:
It seems to work with the quotes on Windows.

deanmarcussen:
Thanks @agriffard updated to just put the quotes around all the time ;)

PR #9307, opened 2021-04-28
Start Bootstrap Coming Soon 5.1.0

agriffard:
https://www.npmjs.com/package/startbootstrap-coming-soon/v/5.1.0
https://github.com/StartBootstrap/startbootstrap-coming-soon/releases/tag/v5.1.0


Skrypt:
I'm just looking at the wwwroot folder and we are kind of hosting the source code. Maybe we should think about moving these to the "Assets" folder and copy the "dist" folder into the wwwroot folder only. Just an idea.

src/OrchardCore.Themes/TheComingSoonTheme/Assets/css/bootstrap-oc.css
```diff
@@ -0,0 +1,307 @@
+/*
```
Skrypt, line 1: This Assets/css/ folder is not necessary. Seems like it generated files in there from a previous gulp build.

src/OrchardCore.Themes/TheComingSoonTheme/Assets/src/scss/sections/_social.scss
```diff
@@ -0,0 +1,67 @@
...
+                    width: 4rem;
+
+                    &:hover {
+                        background-color: $primary;
+                    }
+                }
+            }
+        }
+    }
+}
```
Skrypt, line 67: Lots of EOF missing

Skrypt:
We need to also remove the new /Assets folder from the build. They should not be Embedded Resources on the assembly.

agriffard:
@Skrypt 

In the past, we made those choices:

- Copy the whole folder from the StartBootstrap repository.
- Add bootstrap-oc.scss and modules to generate a separate bootstrap-oc.css file.
- Build css and js and the .min versions using Assets.json
- Add specific Resources to the thelme in its ResourceManifestOptionsConfiguration.

You suggested in https://github.com/OrchardCMS/OrchardCore/pull/9307#issuecomment-828563857 to move the assets in an `Assets` folder and now you are asking to remove it 'from the build'.
 
I don't get what you expect me to do.

src/OrchardCore.Themes/TheComingSoonTheme/Assets.json
```diff
@@ -1,4 +1,18 @@
 [
+  {
+    "copy": true,
+    "inputs": [
+      "Assets/dist/assets/img/bg-mobile-fallback.jpg"
```
Skrypt, line 5:
```json 
 {
    "copy": true,
    "inputs": [
      "Assets/dist/assets/*"
    ],
    "output": "wwwroot/assets/@"
  },
```

jtkech:
@agriffard @Skrypt 

You don't need this in the theme project file

    <EmbeddedResource Remove="\Assets\*;" />

All files under the `Assets` folder are automatically excluded from embedded resources

https://github.com/OrchardCMS/OrchardCore/blob/395fc4295024d40ac9bad3645c2f72b531032355/src/OrchardCore/OrchardCore.Module.Targets/OrchardCore.Module.Targets.props#L9

You only need to sepecify those that we don't want to embedd and that are under the `wwwroot` folder

Skrypt:
@jtkech thank you I forgot that one. But I don't think this will be applied on Themes but only for Modules. Maybe we need to add these Excludes in Themes too then.

Skrypt (approved):
All these EOF remaining are coming from the original theme repository.
LGTM.

PR #9318, opened 2021-04-29
Decouple modules from infrastructure

TFleury:
Only "target" packages should depend on infrastructure.

This PR :
- Removes unnecessary dependency on Infrastructure.
- Replaces Infrastructure dependencies by Infrastructure.Abstractions dependencies when it's required.
- Moves some abstractions from Infrastructure to Infrastructure.Abstractions.
- Moves some infrastructure DI registrations from module startup to the "AddOrchardCms" helper.
- Fixes an issue on ApiAuthenticationHandler not returning appropriate status code.


TFleury:
@sebastienros another decoupling PR.

sebastienros (approved):

PR #9322, opened 2021-04-30
Code analysis

agriffard:
Fixes various alerts detected in the code:
https://lgtm.com/projects/g/OrchardCMS/OrchardCore/?mode=list&lang=csharp


src/OrchardCore.Modules/OrchardCore.Contents/AdminNodes/ContentTypesAdminNodeNavigationBuilder.cs
```diff
@@ -90,8 +90,6 @@ private IEnumerable<ContentTypeDefinition> GetContentTypesToShow(ContentTypesAdm
 
             if (!node.ShowAll)
             {
-                node.ContentTypes = node.ContentTypes;
```
hishamco, line 93: Do we really have such thing ;)

src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/AccessController.cs
```diff
@@ -568,9 +568,6 @@ private async Task<IActionResult> ExchangePasswordGrantType(OpenIddictRequest re
 
         private async Task<IActionResult> ExchangeAuthorizationCodeOrRefreshTokenGrantType(OpenIddictRequest request)
         {
-            var application = await _applicationManager.FindByClientIdAsync(request.ClientId) ??
-                throw new InvalidOperationException("The application details cannot be found.");
```
deanmarcussen, line 572: Not sure abou this change @agriffard 
Why was it being suggested?

agriffard: Because `application` is not used later int he code.

deanmarcussen: Remove the variable then, not the check that it exists. (I'm guessing here, but we wouldn't want to lose the behaviour of checking that the client id exists, I think)
i.e.
```
await _applicationManager.FindByClientIdAsync(request.ClientId) ??
                throw new InvalidOperationException("The application details cannot be found.");
```

agriffard: re-added but it requires to set a variable.

kevinchalet: Feel free to remove it: OpenIddict itself ensures the application exists and this second call will basically return a cached value anyway. If we don't use it, there's no point doing that (we don't allow disabling OpenIddict's L2-like cache anyway).

deanmarcussen (approved):

PR #9328, opened 2021-04-30
ZString 2.4.2

agriffard:
Use `ZStringWriter` available in the new version of ZString.

Related to #9202

`StringBuilderPool` is still used in ViewBufferTextWriterContent and CacheTag.

/cc @lahma


lahma:
I was about to prepare the same PR with adding the missing library doc, you beat me to it 👍🏻

I might do a follow-up to move dependency  from root Abstractions library.

sebastienros:
@lahma are the changes good ?

lahma (approved):
Looking good.

PR #9329, opened 2021-04-30
Negative Role Condition Evaluator

arnaudroystatcan:
The way a role condition with a negative operator is evatuated right now doesn't take multiple roles into account. 
If I have 2 users: 
-- | roleA | roleB
 ------------- | ------------- | -------------
UserA | X | 
UserB | X | X

The evaluator uses .Any() when evaluating the value against every role of the user. If I pick the operator "Equals" with the value "roleB", it will only be true for UserB which has roleB. If I were now to just change the operator to "NotEquals" logically it should be the opposite result of "Equals" but since the value get compare against every role and roleA != roleB it will return true as well.
### .Any() (Current)
Operator | Value | UserA | UserB 
 ------------- | ------------- | ------------- | ------------- 
Equals  | roleB | false | true 
Not Equals | roleB   | true | true 
Contains | roleB   | false | true 
Not Contains | roleB   | true | true 

I suggest using .All() with the negative operators which would give opposite results as intendend
Operator | Value | UserA | UserB 
 ------------- | ------------- | ------------- | ------------- 
Equals  | roleB | false | true 
Not Equals | roleB   | true | false 
Contains | roleB   | false | true
Not Contains | roleB   | true | false


src/OrchardCore.Modules/OrchardCore.Rules/Services/RoleConditionEvaluator.cs
```diff
@@ -31,6 +31,17 @@ public override ValueTask<bool> EvaluateAsync(RoleCondition condition)
             // IsInRole() & HasClaim() are case sensitive.
             var operatorComparer = _operatorResolver.GetOperatorComparer(condition.Operation);
 
+            // Claim all if the operator is negative
+            if(condition.Operation == new StringNotEqualsOperator() ||
```
deanmarcussen, line 35: I agree with the choice to use `.All` 

What I would suggest is to add an interface to the negative operators, `INegateOperator` use something like
`if (condition.Operation is INegateOperator)...`

dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=9329) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=9329) before we can accept your contribution.<br/><br/>:x: arnaudroystatcan [sign now](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=9329)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OrchardCMS/OrchardCore?pullRequest=9329) it.</sub>

deanmarcussen (approved):

agriffard:
Please format the document syntax.

jptissot:
> Please format the document syntax in `RoleConditionEvaluator.cs`

@agriffard can you elaborate on what to do ?

agriffard:
> > Please format the document syntax in `RoleConditionEvaluator.cs`
> 
> @agriffard can you elaborate on what to do ?

Format document: Ctrl + k, d
https://docs.microsoft.com/en-us/visualstudio/ide/productivity-shortcuts?view=vs-2019

jptissot:
Does this work in vscode ? I tried to format document and no changes were applied.

src/OrchardCore.Modules/OrchardCore.Rules/Services/RoleConditionEvaluator.cs
```diff
@@ -31,6 +31,15 @@ public override ValueTask<bool> EvaluateAsync(RoleCondition condition)
             // IsInRole() & HasClaim() are case sensitive.
             var operatorComparer = _operatorResolver.GetOperatorComparer(condition.Operation);
 
+            // Claim all if the operator is negative
+            if(condition.Operation is INegateOperator)
```
deanmarcussen, line 35:
```suggestion
            if (condition.Operation is INegateOperator)
```

deanmarcussen (commented):
Different keystrokes for vscode @jptissot 

And it doesn't always format as well. 
I am guessing this was the bit @agriffard wanted tweaked?

jptissot:
Ah, I formatted the document again and it worked the second time. I wonder if this was an omnisharp lagging issue.

PR #9331, opened 2021-05-01
Refine ZString dependencies

lahma:
* move JSON pooling to OrchardCode.Infrastructure, remove ZString from OrchardCore.Abstractions
* link JSON pooling to YesSql so that it won't get extra dependency
* use ZString when suitable and readily available instead of StringBuilder
* delete ZStringWriter, part of ZString now
* OrchardCode.Abstractions no longer needs InternalsVisibleTo
* PoolingJsonSerializer is now public, at some point should also create custom JSON serialization/deserialization abstraction to hide Newtonsoft.JSON would allow switching between NJSON and STJ

/cc @TFleury @agriffard


src/OrchardCore/OrchardCore.Data.YesSql/OrchardCore.Data.YesSql.csproj
```diff
@@ -24,6 +25,16 @@
...
+    <PackageReference Include="ZString" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Include="..\OrchardCore.Infrastructure\Pooling\JsonArrayPool.cs">
+      <Link>JsonArrayPool.cs</Link>
+    </Compile>
+    <Compile Include="..\OrchardCore.Infrastructure\Pooling\PoolingJsonSerializer.cs">
+      <Link>PoolingJsonSerializer.cs</Link>
+    </Compile>
```
TFleury, line 37: I think we should avoid this

TFleury (changes requested):
ZString dependency at module level is OK.

When I said it should go to Infrastructure, it implies to add an abstraction layer, like an interface, because module should not depend on infrastructure.
It must be possible to replace default infrastructure with custom implementations.

I made a PR to your repo to suggest adding separate `PoolingJsonSerializer` implementation in each module. So it solves dependency and accessibility concerns, and keeps project easy to maintain.

deanmarcussen:
What is the thinking behind module should not depend on infrastructure? (I don’t disagree I just wondering about the logic behind that after seeing you pr yesterday)

Do we just need a infrastructure.core project for some of the more shared code?

TFleury:
> What is the thinking behind module should not depend on infrastructure? (I don’t disagree I just wondering about the logic behind that after seeing you pr yesterday)

It's dependency inversion principle ([wiki](https://en.wikipedia.org/wiki/Dependency_inversion_principle))

1. High-level modules should not depend on low-level modules. Both should depend on abstractions (e.g., interfaces).
2. Abstractions should not depend on details. Details (concrete implementations) should depend on abstractions.

Infrastructure package has dependencies on implementation libraries (Ideally it should be splitted into multiple packages). So if you add a dependency on infrastructure to a module, you will add lots of undesired dependencies to your module (Ex: MimeKit, HtmlSanitizer).

> Do we just need a infrastructure.core project for some of the more shared code?

If it's just shared code, it's not infrastructure. So maybe it could go into something like "OrchardCore.Serialization.Core" or "OrchardCore.Serialization.Shared".
It's infrastructure if it's an implementation of something abstract.

I first tried to add an `IJsonSerializer` interface to Infrastructure.Abstractions, with `PoolingJsonSerializer` as an implementation, but it would have been too tightly coupled to Newtonsoft.Json, and an abstraction may not depend on Newtonsoft.Json which is an implementation.

deanmarcussen:
Cool, I agree, my take is that the infrastructure project has become a dumping ground for shared code, `MimeKit` and `HtmlSanitizer` are good examples.

Also agree they should just become seperate projects, particularly as they are often just light wrappers around an external dependency.

However we are starting to get a point where there are too many projects, and it's becoming more difficult to work on the solution, just due to the build times of all the projects.

I wonder if it is not time to split out some of these minor dependencies / shared code, that has been dumped there, into a second solution / second github project. Possibly even as @sebastienros has suggested with as a shared source package https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/dependencies#nuget-shared-source-packages

PR #9334, opened 2021-05-01
Code generation template: Module with Part

agriffard:
Update the driver to use the same syntax as the other ones in the source code.

Add registration of the viewModel in Startup or else, it won't be rendered using a Liquid template.

```
services.Configure<TemplateOptions>(o =>
            {
                o.MemberAccessStrategy.Register<MyTestPartViewModel>();
            });
```

We should add this note somewhere in the documentation and check that we didn't forget to register some view models in other modules.

### DIFF
--- a/src/Templates/OrchardCore.ProjectTemplates/content/OrchardCore.Templates.Cms.Module/Drivers/MyTestPartDisplayDriver.cs
+++ b/src/Templates/OrchardCore.ProjectTemplates/content/OrchardCore.Templates.Cms.Module/Drivers/MyTestPartDisplayDriver.cs
@@ -29,7 +29,7 @@ namespace OrchardCore.Templates.Cms.Module.Drivers
 
         public override IDisplayResult Edit(MyTestPart part, BuildPartEditorContext context)
         {
-            return Initialize<TestPartViewModel>(GetEditorShapeType(context), model =>
+            return Initialize<MyTestPartViewModel>(GetEditorShapeType(context), model =>
             {
                 model.Show = part.Show;
                 model.ContentItem = part.ContentItem;
--- a/src/Templates/OrchardCore.ProjectTemplates/content/OrchardCore.Templates.Cms.Module/Drivers/MyTestPartDisplayDriver.cs
+++ b/src/Templates/OrchardCore.ProjectTemplates/content/OrchardCore.Templates.Cms.Module/Drivers/MyTestPartDisplayDriver.cs
@@ -1,6 +1,6 @@
-﻿using System.Linq;
 using System.Threading.Tasks;
 using OrchardCore.ContentManagement.Display.ContentDisplay;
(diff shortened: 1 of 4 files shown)


src/Templates/OrchardCore.ProjectTemplates/content/OrchardCore.Templates.Cms.Module/Drivers/MyTestPartDisplayDriver.cs
```diff
@@ -19,42 +19,34 @@ public MyTestPartDisplayDriver(IContentDefinitionManager contentDefinitionManage
...
-        public override IDisplayResult Display(MyTestPart MyTestPart)
+        public override IDisplayResult Display(MyTestPart part, BuildPartDisplayContext context)
         {
-            return Combine(
-                Initialize<MyTestPartViewModel>("MyTestPart", m => BuildViewModel(m, MyTestPart))
-                    .Location("Detail", "Content:20"),
-                Initialize<MyTestPartViewModel>("MyTestPart_Summary", m => BuildViewModel(m, MyTestPart))
-                    .Location("Summary", "Meta:5")
-            );
+            return Initialize<TestPartViewModel>(GetDisplayShapeType(context), m => BuildViewModel(m, part, context))
```
deanmarcussen, line 24:
```suggestion
            return Initialize<MyTestPartViewModel>(GetDisplayShapeType(context), m => BuildViewModel(m, part, context))
```

src/Templates/OrchardCore.ProjectTemplates/content/OrchardCore.Templates.Cms.Module/Startup.cs
```diff
@@ -12,6 +13,7 @@
 using OrchardCore.Templates.Cms.Module.Handlers;
 using OrchardCore.Templates.Cms.Module.Models;
 using OrchardCore.Templates.Cms.Module.Settings;
+using OrchardCore.Templates.Cms.Module.ViewModels;
```
deanmarcussen, line 16: Think you might need a `using OrchardCore.Liquid` to get `TemplateOptions`

agriffard: TemplateOptions is in Fluid.

deanmarcussen (approved):

src/Templates/OrchardCore.ProjectTemplates/content/OrchardCore.Templates.Cms.Module/Drivers/MyTestPartDisplayDriver.cs
deanmarcussen: Sorry I read this and I still don't understand why the view model type has changed? Are there two viewmodels? one for edit, one for display, or something?

PR #9335, opened 2021-05-02
fix double click issue on workflow activity and content items

### LINKED ISSUE #4227: Workflow Activity Double click issue
I accidentally double clicked the add button when adding an activity to the workflow and it added the activity twice with the same ID, causing the UI to throw an exception, even after reloading the site. I had to go into the database document and remove the duplicate entry for the UI to unblock. Maybe we should disable the button on click.

tropcicstefan:
Fix 
#4227 
and partially #8900


deanmarcussen:
I like it. 
I think it would need to be on the Edit Views as well (and later on most forms)

jptissot:
We could also put the button in the [disabled state ](https://getbootstrap.com/docs/4.0/components/buttons/#disabled-state) by adding the `disabled` attribute. 
Another addition could be to add a [spinner ](https://getbootstrap.com/docs/4.3/components/spinners/#buttons) to the button while it's in the disabled state.

tropcicstefan:
> We could also put the button in the [disabled state ](https://getbootstrap.com/docs/4.0/components/buttons/#disabled-state) by adding the `disabled` attribute.

controller won't get button value if button is disabled [link](https://stackoverflow.com/questions/7357256/disabled-form-inputs-do-not-appear-in-the-request/7357314#7357314)
> Another addition could be to add a [spinner ](https://getbootstrap.com/docs/4.3/components/spinners/#buttons) to the button while it's in the disabled state.

I am not sure how to catch which button actually submitted when multiple:
-maybe by document.activeElement
-maybe like [this](https://stackoverflow.com/questions/2066162/how-can-i-get-the-button-that-caused-the-submit-from-the-form-submit-event/61110260#61110260) eventually



> I like it.
> I think it would need to be on the Edit Views as well (and later on most forms)

Sure. will add to Create views at least.

sebastienros:
Disable is fine, no need for a spinner, more issues than anything.

sebastienros:
Needs to be added to Edit.cshtml too.
And probably "automatically" on the links that have the custom attribute that make them form posts.

deanmarcussen:
Replaced by https://github.com/OrchardCMS/OrchardCore/pull/9716/files

Thank you @tropcicstefan it got done on all the other views as well.

PR #9364, opened 2021-05-06
Fix console.log recursivity

deanmarcussen:
Prevents infinite recursion when jsonifying shapes that reference their parents (pagers / menus etc)

and uses the `Fluid` json filter to prevent serializing members that are inaccessible.


jptissot:
Hehehe I actually ran into this one yesterday. Thanks Dean

PR #9368, opened 2021-05-06
Add ability to export templates as files

jptissot:
Adds a checkbox for the AllTemplates and AllAdminTemplates deployment steps that output the templates as files instead of inline in the Recipe.json file.


deanmarcussen:
How would the recipe import step be able to import these from files?

I suspect it couldn't, so my suggestion would be to make another deployment step for this feature, which has a different name, so won't conflict with the import step (i.e. won't be supported because we couldn't import these on a live site)

jptissot:
I tested it and it worked (import). The recipe executor actually transforms the `[file:text('filename')]` first, then the current import steps works as expected.

deanmarcussen (approved):
Great! Didn't see that line

PR #813, opened 2017-06-09
#812 Standardized claims to OpenIdConnectConstants

anoordende:
PR to resolve issue #812 OpenId comparing URI-style claims with OpenIdConnectConstants.

PR #9370, opened 2021-05-06
Check implied permissions

### LINKED ISSUE #9362: OpenId don't handles inherited (ImpliedBy) permissions
Testing Application access with OpenId module with Client Credentials flow, it appears that ImpliedBy permission are not honored.

### To Reproduce

1. Create a tenant with at least one content
2. Enable "OpenID Authorization Server" & "OpenID Token Validation"
3. Create a role for your app and check "Access content via the api" & "Publish or unpublish content for others"
(note the "Edit content for others" is implied)
4. Enable "Allow Client Credentials Flow" in OpenID Connect
5. Add an Application in OpenID Connect with "Allow Client Credentials Flow" and check the role you created at step 3
6. Retrieve a token and a content item using the Contents API controller
7. Try to update the content using the Contents API controller

You will receive a 403Forbidden

### Expected behavior
It should be allowed

TFleury:
Fixes #9362

`PermissionHandler` should check ImpliedBy permissions.
`RolesPermissionsHandler` don't have to check all user roles permissions, just "Anonymous" and "Authenticated" if applicable.
`RolesPermissionsHandler` is scoped, so I assume we can cache "Anonymous" and "Authenticated" permissions and not query them for each permission evaluation.

### DIFF
--- a/src/OrchardCore/OrchardCore.Infrastructure.Abstractions/Security/PermissionExtensions.cs
+++ b/src/OrchardCore/OrchardCore.Infrastructure.Abstractions/Security/PermissionExtensions.cs
@@ -1,0 +1,57 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Claims;
+using OrchardCore.Security.Permissions;
+
+namespace OrchardCore.Security
+{
+    public static class PermissionExtensions
+    {
+        /// <summary>
+        /// Evaluates if the specified <see cref="Permission"/> is granted by provided claims.
+        /// </summary>
+        /// <param name="permission">The <see cref="Permission"/> to test</param>
+        /// <param name="claims">Provided claims.</param>
+        /// <returns>True if the permission is granted, otherwise false.</returns>
+        public static bool IsGranted(this Permission permission, IEnumerable<Claim> claims)
+        {
+            if (claims == null || !claims.Any())
+            {
+                return false;
+            }
+
+            var grantingNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+            GetGrantingNamesInternal(permission, grantingNames);
+
+            // SiteOwner permission grants them all
+            grantingNames.Add(StandardPermissions.SiteOwner.Name);
+
+            return claims.Any(claim => String.Equals(claim.Type, Permission.ClaimType, StringComparison.OrdinalIgnoreCase)
+                && grantingNames.Contains(claim.Value));
+        }
+
+        private static void GetGrantingNamesInternal(this Permission permission, HashSet<string> stack)
+        {
+            // The given name is tested
+            stack.Add(permission.Name);
+
+            // Iterate implied permissions to grant, it present
+            if (permission.ImpliedBy != null && permission.ImpliedBy.Any())
+            {
+                foreach (var impliedBy in permission.ImpliedBy)
+                {
+                    // Avoid potential recursion
+                    if (stack.Contains(impliedBy.Name))
+                    {
+                        continue;
+                    }
+
+                    // Otherwise accumulate the implied permission names recursively
+                    GetGrantingNamesInternal(impliedBy, stack);
+                }
+            }
+        }
+    }
+}
--- a/test/OrchardCore.Tests/Security/PermissionHandlerTests.cs
+++ b/test/OrchardCore.Tests/Security/PermissionHandlerTests.cs
@@ -73,7 +73,7 @@ namespace OrchardCore.Tests.Security
         }
 
         [Fact]
-        public async Task IsCaseIsensitive()
+        public async Task IsCaseInsensitive()
         {
             // Arrange
             var required = new Permission("required");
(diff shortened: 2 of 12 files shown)


deanmarcussen:
I was just reviewing this, and you added the unit tests 👍 

I think it's good, I assume the issue you were actually fixing, maybe was because the roles were not included in the openid request, so they didn't get added to the claims principal / ticket?

always like to know what the issue being fixed is, especially around permissions and changes

TFleury:
OpenId adds roles claims, but it uses a custom claim type : "role", whereas `RolesPermissionsHandler` was handling the standard role claim type : "http://schemas.microsoft.com/ws/2008/06/identity/claims/role", so it was not matching.

I wondered if we have to change the claim type generated by OpenId to match the standard role claim type (I didn't find how 😅), but the issue was wider : every permission handler should check implied permissions, not only `RolesPermissionsHandler`.

Assume that at a point we add User permissions, we need to check implied ones.

Another point is that `RolesPermissionsHandler` was checking granted roles permission at each permission evaluation while role permissions are added to user claims on login, so it's not required.

TFleury:
I also realized that we have "Anonymous" and "Authenticated" strings in many places, shouldn't we use string constants for that ?

jptissot:
> OpenId adds roles claims, but it uses a custom claim type : "role", whereas `RolesPermissionsHandler` was handling the standard role claim type : "http://schemas.microsoft.com/ws/2008/06/identity/claims/role", so it was not matching.
> 
> I wondered if we have to change the claim type generated by OpenId to match the standard role claim type (I didn't find how 😅), but the issue was wider : every permission handler should check implied permissions, not only `RolesPermissionsHandler`.
> 
> Assume that at a point we add User permissions, we need to check implied ones.
> 
> Another point is that `RolesPermissionsHandler` was checking granted roles permission at each permission evaluation while role permissions are added to user claims on login, so it's not required.

/cc @kevinchalet

kevinchalet:
The WS-Fed vs OIDC claim types discrepancy was discussed a few years ago: https://github.com/OrchardCMS/OrchardCore/pull/813#issuecomment-307427150.
AFAIK, no decision has been made so far.

Regarding roles, you could use https://docs.microsoft.com/fr-fr/dotnet/api/system.security.claims.claimsidentity.roleclaimtype?view=net-5.0 to get the type of the claims used to convey the roles (identities created by OpenIddict use `role` as you figured out).

src/OrchardCore/OrchardCore.Infrastructure.Abstractions/Security/PermissionExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+{
+    public static class PermissionExtensions
+    {
+        /// <summary>
+        /// Evaluates if the specified <see cref="Permission"/> is granted by provided claims.
+        /// </summary>
+        /// <param name="permission">The <see cref="Permission"/> to test</param>
+        /// <param name="claims">Provided claims.</param>
+        /// <returns>True if the permission is granted, otherwise false.</returns>
+        public static bool IsGranted(this Permission permission, IEnumerable<Claim> claims)
```
deanmarcussen, line 17: We might want to introduce this as a service - @sebastienros would confirm, but last we spoke about it, the implied by's were showing up on stack traces a lot, and with a service (even if not down now on this pr), we would have the potential to add some caching to the implied by `HashSet`, rather than having to change this api again.

deanmarcussen:
@TFleury can you check the roles controller
https://github.com/OrchardCMS/OrchardCore/blob/395fc4295024d40ac9bad3645c2f72b531032355/src/OrchardCore.Modules/OrchardCore.Roles/Controllers/AdminController.cs#L273

I'm not sure the fake role claim here will be enough anymore to generate effective permissions. Might need to grab the role claims, to populate, or use the claims factory to generate the fake claims

src/OrchardCore/OrchardCore.Infrastructure.Abstractions/Security/PermissionExtensions.cs
TFleury: An alternative could be to add "Anonymous" permissions using a middleware, and "Authenticated" permissions using an `IUserClaimsProvider`, so all authorized permissions will be available as claims.
Then we can get rid of `RolesPermissionsHandler`.

In that case, the `IsGranted` logic will go into `PermissionHandler` - which is already a service - and we could add any required cache logic.

sebastienros:
Ok to make it a service so we can add some state later on a fix the perf impact it will probably have. Like some caching.

test/OrchardCore.Tests/Security/PermissionHandlerTests.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            // Act
+            await permissionHandler.HandleAsync(context);
+
+            // Assert
+            Assert.True(context.HasSucceeded);
+        }
+
+        [Fact]
+        public async Task IsCaseIsensitive()
```
deanmarcussen, line 76:
```suggestion
        public async Task IsCaseInsensitive()
```

deanmarcussen (approved):

PR #9380, opened 2021-05-07
Fix bug RoleConditionEvaluator

arnaudroystatcan:
The `claim.Type == roleClaimType` was causing the `.All()` to always be false. The evaluator now only takes into account the role claims.


jptissot (approved):
Tested locally, now works properly

PR #9381, opened 2021-05-07
Update cypress-commands

jptissot:
Also fix a timing issue in the `uploadRecipeJson()` function


jptissot:
Fyi I also pushed the changes to npm https://www.npmjs.com/package/cypress-orchardcore

PR #9384, opened 2021-05-08
Throws FileStoreException instead of exposing media path while uploading

### LINKED ISSUE #9379: Minor security issue: Internal File Path Disclosure
Hello,

A customer of ours has had a pen test undertaken of a system we have built for them, one of the issues that has been raised is the Media module can potentially expose internal file paths, letting a would be attacker find out what OS the server is using and gaining them knowledge about the file structure being used by the web application.

image

I've tracked it down to this catch statement:

https://github.com/OrchardCMS/OrchardCore/blob/9d4fb5372e6ac0deb9d749dc746a022c3dbe46a4/src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs#L184-L195

It's down as a low priority issue on the pen test but I thought I should log it here so it can be eventually resolved. Part of me is tempted to say have a blanket default error message and if someone is really interested they can check the server logs for the full details.

hishamco:
Fixes #9379

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs
@@ -131,7 +131,7 @@ namespace OrchardCore.Media.Controllers
 
         [HttpPost]
         [MediaSizeLimit]
-        public async Task<ActionResult<object>> Upload(
+        public async Task<ActionResult> Upload(
             string path,
             string contentType,
             ICollection<IFormFile> files)
@@ -146,8 +146,6 @@ namespace OrchardCore.Media.Controllers
                 path = "";
             }
 
-            var result = new List<object>();
-
             // Loop through each file in the request
             foreach (var file in files)
             {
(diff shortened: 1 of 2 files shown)


deanmarcussen (commented):
I'm afraid this is not what I suggested. 

The suggestion was that the `FileSystemStore` should be try catching exceptions, and wrapping the file system exception in a `FileStoreException`

For example as is done here https://github.com/OrchardCMS/OrchardCore/blob/97eaad4e7d7ca49ee7f2831163b60e681259ff22/src/OrchardCore/OrchardCore.FileStorage.FileSystem/FileSystemStore.cs#L90

To clarify:
- Review the `FileSystemStore` and `BlobFileStore` ensuring that exceptions there are caught and replayed as `FileSystemStore` exceptions, with any path information as part of the exception `Message` 
- `AdminController` still returns a result to the UI, where the result exposes no system related information
- no changes to the `AdminController` or `FileStoreException`

hishamco:
Aha, you mean something like TryCreate or TryDelete where the catches should go into the `FileSystemStore`

> no changes to the AdminController or FileStoreException

I think this will lead us to think about All the possible exception may occurs during uploading file

deanmarcussen:
Can you make the suggested changes @hishamco or comment if you are not planning to?

Note: it's not a change of the methods to a `TryCreate` pattern

It's simply try catching the existing filestore methods and rethrowing with a `FileStoreException`
so
```
try
{
  ... File.Create
}
catch (Exception e)
{
_log...
throw new FileStoreException ...
}

hishamco:
Sorry I have an internet issue, it will be it that you handle this

PR #9414, opened 2021-05-10
Attempt to properly build field values

### LINKED ISSUE #9103: Fields are not auto welded onto the kernel
### Describe the bug
When creating a content item via the API fields are not auto welded onto the kernel like parts are. This means that you need to either remember to create a new Field via code for every field in your content item otherwise you will not be able to save the item after editing in the Admin UI OR you might get Indexing errors when you save the item via the API.

### To Reproduce
Create a content item that has fields via the API and don't specify some fields in your code

### Expected behavior
I would expect that fields are automatically initialized like parts when using the API.

### Workaround
Workarounds for this problem are:

Apply a field for each item via the API even if it has no value like so:
```
myPart.Apply("Summary", new TextField());
```
Or make sure you initialize each field in your model if you have a model like so:
```
public TextField Summary { get; set; } = new TextField();
```

jptissot:
fix #9103

### DIFF
--- a/src/OrchardCore/OrchardCore.ContentManagement.Abstractions/ContentExtensions.cs
+++ b/src/OrchardCore/OrchardCore.ContentManagement.Abstractions/ContentExtensions.cs
@@ -115,8 +115,7 @@ namespace OrchardCore.ContentManagement
                 element.Data = JObject.FromObject(element, ContentBuilderSettings.IgnoreDefaultValuesSerializer);
                 element.ContentItem = contentElement.ContentItem;
 
-                contentElement.Data[name] = element.Data;
-                contentElement.Elements[name] = element;
+                contentElement.Apply(name, element);
             }
 
             return contentElement;
--- a/src/OrchardCore/OrchardCore.ContentManagement/Handlers/ContentPartHandlerCoordinator.cs
+++ b/src/OrchardCore/OrchardCore.ContentManagement/Handlers/ContentPartHandlerCoordinator.cs
@@ -62,6 +62,17 @@ namespace OrchardCore.ContentManagement.Handlers
                 // Iterate existing handler registrations as multiple handlers maybe not be registered with ContentOptions.
                 await _partHandlers.InvokeAsync((handler, context, part) => handler.ActivatingAsync(context, part), context, part, _logger);
                 context.Builder.Weld(typePartDefinition.Name, part);
+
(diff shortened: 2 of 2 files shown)


src/OrchardCore/OrchardCore.ContentManagement.Abstractions/ContentExtensions.cs
```diff
@@ -112,7 +112,7 @@ public static ContentElement Weld(this ContentElement contentElement, string nam
         {
             if (!contentElement.Data.ContainsKey(name))
             {
-                element.Data = JObject.FromObject(element, ContentBuilderSettings.IgnoreDefaultValuesSerializer);
+                element.Data = JObject.FromObject(element);
```
jptissot, line 115: Not sure if this is a good change but it makes the structure of the fields appear in the ContentItem.

deanmarcussen: I think this is a good change, as I see why it was not working without it.

src/OrchardCore/OrchardCore.ContentManagement/DefaultContentManager.cs
```diff
@@ -65,13 +65,13 @@ public async Task<ContentItem> NewAsync(string contentType)
             {
                 ContentType = contentTypeDefinition.Name,
                 Definition = contentTypeDefinition,
-                Builder = new ContentItemBuilder(contentTypeDefinition)
+                ContentItem = new ContentItem() { ContentType = contentTypeDefinition.Name }
```
deanmarcussen, line 68: I think given the fix to `Weld` above that maybe there is no need to remove the `Builder` and change the API?

jptissot: We don't really need to but I thought the builder was a bit redundant and did not provide any value. Small cleanup while I was there.

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/ContentExtensions.cs
jptissot: The downside is that  we will have more json in the database potentially? Although all content created with the admin already has these fields populated to blank values if unfilled.

Issue #8194, opened 2020-12-31
Admin UI : Script defered


Skrypt:
**Is your feature request related to a problem? Please describe.**

Consider using defered scripts instead of declaring them in the page footer.
Declare them all in the header instead like they are supposed to.

**Describe the solution you'd like**

Use `<script defer="defer"></script>`

PR #9430, opened 2021-05-13
Fix a bug when a resource at head and a resource at foot has the same dependency the dependency is rendered at foot

### LINKED ISSUE #9394: KendoUI script error after upgrade to latest dev branch
I developed some modules on Kendo UI framework and that works fine. after upgrade to the latest dev branch, this error occurred on my page:
`Uncaught TypeError: jQuery(...).kendoLoader is not a function`
I think this is because ResorceManagement class moved `<script src="/OrchardCore.Resources/Scripts/jquery.js"></script>` to the end of page.
how can I set `OrchardCore.Resources` scripts inserted in `<head>` tag??

MichaelPetrinolis:
Also added a test

fixes #9394

### DIFF
--- a/src/OrchardCore/OrchardCore.ResourceManagement/ResourceManager.cs
+++ b/src/OrchardCore/OrchardCore.ResourceManagement/ResourceManager.cs
@@ -443,9 +443,6 @@ namespace OrchardCore.ResourceManagement
 
             if (dependencies != null)
             {
-                // share search instance
-                var tempSettings = new RequireSettings();
-
                 for (var i = 0; i < dependencies.Count; i++)
                 {
                     var d = dependencies[i];
@@ -458,17 +455,22 @@ namespace OrchardCore.ResourceManagement
                         version = d[(idx + 1)..];
                     }
 
-                    tempSettings.Type = resource.Type;
-                    tempSettings.Name = name;
-                    tempSettings.Version = version;
+                    var tempSettings = new RequireSettings()
+                    {
+                        Type = resource.Type,
+                        Name = name,
+                        Version = version
+                    };
 
                     var dependency = FindResource(tempSettings);
                     if (dependency == null)
                     {
                         continue;
                     }
-
-                    ExpandDependencies(dependency, dependencySettings, allResources);
+                    tempSettings.Combine((RequireSettings)allResources[dependency] ?? dependencySettings)
+                        .UpdatePositionFromDependent(dependencySettings);
+                    ExpandDependencies(dependency, tempSettings, allResources);
+                    dependencySettings.UpdatePositionFromDependency(tempSettings);
                 }
             }
 
--- a/test/OrchardCore.Tests/ResourceManagement/ResourceManagerTests.cs
+++ b/test/OrchardCore.Tests/ResourceManagement/ResourceManagerTests.cs
@@ -615,6 +615,66 @@ namespace OrchardCore.Tests.ResourceManagement
         }
 
         [Fact]
+        public async Task RenderHeadAndFootScriptWithSameDependency()
+        {
+            var options = new ResourceManagementOptions();
+            var manifest = new ResourceManifest();
+
+            manifest.DefineScript("required").SetUrl("required.js")
+                .SetDependencies("dependency");
+            manifest.DefineScript("dependency").SetUrl("dependency.js");
+            manifest.DefineScript("required-at-foot").SetUrl("required-at-foot.js")
+                .SetDependencies("dependency");
+            manifest.DefineScript("not-required").SetUrl("not-required.js");
+
+            options.ResourceManifests.Add(manifest);
+
+            var resourceManager = new ResourceManager(
+                new OptionsWrapper<ResourceManagementOptions>(options),
+                StubFileVersionProvider.Instance
+            );
+
+            // Require resource
+            resourceManager.RegisterResource("script", "required").AtHead();
+
+            // Register custom script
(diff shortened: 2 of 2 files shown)


Skrypt:
The idea is good. Though, nothing says that rendering at foot won't break the script because of the way it's used. Also, an other option that would be equivalent would be to use defered scripts and keeping them at head.

Skrypt:
Yeah, I agree that this is an issue if we have 2 scripts with the same dependency which are asked to be rendered in 2 different positions. But, what I suggest is to then use defered scripts to fix the issue entirely. Of course there are scripts which are exceptions because they need to be executed at head before any DOM is rendered. 

But in this case, JQuery doesn't require it. The issue is that we made it an habit over time to declare scripts in the footer of the page because we want to make sure the DOM is rendered before executing any scripts. When we are using JQuery it forces you to use the document.ready method to encapsulate your code.

https://learn.jquery.com/using-jquery-core/document-ready/

These are equivalent : 
```javascript
// A $( document ).ready() block.
$( document ).ready(function() {
    console.log( "ready!" );
});
```

```javascript
// Shorthand for $( document ).ready()
$(function() {
    console.log( "ready!" );
});
```

If you want to make sure both your JQuery and KendoUI script to be executed after the DOM is rendered then you declare them as defered. Here, I don't know what does KendoUI but it should be rendered after the JQuery script if there is dependency on it.

So, as stated in the issue the KendoUI script is declared at head while the JQuery is declared at foot. JQuery should be declared before KendoUI so KendoUI should be declared at foot with a dependency on JQuery so that it gets rendered under the JQuery script.

Or, keep KendoUI at head, move JQuery back at head too and make them executed as defered.

Or, fiddle the resourcemanager so that a script declared at head to be rendered at foot because of it's dependency ... I don't like that one because no one will know what is happening.

Is there a reason it needs to be declared at head in this case?

MichaelPetrinolis:
Thanx for the explanation. Here is what I encountered in a custom admin theme and the reason I created this PR.
```html
<script asp-name="font-awesome" at="Head" version="5"></script>
<script asp-name="font-awesome-v4-shims" at="Head" version="5"></script>
<script asp-name="jQuery" at="Head"></script>
<script asp-name="kendo-aspnetmvc" at="Head"></script>
<script asp-name="js-cookie" at="Foot" version="2"></script>
<script asp-name="admin" asp-src="~/TheAdmin/Scripts/TheAdmin.min.js" debug-src="~/TheAdmin/Scripts/TheAdmin.js" at="Foot"></script>
```

As you can see, the jQuery is registered at Head but because js-cookie depends on jQuery the jQuery is rendered at Foot. 
I tried to add defer to jQuery and kendo-aspnetmvc but defer attribute is not rendered. Also, suppose it rendered, the jQuery would still appear after kendo unless the order of scripts changes.

I registered all at Foot, but still, it did not work. The reason is that the Kendo's tag helpers render inline scripts like the following (which are not defered)
```javascript
<script>
kendo.syncReady(function () {
    jQuery("#grid").
    kendoGrid({
        "groupable": {
            "enabled": true
        },
        "height": 550,
        "pageable": {
            "buttonCount": 5,
            "refresh": true,
            "pageSizes": [5, 10, 20]
        },
        "sortable": true,
        "filterable": true,
        "columns": [{
                "field": "Field A",
                "title": "Field B",
                "width": 120
            }
        ],
        "toolbar": [{
                "name": "search"
            }, {
                "text": "Add",
                "name": "custom",
                "template": template
            }
        ],
        "dataSource": {
            "pageSize": 20,
            "transport": {
                "read": {
                    "url": "/Admin/Area/Controller/Action"
                }
            }
        }
    });
});
</script>
```
and throw exceptions like ```ReferenceError: kendo is not defined```

Skrypt:
Defered script gets called in the order they are declared. So, if you put your KendoUI script before the JQuery one then it will not work. Also, not sure, but I think we also don't support defered and async scripts in OC. I opened an issue about it a while ago.

https://github.com/OrchardCMS/OrchardCore/issues/8194

I see your point there. If there is inline scripting like this then you have no choice to declare your 2 dependencies at head. I would recommend to see if there is any configurations in KendoUI or ways to make these executed after document.ready though. Else, this is really bad practice.

Skrypt:
Another trick would be to make these inline scripts to be rendered with a defer attribute.

jtkech:
For infos now there is also a new `Position` property that can be set to `First / Last / ByDependency (the default)`, so that we can force a set of scripts to be before / after other ones, so we have to take care to not mix `Location` and `Position`. Note the `Position` property can't be set through a tag helper, only through a resource manifest.

---

That said if A at Foot depends on B at Head, B should stay at Head

I found the issue, i tried it it works, one line of code, will do a PR this night ;)

Skrypt:
Nevermind, you can't defer an inline script (of course) ... it's inline ...

Skrypt:
Ok found something interesting. You can defer inline scripts by doing this to them : 

```javascript
<script type="module">
// your inline javascript
</script>
```

Skrypt:
Also found this : https://docs.telerik.com/aspnet-mvc/getting-started/helper-basics/fundamentals

jtkech:
@Skrypt very interesting 👍

jtkech:
@MichaelPetrinolis 

I did #9437 that is intended to replace this one but never 100% sure ;)

So if you could try it to confirm that it fixes the issue

MichaelPetrinolis:
superseded by #9437

PR #9437, opened 2021-05-14
Fixes #9394 Resource dependency location

### LINKED ISSUE #9394: KendoUI script error after upgrade to latest dev branch

jtkech:
If a resource A at Foot depends on a resource B at Head, B should stay at Head

Fixes #9394


MichaelPetrinolis:
@jtkech this one liner fixes the issue :)
you can also copy the unit test from #9430 or I can send a PR after merging this one

PS
I thought that introducing Position broke it, but Location was always not included in Combine

jtkech:
@MichaelPetrinolis 

Thanks for the feedback

> I thought that introducing Position broke it, but Location was always not included in Combine

Yes this is what i saw while debugging to see step b step where the position was 1st overwritten ;)

deanmarcussen:
@jtkech looks good. can you grab the unit test and bring it along as well (otherwise we will forget) ?

jtkech:
Okay will do

MichaelPetrinolis (approved):

PR #9470, opened 2021-05-18
Deployment steps

agriffard:
Add import/export deployment steps.

FacebookLogin
GoogleAnalytics
Microsoft Azure AD
OpenIdValidationSettings


src/OrchardCore.Modules/OrchardCore.OpenId/Startup.cs
```diff
@@ -204,6 +199,17 @@ public override void ConfigureServices(IServiceCollection services)
             });
         }
 
+        [RequireFeatures("OrchardCore.Deployment")]
```
deanmarcussen, line 202: Needs to be outside the `ServerStartup` class.

And in this case it would be `[RequireFeatures("OrchardCore.Deployment", OpenIdConstants.Features.Server)]` to trigger the startup.

Same below with validation

deanmarcussen (approved):

agriffard:
Related to #5558

PR #9472, opened 2021-05-18
Check allowed file extensions when importing media

deanmarcussen:
via a recipe

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Media/Recipes/MediaStep.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Recipes/MediaStep.cs
@@ -1,8 +1,12 @@
 using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Net.Http;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


src/OrchardCore.Modules/OrchardCore.Media/Recipes/MediaStep.cs
```diff
@@ -32,6 +43,13 @@ public async Task ExecuteAsync(RecipeExecutionContext context)
 
             foreach (var file in model.Files)
             {
+                if (!_allowedFileExtensions.Contains(Path.GetExtension(file.TargetPath), StringComparer.OrdinalIgnoreCase))
+                {
+                    _logger.LogWarning("File extension not allowed: '{0}'", file.TargetPath);
```
sebastienros, line 48:
```suggestion
                    _logger.LogWarning("File extension not allowed: '{Path}'", file.TargetPath);
```

deanmarcussen: Done (and fixed the other place I cut and pasted from)

Issue #9413, opened 2021-05-10
"Database is locked" error when using ChangeEmailAsync or SetEmailAsync with custom action


orchardc:
### Describe the bug
I think this is very much repeatable...if a custom update follows with **_userService.ChangeEmailAsync** or **_userManager.SetEmailAsync**
It will always throw SQLite "database is locked" error    

In a HttpPost action, 
`public async Task<IActionResult> ChangeEmail(string email)
{
  if (await _userService.ChangeEmailAsync(user, model.Email,
                        (key, message) => ModelState.AddModelError(key, message)))
  {
 	//then update the custom table, it will throw SQLite "database is locked" error    
 	if (ModelState.IsValid)
	{
		var myModel = new MyModel(){
			Id = model.Id,
			Email = model.Email
		};
  		await _myRepo.UpdateUser(myModel); //throw "database is locked" error    
	}
  }

}`

Issue #9363, opened 2021-05-06
Cancelling the transaction not works as expected


intellimedhu:
I have do the following in the same transaction:

1. Create a **new version** from a ContentItem
2. Create a new User

I call `ISession.CancelAsync()` when exception is thrown but a new version from a content item is created as latest and unpublished. As if user creation would commit the transaction. (Without creating a new user it works as expected.)

### To Reproduce

```
try
{
    var contentItem = await _contentManager.GetAsync("id", VersionOptions.DraftRequired);
    var user = _userService.CreateUserAsync(...);
    ...
    throw new Exception();

    await _contentManager.PublishAsync(contentItem);
}
catch
{
    await _session.CancelAsync();
    throw;
}
```

### Expected behavior
ContentItem will have the original version as latest and published.

I know reversing step 1 and 2 will solve the issue, but doing with CancelAsync() would be the best.

jtkech:
You don't await `CreateUserAsync()`

intellimedhu:
You're right, it's a misspell, sorry. However the problem still exists.

jtkech:
Yes the user store call directly a session commit, as i remember it is related on how the related aspnetcore services work

In fact now it does a `_session.SaveChangesAsync()` that does a flush as i remember, but normally if you do a session cancel, the transaction should not be committed, at some point it was commited but it has been changed to keep the cancel state even after a flush, i will check it tomorrow

Which version are you using?

intellimedhu:
Ok, thanks!

version: 1.0.0-rc2-16169

One more thing, probably related to this issue: I do the user creation first then updating the content item. When the user is created and an exception is thrown I try to delete the user in two ways:

1:
```
await _userManager.DeleteAsync(user); // UserManager<IUser>
await _session.CancelAsync();
```
The problem is the same: user deletion will commit the transaction, so CancelAsync makes no sense.

2.
```
await _session.CancelAsync();
await _userManager.DeleteAsync(user);
```
Content changes will be rolled back, but deleting the user fails.

deanmarcussen:
> Yes the user store call directly a session commit, as i remember it is related on how the related aspnetcore services work

For infos @jtkech this is a behaviour we have talked about needing to change.

While EF/Identity would do this by default, EF/Identity, by default, difer in that they do not treat a the DbContext as a per request, i.e. cancellable, single transaction oer the whole request, by default.
It's something you can work into EF, but isn't done in the default patterns.

[...]

PR #9481, opened 2021-05-19
New User Event Handlers

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/9413
Fixes https://github.com/OrchardCMS/OrchardCore/issues/9363

- Removes `SaveChangesAsync` from the `UserStore` and `OpenIdStores` (except when pruning, where it still makes sense.
- Adds three new event handlers to the user events, `CreatingAsync`, `DeletingAsync`, and `UpdatingAsync` - this to bring the event handlers inline with what we do with other events, like content item events, allowing custom logic through event handlers to cancel sessions, etc before persisting. Means less having to swap the `UserStore` out to get custom behaviours, depending on other logic.

The reason being removing `SaveChangesAsync` is that as a pattern through OrchardCore we save transactions through the request session, allowing features like `CancelAsync` to work, but when we save early like this, these features can't work.

/cc @kevinchalet for infos - the OpenId not as blocking for us (Users has been the biggest issue), but trying to bring things inline with everything else, it does mean these become YesSql concurrency exceptions, rather then OpenIddict concurrency exceptions. Open to options here.

/cc @jtkech

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Workflows/Controllers/WorkflowTypeController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Workflows/Controllers/WorkflowTypeController.cs
@@ -461,8 +461,8 @@ namespace OrchardCore.Workflows.Controllers
             }
 
             await _workflowTypeStore.SaveAsync(workflowType);
-            await _session.SaveChangesAsync();
             _notifier.Success(H["Workflow has been saved."]);
+
             return RedirectToAction(nameof(Edit), new { id = model.Id });
         }
 
--- a/src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Stores/OpenIdApplicationStore.cs
+++ b/src/OrchardCore/OrchardCore.OpenId.Core/YesSql/Stores/OpenIdApplicationStore.cs
@@ -41,7 +41,7 @@ namespace OrchardCore.OpenId.YesSql.Stores
             => throw new NotSupportedException();
 
         /// <inheritdoc/>
-        public virtual ValueTask CreateAsync(TApplication application, CancellationToken cancellationToken)
+        public virtual async ValueTask CreateAsync(TApplication application, CancellationToken cancellationToken)
         {
             if (application == null)
             {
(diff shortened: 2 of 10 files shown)


jtkech:
Thanks @deanmarcussen, it was on my list but quite busy on Audit Trail.

Hmm, i will have to take into account the new user event handlers ;)

I already use the `IDocumentStore.AfterCommitSuccess()` in the `DocumentManager`, but there is also an `.AfterCommitFailure()` allowing to register a `DocumentStoreCommitFailureDelegate`, if there is no delegate the exception is thrown as is, otherwise we execute the delegates that can e.g. throw custom exceptions.

I will take a look when i will have time

kevinchalet:
> /cc @kevinchalet for infos - the OpenId not as blocking for us (Users has been the biggest issue), but trying to bring things inline with everything else, it does mean these become YesSql concurrency exceptions, rather then OpenIddict concurrency exceptions. Open to options here.

This is a problem: OpenIddict uses this type to ignore exceptions caused by concurrent updates and still return a response that is usable by the client. If the exception is thrown later in the pipeline by YesSql itself, chances are high all clients will get are 500 exceptions.

deanmarcussen:
> This is a problem:

Ok, so given the issue we are fixing presents with `Users`, and we wouldn't expect to handle any `OpenId` operations with Orchard event handlers, or for that matter do any unrelated session work in an `OpenId` request, I will revert the `OpenId` changes

kevinchalet:
FWIW, this may be problematic for Identity too: some of the methods exposed by the user manager rely on the fact DB-related errors are thrown as soon as you call things like `Create` or `Update`. If they pop out at a later stage, you might see weirds things.

deanmarcussen:
Hmm, they don't make it simple. 

So maybe what we need here is the extra event handlers that fire before `SaveChangesAsync` and public access to YesSql's `_cancelled` property to know if a session has been cancelled by a handler, thereby being able to return the correct `IdentityResult` from the store.

deanmarcussen:
Updated to keep `SaveChangesAsync` , i.e. session commit, but allow a `Cancel` from within the `Creating` etc handlers.

This covers the two use cases that keep coming up
- ability to cancel a user operation from a handler (technically this could also be done with an Identity Validation Handler, but for some consistency with OC, seems worthwhile to have this in our code base). Matches the `PublishContentContext` event handler, and keeps the handlers from being fixed to Identity, as they would have been if using a `IdentityResult`.
- ability to mutate user properties from event handlers before they are persisted (the most recent requirement here for us was to update the `EmailConfirmed` property when we have already externally validated the email address through other pre registration channels)
- keeps consistent the `IdentityResult` returned to the `UserManager` so Identity remains happy.

What this won't do is respect a `_session.CancelAsync()` - but doing so loses any `UserManager` consistency.

@kevinchalet does that cover your concerns?

kevinchalet:
Sounds like a reasonable plan! 👍🏻

jtkech (approved):

PR #9495, opened 2021-05-20
Move SelectContentTypesViewComponent to OC.Contents

jtkech:
- For now this component is defined in `OC.ContentTypes` that depends on `OC.Contents`

- `OC.Contents` already uses this component when Workflow is enabled and will use it for AuditTrail

- But `OC.Contents` doesn't depend on `OC.ContentTypes`, this is the opposite.

- That's why I moved the component to `OC.Contents`

Note: `OC.ContentTypes` allows to manage content types through the UI but this is `OC.Contents` that registers content types definitions services


sebastienros:
Ok. It might be used in more locations already. Did you check everything?

jtkech:
I don't remember, at least the new usage i needed in OC.Contents

I will check it in other places

jtkech:
Okay, just tried with

- The list part settings

- Bag part settings

- ContentPublished WF activity

Works as before

deanmarcussen (approved):

PR #9497, opened 2021-05-20
Add DistributedLock for Multi-instance AutoSetup

rserj:
I'm running multiple instances of OrchardCore with AutoSetup.
When all instances start simultaneously and trigger autosetup, some of the instances failed to install with the error below

```
System.InvalidOperationException: Invalid serial number for shell descriptor
   at OrchardCore.Environment.Shell.Data.Descriptors.ShellDescriptorManager.UpdateShellDescriptorAsync(Int32 priorSerialNumber, IEnumerable`1 enabledFeatures)
   at OrchardCore.Setup.Services.SetupService.<>c__DisplayClass13_1.<<SetupInternalAsync>b__1>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell)
   at OrchardCore.Setup.Services.SetupService.SetupInternalAsync(SetupContext context)
   at OrchardCore.Setup.Services.SetupService.SetupAsync(SetupContext context)
   at OrchardCore.AutoSetup.AutoSetupMiddleware.SetupTenantAsync(ISetupService setupService, TenantSetupOptions setupOptions, ShellSettings shellSettings)
   at OrchardCore.AutoSetup.AutoSetupMiddleware.InvokeAsync(HttpContext httpContext)
```

I tried to use Distributed lock implementation while Shell Activatates but it always acquires Local lock, this PR fixes this part.

I have tested the fix by enabling Reis Lock modules like this
```
services.AddOrchardCms()
                .AddTenantFeatures("OrchardCore.Redis", "OrchardCore.Redis.Lock")
                .AddSetupFeatures("OrchardCore.AutoSetup");
```
and configure redis and shell settings via env. variables
```
"OrchardCore__OrchardCore_Redis__Configuration": "localhost",
"OrchardCore__ShellActivateLockTimeout": "5000",
"OrchardCore__ShellActivateLockExpiration": "5000",
"OrchardCore__ShellActivateLockName": "MyRedisLockKey"
```
but didn't test autosetupping multiple instances at the same time.

@jtkech can you please take a look?


jtkech:
@rserj 

Yes, indeed it is not intended to use a distributed lock while the tenant is initiailizing during a setup

So first step, i let you try your change with autosetupping multiple instances at the same time

Then we will see what happens ;)

rserj:
@jtkech should we use Distributed lock in the SetupService instead? by doing this we can rid off of individual locks in  SetupController and AutosetupMiddleware?

src/OrchardCore/OrchardCore.Abstractions/Shell/Builders/ShellContextExtensions.cs
```diff
@@ -32,9 +30,10 @@ public static Task<(ILocker locker, bool locked)> TryAcquireShellActivateLockAsy
                 var options = shellContext.ServiceProvider.GetRequiredService<IOptions<ShellContextOptions>>();
                 timeout = TimeSpan.FromMilliseconds(options.Value.ShellActivateLockTimeout);
                 expiration = TimeSpan.FromMilliseconds(options.Value.ShellActivateLockExpiration);
+                lockName = options.Value.ShellActivateLockName;
```
sebastienros, line 33: Why would it be different than the tenant name?

sebastienros: Unless you want to block all tenants creation "by default"? But with this change this is what it would do, you couldn't create two tenants concurrently.

rserj: I thought about this, but we may run many single-tenant instances on the same Reids in such case lockName always will be equal to "Default", the same issue with DPAPI Redis as I know we have a hardcoded "Default" tenant name if we run orchard in Single Tenant mode.

jtkech: For infos we already setup tenants concurently in unit tests but they all have different generated names

For info our RedisLock automaticaly prefix the lock key with the tenant name, so the unprefixed lock key should be the same, not an option that may be different for 2 instances running for the same tenant, unless we are sure they use the same config source, but i think it is better to let it hardcoded.

Hmm, but do you mean that you may run multiple single-tenant instances that are unrelated, not related to the same 
default tenant, but different ones using different databases ? But using the same Redis instance ?

jtkech: But i don't think so as your issue is a concurrent acces on the shell descriptor so in the same database and for the same tenant `Invalid serial number for shell descriptor`

jtkech:
@rserj 

> Should we use Distributed lock in the SetupService instead? by doing this we can rid off of individual locks in SetupController and AutosetupMiddleware?

Yes better

src/OrchardCore/OrchardCore.Abstractions/Shell/Builders/ShellContextExtensions.cs
rserj: >but different ones using different databases ? But using the same Redis instance ?

Correct, we may run single-instance for each customer but share the same Redis instance, the configuration may vary depending on env. Testing Env, Staging, and Prod, but basically yes this is a scenario.
Having a Single Tenant gives us some benefits like updating the env without worrying about breaking other customers and so on.

jtkech: Okay so that's why you want an option e.g. to differentiate 2 different tenants, not related to the same database, but with the same name e.g. `Default`. Hmm, but then you will need the same e.g. if you want to share data with our RedisCache.

Through the redis config string we can define a `Default database index`, i think this is the right way to differentiate tenants not related to the same database, in place of changing / prefixing lock keys.

https://stackexchange.github.io/StackExchange.Redis/Configuration.html

rserj: I may be wrong, but I think using DatabaseIndex as a Tenant delimiter is not a good idea, we might face some performance issues in such a case.

rserj:
@jtkech 
```
.AddSetupFeatures("OrchardCore.AutoSetup", "OrchardCore.Redis.Lock");
```
I tried this, made one more commits, can you please take a look
If you are ok with the change can you please point out the places where I should delete SemaforLocks?

jtkech:
> can you please point out the places where I should delete SemaforLocks?

If RedisLock is enabled you will get it by resolving ILock without having to remove any semaphore.

But because it is related to the AutoSetup, finally i think it would be better to use the distributed lock per tenant name in the auto setup middleware (you can let the usage of the explicitly defined local semaphore), and let the setup controller and setup service unchanged.

---

So, for a given tenant name you can have multiple instances related to the same database OR NOT, but all sharing the same Redis instance. So, e.g. to not lock 2 instances that are not related to the same database you want to use a different lock key, but as said you will have conflicts with other Redis features as our RedisCache.

But maybe your data are not intended to be changed and you don't want to use the RedisCache, only the Redis lock to be able to use the AutoSetup in this scenario. If so i understand but i still think that it would be better to use different Redis instances or at least different redis database indexes.

Is the perf a concern if the redis instance is only used for the auto setup? And if the Redis instance is also used afterwards, it would mean that you want to use other Redis features, and in that case you will have conflicts.

So it is related to the auto setup and in a specific scenario

Can you provide a more complete description of your possible scenarios? Maybe i didn't understand them.

rserj:
>So, for a given tenant name you can have multiple instances related to the same database OR NOT, but all sharing the same Redis instance. So, e.g. to not lock 2 instances that are not related to the same database you want to use a different lock key, but as said you will have conflicts with other Redis features as our RedisCache.
Is the perf a concern if the redis instance is only used for the auto setup? And if the Redis instance is also used afterwards, it would mean that you want to use other Redis features, and in that case you will have conflicts.

Correct, I have to rewrite the current Redis feature, particularly Redis DPAPI we use custom "Prefix" different from the Tenant name. We do not use RedisCashe yet, but when we will I afraid I need to rewrite it too.

>Can you provide a more complete description of your possible scenarios?

For the QA Environments, there is no need to create Redis instance for each Customer, so we may share one among clients, please see the image below
image

jtkech:
@rserj 

Okay, so in place of configuring key names, what i could do is to add to the Redis config a global prefix (empty by default), not related to a given tenant name and that will be used by all Redis services RedisCache / RedisLock ... including the related DP and `IDistributedCache`, so that by code you will be able to consume all services as before.

So a new configurable `RedisOptions` property (e.g `InstanceName`) allowing to share a given Redis database.

I will do a PR soon

rserj:
@jtkech thank you for your PR

jtkech:
So here there are 2 things

- The ability to share the same Redis instance by 2 sets of OC instances not related to the same database, which was not intended to be supported but could be allowed by introducing a Redis `InstancePrefix` as done in #9522.

- The ability to run the auto setup concurrently by multiple OC instances, which can be done in a separate PR, e.g. this one with a new title because it's not about fixing the distributed lock but using it for auto setup.

That said, some first thoughts about auto setup atomicity

- Autosetup atomicity is not sufficient, in fact i think it needs to be exclusive, after having acquired a lock, a given tenant shoul not be setup if it has been done by another OC instance. This can be done by using a low timeout to acquire a lock, if the lock is acquired do the setup, otherwise just do a tenant release so that it will be rebuild on the next request by using the updated data from the database.

- We could use a lock per tenant so that each running OC instance can setup different tenants at the same time.

- You also need a distributed configuration, e.g. not using `tenants.json` and tenant specific `appsettings.json` files (under each tenant folder), but using `.AddDatabaseShellsConfiguration()` to retrieve these configs from the database.

jtkech:
@rserj No problem, it is interesting;)

Just posted the above comment

rserj:
>.AddDatabaseShellsConfiguration()

Correct, I'm loading shell's configuration instead of files

jtkech:
@rserj so what i suggest

- Replace the semaphore by IDistributedLock as you did

- Then, as said we could use the Distributed Tenants feature that keeps tenants in sync by sharing settings identifiers through the RedisCache, but it only works if the default tenant is running (see `DistributedShellHostedService`).

    So in place of reading the RedisCache, i suggest to use the ShellSettingsManager to see if the tenant was updated. If we don't acquire the lock, we assume that the tenant was setup by another instance, so we just need to reload it.

    Note: Below we use `eventSource: false` because here we don't update the settings before reloading, so we are more reacting to a possible "event" by reloading the settings from the settings sources.

Too lazy to try it but something like this

            if (_setupOptions != null && _shellSettings.State == TenantState.Uninitialized)
            {
                (var locker, var locked) = await _distributedLock.TryAcquireAutoSetupLockAsync(...);
                if (!locked)
                {
                    throw new TimeoutException(...);
                }

                if (_shellSettings.State == TenantState.Uninitialized)
                {
                    var pathBase = httpContext.Request.PathBase;
                    if (!pathBase.HasValue)
                    {
                        pathBase = "/";
                    }

                    var names = await _shellSettingsManager.LoadSettingsNamesAsync();
                    if (names.Contains(_shellSettings.Name))
                    {
                        await _shellHost.ReloadShellContextAsync(_shellSettings, eventSource: false);

                        httpContext.Response.Redirect(pathBase);

                        return;
                    }

                   ...
                   ...

src/OrchardCore.Modules/OrchardCore.AutoSetup/OrchardCore.AutoSetup.csproj
```diff
@@ -18,6 +18,10 @@
     <ProjectReference Include="..\..\OrchardCore\OrchardCore.Setup.Abstractions\OrchardCore.Setup.Abstractions.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="Extensions\" />
+  </ItemGroup>
+
```
jtkech, line 25: You will have to remove these extra item groups

src/docs/OrchardCore.Docs.csproj
```diff
@@ -7,4 +7,8 @@
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Folder Include="reference\modules\AutoSetup\assets\" />
+  </ItemGroup>
+
```
jtkech, line 13: Idem, remove

jtkech:
@rserj Did you try my above example?

The idea being that when an OC instance acquires a lock (distributed), we still check the internal shell settings for local thread safety as before if multiple request e.g. by a browser on the same instance.

Then we check through the shell settings manager if the tenant exists, meaning that it has been setup by another instance that has updated its settings, and in that case we just have to do a reload => the tenant settings are loaded and the tenant is released so that on a new request (here triggered by the redirect) it is fully built.

Otherwise, even it is atomic it will fail if an OC instance try to setup a tenant that is already setup by another instance.

rserj:
I did try to make multiple requests in a browser with code above
```
// Check if tenant was Installed by another instance
                    var names = await _shellSettingsManager.LoadSettingsNamesAsync();
                    if (names.Contains(_shellSettings.Name))
                    {
                        await _shellHost.ReloadShellContextAsync(_shellSettings, eventSource: false);
                        httpContext.Response.Redirect(pathBase);

                        return;
                    }
```

it went to infinite redirection loop when I tried to install /tenant on demand

rserj:
@jtkech  can I somehow refer to current PR artifacts, to nuget packages in order to test this PR?

jtkech:
> it went to infinite redirection loop when I tried to install /tenant on demand

Okay i would need to try it ;) But normally it means that you want to setup a tenant that already exists

Oh yes, you are pre-creating them (not the default one) and save their settings with `UpdateShellSettingsAsync()`, so okay we need to use `LoadSettingsAsync(string tenant)` and check its state, i will update my example.

> can I somehow refer to current PR artifacts, to nuget packages in order to test this PR?

I don't think so, the CI doesn't build packages for a PR, you would need to do it yourself and put the packages e.g. in a local feed.

jtkech:
Okay, can you try the following, again i didn't try it, just to progress step by step ;)

Note: But it still relies on the RedisLock to not run a given setup concurently

        if (_setupOptions != null && _shellSettings.State == TenantState.Uninitialized)
        {
            (var locker, var locked) = await _distributedLock.TryAcquireAutoSetupLockAsync(...);
            if (!locked)
            {
                throw new TimeoutException(...);
            }

            if (_shellSettings.State == TenantState.Uninitialized)
            {
                var pathBase = httpContext.Request.PathBase;
                if (!pathBase.HasValue)
                {
                    pathBase = "/";
                }

                var settings = await _shellSettingsManager.LoadSettingsAsync(_shellSettings.Name);
                if (settings.State != TenantState.Uninitialized)
                {
                    await _shellHost.ReloadShellContextAsync(_shellSettings, eventSource: false);

                    httpContext.Response.Redirect(pathBase);

                    return;
                }

               ...
               ...

src/OrchardCore.Modules/OrchardCore.AutoSetup/Extensions/DistributedLockExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// </returns>
+        public static Task<(ILocker locker, bool locked)> TryAcquireAutoSetupLockAsync(this IDistributedLock distributedLock, string shellName, LockOptions lockOptions)
+        {
+            if (lockOptions == null)
+            {
+                throw new ArgumentNullException(nameof(lockOptions));
+            }
+
+            return distributedLock.TryAcquireLockAsync(
+                $"{LockPrefix}_{lockOptions.LockName ?? shellName}",
```
jtkech, line 52: So here we don't need any prefix, configurable LockName, and don't need the shell name, we only need a lock key e.g. "AUTOSETUP_LOCK", because each tenant will be setup by itself on demand, the tenant name will be used as a prefix automatically by the RedisLock.

And about sharing the same Redis Instance by different applications, as said it is another concern, would be good to first try it with only one set of OC instances related to the same database.

Then if you want to try it with 2 sets of OC instances sharing the same Redis instance, if it is easier for you maybe i will merge #9522, i think it is safe to be merged because it only adds a prefix that is null by default

src/OrchardCore.Modules/OrchardCore.AutoSetup/Extensions/DistributedLockExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+                TimeSpan.FromMilliseconds(lockOptions.LockTimeout == 0 ? DefaultLockTimeout : lockOptions.LockTimeout),
+                TimeSpan.FromMilliseconds(lockOptions.LockExpiration == 0 ? DefaultLockExpiration : lockOptions.LockExpiration));
```
jtkech, line 54: I would have check `<= 0`, just in case of a negative value

jtkech: So, if at the end we only use the distributed lock in a simple way, maybe we don't need this extension at all.

Unless if we check that it is a `ILocalLock`, as in `TryAcquireShellActivateLockAsync()`, and in that case don't use any timeout and expiration by setting them to `TimeSpan.MaxValue`. I did this for shell because it is a sensitive place, so that if it is a local lock it just works as before with a simple semaphore without any timeout / expiration.

So, the only difference with `TryAcquireShellActivateLockAsync()` would be that we would not have to check if it is `Initializing`, just check if it is a local lock.

But as you want, i'm not against using the distributed lock in the simplest way, but in that case maybe this extension is not necessary.

src/OrchardCore.Modules/OrchardCore.AutoSetup/Options/AutoSetupOptions.cs
```diff
@@ -16,6 +16,11 @@ public class AutoSetupOptions : IValidatableObject
         /// </summary>
         public string AutoSetupPath { get; set; }
 
+        /// <summary>
+        /// AutoSetup Lock Options
+        /// </summary>
+        public LockOptions LockOptions { get; set; } = new LockOptions();
```
jtkech, line 22: maybe not necessary

src/OrchardCore.Modules/OrchardCore.AutoSetup/Extensions/DistributedLockExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// </param>
+        /// <param name="lockOptions">
+        /// The setup options.
+        /// </param>
+        /// <returns>
+        /// The <see cref="Task"/>.
+        /// </returns>
+        public static Task<(ILocker locker, bool locked)> TryAcquireAutoSetupLockAsync(this IDistributedLock distributedLock, string shellName, LockOptions lockOptions)
+        {
+            if (lockOptions == null)
```
jtkech, line 46: If we keep this class, maybe to much as it is initialized with `= new LockOptions();`

src/OrchardCore/OrchardCore.Abstractions/Shell/Builders/ShellContextExtensions.cs
```diff
@@ -15,12 +15,10 @@ public static class ShellContextExtensions
         /// </summary>
         public static Task<(ILocker locker, bool locked)> TryAcquireShellActivateLockAsync(this ShellContext shellContext)
         {
-            // If the shell is initializing, force the usage of a local lock.
-            var lockService = shellContext.Settings.State == TenantState.Initializing
-                ? (ILock)shellContext.ServiceProvider.GetRequiredService<ILocalLock>()
-                : shellContext.ServiceProvider.GetRequiredService<IDistributedLock>();
+            var lockService = shellContext.ServiceProvider.GetRequiredService<IDistributedLock>();
```
jtkech, line 19: Revert the change of this class, i explained above why i did it, yes maybe good to not check `Initializing` but for now a setup is not intended to be run concurrently, and if so we just let it fail.

Then, we want to be able to run the AutoSetup concurrently, but this is at another layer that we will allow this, by using the RedisLock in the AutoSetup middleware.

So we can revert the changes of `ShellContextExtensions`.

src/OrchardCore/OrchardCore.Abstractions/Shell/Builders/ShellContextOptions.cs
```diff
@@ -13,5 +13,11 @@ public class ShellContextOptions
         /// Note: Only used if the current distributed lock implementation is not a local lock.
         /// </summary>
         public int ShellActivateLockExpiration { get; set; }
+
+        /// <summary>
+        /// The distributed lock key which used while activating a shell.
+        /// Note: Only used if the current distributed lock implementation is not a local lock.
+        /// </summary>
+        public string ShellActivateLockName { get; set; }
```
jtkech, line 21: Revert these changes too

src/docs/reference/modules/AutoSetup/README.md
```diff
@@ -124,4 +124,40 @@ To enable the Auto Setup feature, it is necessary to add it in the Web project's
...
+| `LockTimeout` | The timeout in milliseconds to acquire a distributed auto setup lock. | 20 seconds |
+| `LockExpiration` | The expiration in milliseconds of the distributed setup lock. | 20 seconds |
+| `LockName` | The distributed setup lock name. | "AUTOSETUP_LOCK_{SHELL-NAME}" |
+
+Lock configuration parameters are optional and can be set via Env. Variables or configuration file 
+
+```
+"OrchardCore__OrchardCore_AutoSetup__LockOptions__LockTimeout": "10000"
+"OrchardCore__OrchardCore_AutoSetup__LockOptions__LockExpiration": "10000"
+"OrchardCore__OrchardCore_AutoSetup__LockOptions__LockName": "Orchard_AutoSetup_Lock"
```
jtkech, line 162: To be updated if we don't use a `LockOptions` child class.

src/OrchardCore/OrchardCore.Abstractions/Shell/Builders/ShellContextExtensions.cs
rserj: Ok, will revert those changes, I tested this part it always resolves ILocalLock -> LocalLock even if RedisLock Feature enabled, meaning that part `shellContext.ServiceProvider.GetRequiredService<IDistributedLock>();` will never be used unless I am  missing something

jtkech: First, if necessary i will have to try it but for now i let you try it with your existing configuration, so for now on my side it's only theorical ;)

So, if you add the RedisLock to the setup features it doesn't mean that when activating the internal shell, that we built for the setup, includes this feature.

Hmm, even if you add the RedisLock feature to your setup recipe, the first time the shell is activated and will execute the Features recipe step, the RedisLock feature is not yet enable.

The important thing to check is, when adding the Redislock to the setup features, that we resolve the RedisLock in the auto setup middleware, is it the case?

jtkech:
@rserj just tried but with only one OC instance and a Redis instance

All as expected, if we add the Redis.Lock to the setup features

            services
                .AddOrchardCms()
                .AddSetupFeatures("OrchardCore.AutoSetup", "OrchardCore.Redis.Lock");

In the AutoSetup middleware, the injected lock service is RedisLock an i can see the prefix being the tenant name.

Then i removed the check of `Initializing` when getting a lock to activate the shell, the first shell activation, before doing the setup, and in the context of which we execute the auto setup middleware, this first activaion resolves the RedisLock.

Then, the first activation, of the internal shell we use to do the setup, get the LocalLock, idem afterwards. Finally if i add the Redis.Lock to the setup recipe, idem for the first activation of this internal shell, because the features step is not yet executed, then subsequents activation resolve the RedisLock.

So, the important thing being to resolve the RedisLock in the middleware seems to be okay.

So, try to revert the changes i mentioned, including the additional prefix, configurable key names and so on, and try the auto setup but with only one set of OC instances related to the same database, then in a second step we will try your scenario by just including #9522.

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
```diff
@@ -94,42 +113,52 @@ public async Task InvokeAsync(HttpContext httpContext)
         {
             if (_setupOptions != null && _shellSettings.State == TenantState.Uninitialized)
             {
-                await _semaphore.WaitAsync();
-                try
+                // Try to acquire a lock before starting installation, it guaranty Atomic setup in multi instance environment.
```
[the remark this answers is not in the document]
rserj: Sorry but I'm not following, if we use 
IDistributedLock which is by default uses LocalLock which uses Semaphore by default, why do we need to use _semaphore again?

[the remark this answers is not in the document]
jtkech: I didn't say that, my above comment is just to correct eventual typos ;)

rserj:
@jtkech  thanks for your review, I made the requested changes might miss something can you please check?
I will push the changes into my NuGet feed and will see how it goes

jtkech:
@rserj 

I did some updates, feel free to revert some if you don't agree ;)

And i merged #9522 so that you can fully test your use case.

If you think it is worth to update `appsettings.json` for the new lock options, i let you update it as i can't edit it in your branch.

    //"OrchardCore_AutoSetup": {
    //  "AutoSetupPath": "",
    //  "LockOptions": { // Optional e.g. if multiple instances are launched, needs a distributed lock as Redis Lock.
    //    "LockTimeout": 10000,
    //    "LockExpiration": 10001
    //  },

jtkech:
@rserj Are you still there?

Did you try it? Does it work?

Would be interesting to know ;)

rserj:
@jtkech I haven't update orchard core for a while, I need to resolve some issues before I try the fix, I'll let you know about the results, thank you

jtkech:
@rserj Okay cool, nothing urgent for me

rserj:
@jtkech I tested it out couple of times here are the test results:

1. **Without** Redis.Lock feature enabled I got  the same result even with  
 `shellSettingsManager.LoadSettingsAsync(_shellSettings.Name)` fix 

```
2021-06-04 01:59:08.0353 ERROR Connection id "0HM96VHALV3VK", Request id "0HM96VHALV3VK:00000001": An unhandled exception was thrown by the application. System.InvalidOperationException: Invalid serial number for shell descriptor
   at OrchardCore.Environment.Shell.Data.Descriptors.ShellDescriptorManager.UpdateShellDescriptorAsync(Int32 priorSerialNumber, IEnumerable`1 enabledFeatures)
   at OrchardCore.Setup.Services.SetupService.<>c__DisplayClass13_1.<<SetupInternalAsync>b__1>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell)
   at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell)
   at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell)
   at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell)
   at OrchardCore.Setup.Services.SetupService.SetupInternalAsync(SetupContext context)
   at OrchardCore.Setup.Services.SetupService.SetupAsync(SetupContext context)
   at OrchardCore.AutoSetup.AutoSetupMiddleware.SetupTenantAsync(ISetupService setupService, TenantSetupOptions setupOptions, ShellSettings shellSettings)
   at OrchardCore.AutoSetup.AutoSetupMiddleware.InvokeAsync(HttpContext httpContext)
   at OrchardCore.AutoSetup.AutoSetupMiddleware.InvokeAsync(HttpContext httpContext)
   at Microsoft.AspNetCore.Localization.RequestLocalizationMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
   at OrchardCore.Modules.ModularTenantRouterMiddleware.Invoke(HttpContext httpContext)
   at OrchardCore.Modules.ModularTenantContainerMiddleware.<>c__DisplayClass4_0.<<Invoke>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell)
   at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell)
   at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell)
   at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell)
   at OrchardCore.Modules.ModularTenantContainerMiddleware.Invoke(HttpContext httpContext)
```

2. With "OrchardCore.Redis.Lock" feature enabled as Setup and as Tenanf feature
I got Timeout on "acquire an auto-setup lock", looks like as planned, only one instance could acquire a lock.

```
Connection id "0HM95HL5INF1L", Request id "0HM95HL5INF1L:00000001": An unhandled exception was thrown by the application.
System.TimeoutException: Fails to acquire an auto setup lock for the tenant: Default
   at OrchardCore.AutoSetup.AutoSetupMiddleware.InvokeAsync(HttpContext httpContext)
   at Microsoft.AspNetCore.Localization.RequestLocalizationMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
   at OrchardCore.Modules.ModularTenantRouterMiddleware.Invoke(HttpContext httpContext)
   at OrchardCore.Modules.ModularTenantContainerMiddleware.<>c__DisplayClass4_0.<<Invoke>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell)
   at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell)
   at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell)
   at OrchardCore.Environment.Shell.Scope.ShellScope.UsingAsync(Func`2 execute, Boolean activateShell)
   at OrchardCore.Modules.ModularTenantContainerMiddleware.Invoke(HttpContext httpContext)
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.ProcessRequests[TContext](IHttpApplication`1 application)
```

And strangely enough, I got another DPAPI error 

```
Microsoft.AspNetCore.DataProtection.Repositories.FileSystemXmlRepository[60]
      Storing keys in a directory '/root/.aspnet/DataProtection-Keys' that may not be persisted outside of the container. Protected data will be unavailable when container is destroyed.
```
even if I enabled "OrchardCore.Redis.DataProtection" as Setup and Tenant feature

jtkech:
- First i assume that you used only one set of OC instances, if not we would need to use our new Redis `InstancePrefix`.

- Here the goal is to not fail on timeout by having both timeout = expiration higher than setup time, so that a given instance has the time to setup a given tenant and then release the lock before the timeout / expiration, so that another instance can acquire the lock (no timeout) but will see that the tenant is already setup **and then reload it**.

    Hmm, maybe the lock is not released when a given instance has setup a given tenant, when the lock is disposed it uses the lock service that created it, but this lock service instance is no more available after a setup that built a new shell.

    If i have time i will try it this week end.

    In the meantime, because the lock is auto released after the expiration, what you could try is to still use an expiration higher than the setup time, **but with a timeout higher than the expiration**, so that after a setup (done by a given instance) the lock expires before any timeout, letting other instances (one after the other) to just do a tenant reload.

- About the DP error we will see, as said during the setup the first shell activation will not use the redis DP because the recipe features step is not yet executed, so we first use the file system DP, so maybe normal to have this "error" saying that you should persist keys outside the container. Note: Normally we store the keys under the app_data folder, and we have a config option to change the location of app_data.

rserj:
I used env. variable `OrchardCore__OrchardCore_Redis__InstancePrefix` to set the prefix, I'll try to increase timeouts and see the result

rserj:
I've increased timeout, no errors!

jtkech:
> I've increased timeout, no errors!

Cool 👍 so we are close to a good solution.

But, even in this auto setup scenario, it would be better if the lock was released by the owner before it expires.

Hmm, maybe something related to the new Redis `InstancePrefix`, maybe good to try it with only one set of OC instances sharing the same database, so that we can try it with a null or empty `InstancePrefix`.

I will look at it this week end.

jtkech:
So, first with only one OC instance and by using the default value of 30s for the auto setup lock timeout and expiration, i could check that after an auto setup the lock is released before the expiration and timeout.

Then i tried the same but with 2 OC instances, not a fully stateless config but by using the following.

            services
                .AddOrchardCms()
                .AddDatabaseShellsConfiguration()
                .AddSetupFeatures(
                    "OrchardCore.AutoSetup",
                    "OrchardCore.Redis.Lock"
                    );

It worked well, only one instance did the setup and released the lock before it expires, the other instance was waiting and, when the first completed the setup, was able to acquire the lock before any timeout and, seeing that the tenant is already setup, was able to reload it (here from the database).

So maybe the timeout and expiration that you first used was too close to the setup time, don't hesitate to use an higher value, e.g. the default of 30s or more if needed, normally it will not be slower because as soon as an instance will finish a setup (e.g. in 5s), it will release the lock and then make it available to another instance and so on.

rserj:
@jtkech thanks for testing it out on your weekends. 
I have a big recipe with lots of Features, 30seconds  wasn't enough. I've increased timeout by setting env variables for 1min
```
OrchardCore__OrchardCore_AutoSetup__LockOptions__LockTimeout: 60000
OrchardCore__OrchardCore_AutoSetup__LockOptions__LockExpiration: 60000
```
it did the trick, no errors in logs. But even with 30_000 when I got a TimeoutException, the Shell wasn't broken and the instance got alive once Setup has been finished by other instance.

rserj:
is it ok to merge this PR?

jtkech:
@rserj

Okay cool, i will approve it, i was just waiting for you to confirm ;)

Hmm, so would it be better to use 60 seconds by default?

rserj:
>use 60 seconds by default?
I think it depends on the scenario, for our Recipes: Blog, Agency even if we copy files during installation to Azure Blob/S3 it should be enough. We should also take into account LocalLock, maybe 60sec for a LocalLock is too much.

BTW, I didn't try running Migration simultaneously as I see ShellActivation lock should trigger? or this is something that can be handled by YesSql already?

jtkech:
Yes, i also thought about updating the shell activation default lock values, maybe in a separate PR if it turns out to be too low too, and knowing that it doesn't interfere with this PR where we lock at another level.

No problem if it is a local lock, here (as for shell activation) we don't use any expiration / timeout, meaning that it never expires / timeout, it just works as before when using a simple semaphore without specifying any max time to wait.

Yes, good idea to adjust the times based on the migrations, but it's not a problem to use higher values, as soon as a process has done its job, it is intended to release its lock. So that's okay to just have them configurable, the goal being to find good default values, not too high but also not too low, so that most of the time we don't have to override them.

So based on your experience i will update these default values to 60s for the auto setup, and then approve this PR.

jtkech (approved):

rserj:
@sebastienros @jtkech if there are no objections can we merge this PR?

jtkech:
@rserj That's okay for me, at least let's wait for the thirsday triage meeting.

PR #9498, opened 2021-05-20
Table in darkmode

jtkech:
So that `table-striped` and `table-hover` work even if in dark mode


src/OrchardCore.Themes/TheAdmin/Assets/scss/modules/theme/darkmode/_table.scss
```diff
@@ -0,0 +1,3 @@
+.table {
+    @extend .table-dark;
```
Skrypt, line 2: This should be added inside bootswatch.scss

https://github.com/OrchardCMS/OrchardCore/blob/e49c72d59990efd816699e886ebc90f032ca66a1/src/OrchardCore.Themes/TheAdmin/Assets/scss/modules/theme/darkmode/_bootswatch.scss#L61-L207

Not sure why it is needed though since we have a bunch of classes already there.

Skrypt: I'm thinking about this and I'm not sure this is the appropriate fix.

.table-dark is used for styling a table with a dark background. Here, we should always be able to decide if we want to have a table-dark or table-light even if we are in the darkmode or not. The only difference is that it should be reverted if we are in the darkmode. So table-dark would become table-light. So .table should never extend .table-dark by default if we want to keep it's usability.

jtkech: > This should be added inside bootswatch.scss

Ahrrr! This is what i first did ;) but thought it was better to not change bootswatch.scss as possible

> I'm thinking about this and I'm not sure this is the appropriate fix.

Here it is a singularity, currently a .table (not a .table-dark) is rendered "darkly" in dark mode but some color contrasts as hover and striped would need to be tuned, so i thought that the easier and better way to do it is to use the Bootstrap contrast tuning. So, if we explicitly specify a table-dark it will be still a dark-table in both modes as before, if we use a simple table, in light mode it is still as before, the only change is when a .table is rendered under the dark mode, it will use the Bootstrap color contrast tuning in place of the styles we may have missed.

But no problem if you think it is better to redefine all the variables a table use so that it becomes a full dark table under our dark mode, but what is better if someone explicitly specify a table-dark, should our .table (that we made darkly with our styles) look like a real .table-dark? If so maybe better to just extend .table to table-dark.

Skrypt: I think that when we switch from lightmode to darkmode the styles won't change from class="table table-dark" to "table table-light" for example. So, when we are in lightmode the table-dark allows to have a highly contrasted black table background. So, if we are in darkmode I think it should simply do the same. Which means table-dark in darkmode would become table-light and vice versa. Else applying directly table-dark to .table will remove the ability to theme those table appropriately I think.

jtkech: Okay, no problem

Hmm, what we could do is to let as is, so that any specific table as .table-dark or .table_light works as before, and just introduce a new class, e.g. `.table-darkly` that we could extend to .table-dark but only if under our dark mode.

What you think? If you agree let me know if you have a better name for this new class

> The table.scss can also be imported inside the bootswatch.scss file

As you want, let me know what you think is better

jtkech:
@Skrypt 

Okay, i did it too quicly, and read your comment too quickly ;)

The only little concern i had is about the contrast of the striped and hover colors in dark mode  when a .table don't specify any specific .table-something.

I'll try to suggest something else this night or just close this PR.

jtkech:
@Skrypt 

For infos in bootstrap 5.0 i saw these variables

    $table-striped-bg-factor: .05;
    $table-striped-bg: rgba($white, $table-striped-bg-factor) !default;

    $table-hover-bg-factor: .075;
    $table-hover-bg: rgba($white, $table-hover-bg-factor) !default;

I could use `$table-hover-bg` that is used in 4.0, but `$table-striped-bg` is not yet used, in 4.0 this is the `$table-accent-bg` that is used for the striped color of a non specific .table

---

So here i just copy pasted the 6 related table variables from the original boostrap _variables, and for our dark mode just replaced $black to $white.

It doesn't affect specific tables as .table-success, in fact it fixes the striped color of these specific tables, before in dark mode all even lines were black.

And a non specific .table in dark mode is well darkened and with visible striped and hover colors.

Skrypt (approved):

PR #9507, opened 2021-05-20
Workflow Editor Quality of life improvements

jptissot:
Small quality of life improvements for the workflow editor.

1. Limit the width of each activity box to 280px and add an ellipsis if the text is longer than the box allows.
Before: 
image

After: 
image

2. If an activity has a negative X or Y coordinate, set the corresponding coordinate to 0 so the Activity is not rendered off-screen. 

I am not sure when or how this happens but I saw some workflows where the X coordinate is outside of the canvas, making it impossible to move / edit an activity without hacking the activity coordinates using the browser's devtools.


src/OrchardCore.Modules/OrchardCore.Workflows/Views/Activity.Design.cshtml
```diff
@@ -10,8 +10,10 @@
     var hasEditor = activity.HasEditor;
     var activityType = isEvent ? "Event" : "Task";
     var activityTypeClass = isEvent ? "fa-bolt" : "fa-caret-right";
+    var xPosition = activityRecord.X <= 0 ? 0 : activityRecord.X;
```
jptissot, line 13: Not sure if this is the right place to do this. We could potentially add a check when saving the activityRecord or on the getter / setter of the X, Y properties on the record.

src/OrchardCore.Modules/OrchardCore.Workflows/Assets/Styles/workflow-editor.scss
```diff
@@ -14,8 +14,9 @@
 .workflow-canvas {
 
     .activity {
+        max-width:280px;
         cursor: move;
-
+
```
sfmskywalker, line 19: Excessive white space added by accident.

src/OrchardCore.Modules/OrchardCore.Workflows/Views/Activity.Design.cshtml
sfmskywalker: I would do it here as well, since it seems to be the safest place. If you would do it when saving, and if for some reason the X Y coordinates gets changed to negative values due to some bug, we'd still have the issue you are fixing.

sfmskywalker (approved):

src/OrchardCore.Modules/OrchardCore.Workflows/Assets/Styles/workflow-editor.scss
jptissot: fixed

Skrypt:
I think there was a previous issue/PR that never got merged. Need to find them and close them if not.

jptissot:
Searched for Workflow and did not find anything similar to this issue.

PR #9509, opened 2021-05-21
Fixes #9269 release view buffer at the end of scope

jtkech:
Fixes #9269

PR #9518, opened 2021-05-21
Fixes #9512 Cancel session on exception

### LINKED ISSUE #9512: Documents are still persisted after an unhandled exception is thrown
### To Reproduce

- enlist an entity for save with `_session.Save(entity)`
- throw an unhandled exception
- entity is still commited by the `RegisterBeforeDispose` in the shell scope

### Expected behavior

We've resolved this ourselves with some exception handling middleware, which specifically calls `await _session.CancelAsync()`

Should we put this, or a similar solution, into the shellscope, so that `RegisterBeforeDispose` can potentially have an option to execute operations, like `SaveChangesAsync()` only if there are no exceptions.

/cc @jtkech have we talked about this before, I thought we had?

jtkech:
Fixes #9512 

@deanmarcussen if you have time, can you check if it fixes your issue?

### DIFF
--- a/src/OrchardCore/OrchardCore.Abstractions/Shell/Scope/ShellScope.cs
+++ b/src/OrchardCore/OrchardCore.Abstractions/Shell/Scope/ShellScope.cs
@@ -231,25 +231,24 @@ namespace OrchardCore.Environment.Shell.Scope
 
             using (this)
             {
-                StartAsyncFlow();
-
-                if (activateShell)
-                {
-                    await ActivateShellInternalAsync();
-                }
-
                 try
                 {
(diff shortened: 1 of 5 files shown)


deanmarcussen:
@jtkech looks like it will be ideal - particularly with the `IExceptionHandlerFeature`

I will test on Monday

jtkech:
Okay cool, i may have some last tweaks to do

jtkech:
Any feedback?

But only if you have time

Just because i need it for #9509 ;)

deanmarcussen (approved):
@jtkech tested, fixes the issue thank you

PR #9523, opened 2021-05-24
Code Optimization for ReCaptchaService

Skrypt:
Adds multipart/form-data Content Type.

### DIFF
--- a/src/OrchardCore/OrchardCore.ReCaptcha.Core/Services/ReCaptchaService.cs
+++ b/src/OrchardCore/OrchardCore.ReCaptcha.Core/Services/ReCaptchaService.cs
@@ -80,12 +80,10 @@ namespace OrchardCore.ReCaptcha.Services
                 return false;
             }
 
-            // If the header Content-Type of the request is "application/json" we get
-            // the ReCaptcha token from the header as this token should not be persisted in the Workflow metadata
+            // We use the header value as default if it's passed
             var reCaptchaResponse = _httpContextAccessor.HttpContext.Request.Headers[Constants.ReCaptchaServerResponseHeaderName];
 
-            // If this is a standard form post we get the token from the form values
-            // only if the header var value is not affected.
+            // If this is a standard form post we get the token from the form values if not affected previously in the header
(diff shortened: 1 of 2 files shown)


src/OrchardCore/OrchardCore.ReCaptcha.Core/Services/ReCaptchaService.cs
```diff
@@ -84,9 +84,9 @@ public async Task<bool> ValidateCaptchaAsync(Action<string, string> reportError)
             var reCaptchaResponse = _httpContextAccessor.HttpContext.Request.Headers[Constants.ReCaptchaServerResponseHeaderName];
 
             // If this is a standard form post we get the token from the form values if not affected previously in the header
-            if(String.IsNullOrEmpty(reCaptchaResponse) && String.Equals(_httpContextAccessor.HttpContext.Request.Headers["Content-Type"], "application/x-www-form-urlencoded", StringComparison.OrdinalIgnoreCase))
+            if(String.IsNullOrEmpty(reCaptchaResponse) && (_httpContextAccessor.HttpContext?.Request?.HasFormContentType ?? false))
```
deanmarcussen, line 87: I have a quick question re this.

Why accessing the request directly? Depending on where and how this is used, just injecting the `IUpdateModelAccessor` and model binding it may (or may not) actually be a one trick binding to get it from header/form/ json body etc.

Might work assuming recaptcha is always coming through a controller. Just for ideas ;)

you would have to try binding it to a class I suspect. like
```
public class RecaptchaBinding
{ 
   public string ResponseHeaderName { get; set; } (or form name)
}
var viewModel = new RecaptchaBinding();
await updater.TryUpdateModelAsync(viewModel, String.Empty);  --> i.e. no prefix
```

Skrypt: I don't want to assume it's coming from a standard controller that request a common application/form. We could also use a reCaptcha on a request that sends a application/text/plain or application/somefiletype data.

jtkech (approved):

Issue #8619, opened 2021-02-17
Tabs do not work without Contents module enabled


deanmarcussen:
### Describe the bug

It's common to use Tabs to arrange custom properties on the Users View.

The `TabContainer` and related views are in the `Contents` module.

When the `Contents` module is disabled, these views are not discovered, and users settings cannot make use of tabs.

I am currently working on a project which is more about Users, and custom entities, than Content Items, and hence have the `Contents` module disabled, because I do not want to expose unwanted API endpoints like `api/content`, as they have no meaning for this particular project.

### Expected behavior

We need to find a module that can support these views, that has no dependencies.

Or consider placing them in the `DisplayManagement` module. 

That's tricky because the `DisplayManagement` module does not currently provide Views. 

It does have the `ShapeAttributes`, so one alternative could be to move these to shape attributes.

jtkech:
Yes, a separate module with no deps and that `OC.Contents` would depend on, or `ShapeAttributes` in `OC.DisplayManagement`, and maybe also for the `CardContainer.cshtml` and `Card.cshtml` and all future ones.

Hmm, maybe better to use a module, if easier to add future views.

deanmarcussen:
I think maybe I will move them to `ShapeAttributes`

Because `DisplayManagement` would technically fail without them.

Originally I only did them as views, as easier to see what the code is, to override them. But really they are all very simple shapes, mostly built with `TagBuilder` anyway, so still easy as a `ShapeAttribute`

I will do a pr soon

PR #9551, opened 2021-05-26
Move card / tab shapes to Shape Attributes

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/8619

/cc @jtkech 

I need to do a little more testing first :)

Need this for audit trail as well, as using it for Users, on the SaaS recipe creates the same issue.


jtkech:
LGTM will take a deeper look this night

jtkech:
Just tried the Tabs, i let you check Columns and  Cards

I committed some minors changes but feel free to revert them if you don't like them ;)

- Remove unused using.

- Add a null check, `shape.GetProperty<>` may return null if the type is not a value type e.g. a class / interface (and if we don't specify a default value).

- Here also i prefer to use a first lower case for method parameters.

- Set first to false once, only if it is true.

Note: I'm now used to not always add a line before an if block, e.g. after initializing a related variable, good compromise between a too compacted code and a too diluted one, but i still always add a new line after an if block. But open to any other rule, even if we say that here everybody can do as they want.

jtkech (commented):
Waiting for testing cards / columns before approving

deanmarcussen:
> Waiting for testing cards / columns before approving

All tested, works fine.

> use a first lower case for method parameters

Did not know we had gone case invariant on attribute method name injection. All good with me.

All changes appreciated :)

jtkech (approved):

PR #9554, opened 2021-05-26
Set the cookie in the ContentRequestCultureProvider

### LINKED ISSUE #5065: ContentRequestCultureProvider issue
I have seen cases where the cookie and Content Culture were off sync (cookie set to english but current page is french because of ContentRequestCultureProvider). This causes weird issues with the ContentCulturePicker and non content links.

For example: I am on a french page, cookie is set to english. When I click on a login link, the text will be in english.

To fix this, I had to add this line of javascript that will set the cookie based on the html lang attribute: 
```
document.cookie = '.AspNetCore.Culture=c%3D'+document.documentElement.lang+'%7Cuic%3D'+document.documentElement.lang +';path=/';
```

Any ideas on a better fix ?

jptissot:
fixes #5065

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentLocalization/Services/ICulturePickerService.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentLocalization/Services/ICulturePickerService.cs
@@ -25,6 +25,6 @@ namespace OrchardCore.ContentLocalization.Services
         /// <summary>
         /// Sets the culture cookie if the setting is enabled
         /// </summary>
-        Task SetContentCulturePickerCookie(string targetCulture);
+        void SetContentCulturePickerCookie(string targetCulture);
     }
 }
--- a/src/OrchardCore.Modules/OrchardCore.ContentLocalization/Services/ICulturePickerService.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentLocalization/Services/ICulturePickerService.cs
@@ -22,5 +22,9 @@ namespace OrchardCore.ContentLocalization.Services
         /// </summary>
         /// <returns>List of Localization for the url or null if not found</returns>
         Task<IEnumerable<LocalizationEntry>> GetLocalizationsFromRouteAsync(PathString url);
+        /// <summary>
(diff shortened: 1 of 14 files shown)


jptissot:
@sebastienros I added a separate setting as suggested.

I also modified the admin menu structure slightly to group the localization settings together. Let me know if this change makes sense.

image

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Services/ICulturePickerService.cs
```diff
@@ -22,5 +22,9 @@ public interface IContentCulturePickerService
         /// </summary>
         /// <returns>List of Localization for the url or null if not found</returns>
         Task<IEnumerable<LocalizationEntry>> GetLocalizationsFromRouteAsync(PathString url);
+        /// <summary>
+        /// Sets the culture cookie if the setting is enabled
```
jptissot, line 26:
```suggestion
        /// Sets the culture cookie to the target culture
```

jptissot:
Anything else needed for this pr ?

src/OrchardCore.Modules/OrchardCore.ContentLocalization/Startup.cs
```diff
@@ -80,15 +75,21 @@ public override void Configure(IApplicationBuilder app, IEndpointRouteBuilder ro
...
     {
+        private readonly IShellConfiguration _shellConfiguration;
+        public ContentPickerStartup(IShellConfiguration shellConfiguration)
+        {
+            _shellConfiguration = shellConfiguration;
+        }
+
         public override void ConfigureServices(IServiceCollection services)
         {
+            services.AddLiquidFilter<SwitchCultureUrlFilter>("switch_culture_url");
```
deanmarcussen, line 86: think you added this one to this startup (which makes way more sense), and forgot to remove it from the above startup?

jptissot: Yes, thank you. Fixed

sebastienros (approved):

PR #9557, opened 2021-05-27
Workflow : add deserializeRequestData scripting method

### LINKED ISSUE #4332: Workflows and Forms : Accessing Form Data for submitting POST HTTP request
Created a ContactUs form with the help of the [link](https://orchardcore.readthedocs.io/en/dev/OrchardCore.Modules/OrchardCore.Forms)

I want to submit the form and store the values in the SQL Server Database so I created Web API with Post method and I tried to add HTTP Request Task in the workflow but I am not able to pass any form value through Body but if I pass simple static string then the record is getting inserted.

image

I want to update the highlighted line with the liquid query so that I can pass the form data to my Web API.

My Workflow:
image

Could someone tell me whether it is possible in beta version or I have to wait for the final version? Also is there any document or link which guide me step by step to submit form data to the database.

### LINKED ISSUE #9491: Jint does not seem to handle Json.NET linq objects as one would expect. Affecting the workflow scripting activity.
### Describe the bug

The Orchard docs state both `queryStringAsJson` and `requestFormAsJson` should return the respective item as a JSON object. Looking at the code, each return a `JObject` type in the `GlobalMethod` delegate. However Jint does not appear to be correctly translating this object into a javasrcript object that can be worked with inside the scripting engine. 

I attempted to inspect what the object was by logging the stringified version, aka

```
var form = requestFormAsJson();
log("Debug", JSON.stringify(form));
```
Which resulted in this output:
```
{}
```

I then tried this:
```
var form = requestFormAsJson();
var formLength = Object.keys(form).length;
log("Debug", "Form Length: " + formLength);
```
Which resulted in this output:
```
Form Length: 0
```

I am using Orchard RC2, so I'm not sure if this is resolved by a newer beta release of Jint...

Originally I found this by adding my own custom global method which also returned a `JToken` object, and I found it was not being handled correctly. For the time being i'm getting around it by retrieving the stringified version then doing a `JSON.parse()` call inside the engine, and this has worked as intended.

Skrypt:
Fixes #9491, Fixes #4332

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Workflows/Http/Scripting/HttpMethodsProvider.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Workflows/Http/Scripting/HttpMethodsProvider.cs
@@ -8,6 +8,7 @@ using Microsoft.AspNetCore.Mvc.Abstractions;
 using Microsoft.AspNetCore.Mvc.Routing;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Net.Http.Headers;
 using Newtonsoft.Json.Linq;
 using OrchardCore.Mvc.Core.Utilities;
 using OrchardCore.Scripting;
@@ -158,7 +159,7 @@ namespace OrchardCore.Workflows.Http.Scripting
                             result = new JObject((from field in httpContextAccessor.HttpContext.Request.Form
                                     select new JProperty(field.Key, JArray.FromObject(field.Value.ToArray()))).ToArray());
                         }
-                        else if (httpContextAccessor.HttpContext.Request.HasJsonContentType())
+                        else if (HasJsonContentType(httpContextAccessor.HttpContext.Request))
                         {
                             string json;
                             using (var sr = new StreamReader(httpContextAccessor.HttpContext.Request.Body))
@@ -179,6 +180,41 @@ namespace OrchardCore.Workflows.Http.Scripting
         public IEnumerable<GlobalMethod> GetMethods()
         {
             return new[] { _httpContextMethod, _queryStringMethod, _responseWriteMethod, _absoluteUrlMethod, _readBodyMethod, _requestFormMethod, _queryStringAsJsonMethod, _requestFormAsJsonMethod, _requestDataAsJsonObjectMethod };
+        }
+
+        /// <summary>
+        /// Checks the Content-Type header for JSON types.
+        /// This method needs to be removed after we drop support for netcoreapp3.1
+        /// It is now part of net5.0 see :
+        /// https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.httprequestjsonextensions.hasjsoncontenttype?view=aspnetcore-5.0
+        /// </summary>
+        /// <param name="request"></param>
+        /// <returns></returns>
+        private static bool HasJsonContentType(HttpRequest request)
+        {
+            if (request == null)
+            {
+                throw new ArgumentNullException(nameof(request));
+            }
+
+            if (!MediaTypeHeaderValue.TryParse(request.ContentType, out var mt))
+            {
+                return false;
+            }
+
+            // Matches application/json
+            if (mt.MediaType.Equals("application/json", StringComparison.OrdinalIgnoreCase))
+            {
+                return true;
+            }
+
+            // Matches +json, e.g. application/ld+json
+            if (mt.Suffix.Equals("json", StringComparison.OrdinalIgnoreCase))
+            {
+                return true;
+            }
+
+            return false;
         }
     }
 }
--- a/src/OrchardCore.Modules/OrchardCore.Workflows/Http/Scripting/HttpMethodsProvider.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Workflows/Http/Scripting/HttpMethodsProvider.cs
@@ -156,8 +156,15 @@ namespace OrchardCore.Workflows.Http.Scripting
                     {
                         if (httpContextAccessor.HttpContext.Request.HasFormContentType)
                         {
-                            result = new JObject((from field in httpContextAccessor.HttpContext.Request.Form
-                                    select new JProperty(field.Key, JArray.FromObject(field.Value.ToArray()))).ToArray());
+                            try
+                            {
+                                result = new JObject((from field in httpContextAccessor.HttpContext.Request.Form
+                                        select new JProperty(field.Key, JArray.FromObject(field.Value.ToArray()))).ToArray());
+                            }
+                            catch
+                            {
+                                throw new Exception("Invalid form data passed in the request.");
+                            }
                         }
                         else if (HasJsonContentType(httpContextAccessor.HttpContext.Request))
                         {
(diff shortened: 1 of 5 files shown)


src/OrchardCore.Modules/OrchardCore.Workflows/Http/Scripting/HttpMethodsProvider.cs
```diff
@@ -134,18 +136,99 @@ public HttpMethodsProvider(IHttpContextAccessor httpContextAccessor)
...
+                    JObject result = null;
+
+                    if(httpContextAccessor.HttpContext != null)
+                    {
+                        if (httpContextAccessor.HttpContext.Request.HasFormContentType)
+                        {
+                            try
+                            {
+                                result = new JObject((from field in httpContextAccessor.HttpContext.Request.Form
+                                        select new JProperty(field.Key, JArray.FromObject(field.Value.ToArray()))).ToArray());
```
jptissot, line 162: I would suggest to differentiate between arrays and simple properties, the resulting object is much cleaner.

Skrypt: Yes, I need to fix this part here still. Felt asleep. I think, since requests can send multiple values for a single form element that it makes sense that these are represented as an array. Though, when the request is sent as JSON using a single property is allowed. So this method needs more validation on the actual JSON passed so that the resulting object be always using the same "schema".

Skrypt: Ok, I see what you mean now that I look at this : 
https://github.com/StatCan/StatCan.OrchardCore/blob/ae8543049fbe3e1e93864ae6dcb6ac60e8668584/src/Modules/StatCan.OrchardCore.Scripting/FormsGlobalMethodsProvider.cs#L53-L58

Sanitizing might also be a good thing.

jptissot: Yeah, I think sanitizing should be a option on the globalMethod and be true by default, but can be disabled per use case.

sebastienros: I don't understand why sanitization is necessary here. Or make it available in the script if the user wants to invoke it.

src/OrchardCore.Modules/OrchardCore.Workflows/Http/Scripting/HttpMethodsProvider.cs
```diff
@@ -134,18 +137,109 @@ public HttpMethodsProvider(IHttpContextAccessor httpContextAccessor)
...
             {
                 Name = "requestFormAsJson",
                 Method = serviceProvider => (Func<JObject>)(() =>
                     new JObject((from field in httpContextAccessor.HttpContext.Request.Form
                                  select new JProperty(field.Key, JArray.FromObject(field.Value.ToArray()))).ToArray()))
             };
+
+            _requestDataAsJsonObjectMethod = new GlobalMethod
+            {
+                Name = "requestDataAsJsonObject",
```
sebastienros, line 151: There is no such thing of a JSON object. It's either json or an object. In this case it's a `JObject` (NewtonSoft) but I don't see why it should be it, and not just a `Dictionary<string, object>`. Or does it matter because it's passed back to C# and it's better than a Dictionary?

jptissot: in my case, it was not passed back to C#. It was used only in scripts and liquid. Dictionary would work.

sebastienros: would probably even make more sense to call it `deserializeRequestData`/`deserializeRequestForm`

src/OrchardCore.Modules/OrchardCore.Workflows/Http/Scripting/HttpMethodsProvider.cs
Skrypt: You could pass `<script>something<script>` as a form value to inject a script. That's the reason why I think it should be sanitized. Or maybe we should not do it here and somewhere else. Making it optional and default to true makes sense if we want this. Else I can remove, let me know.

src/OrchardCore.Modules/OrchardCore.Workflows/Http/Scripting/HttpMethodsProvider.cs
```diff
@@ -134,18 +137,130 @@ public HttpMethodsProvider(IHttpContextAccessor httpContextAccessor)
...
                                  select new JProperty(field.Key, JArray.FromObject(field.Value.ToArray()))).ToArray()))
             };
+
+            _deserializeRequestDataMethod = new GlobalMethod
+            {
+                Name = "deserializeRequestData",
+                Method = serviceProvider => (Func<JObject>)(() =>
+                {
+                    JObject result = null;
+                    var sanitizer = serviceProvider.GetRequiredService<IHtmlSanitizerService>();
```
sebastienros, line 155: We don't do this anywhere else in controllers, editors, ... We handle that when user data is rendered.

sebastienros: Add some validation that will prevent the form from being saved instead if it contains some script. With a custom validation step.

Skrypt: Ok removing. It adds overhead to each requests.

src/OrchardCore.Modules/OrchardCore.Workflows/Http/Scripting/HttpMethodsProvider.cs
Skrypt: Ok last thing remaining to test is `Dictionary<string, object>` instead of `JObject`

Skrypt: While it still works there are some things that don't get serialized properly. This is probably why it got changed to send a JObject directly so that it can be formatted the way we want it exactly. I'm going to try to make it work with `Dictionary<string, object>` but it seems to not work as it should.

Skrypt: application/x-www-form-urlencoded : 

image

application/json : 

image

We can see here that with application/json we can pass fields with single string values <string, string>
But with application/x-www-form-urlencoded we can't pass fields with a single string value because of IDictionary<string, object>.

I think it is fine enough since application/json seems more appropriate for sending "custom json" to the API.

Skrypt:
Done

sebastienros (approved):

sebastienros:
Not sure it's fixing https://github.com/OrchardCMS/OrchardCore/issues/9491 though. The methods in this issue have a wrong name.

Skrypt:
I'm tempted to change their names but it would surely break a lot of Workflows.

sebastienros:
Create new methods, returning Dictionary. Keep current one calling the new ones, with a comment to remove them in the next breaking change version. Update the documentation so they are not found anymore. Add a comment. Different PR.

PR #9569, opened 2021-05-28
Make PortableObjectStringLocalizer simpler for inheritance

### LINKED ISSUE #9542: Modify PortableObjectStringLocalizer to enable correct inheritance
PortableObjectStringLocalizer has non-virtual methods this[string] for getting resources. If you inherit from PortableObjectStringLocalizer, you cannot call the required method using the IStringLocalizer interface.

Need to add a modifier virtual for methods this[string] in PortableObjectStringLocalizer class. Аlso need to change the modifier of GetTranslation methods from private to protected.

Now we have to cast IStringLocalizer to the appropriate type to call the correct method.

hishamco:
Fixes #9542

### DIFF
--- a/src/OrchardCore/OrchardCore.Localization.Core/PortableObject/PortableObjectStringLocalizer.cs
+++ b/src/OrchardCore/OrchardCore.Localization.Core/PortableObject/PortableObjectStringLocalizer.cs
@@ -37,7 +37,7 @@ namespace OrchardCore.Localization.PortableObject
         }
 
         /// <inheritdocs />
-        public LocalizedString this[string name]
+        public virtual LocalizedString this[string name]
         {
             get
             {
@@ -53,7 +53,7 @@ namespace OrchardCore.Localization.PortableObject
         }
 
         /// <inheritdocs />
-        public LocalizedString this[string name, params object[] arguments]
+        public virtual LocalizedString this[string name, params object[] arguments]
         {
             get
             {
@@ -65,7 +65,7 @@ namespace OrchardCore.Localization.PortableObject
         }
 
         /// <inheritdocs />
-        public IEnumerable<LocalizedString> GetAllStrings(bool includeParentCultures)
+        public virtual IEnumerable<LocalizedString> GetAllStrings(bool includeParentCultures)
         {
             var culture = CultureInfo.CurrentUICulture;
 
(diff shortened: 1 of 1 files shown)


Skrypt:
I'm not sure if this is necessary. Why would you want to inherit the PortableObjectStringLocalizer?

hishamco:
There are some case to allow developers to subclass the localizer. I remember one example is to allow multiple path to load the localization resources from. Make the members open some extensibility

@amolodenov you can mention your use case if it's possible

amolodenov:
> There are some case to allow developers to subclass the localizer. I remember one example is to allow multiple path to load the localization resources from. Make the members open some extensibility
> 
> @amolodenov you can mention your use case if it's possible

I need to use few cultures at the same time. For this, I planned to implement an analogue https://github.com/aspnet/Localization/blob/master/src/Microsoft.Extensions.Localization/ResourceManagerWithCultureStringLocalizer.cs, but PortableObjectStringLocalizer did not allow inherit correctly. I know ResourceManagerWithCultureStringLocalizer is obsolete, but changing CurrentCulture doesn't suit for me.
My example
```
public class PortableObjectWithCultureStringLocalizer : PortableObjectStringLocalizer
{
    private readonly ILocalizationManager _localizationManager;
    private readonly bool _fallBackToParentCulture;
    private readonly string _context;
    private readonly CultureInfo _culture;

    public PortableObjectWithCultureStringLocalizer(
        string context,
        ILocalizationManager localizationManager,
        bool fallBackToParentCulture,
        CultureInfo culture,
        ILogger logger)
        : base(context, localizationManager, fallBackToParentCulture, logger)
    {
        _context = context;
        _localizationManager = localizationManager;
        _fallBackToParentCulture = fallBackToParentCulture;
        _culture = culture;
        _logger = logger;
    }

    public override LocalizedString this[string name]
    {
        get
        {
            if (name == null)
            {
                throw new ArgumentNullException(nameof(name));
            }

            var translation = GetTranslation(name, _context, _culture, null);

            return new LocalizedString(name, translation ?? name, translation == null);
        }
    }

    public override LocalizedString this[string name, params object[] arguments]
    {
        get
        {
            var (translation, argumentsWithCount) = GetTranslation(name, arguments);
            var formatted = string.Format(translation.Value, argumentsWithCount);

            return new LocalizedString(name, formatted, translation.ResourceNotFound);
        }
    }
   
   ...
}
```

hishamco:
Any update on this, I think the idea to make the localizer more extensible nothing else

Skrypt:
//cc @sebastienros

hishamco:
Can we triage this?

sebastienros:
Merging, but we should not make it a habit. It's exposing things we might regret as it's increasing the burden on breaking changes for future versions. One can easily copy the code if necessary. There isn't much value in inheriting directly from it.

hishamco:
Sometimes there's a need to customize the default behavior, Copy paste is not good all the time ;)

PR #9581, opened 2021-05-29
Liquid intellisense for monaco

MichaelPetrinolis:
This is a draft, just to start a discussion of how we should handle it.

~It contains a custom build of https://github.com/microsoft/monaco-html/pull/13 PR that supports HTML5 for liquid.~

Should we replace code mirror in templates editor, or we should provide a configurable option?
How can I get a list of registered tags and filters that are registered in Orchard Core to provide suggestions?
Template preview does not render (and this happens also in original code mirror .cshtml code)
Could we add the model for suggestions of variables in object tags? {{}}

and some screenshots of tags,
image
filters
image
and html tags autocompletion
image


deanmarcussen:
Cool, interesting :)

Is the Templates Preview feature broken at the moment? It was working last time I used it.

Registered Tags and Filters are available in the `IOptions<TemplateOptions>` - that would include all the standard tags from Fluid as well.

There isn't currently any metadata available regarding arguments

MichaelPetrinolis:
@deanmarcussen I added the filters, but how can i get the tags? I need to have both comment and endcomment tags to also suggest the end tags. 

Here is a screenshot with the filters.

image

src/OrchardCore.Modules/OrchardCore.Liquid/ScriptsMiddleware.cs
```diff
@@ -0,0 +1,49 @@
...
+        private readonly RequestDelegate _next;
+
+        public ScriptsMiddleware(RequestDelegate next)
+        {
+            _next = next;
+        }
+
+        public async Task Invoke(HttpContext httpContext)
+        {
+            if (httpContext.Request.Path.StartsWithSegments("/OrchardCore.Liquid/Scripts", StringComparison.OrdinalIgnoreCase))
```
sebastienros, line 28: This should be cached per tenant, it doesn't change one the tenant is started.

src/OrchardCore.Modules/OrchardCore.Liquid/ScriptsMiddleware.cs
```diff
@@ -0,0 +1,49 @@
...
+            {
+                var script = default(string);
+                if (Path.GetFileName(httpContext.Request.Path.Value) == "liquid-intellisense.js")
+                {
+                    var templateOptions = httpContext.RequestServices.GetRequiredService<IOptions<TemplateOptions>>();
+
+                    var filters = string.Join(',', templateOptions.Value.Filters.Select(x => $"'{x.Key}'"));
+                    
+                    script = $@"[{filters}].forEach(value=>{{if(!liquidFilters.includes(value)){{ liquidFilters.push(value);}}}});";
```
sebastienros, line 37: Would there be a way to add the dynamic tags too? Or is that you didn't find a way to list them? I might be able to help then.

Would awesome also if we could list the arguments, or a description for the filters/tags, if we can add some metadata when they are registered (would need to change fluid for that, or an other options in orchard).

MichaelPetrinolis: I couldn't find a way to list the tags, that's why i added only the filters.
There are two fields in suggestions, detail(string) and documentation([monaco.IMarkdownString](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.imarkdownstring.html)) which we could use.
The output then would be like 
image
collapsed, and expanded
image

I don't see any obstacle to add the arguments as suggestions (you mean ex. the User object in the above image? ) if we have a list of them.

sebastienros: Try to resolve `LiquidViewParser`, then access the `RegisteredTags` that contains custom tags. If adding documentation requires more work then we can try to handle it in a separate PR.

src/OrchardCore.Modules/OrchardCore.Templates/Views/Template/Create - Copy.cshtml
```diff
@@ -0,0 +1,143 @@
+@model TemplateViewModel
```
jptissot, line 1: Fix the file name of this file

src/OrchardCore.Modules/OrchardCore.Liquid/OrchardCore.Liquid.csproj
```diff
@@ -10,10 +10,28 @@
     <PackageTags>$(PackageTags) OrchardCoreCMS</PackageTags>
   </PropertyGroup>
 
+  <ItemGroup>
+    <EmbeddedResource Remove="wwwroot\monaco\liquid-intellisense.ts" />
+    <EmbeddedResource Remove="wwwroot\monaco\monaco.d.ts" />
+    <EmbeddedResource Remove="wwwroot\monaco\tsconfig.json" />
+  </ItemGroup>
```
jtkech, line 17: Maybe better to use an `Assets` folder that is automatically excluded from embedded resources

And an `Assets.json` to be part of the gulp pipeline that will output the right files under wwwroot

As done in other modules, and then remove these extra ItemGroup

src/OrchardCore.Modules/OrchardCore.Liquid/OrchardCore.Liquid.csproj
```diff
@@ -10,10 +10,28 @@
...
+    <None Include="wwwroot\monaco\liquid-intellisense.ts">
+      <CopyToPublishDirectory>Never</CopyToPublishDirectory>
+    </None>
+    <None Include="wwwroot\monaco\monaco.d.ts">
+      <CopyToPublishDirectory>Never</CopyToPublishDirectory>
+    </None>
+    <None Include="wwwroot\monaco\tsconfig.json">
+      <CopyToPublishDirectory>Never</CopyToPublishDirectory>
+    </None>
+  </ItemGroup>
```
jtkech, line 33: Idem as above, and then remove these extra ItemGroup

src/OrchardCore.Modules/OrchardCore.Liquid/ScriptsMiddleware.cs
MichaelPetrinolis: Thanks, that worked. Is there a way to know if a tag is a block and needs end tag? 
image
I agree to keep the documentation for another PR. We could also implement a completion provider for the shortcodes with documentation first.

src/OrchardCore.Modules/OrchardCore.Templates/Views/Template/Create - Copy.cshtml
MichaelPetrinolis: thanx, i kept it as a reference until the PR was ready

src/OrchardCore.Modules/OrchardCore.Liquid/OrchardCore.Liquid.csproj
MichaelPetrinolis: i left it this way because it was easier to just tsc --watch and debug :)

deanmarcussen:
@MichaelPetrinolis I don't think we'll be able to know that from the current tag registration.

Either we need to add some metadata into fluid to capture that, or register a tag metadata options in OC, to capture the same.

What metadata is useful / necesary? So can figure out if ok to add to fluid, or if it's too much, and needs localizing etc, so better to do seperately in OC.
- if it's a block tag
- arguments, is it just a description string?

Anything else that it could / should take?

MichaelPetrinolis:
@deanmarcussen We could leave arguments out for the moment, and just get the list of end tags. I made a PR for that

sebastienros:
I don't want to change Fluid for that, we should just create an options class that we can feed metadata into specifically for that purpose. The data model would expose what @deanmarcussen said, name, description, maybe example, like he has done for shortcodes. Each module would then be able to register their filters, tags and corresponding documentation.

src/OrchardCore.Modules/OrchardCore.Liquid/ScriptsMiddleware.cs
```diff
@@ -0,0 +1,60 @@
...
+using OrchardCore.DisplayManagement.Liquid;
+using OrchardCore.Entities;
+using Microsoft.Extensions.DependencyInjection;
+using OrchardCore.Settings;
+using Fluid;
+using Microsoft.Extensions.Caching.Memory;
+
+namespace OrchardCore.Liquid
+{
+    public class ScriptsMiddleware
```
sebastienros, line 18: Why not use a static js file for now?
If you use IMemoryCache it would be to add a sliding expiration, or instead use a local variable. But sliding expiration is better. Also the response should have public cache headers so the client doesn't request it again.

src/OrchardCore.Modules/OrchardCore.Liquid/ScriptsMiddleware.cs
```diff
@@ -28,33 +31,44 @@ public async Task Invoke(HttpContext httpContext)
...
+                    if (httpContext.Request.Headers.TryGetValue(HeaderNames.IfNoneMatch, out var v))
                     {
+                        if (v.Contains(etag))
+                        {
+                            httpContext.Response.StatusCode = StatusCodes.Status304NotModified;
+                            return;
+                        }
+                    }
 
+                    var cacheControl = "public, max-age=2592000, s-max-age=31557600";
```
sebastienros, line 45: nit: static value. Could use `$"{Timespan.FromHours().TotalSeconds}"` to make it obvious how much that is.

MichaelPetrinolis: done.

MichaelPetrinolis:
There is an issue registering liquid with html language service in monaco v0.25.0 that is fixed in monaco 0.25.1. I created a PR for that https://github.com/OrchardCMS/OrchardCore/pull/9697

sebastienros:
Ready to merge?

MichaelPetrinolis:
> Ready to merge?

No, not yet... I cannot add the 'do not merge' tag. I must change the registration of the tags to include the Metadata for block tags, description etc.

Skrypt:
@MichaelPetrinolis I suggest that we merge this PR and that we add these in a second PR. We are waiting on this PR to get merged to make the related changes in https://github.com/OrchardCMS/OrchardCore/pull/9623

I believe that at least now we will get basic liquid support and nothing prevents from adding more over time.

MichaelPetrinolis:
@Skrypt sure, we could do that. And get more suggestions and feedback from community.
The only thing I see is the preview. With latest dev merge works but has performance issues.

Skrypt:
Ok I will try the PR to see what you mean by perf issues.

jptissot:
I will update my PR once this one is merged. I got some ideas I'd like to discuss with you guys on how to implement an event system for OC editors.

Skrypt:
Ok I tested the preview and there is some issue going on when trying to render Liquid. But I don't think it is perf related. It seems to not parse Liquid correctly.

MichaelPetrinolis:
Sorry, I was talking about the template preview. In HtmlBody and HtmlField liquid is not rendered.

Skrypt:
So nevermind about the Liquid parsing issue. It's totally fine. It's my mistake.

PR #9638, opened 2021-06-08
Fix buttons with same ids

agriffard:
When you edit a Content type, `Edit templates` and `Edit placements` have the same id.


deanmarcussen (approved):
I can't imagine we're using that id. LGTM

PR #9684, opened 2021-06-14
Monaco 0.25

agriffard:
https://github.com/microsoft/monaco-editor/releases/tag/v0.25.0
https://www.npmjs.com/package/monaco-editor/v/0.25.0

/cc @MichaelPetrinolis


MichaelPetrinolis:
@agriffard just what I needed :) 

There is an issue we must investigate. When enabling monaco editor for example in Project Content Type in Agency Recipe, an error regarding monaco appears in the console. When it is enabled on more content types, only the first group is ok, in the other groups editor fails to initialize. Maybe @jptissot has fixed it in #9623. It seems it has nothing to do with monaco version.

jptissot:
> @agriffard just what I needed :)
> 
> There is an issue we must investigate. When enabling monaco editor for example in Project Content Type in Agency Recipe, an error regarding monaco appears in the console. When it is enabled on more content types, only the first group is ok, in the other groups editor fails to initialize. Maybe @jptissot has fixed it in #9623. It seems it has nothing to do with monaco version.

Yes I did fix that in the above mentionned PR.

Issue #9658, opened 2021-06-11
Invalid column name 'UserId' after applying 1.0.0-rc2-16283


markgould:
After upgrading from 13450 to 16283 I'm getting the error "Invalid column name 'UserId'"

Digging into the logs, I see there is a failed migration:

> OrchardCore.Data.Migration.DataMigrationManager|Error while running migration version 7 for 'OrchardCore.Users'. Microsoft.Data.SqlClient.SqlException (0x80131904): Invalid column name 'IsLockoutEnabled'. Invalid column name 'LockoutEndUtc'. Invalid column name 'AccessFailedCount'. Invalid column name 'IsLockoutEnabled'. Invalid column name 'LockoutEndUtc'. Invalid column name 'AccessFailedCount'.    at Microsoft.Data.SqlClient.SqlCommand.<>c.<ExecuteDbDataReaderAsync>b__188_0(Task`1 result)    at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()    at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state) --- End of stack trace from previous location where exception was thrown ---    at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread) --- End of stack trace from previous location where exception was thrown ---    at YesSql.Commands.BatchCommand.ExecuteAsync(DbConnection connection, DbTransaction transaction, ISqlDialect dialect, ILogger logger)    at YesSql.Session.FlushAsync()    at YesSql.Session.FlushAsync()    at YesSql.Services.DefaultQuery.Query`1.ListImpl()

markgould:
It looks like PR #5785  may have caused this?

https://github.com/OrchardCMS/OrchardCore/commit/f9f93bf7d2f1a0ce5681f1c6cfb3c8bc85323a6e#diff-a3affe1c5497d75ba4867d94fac5be35f17bb7484033e15f75c50331675a6f57R27

markgould:
Looked at the document table and we are currently on migration version 4. 

To me it looks like the _session.Save is the culprit because it is calling _session.Save which since the User object has those fields on it now, it's trying to update columns that don't yet exist, since they get added in migration 10.

Also, since they are added in UpdateFrom10, it looks like UpdateFrom8 would fail too since those columns won't be there yet to add to the index.

[...]

Issue #9654, opened 2021-06-10
Cannot run on MariaDB due to migration error


fuzl-llc:
I cloned the Orchard source code and tried to run it against a local MariaDB instance.  It crashes with the following error:

MySqlConnector.MySqlException: 'Specified key was too long; max key length is 3072 bytes'

The error happens at line 33 of \src\OrchardCore.Modules\OrchardCore.Users\Migrations.cs in the "Create" method of the "Migrations" class.:

SchemaBuilder.AlterIndexTable<UserIndex>(table => table
                .CreateIndex("IDX_UserIndex_DocumentId",
                    "DocumentId",
                    "UserId",
                    "NormalizedUserName",
                    "NormalizedEmail",
                    "IsEnabled",
                    "IsLockoutEnabled",
                    "LockoutEndUtc",
                    "AccessFailedCount")
            );

### To Reproduce

1. Create a local database in MariaDB called test.
2. Get the latest code.
3. Build
4. Run OrchardCore.Cms.Web
5. On setup screen, choose the blank recipe, select "MySql," and put connection string "server=localhost; database=test; uid=root; pwd=thepassword;"

### Notes

I found two issues (#2509 and #4753) that may be related but neither seem to help.

If I show tables on the test db, I see this (it seems there are some tables missing):

```
+--------------------+
| Tables_in_test     |
+--------------------+
| aliaspartindex     |
| autoroutepartindex |
| contentitemindex   |
| document           |
| identifiers        |
| userindex          |
+--------------------+
```

deanmarcussen:
Yes looks like the functional tests are failing on MySql 

https://github.com/OrchardCMS/OrchardCore/runs/2794622153

A slight refactor required of the new lockout feature required.

Related https://github.com/OrchardCMS/OrchardCore/issues/9641

fuzl-llc:
Thanks!  Any workarounds you can suggest aside from moving to an earlier snapshot of the code?

Issue #9641, opened 2021-06-08
Claims Argument Null Exception When Migrating Site


davidhayden:
Attempting to migrate a website from RC2-15222 and received the attached error after upgrading to the latest source code in the Dev branch. Occurs when I "successfully" log in as an administrator. For completeness, also experienced another issue on the frontend as well (#9640). No place to go from here with both frontend and backend out of commission. Changed the framework to .NET5 as well if that matters.

Curious if this is expected. Nothing fancy about the user. Just a plain 'ol administrator account.

![claims-error](https://user-images.githubusercontent.com/96801/121252117-a632ad80-c875-11eb-9a95-a29dc97eeaa7.png)

deanmarcussen:
@davidhayden this and your other issue look like a migration error.

See here for notes I wrote about upgrades
https://github.com/OrchardCMS/OrchardCore/issues/7206#issuecomment-796536006

Probably you need to check your log files, as another migration may have failed previously causing both the users and layers/ rules to fail during migration.

deanmarcussen:
> 'OrchardCore.Users'. Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 1: 'table UserIndex has no column named IsLockoutEnabled'.

Yes, looks like an issue with that migration. I'll check it tomorrow if noone else gets there first.

Short term solution would be to back down the nuget packages by a week I imagine?

deanmarcussen:
@jtkech I haven't had time to pr it, but will run what I think we need to do past you, in case you have a better idea.

Migration 7 is failing now, where it updates the User, because there are now new properties added in Migration 10, and of course the index provider being activated during MIgration 7 is expecting them.

I don't think we can rearrange the migration order, in a way that will be succesful for all of the various versions that may use it.

So I am think of moving guts of migration 7, i.e. the query and saves, into a deferred task, so it isn't run until all other migration steps have completed.

It's not ideal, so hoping for a better idea...

jtkech:
@deanmarcussen

Looks like the same kind of issue we had with contents creation in a migration recipe and that we fixed by using a deferred task in the content step, here it's not about contents but users, not creation but updating, but same issue related to missing migration steps before doing some operations.

[...]

PR #9687, opened 2021-06-14
Fix user lockout migrations

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/9641
Fixes https://github.com/OrchardCMS/OrchardCore/issues/9654
Fixes https://github.com/OrchardCMS/OrchardCore/issues/9658

Just needs running on MySql to see if it actually fixes it

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Users/Migrations.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Migrations.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
 using OrchardCore.Data.Migration;
@@ -138,8 +137,6 @@ namespace OrchardCore.Users
                     user.UserId = user.UserName;
                     session.Save(user);
                 }
-
-                await session.SaveChangesAsync();
             });
 
             return 6;
@@ -166,11 +163,6 @@ namespace OrchardCore.Users
                     user.UserName = user.UserName.Replace('@', '+');
                     user.NormalizedUserName = user.NormalizedUserName.Replace('@', '+');
                     session.Save(user);
-                }
-
-                if (users.Any())
-                {
-                    await session.SaveChangesAsync();
                 }
             });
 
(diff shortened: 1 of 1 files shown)


deanmarcussen:
Passed https://github.com/OrchardCMS/OrchardCore/actions/runs/935949713

src/OrchardCore.Modules/OrchardCore.Users/Migrations.cs
```diff
@@ -123,14 +126,21 @@ public int UpdateFrom4()
...
-            }
+                var session = scope.ServiceProvider.GetRequiredService<ISession>();
+                var users = await session.Query<User>().ListAsync();
+                foreach (var user in users)
+                {
+                    user.UserId = user.UserName;
+                    session.Save(user);
+                }
+
+                await session.SaveChangesAsync();
```
jtkech, line 142: Each deferred task is executed in its own scope, so normally `SaveChangesAsync()` is automatically called before disposing the scope. Can you retry it but without this line.

src/OrchardCore.Modules/OrchardCore.Users/Migrations.cs
```diff
@@ -144,15 +154,25 @@ public int UpdateFrom6()
...
+                {
+                    user.UserName = user.UserName.Replace('@', '+');
+                    user.NormalizedUserName = user.NormalizedUserName.Replace('@', '+');
+                    session.Save(user);
+                }
+
+                if (users.Any())
+                {
+                    await session.SaveChangesAsync();
+                }
```
jtkech, line 174: Idem as above

jtkech (commented):
@deanmarcussen 

LGTM, the only thing is that you don't need to explicitly call `SaveChangesAsync()`.

I let you retry it without these calls.

src/OrchardCore.Modules/OrchardCore.Users/Migrations.cs
deanmarcussen: Cool, I couldn't remember. Tested without, works fine

davidhayden:
Looking forward to this! Thank you for all your effort @deanmarcussen @jtkech 👍

jtkech (approved):

PR #9707, opened 2021-06-17
Fix Minor security issue: Internal File Path Disclosure

### LINKED ISSUE #9379: Minor security issue: Internal File Path Disclosure

Skrypt:
Fixes #9379

### DIFF
--- a/src/OrchardCore/OrchardCore.FileStorage.AzureBlob/BlobFileStore.cs
+++ b/src/OrchardCore/OrchardCore.FileStorage.AzureBlob/BlobFileStore.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Linq;
 using System.Net;
 using System.Text;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


src/OrchardCore/OrchardCore.FileStorage.AzureBlob/BlobFileStore.cs
```diff
@@ -61,43 +60,65 @@ public BlobFileStore(BlobStorageOptions options, IClock clock, IContentTypeProvi
...
 
-            var properties = await blob.GetPropertiesAsync();
+                var properties = await blob.GetPropertiesAsync();
 
-            return new BlobFile(path, properties.Value.ContentLength, properties.Value.LastModified);
+                return new BlobFile(path, properties.Value.ContentLength, properties.Value.LastModified);
+            }
+            catch
+            {
+                throw new FileStoreException($"Cannot get file info with path '{path}'.");
```
deanmarcussen, line 78:
```
catch (Exception ex)
{
    throw new FileStoreException($"Cannot get file info with path '{path}'.", ex);
}
```

src/OrchardCore/OrchardCore.FileStorage.AzureBlob/BlobFileStore.cs
```diff
@@ -180,106 +201,147 @@ public async Task<bool> TryCreateDirectoryAsync(string path)
...
 
-            if (!await blob.ExistsAsync())
+                if (!await blob.ExistsAsync())
+                {
+                    throw new FileStoreException($"Cannot get file stream because the file '{path}' does not exist.");
+                }
+
+                return (await blob.DownloadAsync()).Value.Content;
+            }
+            catch
```
deanmarcussen, line 341: catch (FileStoreException)
{
    throw;
}
catch (Exception ex)
{
    throw new FileStoreException($"Cannot get file info with path '{path}'.", ex);
}

deanmarcussen (commented):
Great.

What I would suggest is on the methods where we don't internally throw a `FileStoreException` already, is that we catch the unknown exception, and add it to the `FileStoreException` as the `InnerException`

This way we keep the stack trace, for the logs, but the media app / controllers won't be reading the `InnerException`

On the ones where we are throwing a `FileStoreException`, for known errors, we can catch that as `catch (FileStoreException)` and directly rethrow it with `throw;`, and then we have that known message, and no break in the strack trace.
And still catch the general exceptions and rethrow them.

Hopefully makes sense

deanmarcussen (approved):

PR #9726, opened 2021-06-18
DefaultContentManager CreateAsync with VersionOptions.Published issue

### LINKED ISSUE #9694: DefaultContentManager.CreateAsync doesn't publish any version after the first version
### Describe the bug
In my code I have something like this
```
await _contentManager.UpdateValidateAndCreateAsync(contentItem, publish ? VersionOptions.Published : VersionOptions.Draft);
```

This works for the initial creation of the content item, but doesn't work if this is a new version for an existing content item.

### Expected behavior
If I pass VersionOptions.Published I expect that the new version will be published.

### Screenshots
I think the problem is in the `DefaultContentManager.CreateAsync` method.
image

At line 431 the Published flag is set to true, but this only happens if the content item doesn't already have a content item version ID. Furthermore, at line 452 the Publishing and Published events are fired, but the Published flag on the content item is still false.

I'm creating my new version like this
```
await _contentManager.GetAsync(contentItemId, VersionOptions.DraftRequired);
```
 which appears to set the content item version ID right away.

### Workaround
Do something like the following
```
ContentValidateResult contentValidateResult = await _contentManager.UpdateValidateAndCreateAsync(contentItem, VersionOptions.Draft);
if (contentValidateResult.Succeeded && publish)
{
    await _contentManager.PublishAsync(contentItem);
}
```

Skrypt:
Forces ContentItem.Published when VersionOptions.Published in DefaultContentManager.CreateAsync.

Fixes #9694

### DIFF
--- a/src/OrchardCore/OrchardCore.ContentManagement/DefaultContentManager.cs
+++ b/src/OrchardCore/OrchardCore.ContentManagement/DefaultContentManager.cs
@@ -395,7 +395,7 @@ namespace OrchardCore.ContentManagement
 
                 if (latestVersion != null)
                 {
-                    _session.Save(latestVersion);
+                    _session.Save(latestVersion, checkConcurrency: true);
                 }
             }
 
@@ -444,13 +444,14 @@ namespace OrchardCore.ContentManagement
             // invoke handlers to add information to persistent stores
             await Handlers.InvokeAsync((handler, context) => handler.CreatingAsync(context), context, _logger);
 
-            _session.Save(contentItem);
+            _session.Save(contentItem, checkConcurrency: true);
             _contentManagerSession.Store(contentItem);
 
             await ReversedHandlers.InvokeAsync((handler, context) => handler.CreatedAsync(context), context, _logger);
 
             if (options.IsPublished)
             {
+                contentItem.Published = true;
                 var publishContext = new PublishContentContext(contentItem, null);
 
                 // invoke handlers to acquire state, or at least establish lazy loading callbacks
@@ -607,7 +608,7 @@ namespace OrchardCore.ContentManagement
 
             await Handlers.InvokeAsync((handler, context) => handler.UpdatingAsync(context), context, _logger);
 
-            _session.Save(contentItem);
+            _session.Save(contentItem, checkConcurrency: true);
 
             await ReversedHandlers.InvokeAsync((handler, context) => handler.UpdatedAsync(context), context, _logger);
         }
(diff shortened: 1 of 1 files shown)


Skrypt:
@rjpowers10 See if that fixes the issue for you.

src/OrchardCore/OrchardCore.ContentManagement/DefaultContentManager.cs
```diff
@@ -438,13 +438,18 @@ public async Task CreateAsync(ContentItem contentItem, VersionOptions options)
                 contentItem.Published = false;
             }
 
+            if (options.IsPublished)
+            {
+                contentItem.Published = true;
+            }
```
rjpowers10, line 444: I haven't run the code but this is the fix I was expecting.

On line 457 it seems like it would be safer to do `if (contentItem.Published)` instead of `if (options.IsPublished)` when deciding to fire the publish events.

src/OrchardCore/OrchardCore.ContentManagement/DefaultContentManager.cs
```diff
@@ -395,7 +395,7 @@ protected async Task<ContentItem> BuildNewVersionAsync(ContentItem existingConte
 
                 if (latestVersion != null)
                 {
-                    _session.Save(latestVersion);
+                    _session.Save(latestVersion, checkConcurrency: true);
```
Skrypt, line 398: Diff is bugged, we just merged this.

src/OrchardCore/OrchardCore.ContentManagement/DefaultContentManager.cs
Skrypt: I'm not sure why we assumed that the ContentItem would be set to Published = true here. I think the method was used to publish all the time at first. And then we decided that we would need to create a draft and then publish it. Need to see before merging if the logic make sense in all use cases we have.

Skrypt: Else, not sure why we would need to have a VersionOptions passed in these methods. There's probably a reason and I need to find why first.

Skrypt: I think the idea is that you could alter the ContentItem in any of the Events fired. So, if we pass a VersionOptions.Published then we retain the main intent and then don't need to look if the ContentItem.Published = true. Though, the issue here is that if it's not set to Published it won't create a Published content item. Then, I think you could have set your ContentItem.Published = true before calling UpdateValidateAndCreateAsync method. Just thinking out loud but need to validate.

PR #9766, opened 2021-06-24
Fix bootstrap select

deanmarcussen:
Somewhere recently, perhaps with the upgrades to bootstrap 5? the content list filters seem to have stopped working.

this just adds a dependency to the `bootstrap-select` to get them running again


src/OrchardCore.Modules/OrchardCore.Resources/ResourceManagementOptionsConfiguration.cs
```diff
@@ -195,7 +195,7 @@ ResourceManifest BuildManifest()
 
             manifest
                 .DefineScript("bootstrap-select")
-                .SetDependencies("jQuery")
+                .SetDependencies("bootstrap:4.6.0")
```
jptissot, line 198: This causes a javascript error 🤔

PR #9087, opened 2021-04-07
GraphQL 4

agriffard:
Giving another try to see if we can migrate to GraphQL 4.

Many changes in versions 3 and 4:
https://graphql-dotnet.github.io/docs/migrations/migration3/
https://graphql-dotnet.github.io/docs/migrations/migration4/

update from @carlwoodhouse 

I fixed a bunch of the tests and issues, still some stuff outstanding. Not sure of the state of the lucene stuff either as im not as familiar with that integration but suggest we fix the other issues before that

- [x]  Dynamic fields are broken
- [x]  Custom validators for permissions need re-implementing to new spec
- [x]  Custom validators for max results need re-implementing to new spec - fixed in #10782
- [x]  Check metadata thread safety changes for permissions/part attributes meta
- [ ]  Remove asynclock resolver and replace with built in scoped resolver / check this works with parallel execution strategy
- [ ]  Check dataloader changes are actually deferring load
- [x]  Check GraphQLFieldNameAttribute changes; how we've resolved it atm doesnt seem right to me ...
- [x] Not correctly setting response code on auth issues - fixed in #10782
- [ ] buffered stream issue from (https://github.com/OrchardCMS/OrchardCore/pull/10782)

PR #9769, opened 2021-06-24
Fix DataLoader concurrent issues

### LINKED ISSUE #9651: intermittent error 'Error trying to resolve contentItems' when executing graphQL
It's a super rare issue, but I'm getting error like this occasionally.
I don't know how to reproduce this.
Same GQL query is returning proper data, but once in a time I was able to observe result like this:

image

sebastienros:
Fixes #9651

### DIFF
--- a/src/OrchardCore/OrchardCore.Apis.GraphQL.Abstractions/Extensions/DataLoaderExtensions.cs
+++ b/src/OrchardCore/OrchardCore.Apis.GraphQL.Abstractions/Extensions/DataLoaderExtensions.cs
@@ -7,16 +7,16 @@ namespace OrchardCore.Apis.GraphQL
 {
     public static class DataLoaderExtensions
     {
-        public static Task<T[]> LoadAsync<TKey, T>(this IDataLoader<TKey, T> dataLoader, IEnumerable<TKey> keys)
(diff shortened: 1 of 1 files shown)


sebastienros:
No it won't defer the call, was that the intent?

carlwoodhouse:
How the loader works (or did for this version of graphqldotnet) is to defer
the calls till it has them all so it can make one call at the end of the
request to sql instead of many.

The lockasyncresolver we added is supposed to stop the concurrency issues;
though clearly some issue remains Maybe a resolver isn’t using it ?

sebastienros:
> so it can make one call

where is this logic?

carlwoodhouse:
If I remember right it’s In the graphqldotnet dataloader implementation. If
you look in our graphqlmiddleware you can see we register the listeners for
dataloaders; but it just uses their implementation.

The general idea is that you load the calls into the dataloaders  by key
then the listener executes them all as one for that key to reduce sql
queries and reduce graphqls n+1 issue.

If you profile a query which resolves lists of contentitems as fields on a
type you should see it just runs one query for that dataloader key vs 1
additional query for each result.

Gonna try and pick the graphql update branch/or back up tomorrow which does
handle dataloaders much better.

Piedone:
We don't have this class anymore, nor anything else with `Task.WhenAll()` for GraphQL. Was removed in https://github.com/OrchardCMS/OrchardCore/pull/9087.

PR #9775, opened 2021-06-25
Fix regression renaming admin style resource

Skrypt:
There was none issue with it but just for consistency and in case of.


jptissot:
I found something else..  https://github.com/StatCan/StatCan.OrchardCore/runs/2917054863?check_suite_focus=true#step:7:266

PR #9781, opened 2021-06-25
Support multipart content type with email service.

### LINKED ISSUE #9751: Workflow email task ability to send both HTML and plain-text versions
Right now when we send an email as HTML by checking the : "Does the Body contain HTML" checkbox option on the email task it will be sent only in HTML. Some email softwares will often transfer these emails as plain-text so we should provide also a plain-text version of the email for the email softwares to be able to use. This way, we can provide 2 different Body versions of the email.

When the "Does the Body contain HTML" checkbox is checked we should have a second Body (plain-text) textfield to provide the email as plain text.

Using a second Email task in the workflow without the HTML checkbox checked ; but it will send a second email too. Which is not wanted.

Skrypt:
Fixes #9751

See email example below : 

```
From: jasmin.savard@gmail.com
Date: Fri, 25 Jun 2021 17:34:43 -0400
Subject: Test
Message-Id: <OUJBALN63EU4.IS3IGSIZUPB3@skrypt>
Sender: jasmin.savard@gmail.com
To: jasmin.savard@gmail.com
Reply-To: jasmin.savard@gmail.com
MIME-Version: 1.0
Content-Type: multipart/alternative; boundary="=-9nfRnDakF+Ib69kpAIdY3A=="

--=-9nfRnDakF+Ib69kpAIdY3A==
Content-Type: text/plain; charset=utf-8

Test text
--=-9nfRnDakF+Ib69kpAIdY3A==
Content-Type: text/html; charset=utf-8

<p>Test html</p>
--=-9nfRnDakF+Ib69kpAIdY3A==--
```

If someone was sending an email as plain-text using the Body textarea ; it will try to send it as HTML after merging this PR.
There is now 2 different textareas for defining a Html and a Plain text body.

image


hishamco (changes requested):
I need some time to review this, but my first comment is, please simplify this by adding a single checkbox "HTML Body?" which is checked by default and one textbox for the body, then move the heavy lifting to the code to toggle the Body (Text or HTML)

Skrypt:
It's not a toggle. You are supposed to be able to send a email with both HTML and plain text at the same time. This is called a multipart content type email. As you can see in the email example it allows to send the email as HTML and plain text at the same time. This means that if someones receives his emails only in plain-text they won't be displaying strange chars. Here, what we had is the option to send it as html or plain-text but this is not how a `BodyBuilder` is working. It allows to provide both plain-text and html body so that it can be sent as a multipart content type email. So, the changes here actually are reflecting how a email `BodyBuilder` works.

hishamco:
> You are supposed to be able to send a email with both HTML and plain text at the same time.

What if I want to send only HTML or Text?

Skrypt:
This is what it does. It allows you to send it as HTML or Plain Text only. And allows you also to send it as Html and Plain-Text as a single email with multipart content type.

hishamco (approved):
LGTM

hishamco:
Please some can review the workflow task, but it looks good in general

agriffard:
Please solve conflict after adding new project.

jptissot:
Will this break all current emails ?

Skrypt:
It won't break them. Like I said previously. The only difference is that the `Body` field only sends HTML content now.
So, for those that we're using the Body field to send as plain-text they will just need to move their email content to the proper `Body Plain Text` field. So basically, no issue, the email will be sent with HTML content type containing plain text in it. It should display fine still.

jptissot:
Ok perfect. I don't think I use Plain Text anyways ;)

sebastienros:
If you can provide both fields, then why provide the checkboxes? It knows if there is a content in either text area, and they both have to be passed if available, or just one. Both text areas would be optional, but at least one should be not empty.

Skrypt:
I kept the checkboxes for having the ability to toggle it. Even if someone composed a plain-text version it doesn't imply that he want to send it as plain-text. You can then disable the plain-text version while keeping what you composed in the textarea. Same for the HTML version. I don't think a body is mandatory for sending a email so not sure about making a Body mandatory.

Skrypt:
Something that could be done is creating Email Liquid Templates. This way we could define if a email would be sent as HTML and/or Plain-text by defining these 2 templates with a prefix. If one prefix is not found then don't send it as multi-part. Here, the thing I don't really like about these textareas is the fact that they should be Templates defined in the Admin UI in the Admin Templates. So, I kind of agree with what you are suggesting, but here since we can't assign any templates to the Workflow task I think it is better to just have checkboxes to toggle these.

deanmarcussen:
Would quite like a wrapper shape that is empty by default @Skrypt - such that a site may set a email header and footer, across all workflow emails, and the user / editor only has to provide the content / body. Was going to mention it on this pr, but this pr was for a different feature :)

PR #9808, opened 2021-06-29
URI components Liquid accessors

jtkech:
Related to #9617 and #9618

Provides correct escaping to reconstruct a request or redirect URI.


sebastienros:
Can these replaces the existing ones?

jtkech:
> Can these replaces the existing ones?

No, the existing ones are still useful to get the non escaped original values.

deanmarcussen:
Needs adding to the docs as well @jtkech  :) otherwise looks fine

jtkech:
@deanmarcussen done through #10179 ;)

PR #9813, opened 2021-06-29
Fix admin menu links to 403 forbidden pages

scleaver:
Fix a couple of admin menu issues for roles other than Administrator where the menu shows a link to a page but the user has no permissions so is presented with 403 forbidden page. Namely:

- Media options page - permissions where different between `AdminController.cs` and `adminmenu.cs`. Corrected the permission name on menu to match controller.
- Culture settings page - editor has `ManageCultures` but does not have `ManageSettings`. The Cultures page being a settings page needs both permissions to view. Removed Editor from the `ManageCultures` permission.

Tested and fixes the issues


scleaver:
Reviewers @deanmarcussen and @agriffard

deanmarcussen (approved):

PR #9815, opened 2021-06-29
Fix rule conditions localization

deanmarcussen:
For some reason I forgot that `IConfigureOptions` are singletons, so there is no point injecting a string localizer to them.

So fixing the localization to be scoped.


sebastienros:
How will the localizer find the strings from the class where they are defined since it's passed from a different component?

deanmarcussen:
It's resolved by making and resolving a `IStringLocalizer<T>` generic type, where `T` is the the class where the strings are defined.

`(sp) => displayText((IStringLocalizer)sp.GetService(typeof(IStringLocalizer<>).MakeGenericType(typeof(TLocalizer))))`

But the useful thing is that the po extractor also finds these, when the func signature is of type `Func<IStringLocalizer...` and the property name is `S`

It could probably cache the `MakeGenericType` once

PR #9821, opened 2021-06-30
Fix task.json for powershell

ns8482e:
Fixed following build error  encountered while using powershell in vscode.

image

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -21,7 +21,6 @@
   "tasks": [
     {
       "label": "Dotnet Restore",
-      "type": "shell",
       "command": "dotnet",
       "args": [
         "restore"
@@ -29,7 +28,6 @@
     },
     {
       "label": "Dotnet Restore Force",
-      "type": "shell",
       "command": "dotnet",
       "args": [
         "restore",
@@ -38,7 +36,6 @@
     },
     {
       "label": "Dotnet clean (debug)",
-      "type": "shell",
       "command": "dotnet",
       "args": [
         "clean",
@@ -48,7 +45,6 @@
     },
     {
       "label": "Dotnet clean (release)",
-      "type": "shell",
       "command": "dotnet",
       "args": [
         "clean",
@@ -58,7 +54,6 @@
     },
     {
       "label": "Web App: Dotnet Build Debug - net5.0",
-      "type": "shell",
       "command": "dotnet",
       "args": [
         "build",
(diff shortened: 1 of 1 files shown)


Skrypt:
Have you tried to see if it still works in Linux then?

ns8482e:
@Skrypt  no I haven’t tried in Linux

Skrypt:
Please try it, I'm pretty sure it will break the tasks. That's why we use the "Shell" type of command because it is universal across different OS's. Also, powershell is not necessary.

ns8482e:
My vs code uses powershell For all task.json related task

deanmarcussen:
I’ll try it on the Mac tomorrow and let you know

deanmarcussen:
> Please try it, I'm pretty sure it will break the tasks. That's why we use the "Shell" type of command because it is universal across different OS's. Also, powershell is not necessary.

Pretty much what @Skrypt says, doesn't seem to be the right fix for the issue with powershell.

ns8482e:
@deanmarcussen, @Skrypt I guess that's correct, But I don't understand why its taking powershell instead of cmd on windows?

Skrypt:
Maybe because your default shell on Windows is set to Powershell.

ns8482e:
@Skrypt  understood - I do have power shell as default in vscode that i  use in other repos. So I don't want to change it in global settings.

 What I don't understand is, Task.json has already defined following in OC repo 

```json
"windows": {
    "command": "cmd",
    "args": [
      "/C"
    ]
  },
```

Why it's not taking `cmd` and uses still `powershell`  for OC

Skrypt:
image

I think that the description here explains what it does. It overrides any tasks.json settings to use what you defined globally in this "windows" settings.

Skrypt:
I'm doing some reading and it would be possible to only use Powershell for any OS since you can install it on Linux and OSX. But the issue is that it requires to configure VS Code accordingly by adding this in the settings.json file of our project : 

```
{
    "terminal.integrated.automationShell.windows": "C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe"
}
```

Skrypt:
Here some reading : https://code.visualstudio.com/docs/editor/integrated-terminal#_configuring-profiles

ns8482e:
@Skrypt  @deanmarcussen  I am not sure why it's not working for me - I am using default installation of VS Code. All other .NET projects builds fine except orchard core.

Based your above suggestion, I updated the `task.json` as following for `windows` section and it works fine - update the PR

```json
"windows": {
    "options": {
      "shell": {
        "executable": "cmd.exe",
        "args": [
          "/C"
        ]
      }
    }
  }
```

Skrypt:
I think this will work.

PR #9826, opened 2021-06-30
Audit Trail packages

agriffard:
/cc @domonkosgabor Can you please check that the audit trail module is still working with the updated packages to the latest version?


domonkosgabor:
@agriffard It looks fine with the updated packages.

PR #9833, opened 2021-06-30
Preview: RenderBody needs to be called before RenderBodyAsync

john-s-morgan:
RenderBodyAsync() seems to be throwing the following:

InvalidOperationException: RenderBody has not been called for the page at '/Areas/OrchardCore.ContentPreview/Views/_Layout.cshtml'. To ignore call IgnoreBody().

From looking at previous PRs, this is expected if RenderBody() has never been called, even if the _ViewImports.cshtml include `@inherits OrchardCore.DisplayManagement.Razor.RazorPage<TModel>`

This creates the initial RenderBody call

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentPreview/Views/_Layout.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.ContentPreview/Views/_Layout.cshtml
@@ -6,6 +6,6 @@
       <title>@RenderTitleSegments(Site.SiteName, "before")</title>
   </head>
   <body dir="@Orchard.CultureDir()">
-          @await RenderBodyAsync()
+          @RenderBody()
   </body>
 </html>


Skrypt:
I remember vaguely @sebastienros saying why we we're leaving this Layout to null. Maybe it will work with your fix here. Need to investigate further.

Skrypt:
I think we will need to revert the previous PR because even when using this PR it fails rendering the Preview. Need to see why.

jtkech:
We set `Layout` to null to prevent our hard coded theme layout from being executed twice, if you remove this line and put a break point on the following line, you will see that it is called twice, the 2nd one when the iframe is loaded.

https://github.com/OrchardCMS/OrchardCore/blob/16a06e80384dde3d7eaa69d0f73259946892464d/src/OrchardCore/OrchardCore.DisplayManagement/Theming/ThemeLayout.cs#L22

Notice that `RenderLayoutBody()` calls `RenderBody()`.

https://github.com/OrchardCMS/OrchardCore/blob/16a06e80384dde3d7eaa69d0f73259946892464d/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs#L253

Hmm, yes as for a regular mvc controller view you can override the `Layout`, i suggest to put this `_Layout.cshtml` file under the `Views/Preview` folder and then just use `Layout = "_Layout"` (the .cshtml is not required).

Then, yes `RenderBodyAsync()` needs that `RenderBody()` is already called, this is what our coded theme layout does, but here, as for a regular mvc view for which you override our theme layout (but only for the 1st rendering, not the following ones for previewing), here normally you can add what is missing and then you just need to call `RenderBody()`.

Just tried, seems to work, but unless @Skrypt saw another issue i'm not aware as it seems to not work for him.

Skrypt:
I tried it and it seems like it works only the third time I refresh the page on my side. First request I get an "undefined" text in the page and nothing else.

PR #9819, opened 2021-06-29
Add mobile viewport to content preview.

john-s-morgan:
Resolves #9818 by adding a basic layout with a viewport.

PR #9837, opened 2021-07-01
Preview : fix javascript displaying "undefined" 

Skrypt:
Because not displaying the proper iframe.


jtkech:
@Skrypt Just for info

- I was not able to repro the `undefined` rendering but i trust you ;)

- I saw that you removed the usage of `iframe2` in some places, so i tried to remove its usage everywhere and it seems to be working the same, so maybe we could do some cleanups if we don't re-use it, but see below.

- As i remember `iframe2` was to keep the current scrolling position of the preview, e.g. in case of a long document, and without seeing the preview flickering. Now when you update a document, the preview returns and stays to the top.

    So i returned back before #9819, indeed the scrolling position was kept, no flickering, but the preview was first scrolling up to the top and then scrolling back to its previous position, but as i remember the goal was to just see it keeping its position without flickering but also without seeing it moving.

So maybe something was already changed before #9819, if you think it's worth trying to retrieve the original behavior we could see it together, otherwise maybe only cleanup the usage of `iframe2` if we don't re-use it, let me know.

Skrypt:
I thought it was use because of some flickering happening when a single iframe was used. I think it still use both iframes but seems like there was an issue on startup of the preview where it was assigning the content to the second iframe without showing it. I did not analyze this further though. For now, it works. If you want to push more work on this no problem.

jtkech:
Okay no problem, i will see, in the meantime here what i noticed.

- I also noticed that there is another preview index view under the OC.Templates module using the same technique, so we would need to also update this view.

- Yes it works but the scrolling position is no more preserved, keeping the position is useful for a long document, when you have scrolled down to a position related to the data you are updating. Yes we were swapping the iframes to not see the current one scrolling up and down while updating it, as i remember the scrolling was fast that's why we were talking about flickering. But as said, now when i use the previous code i see the preview scrolling up and down slowly.

    So i think that something else has changed before the 2 related PRs. Maybe a new jQuery version made this scrolling "animated / delayed", this would explain that now the previous code render the other iframe before its scrolling has completed.

Maybe i will do a PR if i find a solution to retrieve the old behavior, if so I will ask you to check if you still have the "undefined" issue.

Skrypt:
All good. I just wanted to "at least" commit bites that works first without thinking about it too much. Maybe we could easily add back these 2 lines where it scrolls the page but need to test more.

jtkech:
Okay cool, and you were right to fix it ;)

I will give it a last try this night, maybe i found something.

PR #9844, opened 2021-07-03
Subject & Body are not required if the confirmation is false

### LINKED ISSUE #9760: RegisterUserTask: Subject & Template for confirmation email should not be required if Send Email is unchecked
### Describe the bug
Subject & Template for confirmation email are required even if Send Confirmation Email is not checked
### To Reproduce
1. Workflows
2. Create Workflow
3. Add Task "Register User"

### Expected behavior
Subject & Template for confirmation email fields should be required only if Send Confirmation Email is checked

### Screenshots
image

hishamco:
Fixes #9760


src/OrchardCore.Modules/OrchardCore.Users/Views/Items/RegisterUserTask.Fields.Edit.cshtml
```diff
@@ -11,14 +11,27 @@
...
     <span asp-validation-for="ConfirmationEmailTemplate"></span>
     <span class="hint">@T["The message to send. Access email confirmation url with Workflow.Properties.EmailConfirmationUrl liquid property."]</span>
 </div>
+
+<script at="Foot">
+    $(function () {
+        var required = false;
+
+        $("#@Html.IdFor(m => m.SendConfirmationEmail)").change(function () {
+            $("#@Html.IdFor(m => m.ConfirmationEmailSubject)").attr('required', required);
```
deanmarcussen, line 31: Fine to add some javascript helpers, but the validation also needs to be performed in the display driver

hishamco: I will try to add some validation on the display driver, coz all the time I see the validation on view model using data annotations

hishamco: @deanmarcussen I added a validation on the display driver, but I think it's not needed because the workflow already checked the `SendConfirmationEmail` before set the expressions for other properties

src/OrchardCore.Modules/OrchardCore.Users/Workflows/Drivers/RegisterUserTaskDisplayDriver.cs
```diff
@@ -10,15 +10,23 @@ public class RegisterUserTaskDisplayDriver : ActivityDisplayDriver<RegisterUserT
...
 
         protected override void UpdateActivity(RegisterUserTaskViewModel model, RegisterUserTask activity)
         {
             activity.SendConfirmationEmail = model.SendConfirmationEmail;
-            activity.ConfirmationEmailSubject = new WorkflowExpression<string>(model.ConfirmationEmailSubject);
-            activity.ConfirmationEmailTemplate = new WorkflowExpression<string>(model.ConfirmationEmailTemplate);
+
+            if (model.SendConfirmationEmail)
+            {
+                activity.ConfirmationEmailSubject = new WorkflowExpression<string>(model.ConfirmationEmailSubject);
```
deanmarcussen, line 27: This isn't validation.

Here is an example of how to add validation to a driver
https://github.com/OrchardCMS/OrchardCore/blob/af6e3313f7a650ea3e2c4931218fb622173957b4/src/OrchardCore.Modules/OrchardCore.ContentFields/Drivers/TextFieldDisplayDriver.cs#L55

hishamco: @deanmarcussen if I use the overload that has the `IUpdaterModel` I can't access the `RegisterUserTask`, any idea?

deanmarcussen: `public async override Task<IDisplayResult> UpdateAsync(TActivity model, IUpdateModel updater)`

try this method

hishamco: I already did that but if you check my previous comment the missing part is `RegisterUserTask` which I can't find

deanmarcussen: @hishamco if you take the time to look and see what that method does, you will see it is merely a helper to create the viewmodel, and bind it. Override it, and change the code to something like this

```
        public async override Task<IDisplayResult> UpdateAsync(TActivity model, IUpdateModel updater)
        {
            var viewModel = new TEditViewModel();
            if (await updater.TryUpdateModelAsync(viewModel, Prefix))
            {
                //await UpdateActivityAsync(viewModel, model);
                Do the regular viewmodel -> model mappings here.
                Validate the values here.
             }

            return Edit(model);
        }
```

hishamco: I though there's an overload with both `IUpdateModel` and `RegisterUserTask` but I don't find it, anyhow I updated the PR, but the strange part that the validation error messages don't show up, but when I place validation summary they show up

deanmarcussen: Looks like you're missing the `Prefix` property when adding the error to the model state

can you merge dev as well, and add the new property in

hishamco: I was wonder why it's important

src/OrchardCore.Modules/OrchardCore.Users/Workflows/Drivers/RegisterUserTaskDisplayDriver.cs
```diff
@@ -15,12 +27,24 @@ protected override void EditActivity(RegisterUserTask activity, RegisterUserTask
...
+            {
+                model.SendConfirmationEmail = viewModel.SendConfirmationEmail;
+                model.RequireModeration = viewModel.RequireModeration;
+                model.ConfirmationEmailSubject = new WorkflowExpression<string>(viewModel.ConfirmationEmailSubject);
+                model.ConfirmationEmailTemplate = new WorkflowExpression<string>(viewModel.ConfirmationEmailTemplate);
+
+                if (model.SendConfirmationEmail)
+                {
+                    updater.ModelState.AddModelError(Prefix, nameof(viewModel.ConfirmationEmailSubject), S["A value is required for {0}.", nameof(viewModel.ConfirmationEmailSubject)]);
+                    updater.ModelState.AddModelError(Prefix, nameof(viewModel.ConfirmationEmailTemplate), S["A value is required for {0}.", nameof(viewModel.ConfirmationEmailTemplate)]);
```
deanmarcussen, line 43: you need to test here if the `ConfirmationEmailSubject` is empty or not.
Otherwise this will error every time you try and save.

hishamco: Seems we need to check both subject and template before adding the error to the model state

src/OrchardCore.Modules/OrchardCore.Users/Workflows/Drivers/RegisterUserTaskDisplayDriver.cs
```diff
@@ -15,12 +27,31 @@ protected override void EditActivity(RegisterUserTask activity, RegisterUserTask
...
+            if (await updater.TryUpdateModelAsync(viewModel, Prefix))
+            {
+                model.SendConfirmationEmail = viewModel.SendConfirmationEmail;
+                model.RequireModeration = viewModel.RequireModeration;
+                model.ConfirmationEmailSubject = new WorkflowExpression<string>(viewModel.ConfirmationEmailSubject);
+                model.ConfirmationEmailTemplate = new WorkflowExpression<string>(viewModel.ConfirmationEmailTemplate);
+
+                if (model.SendConfirmationEmail)
+                {
+                    if (string.IsNullOrEmpty(viewModel.ConfirmationEmailSubject))
```
sebastienros, line 42: OrWhiteSpace

src/OrchardCore.Modules/OrchardCore.Users/Views/Items/RegisterUserTask.Fields.Edit.cshtml
```diff
@@ -11,14 +11,14 @@
 
 <div class="form-group" asp-validation-class-for="ConfirmationEmailSubject">
     <label asp-for="ConfirmationEmailSubject">@T["Subject for confirmation email"]</label>
-    <input type="text" asp-for="ConfirmationEmailSubject" class="form-control code" />
+    <input type="text" asp-for="ConfirmationEmailSubject" class="form-control code" required />
```
sebastienros, line 14: Doesn't it make it required by default, thought the Send Email is not checked by default? (or is it checked)

hishamco: I think it's let me double check ..

hishamco: Register User Task

deanmarcussen: That will only work on Create not on Edit, where the value may be different to the default js you are suggesting.

I would drop the javascript validation personally

hishamco: > That will only work on Create not on Edit

How?

> I would drop the javascript validation personally

We could remove the client-side validation and rely on server-side instead

deanmarcussen: Because your js code assumes the box is checked. When editing the box maybe unchecked and the js code still makes it required.

hishamco: I remove the client-side validation and rely on server-side once

deanmarcussen (approved):

Skrypt (approved):

Issue #9856, opened 2021-07-05
Support custom parameters in the OpenId Connect client


deanmarcussen:
### Is your feature request related to a problem? Please describe.

The `OpenIdConnectOptions` support an `OnRedirectToIdentityProvider` event feature which allows setting of custom parameters on the protocal message when generating an `OpenIdConnectMessage` to an external provider.

We need to be able to send some custom parameters to some of our tenants `AzureB2C` auth servers (but not all, and it varies per tenant)

i.e.
```
            options.Events.OnRedirectToIdentityProvider = (context) =>
            {
                context.ProtocolMessage.SetParameter("foo", "bar");

                return Task.CompletedTask;
            };
```

### Describe the solution you'd like

Editable parameter (kvp) options on the `OpenIdConnectSettings` so we can configure different tenants to use different custom parameters.

### Describe alternatives you've considered

`OpenIdConnectOptions` is configurable vis `PostConfigure` so we can do this manually, in code, but it would be nicer if we were able to support this through UI based settings.

// cc @MichaelPetrinolis @kevinchalet 

Currently working on a pr for this unless anyone sees any issues with supporting this through the UI?

MichaelPetrinolis:
The only thing is that we will cover only a fraction of the functionality provided by the Events. But could be extended as needed :)

PR #9863, opened 2021-07-06
OpenID Client parameters

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/9856

Adds a parameters option to set custom parameters on the authenticate request.

Screenshot 2021-07-05 at 16 06 03


src/OrchardCore.Modules/OrchardCore.OpenId/Views/OpenIdClientSettings.Edit.cshtml
```diff
@@ -136,6 +144,79 @@
...
+</script>
+
+<div id="@Html.IdFor(m => m)" class="field-editor field-editor-parameters">
+    <div class="row mb-3">
+        <div class="col-12 col-lg-6">
+            <label asp-for="Parameters">@T["Parameters"]</label>
+            <a href="javascript:void(0)" v-on:click="showModal" class="float-right" title="@T["Edit Data"]"><i class="fas fa-edit"></i></a>
+            <parameters-table :data="sharedState"></parameters-table>
+            <parameters-modal :data="sharedState" v-on:reload-data="reloadData()"></parameters-modal>
+            <div class="hint">@T["Optional parameters that can sent with the authorize request."]</div>
```
jtkech, line 215:
```suggestion
            <div class="hint">@T["Optional parameters that can be sent with the authorize request."]</div>
```

But not sure, maybe it is allowed in english ;)

deanmarcussen: No, just a bad English sentence from me ;)

src/OrchardCore.Modules/OrchardCore.OpenId/Configuration/OpenIdClientConfiguration.cs
```diff
@@ -97,6 +97,20 @@ public void Configure(string name, OpenIdConnectOptions options)
...
             }
+
+            if (settings.Parameters.Any())
+            {
+                var parameters = settings.Parameters;
+                options.Events.OnRedirectToIdentityProvider = (context) =>
+                {
+                    foreach(var parameter in parameters)
+                    {
+                        context.ProtocolMessage.SetParameter(parameter.Name, parameter.Value);
```
kevinchalet, line 108: It's worth noting that `SetParameter()` allows overriding existing parameters. Do we want to support that kind of extensibility?

deanmarcussen: I wondered if that would be a concern - but the extensibility is there anyway, this just reveals it via a ui, so adds more flexibility - something we need for a particular azure client connection

kevinchalet: I'm asking that because I regularly people trying to replace the default parameters already generated by an OIDC client without realizing it's often not a good idea. To emphasize that, you may want to rename the title to "Additional parameters" or add an "advanced" label somewhere 😃

kevinchalet: Also, I'm curious: why not using the Microsoft.Identity.Web integration for Azure?

deanmarcussen: Updated to `Advanced Parameters` and I made a note in the docs as well. Hopefully suits @kevinchalet 😄 

Re : Microsoft.Identity.Web for azure.
I'm using that as well, just have (currently) two external providers, which both happen to be different variants of AzureAD, and only one of is after some extra params.

We're also tracking https://github.com/OrchardCMS/OrchardCore/issues/8752 upgrading the current AzureAD provider to `Microsoft.Identity.Web` as currently it's still on the older versions.

vitalybrandes:
@deanmarcussen 
Is this fix somehow affect recipe setup of :
"name": "OpenIdClientSettings"?

PR #9867, opened 2021-07-06
Google Tag Manager

Skrypt:
# TODO

- [x] ~Add missing script after body tag in frontend when module is enabled and a ContainerId is set.~ Won't do, it's an unecessary script for people not allowing javascript in their browsers. It can be added manually in a theme if needed.
- [x] Use newer generic site settings method for DeploymentStep and RecipeStepHandler.


src/OrchardCore.Modules/OrchardCore.Google/TagManager/Deployment/GoogleTagManagerDeploymentStepDriver.cs
```diff
@@ -0,0 +1,23 @@
+using OrchardCore.Deployment;
+using OrchardCore.DisplayManagement.Handlers;
+using OrchardCore.DisplayManagement.Views;
+
+namespace OrchardCore.Google.TagManager.Deployment
+{
+    public class GoogleTagManagerDeploymentStepDriver : DisplayDriver<DeploymentStep, GoogleTagManagerDeploymentStep>
+    {
+        public override IDisplayResult Display(GoogleTagManagerDeploymentStep step)
```
deanmarcussen, line 9: We should be able to skip all the `DeploymentStep` and `RecipeStepHandler` for these by using the newer generic site settings methods

See here for when I am improving it (but the old code works just fine too)

https://github.com/OrchardCMS/OrchardCore/pull/9728

just saves having more unecesary code to maintain for deployments and recipes.

Other than that all looks good

Skrypt: Can we merge and fix the remaining when your PR is merged. I have also changes in the Google Analytics feature to allow using multiple MeasurementId's that I will push in a second PR.

deanmarcussen: The idea is to not have these classes views at all.
So for now you can still use the existing deployment / recipe registration (my pr is just some nicer registration, so I can update that later after this is merged)

```
            services.AddTransient<IDeploymentSource, SiteSettingsPropertyDeploymentSource< GoogleTagManagerSettings >>();
            services.AddScoped<IDisplayDriver<DeploymentStep>>(sp =>
            {
                var S = sp.GetService<IStringLocalizer<DeploymentStartup>>();
                return new SiteSettingsPropertyDeploymentStepDriver< GoogleTagManagerSettings >(S["Admin settings"], S["Exports the admin settings."]);
            });
            services.AddSingleton<IDeploymentStepFactory>(new SiteSettingsPropertyDeploymentStepFactory< GoogleTagManagerSettings >());
```

So if you add this instead (with the localization), to the`Startup`, and then you can delete all the deployment views / drivers and recipe steps - if we do it after merge, then we get stuck with a bunch of code that can't be removed.

Just about less code to maintain ;)

Skrypt: ok done

src/OrchardCore.Modules/OrchardCore.Google/TagManager/Recipes/GoogleTagManagerSettingsStep.cs
```diff
@@ -0,0 +1,38 @@
...
+using OrchardCore.Recipes.Models;
+using OrchardCore.Recipes.Services;
+using OrchardCore.Settings;
+
+namespace OrchardCore.Google.TagManager.Recipes
+{
+    /// <summary>
+    /// This recipe step sets Google TagManager settings.
+    /// </summary>
+    public class GoogleTagManagerSettingsStep : IRecipeStepHandler
```
deanmarcussen, line 15: We can drop this now too :) it gets handled by the site settings recipe step handler

src/OrchardCore.Modules/OrchardCore.Google/Views/GoogleTagManagerSettings.Edit.cshtml
```diff
@@ -0,0 +1,18 @@
+@using OrchardCore.Google.TagManager.ViewModels;
+@model GoogleTagManagerSettingsViewModel;
+
+<p class="alert alert-warning">
+    @T["The current tenant will be reloaded when the settings are saved."]
```
deanmarcussen, line 5: I don't see a reset for the shell in the driver?

It doesn't look like it would be necesary to reset, or has the filter cached the result?

Skrypt: Yes I think _scriptsCache is a cached value in the GoogleTagManagerFilter. But I don't see where we reset the tenant. If we don't reset it then we don't reset it also for the GoogleAnalytics MeasurementId.

Skrypt: I think the issue is that it is not cached and that it should be but I removed the message because we could simply invalidate that cached value without restarting the tenant.

Skrypt: Ok, I get it, now the value is cached in the `await _siteService.GetSiteSettingsAsync()` so it is invalidated when we update the value of the setting simply. So it is cached.

PR #9889, opened 2021-07-11
Preview issues

### LINKED ISSUE #9888: Preview issues
- When doing an update in the editor view, the preview doesn't keep its scrolling position.

- Previewing the Landing Page in the Agency is very slow, around 3 sec, e.g. entering 3 chars will take 3 * 3 = 9 sec.

- The Preview view in the Templates module need to be updated as well.

There were some recent changes on the Preview where @Skrypt already fixed an issue, here i saw another one not related to these changes, and about the scrolling position, anyway if we re-use an older code, even if the scolling position is kept, on the first editor update we see the preview (if not at the top) moving to the top and then returning back to its position.

I have a working solution that i will comment, will do a PR.

Maybe good to be considered before 1.0 if it is not too late ;)

jtkech:
Fixes #9888

There were some recent changes on the Preview where @Skrypt already fixed an issue, here i saw another one not related to these changes, and about the scrolling position, anyway if we re-use an older code, even if the scolling position is kept, on the first editor update we see the preview (if not at the top) moving to the top and then returning back to its position.

Here the issues as described in the related #9888 and their related fix.

- When doing an update in the editor view, the preview doesn't keep its scrolling position.

    I pass the details but here i first removed all the code to swap between 2 frames. Then all was working unless that the preview was still moving (if not at the top) on the first editor update, as with an older code before the recent changes.

    I saw that to fix it we only need some delay in the first `iframe.onload()` that really block the thread. Another way is to do a page refresh once after this first frame loading. Here, we just rewrite the frame document in this first loading.

- Previewing the Landing Page in the Agency is very slow, around 3 sec, e.g. entering 3 chars will take 3 * 3 = 9 sec.

    Here in the Agency layout, for the `font-awesome` I needed to re-use the `style` tag in place of the `script` tag. It seems that using the `font-awesome` script adds many elements to the dom that don't play well while previewing.

- The Preview view in the Templates module need to be updated as well.

    Here to be consistent, I also updated the Preview in the Templates module


Skrypt:
The script converts the icons as .svg files. Using the .css will use the fontawesome font to render icons. Not sure what is better.

deanmarcussen:
>  Using the .css will use the fontawesome font to render icons. Not sure what is better.

Given how much performance can be impacted by using the font awesome javascript, I'm ok with moving it back to css. There's lots of commentary on stackoverflow about performance issues re fa js, and I already had to fix the media app for it once before in the admin.

jtkech:
@Skrypt @deanmarcussen Thanks for the info

And as i remember i was seeing hundreds of elements added to the frame document.

So i let you try it and see if it is worth to merge it knowing that currently

- The scrolling position of the Preview is not preserved on an editor update.

- Previewing the Landing page of the Agency theme is very very slow.

- The Preview view of the Templates module is no longer consistent with the one of the Preview module.

Maybe there are other issues but it fixes the above issues that i could see.

And Italy won ;)

Skrypt (approved):

Issue #9881, opened 2021-07-09
Theadmin.js cause drop-down menu needs to be clicked twice


EternityTeam:
All drop-down menus need to be clicked twice to display.
Block the request of theadmin.js in devtools can solve this.

### Screenshots
![image](https://user-images.githubusercontent.com/14150259/125071378-c57a6000-e0eb-11eb-9a4f-04842eb37d3b.png)
![image](https://user-images.githubusercontent.com/14150259/125071425-d5923f80-e0eb-11eb-9b51-4835db3c07aa.png)

Skrypt:
I does that on very specific pages in the admin. Not all of them.

EternityTeam:
So this is the expected effect. Using bootstrap select in my custom page is also affected. How can I avoid this feature?

hishamco:
@Skrypt why it need to be clicked twice?

Skrypt:
I don't know yet. Maybe some recent PR merge that does this. I removed the form double-click code I added recently to see if it was not that and it's not.

jtkech:
Yes I confirm, just tried with the last dev with the dropdowns of the content items page

Oops maybe too late to fix it before the 1.0 release ;)

After a search on Google, it seems that most of the time this is because bootstrap.js is loaded twice.

In the dev tools I saw TheAdmin.js script which embeds bootstrap.js

    "inputs": [
      "node_modules/popper.js/dist/umd/popper.js",
      "node_modules/bootstrap/dist/js/bootstrap.js",
      "Assets/js/*.js"
    ],
    "output": "wwwroot/js/TheAdmin.js"

But I also saw before the bootstrap script explicitly as a dependency of bootstrap-select that we also use

            manifest
                .DefineScript("bootstrap-select")
                .SetDependencies("jQuery", "bootstrap:4.6.0")

But not sure, not yet tried

PR #9774, opened 2021-06-25
Fix regression by renaming Admin Script resource

Skrypt:
Fixes #9771

image

PR #9893, opened 2021-07-11
Fix bootstrap regression

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/9881

This was a regression from here https://github.com/OrchardCMS/OrchardCore/pull/9766
before we fixed another regression here https://github.com/OrchardCMS/OrchardCore/pull/9774
when the admin script name got changed, and scripts were incorrectly ordered.

Good catch @jtkech

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Resources/ResourceManagementOptionsConfiguration.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Resources/ResourceManagementOptionsConfiguration.cs
@@ -195,7 +195,7 @@ namespace OrchardCore.Resources
 
             manifest
                 .DefineScript("bootstrap-select")
-                .SetDependencies("jQuery", "bootstrap:4.6.0")
+                .SetDependencies("jQuery")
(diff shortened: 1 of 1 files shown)


jtkech:
Okay cool, yes i saw it just before stopping to work ;)

So I didn't take the time to do a new gulp build to try it.

Hmm, but my first thinking was

- If bootstrap-select really depends on bootstrap, i would keep this dependency.

- Then, i would add the same bootstrap, and maybe popper, as new dependencies to the "admin" script resource, this in its resource definition and / or when injecting it, but no longer embed them directly in TheAdmin.js output file.

deanmarcussen:
> If bootstrap-select really depends on bootstrap, i would keep this dependency.

Then, i would add the same bootstrap, and maybe popper, as new dependencies to the "admin" script resource, this in its resource definition and / or when injecting it, but no longer embed them directly in TheAdmin.js output file.

@jtkech I agree, was my thinking as well, it is a little odd to bundle it in. however in the short term (before v1) I don't think we should rearrange how it has been done. so for later?

Issue #9907, opened 2021-07-13
OpenIdClientSettings in recipe  get exeption


vitalybrandes:
### Describe the bug
Update to latest dev nugets,

Added to recipe
 {
      "name": "OpenIdClientSettings",
      "Authority": "https://AUTORITY",
      "DisplayName": "Login",
      "ClientId": "Manager",
      "CallbackPath": "/signin-oidc",
      "SignedOutCallbackPath": "/signout-callback-oidc",
      "Scopes": "email profile",
      "ResponseMode": "form_post",
      "ResponseType": "code id_token",
      "ClientSecret": "secret"
    },

error:
```

ArgumentNullException: Value cannot be null. (Parameter 'source')

    System.Linq.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
    System.Linq.Enumerable.Any<TSource>(IEnumerable<TSource> source)
    OrchardCore.OpenId.Configuration.OpenIdClientConfiguration.Configure(string name, OpenIdConnectOptions options)
    Microsoft.Extensions.Options.OptionsFactory<TOptions>.Create(string name)
    Microsoft.Extensions.Options.OptionsMonitor<TOptions>+<>c__DisplayClass11_0.<Get>b__0()
    System.Lazy<T>.ViaFactory(LazyThreadSafetyMode mode)
    System.Lazy<T>.ExecutionAndPublication(LazyHelper executionAndPublication, bool useDefaultConstructor)
    System.Lazy<T>.CreateValue()
    Microsoft.Extensions.Options.OptionsMonitor<TOptions>.Get(string name)
    Microsoft.AspNetCore.Authentication.AuthenticationHandler<TOptions>.InitializeAsync(AuthenticationScheme scheme, HttpContext context)
    Microsoft.AspNetCore.Authentication.AuthenticationHandlerProvider.GetHandlerAsync(HttpContext context, string authenticationScheme)
    Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
    Microsoft.AspNetCore.ResponseCompression.ResponseCompressionMiddleware.Invoke(HttpContext context)
    Microsoft.AspNetCore.Localization.RequestLocalizationMiddleware.Invoke(HttpContext context)
    OrchardCore.Diagnostics.DiagnosticsStartupFilter+<>c__DisplayClass3_0+<<Configure>b__1>d.MoveNext()
[...]

PR #9919, opened 2021-07-15
Null check on new OpenID Parameters setting

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/9907

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.OpenId/Configuration/OpenIdClientConfiguration.cs
+++ b/src/OrchardCore.Modules/OrchardCore.OpenId/Configuration/OpenIdClientConfiguration.cs
@@ -98,7 +98,7 @@ namespace OrchardCore.OpenId.Configuration
                 }
             }
 
-            if (settings.Parameters.Any())
+            if (settings.Parameters?.Any() == true)
             {
                 var parameters = settings.Parameters;
                 options.Events.OnRedirectToIdentityProvider = (context) =>


src/OrchardCore.Modules/OrchardCore.OpenId/Configuration/OpenIdClientConfiguration.cs
```diff
@@ -98,7 +98,7 @@ public void Configure(string name, OpenIdConnectOptions options)
                 }
             }
 
-            if (settings.Parameters.Any())
+            if (settings.Parameters?.Any() == true)
```
kevinchalet, line 101: You need more C# 9 love in your life:

```suggestion
            if (settings.Parameters is { Length: > 0 })
```

deanmarcussen: Would love to ;)

https://github.com/OrchardCMS/OrchardCore/blob/20d5554f78794d85acdb8b9e61a365f12b514b7b/src/OrchardCore.Build/OrchardCore.Commons.props#L9-L10

kevinchalet: Weird, I've never had any issue with that combination 😄 
Well, then... 👍🏻

jtkech: Maybe we could just use ;)

    <PropertyGroup>
      <LangVersion>9.0</LangVersion>
      <RazorLangVersion>Latest</RazorLangVersion>
    </PropertyGroup>

Will do a PR this night if it works

deanmarcussen: Worth a crack @jtkech, super sad

Where it fails (for me - Seb had it failing on his perf tests - and found somewhere in the docs that it isn't supported) is if you try and use a C# 9 feature in a razor page when building with core 3

PR #9927, opened 2021-07-16
Fix dev feed build number

deanmarcussen:
We renamed the preview action last night, which caused the run number to reset to 0
so the cloudsmith version number is now wrong.

This when merged will be build number 4
latest on cloudsmith is 16371 

4 + 16368 = 16372

w / any luck


deanmarcussen:
@agriffard needs to be merged before anything else or the number is wrong ;)

jptissot (approved):

PR #9928, opened 2021-07-17
Configure ApiExplorer at the tenant level.

### LINKED ISSUE #9921: Orchard can't discovery pages and controllers when IActionDescriptorCollectionProvider implementation is added to the host application
### Describe the bug
Adding default MVC services to the shell application results in controllers and pages not being found:

### To Reproduce
1. Make a new project targeting .NET 5 (could be any version)
2. Use the following code:

```C#
using Microsoft.AspNetCore.Builder;
using Microsoft.AspNetCore.Hosting;
using Microsoft.AspNetCore.Mvc.Abstractions;
using Microsoft.AspNetCore.Mvc.Infrastructure;
using Microsoft.Extensions.FileProviders;
using Microsoft.Extensions.Hosting;
using Microsoft.Extensions.Primitives;

Host.CreateDefaultBuilder(args).ConfigureWebHostDefaults(webbuilder =>
{
    webbuilder.Configure(app =>
    {
        app.UseOrchardCore();
    });
})
.ConfigureServices(services =>
{
    services.AddSingleton<IActionDescriptorCollectionProvider, Collection>();
    services.AddOrchardCms();
})
.Build()
.Run();

class Collection : ActionDescriptorCollectionProvider
{
    public override ActionDescriptorCollection ActionDescriptors { get; } = new(new ActionDescriptor[0], 0);

    public override IChangeToken GetChangeToken()
    {
        return NullChangeToken.Singleton;
    }
}
```
3. Run the main page
4. Get a 404

### Expected behavior
The modules should still be able to discover their controllers and pages regardless of what the shell application does.

This was discovered when we [added more default services](https://github.com/dotnet/aspnetcore/commit/2af293dc72625b865fa7fcad5aeb510df46386bd) to the main web host. Even though the change was [reverted](https://github.com/dotnet/aspnetcore/pull/34257), it can still break orchard if this method call is made or if a custom implementation is added.

Scanning the orchard code it look as though [this method](https://github.com/OrchardCMS/OrchardCore/blob/1ec2e9898ff37e49d4224732c8a3e0acb5ed2fa6/src/OrchardCore/OrchardCore/Shell/Builders/Extensions/ServiceProviderExtensions.cs#L15) needs to special case descriptors that can affect the tenant container. I don't have any expertise here though this was just me debugging to figure out why orchard broke on .NET 6.

jtkech:
Fixes #9921

### DIFF
--- a/src/OrchardCore/OrchardCore/Modules/Extensions/ServiceCollectionExtensions.cs
+++ b/src/OrchardCore/OrchardCore/Modules/Extensions/ServiceCollectionExtensions.cs
@@ -10,6 +10,7 @@ using Microsoft.AspNetCore.DataProtection.KeyManagement;
 using Microsoft.AspNetCore.DataProtection.XmlEncryption;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc.ApiExplorer;
 using Microsoft.AspNetCore.Mvc.Infrastructure;
 using Microsoft.AspNetCore.Mvc.Localization;
 using Microsoft.AspNetCore.Routing;
@@ -270,8 +271,8 @@ namespace Microsoft.Extensions.DependencyInjection
                 // Remove the related host singletons as they are not tenant aware.
                 var descriptorsToRemove = collection
                     .Where(sd => sd is ClonedSingletonDescriptor &&
-                        (sd.ServiceType == typeof(IActionDescriptorCollectionProvider)/* ||
-                        sd.ServiceType == typeof(IApiDescriptionGroupCollectionProvider)*/))
+                        (sd.ServiceType == typeof(IActionDescriptorCollectionProvider) ||
+                        sd.ServiceType == typeof(IApiDescriptionGroupCollectionProvider)))
                     .ToArray();
 
                 foreach (var descriptor in descriptorsToRemove)
@@ -279,8 +280,10 @@ namespace Microsoft.Extensions.DependencyInjection
                     collection.Remove(descriptor);
                 }
 
-                // TODO: Configure ApiExplorer at the tenant level.
-                // collection.AddEndpointsApiExplorer();
+#if NET6_0_OR_GREATER
+                // Configure ApiExplorer at the tenant level.
+                collection.AddEndpointsApiExplorer();
+#endif
             },
             order: Int32.MinValue + 100);
         }
@@ -363,7 +366,7 @@ namespace Microsoft.Extensions.DependencyInjection
 
             if (options.SameSite == SameSiteMode.None)
             {
-                if (string.IsNullOrEmpty(userAgent))
+                if (String.IsNullOrEmpty(userAgent))
                 {
                     return;
                 }
--- a/src/OrchardCore/OrchardCore/Modules/Extensions/ServiceCollectionExtensions.cs
+++ b/src/OrchardCore/OrchardCore/Modules/Extensions/ServiceCollectionExtensions.cs
@@ -8,10 +8,13 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.DataProtection;
 using Microsoft.AspNetCore.DataProtection.KeyManagement;
 using Microsoft.AspNetCore.DataProtection.XmlEncryption;
+using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc.Infrastructure;
 using Microsoft.AspNetCore.Mvc.Localization;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Hosting;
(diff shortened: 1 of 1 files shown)


src/OrchardCore/OrchardCore/Modules/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -251,7 +256,36 @@ private static void AddRouting(OrchardCoreBuilder builder)
...
+                    .ToArray();
+
+                foreach (var descriptor in descriptorsToRemove)
+                {
+                    collection.Remove(descriptor);
+                }
+
+#if NET6_0_OR_GREATER
+                // Configure ApiExplorer at the tenant level.
+                collection.AddEndpointsApiExplorer();
```
sebastienros, line 285: This is enabling the api explorer by default. Can't we just do that if a service was previously registered in the `collection`? This way we only enable it at the tenant level if it was invoked by the main Startup class of the user's app.

jtkech: I did this because it will be called automatically by the web host, just after `AddRouting()`. It will register the 2 default singletons of `IActionDescriptorCollectionProvider` and `IApiDescriptionGroupCollectionProvider`, see the comment below => `Try to add default services in case MVC services aren't added.`

    public static IServiceCollection AddEndpointsApiExplorer(this IServiceCollection services)
    {
      // Try to add default services in case MVC services aren't added.
      services.TryAddSingleton<IActionDescriptorCollectionProvider, DefaultActionDescriptorCollectionProvider>();
      services.TryAddSingleton<IApiDescriptionGroupCollectionProvider, ApiDescriptionGroupCollectionProvider>();


We can still register a custom implementation at the host level, will be removed here but just at the tenant level. Then it's up to the user to register a custom implementation at the tenant level by using a module startup or our app helper. Then, as we do for `AddRouting()`, here we call `AddEndpointsApiExplorer()` that adds the defaults singletons.

davidfowl: This seems too special, can't we do something like make a fake service collection and add a bunch of services we don't want to chain back to the host application then special case those?

jtkech: @davidfowl Thanks for the feedback, yes I thought about it

First some contextual info, yes we build a child container and pipeline for each tenant, a tenant is composed of modules that have their own startup so that they can collaborate to build a tenant container and pipeline, and we have helpers allowing the application to register some delegates to do the same collaboration at the tenant level.

- Yes, the first time we filtered some services we did it directly when we create a tenant child container, it was for the host `IStartupFilter` to prevent them to re-add middlewares when we build a given tenant pipeline. https://github.com/OrchardCMS/OrchardCore/blob/9d522b4895f320949e0c97992b03a14245fac0dd/src/OrchardCore/OrchardCore/Shell/Builders/Extensions/ServiceProviderExtensions.cs#L23

- Most of the time we only need to re-register / remove some services, **for now we did it step by step on each new issue related to a specific `AddSomething()`**, this in different modules startups or with helpers from the app.

- The problem is mainly related to host singletons / options, particularly when they hold global collections. One problem is when a service is registered with an implementation instance whose type is not public, in that case we use a fake collection on which we do the same `AddSomething()` to retrieve the implementation types.

- E.g. when we migrated from 2.0 to 3.0 with the routing services, since 3.0 the route `IOptions` is mutated each time `UseEndpoints()` is called, but our tenant pipelines may be built and then call `UseEndpoints()` concurrently, so we needed isolated ones per tenant. See the following and other `AddSomething()` in the same file. https://github.com/OrchardCMS/OrchardCore/blob/9d522b4895f320949e0c97992b03a14245fac0dd/src/OrchardCore/OrchardCore/Modules/Extensions/ServiceCollectionExtensions.cs#L242

Just to say that we already did it in many places in modules or with helpers from the app and for now most of the time related to different specific `AddSomething()`, that's why here I also did it specifically. But yes, maybe better to do it in a more central place e.g. when we first create a given tenant child container.

That said we may not be able to do it in only one place, the library where we create tenant containers may not reference some needed packages to use some service types and `AddSomething()` (e.g. with `OpenIddict`). As soon as I have time I will focus on it and will try to filter services in a a more centralized way, thanks for the suggestion.

kevinchalet: > That said we may not be able to do it in only one place, the library where we create tenant containers may not reference some needed packages to use some service types and AddSomething() (e.g. with OpenIddict). 

If you want to discuss about potential ways to improve things in OpenIddict, let me know.

jtkech: @kevinchalet Okay cool!

When I will have time to work on it, I will let you know, thanks.

sebastienros:
This doesn't build for me on net60, maybe I need to update my sdk

davidfowl:
I think the build is broken right now.

sebastienros:
I updated to 6.0-preview6 and it's building.

sebastienros:
And got it building in preview7, with some small changes I will submit

PR #9929, opened 2021-07-17
FlowPart--always show "widget" classes

infofromca:


sebastienros:
Can you create an issue describing the scenario where the issue happen? The change doesn't look wrong, but I'd rather understand when it happens and if that makes sense.

ns8482e:
`widget` classes are added for widget content type -  not related to `FlowmeataData` - IMHO this change is not required

PR #9931, opened 2021-07-17
Running Console Log on window load

### LINKED ISSUE #9477: console_log filter should render in the FootScript
When using a framework such as Vue, rendering a script inline can cause console errors, we should probably inject the script at the FootScript zone

ns8482e:
Running Console Log on `window.load`
Fixes #9477


jptissot:
Won't fix the issue I am having. The issue is that in VueJs, you cannot have scripts in the middle of the html. This inline script should be placed in the FootScripts section.

ns8482e:
IMHO `console_log` is diagnostic related feature and not functional feature. So It would be better if that renders on some sort of predefined `Diagnostic` zone that gets added and rendered only if `OrchardCore.Diagnostics` module is enabled.

deanmarcussen:
Why don’t you just register it as a footscript with the resource manager?

Issue #9941, opened 2021-07-19
Check SubResourceIntegrity


agriffard:
In order to avoid incorrect SRIs, we should find a way to check the resources that have cdn versions declared.

Ex:
https://github.com/Dotnet-Boxed/Framework/blob/1fd6ecc5966d820166ca633bc1cefa91bbcb4369/Source/Boxed.AspNetCore.TagHelpers/SubresourceIntegrityTagHelper.cs#L243

deanmarcussen:
This should be as simple as having a test, or perhaps a console app, which discovers all registered resource manifests, checks if they have an integrity set, reads the file from the `src` and `debug-src` property, and compares it against the calculated one (from the code ag linked above).

Additionally it could output the correct one.

hishamco:
I will try to add a unit test for this one

hishamco:
I think we can calculate the SRI instead of entering them manually during the resource manifest registration

hishamco:
There's a related PR ready to merge, just waiting for Zoltan approval

PR #9947, opened 2021-07-21
Check SubResourceIntegrity

hishamco:
Address #9941

### DIFF
--- a/test/OrchardCore.Tests/Modules/OrchardCore.Resources/SubResourceIntegrityTests.cs
+++ b/test/OrchardCore.Tests/Modules/OrchardCore.Resources/SubResourceIntegrityTests.cs
@@ -1,0 +1,97 @@
+using System;
+using System.IO;
+using System.Linq;
+using System.Security.Cryptography;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.Hosting;
+using Moq;
+using OrchardCore.Environment.Shell;
+using OrchardCore.ResourceManagement;
+using OrchardCore.Resources;
+using OrchardCore.Settings;
+using Xunit;
+
+namespace OrchardCore.Tests.Modules.OrchardCore.Resources
+{
+    public class SubResourceIntegrityTests
+    {
+        private const string WebRoot = "wwwroot";
+
+        private string RootPath => new DirectoryInfo("../../../../../").FullName;
+
+        [Fact]
+        public void CheckSubResourceIntegrity()
+        {
+            // Arrange
+            var orchardCoreResourcesPath = Path.Combine(RootPath, @"src\OrchardCore.Modules\OrchardCore.Resources");
+            var siteServiceMock = new Mock<ISiteService>();
+            siteServiceMock
+                .Setup(s => s.GetSiteSettingsAsync())
+                .Returns(Task.FromResult<ISite>(new SiteSettings { ResourceDebugMode = ResourceDebugMode.Enabled }));
+            var hostEnvironmentMock = new Mock<IHostEnvironment>();
+            hostEnvironmentMock
+                .Setup(e => e.ContentRootFileProvider)
+                .Returns(new PhysicalFileProvider(Path.Combine(orchardCoreResourcesPath, WebRoot)));
+            var httpContextAccessorMock = new Mock<IHttpContextAccessor>();
+            httpContextAccessorMock
+                .Setup(a => a.HttpContext)
+                .Returns(new DefaultHttpContext());
+            var shellSettings = new ShellSettings();
+            var configurationOptions = new ResourceManagementOptionsConfiguration(
+                siteServiceMock.Object,
+                hostEnvironmentMock.Object,
+                httpContextAccessorMock.Object,
+                shellSettings
+            );
+            var resourceManagementOptions = new ResourceManagementOptions();
+
+            // Act & Assert
+            configurationOptions.Configure(resourceManagementOptions);
+
+            var resourceManifest = resourceManagementOptions.ResourceManifests.First();
+
+            ValidateSubResourceIntegrity("script");
+            ValidateSubResourceIntegrity("style");
+
+            void ValidateSubResourceIntegrity(string resourceType)
+            {
+                var orchardCoreResourceFolder = "~/OrchardCore.Resources";
+                var contentRootProvider = hostEnvironmentMock.Object.ContentRootFileProvider;
+                foreach (var resource in resourceManifest.GetResources(resourceType))
+                {
+                    foreach (var resourceDefinition in resource.Value)
+                    {
+                        if (!String.IsNullOrEmpty(resourceDefinition.CdnIntegrity))
+                        {
+                            var resourcePath = contentRootProvider.GetFileInfo(resourceDefinition.Url[orchardCoreResourceFolder.Length..]).PhysicalPath;
+                            var resourceIntegrity = GetSubResourceIntegrity(resourcePath);
+                            Assert.Equal(resourceIntegrity, resourceDefinition.CdnIntegrity);
+                        }
+
+                        if (!String.IsNullOrEmpty(resourceDefinition.CdnDebugIntegrity))
+                        {
+                            var resourceDebugPath = contentRootProvider.GetFileInfo(resourceDefinition.UrlDebug[orchardCoreResourceFolder.Length..]).PhysicalPath;
+                            var resourceDebugintegrity = GetSubResourceIntegrity(resourceDebugPath);
+                            Assert.Equal(resourceDebugintegrity, resourceDefinition.CdnDebugIntegrity);
+                        }
+                    }
+                }
+            }
+        }
+
+        private static string GetSubResourceIntegrity(string resourcePath)
+        {
+            var resourceIntegrity = String.Empty;
+            var resourceBytes = File.ReadAllBytes(resourcePath);
+            using (var sha384 = SHA384.Create())
+            {
+                var hash = sha384.ComputeHash(resourceBytes);
+                resourceIntegrity = "sha384-" + Convert.ToBase64String(hash);
+            }
+
+            return resourceIntegrity;
+        }
+    }
+}


hishamco:
@agriffard @deanmarcussen what I notice that many files in my local don't have the same SRI, for example `~/OrchardCore.Resources/Scripts/js.cookie.js` so when I download that file from CDN I got the same exact SRI

Could you please let me know if there's something I can do? That's why the Test fails in both my end and GitHub

hishamco:
The following are the file that have the exact same SRI:

- `~/OrchardCore.Resources/Scripts/jquery-ui-i18n.min.js`
- `~/OrchardCore.Resources/Scripts/vue-multiselect.min.js`
- `~/OrchardCore.Resources/Scripts/vue-multiselect.min.js`

deanmarcussen:
I would suggest picking a specific file that doesn't match, and using the online https://www.srihash.org/ that I suspect @agriffard is using to test why one is not working, perhaps they are using different hashing methods.

hishamco:
@agriffard are using `SHA384` when you generate the hash in SRI? Or are you using another one?

agriffard:
> @agriffard are using `SHA384` when you generate the hash in SRI? Or are you using another one?

Yes, I do. I always use https://www.srihash.org/ and the default encryption is `SHA-384`.

hishamco:
@agriffard how can I ensure that all the resources in my side are up to date, so I can validate the SRI for them

hishamco:
FYI I tried `npm run` but I'm getting an error

hishamco:
@agriffard could you let me know what should I do here to make a successful restore for npm packages. I need to finalize this PR

agriffard:
> @agriffard could you let me know what should I do here to make a successful restore for npm packages. I need to finalize this PR

`npm install`

hishamco:
NPM

Skrypt:
@hishamco  Update your node.js and/or npm to latest LTS version.

hishamco:
I already updated Node JS, I need to check again, coz it success with UI Testing project .. I will have a try

hishamco:
Gulp

Skrypt:
Once you installed the new node.js and npm try and do a cleanup of the node_modules folder. Then redo npm install.

hishamco:
@Skrypt is `package-lock.json` play an important role, what I notice is when I download the exact file from the CDN the SRI validated successfully. This make me think that some of my local assest are not updated

I tried `npm run rebuild` with no luck :(

hishamco:
@agriffard this PR is ready, I know the build will fail because I discover there are many invalid SRI, so I will create another PR to fix them. Once you merge the other PR we could rebase on main then the build will pass

agriffard:
I cancelled the workflow, the build was running without progress for 1h40.

hishamco:
That's why I didn't the build failed, I might need to check the logs

hishamco:
Seems the resources take time to validate the SRI, which is weird!!

Piedone:
Is this something you'd like to revisit any time soon @hishamco? Would be quite useful.

hishamco:
Sure, there are many PRs on my list :)

hishamco:
Seems the old issue in the testing has been resolved https://github.com/OrchardCMS/OrchardCore/pull/9947#issuecomment-1568750419

@Piedone I think we're ready to merge this now

test/OrchardCore.Tests/Modules/OrchardCore.Resources/SubResourceIntegrityTests.cs
```diff
@@ -0,0 +1,73 @@
...
+            var resourceOptions = Options.Create(new ResourceOptions());
+            var httpContextAccessorMock = new Mock<IHttpContextAccessor>();
+            httpContextAccessorMock
+                .Setup(a => a.HttpContext)
+                .Returns(new DefaultHttpContext());
+            var configurationOptions = new ResourceManagementOptionsConfiguration(
+                resourceOptions,
+                Mock.Of<IHostEnvironment>(),
+                httpContextAccessorMock.Object
+            );
```
Piedone, line 22:
```suggestion
                httpContextAccessorMock.Object);
```

test/OrchardCore.Tests/Modules/OrchardCore.Resources/SubResourceIntegrityTests.cs
```diff
@@ -0,0 +1,73 @@
...
+            var resourceManagementOptions = new ResourceManagementOptions();
+
+            // Act & Assert
+            configurationOptions.Configure(resourceManagementOptions);
+
+            var resourceManifest = resourceManagementOptions.ResourceManifests.First();
```
Piedone, line 28:
```suggestion
            // Act
            configurationOptions.Configure(resourceManagementOptions);
            
            // Assert
            var resourceManifest = resourceManagementOptions.ResourceManifests.First();
```

test/OrchardCore.Tests/Modules/OrchardCore.Resources/SubResourceIntegrityTests.cs
```diff
@@ -0,0 +1,73 @@
...
+                    foreach (var resourceDefinition in resource.Value)
+                    {
+                        if (!String.IsNullOrEmpty(resourceDefinition.CdnIntegrity) && !String.IsNullOrEmpty(resourceDefinition.UrlCdnDebug))
+                        {
+                            var resourceIntegrity = GetSubResourceIntegrityAsync(resourceDefinition.UrlCdnDebug).Result;
+
+                            Assert.True(resourceIntegrity.Equals(resourceDefinition.CdnDebugIntegrity), $"The {resourceType} {resourceDefinition.UrlCdnDebug} has invalid SRI hash, please use '{resourceIntegrity}' instead.");
+                        }
+
+                        if (!String.IsNullOrEmpty(resourceDefinition.CdnIntegrity) && !String.IsNullOrEmpty(resourceDefinition.UrlCdn))
```
Piedone, line 46:
```suggestion
                        if (!string.IsNullOrEmpty(resourceDefinition.CdnIntegrity) && !string.IsNullOrEmpty(resourceDefinition.UrlCdnDebug))
                        {
                            var resourceIntegrity = GetSubResourceIntegrityAsync(resourceDefinition.UrlCdnDebug).Result;

                            Assert.True(resourceIntegrity.Equals(resourceDefinition.CdnDebugIntegrity), $"The {resourceType} {resourceDefinition.UrlCdnDebug} has invalid SRI hash, please use '{resourceIntegrity}' instead.");
                        }

                        if (!string.IsNullOrEmpty(resourceDefinition.CdnIntegrity) && !string.IsNullOrEmpty(resourceDefinition.UrlCdn))
```

test/OrchardCore.Tests/Modules/OrchardCore.Resources/SubResourceIntegrityTests.cs
```diff
@@ -0,0 +1,73 @@
+using System.Security.Cryptography;
+using OrchardCore.ResourceManagement;
+using OrchardCore.Resources;
+
+namespace OrchardCore.Tests.Modules.OrchardCore.Resources
+{
+    public class SubResourceIntegrityTests
+    {
+        [Fact]
+        public void CheckSubResourceIntegrity()
```
Piedone, line 10:
```suggestion
        public void SavedSubResourceIntegritiesShouldMatchCurrentResources()
```

test/OrchardCore.Tests/Modules/OrchardCore.Resources/SubResourceIntegrityTests.cs
```diff
@@ -0,0 +1,73 @@
...
+            {
+                foreach (var resource in resourceManifest.GetResources(resourceType))
+                {
+                    foreach (var resourceDefinition in resource.Value)
+                    {
+                        if (!String.IsNullOrEmpty(resourceDefinition.CdnIntegrity) && !String.IsNullOrEmpty(resourceDefinition.UrlCdnDebug))
+                        {
+                            var resourceIntegrity = GetSubResourceIntegrityAsync(resourceDefinition.UrlCdnDebug).Result;
+
+                            Assert.True(resourceIntegrity.Equals(resourceDefinition.CdnDebugIntegrity), $"The {resourceType} {resourceDefinition.UrlCdnDebug} has invalid SRI hash, please use '{resourceIntegrity}' instead.");
```
Piedone, line 43: Some line breaks here would be nice so it's not that long of a line, like along arguments. Below too.

test/OrchardCore.Tests/Modules/OrchardCore.Resources/SubResourceIntegrityTests.cs
```diff
@@ -0,0 +1,73 @@
...
+                        }
+
+                        if (!String.IsNullOrEmpty(resourceDefinition.CdnIntegrity) && !String.IsNullOrEmpty(resourceDefinition.UrlCdn))
+                        {
+                            var resourceIntegrity = GetSubResourceIntegrityAsync(resourceDefinition.UrlCdn).Result;
```
Piedone, line 48: The test method can be async, `.Result` is not needed.

test/OrchardCore.Tests/Modules/OrchardCore.Resources/SubResourceIntegrityTests.cs
```diff
@@ -0,0 +1,73 @@
...
+                            Assert.True(resourceIntegrity.Equals(resourceDefinition.CdnIntegrity), $"The {resourceType} {resourceDefinition.UrlCdn} has invalid SRI hash, please use '{resourceIntegrity}' instead.");
+                        }
+                    }
+                }
+            }
+        }
+
+        private static async Task<string> GetSubResourceIntegrityAsync(string url)
+        {
+            var client = new HttpClient
```
Piedone, line 59: `HttpClient` [mustn't be instantiated frequently](https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-net-http-httpclient#instancing).

Piedone (changes requested):

test/OrchardCore.Tests/Modules/OrchardCore.Resources/SubResourceIntegrityTests.cs
hishamco: The following error shows up when we use the same instance

**System.InvalidOperationException : This instance has already started one or more requests. Properties can only be modified before sending the first**

hishamco: Now the tests passed after I revert the last changes, you could check the logs

Piedone: Please fix that first.

hishamco: I already did then revert the changes when the build failed. Let me do it again

Piedone:
Let me know if you'd like a review here.

test/OrchardCore.Tests/Modules/OrchardCore.Resources/SubResourceIntegrityTests.cs
Piedone: This is still the case.

Piedone (changes requested):
See https://github.com/OrchardCMS/OrchardCore/pull/9947#discussion_r1477471561

test/OrchardCore.Tests/Modules/OrchardCore.Resources/SubResourceIntegrityTests.cs
hishamco: There's something strange when using the await in this case, I need to check again

hishamco: @Piedone the reason why the test failed after using static is

https://github.com/dotnet/corefx/blob/b8b81a66738bb10ef0790023598396861d92b2c4/src/System.Net.Http/src/System/Net/Http/HttpClient.cs#L665-L672

Piedone: Please don't request a review until this is addressed. Check out the docs I linked.

Piedone: Until this works without instantiating `HttpClient` repeatedly, your code needs to be adjusted.

Piedone: You get this because the code sets the `BaseAddress` on `HttpClient`. You can't change this on an existing instance. However, you don't need to do this anyway, because in `GetByteArrayAsync()` you set the URL too. You only need to set it there, and reuse the instance.

`HttpClient` can, of course, open multiple different requests with the same instance, but not if you set `BaseAddress`.

hishamco:
@Piedone time to merge :)

test/OrchardCore.Tests/Modules/OrchardCore.Resources/SubResourceIntegrityTests.cs
```diff
@@ -6,6 +6,8 @@ namespace OrchardCore.Tests.Modules.OrchardCore.Resources;
 
 public class SubResourceIntegrityTests
 {
+    private static readonly HttpClient _httpClient = new();
```
Piedone, line 9: You also need to dispose this. It doesn't actually need to be static, you can just add a `using var` into the test method.

Piedone (changes requested):

test/OrchardCore.Tests/Modules/OrchardCore.Resources/SubResourceIntegrityTests.cs
hishamco: > You also need to dispose this

Agree

> It doesn't actually need to be static

The `GetSubResourceIntegrityAsync()` is `static`

Piedone: And you can change that, because you have the infinite power of Visual Studio at your hand :). However, I'm not arguing for that. You can just pass `HttpClient` into the method. The latest implementation of disposing a static field from an instance method is a wrong approach to this.

hishamco: >And you can change that, because you have the infinite power of Visual Studio at your hand

Sure :)

Piedone: I committed what I think is simpler, please check.

test/OrchardCore.Tests/Modules/OrchardCore.Resources/SubResourceIntegrityTests.cs
```diff
@@ -25,10 +23,11 @@ public async Task SavedSubResourceIntegritiesShouldMatchCurrentResources()
 
         // Act
         configurationOptions.Configure(resourceManagementOptions);
-        
+
         // Assert
         var resourceManifest = resourceManagementOptions.ResourceManifests.First();
 
+        using var httpClient = new HttpClient();
```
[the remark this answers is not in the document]
hishamco: No problem, I think we are fine now for merge this, right?

Piedone (approved):

test/OrchardCore.Tests/Modules/OrchardCore.Resources/SubResourceIntegrityTests.cs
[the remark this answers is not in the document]
Piedone: If you're OK with it then yes.

sebastienros:
Use the new static instances instead of creating new ones: https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.sha384.hashdata?view=net-8.0

sebastienros:
I am right that these will run also locally?
If so would you want to create an XUnit attribute to add on the methods which would skip the test if the "CI" environment variable is defined? 

```
public class CiOnlyFactAttribute : FactAttribute
{
    public override string Skip
    {
        get
        {
            // "CI" is defined by GitHub actions
            // "BUILD_BUILDID" is defined by Azure DevOps

            if (Environment.GetEnvironmentVariable("BUILD_BUILDID") != null || 
                Environment.GetEnvironmentVariable("CI") != null)
            {
                return $"nameof(CiOnlyFactAttribute) tests are not run locally. To run them locally create a \"CI\" environment variable.";
            }

            return null!;
        }
    }
}
```

hishamco:
We could do that, but I'm not sure do we still need Azure DevOps, while we're using GitHub action?

Piedone:
Why would we skip these in CI?

sebastienros:
The opposite, skip these locally but not in CI. My wording was wrong, the code is correct ;)

hishamco:
I thought what the Seb meant at the first glance 😂, I will submit a PR

hishamco:
> The opposite, skip these locally but not in CI. My wording was wrong, the code is correct ;)

Seb seems the code is wrong, because the fact should be `null` when the environment valriable exists not the opposite :)

PR #9975, opened 2021-07-25
Update README.md

matiasmolleja:
As we are targetting net5.0 and netcoreapp3.1 running 'dotnet run' without passing the framework parameter results in an error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Here is a more detailed [roadmap](https://github.com/OrchardCMS/OrchardCore/wiki
 
 - Install the latest version of the .NET SDK from this page <https://dotnet.microsoft.com/download>
 - Next, navigate to `D:\OrchardCore\src\OrchardCore.Cms.Web` or wherever your folder is on the commandline in Administrator mode.
-- Call `dotnet run`.
+- Call `dotnet run -f net5.0`.
 - Then open the `http://localhost:5000` URL in your browser.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,7 +38,7 @@ Here is a more detailed [roadmap](https://github.com/OrchardCMS/OrchardCore/wiki
 
 - Install the latest version of the .NET SDK from this page <https://dotnet.microsoft.com/download>
 - Next, navigate to `D:\OrchardCore\src\OrchardCore.Cms.Web` or wherever your folder is on the commandline in Administrator mode.
-- Call `dotnet run`.
+- Call `dotnet run -f net5.0`.
```
hishamco, line 41: But this will target `net5.0` only, can we make it

- Call `dotnet run -f {targetFramework}`.

matiasmolleja: It's just that I would prefer to keep instructions as precise as possible. If the user follow the 3 points above, the framework will be 5.0. 
And if we use the word  '{targeFramework}' we will force the user to guess, or to look at the source code to know which versions we support.

But I see your point. 
What about this?

'Call `dotnet run -f net5.0`  (or `dotnet run -f netcoreapp3.1` depending on your version of .NET SDK).

hishamco (approved):
LGTM

agriffard:
Good to see you back @matiasmolleja

PR #9976, opened 2021-07-25
Flow alignment icon always render as left aligned.

microposmp:


sebastienros:
Can you share a screenshot where it's on the right?

microposmp:
Sorry, the title was not clear enough.

It's the icon in the picture below that always renders as if the setting is left aligned after saving/publishing.

image

The change should make it render as the selected alignment. Now it's not possible to see what alignment is set without expanding the dropdown. Now it always shows as left aligned.

Skrypt (approved):

PR #9985, opened 2021-07-27
Not use a current culture when checking invalid autoroute path

aaronamm:
# Existing issue:
If you enable Localization module and set a default culture to non en-US culture, e.g. th-TH, you will get a validation error when trying to save a content item with an Autoroute part.

Here is the screenshot when trying to save a content item and a default culture is `th-TH`.

image

Note, you can switch to any culture by appending culture=<culture-name> query string to a current URL.

# Root cause
- This issue is from [AutoroutePartExtensions.ValidatePathFieldValue method](https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore.Modules/OrchardCore.Autoroute/Models/AutoroutePartExtensions.cs#L17)
- Inside the method, there is a call of `autoroute.Path?.IndexOf("//")`  which returns 0 instead of -1 when a current culture is `th-TH`.
- According to the [MS document](https://docs.microsoft.com/en-us/dotnet/api/system.string.indexof?view=net-5.0#System_String_IndexOf_System_String_), `IndexOf(String)` search a word by using culture and case sensitive.
- Please check `IndexOf` test cases that I've made https://github.com/codesanook/codesanook-examples/blob/master/Codesanook.Examples.DotNet/Localization/LocalizationTest.cs#L24. 

# How this PR fixes
- Refer to [Recommendations for string usage
](https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings#recommendations-for-string-usage), we should use StringComparison.Ordinal or StringComparison.OrdinalIgnoreCase for the non-linguistic string search.
- There is no case in `a space` and `//`, so I pick `StringComparison.Ordinal` and to match IndexOf which is `case-sensitive`.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Autoroute/Models/AutoroutePartExtensions.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Autoroute/Models/AutoroutePartExtensions.cs
@@ -18,8 +18,7 @@ namespace OrchardCore.Autoroute.Models
             if (HasInvalidCharacters(autoroute.Path))
             {
                 var invalidCharactersForMessage = string.Join(", ", AutoroutePart.InvalidCharactersForPath.Select(c => $"\"{c}\""));
-                const string errorMessage = "Please do not use any of the following characters in your permalink: {0}. No spaces, or consecutive slashes, are allowed (please use dashes or underscores instead).";
-                yield return new ValidationResult(S[errorMessage, invalidCharactersForMessage], new[] { nameof(autoroute.Path) });
+                yield return new ValidationResult(S["Please do not use any of the following characters in your permalink: {0}. No spaces, or consecutive slashes, are allowed (please use dashes or underscores instead).", invalidCharactersForMessage], new[] { nameof(autoroute.Path) });
             }
 
             if (autoroute.Path?.Length > AutoroutePart.MaxPathLength)
@@ -30,9 +29,21 @@ namespace OrchardCore.Autoroute.Models
 
         private static bool HasInvalidCharacters(string path)
         {
-            if (path?.IndexOfAny(AutoroutePart.InvalidCharactersForPath) > -1) return true; // IndexOfAny performs culture-insensitive search.
-            if (path?.IndexOf(' ', StringComparison.Ordinal) > -1) return true;
-            if (path?.IndexOf("//", StringComparison.Ordinal) > -1) return true;
+            // IndexOfAny performs culture-insensitive and case-sensitive search.
+            if (path?.IndexOfAny(AutoroutePart.InvalidCharactersForPath) > -1)
+            {
+                return true;
+            }
+
+            if (path?.IndexOf(' ', StringComparison.Ordinal) > -1)
+            {
+                return true;
+            }
+
+            if (path?.IndexOf("//", StringComparison.Ordinal) > -1)
+            {
+                return true;
+            }
 
             return false;
         }
--- a/src/OrchardCore.Modules/OrchardCore.Autoroute/Models/AutoroutePartExtensions.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Autoroute/Models/AutoroutePartExtensions.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.Autoroute/Models/AutoroutePartExtensions.cs
```diff
@@ -14,16 +15,26 @@ public static IEnumerable<ValidationResult> ValidatePathFieldValue(this Autorout
                 yield return new ValidationResult(S["Your permalink can't be set to the homepage, please use the homepage option instead."], new[] { nameof(autoroute.Path) });
             }
 
-            if (autoroute.Path?.IndexOfAny(AutoroutePart.InvalidCharactersForPath) > -1 || autoroute.Path?.IndexOf(' ') > -1 || autoroute.Path?.IndexOf("//") > -1)
+            if (HasInvalidCharacters(autoroute.Path))
             {
                 var invalidCharactersForMessage = string.Join(", ", AutoroutePart.InvalidCharactersForPath.Select(c => $"\"{c}\""));
-                yield return new ValidationResult(S["Please do not use any of the following characters in your permalink: {0}. No spaces, or consecutive slashes, are allowed (please use dashes or underscores instead).", invalidCharactersForMessage], new[] { nameof(autoroute.Path) });
+                const string errorMessage = "Please do not use any of the following characters in your permalink: {0}. No spaces, or consecutive slashes, are allowed (please use dashes or underscores instead).";
```
deanmarcussen, line 21: doing this means it will no longer be localized. please do that as it was before.

src/OrchardCore.Modules/OrchardCore.Autoroute/Models/AutoroutePartExtensions.cs
```diff
@@ -14,16 +15,26 @@ public static IEnumerable<ValidationResult> ValidatePathFieldValue(this Autorout
...
 
             if (autoroute.Path?.Length > AutoroutePart.MaxPathLength)
             {
                 yield return new ValidationResult(S["Your permalink is too long. The permalink can only be up to {0} characters.", AutoroutePart.MaxPathLength], new[] { nameof(autoroute.Path) });
             }
         }
+
+        private static bool HasInvalidCharacters(string path)
+        {
+            if (path?.IndexOfAny(AutoroutePart.InvalidCharactersForPath) > -1) return true; // IndexOfAny performs culture-insensitive search.
```
deanmarcussen, line 33:
```suggestion
            // IndexOfAny performs culture-insensitive search.
            if (path?.IndexOfAny(AutoroutePart.InvalidCharactersForPath) > -1) 
            {
                return true;
            }
```

src/OrchardCore.Modules/OrchardCore.Autoroute/Models/AutoroutePartExtensions.cs
```diff
@@ -14,16 +15,26 @@ public static IEnumerable<ValidationResult> ValidatePathFieldValue(this Autorout
...
+            if (path?.IndexOf(' ', StringComparison.Ordinal) > -1) return true;
+            if (path?.IndexOf("//", StringComparison.Ordinal) > -1) return true;
```
deanmarcussen, line 35: as above with the formatting thanks

deanmarcussen:
This would very much benefit from a unit test with the issue you have demonstrated

src/OrchardCore.Modules/OrchardCore.Autoroute/Models/AutoroutePartExtensions.cs
aaronamm: @deanmarcussen I pull this error message to a variable for better code formatting.
However, I think with this code `S[errorMessage, invalidCharactersForMessage]` it should be get translated as it was before.

Please correct me if I am wrong for this.

src/OrchardCore.Modules/OrchardCore.Autoroute/Models/AutoroutePartExtensions.cs
aaronamm: @deanmarcussen code for early exit I prefer to use if without a curry brace. 
If without a curry brace has some drawback but I think it is okay for this situation.
However, as your suggestion, I will change.

aaronamm:
@deanmarcussen Thanks for your prompt reply. I have a question for you if translation still works.
After you have answered the question, I will fix this PR from your feedback.
Thanks.

src/OrchardCore.Modules/OrchardCore.Autoroute/Models/AutoroutePartExtensions.cs
aaronamm: @deanmarcussen  Here is my test result with `th-TH` culture.
The translation still works.

image

Here is a content in th-TH.po.
```
msgid "Please do not use any of the following characters in your permalink: {0}. No spaces, or consecutive slashes, are allowed (please use dashes or underscores instead)."
msgstr "กรุณาอย่าใช้ตัวอักษรดังต่อไปนี้ใน permalink : {0}. ห้ามใส่ช่องว่าง, consecutive slashes (กรุณาใช้ dash หรือ underscores แทน)"
```

deanmarcussen: Please revert the change to the localization. It will not work as the poextractor relies on the string being provided in the S accessor.

aaronamm: @deanmarcussen I have already reverted the localization message, please check all changes to this PR.
https://github.com/OrchardCMS/OrchardCore/pull/9985/files

I think is ready to be merged.
Thank you so much.

PR #9994, opened 2021-07-29
Reduce allocations in CamelFriendly

lahma:
* Tested with agency and blog theme
* Using `ZString` instead of `new StringBuilder`
* Using "normal algorithm" which start from beginning and no need to shuffle data in array (insert vs append)
* Rider was complaining about duplicate reference `OrchardCore.Data.Abstraction` in `OrchardCore.Alias.csproj` so fixed that too

### DIFF
--- a/src/OrchardCore/OrchardCore.Mvc.Core/Utilities/StringExtensions.cs
+++ b/src/OrchardCore/OrchardCore.Mvc.Core/Utilities/StringExtensions.cs
@@ -21,11 +21,6 @@ namespace OrchardCore.Mvc.Utilities
                 return "";
             }
 
-            if (camel == "RawHtml")
-            {
-                return "Raw Html";
-            }
-
             using var sb = ZString.CreateStringBuilder();
             for (var i = 0; i < camel.Length; ++i)
             {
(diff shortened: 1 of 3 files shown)


src/OrchardCore/OrchardCore.Mvc.Core/Utilities/StringExtensions.cs
```diff
@@ -14,19 +15,26 @@ public static class StringExtensions
...
         public static string CamelFriendly(this string camel)
         {
+            // optimize common cases
             if (string.IsNullOrWhiteSpace(camel))
             {
                 return "";
             }
 
-            var sb = new StringBuilder(camel);
+            if (camel == "RawHtml")
```
sebastienros, line 24: Do you where this one is coming from exactly?

sebastienros: @deanmarcussen found it, it's from the Widget name that is included as a class in the default template. That's border-line. An mru cache would be better in that case since it's only because you used The Agency template apparently, and each site will have different common cases.

lahma: OK, I don't start the dance with MRU as it will need to have locking etc, so removed the special case and now always falls to building the string, should be quite good as ZString will use array pool anyway.

deanmarcussen (approved):

PR #10008, opened 2021-08-01
Sonar Integration

Jetski5822:


Skrypt:
Impossible that we are doing bad code right? 😉

Jetski5822:
HA - Get ready!

Jetski5822:
hmm, sonarcloud ups the build by about 30 minutes :/

Skrypt:
Should probably set as a Github workflow that we can trigger manually.

Skrypt:
Merge latest main branch here.

Jetski5822:
yeah maybe - it depends on how much we care about the build time...

Skrypt:
> yeah maybe - it depends on how much we care about the build time...

We care, it can't be that long. We switched to Github Actions because Appveyor was taking too long.
We sometimes need to actually debug the CI and waiting 40 minutes between builds will become a problem.

jptissot:
You can also have it setup on a cron schedule like this https://github.com/StatCan/StatCan.OrchardCore/blob/master/.github/workflows/codeql-analysis.yml#L11

Skrypt:
I actually preferred seeing actual bugs to fix.

Jetski5822:
ah - its because i changed the shallow depth

Jetski5822:
@Skrypt down to 22 minutes... I think the issue with the zero reported issues is that we havent analysed the main branch, so it has nothing to base anything off of

sebastienros:
image

To me it shows it doesn't work, right?

vg:
If you are trying to get the build time lower, how about using [Codacy](https://www.codacy.com) instead of Sonar. 
Codacy uses the same [Sonar c#](https://docs.codacy.com/getting-started/supported-languages-and-tools/) tool but runs in its own instance instead of using GitHub actions minutes.

Piedone:
Is this something you'd like to revisit any time soon @Jetski5822 or should we close?

I think a better approach is to add the Sonar analyzer package to the solution. That way, you can see violations (as warnings) locally too, with automatic fixes available. Then, the same analyzers would be used during CI builds too, but due to turning those warnings into errors, violations would also fail the build.

Also see https://github.com/OrchardCMS/OrchardCore/issues/7950 and https://github.com/OrchardCMS/OrchardCore/issues/12929.

Piedone:
Closing due to inactivity.

PR #10019, opened 2021-08-02
Expanded strongly typed example with update endpoint

brunoAltinet:
When trying to find good info about how updates work for a strongly typed part, this example would show up frequently but it hardly shows anything. I think this kind of an expansion gives some guidance on where to look when doing updates via code.

### DIFF
--- a/src/docs/reference/modules/ContentTypes/README.md
+++ b/src/docs/reference/modules/ContentTypes/README.md
@@ -192,6 +192,7 @@ public class ProductController : Controller
     [HttpPost("/api/product/{productId}/price/{price}")]
     public async Task<ContentValidateResult> UpdateProductPriceAsync(string productId, int price)
     {
+        //this call will only fetch published content item, which makes publishing after update redundant
         var product = _orchardHelper.GetContentItemByIdAsync(productId);
 
         if (product == null) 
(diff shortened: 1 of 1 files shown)


src/docs/reference/modules/ContentTypes/README.md
```diff
@@ -186,5 +188,26 @@ public class ProductController : Controller
...
+
+        var productPart = product.As<Product>();
+        productPart.Price.Value = price;
+        
+        product.Apply(productPart) //apply modified part to a content item
+        
+        await _contentManager.UpdateAsync(product); //update will fire handlers which could alter the content item.
+
+        //validation will cancel changes if product is not valid. It's fired after update since handlers could change the object.
+        return await _contentManager.ValidateAsync(product);
```
deanmarcussen, line 210: This probably needs a publishasync option shown as well, otherwise the item will be left in an unknown state

brunoAltinet: Looking at the code of DefaultContentManager, it basically calls save on _session and a bunch of handlers. So i think what it does it leaves the current state as is (published or draft). I think this is good enough for a sample. I can add a note about publishing though?

PR #10085, opened 2021-08-09
Fallback to the "Default" Picker Field Result Provider

### LINKED ISSUE #10084: Fallback to the "Default" Picker Field Result Provider
For the Content and User Picker fields, when you define a custom editor with a given name (and its related driver), you should also register a custom `IXxxPickerResultProvider` with the same name as the editor.

The problem is when you just want to have a custom editor / driver without changing the way that items are searched to populate the picker, so here I suggest to fallback to the default `IXxxPickerResultProvider` whose name is `Default`.

jtkech:
Fixes #10084 

For the Content and User Picker fields, when you define a custom editor with a given name (and its related driver), you should also register a custom `IXxxPickerResultProvider` with the same name as the editor.

The problem is when you just want to have a custom editor / driver without changing the way that items are searched to populate the picker, so here I suggest to fallback to the default `IXxxPickerResultProvider` whose name is `Default`.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Controllers/ContentPickerAdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Controllers/ContentPickerAdminController.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


deanmarcussen (commented):
Funny, normally I want the opposite - a different result provider, but the same editor shape (I often morph the shape back into the standard).

Maybe what we should do is have a field on the settings for which result provider to use (and when it's empty use Default), rather than just relying on the editor shape name. Would be easy enough to have a select list.

If you disagree I'm ok with this

jtkech:
@deanmarcussen 

Yes good idea, hmm but another setting to manage, in fact to not break would need by default to be the same name as the `Editor` but that is set with the `DisplayMode` at the part field level, hmm but then if not registered would be still useful to fallback to the default one ;)

So yes good idea but in the meantime maybe useful to use this PR.

deanmarcussen (approved):

PR #10118, opened 2021-08-15
Use of new StaticFileOptions for Modules

### LINKED ISSUE #2966: do we have plan to use Blazor?
it seems that after we use Blazor in client (webAssembly), we do not need Liquid.

https://blazor.net/docs/get-started.html

### LINKED ISSUE #6474: Blazor WebAssembly Hosted Template
I have created a new solution using the template in Visual Studio 2019 (Blazor App->Blazor WebAssembly App with ASP.NET Core hosed ticked), refernced `OrchardCore.Application.Mvc.Targets` in the Server project and added the tenants using the sample from https://github.com/OrchardCMS/OrchardCore.Samples/blob/master/MultiTenantApplication/appsettings.json

Moving `UseEndpoints` from `StartUp.Configure` to `OrchardCoreBuilder.Configure` allows me to inject `ShellSettings` into the constructor of `WeatherForecastController` and correctly retrieve tenant specific information such as `CustomTitle`.

When running the solution the Blazor client app can't resolve the page returns a 404 however https://localhost:44338/WeatherForecast and https://localhost:44338/customer-a/WeatherForecast work as expected.

My Setup.cs looks as follows:

````
public class Startup
  {
    public Startup(IConfiguration configuration)
    {
      Configuration = configuration;
    }

    public IConfiguration Configuration { get; }

    // This method gets called by the runtime. Use this method to add services to the container.
    // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
    public void ConfigureServices(IServiceCollection services)
    {
      services.AddControllersWithViews();
      services.AddRazorPages();

      services.AddOrchardCore().WithTenants().Configure(app =>
      {
        app.UseEndpoints(endpoints =>
        {
          endpoints.MapRazorPages();
          endpoints.MapControllers();
          endpoints.MapFallbackToFile("index.html");
        });
      });
    }

    // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
    public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
    {
      if (env.IsDevelopment())
      {
        app.UseDeveloperExceptionPage();
        app.UseWebAssemblyDebugging();
      }
      else
      {
        app.UseExceptionHandler("/Error");
        // The default HSTS value is 30 days. You may want to change this for production scenarios, see https://aka.ms/aspnetcore-hsts.
        app.UseHsts();
      }

      app.UseHttpsRedirection();
      app.UseBlazorFrameworkFiles();
      app.UseStaticFiles();

      app.UseRouting();



      app.UseOrchardCore();
    }
}
````

Is it possible to add multi tenant functionality to the Server project?

### LINKED ISSUE #6505: Serve static content of web application to all tenants
When static contents are added to web application `wwwroot ` which are not embedded in assembly - i.e. static content added by Blazor web assembly or  static css, javascript or images etc,  `app.UseStaticFiles()` only serves content for `Default` tenant.

In multi-tenant scenario, If  there is `'wwwroot/appfile1.html'` and have  `Default`, `tenant1`, `tenant2` then it should serve `appfile1.html` to all tenants relative to tenant prefix.

i.e. for default tenant `appfile1.html` should be served with url `/appfile1.html` and for `tenant1`and `tenant2` served with url `/tenant1/appfile1.html` and `/tenant2/appfile1.html`.

Ref issue # https://github.com/OrchardCMS/OrchardCore/issues/6474#issuecomment-649139631

/cc @jtkech

### PR DESCRIPTION
**Pull request by ns8482e**
Use of `new StaticFileOptions()` for module files, instead resolving from DI. 

~~Replace `app.UseStaticFiles(options);` to use  `app.UseStaticFiles();` and configuring Module's FileProviders in `PostConfigure`.~~

~~This allows external module ( e.g. Blazor Server) to inject any additional FileProviders in tenant's pipeline -  that ultimately used by StaticFileMiddleware.~~

Static Files from HostWebRoot can be served using `app.UseStaticFiles()` in tenant pipeline.

```cs
services.AddOrchardCore()
                .AddMvc()
                .WithTenants()
                .Configure((app, routes, serviceProvider) =>
                {
                    app.UseStaticFiles();
                });
```

Fixes #6505 
Fixes #2966 
Fixes #6474


willnationsdev:
lol, wow, I *just* started investigating the possibilities, ran into the issue, saw the workaround, and found it was being fixed soon. Gotta love open-source. :-D

src/OrchardCore/OrchardCore/Modules/Extensions/ConfigureModuleStaticFileOptions.cs
```diff
@@ -0,0 +1,69 @@
...
+                // Serve Tenant Static files from module and from HostWebRoot
+                options.FileProvider = new CompositeFileProvider(_fileProvider, _environment.WebRootFileProvider);
+            }
+            else
+            {
+                // Serve Tenant Static files only from module
+                options.FileProvider = _fileProvider;
+            }
+
+            var cacheControl = _shellConfiguration.GetValue("StaticFileOptions:CacheControl", $"public, max-age={TimeSpan.FromDays(30).TotalSeconds}, s-max-age={TimeSpan.FromDays(365.25).TotalSeconds}");
```
sebastienros, line 52: `s-maxage` ?

sebastienros: Since there are settings, maybe put these values in there too.

src/docs/reference/core/Modules/README.md
```diff
@@ -60,9 +60,12 @@ You can find a sample application here: [`OrchardCore.Mvc.Web`](../../../../Orch
 
 The following configuration values are used by default and can be customized:
 
+By default each tenant doesn't serve static files from Host `wwwroot`. To serve Host `wwwroot` for each tenant, set `ServeHostWebRoot` to `true` in tenant config.
```
sebastienros, line 63: Has it always been the case?

ns8482e: Currently  static files from Host `wwwroot` is only served to `Default` tenant  (and not to other tenants) by configuring `app.UseStaticFiles()` in Host startup.

Keeping the same behavior by setting default value to `false`

src/OrchardCore/OrchardCore/Modules/Extensions/ConfigureModuleStaticFileOptions.cs
ns8482e: looks like typo, will fix it.  It's reading from settings if available- the second param is default fallback

ns8482e: Updated code and docs to `s-maxage`

src/OrchardCore/OrchardCore/Modules/Extensions/ConfigureModuleStaticFileOptions.cs
```diff
@@ -0,0 +1,69 @@
...
+        private readonly IWebHostEnvironment _environment;
+        public ConfigureModuleStaticFileOptions(IModuleStaticFileProvider fileProvider,
+            IShellConfiguration shellConfiguration,
+            IWebHostEnvironment environment)
+        {
+            _fileProvider = fileProvider;
+            _shellConfiguration = shellConfiguration;
+            _environment = environment;
+        }
+        public void PostConfigure(string name, StaticFileOptions options)
```
jtkech, line 30: Add new line before

src/OrchardCore/OrchardCore/Modules/Extensions/ConfigureModuleStaticFileOptions.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            name = name ?? throw new ArgumentNullException(nameof(name));
+            options = options ?? throw new ArgumentNullException(nameof(options));
+
+            if (name != Microsoft.Extensions.Options.Options.DefaultName)
```
jtkech, line 35: Maybe add an using above so that here we could just use `Options.DefaultName`

src/OrchardCore/OrchardCore/Modules/Extensions/ConfigureModuleStaticFileOptions.cs
```diff
@@ -0,0 +1,69 @@
...
+            if (serveWebRoot)
+            {
+                // Serve Tenant Static files from module and from HostWebRoot
+                options.FileProvider = new CompositeFileProvider(_fileProvider, _environment.WebRootFileProvider);
+            }
+            else
+            {
+                // Serve Tenant Static files only from module
+                options.FileProvider = _fileProvider;
+            }
```
jtkech, line 50: Maybe we still need to set `options.RequestPath = ""`

src/OrchardCore/OrchardCore/Modules/Extensions/ConfigureModuleStaticFileOptions.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        private static void OnPrepareResponseCacheControl(StaticFileResponseContext ctx, string cacheControl, Action<StaticFileResponseContext> beforePrepare)
+        {
+            if (beforePrepare != null)
+            {
+                beforePrepare(ctx);
+            }
+
+            // Cache static files for a year as they are coming from embedded resources and should not vary
+            ctx.Context.Response.Headers[HeaderNames.CacheControl] = cacheControl;
```
jtkech, line 66: Hmm, but not fully true if we compose with the `WebRootFileProvider` that may be related to not embedded files.

Maybe we need 2 `UseStaticFiles()`, then maybe the same options but at least with 2 separate default values.

jtkech:
@ns8482e @sebastienros Just did a little review

Hmm, for content root files each tenant already compose with `ContentRootFileProvider` without any option https://github.com/OrchardCMS/OrchardCore/blob/ac7598ceaf54f5c5a5fed01e8fd0a2c3a7f2f3a3/src/OrchardCore/OrchardCore/Modules/Extensions/ApplicationBuilderExtensions.cs#L20-L22

Maybe at the tenant level we could compose with the `WebRootFileProvider` without needing an option for this.

Now not sure we need a post configuration, maybe jut a regular one, then as said in the review maybe a separate static file options as the `WebRootFileProvider` may be related to non embedded files.

Need a little more time to  focus on it again ;) I will complete my review asap.

ns8482e:
@jtkech 
> Maybe at the tenant level we could also compose with the `WebRootFileProvider` without needing an option for this.

The real issue is due to following lines,

https://github.com/OrchardCMS/OrchardCore/blob/ac7598ceaf54f5c5a5fed01e8fd0a2c3a7f2f3a3/src/OrchardCore/OrchardCore/Modules/Extensions/ServiceCollectionExtensions.cs#L210-L213

Line 210 evaluates `IOptions<StaticFileOptions>` outiside `StaticMiddleware` 
Line 213 replaces all framework provided `FileProviders`  resolved via `IOptions<StaticFileOptions>` including  `WebRootFileProvider`, `StaticWebAssetsFileProvider` and `ManifestEmbeddedFileProvider`

So even if a developer uses `UseStaticFiles()` inside tenant pipeline, the resolved FileProviers are always `ModuleStaticFileProviders` 

IMHO that value of `IOptions<StaticFileOptions>` must only be resolved inside `StaticMiddleware` so that all Framework provided FileProviders  and Custom File provided that uses `IOptions` pattern to configure are available to  `StaticMiddleware` when use `UseStaticFiles()` without any params.

May be better option is not to use or configure `IOptions<StaticFileOptions>` at all and change the **line 210** as following and not interfere any values of `IOptions<StaticFileOptions>`

```cs
var options = new StaticFileOptions(); 
  
 options.RequestPath = ""; 
 options.FileProvider = fileProvider;  
```

And Let developer control whether to serve files from `wwwroot` in their modules in tenant pipeline using  `UseStaticFiles()`

src/OrchardCore/OrchardCore/Modules/Extensions/ConfigureModuleStaticFileOptions.cs
ns8482e: As this could be vary based on source of static file, may be we don't use PostConfigure and we just use `new StaticFileOptions()` for module files?

src/OrchardCore/OrchardCore/Modules/Extensions/ConfigureModuleStaticFileOptions.cs
ns8482e: As this could be vary based on source of static file and developer's configuration, may be we don't use PostConfigure and we just use `new StaticFileOptions()` for module files?

src/OrchardCore/OrchardCore/Modules/Extensions/ConfigureModuleStaticFileOptions.cs
ns8482e: `using Microsoft.Extensions.Options;` is already included at line 7, but compile complains

ns8482e:
@sebastienros  @jtkech  removed, IPostConfigureOptions

jtkech:
@ns8482e Yes, I prefer this ;)

And let people do what they want from the app with our helpers, e.g.

    services.AddOrchardCms()
        .Configure((app, routes, serviceProvider) =>
        {
            app.UseStaticFiles();
        });

**Maybe add it to the doc, and maybe saying that for mutating / collaborate to the related host options better to still do it at the host level**. Note: As I remember, not so good at the tenant level to mutate host level options because we may build tenants concurrently, we had this kind of problem with routing services / options where we needed at the tenant level to remove from the DI some host level singleton services / configuration options before re-registering them.

So okay for now that at the tenant level we always use a `new StaticFileOptions()` as we already do for tenant static files.

https://github.com/OrchardCMS/OrchardCore/blob/c1f261fc59a9f11d46b5c8faf8715cd4619da9b2/src/OrchardCore.Modules/OrchardCore.Tenants/Startup.cs#L106-L117

Notice that here the `CacheControl` doesn't come from the shell configuration, hmm but we would need to have separate tenant level options, so that's okay for now, will see if needed.

For module embedded static files maybe we could use the same code pattern as above, then not sure about setting other properties e.g. `DefaultContentType` and `ServeUnknownFileTypes` as done above. Before they were coming from the current host level options.

Hmm, so one idea would be to still resolve the host level `StaticFileOptions` but in place of mutating it, clone it to a new one with the same properties unless for `RequestPath`, `FileProvider` and `OnPrepareResponse`. So exactly as before but without mutating the current host level `StaticFileOptions`.

ns8482e:
@jtkech  @agriffard  what would be accurate path to document this? IMHO This is related to framework and and not related to Tenant module.

jtkech:
@ns8482e 

> what would be accurate path to document this? IMHO This is related to framework and and not related to Tenant module.

Or maybe in a section dedicated to static files but didn't find it, anyway should not block this PR.

Last thing, what you think about still resolving the host level `StaticFileOptions` to still read the properties that we don't override, but this to populate the new `StaticFileOptions` that you created.

The idea is to work as before for modules static files but just without mutating the host `StaticFileOptions`. Hmm, but maybe better now that host `StaticFileOptions` don't act on module static files.

So as you want ;) let me know before approving.

PR #7237, opened 2020-10-09
Common properties for NuGet packages

ns8482e:
### Todo ( In separate PR )
- All Module title & description.
- Publish Symbol package to nuget.org and use of nuget v3 api in appveyor.yml for master branch (or for dev branch if cloudsmith.io supports symbol package)

### Completed
- Added Common Package description
- Update build script in appveyor.yml  and travis.yml for passing properties `SourceRevisionId`,  `GitBranch` to `dotnet build` to support Source Link
- Themes
   - [x] TheAdmin, 
   - [x] TheAgency
   - [x]  TheCommingSoon
   - [x] TheBlog 

- Add Source Link Properties 
- Added Git Commit hash for Product Version and build by passing git-commit-hash as in following command

Script to include git commit hash during build
```bash
dotnet build -c Release /p:SourceRevisionId=c990b5c38f0a44fb4f2ecb53b9848f8f4ceba0d0  /p:GitBranch=dev
dotnet pack -c Release --no-build
```
### Generated DLL Properties
image

### Generated Package Properties
image

Fixes #6404

PR #10129, opened 2021-08-16
Publish symbol for v1.0

### LINKED ISSUE #9990: For version 1.0 - Symbols are not published to nuget.org
### Describe the bug
For version 1.0 - Symbols are not published to nuget.org

### Expected behavior
Symbol to be published to nuget.org for  V 1.0

ns8482e:
Publish symbol - removed -n switch to allow publish snupkg

Fixes #9990 

/cc @sebastienros  @agriffard  @deanmarcussen

### DIFF
--- a/.github/workflows/release_ci.yml
+++ b/.github/workflows/release_ci.yml
@@ -63,7 +63,7 @@ jobs:
         VERSION="${arrTag[2]}"
         VERSION="${VERSION//v}"
         dotnet pack -c Release --no-restore --no-build -p:Version=$VERSION -p:TreatWarningsAsErrors=false
-        dotnet nuget push './src/**/*.nupkg' -k ${{secrets.NUGET_API_KEY}} -n true -s https://api.nuget.org/v3/index.json --skip-duplicate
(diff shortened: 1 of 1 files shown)


Skrypt:
I'm not sure if we should push the symbols to debug OC in a "release" version. Maybe we should provide a version with symbols and a version without because else the libraries are bigger and it could be a concern for certain people. Also, you generally don't need to debug a final release, you could equally use the preview feed at the next or previous commit after 1.0 to do so.

I need to check with @sebastienros if this was intended to be published without symbols.

ns8482e:
@Skrypt 

> else the libraries are bigger and it could be a concern

See size comparison discussion in original PR for symbols https://github.com/OrchardCMS/OrchardCore/pull/7237#issuecomment-715665839

> I need to check with @sebastienros if this was intended to be published without symbols.

May be it's copied from preview feed, and  `-n` is there for preview feed as cloudsmith doesn't support symbol server yet.

Skrypt:
Maybe I'm wrong here, I will let @sebastienros decide. The size is not that much bigger at least.

sebastienros:
The -s might have been for cloudsmith because it doesn't support them, and when I created the release pipeline I just copy-pasted it. I am ok with publishing the symbols on nuget.

sebastienros (approved):

PR #10131, opened 2021-08-17
Allow to input seconds in TimeField editor

jtkech:
Just by setting the step attribute to 1, otherwise the seconds are not editable.

If it is not a problem to have the seconds editable by default.

Otherwise would need to be a setting, knowing that if you set the step to e.g. `0.2` the milliseconds will be editable, so maybe for the step value we could only use an integer value < 60 (that only acts on the seconds), or for simplicity only allow `step=1` as done here. If so the setting could be a boolean => `Whether the seconds are editable`.


hishamco:
> If so the setting could be a boolean => Whether the seconds are editable

May be this a good option, coz the common usage for time picker is to set the hours & minutes

Skrypt:
Add a precision setting on the timefield which will result in setting the value of that `step="Model.Precision"`

jtkech:
@hishamco @Skrypt was too lazy ;)

Will do another PR with a setting

PR #10151, opened 2021-08-20
Update LICENSE

hishamco:


sebastienros:
I don't think mentioning years in copyrights is necessary, or even meaningful. I wouldn't change anything that is not 100% understood.

hishamco:
I saw such thing in some repos that are a part of .NET Foundation, you can close the PR if it's not needed

agriffard:
I don't think it is necessary either.

PR #10162, opened 2021-08-25
Messaging when getting Part InvalidCastException

brunoAltinet:
Ran into this a couple times and it always gets me. Something to help newbies when using As<Part>


src/OrchardCore/OrchardCore.ContentManagement.Abstractions/ContentExtensions.cs
```diff
@@ -23,7 +23,14 @@ public static class ContentExtensions
         /// <returns>The content element instance or <code>null</code> if it doesn't exist.</returns>
         public static TElement Get<TElement>(this ContentElement contentElement, string name) where TElement : ContentElement
         {
-            return (TElement)contentElement.Get(typeof(TElement), name);
+            try
```
sebastienros, line 26: Can we test the types instead of relying on an exception?

brunoAltinet: I was thinking about that but I didnt want to cause any regressions by demanding a certain type when previously a ContentElement would do. It would be more correct but it's potentially a breaking change.

sebastienros: I don't understand.
Let's imagine this code is written like this:

```csharp
var foo = contentElement.Get(typeof(TElement), name);
return (TElement)foo;
```
What is causing the `InvalidCastException` that is caught:

1. `var foo = contentElement.Get(typeof(TElement), name);`
2. `return (TElement)foo`

if the answer is "2", why not check with `if (foo is TElement)` instead of catching an exception?

brunoAltinet: Ok, but what to do after checking for type? Throw exception?

PR #10166, opened 2021-08-26
Preserve browser Autofill for bag, flow and widgetList parts

jtkech:
Currently while editing a flow / bag / widgetList part the html input ids are based on the related embedded item ids that always have different values. So we loose the browser autofill based on these html input ids when editing another parent.

So here the idea is to generate ids that are still unique while editing in the context of a given parent item, but that are the same while editing another parent content item of the same type.

This by generating ids using a prefix composed of the part name of a given unique / named part, a delimiter, then a number equals to the current items number. For the widgetList part we also use the related zone name in the prefix.


deanmarcussen (commented):
If you are using the index, how does that work when items are reordered? I assume ok, beacuse although it is no longer the correct index, it is still unique.

But what about adding via ajax, how does it remain unique then?

jtkech:
> If you are using the index, how does that work when items are reordered? I assume ok, beacuse although it is no longer the correct index, it is still unique.

Yes, it is only used for unicity, not for ordering that works as before.

> But what about adding via ajax, how does it remain unique then?

The prefix passed to the controller is also based on the items count, e.g.

    var prefix = partName + '-' + zone + '-' + $('#' + targetId + " .widget-editor-body").length.toString();

deanmarcussen:
Ok, cool, what about nested bags? Is the prefixing part names you’re constructing nested?

jtkech:
Yes I thought about this in case a nested type e.g. has the same bagPart name e.g. `MyBag1`.

Need to use `ParentPartName-ParentIndex_ChildPartName-ChildIndex`

We have collections to bind that don't need to be differentiated by the child index

Just tried, here for them we already generate different ids also based on the parent e.g.

    MyBag1-2_MyBag1_Prefixes

But I missed it for other fields, just tried, yes here we may generate another e.g. `MyBag1-0_TitlePart.Title`

This in place of `MyBag1-2_MyBag1-0_TitlePart.Title`.

Okay, in fact `MyBag1-2_MyBag1` is the current `HtmlPrefix` that I need to use

I will update it when I will have time

deanmarcussen:
Ok, could get pretty big on multiple contained bags / widgets?

Wonder if you've considered wrapping the model updater with a `PrefixedModelUpdater` when making the calls to build editor for the bag/widgets, and the bags contentitem id as part of the prefix, rather than the guid we were using before?

This _might_ also have the advantage of maintaining a valid prefix for model state errors, fixing the issue that the validation classes / model state errors don't correctly display on things inside a bag (just an idea / not checked at all)

src/OrchardCore.Modules/OrchardCore.Flows/Controllers/AdminController.cs
```diff
@@ -83,7 +83,7 @@ public async Task<IActionResult> BuildEditor(string id, string prefix, string pr
                 CanDelete: true,
                 //Input hidden
                 //Prefixes
-                HtmlFieldPrefix: prefix,
```
sebastienros, line 86: Removing this concerns me.
Have you checked that recursive parts will still work after this change? Meaning a bag part inside a flow part, something that has multiple levels of "dynamic" injection of an edit template.

jtkech: Yes, not all use cases but I tried nested things.

The `HtmlFieldPrefix` of the `ContentCard` was never used, only the `PrefixValue`.

Now we include the `HtmlFieldPrefix` (retrieved from the `ViewData.TemplateInfo`) in the `PrefixValue` on which we add an index value.

We also pass this `HtmlFieldPrefix` to the script through an html attribute so that, when we hit this controller through an ajax call, it is also included in the prefix parameter that we pass to the controller, this to set again here the `PrefixValue` of the `ContentCard`.

jtkech:
@deanmarcussen 

> Wonder if you've considered wrapping the model updater with a PrefixedModelUpdater when making the calls to build editor for the bag/widgets, and the bags contentitem id as part of the prefix, rather than the guid we were using before?

Yes but even if we use the contentitem id in place of a guid, we still loose the autofill e.g. when we create a new main parent contentitem. The goal is to have autofill when we add / insert a nth child if we already did it in another main parent item at the same place.

> This might also have the advantage of maintaining a valid prefix for model state errors, fixing the issue that the validation classes / model state errors don't correctly display on things inside a bag (just an idea / not checked at all)

Forgot to try it, maybe it fixes it ;)

> Ok, could get pretty big on multiple contained bags / widgets?

Yes maybe it is the main concern.

Hmm but here we retrieve more "regular" prefixes on which we only append indexes values. Note: Here we use the `HtmlFieldPrefix` retrieved from the `ViewContext.ViewData.TemplateInfo.HtmlFieldPrefix`.

sebastienros:
Let us know when you think it's ready to be nerged

jtkech:
Okay, hmm maybe better to retry it with some more nested things, i will let you know

jtkech:
@sebastienros just re-tried more nested things and checked that there are well edited and saved.

As pointed out by @deanmarcussen it generates longer ids / names, otherwise LGTM.

deanmarcussen (approved):
Thanks for checking with deeper nested ones

PR #10167, opened 2021-08-26
Allow custom editors for bag / flow / widgetsList parts

jtkech:
Just by using `GetEditorShapeType()` in the related drivers

And providing the related default / standard options views.


deanmarcussen (approved):
Cool, I didn't do this originally, because @ns8482e had built all the Content containery shapes for flows etc, but no problem to add it.

PR #10173, opened 2021-08-27
Lucene : Fix RangeQuery when parsing integer values from JSON

### LINKED ISSUE #10172: Lucene : RangeQuery with integer
When using a RangeQuery on a NumericField that stores a integer in Lucene it fails.
The issue is that the RangeQuery tries to parse a Int64 in the Lucene index while we are storing a Int32 value in Lucene.

We parse a Int64 value in the RangeQueryProvider because this values comes from JSON which means it could easily use a number up to a Int64 range.

I think it makes sense then to always index a Int64 value for numeric fields in Lucene so that we can be able to use a RangeQuery on these values. Else, we would need to make a distinction on the NumericField itself about if it should be a Int, Int32 or Int64 value.

Here, when we use a RangeQuery on a Field, it doesn't take in consideration on which field it evaluates the range so it doesn't know which type of Int to evaluate. So, I suggest we simply use Int64 for now.

Skrypt:
Fixes #10172

See : https://github.com/OrchardCMS/OrchardCore/blob/8ff2453b72b3bcfc3e2c0181b220be5bc720fd96/src/OrchardCore/OrchardCore.Lucene.Core/QueryProviders/RangeQueryProvider.cs#L70

As you can see it evaluates a Int64 value but we store a Int32 value so it always returns no results.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Lucene/Services/LuceneIndexManager.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lucene/Services/LuceneIndexManager.cs
@@ -231,9 +231,9 @@ namespace OrchardCore.Lucene
                         break;
 
                     case DocumentIndex.Types.Integer:
-                        if (entry.Value != null && Int32.TryParse(entry.Value.ToString(), out var value))
+                        if (entry.Value != null && Int64.TryParse(entry.Value.ToString(), out var value))
                         {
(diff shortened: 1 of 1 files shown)


jtkech:
Otherwise LGTM ;)

Let me know if you change something and if I need to review it again.

Issue #3973, opened 2019-07-25
Add Content-Language Header


hishamco:
Localization again 😄, while `Accept-Language` header is useful for the request, `Content-Language` is very useful too for the response, there's no support for this yet. I already started a PR in AspNetCore meanwhile, I suggest to add this as a feature for a Localization Module. after that we can integrate with the `RequestLocalizationMiddleware`

If sound is good, I will send a PR for this one

/cc @sebastienros @Skrypt

Skrypt:
Some reading : 
https://www.w3.org/International/questions/qa-http-and-lang

hishamco:
Thanks, AFAIK send the `Content-Language` header via response is way better that using meta tag

sebastienros:
I think this article is more descriptive of the issue: https://www.w3.org/International/questions/qa-text-processing-vs-metadata

It explains why the `lang` attribute and the `content-language` header are different.

PR #10178, opened 2021-08-28
Add Content-Language  HTTP header feature

hishamco:
Addresses #3973

I just wondering if `ContentLocalization` is a proper place!!

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Localization/Manifest.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Localization/Manifest.cs
@@ -9,3 +9,10 @@ using OrchardCore.Modules.Manifest;
     Category = "Internationalization",
     Dependencies = new[] { "OrchardCore.Settings" }
 )]
+
+[assembly: Feature(
+    Id = "OrchardCore.Localization.ContentLanguageHeader",
+    Name = "Content Language Header",
+    Description = "Adds the Content-Language HTTP header, which describes the language(s) intended for the audience.",
+    Category = "Internationalization"
+)]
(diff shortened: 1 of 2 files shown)


hishamco:
Nah, seems it's multi-targeting issue .. I need to figure out ..

src/OrchardCore.Modules/OrchardCore.Localization/Manifest.cs
```diff
@@ -9,3 +9,11 @@
     Category = "Internationalization",
     Dependencies = new[] { "OrchardCore.Settings" }
 )]
+#if NET5_0
```
sebastienros, line 12: OR GREATER ?

src/OrchardCore.Modules/OrchardCore.Localization/Startup.cs
```diff
@@ -62,4 +62,22 @@ public override void ConfigureServices(IServiceCollection services)
             services.AddSiteSettingsPropertyDeploymentStep<LocalizationSettings, LocalizationDeploymentStartup>(S => S["Culture settings"], S => S["Exports the culture settings."]);
         }
     }
+#if NET5_0
```
sebastienros, line 65: OR GREATER ?

src/OrchardCore.Modules/OrchardCore.Localization/Manifest.cs
hishamco: I tried this before I submit the PR but it doesn't work, and the code is grayed out, I think `OrGreater` is not shipped (frankly I'm not sure)

hishamco: https://github.com/dotnet/designs/blob/main/accepted/2020/or-greater-defines/or-greater-defines.md

sebastienros: worst case you can do both NET5_0 and NET5_0_OR_GREATER

PR #10202, opened 2021-09-02
Update README.md

### LINKED ISSUE #10169: Orchard.Setup define defaults when debugging
When developing, I often delete app_data and start from scratch to see if all my recipes etc work as intended. I have to input username password and select theme every time. 

It would be great if we could automate this initialization with some sort of json or other equivavelnt, that would also be useful for deployments.

No alternatives have been considered. Checked the docs, there's a section on OrchardCore.Setup but it's pretty rudimentary and mostly concerned with localization.

brunoAltinet:
fixes #10169


agriffard:
Please use more explicit titles for your PRs.

You could have made this change in the #10201 PR, no need to split them as they fix the same issue.

brunoAltinet:
Ok, you want me to add it there?

PR #10205, opened 2021-09-02
Docs - some clarity on AutoSetup parameters

brunoAltinet:
Current docs leave out these simple information which can bring more clarity to newcomers.

### DIFF
--- a/src/docs/reference/modules/AutoSetup/README.md
+++ b/src/docs/reference/modules/AutoSetup/README.md
@@ -4,7 +4,9 @@ The auto-setup module allows to automatically install the application/tenants on
 
 ## JSON Configuration Parameters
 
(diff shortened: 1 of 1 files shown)


Skrypt:
The "OrchardCore" section is not required if you are using the appsettings.json from the tenant folder. So this is a configuration detail that we generally don't add on every configuration examples because it varies depending on the location of the appsettings.json file.

brunoAltinet:
Would Orchard_AutoSetup work if it's put directly into root? I'm not sure i understand. App_Data doesnt exist on first call. Where should i put that node so that my site is up and running from scratch on dev without entering any setup details manually?

Skrypt:
Makes sense, forget my comment.

PR #10206, opened 2021-09-03
Diagnostics module fix (#9987)

### LINKED ISSUE #9987: error page from "OrchardCore.Diagnostics" module isn't presented when [ValidateAntiForgeryToken] attribute is used and validation failed
### Describe the bug
Sometimes, when my "**orchantiforgery_**" cookie expires, and when login page is still opened in browser, I can't log in, because of **[AntiforgeryValidationException](https://github.com/dotnet/aspnetcore/blob/main/src/Antiforgery/src/Internal/DefaultAntiforgery.cs)** which is expected behavior, but instead of seeing error page from 'OrchardCore.Diagnostics' I see default one.

### You can reproduce the issue by removing "**orchantiforgery_**" cookie, or by manipulating with html code.
1. Make sure 'OrchardCore.Diagnostics' module is enabled,
2. Go to 'https://try.orchardcore.net/Login',
3. Inspect page and change value of '__RequestVerificationToken' input,
4. Click on 'Log in',
5. See default 400 error page.
image

### Expected behavior
This error page should be presented to the user:
image

lampersky:
fixes #9987

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Diagnostics/Controllers/DiagnosticsController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Diagnostics/Controllers/DiagnosticsController.cs
@@ -4,6 +4,7 @@ using Microsoft.AspNetCore.Mvc;
 
 namespace OrchardCore.Diagnostics.Controllers
 {
+    [IgnoreAntiforgeryToken]
     public class DiagnosticsController : Controller
     {
         public IActionResult Error(int? status)
--- a/src/OrchardCore.Modules/OrchardCore.Diagnostics/Controllers/DiagnosticsController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Diagnostics/Controllers/DiagnosticsController.cs
@@ -4,9 +4,9 @@ using Microsoft.AspNetCore.Mvc;
 
 namespace OrchardCore.Diagnostics.Controllers
 {
-    [IgnoreAntiforgeryToken]
     public class DiagnosticsController : Controller
     {
+        [IgnoreAntiforgeryToken]
         public IActionResult Error(int? status)
         {
             // Most commonly used error messages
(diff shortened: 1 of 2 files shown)


src/OrchardCore.Modules/OrchardCore.Diagnostics/Controllers/DiagnosticsController.cs
```diff
@@ -4,6 +4,7 @@
 
 namespace OrchardCore.Diagnostics.Controllers
 {
+    [IgnoreAntiforgeryToken]
```
sebastienros, line 7: Can it be added to the method instead?

src/OrchardCore.Modules/OrchardCore.Diagnostics/OrchardCore.Diagnostics.csproj
```diff
@@ -19,4 +19,8 @@
     <ProjectReference Include="..\..\OrchardCore\OrchardCore.Module.Targets\OrchardCore.Module.Targets.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.ViewFeatures" Version="2.2.0" />
```
sebastienros, line 23: Is that necessary?

lampersky: good point, it's not, I will remove it

src/OrchardCore.Modules/OrchardCore.Diagnostics/Controllers/DiagnosticsController.cs
lampersky: sure, I will move it there

hishamco (approved):

PR #10228, opened 2021-09-06
Remove unnecessary references

EternityTeam:


jptissot (approved):

Skrypt:
Are you sure that these references are not there because of Nuget Packages solution?

EternityTeam:
> Are you sure that these references are not there because of Nuget Packages solution?

Checking the source code does not find the code that needs these references, and the compilation can succeed. Will this cause any problems？

sebastienros:
> Checking the source code does not find the code that needs these references

that's good

> the compilation can succeed

This is necessary obviously but not sufficient, because this module might need a project which is imported indirectly. For instance `OrchardCore.DisplayManagement` might already have a reference to `Liquid.Abstractions`, so you could think removing `Liquid.Abstractions` is fine since it builds without it, but if one day `OrchardCore.DisplayManagement` doesn't require `Liquid.Abstractions` anymore then it will fail. So what is important is to check that this module does require dependencies that are implicitly imported.

EternityTeam:
> > Checking the source code does not find the code that needs these references
> 
> that's good
> 
> > the compilation can succeed
> 
> This is necessary obviously but not sufficient, because this module might need a project which is imported indirectly. For instance `OrchardCore.DisplayManagement` might already have a reference to `Liquid.Abstractions`, so you could think removing `Liquid.Abstractions` is fine since it builds without it, but if one day `OrchardCore.DisplayManagement` doesn't require `Liquid.Abstractions` anymore then it will fail. So what is important is to check that this module does require dependencies that are implicitly imported.

This module implements shape cache output through IshapeDisplayEvents , the focus should be on shape result and cache policy. Dependence on other files should not be solved by this module. Otherwise, it is equivalent to dependence on CMS features. Excluding these references is helpful to use this module independently without using CMS features.

PR #10237, opened 2021-09-09
Fixes #10211: Localization feature not listed in the admin

### LINKED ISSUE #10211: select language (Culture) doesn't exists anymore
After  update to new version "1.1.0-preview-16455", the culture menu doesn't appear anymore. It seems the module haven't attached in this version.

image

jtkech:
Fixes #10211

Also if both `OC.Localization` and `OC.Localization.ContentLanguageHeader` are enabled, `UseRequestLocalization()` may be call twice resulting in 2 middlewares in the pipeline.

I will merge this PR as it fixes the related issue.

And for now I added to the new feature a dependency on `OC.Localization` and removed its call to `UseRequestLocalization()`.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Localization/Manifest.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Localization/Manifest.cs
@@ -4,16 +4,23 @@ using OrchardCore.Modules.Manifest;
     Name = "Localization",
     Author = ManifestConstants.OrchardCoreTeam,
     Website = ManifestConstants.OrchardCoreWebsite,
-    Version = ManifestConstants.OrchardCoreVersion,
-    Description = "Provides support for UI localization.",
-    Category = "Internationalization",
-    Dependencies = new[] { "OrchardCore.Settings" }
+    Version = ManifestConstants.OrchardCoreVersion
 )]
-#if (NET5_0 || NET5_0_OR_GREATER)
+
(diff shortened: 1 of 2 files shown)


src/OrchardCore.Modules/OrchardCore.Localization/Manifest.cs
```diff
@@ -4,16 +4,23 @@
...
+    Name = "Localization",
     Description = "Provides support for UI localization.",
-    Category = "Internationalization",
-    Dependencies = new[] { "OrchardCore.Settings" }
+    Dependencies = new[] { "OrchardCore.Settings" },
+    Category = "Internationalization"
 )]
-#if (NET5_0 || NET5_0_OR_GREATER)
+
+#if NET5_0_OR_GREATER
```
hishamco, line 18: @jtkech are you sure this work at your side, I presume the `NET5_0_OR_GREATER` is not supported yet

jtkech: @hishamco Yes it is, just tried through the donet cli

It compiles well under 3.1 but only if we use the directive, so the directive works.

hishamco: I will try it from VS 2019, coz the code is grayed out at the time of pushing the PR

PR #10240, opened 2021-09-10
Create KeyValuePairs field

sebastienros:
Example:

image

- [ ] Documentation
- [ ] Create migration for SQL index
- [ ] Optional: Option to prevent duplicate keys
- [ ] Optional: Option to prevent empty values
- [ ] Optional: Option to allow empty keys


src/OrchardCore.Modules/OrchardCore.ContentFields/Indexing/KeyValueParisFieldIndexHandler.cs
```diff
@@ -0,0 +1,25 @@
+using System.Threading.Tasks;
```
agriffard, line 1: Typo in file name: KeyValue**Paris**FieldIndexHandler.cs

sebastienros: It's actually easier to type, maybe I should change it everywhere.
I am not a big fan of KVP as a field name. 
- KeyValueField
- NameValueField
- PropertiesField

Also for the physical storage, wondering if it should use it's own struct instead of KVP, I haven't checked how it's serialized in Json. Another option could be to store it as `string[] keys` and `string[] values`. Or expose such helpers/filters to retrieve all keys and all values if it's an array of `Key/Value` structs

agriffard: https://github.com/EtchUK/Etch.OrchardCore.Fields#dictionary

deanmarcussen: Isnt it a name value field?

because the keys are not unique like a dictionary?

struct sounds sensible

sebastienros: The field from Etch looks good actually.

Currently there it doesn't prevent from having the same key multiple times. It could be an option to block that in the validation.

PR #10274, opened 2021-09-15
Taxonomy field editors should work with published versions

tropcicstefan:
#10288


sebastienros:
Can you file an issue and explain the problem, I want to be sure I understand the issue, and others too.

PR #10275, opened 2021-09-15
Align taxonomy filter to the right

tropcicstefan:
#10289 

Similar to #8079

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Taxonomies/Views/ContentsAdminListTaxonomyFilter.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Taxonomies/Views/ContentsAdminListTaxonomyFilter.cshtml
@@ -2,7 +2,7 @@
 
 <div class="form-group mb-n1 filter">
     <div class="btn-group mt-1">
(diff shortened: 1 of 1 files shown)


sebastienros:
Can you share a screenshot of the issue and with your fix?
I trust the change but I want to see it to understand if that is the best end result.

Skrypt (approved):

agriffard:
Do you approve because the code seems ok for you or because you actually tested it (LTR and RTL) ?
We asked for a screenshot so it would be easier to see the change.

tropcicstefan:
> Do you approve because the code seems ok for you or because you actually tested it (LTR and RTL) ?
> We asked for a screenshot so it would be easier to see the change.
I put the screenshots in the issue, sry, should have put them here

agriffard:
Thank you, sorry we didn't check in the issue.

Can you please confirm that it is working with a Right To Left culture?

hishamco:
> Can you please confirm that it is working with a Right To Left culture?

I may check this one

hishamco:
Please don't merge this, I need to check few cases, seems the filter picker looks ugly in RTL. I will check and fix it requires

hishamco:
Filter

Skrypt:
If it doesn't work with RTL then it might not work for the others too as the fix is exactly what I did before. So, yeah, test it, it will be safer in that case.

Skrypt:
All good I'm merging.

image

Issue #10282, opened 2021-09-16
Convert INotifier to async


deanmarcussen:
### Is your feature request related to a problem? Please describe.

I've replaced the `INotifier` on the front end with a signalR version, which works asyncronously.

### Describe the solution you'd like

Update `INotifier` to use `ValueTask` when adding notifications.

### Describe alternatives you've considered

Prefer not to do a custom implementation, as using `INotifier` allows the front end to also recieve messaging notifications from code in the `OrchardCore` code base, which is very convenient for us.

sebastienros:
Add `AddAsync` implementation that defaults to the `Add` one so it's not breaking.

sebastienros:
With default interface implementation, and test with a module (from package) built for the old version, to see if it still works.

PR #10301, opened 2021-09-19
Migrate INotifier to support async implementations

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/10282

Default interface implementation used for backwards compatability.


hishamco:
Can we add `Async` version with keeping the old methods instead of removing them?

Skrypt (approved):

PR #10303, opened 2021-09-19
JSON Recipe Deployment Step

deanmarcussen:
Adds a deployment step where you can just write a JSON Recipe and add it to the plan.

Useful to create ad-hoc steps
image


src/OrchardCore.Modules/OrchardCore.Deployment/Views/Items/JsonRecipeDeploymentStep.Fields.Edit.cshtml
```diff
@@ -0,0 +1,85 @@
+@using Newtonsoft.Json.Linq;
+
+@model JsonRecipeDeploymentStepViewModel
+@inject IStringLocalizer<ViewContext> S
+
+@{
+    var schema = new JObject
```
sebastienros, line 7: Static string until it's really dynamic and then it will come from an action.

sebastienros (approved):

PR #10314, opened 2021-09-20
Update to .NET 6

vritant24:


Skrypt:
We should probably also move away from 3.1 and keep 5.0 instead.

hishamco:
Duplicate of #10317

hishamco:
I think we need to close this while the other PR is almost complete

Thanks

PR #10317, opened 2021-09-21
net6.0 default :arrow_up:

sebastienros:


src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
```diff
@@ -1,16 +1,16 @@
...
   <Import Project="..\OrchardCore\OrchardCore.Application.Cms.Core.Targets\OrchardCore.Application.Cms.Core.Targets.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>$(AspNetCoreTargetFrameworks)</TargetFrameworks>
+    <TargetFrameworks>$(CommonTargetFrameworks)</TargetFrameworks>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <WarningsNotAsErrors>612,618</WarningsNotAsErrors>
     <AspNetCoreHostingModel>InProcess</AspNetCoreHostingModel>
-    <PreserveCompilationReferences>true</PreserveCompilationReferences>
+    <!--<PreserveCompilationReferences>true</PreserveCompilationReferences>-->
```
sebastienros, line 13: @jtkech if I don't do this, it doesn't work when started from VS. Getting a lot of reference errors.

jtkech: Okay, I will check

sebastienros: Pranav is dubious too, maybe a bug somewhere. I have other compilation issues if I set the `DefaultTargetFramework` to `net5.0`.

jtkech: Just finished to update VStudio preview with my slow connection, will be soon connected to fiber optic ;)

No problem to build with the dotnet CLI with or without `PreserveCompilationReferences`.

Just tried under VStudio, for now it can't build saying that it can't load some analyzer dlls `CS8032`, indeed there are not under the dotnet/packs folder, related to `*.Generators.dll`.

jtkech: So with last preview of VStudio 2019 I was not able to build, there were missing analyzers dlls

Finally at the end of the night I decided to load VStudio 2022 Preview ;)

Okay I could repro the razor compile errors, missing assembly refs, even if the refs folder is well populated.

It works in prod mode OR if we set `PreserveCompilationReferences` to false

Now that I can repro, I will be able to debug it tomorrow.

ns8482e: It’s annoying to keep changing in project file to enable compilation and build project

is there possibility to keet it false in project?  As compilation can be enabled using env var  `"ASPNETCORE_HOSTINGSTARTUPASSEMBLIES": "Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation"`

sebastienros: c.f. https://github.com/aspnet/Announcements/issues/460

I am not sure from the description if it means that we should not use the property since it's already done by another package and it's then conflicting.

jtkech: @ns8482e 

The "problem" is that we control the **razor compilation services registration at the tenant level**, the package being referenced automatically when we reference `OC.Mvc.Core`. Hmm, maybe we could also test the related env variable but still at the tenant level.

https://github.com/OrchardCMS/OrchardCore/blob/1586fa0b18a41d6472116fd75e232543a53fe5a1/src/OrchardCore/OrchardCore.Mvc.Core/Startup.cs#L111-L117

Then `PreserveCompilationReferences` at the app level still allows to control the outputing of ref assemblies in the refs folder. **Here I now suggest to set it to false by default.**

jtkech:
@sebastienros 

`OC.Cms.Web` still need to target all TFMs because the CI build on all TFMs
Thats why it failed, then yes only the default TFM if e.g. under Visual Studio

jtkech:
`Dependencies.props` is no more shown under the build solution folder, is it intended?

What you think about setting `PreserveCompilationContext` to false by default?

jtkech:
> `Dependencies.props` is no more shown under the build solution folder, is it intended?

> What you think about setting `PreserveCompilationContext` to false by default?

Just saw that you updated the code around this

Just to say that if `PreserveCompilationContext` is set to false the refs folder is not outputed, so it doesn't exixt.

sebastienros:
Good point ... might be a better option. And we could not generate it y default in the templates too after that. Testing ...

.github/workflows/pr_ci.yml
```diff
@@ -30,13 +30,18 @@ jobs:
...
       run: |
-        dotnet build --configuration Release --framework net5.0
-        dotnet build --configuration Release --framework netcoreapp3.1
+        dotnet build --configuration Release --framework net6.0 /p:DefaultTargetFramework=net6.0
+        dotnet build --configuration Release --framework net5.0 /p:DefaultTargetFramework=net5.0
+        dotnet build --configuration Release --framework netcoreapp3.1 /p:DefaultTargetFramework=netcoreapp3.1
     - name: Unit Tests
       run: |
-        dotnet test -c Release --no-restore --no-build ./test/OrchardCore.Tests/OrchardCore.Tests.csproj 
+        dotnet test -c Release --no-restore --no-build ./test/OrchardCore.Tests/OrchardCore.Tests.csproj /p:CommonTargetFramework=""net6.0;net5.0;netcoreapp3.1"
```
jtkech, line 44: As I remember it is problematic to run the unit tests on all TFMs concurrently

.github/workflows/pr_ci.yml
```diff
@@ -30,13 +30,18 @@ jobs:
...
+    - name: Setup .NET Core 6.0 on ${{ matrix.os }}
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 6.0.100-rc.1.21458.32
     - name: Build
       run: |
-        dotnet build --configuration Release --framework net5.0
-        dotnet build --configuration Release --framework netcoreapp3.1
+        dotnet build --configuration Release --framework net6.0 /p:DefaultTargetFramework=net6.0
+        dotnet build --configuration Release --framework net5.0 /p:DefaultTargetFramework=net5.0
```
jtkech, line 40: So what happens if e.g. we do

    dotnet build --configuration Release --framework net6.0 /p:DefaultTargetFramework=net5.0

src/OrchardCore.Build/Dependencies.AspNetCore.props
```diff
@@ -1,43 +1,55 @@
...
+  <!-- Default TFMs to use when building and packaging -->
   <PropertyGroup>
-    <DefaultTargetFramework>net5.0</DefaultTargetFramework>
-    <AspNetCoreTargetFrameworks Condition="'$(AspNetCoreTargetFrameworks)' == ''">net5.0;netcoreapp3.1</AspNetCoreTargetFrameworks>
+
+    <!-- TFM used when coding in Visual Studio and when creating templates -->
+    <DefaultTargetFramework Condition="'$(DefaultTargetFramework)' == ''">net6.0</DefaultTargetFramework>
+    
+    <!-- TFMs used to build the abstractions and modules-->
+    <CommonTargetFrameworks Condition="'$(CommonTargetFrameworks)' == ''">$(DefaultTargetFramework)</CommonTargetFrameworks>
```
jtkech, line 10: So, if we do this here, what we do below line 15 is useless

Before `DefaultTargetFramework` was just used as an inner variable, that's why it was set without any condition

And was just used to set the TFMs if e.g. under VStudio

src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
```diff
@@ -1,16 +1,17 @@
...
 
   <PropertyGroup>
-    <TargetFrameworks>$(AspNetCoreTargetFrameworks)</TargetFrameworks>
+    <TargetFrameworks>$(CommonTargetFrameworks)</TargetFrameworks>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <WarningsNotAsErrors>612,618</WarningsNotAsErrors>
     <AspNetCoreHostingModel>InProcess</AspNetCoreHostingModel>
-    <PreserveCompilationReferences>true</PreserveCompilationReferences>
+    <!-- Required for runtime compilation because the dependency on RuntimeCompilation package is not transitive -->
+    <PreserveCompilationContext Condition="'$(TargetFramework)' != 'net6.0'">true</PreserveCompilationContext>
```
jtkech, line 14: I would have just set it to false

jtkech:
@sebastienros fir info I did some tests

So we have `PreserveCompilationContext` and `PreserveCompilationReferences` that I renamed here `PCContext` and `PCReferences`. One problem is that if now we only use `PCContext`, we can't control anymore the refs folder for net5.0, it is always outputed. So I tested both properties for both net6.0 and net5.0, the only difference is that for net 6.0 `PCContext` is **false by default**, for net 5.0 **it is true by default**, this in our context (see below about `ExcludeAssets`).

    net6.0 => By default PCContext = false - PCReferences = false
    PCContext = true - PCReferences = true => okay (refs outputed)
    PCContext = false - PCReferences = false => okay (refs not outputed)
    PCContext = true - PCReferences = false => okay (refs not outputed)
    PCContext = false - references = true => fails at runtime
    net5.0 => Idem unless that PCContext is true by default

That said in the `OC.Mvc.Core` project we have

    <PackageReference Include="Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation"
        ExcludeAssets="build;buildTransitive" />

If we remove `ExcludeAssets="build;buildTransitive"`, it doesn't fail with `PCContext = false` and `PCReferences = true`, but in that case we can't control anymore the refs folder with `PCContext` and / or `PCReferences`, here it is always outputed., that's why we used `ExcludeAssets`, so i re-used it to retrieve the same above test results.

---

So there are different solutions

- We could let it as it was before unless the better names that you use here, then we could have a target that executes after but not too late to do the right things. This by only exposing only one property e.g. still `PCReferences` and then setting `PCContext = PCReferences` in the target, or only expose `PCContext` and then set `PCReferences = PCContext`.

    In that case, in our projects I would set this property to false by default.

- Another approach woul be to not expose anymore both properties and let people referencing the `RuntimeCompilation` package themselves in their projects. Just tried with net5.0 it works, will try with net6.0

   In that case the `RuntimeCompilation` may be referenced twice, one for our purposes, and one from a given main project, directly or transitively but without the `ExcludeAssets`, this to validate or not the runtime compilation. This knowing that in production we will still disallow runtime compilation, but the refs folder may have been populated for nothing, but as it could be before with the project property `PCReferences`.

jtkech:
So the best and less confusing way I found is to just let people reference or not the `RuntimeCompilation` package as they would do for a regular application, and let them play with `PreserveCompilationContext` and `PreserveCompilationReferences` if they want. Then in our web and template projects we could just provide an example e.g. with a dedicated property

    <PropertyGroup>
      <RazorRuntimeCompilation>false</RazorRuntimeCompilation>
    </PropertyGroup>

    <ItemGroup>
      <PackageReference Include="Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation"
        Condition="'$(RazorRuntimeCompilation)' == 'true'" />
  </ItemGroup>

agriffard:
In https://github.com/OrchardCMS/OrchardCore/blob/a4081acd2d77349530f741d88e4ab4e520d9037a/.github/workflows/functional_all_db.yml#L26

The dotnet-version is used. 
Do we add a new line or replace it by `6.0.*`?

jtkech:
Yes if @sebastienros still want to support 5.0 and 3.1 as it seems to be
We would need to add the 4 related lines for 6.0.

hishamco:
> Yes if @sebastienros still want to support 5.0 and 3.1 as it seems to be

I think we should remove 3.1, while 6.0 is LTS

.github/workflows/pr_ci.yml
jtkech: Hmm, forgot, as I remember now this is dotnet build all projects  that was problematic on all TFMs

Here there is a parsing issue

    ParserError: D:\a\_temp\53ef018a-7f1d-48b6-bd60-6e201c3d939a.ps1:2
    Line |
       2 |  … .Tests.csproj /p:CommonTargetFramework=""net6.0;net5.0;netcoreapp3.1"
         |                                                                        ~
         | The string is missing the terminator: ".

I will update it and see if the CI build

.github/workflows/pr_ci.yml
```diff
@@ -30,13 +30,18 @@ jobs:
...
+        dotnet build --configuration Release --framework netcoreapp3.1 /p:DefaultTargetFramework=netcoreapp3.1
     - name: Unit Tests
       run: |
-        dotnet test -c Release --no-restore --no-build ./test/OrchardCore.Tests/OrchardCore.Tests.csproj 
+        dotnet test -c Release --no-restore --no-build ./test/OrchardCore.Tests/OrchardCore.Tests.csproj /p:CommonTargetFramework=""net6.0;net5.0;netcoreapp3.1""
```
jtkech, line 44: There was a parsing issue I fixed by adding a `"` at the end

Hmm, but not sure we can pass an msbuild property with `dotnet test`

Will see tomorrow

jtkech:
Summary as a reminder

- Okay, I understood what you did around `DefaultTargetFramework`, but then the following dotnet cmds fail

      dotnet build -f net5.0
      dotnet build -f net5.0 /p:DefaultTargetFramework=net6.0

- We can't pass an msbuild prop to `dotnet test`, so we can't pass `CommonTFMs`, and here if we don't pass them it falls back to the only default net6.0. Note: Hmm when I tried it without passing them no test was run.

- E.g. under net5.0 the following doesn't work, we still need `PCReferences` (see my related comments), or another way e.g. by referencing the razor package as in #10380 (without talking about the hot reload).

      <PreserveCompilationContext Condition="'$(TargetFramework)' != 'net6.0'">true</PreserveCompilationContext>

src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
ns8482e: @jtkech as when looked at `refs` folder - it also compiles all views embedded in  nuget module -  where intention should be to only compile project in solution.

jtkech: Did you try it? As I remember we check if a given reference is a project or package.

Also, if you have time you could try the PR #10380 I did based on this branch, and where I use a more regular approach by referencing (or not) the razor compile package directly from the main app.

PR #10363, opened 2021-09-28
Prevent Displaying Error Stack Page (NRE) in case of missing Content Type

urbanit:


src/OrchardCore.Modules/OrchardCore.Layers/Controllers/AdminController.cs
```diff
@@ -100,8 +100,16 @@ public async Task<IActionResult> Index()
                 {
                     model.Widgets.Add(zone, list = new List<dynamic>());
                 }
+                try
+                {
+                    var widgetDisplay = await _contentItemDisplayManager.BuildDisplayAsync(widget.ContentItem, _updateModelAccessor.ModelUpdater, "SummaryAdmin");
+                    list.Add(widgetDisplay);
+                }
+                catch (Exception _e) {
+                    await _notifier.ErrorAsync(H[_e.Message]);
```
sebastienros, line 109: We don't translate dynamic messages.
If the exception is specific we can catch it and render a custom message.

In this case logging an error would be fine. Or even better if it's done in the BuildDisplay if it can detect the missing content type. And not throw anything.

src/OrchardCore/OrchardCore.ContentManagement.Display/ContentItemDisplayManager.cs
```diff
@@ -57,6 +57,10 @@ public async Task<IShape> BuildDisplayAsync(ContentItem contentItem, IUpdateMode
             }
 
             var contentTypeDefinition = _contentDefinitionManager.GetTypeDefinition(contentItem.ContentType);
+            if (contentTypeDefinition == null)
+            {
+                throw new ArgumentNullException(contentItem.ContentType);
```
sebastienros, line 62: Right, so here just log an error and don't render anything. Or throw an OrchardException with a meaningful error message.

urbanit: Logger is not referenced, is it ok to add it?

src/OrchardCore.Modules/OrchardCore.Layers/Controllers/AdminController.cs
urbanit: I will add error but I think that Admin should be notified somehow, shouldn't?. It's clear about dynamic messages, but we have same exception for null content item and null definition. So, a general error string or no notification?

src/OrchardCore/OrchardCore.ContentManagement.Display/ContentItemDisplayManager.cs
hishamco: No problem if you reference logging abstractions

PR #10368, opened 2021-09-29
Encode redirect urls

### LINKED ISSUE #10364: Redirect URL encoding issue on Azure
### Describe the bug
When publishing this ```Admin/Contents/ContentItems/{ContentItemId}/Edit?returnUrl={greek or unicode characters}``` the redirect goes to ```{some encoded or strange characters url}``` and of course we are getting page not found... It happens on Azure, not on local dev env.. Both on .net 3.1 and 5

### To Reproduce
1. Publish a blogpost with permalink that contains non English characters
2. Edit this blogpost and have a redirect url the above permalink
3. See error

### Expected behavior
Non English characters should not be encoded

### Screenshots
image

image

image

sebastienros:
Fixes #10364

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
+++ b/src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
@@ -133,7 +133,7 @@ namespace OrchardCore.AutoSetup
                     if (settings.State != TenantState.Uninitialized)
                     {
                         await _shellHost.ReloadShellContextAsync(_shellSettings, eventSource: false);
-                        httpContext.Response.Redirect(pathBase);
+                        httpContext.Response.Redirect(Uri.EscapeUriString(pathBase));
 
                         return;
                     }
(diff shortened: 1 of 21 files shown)


src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
```diff
@@ -133,7 +133,7 @@ public async Task InvokeAsync(HttpContext httpContext)
                     if (settings.State != TenantState.Uninitialized)
                     {
                         await _shellHost.ReloadShellContextAsync(_shellSettings, eventSource: false);
-                        httpContext.Response.Redirect(pathBase);
+                        httpContext.Response.Redirect(Uri.EscapeUriString(pathBase));
```
jtkech, line 136: Here the implicit `PathString` ToString() already encodes it

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
```diff
@@ -153,7 +153,7 @@ public async Task InvokeAsync(HttpContext httpContext)
                             }
                         }
 
-                        httpContext.Response.Redirect(pathBase);
+                        httpContext.Response.Redirect(Uri.EscapeUriString(pathBase));
```
jtkech, line 156: idem

PR #10373, opened 2021-09-30
Add store external tokens to OpenID Client recipe step

deanmarcussen:
Allows to set the store external tokens property in a recipe


kevinchalet (approved):
👍🏻

PR #10391, opened 2021-10-01
Added feature ID in braces to feature name for clarity

brunoAltinet:
This one got me banging head against the wall for hours. I assumed that the feature is calle OrchardCore.Media.Azure, but it's actually called OrchardCore.Media.Azure.Storage. There's no way to see that unless you have OrchardCore source open, and i'm explictliy building against packages. We need to show feature id somewhere, somehow.


src/OrchardCore.Modules/OrchardCore.Taxonomies/Liquid/TaxonomyTermFilter.cs
```diff
@@ -1,4 +1,3 @@
-using System;
```
brunoAltinet, line 1: this kinda slipped in, but it's unnecessary using. I can undo it if you want

deanmarcussen:
Can you send a screenshot with the changes?

Looks like a good idea though

Piedone:
I checked out the PR, going over old ones, and this would indeed be quite useful. So, I fixed the merge conflict.

Old UI:

image

New UI with @brunoAltinet's changes:

image

After I tweaked it with `<code>`:

image

deanmarcussen (approved):

PR #10392, opened 2021-10-01
Remove EscapeUriString usage

sebastienros:
Deprecated in net6.0


sebastienros:
It's also removing some escaping when it was not necessary, for instance when urls are generated by UrlHelper.

jtkech:
@sebastienros 

Just tried your new extension on my working branch

For now looks good, it seems that it escapes but only if needed

Tried with special chars in the path and / or the queryString of the returnUrl

Not tried if the returnUrl contains an host part but I think it is okay

Will still try to repro the use case of @urbanit, maybe still an extra encoding (not url related) before passing the returnUrl.

PR #10425, opened 2021-10-06
Refactor ContainedPart indexing to consider display Order data.

### LINKED ISSUE #10424: ContainedPart Order is not indexed in Lucene
We index Content.ContentItem.ListContentItemId in Lucene to keep the parent reference in list content type children content items. We should also keep the Order set from the admin UI to order properly these records from a Lucene Query.

Skrypt:
Fixes #10424


sebastienros:
TODO: Document change for lucene queries

PR #10443, opened 2021-10-10
Added support for Razor Component (*.razor files)

### LINKED ISSUE #10442: Support for Razor files
Support  Razor component (*.razor ) files in module and to allow rendering using component tag helper in liquid

Update tag helper bridge tp component tag helper

Currently to support razor file compilation in module, requires to explicitly exclude *.razor from embedded resource and from none and include in content in project file.

content tag helper is currently not supported in liquid.

ns8482e:
Added support for razor files and component tag helper

For example
```liquid
{% helper "component", type: "OrchardCore.Demo.Components.HelloBlazor, OrchardCore.Demo", render_mode: "Static", param_Title: Model.Title   %}
```

Fixes #10442


Skrypt:
Ok, I need to understand why we couldn't make an OC module that would use a Blazor Client Side Webassembly project. It should work if it's not using any other MVC dependencies right?

Skrypt:
So this would require that we provide a complete admin UI for the SPA application and that this be run alongside the current admin right?

ns8482e:
Closing as this can be achieved by defining on project file too

PR #10444, opened 2021-10-10
jQuery UI 1.13.0 :arrow_up:

agriffard:
https://www.npmjs.com/package/jquery-ui/v/1.13.0

CDN does not respond for the moment: https://code.jquery.com/ui/1.13.0/jquery-ui.js , SRIs still need to be updated.


hishamco:
@agriffard could you please wait before merge, I need to check the SRI for this using code

Skrypt:
This is a tricky one. Somehow the DateTime picker may be broken because of these JQueryUI updates but now since it's broken I'd say let's merge it in the main branch but also make tests where it is used. To be honest I'd love that we move away from JQuery UI at some point.

Issue #10104, opened 2021-08-12
Cors module bug while deleting any origin


vitalybrandes:
Add 2 hosts to origin and try to delete first
the second from the list deleted instead first.

![image](https://user-images.githubusercontent.com/61165619/129150367-e804b984-c7fb-4c08-b3d5-217e698b3b76.png)

![image](https://user-images.githubusercontent.com/61165619/129150416-da30085e-82ab-4d93-9c88-7ff18c3b4e53.png)

sebastienros:
Do you feel like you could try fixing it? Might be an array index issue.

Skrypt:
This UI would need some ❤️ 
I listed the missing bits on the original PR.

PR #10447, opened 2021-10-10
CORS UI: fix delete, do not add empty/duplicate entries

kaipm:
The delete function for list items in the CORS admin UI did not work properly (see #10104). This affected all lists on the CORS admin page.

The PR makes the following changes:
- `Delete` buttons remove the correct items
- `Add` button in lists **ignore**:
  - null or empty entries
  - duplicate entries (case insensitive)

The fix is quite minimal, because I know next to nothing of vue.js, but the UI should be less error prone that way.


deanmarcussen (approved):

Skrypt:
This one has never been triaged and could be added to 1.2 @sebastienros

PR #10528, opened 2021-10-21
Link field escaping

### LINKED ISSUE #10520: Decode UTF in Link Field
The item that holds the URL, Link Field, does not accept non-Unicode characters.
For example the link /τροφιμα/ψαρια must encode to %2F%CF%84%CF%81%CE%BF%CF%86%CE%B9%CE%BC%CE%B1%2F%CF%88%CE%B1%CF%81%CE%B9%CE%B1 and then save.
The annoying thing is that when I edit it I, must decode it, if I want to read it.

It will be nice if non-latin characters encoding and decoding in background

jtkech:
Fixes #10520

So that a link with the following href works

    /τροφιμα/ψαρια

Note: Currently an url starting with a tilde `~`doesn't work, should we support it?


Skrypt:
You merged it without any approval. 😭 
Don't worry, we won't fire you 😉

Issue #4563, opened 2019-10-18
How Attached MediaField keeps the original file name？


chinasqzl:
Attached MediaField uploaded as a GUID + real file name, saved and changed to MD5.xxx, is there such a necessary, or has other options to retain the original file name?

matiasmolleja:
No other options as it is now.
When we designed how it should work we tried to preserve disk space as much as possible.
We use the hash to check if a newly added file is already there. If it is there we don't upload the new file.

We do that here:
https://github.com/OrchardCMS/OrchardCore/blob/f21b24cf4f2a7ab441bae03eba09f9322ab235f8/src/OrchardCore.Modules/OrchardCore.Media/Services/AttachedMediaFieldFileService.cs#L105

Skrypt:
Hash is usefull for avoiding the issue of people naming their files with similar names. You don't want to override files when uploading new ones and you can't really remember all the file names you used before. I think here, what @chinasqzl wants is to add a name portion after the hash so that the file names themselves could be indexed by a search engine for SEO or could be also for adding to a Lucene index. For that matter, I prefer that we make use of a Title field for images. Most of the time people don't look at their image names and they just upload it even if they have strange chars and else. Ex: ( "my-file_06-copy2.jpg")

deanmarcussen:
Yes I like the idea of appending the source file as an extension to the hash. We have had a requirement to know the original file name with this field, but had to capture it manually in a text field. 
I can imagine it’s not totally uncommon to need to maintain the original file name

sebastienros:
That could be part of the data stored on the field instead.

deanmarcussen:
What about a dictionary on the field for metadata such as this, and perhaps an alt tag value from an alt-editor @sebastienros ?

sebastienros:
We have talked about tags already for standard media fields. I would be ok with that only as another property on the field. But that means changing the ui significantly. But no dictionary, it's not open ended. Well-known properties only.

chinasqzl:
File name is especially important when the user uploads an attachment, not a picture

DrewBrasher:
@deanmarcussen I've had to go ahead and start using that pull request in a production site. Is this still something the Orchard team is interested in supporting or should I work on a way to achieve this through a separate module?

Skrypt:
We need to review back the PR. It is summer time, people are busy, sorry for that.

DrewBrasher:
I understand

hishamco:
[...]

PR #10570, opened 2021-10-27
Keep file name of attached media files

DrewBrasher:


DrewBrasher:
> It's making us wonder if we want to display something that actually doesn't exist. Also if it would make sense to keep the original filename in the new file computed one.

I would prefer the original name be kept if it can be.  That way if a user downloads the file it will have the original name.  What do you think of the suggestion in my previous comment?

src/OrchardCore.Modules/OrchardCore.Media/Drivers/MediaFieldDriver.cs
```diff
@@ -123,6 +145,8 @@ public override async Task<IDisplayResult> UpdateAsync(MediaField field, IUpdate
 
                 field.Paths = items.Where(p => !p.IsRemoved).Select(p => p.Path).ToArray() ?? new string[] { };
 
+                field.SetAttachedFileNames(items.Where(i => !i.IsRemoved).Select(i => i.AttachedFileName).ToArray());
```
deanmarcussen, line 148: I think this should only happen when the editor is attached. Around line 115

src/OrchardCore.Modules/OrchardCore.Media/Drivers/MediaFieldDriver.cs
```diff
@@ -110,6 +122,16 @@ public override async Task<IDisplayResult> UpdateAsync(MediaField field, IUpdate
                 // If it's an attached media field editor the files are automatically handled by _attachedMediaFieldFileService
                 if (string.Equals(context.PartFieldDefinition.Editor(), "Attached", StringComparison.OrdinalIgnoreCase))
                 {
+                    for (var i = 0; i < items.Count; i++)
+                    {
+                        if (items[i].IsNew)
+                        {
+                            var lastSlash = items[i].Path.LastIndexOf('/');
+                            var nameOnly = items[i].Path[(lastSlash + 1)..];
+                            items[i].AttachedFileName = nameOnly[36..]; // remove unique id from name
```
deanmarcussen, line 131: It would be better if the javascript app includes the `AttachedFileName` on it's  post, than having to do this .

src/OrchardCore.Modules/OrchardCore.Media/wwwroot/Scripts/media.js
```diff
@@ -3351,7 +3356,7 @@ Vue.component('mediaFieldThumbsContainer', {
                         <div class="media-container-main-item-title card-body">\
                             <a href="javascript:;" class="btn btn-light btn-sm float-right inline-media-button delete-button"\
                                 v-on:click.stop="selectAndDeleteMedia(media)"><i class="fa fa-trash" aria-hidden="true"></i></a>\
-                            <span class="media-filename card-text small text-danger" :title="media.name">{{ media.name }}</span>\
+                            <span class="media-filename card-text small text-danger" :title="media.name">{{ media.attachedFileName == null ? media.name : media.attachedFileName }}</span>\
```
deanmarcussen, line 3359: I agree with @sebastienros here, we shouldn't be showing anything different in the edit UI here.

deanmarcussen (commented):
Reviewing the issue, I agree with @matiasmolleja when this was originally designed, the hash was an important part of the design for the feature.

The suggestion was to provide a less well known field, which could be used when displaying the media item (in the front end, not the admin) to provide some context to the user as to the file they might be downloading.

I'm open to being convinced we could show this information in the admin as well, but it would have to be an addition the the actual file name created from the hash.

I'm open to seeing how we could maintain the hash and the original file name (without hacks), but I don't see a particular way forward there.

DrewBrasher:
@deanmarcussen I will update my changes to only store the file name, not display it and make the other changes you suggested here.  

I do think something needs to be done to the admin display for files that don't have a thumbnail image like pdfs because right now it just shows a blank thumbnail and a hashed name so admin users have no indicator of what files they have already attached.  But maybe that should be a separate issue or discussion.

deanmarcussen:
merge conflicts to fix

src/OrchardCore.Modules/OrchardCore.Media/Drivers/MediaFieldDriver.cs
DrewBrasher: I moved that line up in  https://github.com/OrchardCMS/OrchardCore/pull/10570/commits/5280f52ecdeb677b0b64216b5e99ffa3a1bd8f95

src/OrchardCore.Modules/OrchardCore.Media/wwwroot/Scripts/media.js
DrewBrasher: I reverted my changes to the admin display in https://github.com/OrchardCMS/OrchardCore/pull/10570/commits/5cd1a8151491adf11b7c83334e50b226c6bae8cf

src/OrchardCore.Modules/OrchardCore.Media/Drivers/MediaFieldDriver.cs
DrewBrasher: I removed this code in https://github.com/OrchardCMS/OrchardCore/pull/10570/commits/99ece1690cb60ed361b68ef50f7b7924805aa985 but I'm still trying to figure out where to include it in the JavaScript app.

DrewBrasher:
> merge conflicts to fix

The conflicts were with the wwwroot\Scripts\media.js and media.min.js files.  I resolved the conflict by overwriting mine with the OrchardCMS main.  These files were updated by gulp.  Should they be excluded from git?  If not, I will add my generated files back to this once it is working.

deanmarcussen:
> These files were updated by gulp. Should they be excluded from git? If not, I will add my generated files back to this once it is working.

You'll need to regenerate with gulp, and commit to source control (we have to keep the generated files in source control because not all users, use the NuGet releases, some use the repository directly)

I recomend using `gulp watch` and just changing the media.js file so it rebuilds, rather than trying to rebuild the entire pipeline

src/OrchardCore.Modules/OrchardCore.Media/Drivers/MediaFieldDriver.cs
DrewBrasher: @deanmarcussen, If I understand how this works correctly, when you select or drop in a file it immediately posts the file and saves it to a temporary location and then moves it once you save the content type the file is attached to.  Is that correct and if so, at which step should the AttachedFileName be set?

DrewBrasher: I think I figured it out but I'm not sure I did it the best way.  Stepping through the JavaScript I did not see another way but I may have missed something.  It's in this commit: https://github.com/OrchardCMS/OrchardCore/pull/10570/commits/481d4f3e282148802fe8a59bb16d9d5205851121

src/OrchardCore.Modules/OrchardCore.Media/Drivers/MediaFieldDriver.cs
```diff
@@ -87,6 +87,18 @@ public override IDisplayResult Edit(MediaField field, BuildFieldEditorContext co
                     }
                 }
 
+                var filenames = field.GetAttachedFileNames();
```
ns8482e, line 90: Why use extension method here? Why not just add property to MediaField?

src/OrchardCore.Modules/OrchardCore.Media/Drivers/MediaFieldDriver.cs
```diff
@@ -112,6 +124,7 @@ public override async Task<IDisplayResult> UpdateAsync(MediaField field, IUpdate
                 {
                     try
                     {
+                        field.SetAttachedFileNames(items.Where(i => !i.IsRemoved).Select(i => i.AttachedFileName).ToArray());
```
ns8482e, line 127: Add property to media field and use here

ns8482e (changes requested):

src/OrchardCore.Modules/OrchardCore.Media/Drivers/MediaFieldDriver.cs
DrewBrasher: I did that because it was how the anchor was done.

src/OrchardCore.Modules/OrchardCore.Media/Drivers/MediaFieldDriver.cs
DrewBrasher: @ns8482e  This only applies to attached media. Do you still want the field on [src/OrchardCore/OrchardCore.Media.Core/Fields/MediaField.cs](https://github.com/OrchardCMS/OrchardCore/blob/c02fedd8c3bf1bb1b892e3c5fc0c000d27a9ca80/src/OrchardCore/OrchardCore.Media.Core/Fields/MediaField.cs)?

DrewBrasher: Thank you, I just wanted to make sure.

deanmarcussen: > Add property to media field and use here

I'd suggest not doing this @ns8482e .

The value is only going to be supplied when using the attached media field editor. Not the standard editor.

So the design choice there is to attach things to the field, as extra meta data with extension methods.

ns8482e: @DrewBrasher @deanmarcussen sounds good

sebastienros: Also using a metadata and not adding a property to the field make it possible to create this as a separate feature. The feature would add the UI to edit and store the metadata. Right now it seems like it's not a separate feature, but at least it's not breaking anything, as long as the value is not mandatory.

sebastienros:
Is it breaking anything in its current state when migrating from a previous version? What happens if the metadata is not preset?

DrewBrasher:
> Is it breaking anything in its current state when migrating from a previous version? What happens if the metadata is not preset?

@sebastienros It does not seem to be breaking anything when migrating from a previous version. Files that were attached before the migration will have the metadata added with a null value.

Here is what I did to test it:

1. Checked out the release/1.4 branch and ran the cms.web project.
2. Created a site with the Agency Theme and added a content type with a media field set to use the attached editor
3. Created a new item, attached a document, and published.
4. Checked out my branch and ran the cms.web project.
5. Viewed the content item and inspected the file object using the Vue explorer in Chrome.
image
6. Attached another file to the same content item as before and published.
7. Viewed the content item and inspected the file objects using the Vue explorer in Chrome.
image

PiemP:
Resolve #4563

Issue #10575, opened 2021-10-28
BackgroundTasks Index action not showing pager


tropcicstefan:
### Describe the bug
BackgroundTaskController Index action not showing pager at https://.../Admin/BackgroundTasks when having more Tasks than default site settings PageSize.
E.g. if there are 2 tasks and PageSize is set to 1, one task won't be on the list and is only accessible by adding query param 'page=2' because pager wont render.

### To Reproduce
1. Set Configuration->Settings->General Page Size to 1
2. Create 2 Background Tasks and go to Configuration->Tasks->Background Tasks
3. Only one is visible and no pager

### Expected behavior
Pager should be rendered

### Screenshots
![image](https://user-images.githubusercontent.com/30324480/139333265-e0151a6f-a8f3-4579-b6ad-ca05bffc8a4e.png)

PR #10576, opened 2021-10-28
BackgroundTaskController count query looked at wrong collection

tropcicstefan:
#10575 

image

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.BackgroundTasks/Controllers/BackgroundTaskController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.BackgroundTasks/Controllers/BackgroundTaskController.cs
@@ -69,7 +69,7 @@ namespace OrchardCore.BackgroundTasks.Controllers
             .Take(pager.PageSize)
             .ToList();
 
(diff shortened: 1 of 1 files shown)


tropcicstefan:
> Wait, is it not intended that the taskEntries are used instead of the _backgroundTasks here?

taskEntries is always limited by .Take and will always give limited results, so no

> Basically, it will list only Background Tasks which have BackgroundTaskSettings.

NO,  in `if` there is `settings = t.GetDefaultSettings();` which will show all BT with default settings which correlates with `ModularBackgroundService` where default settings are also applied



> The question is, will it fail to save any BackgroundTaskSettings on these background tasks if we don't provide any default?

It won't fail to save, because Edit action checks if settings are saved for that BT in document, if not it redirects to Create action.

> And is it why it's been listing only those

By now I hope you see there is a bug.

P.S.
On a lighter note, @hishamco do you have something against me, or you just like adding dontmerge labels on my PRs 🤣

jtkech:
@tropcicstefan 

> taskEntries is always limited by .Take and will always give limited results, so no

Good catch ;)

jtkech (approved):

Skrypt:
I see where the issue is now. And you are right. Merged.

PR #10581, opened 2021-10-29
Avoid NRE in tenant creation

hishamco:
While I'm working on other PR, I just found a NRE when I create a new tenant, this is because the feature profles is `null`

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
@@ -324,10 +324,11 @@ namespace OrchardCore.Tenants.Controllers
                 await ValidateViewModel(model, true);
             }
 
-            // Creates a default shell settings based on the configuration.
-            var shellSettings = _shellSettingsManager.CreateDefaultSettings();
             if (ModelState.IsValid)
             {
+                // Creates a default shell settings based on the configuration.
+                var shellSettings = _shellSettingsManager.CreateDefaultSettings();
+
                 shellSettings.Name = model.Name;
                 shellSettings.RequestUrlHost = model.RequestUrlHost;
                 shellSettings.RequestUrlPrefix = model.RequestUrlPrefix;
@@ -350,7 +351,7 @@ namespace OrchardCore.Tenants.Controllers
             var recipeCollections = await Task.WhenAll(_recipeHarvesters.Select(x => x.HarvestRecipesAsync()));
             var recipes = recipeCollections.SelectMany(x => x).Where(x => x.IsSetupRecipe).OrderBy(r => r.DisplayName).ToArray();
             model.Recipes = recipes;
-            model.FeatureProfiles = await GetFeatureProfilesAsync(shellSettings["FeatureProfile"]);
+            model.FeatureProfiles = await GetFeatureProfilesAsync(model.FeatureProfile);
 
             // If we got this far, something failed, redisplay form
             return View(model);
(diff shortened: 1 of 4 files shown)


jtkech:
@hishamco @deanmarcussen

Yes I can confirm, if initially there is no `FeatureProfiles` in the `Create / Edit` post actions the `FeatureProfiles` of the bounded `EditTenantViewModel model` is null, so if it fails to validate we do `return View(model)` to re-display the form and then we get an NRE in `Create / Edit.cshtml` on `Model.FeatureProfiles.Any()`.

So maybe we not only need a null check in `Create.cshtml` but also in `Edit.cshtml`.

hishamco:
Thanks @jtkech , frankly I didn't check the edit view but seems you confirm that NRE occurs

jtkech:
@hishamco 

Look at my updated comment above, I think it's beter to re-init the `FeatureProfiles` list in place of adding null checks.

hishamco:
> Look at my updated comment above, I think it's beter to re-init the FeatureProfiles list in place of adding null checks.

I already suggested that too ;) I will update the PR

jtkech:
@hishamco 

See the updated part of my previous comment

> Or better I think, in the post actions, if it fails to validate we need to re-populate the FeatureProfiles list, as we do in the related get actions, even if it ends up in an empty list (but not null).

In fact I was thinking about re-populating the FeatureProfiles list, as it is done in the related get actions, not just init it as an empty list through a new ctor.

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -329,11 +324,10 @@ public async Task<IActionResult> Create(EditTenantViewModel model)
                 await ValidateViewModel(model, true);
             }
 
+            // Creates a default shell settings based on the configuration.
+            var shellSettings = _shellSettingsManager.CreateDefaultSettings();
```
deanmarcussen, line 328: don't move this

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -356,6 +350,7 @@ public async Task<IActionResult> Create(EditTenantViewModel model)
             var recipeCollections = await Task.WhenAll(_recipeHarvesters.Select(x => x.HarvestRecipesAsync()));
             var recipes = recipeCollections.SelectMany(x => x).Where(x => x.IsSetupRecipe).OrderBy(r => r.DisplayName).ToArray();
             model.Recipes = recipes;
+            model.FeatureProfiles = await GetFeatureProfilesAsync(shellSettings["FeatureProfile"]);
```
deanmarcussen, line 353:
```suggestion
            model.FeatureProfiles = await GetFeatureProfilesAsync(model.FeatureProfile);
```

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -485,6 +476,7 @@ public async Task<IActionResult> Edit(EditTenantViewModel model)
...
```
deanmarcussen, line 479:
```suggestion
            model.FeatureProfiles = await GetFeatureProfilesAsync(model.FeatureProfile);
```

deanmarcussen (commented):
ok, so from what @jtkech says and I can see you identified an issue when the form is invalid.
So the suggestion is to use the values from the model, as it was presented on the form post.

deanmarcussen (approved):

Issue #10402, opened 2021-10-04
Tenants Classification


hishamco:
> Tenants enhancement: In the case of too many tenants, whether to support the classification of the tenant list

I think it's good idea to categorize the tenants when the list get larger

related to https://github.com/OrchardCMS/OrchardCore/discussions/10397

PR #10586, opened 2021-10-29
Categorized tenants

hishamco:
Addresses #10402

Screenshot 2021-10-29 182651

- [x] Sorting
- [x] Filtering
- [x] Pagination
- [x] Searching


hishamco:
Screenshot 2021-10-29 213421

src/OrchardCore/OrchardCore.Abstractions/Shell/ShellSettings.cs
```diff
@@ -38,6 +38,12 @@ public ShellSettings(ShellSettings settings)
 
         public string Name { get; set; }
 
+        public string Category
+        {
+            get => _settings[nameof(Category)];
+            set => _settings[nameof(Category)] = value;
+        }
```
jtkech, line 45: `ShellSettings` holds 2 `ShellConfiguration` config sources

https://github.com/OrchardCMS/OrchardCore/blob/dbd83c6f226c15db5e6ba82b0e2c1609aa693d77/src/OrchardCore/OrchardCore.Abstractions/Shell/ShellSettings.cs#L20-L24

For both of them, a value can be pre-configured e.g. through the main `appsettings.json`, but when a value is added / updated at runtime, e.g. through a setup or while editing a tenant, a value updated in `_configuration` will be persisted in the tenant specific `appsettings.json` under the tenant folder (can be configured to be in an azure blob or database), a value updated in `_settings` will be persisted in the `tenants.json` (shared by all tenants) directly under the `App_Data` folder (can be configured to be elsewhere, in an azure blob or database).

The `tenants.json` file (related to `_settings`) is intended to hold the minimum set of values (not already pre-configured) required to init a tenant, the other values held by tenant specific sources (related to `_configuration`) being used to lazily build a tenant.

So here for `Category` better to not use a dedicated field using `_settings`, but instead the following indexer that uses `_configuration`, as we already do e.g. for the tenant `Description`.

https://github.com/OrchardCMS/OrchardCore/blob/dbd83c6f226c15db5e6ba82b0e2c1609aa693d77/src/OrchardCore/OrchardCore.Abstractions/Shell/ShellSettings.cs#L69-L74

So nothing to do here, and in the other place just do what we already did for the tenant `Description`.

hishamco: Thanks for the info @jtkech, I will use indexer instead

deanmarcussen:
I don't like `Not Categorized`. For any existing user it would force them to go through and categorise all their tenants, just to get rid of an odd bit of text.

Categories (if we do it), should be optional.
So just use `String.Empty`, and `None` in the filter list (like we do elsewhere`), and don't show anything if it is `String.Empty`

hishamco:
> And equally don't show the filter if there are no categories.

So, how can we list the tenants that they don't have category?

> And equally don't show the filter if there are no categories.

+1

deanmarcussen:
> So, how can we list the tenants that they don't have category?

you would have an All. It would be the default.

We would never have a filter that forces people to put things into categories.

hishamco:
> you would have an All. It would be the default.

No make sense, coz when you said None before ;) I will update the PR soon ...

hishamco:
Any updates on this?

sebastienros:
Can you update the screenshots if you removed the "not categorized" ?

hishamco:
Tenants categories

hishamco:
Please don't merge after I double check the merge conflict

deanmarcussen:
@hishamco you have missed my previous review.

Where I said that `None` shouldn't be showing.

If people aren't wanting to categorise their tenants (I can't think of any category I'd give mine, they are all of the same type), then having a ranom None tag showing up makes no sense.

i.e. Ok, with categorising, not ok with forcing it on users.

hishamco:
Oops, seems I forgot this one, I will update the PR soon ...

hishamco:
If there's no other comment, then the PR is ready to merge

src/OrchardCore.Modules/OrchardCore.Tenants/ViewModels/AdminIndexViewModel.cs
```diff
@@ -59,7 +69,7 @@ public class TenantIndexOptions
         public List<SelectListItem> TenantsBulkAction { get; set; }
     }
 
-    public enum TenantsFilter
+    public enum TenantsStatus
```
deanmarcussen, line 72: If we're going to rename this is should be `TenantState`  or `TenantsState` not status.

Otherwise looks good. 
Can you provide a new graphic please.

hishamco: > Can you provide a new graphic please.

For? I already attached the latest screenshoort [here](https://github.com/OrchardCMS/OrchardCore/pull/10586#issuecomment-991077437)

deanmarcussen: Because the last screenshot didn't make sense, and I had to comment on it requesting more changes.

So I can see what you have done this time, and make sure it makes sense, please include a new screenshot

hishamco: Sorry, seems I removed the `None` badge as you suggested earlier. I need to update the screenshot

hishamco:
Tenants categories

hishamco:
@deanmarcussen anything left here or can we approve then merge?

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -120,15 +123,20 @@ public async Task<IActionResult> Index(TenantIndexOptions options, PagerParamete
                      (t.ShellSettings.RequestUrlPrefix != null && t.ShellSettings.RequestUrlPrefix.IndexOf(options.Search, StringComparison.OrdinalIgnoreCase) > -1)))).ToList();
             }
 
-            switch (options.Filter)
+            if (!String.IsNullOrWhiteSpace(options.Category) && options?.Category != DefaultCategory)
+            {
+                entries = entries.Where(t => t.Category == options.Category).ToList();
```
deanmarcussen, line 128: Since they are inserted (on the edit screen) manually I think you should use ordinal ignore case here.
Otherwise alpha and Alpha are different categories which would be odd

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -167,11 +176,19 @@ public async Task<IActionResult> Index(TenantIndexOptions options, PagerParamete
             };
 
             // We populate the SelectLists
+            model.Options.TenantsCategories = allSettings
+                .GroupBy(t => t["Category"])
+                .Where(t => !String.IsNullOrEmpty(t.Key))
+                .Select(t => new SelectListItem(t.Key, t.Key))
+                .ToList();
+
+            model.Options.TenantsCategories.Insert(0, new SelectListItem(S["All"], DefaultCategory, selected: true));
```
deanmarcussen, line 185: This is only selected:true when it’s all or empty isn’t it?

so here and in the above you need to do a comparison against the currently selected value to set the is selected property

deanmarcussen (commented):
Two minor issues to fix. Otherwise looks good

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
hishamco: I think no need to ignore the comparison casing here because the `DefaultCategory` is not editable. Regarding the check you are right need to be changed

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -167,11 +175,22 @@ public async Task<IActionResult> Index(TenantIndexOptions options, PagerParamete
             };
 
             // We populate the SelectLists
+            model.Options.TenantsCategories = allSettings
+                .GroupBy(t => t["Category"])
+                .Where(t => !String.IsNullOrEmpty(t.Key))
+                .Select(t => new SelectListItem(t.Key, t.Key))
```
deanmarcussen, line 181: here is where you need to add the comparison, otherwise when these items are selected they will never be shown as selected.

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -167,11 +175,22 @@ public async Task<IActionResult> Index(TenantIndexOptions options, PagerParamete
...
+                .ToList();
+
+            model.Options.TenantsCategories.Insert(0, new SelectListItem(
+                S["All"],
+                DefaultCategory,
```
deanmarcussen, line 186: You don't need `DefaultCategory` at all. when it is Empty it is the default category. Just remove it

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
hishamco: I didn't get it, I'm just checking if the category is `null` or empty, could you please write a snippet of what you mean

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
hishamco: Right

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
hishamco:
```csharp
model.Options.TenantsCategories = allSettings
    .GroupBy(t => t["Category"])
    .Where(t => !String.IsNullOrEmpty(t.Key) && !t.Key.Equals(DefaultCategory, StringComparison.OrdinalIgnoreCase))
    .Select(t => new SelectListItem(t.Key, t.Key))
    .ToList();
```

Is it what you mean?

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
deanmarcussen, line 186:
```suggestion
                String.Empty,
```

And remove `DefaultCategory` - it works, and we do this everywhere else.

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
deanmarcussen, line 181:
```
model.Options.TenantsCategories = allSettings
    .GroupBy(t => t["Category"])
    .Where(t => !String.IsNullOrEmpty(t.Key))
    .Select(t => new SelectListItem(t.Key, t.Key, String.Equals(model.Category, t.Key, StringComparison.OrdinalIgnoreCase)))
    .ToList();
```

where model.Category = Selected Category, and you set whether the `SelectListItem` is selected or not

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
hishamco: Same thing

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
hishamco: Go it, you want to set th select property, I will update the PR

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -167,11 +175,22 @@ public async Task<IActionResult> Index(TenantIndexOptions options, PagerParamete
...
+            model.Options.TenantsCategories = allSettings
+                .GroupBy(t => t["Category"])
+                .Where(t => !String.IsNullOrEmpty(t.Key))
+                .Select(t => new SelectListItem(t.Key, t.Key, String.Equals(options.Category, t.Key, StringComparison.OrdinalIgnoreCase)))
+                .ToList();
+
+            model.Options.TenantsCategories.Insert(0, new SelectListItem(
+                S["All"],
+                String.Empty,
+                selected: String.IsNullOrEmpty(options.Category) || model.Options.Category == DefaultCategory));
```
deanmarcussen, line 187:
```suggestion
                selected: String.IsNullOrEmpty(options.Category));
```

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -120,15 +122,20 @@ public async Task<IActionResult> Index(TenantIndexOptions options, PagerParamete
                      (t.ShellSettings.RequestUrlPrefix != null && t.ShellSettings.RequestUrlPrefix.IndexOf(options.Search, StringComparison.OrdinalIgnoreCase) > -1)))).ToList();
             }
 
-            switch (options.Filter)
+            if (!String.IsNullOrWhiteSpace(options.Category) && options?.Category != DefaultCategory)
```
deanmarcussen, line 125:
```suggestion
            if (!String.IsNullOrWhiteSpace(options.Category))
```

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -20,13 +20,14 @@
...
 using OrchardCore.Routing;
 using OrchardCore.Settings;
-using OrchardCore.Tenants.Services;
 using OrchardCore.Tenants.ViewModels;
 
 namespace OrchardCore.Tenants.Controllers
 {
     public class AdminController : Controller
     {
+        private static readonly string DefaultCategory = "All";
```
deanmarcussen, line 29: Remove

deanmarcussen (commented):
You forgot to remove the default category - it's unnecesary

hishamco:
Seems I forgot to push a local commit from awhile ;)

deanmarcussen (approved):

PR #10601, opened 2021-10-31
Fix audit trail date parsing (greater than or equal to)

deanmarcussen:
Working on something else and realised the `>=` literals need to be parsed first.

Should really fix the other tests at some point to (but they work in actual real life, so another pr).


hishamco (approved):

hishamco:
> Should really fix the other tests at some point to (but they work in actual real life, so another pr).

I could handle this, once this got merge

PR #10618, opened 2021-11-01
Make https url first in launchsettings

Skrypt:
So that it doesn't try to open browser on https://localhost:5000 when debugging with VS Code.

### DIFF
--- a/src/OrchardCore.Cms.Web/Properties/launchSettings.json
+++ b/src/OrchardCore.Cms.Web/Properties/launchSettings.json
@@ -22,7 +22,7 @@
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development"
       },
-      "applicationUrl": "http://localhost:5000;https://localhost:5001"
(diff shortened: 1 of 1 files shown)


hishamco:
@Skrypt just a question related to VS Code, how can I unload or disable some projects, coz when I open OC using VS Code it takes much memory, which leads to hand my laptop ;)

Any idea?

Skrypt:
Try disabling IntelliSense.

hishamco:
Really is that matter, I think the issue is all the ~ 175 projects are compiled during the build

Skrypt:
Sure, that's a problem too. I'm not sure your PR would work with VS Code. But there is definitely something about dotnet restore too that executes when opening the project to load intellisense too.

hishamco:
> using source code will always take a lot of resources.

Exactly, I see multiple instances of .NET Host which take a lot of RAM!! I like Visual Studio because you can unload or remove some projects from build process

PR #10638, opened 2021-11-05
Starting background tasks without waiting for initialized tenants

jersiovic:


jtkech:
@jersiovic 

Thanks, I will review it asap, but looks like so simple ;)

I think we need at least a config option to lazily build or not tenants

Something like under the `OrchardCore` section => `OrchardCore_Shells` section => `LazyBuild`

Hmm, and maybe allow to set it per tenant

jersiovic:
> I think we need at least a config option to lazily build or not tenants

But the feature I added doesn't change if the tenant is lazily build, in fact, I don't see the problem with lazy build them on a first request, it is fine, cause it answer the request as soon it is build and now it doesn't stop background tasks from running. 
@jtkech Do you still want to make it configurable? Is there any reason?

jtkech:
> But the feature I added doesn't change if the tenant is lazily build, in fact, I don't see the problem with lazy build them on a first request, it is fine, cause it answer the request as soon it is build and now it doesn't stop background tasks from running.

On the first request on any tenant, we don't fully build all tenants, we only pre-create them as done in `PreCreateAndRegisterShellsAsync()`, this to save resource e.g. if there are many tenants.

If we don't check the tenant pipeline anymore, when a bg task will start on a given tenant we will get a scope on its shell, if its shell is only pre-created, it will be fully built even if no request has been done yet on this tenant, which we may not want if there are many tenants (and that have at least one bg task e.g. for lucene indexing).

jersiovic:
> > But the feature I added doesn't change if the tenant is lazily build, in fact, I don't see the problem with lazy build them on a first request, it is fine, cause it answer the request as soon it is build and now it doesn't stop background tasks from running.
> 
> On the first request on any tenant, we don't fully build all tenants, we only pre-create them as done in `PreCreateAndRegisterShellsAsync()`, this to save resources e.g. if there are many tenants.
> 
> If we don't check the tenant pipeline anymore, when a bg task will start on a given tenant we will get a scope on its shell, if its shell is only pre-created, it will be fully built even if no request has been done yet on this tenant, which we may not want if there are many tenants (and that have at least one bg task e.g. for lucene indexing).

What do you think of this other alternative? Instead of global config option we can add to IBackgroundTask a property to indicate if the task can wait for the tenant lazy build when the first request arrive or if it should force the tenant build to start when orchard is started?
We can make this property configurable from the dashboard UI for background tasks

jersiovic:
done

src/OrchardCore/OrchardCore/Modules/ModularBackgroundService.cs
```diff
@@ -54,6 +64,12 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
                 _logger.LogInformation("'{ServiceName}' is stopping.", nameof(ModularBackgroundService));
             });
 
+            if (_lazyBuildDisabled)
+            {
+                // Ensure all ShellContext are loaded and available.
+                await _shellHost.InitializeAsync();
```
jtkech, line 70: Okay so would simulate a first request

jtkech:
Okay looks good, just suggested a little change

As I remember before checking the shell Pipeline we were doing a null check on its ServiceProvider.

I think that's okay but just need to think about it a little.

src/OrchardCore/OrchardCore/Modules/ModularBackgroundService.cs
```diff
@@ -34,17 +36,25 @@ internal class ModularBackgroundService : BackgroundService
...
             _shellHost = shellHost;
             _httpContextAccessor = httpContextAccessor;
             _logger = logger;
             _clock = clock;
+
+            _lazyBuildDisabled = configuration
+                .GetSection("OrchardCore")
+                .GetSectionCompat("OrchardCore_Shells_LazyBuild")
+                .GetSectionCompat("Disable")
+                .Get<bool>();
```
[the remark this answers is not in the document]
jtkech: Note: `GetSectionCompat()` is only useful when the key contains underscores.

jtkech:
@jersiovic 

If you agree with my suggested change (see the review comment)

Can you apply it? Then I will approve

src/OrchardCore/OrchardCore/Modules/ModularBackgroundService.cs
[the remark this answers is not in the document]
jtkech: I also suggest to use `Disabled` for the key, in place of `Disable`

src/OrchardCore/OrchardCore/Modules/ModularBackgroundService.cs
```diff
@@ -34,17 +36,25 @@ internal class ModularBackgroundService : BackgroundService
...
         private readonly IClock _clock;
+        private readonly bool _lazyBuildDisabled;
 
         public ModularBackgroundService(
             IShellHost shellHost,
             IHttpContextAccessor httpContextAccessor,
             ILogger<ModularBackgroundService> logger,
-            IClock clock)
+            IClock clock,
+            Microsoft.Extensions.Configuration.IConfiguration configuration)
```
sebastienros, line 46: Shouldn't it be an options class?

jtkech: Yes, useful if multiple option fields and/or injected in multiple places. But not necessarily, we can get any config section "field" separately. But maybe still better to use an option class even if only one field and only used here. Note: Can be done later on without breaking change.

I'am also thinking about the section name `OC_Shells_LazyBuild` and its field `Disable`, maybe an `OC_Shells` section with a new `ShellsOptions` with a first `LazyBuild` field. Note: We already have a `ShellOptions` (for `App_Data` config) but only used at the tenant level, here it is at the host level.

Hmm, but maybe it is a background service option, e.g. if there is no bg task defined, even if you disable the lazy build option, a given shell will be still lazy built on a first request. So here it is to say to the bg service (if there are some bg tasks) that it can warmup a shell by executing a bg task.

I will try to think about it and suggest something this week end.

jersiovic:
> Can you apply it? Then I will approve

done

src/OrchardCore/OrchardCore/Modules/ModularBackgroundService.cs
```diff
@@ -34,17 +36,24 @@ internal class ModularBackgroundService : BackgroundService
...
         {
             _shellHost = shellHost;
             _httpContextAccessor = httpContextAccessor;
             _logger = logger;
             _clock = clock;
+
+            _lazyBuildDisabled = configuration
+                .GetSection("OrchardCore")
+                .GetSectionCompat("OrchardCore_Shells_LazyBuild")
+                .GetValue("Disabled", true);
```
deanmarcussen, line 56: Shouldn’t this be false by default?

jtkech: Yes you're right, because here true means that lazy build would be `Disabled` ;)

My other little concern is that maybe it is better to make it a bg service config option

I will think about it this night

jtkech:
@jersiovic Sorry for the delay

Just did some minor changes locally but didn't have enough time to complete / commit them.

Will do it tomorrow night, then you will be able to revert them if you don't agree.

jtkech:
@jersiovic Here some last thoughts but that should not block this PR.

Yes, `_shellHost.InitializeAsync()` is a warmup where features and shells settings are loaded, but only light shells are pre-created, they are still not yet fully built (see below). So could be a 1st option, maybe a `Shells` option even if we don't use a dedicated shell warmup hosted service, or consider it as a `BackgroundService` behavior.

Then here this is on the first call of `_shellHost.GetScopeAsync()` before a background task that a given tenant will be fully built (still not its pipeline). Could be a 2nd option but here more like a `BackgroundService` option allowing to eagerly build a shell before executing a background task.

But to make it simpler, knowing that the 2nd option requires the 1st one (but not the opposite), that's okay to only have one option, but then the whole is more clearly a `BackgroundService` option.

Anyway I can't neither add nor edit any file on your branch, so not easy to show what I mean, so here the simplest compromise that I found and again that should not block this PR.

- So okay to only have one option, but more like a background service option allowing to warmup shells and then allowing to eagerly build a shell (if not yet done) just before first executing one of its background task.

- Modules can't participate to this config option by code, because would need to be done at the app / host level, so that's okay to still not use the `IConfigureOptions<>` pattern, as done e.g. for `DatabaseShellsStorageOptions`.

- But still useful to have a `BackgroundServiceOptions` as suggested by @sebastienros so that it will be easier to add future option fields.

As said I can't edit your files, so here my suggested changes, feel free to not use or adjust them if you don't agree.

**appsettings.json**

    // By default background tasks are waiting for their shell to be lazily built on a first matching request.
    //"OrchardCore_BackgroundService": {
    //  "ShellWarmup": true // Allows to eagerly build shells just before executing one of their background tasks.
    //},

**BackgroundServiceOptions.cs** maybe in the `Abstrations` project, but I'm fine to keep it in the same project, or even more to define it in the same file.

    public class BackgroundServiceOptions
    {
        public bool ShellWarmup { get; set; }
    }


**ModularBackgroundService.cs**

        ...

        private readonly BackgroundServiceOptions _options;

        ...
            _options = configuration
                .GetSection("OrchardCore")
                .GetSectionCompat("OrchardCore_BackgroundService")
                .Get<BackgroundServiceOptions>()
                ?? new BackgroundServiceOptions();
        ...

            if (_options.ShellWarmup)
            {
                // Ensure all Shells are loaded and available.
                await _shellHost.InitializeAsync();
            }

            // And so on, replacing '_lazyBuildDisabled' by '_options.ShellWarmup'

Finally, if we let this PR as it is without applying the above suggestions, at least, as stated by @deanmarcussen,  change the default value of your current option from true to false.

jersiovic:
Changes proposed done. Thank you guys

src/OrchardCore/OrchardCore/Modules/ModularBackgroundService.cs
```diff
@@ -104,10 +121,10 @@ private async Task RunAsync(IEnumerable<ShellContext> runningShells, Cancellatio
                     {
                         break;
                     }
-
+
```
jtkech, line 124: Remove extra spaces

jtkech (approved):

jtkech:
@jersiovic 

Thanks for the changes, I approved the PR

Just some extra spaces to remove if you have time but not crucial ;)

src/OrchardCore/OrchardCore/Modules/ModularBackgroundService.cs
```diff
@@ -34,17 +37,25 @@ internal class ModularBackgroundService : BackgroundService
...
-            IClock clock)
+            IClock clock,
+            Microsoft.Extensions.Configuration.IConfiguration configuration)
         {
             _shellHost = shellHost;
             _httpContextAccessor = httpContextAccessor;
             _logger = logger;
             _clock = clock;
+
+            _options = configuration
```
deanmarcussen, line 54: Shouldn’t this be registered as an IOptions?

here it’s just reading from IConfiguration directly?

The naming here looks much better thanks @jersiovic

jtkech: > Shouldn’t this be registered as an IOptions?

Yes could be and then e.g. a default config setup that would read from IConfiguration.

But as said in my last comment, this option is only used here and because it is at the app level other modules can't collaborate to set it by code, would need to be done by app level code.

So to keep it simpler I think it is okay to not use the `IConfigureOptions<>` pattern just to do the same thing (read from configuration), as we do for `DatabaseShellsStorageOptions`.

But I'm not against ;)

deanmarcussen: Since we’ve added an Options class it would make sense to me. 

Then it is actually configurable via the class as well, otherwise the class doesn’t do anything I think?

In this way you can configure it via appsettings, or at app level in code.

not to worried either way @jtkech and @jersiovic  just saw the class and couldn’t figure why we’d add the class without registering it.

jtkech: Okay cool then, could be a simple `services.Configure<>()` in the `AddBgService()` helper.

I will suggest the few lines of code to do it (can't edit the files)

jtkech: > otherwise the class doesn’t do anything I think?

Yes and no, the class is an helper in itself when binding to a `BackgroundServiceOptions` field, so if we add more options to this class, we will not have to change the constructor code.

jtkech:
@jersiovic 

Sorry but as a last update to do

In the already existing `AddBackgroundService()` helper

        public static OrchardCoreBuilder AddBackgroundService(this OrchardCoreBuilder builder)
        {
            builder.ApplicationServices.AddHostedService<ModularBackgroundService>();

            builder.ApplicationServices
                .AddOptions<BackgroundServiceOptions>()
                .BindConfiguration("OrchardCore:OrchardCore_BackgroundService");

            return builder;
        }

Then in the `ModularBackgroundService` ctor just inject the options.

        public ModularBackgroundService(
            IShellHost shellHost,
            IOptions<BackgroundServiceOptions> options,
            ILogger<ModularBackgroundService> logger,
            IClock clock)
        {
            _shellHost = shellHost;
            _options = options.Value;
            _logger = logger;
            _clock = clock;
        }

saleemhaq:
admin can you please approved that pr because i need to achieve by that way here is my question raised regarding this: https://github.com/OrchardCMS/OrchardCore/discussions/10938

jtkech:
@jersiovic 

Okay, I can't edit your files but no problem I will provide a little PR

This just to use the last suggested `IOptions<>` pattern (no breaking change)

So in the meantime I merge this one

saleemhaq:
So now it workable without having to need the first request?

jtkech:
Yes but this is still not the default

You need to add the following option in your `appsettings.json` as commented in our source code

    // By default background tasks are waiting for their shell to be lazily built on a first matching request.
    //"OrchardCore_BackgroundService": {
    //  "ShellWarmup": true // Allows to eagerly build shells just before executing their first background task.
    //},

And you will need to reference the last packages from the cloudsmith feed

But wait a little, I'm doing a little patch #10943 related to this that I will merge very soon.

agriffard:
Any chance you can add some doc on https://docs.orchardcore.net/en/latest/docs/reference/modules/BackgroundTasks/ @jersiovic ?

saleemhaq:
Can you please merge it by today if possible?

Thanks..

jersiovic:
Thank you guys, I was on vacation, I see you merged it

PR #10671, opened 2021-11-08
.NET 6.0 docs and templates

agriffard:
Update version to 6.0 in docs and templates help.


hishamco (commented):
I think we should update templates to 1.1.0, right?

PR #10673, opened 2021-11-08
Make UpdateAtomic Delegates scoped and distinct

### LINKED ISSUE #10672: Make UpdateAtomic Delegates scoped and distinct
Related to `IVolatileDocumentManager.UpdateAtomicAsync()` that for now is never used in OC.

- One problem since we made `IVolatileDocumentManager` a singleton is that the same delegate list is always used / executed and grow up, **so the delegate list should be still held at the scope level which I forgot to do**.

- The 2nd problem is when enlisting multiple delegates (not allowed with `UpdateAsync()`), we can enlist multiple time the same delegate (targetting the same method and so on), not a problem most of the time but not good when we do things by batches where the same one can be added many times, **so the delegate list should only contain distinct delegates**.

jtkech:
Fixes #10672

Related to `IVolatileDocumentManager.UpdateAtomicAsync()` that for now is never used in OC.

- One problem since we made `IVolatileDocumentManager` a singleton is that the same delegate list is always used / executed and grow up, **so the delegate list should be still held at the scope level which I forgot to do**.

- The 2nd problem is when enlisting multiple delegates (not allowed with `UpdateAsync()`), we can enlist multiple time the same delegate (targetting the same method and so on), not a problem most of the time but not good when we do things by batches where the same one can be added many times, **so the delegate list should only contain distinct delegates**.

### DIFF
--- a/src/OrchardCore/OrchardCore.Infrastructure/Documents/VolatileDocumentManager.cs
+++ b/src/OrchardCore/OrchardCore.Infrastructure/Documents/VolatileDocumentManager.cs
@@ -1,10 +1,12 @@
 using System;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Caching.Distributed;
 using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Options;
 using OrchardCore.Data.Documents;
 using OrchardCore.Documents.Options;
+using OrchardCore.Environment.Shell.Scope;
 using OrchardCore.Locking.Distributed;
 
 namespace OrchardCore.Documents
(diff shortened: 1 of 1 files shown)


src/OrchardCore/OrchardCore.Infrastructure/Documents/VolatileDocumentManager.cs
```diff
@@ -82,5 +88,11 @@ public Task UpdateAtomicAsync(Func<Task<TDocument>> updateAsync, Func<TDocument,
 
             return Task.CompletedTask;
         }
+
+        private class UpdateDelegates
```
sebastienros, line 92: That is called a `Tuple`, there are nice syntaxes in C# for that, and you can even do a struct record otherwise, but tuples are better here.

sebastienros (approved):

src/OrchardCore/OrchardCore.Infrastructure/Documents/VolatileDocumentManager.cs
jtkech: Yes good ideas

But for the struct, I'm using a `GetOrCreateFeature()` that is waiting for a class, then for the tuple, because we can pass a factory I also tried a `Tuple.Create<UpdateDelegate, AfterUpdateDelegate>(null, null))`, but then we can't do `delegates.Item1 += () => updateAsync()` as it is readonly.

PR #10674, opened 2021-11-09
Update YesSql

sebastienros:


sebastienros:
To be discussed. Current Sqlite 6.0.0 will fail on netcoreapp3.1, on some environments.

deanmarcussen:
It might be good to do another release with 6.0 before dropping 3.1.

src/OrchardCore.Modules/OrchardCore.Workflows/Http/Scripting/HttpMethodsProvider.cs
```diff
@@ -174,13 +173,13 @@ public HttpMethodsProvider(IHttpContextAccessor httpContextAccessor)
                                     throw new Exception("Invalid form data passed in the request.");
                                 }
                             }
-                            else if (HasJsonContentType(httpContextAccessor.HttpContext.Request))
```
agriffard, line 177: Then, remove `HttpMethodsProvider.HasJsonContentType` ?

sebastienros: I did, didn't I ?

sebastienros:
I handled the issue with netcoreapp3.1 by using the previous sqlite version in this case.

PR #10677, opened 2021-11-09
Better NRE handling for Widget and MenuItem stereotypes

### LINKED ISSUE #1014: NullReferenceException on editor if embedded content type definiton is deleted
1. Create site using Agency recipe.
2. Go to Content Definition -> Content Types menu.
3. Remove Milestone type definition.
4. Try to edit or create new _Agency Start Bootstrap Theme_ (landing page).
5. Observe NullReferenceException in `ContentDisplayManager.BuildEditorAsync()`

When you delete Milestone `ContentType`, Orchard probably doesn't check if it's contained in any Bags. When trying to edit a `ContentItem` containing Milestone, `ContentDisplayManager.BuildEditorAsync()` cannot find a proper `ContentDefinition` for still existing parts of the ContentItem, but still tries to extract metadata from it.

The solution would be one of the following:
1. check for all `ContentItems` of a type and delete them
2. forbid deleting the `ContentType` until all its items are deleted, including embedded items
3. delete the `ContentType`; if Orchard later stumbles upon a `ContentItem` with unknown definition, simply ignore it
Since solutions 1 and 2 are harder to implement and can incur severe performance penalties, I think it's best to simply ignore items of unknown type/definition.

### LINKED ISSUE #1794: Error when editing menu containing menu items whose content type has been deleted
Repro steps:

1. Create a new content type with stereotype = "MenuItem".
2. Add a menu item of that type to a menu.
3. Delete the content type created in step 1.
4. Go back to the menu you created in step 2.
5. Observe a Null Reference Exception.

### LINKED ISSUE #6751: Deleteing a content-type that is inuse causes an exception

### LINKED ISSUE #7963: NRE When Displaying Content Items That Have Null Content Type Definition
Similar to Issue #1794.

If one creates a new tenant using _TheBlogTheme_ and subsequently delete the `RawHtml` Content Type, one will receive a null reference exception both on the frontend and in the backend when [ContentDisplayManager](https://github.com/OrchardCMS/OrchardCore/blob/b5fa5f6e73526e58050482b7ccdea0c640232409/src/OrchardCore/OrchardCore.ContentManagement.Display/ContentDisplayManager.cs#L63) interrogates a null content type definition when it attempts to display instances of `RawHtml` that were not deleted before deleting the content type.

```csharp
var contentTypeDefinition = _contentDefinitionManager.GetTypeDefinition(contentItem.ContentType); // This will be null

var stereotype = contentTypeDefinition.GetSettings<ContentTypeSettings>().Stereotype; // NRE
```

I suspect this exception will occur more often when experimenting than in actual production, but in general, **it's assumed that when displaying a content item that the content type definition will not be null**, which can make the frontend and backend fairly unusable when it is null. One can't actually fix the problem that I know of other than to realize by some luck that a content type definition has been deleted and add it back so one can then go back and delete the offending content items.

I'm wondering if there is a more graceful solution that can handle this case, such that one can be better notified of the problem, possibly avoid it, and easily fix the issue without making the frontend and backend unusable.

I was thinking one could be warned about deleting a content type definition that has content items in play, and perhaps rather than generating a NRE, one could get some type of overridable shape in its place that would be more descriptive of the problem and allow one to deal with it more gracefully. In the case of the `RawHtml` widget, it would be nice to be able to still view widgets in the design admin menu and see a shape where `RawHtml` would normally be that notifies one of the problem and still gives me the option to delete it if I wish. [Note: And possibly when an Audit Trail Module is in production, give me the option of re-creating the deleted definition.]

I've only run into the problem a few times while experimenting, but it's enough to get one thinking it could be handled much more gracefully.

### LINKED ISSUE #8969: Widget shape error handling
Throw a more explicit exception when a Content Type has been removed.

### LINKED ISSUE #10592: Unable to see the dropdown values of widget in decouple cms
### Discussed in https://github.com/OrchardCMS/OrchardCore/discussions/10590

<div type='discussions-op-text'>

<sup>Originally posted by **saleemhaq** October 29, 2021</sup>
Unable to see the dropdown values of widget in decouple cms when click on add widget button:

widgetarea
</div>

Skrypt:
Fixes #10592, Closes #10363, Fixes #1014, Fixes #8969, Fixes #6751, Fixes #7963, Fixes #1794

image

image

image

image

image


src/OrchardCore.Modules/OrchardCore.Flows/Views/FlowPart.Edit.cshtml
```diff
@@ -64,18 +67,25 @@
                     <i class="fa fa-plus" aria-hidden="true"></i>
                 </button>
                 <div class="dropdown-menu">
-                    @foreach (var type in widgetContentTypes.OrderBy(w => w.DisplayName))
+                    @if (widgetContentTypes.Count() == 0)
```
deanmarcussen, line 70:
```suggestion
                    @if (!widgetContentTypes.Any())
```

src/OrchardCore.Modules/OrchardCore.Flows/Drivers/FlowPartDisplayDriver.cs
```diff
@@ -47,11 +59,29 @@ public override IDisplayResult Display(FlowPart flowPart, BuildPartDisplayContex
...
             {
-                m.FlowPart = flowPart;
-                m.Updater = context.Updater;
-                m.ContainedContentTypeDefinitions = GetContainedContentTypes(context.TypePartDefinition);
+                var containedContentTypes = GetContainedContentTypes(context.TypePartDefinition);
+                var notify = false;
+
+                foreach (var widget in flowPart.Widgets)
+                {
+                    if (!containedContentTypes.Any(c => c.Name == widget.ContentType))
```
deanmarcussen, line 69: would it be better if here we constructed a new list of widgets that do have a content type, and pass that to the view model?

Savings us the need to make any changes and checks in the View

deanmarcussen (commented):
Couple of thoughts, but I like it.

src/OrchardCore.Modules/OrchardCore.Flows/Drivers/FlowPartDisplayDriver.cs
Skrypt: I agree.

src/OrchardCore.Modules/OrchardCore.Flows/Drivers/FlowPartDisplayDriver.cs
```diff
@@ -47,11 +59,29 @@ public override IDisplayResult Display(FlowPart flowPart, BuildPartDisplayContex
...
+                    {
+                        _logger.LogWarning("The Widget content item with id {0} has no matching {1} content type definition.", widget.ContentItem.ContentItemId, widget.ContentItem.ContentType);
```
sebastienros, line 71:
```suggestion
                        _logger.LogWarning("The Widget content item with id {ContentItemId} has no matching {ContentType} content type definition.", widget.ContentItem.ContentItemId, widget.ContentItem.ContentType);
```

sebastienros: And everywhere else

src/OrchardCore.Modules/OrchardCore.Flows/Drivers/FlowPartDisplayDriver.cs
deanmarcussen: Just this to do before merging I think?

Skrypt: We could merge and optimize the code later on too. Should end up with the same result.

Skrypt: Done

PR #10687, opened 2021-11-10
Cypress 9.0

agriffard:
https://www.npmjs.com/package/cypress/v/9.0.0

### DIFF
--- a/.github/workflows/functional_all_db.yml
+++ b/.github/workflows/functional_all_db.yml
@@ -10,7 +10,7 @@ jobs:
     name: Functional Tests - MVC
     runs-on: ubuntu-latest
     container:
-      image: cypress/included:8.7.0
+      image: cypress/included:9.0.0
     steps:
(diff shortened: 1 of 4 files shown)


Skrypt:
Does it fix the npm install issues with `har-validator`?

agriffard:
> Does it fix the npm install issues with `har-validator`?

I never encountered this `har-validator` issue by executing `npm i`.

Skrypt:
You need to do `cd /test/Functional` and `npm install`
And `cd /cypress-commands` and `npm install`

This is not part of our main package.json file so it does not execute npm install for these.

agriffard:
> You need to do `cd /test/Functional` and `npm install` And `cd /cypress-commands` and `npm install`
> 
> This is not part of our main package.json file so it does not execute npm install for these automatically.

No errors restoring packages in those folders.

Skrypt (approved):

Issue #5522, opened 2020-02-11
Trumbowyg is broken: Cannot read property 'x-small' of undefined


marc3lsk:
With latest preview nuget, when I choose trumbowyg as HTML editor, there is only empty textarea with following error in console:
Uncaught TypeError: Cannot read property 'x-small' of undefined
    at String.<anonymous> (trumbowyg-plugins.js?v=4PAOJjGgwwaNinsO0B7XBZrJ9d_lTVizlnxiHBl3eac:1531)
    at Function.each (jquery-3.4.1.js:367)
    at t (trumbowyg-plugins.js?v=4PAOJjGgwwaNinsO0B7XBZrJ9d_lTVizlnxiHBl3eac:1529)
    at Object.init (trumbowyg-plugins.js?v=4PAOJjGgwwaNinsO0B7XBZrJ9d_lTVizlnxiHBl3eac:1741)
    at Object.<anonymous> (trumbowyg.js:1910)
    at Function.each (jquery-3.4.1.js:373)
    at Trumbowyg.initPlugins (trumbowyg.js:1908)
    at Trumbowyg.init (trumbowyg.js:476)
    at new Trumbowyg (trumbowyg.js:460)
    at HTMLTextAreaElement.<anonymous> (trumbowyg.js:144)

file: OrchardCore.Modules\OrchardCore.Resources\Assets\trumbowyg\plugins\fontsize\trumbowyg.fontsize.js
line 222: text: '<span style="font-size: ' + size + ';">' + (trumbowyg.lang.fontsizes[size] || size) + '</span>',

agriffard:
Not yet solved. It only works when the current culture is different than `en` for the moment.

ChocoMachine:
We have this issue again in Version="1.0.0-rc2-13809"...

![image](https://user-images.githubusercontent.com/7971514/87390481-48a5ca80-c5ec-11ea-8303-fe3f827aa0b9.png)

ChocoMachine:
![image](https://user-images.githubusercontent.com/7971514/87391029-5445c100-c5ed-11ea-9b60-f48fbc9bd2c6.png)


and lang = "ltr"
![image](https://user-images.githubusercontent.com/7971514/87391145-8c4d0400-c5ed-11ea-99cb-d9331549cc5c.png)

deanmarcussen:
Yes, I just fixed the issue here for another pr.

Someone has added this code in a few places.
```
[...]

Issue #10654, opened 2021-11-07
HTML Editor is broken in localized content (RTL)


dodyg:
### Describe the bug

WYSIWYG Editor in Localized content (Arabic) is broken in version 1.0.

This is the error at the browser
```Uncaught TypeError: langs is undefined```

### To Reproduce
1. Create a new content definition with HtmlBody Part and set editor to WYSIWYG .
2. Add Localization Part.
3. Create content in English.
4. Translate content to Arabic. You will see the HTML Editor is broken (missing). 

### Expected behavior

For HTML Editor to work. Right now it just shows a multiline field.

hishamco:
Is this related to Arabic (which I don't think so), could you try another language and check if the issue still exist

dodyg:
Arabic, English, French

![arabic](https://user-images.githubusercontent.com/1761249/140645723-3c135e17-117c-4211-9ec9-25e7c55f8b48.png)
![english](https://user-images.githubusercontent.com/1761249/140645726-0ee64b27-de44-4421-ac72-9fd7e56fe0f6.png)
![french](https://user-images.githubusercontent.com/1761249/140645727-456a9cdf-1022-4c62-9325-b014d3e8bc16.png)

hishamco:
Please try another RTL language, I'm expecting the issue trelated to RTL script

dodyg:
Farsi
![farsi](https://user-images.githubusercontent.com/1761249/140646312-15e18513-8e7f-4e0e-b12e-b04433bee538.png)

dodyg:
Hebrew
![hebrew](https://user-images.githubusercontent.com/1761249/140646377-37fcdd57-42a6-423e-a84e-a5de8cce89c8.png)

hishamco:
As expected, let me check ..

Skrypt:
Duplicate of https://github.com/OrchardCMS/OrchardCore/issues/4709

Issue #4709, opened 2019-11-02
Html Wysiwyg is not RTL when the direction of current culture is RTL!!


ampadashi:
When the direction of the default culture in CMS is RTL, the direction of the Html Wysiwyg is not RTL. Only text-align property is set to right. but, the direction property in editor is still LTR.

hishamco:
@deanmarcussen I already fixed this long time ago, did you change anything related to RTL while working on gulp and resources

I remember you mentioned that before

deanmarcussen:
@hishamco I think maybe the original fix is not enough.

as @ampadashi just setting the `text-align` property is not enough for the wysiwig editor

```
    @if (culture.IsRightToLeft())
    {
        <div style="text-align:right">
            <textarea asp-for="Html" class="form-control"></textarea>
        </div>
    }
    else
    {
        <textarea asp-for="Html" class="form-control"></textarea>
    }
```

That would work ok for the standard editor, but for trumbowyg it will not override any of the generated `trumbowyg` rtl css.

Correct me if I'm wrong?

hishamco:
I was talking about `trumbowyg` style itself not the above markup, anyhow I will have a look again

ampadashi:
Right now, only text-align property is set to right. Direction property should be set to rtl too. Another thing which needs to be considered is that the editor has an option letting us see the source of html. Source view must be LTR and normal view must be rtl in all rtl cultures.

hishamco:
> Source view must be LTR and normal view must be rtl in all rtl cultures.

I don't think so while it has an HTML markup

@ampadashi could you please show us a screenshot for the RTL issue?

ampadashi:
Please take a look at the photo below. Current culture is fa-IR which is RTL. Direction of HTML editor is still LTR. Only text-align was set to right.
![image](https://user-images.githubusercontent.com/32009041/68083829-eaacca80-fe42-11e9-94b2-98a3a31ac2d2.png)

hishamco:
I see, I will try to handle this

ampadashi:
Please consider the source view mode. It must stay LTR.

hishamco:
@agriffard do you finished your PR that are related to `trumbowyg` coz I need to tweek it little bit for better RTL support

ns8482e:
Trumbowyg language files are not included in OrchardCore.Resources

https://alex-d.github.io/Trumbowyg/documentation/#add-localization

PR #10697, opened 2021-11-12
Fix RTL HTML editors

deanmarcussen:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/5522
https://github.com/OrchardCMS/OrchardCore/issues/4709
https://github.com/OrchardCMS/OrchardCore/issues/10654

Some of these are marked as closed already, but the issues were never fixed.


deanmarcussen:
There is room for a reasonable amount of weirdness here, because by default the trumbowyg lang scripts are not included in Orchard Core. 

So what was done originally never worked properly.

Here I am allowing forcing the lang dir to _rtl, but using the en dir (the only one supplied oob)

Otherwise everything was broken

Skrypt (approved):

Issue #10680, opened 2021-11-09
Multiple GetMedia calls when adding a folder


dannoh:
When you click OK on the new Media Folder name dialog the first thing it does after creating the new folder is emits 'beforeFolderAdded' which loads the folder we are currently on.  Then it emits the 'folderAdded' which selects the new folder (triggering a load of the new folder through the watcher)
It appears that both requests are in flight at the same time, and we are seeing that occasionally the beforeFolderAdded request comes in last and then makes it appear as if the new folder has all the contents of the previously selected folder.

```javascript
bus.$on('addFolder', function (target, folder) {
    if (self.model == target) {

        bus.$emit('beforeFolderAdded', self.model);
        if (self.children !== null) {
            self.children.push(folder);
        }                
        folder.parent = self.model;
        bus.$emit('folderAdded', folder);
    }
});
```

It wasn't immediately obvious to me why we load the folder before adding a new folder.  Can we just not do that? Or do we need some code to ensure we don't load the out of date data of the first request.

sebastienros:
Do you think you can provide a PR to fix the issue?

dannoh:
Created the PR
#10701

PR #10701, opened 2021-11-12
Removed beforeFolderAdded to stop multiple calls to fetch folders.

dannoh:
PR to address Issue #10680 

The 'beforeFolderAdded' loads the folder that we are on already.

When we create a folder we fire the 'beforeFolderAdded' then 'folderAdded' both of which load contents of a folder, but there is no guarantee on the order these calls will resolve in.  Occasionally we are seeing the first call to the original folder come in second and the new folder appears to be populated with images already.

Can't see why we need to fetch the contents of the first folder when adding a new one.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=10701) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=10701) before we can accept your contribution.<br/><br/>:x: dannoh [sign now](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=10701)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OrchardCMS/OrchardCore?pullRequest=10701) it.</sub>

sebastienros:
Do the assets need to be built with the gulp pipepline to generate the bundles?

dannoh:
Yes these get built into media.js.  Is there something I need to do as part of the PR?

sebastienros:
Run the gulp pipeline I assume that is at the root. @Skrypt @agriffard I don't think we have that documented?

Skrypt:
https://docs.orchardcore.net/projects/O1/en/latest/Documentation/Processing-client-side-assets/

Documentation is there and needs to be adapted for Orchard Core.

dannoh:
I'm familiar with the gulp pipeline and did run it (`gulp build`), but it doesn't generate anything that Git is interested in.  Is there an extra / different pipeline you are looking to have run for this to generate bundles?

sebastienros:
It should have updated `media.js` based on this file: https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore.Modules/OrchardCore.Media/Assets.json#L11-L16

Skrypt:
As declared here in the Assets.json file; everything in the "Assets/js/**/*.js" folder should be compiled as media.js.

```json
  {
    "inputs": [
      "./node_modules/blueimp-file-upload/js/jquery.fileupload.js",
      "./node_modules/blueimp-file-upload/js/jquery.iframe-transport.js",
      "Assets/js/**/*.js"
    ],
    "output": "wwwroot/Scripts/media.js"
  }
```

So, if it's not recompiling media.js then maybe you could try to do npm install directly on the module folder and then a gulp build afterward. But, our main package.json file is definitely doing a npm install on that folder.

`"install:orchardcore.media": "cd src/OrchardCore.Modules/OrchardCore.Media && npm install",`

sebastienros:
Perfect world: There is a github workflow that runs on PRs to run the gulp pipeline and detect discrepancies that would indicate the tool was not run. It would also create a PR to this PR! It's all totally possible and would help ensure we have the correct assets, and help contributors with this pipeline.

Then cherry on the cake if the validation step can check that the resources have the correct SHA.

sebastienros:
What did you finally do to get it working? Do you want to create another PR to improve the documentation with the solution?

dannoh:
To be honest it was a bit of a mess, I'm guessing due to a tooling version mismatch?
1. Every package-lock.json was modified after my npm install
2. I had to do a gulp rebuild to trigger the change. This created 200+ changes.  You can see in the media.js that I generated one of the methods was moved, I'm guessing the other 200 changes were similar.

jtkech:
It also happened to me ;) As I remember I needed tu update Node.js

Issue #9634, opened 2021-06-08
What exactly is the syntax supported  at the correlation Task?


hyzx86:
Tested not recognized `{{ Workflow.Properties.Doc.Id }}` throws an exception: Workflow is not defined

hyzx86:
When I directly enter the number：
`Unable to cast object of type 'System.Double' to type 'System.String'.`


Strangely enough, this way it works properly
![image](https://user-images.githubusercontent.com/15613121/121306846-ebb4b080-c931-11eb-8268-7fc8e791d05b.png)

chinasqzl:
具我所知，Luquid中，要对对象使用.访问，对象要先注册。不清楚你是否这种情况。
![image](https://user-images.githubusercontent.com/16968103/122846456-6de0a400-d338-11eb-980e-d13e61b45198.png)

hyzx86:
我这是工作流里的json啊



---Original---
From: ***@***.***&gt;
Date: Tue, Jun 22, 2021 09:02 AM
To: ***@***.***&gt;;
Cc: "Tony ***@***.******@***.***&gt;;
Subject: Re: [OrchardCMS/OrchardCore] What exactly is the syntax supported  at the correlation Task? (#9634)




 
具我所知，Luquid中，要对对象使用.访问，对象要先注册。不清楚你是否这种情况。

PR #10712, opened 2021-11-16
Fix CorrelateTask

hyzx86:
Fix :https://github.com/OrchardCMS/OrchardCore/issues/9634

The prompt states that it is using Liquid syntax, but the code incorrectly uses the Javascript parser

image

image


deanmarcussen:
This is quite breaking, because every user will be using the script evaluator.

Would it make more sense to just update the view so it says it uses a script, and remove the note about liquid?

hyzx86:
> This is quite breaking, because every user will be using the script evaluator.
> Would it make more sense to just update the view so it says it uses a script, and remove the note about liquid?

You're right, I've fixed the prompt message and changed the editor to codemirror

image

deanmarcussen (approved):

sebastienros:
Better to add an option here not to break. That's what Elsa does too.

sfmskywalker (approved):

sfmskywalker:
Could be done similar to this:
syntax-switching

Skrypt:
@sfmskywalker What happens when you have values in both these editors? Should we also save then the editor choice on the task?

hyzx86:
> @sfmskywalker What happens when you have values in both these editors? Should we also save then the editor choice on the task?

That will imporve more code changes

https://github.com/elsa-workflows/elsa-core/blob/cc7acb1790dcb2690b786d45ce56b47bf00928f4/src/core/Elsa.Abstractions/Attributes/ActivityInputAttribute.cs#L51

sfmskywalker:
> @sfmskywalker What happens when you have values in both these editors? Should we also save then the editor choice on the task?

@Skrypt Exactly, you would save the selected syntax.

sebastienros (approved):

Skrypt:
@sebastienros Are you sure? Because seems to me like it is missing something here that would define if the editor should parse javascript or Liquid based on the editor used.

Skrypt:
Ok, merge but open another issue to support Liquid and Javascript then.

PR #10739, opened 2021-11-23
Fix Liquid QueryCollection access

### LINKED ISSUE #10737: Liquid : Request.Query.value not working
### Describe the bug

{{ Request.Query.something }} should return the QueryString element value and it doesn't as of now from the latest main branch.

### Screenshot

https://user-images.githubusercontent.com/3228637/142957299-c48141ca-078f-4f22-93a4-6e4365e6e435.mp4

jtkech:
Fixes #10737 

The underlying type being a QueryCollectionInternal on which we don't have access

Fixed by creating a QueryCollection


Skrypt:
Should it not be the same for the Request.Form then?

jtkech:
Only checked for header collection that was not an internal class

Forgot to check all, just need to put a breakpoint and see what types the Request holds

Skrypt:
I'm testing it right now.
It returns nothing.

image

jtkech:
I tested it under the Blog recipe in the Article page, it was working

Yes it still says `QueryCollectionInternal`, did you applied my change?

Skrypt:
I mean I'm testing the Request.Form now and it's returning nothing. I assume that your fix does work for the Request.Query.

jtkech:
Ah okay, yes maybe the same issue with Request.Form

I'm on other things, I will check it asap

Skrypt:
Alright never mind, it works. I forgot while testing to add the antiforgery token when using a form post.

image

Skrypt (approved):

jtkech:
> Alright never mind, it works. I forgot while testing to add the antiforgery token when using a form post.

Cool cool ;)

PR #10743, opened 2021-11-23
Add current port to the setup tenant button url

### LINKED ISSUE #10742: Tenant "setup" and "view" buttons not working properly.
When trying to setup a new tenant by clicking on the setup button it returns a server not found error in the browser.
This is because we don't append the current server port to the tenant setup page url.

https://user-images.githubusercontent.com/3228637/142970670-201075cf-0b00-4e91-97a2-fa98f5df3288.mp4

Skrypt:
if using port 80 it will be cloaked naturally in the browser url.

Fixes #10742

## TODO

- [x] Fix functional tests

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Index.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Index.cshtml
@@ -18,8 +18,9 @@
         var shellSettings = shellSettingsEntry.ShellSettings;
         var host = shellSettings.RequestUrlHost?.Split(new[] { ',', ' ' }, StringSplitOptions.RemoveEmptyEntries).FirstOrDefault();
         
+        // if first host set is not specifying a port.
         // When developping we append the current port assuming it should never use port 80 or 443
-        if (Context.Request.Host.Port != 80 && Context.Request.Host.Port != 443)
+        if (!host.Contains(":") && Context.Request.Host.Port != 80 && Context.Request.Host.Port != 443)
         {
             host = String.Concat(host, ":", Context.Request.Host.Port.ToString());
         }
@@ -46,8 +47,9 @@
         var shellSettings = shellSettingsEntry.ShellSettings;
         var host = shellSettings.RequestUrlHost?.Split(new[] { ',', ' ' }, StringSplitOptions.RemoveEmptyEntries).FirstOrDefault();
 
+        // if first host set is not specifying a port.
         // When developping we append the current port assuming it should never use port 80 or 443
-        if (Context.Request.Host.Port != 80 && Context.Request.Host.Port != 443)
+        if (!host.Contains(":") && Context.Request.Host.Port != 80 && Context.Request.Host.Port != 443)
         {
             host = String.Concat(host, ":", Context.Request.Host.Port.ToString());
         }
--- a/src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Index.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Index.cshtml
@@ -3,6 +3,7 @@
 @using Microsoft.AspNetCore.Http
 @using OrchardCore.Environment.Shell
 @using OrchardCore.Environment.Shell.Models
+@inject Microsoft.AspNetCore.Hosting.IHostEnvironment env
 
 @{
     var originalPathBase = Context.Features.Get<ShellContextFeature>().OriginalPathBase;
@@ -20,7 +21,7 @@
         
         // if first host set is not specifying a port.
         // When developping we append the current port assuming it should never use port 80 or 443
-        if (!host.Contains(":") && Context.Request.Host.Port != 80 && Context.Request.Host.Port != 443)
+        if (env.IsDevelopment() && !host.Contains(":") && Context.Request.Host.Port != 80 && Context.Request.Host.Port != 443)
         {
             host = String.Concat(host, ":", Context.Request.Host.Port.ToString());
         }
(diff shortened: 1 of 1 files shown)


jtkech:
Normally the `HostString` includes the port if one

Then it take it into account when doing .ToString() => .ToUriComponents() see https://github.com/dotnet/aspnetcore/blob/465f9294d0b9f5403ab1a25c42ca9cf75e3cdeae/src/Http/Http.Abstractions/src/HostString.cs#L148-L150

When setting your tenant host you should include the port if one.

Skrypt:
Yes but as you can see we don't use the HostString in all cases. it ends up having no ports. Maybe I should add the port manually back to the HostString when it's null.

jtkech:
Normally the Host comes from the settings or default to the current request one

https://github.com/OrchardCMS/OrchardCore/blob/0e089fba606c56784e6b4b24ce37b455919faa5b/src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Index.cshtml#L17-L19

In both case it should include the port (if needed), this is the case when using the one of the current request, if coming from the settings and is missing we need to edit the tenant and update its host field

Skrypt:
Yes but this HostString should not include any specific port. So we need to append the currently used port so that it works.

jtkech:
Okay I understand but just to say that normally the port( that may be empty) is part of the hoststring, Port is a property of the Host, when it is implicitly converted to string (ToString()) it apends its port (if not empty).

What happens if your tenant host should not have a port and we still append e.g. the port of the current request, or if your tenant host need a specific port (defined with the host settings) we will append another one. Here one compromise would be to check if the host settings include a port, but still not good for the first case.

Skrypt:
Yes, I saw that HostString.Port only has a Getter. So you need to set it from its constructor. And it definitely returns the port when doing a ToString() as you did so that's why I changed the code. I don't think it is necessary to make the port configurable at all, to be honest.

Skrypt:
Actually, I think you are right. I should not assume it is always port 80. https runs on port 443

http://www.google.com:80
https://www.google.com:443

doesn't work : 
https://www.google.com:80
http://www.google.com:443

jtkech:
> So you need to set it from its constructor

Yes, just by passing a string that includes (or not) the port e.g. new HostString("test.com:2443")

What we use from the current request is the scheme, maybe we would need a setting ;)

Skrypt:
Maybe that works but there is a constructor that takes the port too. Maybe passing the port in the host is just a "hack".

Skrypt:
Ok then, I see your point maybe I could just not send the port by using this method.

Skrypt:
There. I think it's fairly non-breaking and now it takes into consideration if the first domain set has already a port. This is still tricky as this is not really documented.

Skrypt:
It is maybe still a little weak as we should only do this in development mode.

src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Index.cshtml
```diff
@@ -16,6 +19,14 @@
     {
         var shellSettings = shellSettingsEntry.ShellSettings;
         var host = shellSettings.RequestUrlHost?.Split(new[] { ',', ' ' }, StringSplitOptions.RemoveEmptyEntries).FirstOrDefault();
+        
+        // if first host set is not specifying a port.
+        // When developping we append the current port assuming it should never use port 80 or 443
+        if (env.EnvironmentName.Equals("development", StringComparison.OrdinalIgnoreCase) && !host.Contains(":") && Context.Request.Host.Port != 80 && Context.Request.Host.Port != 443)
```
hishamco, line 24: We can use function, and introduce static property or constants for ports. Simply we can do

```
 public int[] ReservedPorts = new [] { 80, 443 };
```

Skrypt: These are methods in a `Razor View`; they can't access static properties.

Skrypt: I'm starting to think that the only real solution is to use multiple hostnames with ports in the tenant configurations and to document it.

Skrypt:
How about if we would introduce something like : domain.local:*
Which would mean that if we find this ":*" it would bind to any current or future used port?

Skrypt:
After thinking. I removed the port from the first domain set in the hostnames. Then I concatenated the current context port to the domain. So, by deduction, the first URL set in the hostnames will be the one generally used in development. If that domain is used in production then it will use the current running port which could be any port. Doesn't break any websites so far. Then, if you want to add domains that include ports you can do it with all the other domains except the first one. Deal?

Skrypt:
I'm thinking that we should probably add a tenant setting that says: setup hostname. But, to not make any big changes here I basically use what we were and simply make sure that it will take the current running port.

src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Index.cshtml
```diff
@@ -16,6 +17,13 @@
     {
         var shellSettings = shellSettingsEntry.ShellSettings;
         var host = shellSettings.RequestUrlHost?.Split(new[] { ',', ' ' }, StringSplitOptions.RemoveEmptyEntries).FirstOrDefault();
+        
+        if (host != null)
```
Skrypt, line 21: This should basically never happen but there is another issue making this to happen.

Skrypt:
I've merged because it fixes the issue. If my fix creates a regression I'll take care of it but I would be really surprised.

src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Index.cshtml
```diff
@@ -45,7 +60,7 @@
             pathString = pathString.Add('/' + shellSettings.RequestUrlPrefix);
         }
 
-        return $"{Context.Request.Scheme}://{hostString.Value + pathString.Value}";
+        return $"{Context.Request.Scheme}://{hostString + pathString.Value}";
```
Skrypt, line 63: Here it is wrong to assume that the hostnames are all using {Context.Request.Scheme}

PR #4088, opened 2019-08-23
Localization files publishing

jtkech:
Prevents localization files from being outputed under the bin folder on building.

Just output them on publishing.

And other tweaks.

PR #10750, opened 2021-11-24
Add Localization to Docker images

### LINKED ISSUE #10678: Docker Localization .po files issue
I'm testing with the latest OrchardCore Linux Docker image and the .po files don't seem to be loaded or present at all.
This works though locally when building Orchard Core source code.

image

Skrypt:
Fixes #10678

### DIFF
--- a/.github/workflows/preview_ci.yml
+++ b/.github/workflows/preview_ci.yml
@@ -76,7 +76,7 @@ jobs:
       run: |
         rm -rf ./src/OrchardCore.Cms.Web/App_Data
         rm -rf ./src/OrchardCore.Cms.Web/App_Data_Tests
-        dotnet publish -c Release -o ./.build/release --no-build --no-restore --framework net6.
+        dotnet publish -c Release -o ./.build/release --no-build --no-restore --framework net6.0
         cp -R ./src/OrchardCore.Cms.Web/Localization ./.build/release/Localization
         docker build -f Dockerfile-CI -t orchardproject/orchardcore-cms-linux:dev .
         docker push orchardproject/orchardcore-cms-linux:dev
--- a/.github/workflows/preview_ci.yml
+++ b/.github/workflows/preview_ci.yml
@@ -77,7 +77,6 @@ jobs:
         rm -rf ./src/OrchardCore.Cms.Web/App_Data
         rm -rf ./src/OrchardCore.Cms.Web/App_Data_Tests
         dotnet publish -c Release -o ./.build/release --no-build --no-restore --framework net6.0
-        cp -R ./src/OrchardCore.Cms.Web/Localization ./.build/release/Localization
         docker build -f Dockerfile-CI -t orchardproject/orchardcore-cms-linux:dev .
         docker push orchardproject/orchardcore-cms-linux:dev
     - name: Deploy preview docker image for windows
--- a/.github/workflows/preview_ci.yml
+++ b/.github/workflows/preview_ci.yml
@@ -76,7 +76,8 @@ jobs:
       run: |
         rm -rf ./src/OrchardCore.Cms.Web/App_Data
         rm -rf ./src/OrchardCore.Cms.Web/App_Data_Tests
-        dotnet publish -c Release -o ./.build/release --no-build --no-restore --framework net6.0
+        dotnet publish -c Release -o ./.build/release --no-build --no-restore --framework net6.
+        cp -R ./src/OrchardCore.Cms.Web/Localization ./.build/release/Localization
         docker build -f Dockerfile-CI -t orchardproject/orchardcore-cms-linux:dev .
         docker push orchardproject/orchardcore-cms-linux:dev
     - name: Deploy preview docker image for windows
@@ -85,6 +86,7 @@ jobs:
         Get-ChildItem ./src/OrchardCore.Cms.Web/App_Data -Recurse | Remove-Item
         Get-ChildItem ./src/OrchardCore.Cms.Web/App_Data_Tests -Recurse | Remove-Item
         dotnet publish -c Release -o ./.build/release --no-build --no-restore --framework net6.0
+        Copy-Item -Path "./src/OrchardCore.Cms.Web/Localization" -Destination "./.build/release/Localization" -Recurse
         docker build -f Dockerfile-CI -t orchardproject/orchardcore-cms-windows:dev .
         echo "${{ secrets.DOCKERHUB_PASSWORD }}" | docker login -u="${{ secrets.DOCKERHUB_USERNAME }}" --password-stdin
         docker push orchardproject/orchardcore-cms-windows:dev
--- a/.github/workflows/preview_ci.yml
+++ b/.github/workflows/preview_ci.yml
@@ -86,7 +86,6 @@ jobs:
         Get-ChildItem ./src/OrchardCore.Cms.Web/App_Data -Recurse | Remove-Item
         Get-ChildItem ./src/OrchardCore.Cms.Web/App_Data_Tests -Recurse | Remove-Item
         dotnet publish -c Release -o ./.build/release --no-build --no-restore --framework net6.0
-        Copy-Item -Path "./src/OrchardCore.Cms.Web/Localization" -Destination "./.build/release/Localization" -Recurse
         docker build -f Dockerfile-CI -t orchardproject/orchardcore-cms-windows:dev .
         echo "${{ secrets.DOCKERHUB_PASSWORD }}" | docker login -u="${{ secrets.DOCKERHUB_USERNAME }}" --password-stdin
         docker push orchardproject/orchardcore-cms-windows:dev
--- a/.github/workflows/preview_ci.yml
+++ b/.github/workflows/preview_ci.yml
@@ -77,7 +77,6 @@ jobs:
         rm -rf ./src/OrchardCore.Cms.Web/App_Data
         rm -rf ./src/OrchardCore.Cms.Web/App_Data_Tests
         dotnet publish -c Release -o ./.build/release --no-build --no-restore --framework net6.0
-        cp -R ./src/OrchardCore.Cms.Web/Localization ./.build/release/Localization
         docker build -f Dockerfile-CI -t orchardproject/orchardcore-cms-linux:dev .
         docker push orchardproject/orchardcore-cms-linux:dev
     - name: Deploy preview docker image for windows
@@ -86,7 +85,6 @@ jobs:
         Get-ChildItem ./src/OrchardCore.Cms.Web/App_Data -Recurse | Remove-Item
         Get-ChildItem ./src/OrchardCore.Cms.Web/App_Data_Tests -Recurse | Remove-Item
         dotnet publish -c Release -o ./.build/release --no-build --no-restore --framework net6.0
-        Copy-Item -Path "./src/OrchardCore.Cms.Web/Localization" -Destination "./.build/release/Localization" -Recurse
         docker build -f Dockerfile-CI -t orchardproject/orchardcore-cms-windows:dev .
         echo "${{ secrets.DOCKERHUB_PASSWORD }}" | docker login -u="${{ secrets.DOCKERHUB_USERNAME }}" --password-stdin
         docker push orchardproject/orchardcore-cms-windows:dev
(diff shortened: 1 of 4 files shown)


.github/workflows/release_ci.yml
```diff
@@ -97,6 +98,7 @@ jobs:
         Get-ChildItem ./src/OrchardCore.Cms.Web/App_Data -Recurse | Remove-Item
         Get-ChildItem ./src/OrchardCore.Cms.Web/App_Data_Tests -Recurse | Remove-Item
         dotnet publish -c Release -o ./.build/release --no-build --no-restore --framework net6.0
+        Copy-Item -Path "./src/OrchardCore.Cms.Web/Localization" -Destination "./.build/release/Localization" -Recurse
```
sebastienros, line 101: Sound like an issue with our dotnet publish that it's not copied automatically.

Skrypt: It's more because of the previous `dotnet build` that doesn't copy them automatically because here we just take the files from the previous build made in the "build step". Here, we save time by doing `dotnet publish --no-build` which will take the files previously built in the /src/* folder.

Skrypt: So I just tried locally a `dotnet build --configuration Release --framework net6.0` directly on the root folder of my solution just after doing a `git clean -xdf` and the Localization folder gets updated with .po files at the end of the build process. So, I'm questioning myself now if this folder is there after doing the build on the CI. We could do a `cd ./src/OrchardCore.Cms.Web/Localization/fr` for example.

Skrypt: But I think that this folder is not taken into consideration because on the dotnet publish we say --no-restore and --no-build
So it takes only the files that would be generated without building and restoring which is what is required to get the Localization folder. It kind of does its job correctly.

Skrypt: I need to take a look at #4088

Skrypt: image

You can see here that the `wwwroot` folder is there but it may be created at the same time than creating the ìs-cache` folder.

Skrypt: Here locally on my PC when I execute

`dotnet publish -c Release -o ./.build/release --no-build --no-restore --framework net6.0`

It doesn't create any `Localization` folder but does create a `wwwroot` folder.

image

Skrypt: Another attempt here : 

image

Removing Localization in the first ItemGroup items. Redo a dotnet publish and still nothing.

jtkech: Can't now but I will take a look this night

We have the following msbuild target, maybe the problem is that it only executes after a buid


    <Target
      Name="CopyPackageTranslationFiles"
      Condition="'$(IsCrossTargetingBuild)' != 'true' and ('$(IsCrossTargetedBuild)' != 'true' or '$(TargetFrameworks' == '' or '$(TargetFrameworks.IndexOf($(TargetFramework)))' == '0')"
      AfterTargets="Build"> <== Maybe the problem

      <Message Text="Copying translation files: $(MSBuildProjectName)" Importance="high" />
      <Copy
        SourceFiles="%(PackageTranslationFiles.FullPath)"
        DestinationFolder="Localization\%(RecursiveDir)"
        Condition="'@(PackageTranslationFiles)' != ''"
        SkipUnchangedFiles="true" />

      <ItemGroup>
        <LocalizationFiles Include="Localization\**" Exclude="$(DefaultItemExcludes)" />
        <ResolvedFileToPublish Include="@(LocalizationFiles)">
          <RelativePath>Localization\%(RecursiveDir)%(Filename)%(Extension)</RelativePath>
        </ResolvedFileToPublish>
     </ItemGroup>
    </Target>

Skrypt: Ok, so the way it works is that here we remove any App_Data and Localization folder in every project that imports it.
This is kind of a reset function for all projects.

https://github.com/OrchardCMS/OrchardCore/blob/aa724a3a56c71f4d32b1973e4a2e73255b90d5cf/src/OrchardCore/OrchardCore.Application.Cms.Core.Targets/OrchardCore.Application.Cms.Core.Targets.targets#L3-L8

But then in the OrchardCore.Cms.Web.csproj we override these settings here : 

https://github.com/OrchardCMS/OrchardCore/blob/aa724a3a56c71f4d32b1973e4a2e73255b90d5cf/src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj#L6-L19

And you can see that we import the targets file before the override in the specific project.

We do this also in the OrchardCore.Application.Pages test project.

I'm inspecting the resulting OrchardCore.Cms.Web.exe and .dll with dotPeek and there are no resource files added to the assembly. Which means it's ok.

Skrypt: So the wwwroot folder and Localizations folder are really only added when doing a dotnet publish only. At least this part still works.

Skrypt:
```
 <ItemGroup> 
   <Compile Remove="App_Data\**;Localization\**" /> This removes from compilation
   <Content Remove="App_Data\**;Localization\**" /> This removes from publish
   <EmbeddedResource Remove="App_Data\**;Localization\**" /> This removes from assembly as embedded resource
   <None Remove="App_Data\**;Localization\**" /> This removes from assembly as embedded resource
 </ItemGroup> 
```

Skrypt: I don't know, this is SO weird! I give up for now. 🙈

Skrypt: @jtkech just saw your message. Yes, but here when we do a dotnet publish --no-build it won't recompile anything. Just use what has already been compiled in the /src folder. So the msbuild target here is probably not used at all when doing the dotnet publish.

jtkech: Yes this is what I meant

Skrypt: Here is another test.

```
git clean -xdf
dotnet build --configuration Release --framework net5.0
dotnet publish -c Release -o ./.build/release --no-build --no-restore --framework net5.0
```

image

image

Same result. The Localization folder is not there. I will make an other test by moving to an older commit on the main branch before we merged the .NET 6.0 PR. and report here what I find.

Skrypt: image

Same process and here the result : 

image

No Localization folder. Now I'm moving on testing older Docker images.

Skrypt: RC2 image works 👍🏼 

image

1.1.0 doesn't work 👎🏼 

image

Skrypt: So the only thing I'm thinking right now is that `dotnet publish --no-build` used to trigger the .targets file that was downloading the Localization folder. And now with .NET 6.0 it doesn't anymore. So, that's an MSBuild issue, or actually, it has been fixed because --no-build makes sense to not trigger any building task.

Skrypt: My fix doesn't hurt to be added to make a second pass and to be sure this Localization folder is there.

sebastienros: Ok, good job. But then we should have a hook to copy the localization files on publish too, when --no-build is used.

Skrypt: I'm testing more and I found out that RC2 was using AppVeyor.

jtkech:
@Skrypt 

Normally we would just need to mark them as Content to be `CopyToPublishDirectory`, but we don't want then to be shown under the solution folder, so we remove them from e.g. the `Content` collection.

That's why in a target after a build we do

    <ItemGroup>
      <LocalizationFiles Include="Localization\**" Exclude="$(DefaultItemExcludes)" />
      <ResolvedFileToPublish Include="@(LocalizationFiles)">
        <RelativePath>Localization\%(RecursiveDir)%(Filename)%(Extension)</RelativePath>
      </ResolvedFileToPublish>
    </ItemGroup>

But that doesn't get called on a publish only (maybe before it was)

Just tried the following by just adding in `OrchardCore.Application.Cms.Core.Targets.targets` this


    <Target Name="PublisTranslationFiles" BeforeTargets="GetCopyToPublishDirectoryItems">
      <Message Text="Publish translation files: $(MSBuildProjectName)" Importance="high" />

      <ItemGroup>
        <_WebProjectGeneratedContent Include="Localization\**" Exclude="$(DefaultItemExcludes)">
        </_WebProjectGeneratedContent>
      </ItemGroup>
    </Target>

It works with net6.0, will try with 5.0 and 3.1

Skrypt:
If we add the settings in the .targets file this will get applied on 2 different projects.
One is OrchardCore.Cms.Web and the other one OrchardCore.Tests.Pages.
There is an issue with the OrchardCore.Tests.Page where it doesn't add the Localization folder too.

jtkech:
Okay also works with 5.0 and 3.1, as anyway it uses the same msbuild scripts of the currently installed dotnet sdk.

So by adding the following in `OrchardCore.Application.Cms.Core.Targets.targets` 


    <Target Name="PublisTranslationFiles" BeforeTargets="GetCopyToPublishDirectoryItems">
      <Message Text="Publish translation files: $(MSBuildProjectName)" Importance="high" />

      <ItemGroup>
        <_WebProjectGeneratedContent Include="Localization\**" Exclude="$(DefaultItemExcludes)">
        </_WebProjectGeneratedContent>
      </ItemGroup>
    </Target>

Skrypt:
Ok I prefer your solution except for the typo PublisTranslationFiles = PublishTranslationFiles

```
<Target Name="PublisTranslationFiles" BeforeTargets="GetCopyToPublishDirectoryItems">
  <Message Text="Publish translation files: $(MSBuildProjectName)" Importance="high" />

  <ItemGroup>
    <_WebProjectGeneratedContent Include="Localization\**" Exclude="$(DefaultItemExcludes)">
    </_WebProjectGeneratedContent>
  </ItemGroup>
</Target>
```

image

Skrypt:
Ok so, while it is not necessary to have the Localizations in the OrchardCore.Tests.Pages. I would expect that it works there too. But since it is not really an issue for now let's do this : 🙈

Skrypt:
image

here is the proof that it would work if there were actually files in that folder.

jtkech:
> Ok so, while it is not necessary to have the Localizations in the OrchardCore.Tests.Pages

This is because it references `Cms.Core.Targets`, not `Cms.Targets` that references the localization packages.

> Ok I prefer your solution except for the typo PublisTranslationFiles = PublishTranslationFiles

Yes sorry for the typo ;)

.github/workflows/preview_ci.yml
```diff
@@ -77,6 +77,7 @@ jobs:
         rm -rf ./src/OrchardCore.Cms.Web/App_Data
         rm -rf ./src/OrchardCore.Cms.Web/App_Data_Tests
         dotnet publish -c Release -o ./.build/release --no-build --no-restore --framework net6.0
+        cp -R ./src/OrchardCore.Cms.Web/Localization ./.build/release/Localization
```
jtkech, line 80: Do we still need this line then?

Normally with the new little target it populates the translation files under the publish output path

Idem below

Skrypt: Yes if we want to fix the current 1.1.0 Docker image we need to fix it from the CI. It can be removed afterward.

Skrypt: Now the other question is: is this necessary that we keep this part?

https://github.com/OrchardCMS/OrchardCore/blob/aa724a3a56c71f4d32b1973e4a2e73255b90d5cf/src/OrchardCore/OrchardCore.Application.Cms.Core.Targets/OrchardCore.Application.Cms.Core.Targets.targets#L36-L41

Skrypt: Removed it as it's kind of doing it afterward anyway.

jtkech: Yes it looks like, I assume that you tried it ;)

About not modifying the _ci.yml

> Yes if we want to fix the current 1.1.0 Docker image we need to fix it from the CI. It can be removed afterward.

Could we just do the same (only the suggested change in the .targets) in separate PRs targetting the version branches that need to be updated?

Skrypt: I think we use tags for releases

jtkech (approved):

PR #10777, opened 2021-11-29
Make the Monaco editor Fields  support preview

hyzx86:
Modify HtmlField and TextField to use Ctrl-S to trigger previews when using the Monaco editor

动画


sebastienros:
Why not on each key stroke like the other editors?

hyzx86:
I don't know much about Monaco, but I see that this code is only going to trigger the get value action when the submit is clicked, so I think this is a performance optimization.

```
 window.addEventListener("submit", function () {
                textArea.value = editor.getValue();
});
```
Of course I could also add custom switch options for the Monaco editor, for example, the default is to trigger the update by each key, and Ctrl+S only after the switch is turned on , I think this would be a better solution.

deanmarcussen:
Noone will ever know to enable it with Ctrl+S

Can you try with just keystrokes and if the performance is fine, I think that will be ok

MichaelPetrinolis:
If there are still any performance issues, adding a throttle on keystrokes might solve them.

hyzx86:
I don't know how to monitor the change evnet of  monaco-editor .. 😂
could some  one give me a sample ?

hyzx86:
updated , and change the preview fonction use a named window to avoid always open new window

preview2

src/OrchardCore.Modules/OrchardCore.ContentPreview/Views/ContentPreview.Button.cshtml
```diff
@@ -10,7 +10,7 @@
 
 @if (hasPreviewPermission)
 {
-    <a id="previewButton" target="_blank" class="btn btn-info" href="@Url.Action("Index", "Preview", new { area = "OrchardCore.ContentPreview", id = previewId })">@T["Preview"]</a>
+    <a id="previewButton" target="previewWindow" class="btn btn-info" href="@Url.Action("Index", "Preview", new { area = "OrchardCore.ContentPreview", id = previewId })">@T["Preview"]</a>
```
deanmarcussen, line 13: @Skrypt do you agree with this change? I think I do (but equally if I have multiple Orchard instances running (dev and live), it will confuse me when it keeps swapping windows around on me?

Other changes look good to me.

Skrypt: I'm not sure, to be honest. Could we not have a unique identifier for this target instead?

hyzx86: add the host name as perfix?

deanmarcussen: I would prefer removing it, and opending a seperate issue sowe can have a proper review of it.

src/OrchardCore.Modules/OrchardCore.ContentFields/Views/HtmlField-Monaco.Edit.cshtml
```diff
@@ -46,7 +46,7 @@
 
             var editor = monaco.editor.create(document.getElementById('@Html.IdFor(x => x.Html)_editor'), settings);
             var textArea = document.getElementById('@Html.IdFor(x => x.Html)');
-            
+
             editor.getModel().onDidChangeContent((event) => {
```
deanmarcussen, line 50: How come I see the exact same code that you're adding up on L50?

i.e. don't we already do this?
```
           //for preview
            if(document.getElementById("previewButton")){
                editor.getModel().onDidChangeContent((event) => {
                    textArea.value = editor.getValue();
                    $(document).trigger('contentpreview:render');
                });
            }
```

hyzx86: Yes , you are right .. 
I didn't notice the codes 。

src/OrchardCore.Modules/OrchardCore.ContentFields/Views/TextField-Monaco.Edit.cshtml
```diff
@@ -49,6 +49,13 @@
             window.addEventListener("submit", function () {
                 textArea.value = editor.getValue();
             });
+            //for preview
+             if(document.getElementById("previewButton")){
```
deanmarcussen, line 53: Don't think we need this check, we don't use it anywhere else.

Can you remove for consistency?

Thanks

hyzx86: done😊

deanmarcussen (approved):

Skrypt (approved):

src/OrchardCore.Modules/OrchardCore.ContentFields/Views/TextField-Monaco.Edit.cshtml
```diff
@@ -49,6 +49,11 @@
             window.addEventListener("submit", function () {
                 textArea.value = editor.getValue();
             });
+            editor.getModel().onDidChangeContent((event) => {
+                 textArea.value = editor.getValue();
+                 $(document).trigger('contentpreview:render');
+            }); 
+
```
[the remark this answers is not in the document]
Skrypt: I agree. Please format the code appropriately.

PR #10788, opened 2021-12-01
Preserve contained content items ids

### LINKED ISSUE #10723: Maintain Content Item Id across BagPart ContentItems Not Working?
### Describe the bug
On issue #5653 it appears the intent was to have content item ids on Bag Parts be persistent (i.e. not change). However, as of today, I don't see that it is working / could work. The Bag Part Content Item Id isn't persisted. Before I go and make a static Id using the IdGenerator that doesn't change per "content" item, I wanted to check if it is understood that this doesn't seem to work. The area of the code that seems to try and address this is [here](https://github.com/OrchardCMS/OrchardCore/blob/9ba5a4015b122ec2c2bbae3c864cc86a61f75d1a/src/OrchardCore.Modules/OrchardCore.Flows/Drivers/BagPartDisplay.cs#L72) but again, I don't think it is working anymore.  Reference @deanmarcussen as back in the day you had implemented it but maybe it is no longer desired.

### To Reproduce
Just made a parent content type with a child content type like so:
image

Then I added in a Parent with a couple of children and published it to get this result (it is outputting the child content item id):
image

But every time I Publish the Parent Content Item the child content item ids change:
image

### Expected behavior
Based on the issue #5653 I had assumed it would retain it's child content item ids.

Again, it is not to big of a deal to have a static field (like an alias) that autogenerates a unique id every time. Just didn't want to bypass this if the goal was to have child content item ids persist.

jtkech:
Fixes #10723 

- By using a dedicated `ContentItems` array to hold the Ids.

- Prevent conflicts when removing (or moved to another container) and then adding an item, by computing indexes differently when adding an item, was based on the number of items + 1, now based on the existing max value + 1.

- Fixes other little issues when dragging between containers more than one level up / down.

Note: Can't preserve an item ID when moved between containers, as it involves 2 flow part driver executions.


sebastienros (approved):

sebastienros:
When it's ready

Issue #10794, opened 2021-12-02
Distributed tenants still produce flat files


sonicmouse:
I need to make sure that I set up a SaaS that has tenants that can run on multiple instances. I believe I have set this up correctly, but the tenant is still producing files.

I am using .NET Core 3.1 and the NuGet packages to accomplish this. My `appsettings.json` is set up as so:

```
{
	"EnvironmentName": "Development",
	"OrchardCoreFeatures": {
		"Global": [
			"OrchardCore.Redis",
			"OrchardCore.Redis.DataProtection",
			"OrchardCore.Redis.Lock",
			"OrchardCore.Redis.Cache",
			"OrchardCore.DataProtection.Azure"
		]
	},
	"OrchardCore": {
		"DatabaseProvider": "SqlConnection",
		"ConnectionString": "...",
		"OrchardCore_Admin": {
			"AdminUrlPrefix": "SomethingAdmin"
		},
		"OrchardCore_Shells_Database": { // needs AddDatabaseShellsConfiguration() registered in startup
			"DatabaseProvider": "SqlConnection",
			"ConnectionString": "...",
			"TablePrefix": "something-cms-shellsettings",
			"MigrateFromFiles": true
		},
		"OrchardCore_Media_Azure": {
			"ConnectionString": "...",
			"ContainerName": "something-cms-{{ ShellSettings.Name }}-media",
			"BasePath": "Media",
			"CreateContainer": true
		},
		"OrchardCore_DataProtection_Azure": {
			"ConnectionString": "...",
			"ContainerName": "something-cms-dataprotection",
			"BlobName": null, // auto-generate based on tenant name
			"CreateContainer": true
		},
		"OrchardCore_Redis": {
			"Configuration": "...",
			"InstancePrefix": "something-cms-"
		}
	}
}
```
How my startup.cs looks:
```
public class Startup
{
	private readonly IConfiguration _config;

	public Startup(IConfiguration configuration)
	{
		_config = configuration;
	}

	private string[] GetStringsFromConfig(string section) =>
		_config.GetSection(section).Get<string[]>() ?? Array.Empty<string>();

	public void ConfigureServices(IServiceCollection services)
	{
		services.AddOrchardCms(builder => builder
			.AddGlobalFeatures(GetStringsFromConfig("OrchardCoreFeatures:Global"))
			.AddDatabaseShellsConfiguration() /* config in OrchardCore_Shells_Database appsetting */);
	}

[...]

PR #10799, opened 2021-12-02
Don't enable FileContentDefinition feature by default in our themes

Skrypt:
Because it is not consistent with other themes we have. And it should be the other way around. If you want to have your own Content Definitions locally then, enable this module for yourself. Then, commit to the database later on.

Related to #10794

### DIFF
--- a/src/OrchardCore.Themes/TheAgencyTheme/Recipes/agency.recipe.json
+++ b/src/OrchardCore.Themes/TheAgencyTheme/Recipes/agency.recipe.json
@@ -40,7 +40,6 @@
         "OrchardCore.ContentFields",
         "OrchardCore.ContentPreview",
         "OrchardCore.Contents",
-        "OrchardCore.Contents.FileContentDefinition",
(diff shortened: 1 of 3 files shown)


sebastienros:
But we did add them there by default so devs can shared their changes in source control instead of the db. Would it make you happy if it was a default feature, and storing it to the database was the actual optional one?

deanmarcussen:
This has become my least favorite feature due to the number of times I’ve had to explain how to turn it off :(

Skrypt:
FileContentDefinitions feature is nice for local development, but it is more prone to bad practice usage than the DB version. This is mainly why the default should be DB I think.

sebastienros:
I am fine with both options. And not a breaking change since it's for new sites. We should add a comment on the original issue too so that people who cared are aware of the change. And update the documentation to mention this feature.

Skrypt:
Documentation done.

jtkech (approved):

PR #10807, opened 2021-12-03
Use current port for Tenant urls in the Tenants admin page

### LINKED ISSUE #10742: Tenant "setup" and "view" buttons not working properly.

jtkech:
Fixes #10742


Skrypt:
Looks good. It doesn't cover all use cases but that's what should at least work.

Skrypt (approved):

jtkech:
Yes I agree but the best compromise I found

src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Index.cshtml
```diff
@@ -37,7 +46,16 @@
     {
         var shellSettings = shellSettingsEntry.ShellSettings;
         var host = shellSettings.RequestUrlHost?.Split(new[] { ',', ' ' }, StringSplitOptions.RemoveEmptyEntries).FirstOrDefault();       
-        var hostString = host != null ? new HostString(host) : Context.Request.Host;
+
+        var hostString = Context.Request.Host;
```
hishamco, line 50: I'm not sure why we use function here to remove the duplicate code?!!

sebastienros: Agreed, reuse code in private method

jtkech: > I'm not sure why we use function here to remove the duplicate code?!!

Yes, in this razor view we have 2 `@function`s with specific code, one is called twice, so yes to not have duplicate code, and one is called once but to still not have too much specific code mixed with html tags.

> Agreed, reuse code in private method

Not sure I undestand, we already reuse this code through this razor `@function` being a kind of private method.

sebastienros: Take this part of the code and create a 3rd function `BuildUrlWithHost`

jtkech:
@Skrypt

Thanks for the change, yes the duplicated part is to build an `HostString` (the rest looks similar but is not)

Hmm, but then I think that the name suggested by @sebastienros is not right, `BuildUrlWithHost` means build an url, but here we now build an `Hostring` with this new method, so maybe something like.

    private HostString BuildHostString(string host)

Skrypt:
Have fun 😉

jtkech:
Yes, you're right ;)

We could also add to this shared method the duplicated code to build the `PathString` and returns a string generated from both `Hostring` + `PathString`, because the consumer needs it, but one consumer need it in an escaped form. Or return a Tuple of both so that the consumer can escape them or not. Or use a `bool escape` parameter ;)

That's also why I didn't understand what to do, I will look at it tomorrow

Skrypt:
If you want to compose the perfect code for this simple view I think it's a waste of time. But, yeah, if you have time and want to have fun finding the perfect thing ... I won't argue 😄

Skrypt:
I would probably take more time to analyze all the use cases we currently have with the HostString. We can host more than just domains on port 5000 and 5001 on the same application. I was able the other day to assign a specific tenant to work on a specific port only. So, there is also the HTTP vs HTTPS issue too. So, I think that here, as we discussed at last Thursday's meeting, we just want new tenants' setup buttons to work at least.

jtkech:
So I thought about suggesting only one private method having a `bool displayOnly = false` parameter

But the funny thing is that `GetDisplayUrl()` was done to get a non escaped url, this to be used as the text of a link, the `href` attribute still using `GetEncodedUrl()` to get an escaped url.

But I retrieved a commit where the usage of the link text has been removed, so we now always need to escape, but currently there is an `href` that is still set with the non escaped `GetDisplayUrl()`.

In fact now we always need to escape, the only option is to append or not a query string with the tenant token, this if the tenant is already setup or not. So `GetDisplayUrl()` is used to set an `href` but just to not append the query string.

I will update it

jtkech:
Just saw your last comments, yes I agree

> Thursday's meeting, we just want new tenants' setup buttons to work at least.

Me too ;)

But a little refactoring was requested, which finally is not so bad because it let me see the above.

jtkech:
> If you want to compose the perfect code for this simple view I think it's a waste of time. But, yeah, if you have time and want to have fun finding the perfect thing ... I won't argue

I was thinking the exact same thing, what I explained about returning a Tuple with a HostString and a Pathstring was a joke, ending with a ";)" In fact just to say the same thing that it looks a little useless

jtkech:
So, because now we don't use any link text, only `href`, we don't need anymore to get an unescaped url for display only, we only need escaped urls, so i ended up with only one method that always encode the url.

The only option is to append a query or not, used to pass a tenant token if the tenant is not yet setup.

Note: If we need again an url for display only, we can use the `uri.GetComponents` helper (that we already use for redirecting) that can act on a whole url (all uri components) with the option to escape it or unescape it.

PR #10813, opened 2021-12-05
Fix indexes namespace

hishamco:
Seems there are some incosistency in indexes namespace, we usually used `OrchardCore.XXX.Indexes` for YesSQL indexes, while we use `OrchardCore.XXX.Indexing` for `OrchardCore.Indexing` related things such as implementation of `ContentFieldIndexHandler`, so this PR fix this issue to make clear for everyone


deanmarcussen (commented):
While I appreciated the concept of cleaning up code, this actually represents a breaking change for every consumer of the `ContentItemIndex`, and other indexes you changed.

Because every user who is using it (and there will be many) will also have to update their namespace usage.

Every third party nuget package would have to also republish, to prevent missing method exceptions.

I don't think there's enough value in it.

It would be much better if you opened issues, instead of prs, for things like this.

hishamco:
I agree that this may introduce a breaking changes, but we could defer this PR to get merged in the upcoming major release, we could at least use `Obsolete` like what we did in many places

> I don't think there's enough value in it.

I disagree, our APIs should be consistent, from a consumer perspective. Again I agree to avoid breaking changes

hishamco:
@Skrypt I think you use `Obsolete` to refactor lucene queries If I'm not wrong to avoid breaking-changes. Am I right?

Skrypt:
Ok, it is good to take initiatives likes these but we need to first decide that we may introduce breaking changes first. This is something that needs to be done but for 2.0 only. I don't think we are there yet so let's keep it open, triage this and decide when it will be merged if so.

sebastienros:
Nope ;)

PR #10835, opened 2021-12-07
Fix docker image missing localization files in 1.1.0 release

Skrypt:
As @jtkech suggested.

Will require that we change the release Tag on the repository to a different commit.


jtkech (approved):

Skrypt:
@sebastienros I think I will let you merge this one and decide what we do with it. Should we have a late 1.1.1 release or replace the release by adding this commit?

sebastienros (approved):

sebastienros:
re-targeted to main, or 1.2 only

Skrypt:
It is already merged in main. It's a second PR to target 1.1.0 specifically because the Docker image ...

PR #10852, opened 2021-12-10
Allow to provide CdnUrl or CdnDebugUrl

hishamco:
The `ResourceDefinition` fluent APIs allow us to set the Cdn Url & Cdn Debug Url if we have both the original and minified versions, also have an overload to accept Cdn Url in case the minified version is not present. In some cases we may have the minified version but the original resource is missing a real example is `jquery-ui-i18n.js`

https://code.jquery.com/ui/1.7.2/i18n/jquery-ui-i18n.js is missing
https://code.jquery.com/ui/1.7.2/i18n/jquery-ui-i18n.min.js is present

So, there's no way to support this currently, unless you provide the Cdn Debug for both source and debug versions which is wrong. This PR introduces a new parameter `debug` which is optional to avoid breaking-changes

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Resources/Liquid/ScriptTag.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Resources/Liquid/ScriptTag.cs
@@ -74,16 +74,30 @@ namespace OrchardCore.Resources.Liquid
                     var s = src.ToLowerInvariant();
 
                     var definition = resourceManager.InlineManifest.DefineScript(s);
-                    definition.SetUrl(src, debugSrc);
+
+                    if (String.IsNullOrEmpty(debugSrc))
+                    {
+                        definition.SetUrl(src);
+                    }
+                    else
+                    {
+                        definition.SetUrl(src, debugSrc);
+                    }
 
                     if (!String.IsNullOrEmpty(version))
                     {
                         definition.SetVersion(version);
                     }
 
-                    if (!String.IsNullOrEmpty(cdnSrc))
+                    if (!String.IsNullOrEmpty(cdnSrc) && !String.IsNullOrEmpty(debugCdnSrc))
                     {
                         definition.SetCdn(cdnSrc, debugCdnSrc);
+                    }
+                    else
+                    {
+                        var isDebug = !String.IsNullOrEmpty(debugCdnSrc);
+
+                        definition.SetCdn(isDebug ? debugCdnSrc : cdnSrc, debug: isDebug);
                     }
 
                     if (!String.IsNullOrEmpty(culture))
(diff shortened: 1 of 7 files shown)


hishamco:
@agriffard I'm not sure why this is failing in non-windows, should I close and re-open the PR?

Skrypt:
@hishamco You can also restart the Action workflow directly instead.

Skrypt:
Functional tests are failing. You will need to run them locally on your PC to figure out what is happening.

hishamco:
I didn't touch them, but I will have a look into them

Thanks

hishamco:
The strange part that the testing fails on non-windows?!!

jtkech:
> The strange part that the testing fails on non-windows?!!

Because the functional tests are only run under ubuntu

hishamco:
So I don't have any idea what's going on

jtkech:
@hishamco 

I run cypress tests locally and could find the issue, it fails when creating a tenant

cypress test1

    Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware|ERROR|An unhandled exception has occurred while executing the request. System.ArgumentNullException: Value cannot be null. (Parameter 'urlDebug')
    at OrchardCore.ResourceManagement.ResourceDefinition.ThrowArgumentNullException(String paramName) in C:\Sources\OrchardCoreTest\src\OrchardCore\OrchardCore.ResourceManagement.Abstractions\ResourceDefinition.cs:line 414
    at OrchardCore.ResourceManagement.ResourceDefinition.SetUrl(String url, String urlDebug) in C:\Sources\OrchardCoreTest\src\OrchardCore\OrchardCore.ResourceManagement.Abstractions\ResourceDefinition.cs:line 91
    at OrchardCore.ResourceManagement.TagHelpers.ScriptTagHelper.ProcessAsync(TagHelperContext context, TagHelperOutput output) in C:\Sources\OrchardCoreTest\src\OrchardCore\OrchardCore.ResourceManagement\TagHelpers\ScriptTagHelper.cs:line 71

It throws at ResourceDefinition.cs:line 414 on a null urlDebug, called from ScriptTagHelper line 71

        // Called from ScriptTagHelper line 71
        public ResourceDefinition SetUrl(string url, string urlDebug)
        {
            if (String.IsNullOrEmpty(url))
            {
                ThrowArgumentNullException(nameof(url));
            }

            if (String.IsNullOrEmpty(urlDebug))
            {
                ThrowArgumentNullException(nameof(urlDebug)); <= Throws line 414
            }

            return SetUrlInternal(url, urlDebug);
        }

hishamco:
Thanks a lot @jtketch I will do a unit test for that, then update the PR soon

hishamco:
https://github.com/OrchardCMS/OrchardCore/blob/339d74b251b53393dd510f8ad3027aa17d78500d/src/OrchardCore/OrchardCore.ResourceManagement/TagHelpers/ScriptTagHelper.cs#L71

I will check why this line provides one URL instead of two

deanmarcussen:
I don't think these are good changes to the api.

> So, there's no way to support this currently, unless you provide the Cdn Debug for both source and debug versions which is wrong.

Yet this is eactly what you need to do in this scenario.

You may have confused this by thinking that debug = minified, and not debug = not minified. They don't always match exactly that, as it is based on the resource Debug Mode setting, not an exact, it must be minified scenario.

You've also added, somewhat odd `.SetCdnIntegrity` apis seperately. Which doesn't add up to me.

hishamco:
> Yet this is eactly what you need to do in this scenario.

Nothing but, this introduce the confusion again, because source not equal debug version

> It's also obvious that the changes you have made are going to be breaking on customers, as even just an orchard core test run fails.

> You've also added, somewhat odd .SetCdnIntegrity apis seperately. Which doesn't add up to me.

Let me check again

No, this is something else because the `StyleTagHelper` & `ScriptTagHelper` should use the proper overload, we should pass two versions if they are present

hishamco:
@jtkech can these tests runs on windows, or is there something special for Ubuntu

/cc @jptissot

jtkech:
> @jtkech can these tests runs on windows, or is there something special for Ubuntu

Yes you can, first I suggest that you install the last version of node.js

Then under the functional project run `npm install` and wait until it completes

Then e.g. `npm run mvc:test` for a first test then e.g. `npm run cms:test`

Look at `package.json` for the list of possible scripts commands

E.g. you can also open the cypress UI with `npm run cms:cypress`

But in that case you first need to setup (see below) and then start the cms app manually from another console

    dotnet bin/debug/net6.0/orchardcore.cms.web.dll

The cms app needs to be setup with the admin user name and password that cypress scripts will use

Look at the `cypress.json` file under the `cms-tests` folder to see the user name and password that should be used

hishamco:
Cypress executable not found at: \AppData\Local\Cypress\Cache\9.1.1\Cypress\Cypress.exe

PR #10853, opened 2021-12-10
Allow sending email without "To" if "Cc" or "Bcc" is provided

hishamco:
Addresses #10840

### DIFF
--- a/src/OrchardCore/OrchardCore.Email.Core/Services/SmtpService.cs
+++ b/src/OrchardCore/OrchardCore.Email.Core/Services/SmtpService.cs
@@ -51,6 +51,11 @@ namespace OrchardCore.Email.Services
         /// <remarks>This method allows to send an email without setting <see cref="MailMessage.To"/> if <see cref="MailMessage.Cc"/> or <see cref="MailMessage.Bcc"/> is provided.</remarks>
         public async Task<SmtpResult> SendAsync(MailMessage message)
         {
+            if (_options == null)
+            {
+                return SmtpResult.Failed(S["SMTP settings must be configured before an email can be sent."]);
+            }
+
             try
             {
                 // Set the MailMessage.From, to avoid the confusion between _options.DefaultSender (Author) and submitter (Sender)
@@ -67,7 +72,7 @@ namespace OrchardCore.Email.Services
 
                 if (mimeMessage.From.Count == 0 && mimeMessage.Cc.Count == 0 && mimeMessage.Bcc.Count == 0)
                 {
-                    return SmtpResult.Failed(S["SMTP settings must be configured before an email can be sent."]);
+                    return SmtpResult.Failed(S["The mail message should have at least one of these headers: To, Cc or Bcc."]);
                 }
 
                 switch (_options.DeliveryMethod)
--- a/src/OrchardCore/OrchardCore.Email.Core/Services/SmtpService.cs
+++ b/src/OrchardCore/OrchardCore.Email.Core/Services/SmtpService.cs
@@ -13,6 +13,9 @@ using MimeKit;
 
 namespace OrchardCore.Email.Services
 {
+    /// <summary>
+    /// Represents a SMTP service that allows to send emails.
+    /// </summary>
     public class SmtpService : ISmtpService
     {
         private const string EmailExtension = ".eml";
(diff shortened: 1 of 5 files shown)


src/OrchardCore/OrchardCore.Email.Core/Services/SmtpService.cs
```diff
@@ -34,22 +43,33 @@ IStringLocalizer<SmtpService> stringLocalizer
             S = stringLocalizer;
         }
 
+        /// <summary>
+        /// Sends the specified message to an SMTP server for delivery.
+        /// </summary>
+        /// <param name="message">The message to be sent.</param>
+        /// <returns>A <see cref="SmtpResult"/> that holds information about the sent message, for instance if it has sent successfully or if it has failed.</returns>
+        /// <remarks>This method allows to send an email without setting <see cref="MailMessage.To"/> if <see cref="MailMessage.Cc"/> or <see cref="MailMessage.Bcc"/> is provided.</remarks>
```
hishamco, line 51: I didn't see a suitable place to mention this in our docs, so I just write here, so any developer can notice that the email can required on of the three headers

hishamco:
I may need to check work flow task too, to react to this change

Skrypt:
Yes, here.

https://github.com/OrchardCMS/OrchardCore/blob/e73a81ee82bb55fad2a4ac5d3d1fe6fd986f28b4/src/OrchardCore.Modules/OrchardCore.Email/Workflows/ViewModels/EmailTaskViewModel.cs#L17-L18

src/OrchardCore/OrchardCore.Email.Core/Services/SmtpService.cs
```diff
@@ -34,22 +43,38 @@ IStringLocalizer<SmtpService> stringLocalizer
...
+                if (!String.IsNullOrWhiteSpace(senderAddress))
+                {
+                    message.From = senderAddress;
+                }
+
                 var mimeMessage = FromMailMessage(message);
 
+                if (mimeMessage.From.Count == 0 && mimeMessage.Cc.Count == 0 && mimeMessage.Bcc.Count == 0)
+                {
+                    return SmtpResult.Failed(S["The mail message should have at least one of these headers: To, Cc or Bcc."]);
```
Skrypt, line 75: I'm asking myself if the Mailkit library now won't return proper error messages when we update it to the latest version like you saw here https://github.com/OrchardCMS/OrchardCore/pull/10863

Skrypt, line 75: I'm asking myself if the Mailkit library now won't return proper error messages when we update it to the latest version like you saw here #10863

I would tend to let the message fail without throwing a specific error message here. Let the library do its work if that's possible.

test/OrchardCore.Tests/Modules/OrchardCore.Email/Workflows/EmailTaskTests.cs
```diff
@@ -0,0 +1,73 @@
...
+using OrchardCore.Email.Workflows.Activities;
+using OrchardCore.Workflows.Models;
+using OrchardCore.Workflows.Services;
+using Xunit;
+
+namespace OrchardCore.Tests.Modules.OrchardCore.Email.Workflows
+{
+    public class EmailTaskTests
+    {
+        private static readonly IDictionary<string, object> EmptyDictiony = new Dictionary<string, object>();
```
Skrypt, line 21: EmptyDictionary

test/OrchardCore.Tests/Modules/OrchardCore.Email/Workflows/EmailTaskTests.cs
```diff
@@ -0,0 +1,73 @@
...
+                HtmlEncoder.Default)
+            {
+                Subject = new WorkflowExpression<string>("Test"),
+                Body = new WorkflowExpression<string>("Test message!!")
+            };
+
+            var executionContext = new WorkflowExecutionContext(
+                new WorkflowType(),
+                new Workflow(),
+                EmptyDictiony,
```
Skrypt, line 40: EmptyDictionary

test/OrchardCore.Tests/Modules/OrchardCore.Email/Workflows/EmailTaskTests.cs
hishamco: Good catch

src/OrchardCore/OrchardCore.Email.Core/Services/SmtpService.cs
hishamco: I think there's no throwing here, I suggest to log an explicit message to make it clear. I'm not sure if the server reponse returns a specific or general message

Let's accept this as its, then we can add the server response to `SmtpResult` as I suggested in related link

Skrypt (approved):

PR #10865, opened 2021-12-12
admin filter term names

### LINKED ISSUE #10862: Admin filter terms should be more descriptive
Currently we have the following filter terms  for content items : 

- status
- sort
- type
- text

Propose to replace with more descriptive term names like : 

- contentitem.status
- contentitem.sort
- contentitem.contenttype
- contentitem.displaytext

The reasons are : 

- Easier to understand where the mapping is going on.
- Less error prone to conflicts from custom indexes. Eg I had a custom index on an Issue contenttype with a status field .  I named my custom term  "status" which conflicted with the existing status term.  By having a guideline to prefix with contenttype  its more unlikely to have conflicts . eg so I could write contentitem.status:Published issue.status=InProgress  

@deanmarcussen

giannik:
fixes #10862


giannik:
wont implement. see issue comments

PR #10881, opened 2021-12-15
Fix VS Code launcher and tasks for dotnet watch

Skrypt:
WIP

- [x] Remove publish launchers. They can be executed as tasks from the command palette in VS Code.
- [x] Add `dotnet watch` task.
- [x] Make `attach to process` work when using `dotnet watch`
- [x] Fix `gulp build` task
- [x] Fix `dotnet run` task. Specifying a framework target since now we use multi-targetting. 
- [x] Make `dotnet watch` launcher config so that we can also launch the browser automatically like before. 

The `dotnet watch` solution depends on : 
See https://github.com/Trottero/dotnet-watch-attach
VS Code Extension : https://marketplace.visualstudio.com/items?itemName=Trottero.dotnetwatchattach

## Launchers

image

## Tasks

https://user-images.githubusercontent.com/3228637/156949586-64a3d510-99cf-4b36-9576-7c530dc98044.mp4


src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
```diff
@@ -13,6 +13,11 @@
     <RazorRuntimeCompilation>false</RazorRuntimeCompilation>
   </PropertyGroup>
 
+  <!-- Watcher include and excludes -->
+  <ItemGroup>
+      <Watch Include="**\*.cs" Exclude="Recipes\**;Assets\**;node_modules\**\*;**\*.js.map;obj\**\*;bin\**\*" />
```
Skrypt, line 18: Fixes issue with watcher reporting : too many changes at once in directory.

Skrypt:
This works perfectly without any issues now.

.vscode/tasks.json
```diff
@@ -198,23 +235,25 @@
...
+        "dotnet publish (framework dependent) - net6.0"
       ]
     },
     {
-      "label": "Gulp Build",
+      "label": "gulp build",
+      "type": "shell",
+      "command": "gulp",
       "args": [
-        "gulp build"
```
jptissot, line 217: Should this be npm run build instead ?

Skrypt: never tried but that's for debugging the gulpfile.js

jptissot: Yes, but this required gulp to be installed globally. `npm run build` does not as it uses gulp from node_modules.

Skrypt: That's the default task suggested. npm run build will try to install it if not found every time whereas this will just execute what is needed knowing that it is already installed globally or locally. I don't mind changing it, but need to test it first.

jptissot: Npm run build does not install first as far as I know. It simply runs a script as defined in the package.json script section.

Skrypt: You're right. But then it will do the same if the gulp package is not installed. So, here it is just a task that executes the gulp build command to debug it without knowing if it's installed or not. What you suggest is to run the gulp pipeline by passing through npm which could lead to executing also something else than gulp.

Skrypt: To be honest, we keep it there because it was there and I updated the task to comply with the latest VS Code tasks implementation details; but it's not something that we use much. Also, it is looking like we might need to move to WebPack or something else at some point.

Skrypt:
Is anyone interested in this pull request to be merged or should I remove the "dotnet watch" part of it?

jptissot:
I like it. Personally I just have powershell alias so I can write `oc` or `ocw` and it runs or watches the project in the console. And then I just attach to the process when I need to debug.

Skrypt:
Ok, well I will wait for someone to approve the PR 😄

jptissot (approved):

PR #10888, opened 2021-12-15
Add dotnet watch excludes

Skrypt:
To lighten the filewatcher process. It fixes the issue where it reports in the console that there are too many changes in a folder.

### DIFF
--- a/src/Templates/OrchardCore.ProjectTemplates/content/OrchardCore.Templates.Cms.Web/OrchardCore.Templates.Cms.Web.csproj
+++ b/src/Templates/OrchardCore.ProjectTemplates/content/OrchardCore.Templates.Cms.Web/OrchardCore.Templates.Cms.Web.csproj
@@ -11,6 +11,11 @@
     <Folder Include="Localization\" />
   </ItemGroup>
 
+  <!-- Watcher include and excludes -->
+  <ItemGroup>
+      <Watch Include="**\*.cs" Exclude="Recipes\**;Assets\**;node_modules\**\*;**\*.js.map;obj\**\*;bin\**\*" />
+  </ItemGroup>
+
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


src/Templates/OrchardCore.ProjectTemplates/content/OrchardCore.Templates.Cms.Web/OrchardCore.Templates.Cms.Web.csproj
```diff
@@ -11,6 +11,11 @@
     <Folder Include="Localization\" />
   </ItemGroup>
 
+  <!-- Watcher include and excludes -->
+  <ItemGroup>
+      <Watch Include="**\*.cs" Exclude="Recipes\**;Assets\**;node_modules\**\*;**\*.js.map;obj\**\*;bin\**\*" />
```
sebastienros, line 16: Why Include and Exclude attributes?
`Include="**\*.cs"` does it mean that they are not watched otherwise, but then the Exclude would be useless.
What about cshtml, liquid, ?

Skrypt:
That's because it requires an include and exclude to work. Else it complains about it missing an include so I added the a value already present in the defaults.

src/Templates/OrchardCore.ProjectTemplates/content/OrchardCore.Templates.Cms.Web/OrchardCore.Templates.Cms.Web.csproj
Skrypt: That's because it requires an include and exclude to work. Else it complains about it missing an include so I added the a value already present in the defaults. Here is the documentation about this : https://docs.microsoft.com/en-us/aspnet/core/tutorials/dotnet-watch?view=aspnetcore-6.0#customize-files-list-to-watch

PR #10889, opened 2021-12-16
Add migration for LocalizedContentItemIndex Latest column

### LINKED ISSUE #10756: LocalizedContentItemIndex Latest column, added but not migrated?
Recently upgraded our version of OrchardCore, and we ran into something that I believe is an issue, but wanted to double check.

The ContentLocalization Migrations.cs added a new migration to add `Latest` to the `LocalizedContentItemIndex` table.
In addition the `DefaultContentLocalizationManager` updated some queries (`GetItemsForSetAsync`) that use this new value.

Problem we see with older data is that this column is NULL for versions that are Latest, so they don't get returned by the query anymore.

Seems like we will need to backfill this column properly, sound correct?

Skrypt:
Fixes #10756 

So, this will remove every record that is not `Latest = true` in the `LocalizedContentItemIndex` because of the `LocalizedContentItemIndexProvider`.

Else solution will be to simply do a SQL Query snippet and add it to documentation. But I believe that this works fine since removed records should also be removed from the Index. I'm not sure what is the reason why we keep removed records in there. @jptissot

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentLocalization/Migrations.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentLocalization/Migrations.cs
@@ -1,14 +1,11 @@
-using System.Linq;
 using Microsoft.Extensions.DependencyInjection;
 using OrchardCore.ContentLocalization.Models;
 using OrchardCore.ContentManagement;
 using OrchardCore.ContentManagement.Metadata;
 using OrchardCore.ContentManagement.Metadata.Settings;
-using OrchardCore.ContentManagement.Records;
 using OrchardCore.Data.Migration;
 using OrchardCore.Environment.Shell.Scope;
 using YesSql;
-using YesSql.Services;
 using YesSql.Sql;
 
 namespace OrchardCore.ContentLocalization.Records
--- a/src/OrchardCore.Modules/OrchardCore.ContentLocalization/Migrations.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentLocalization/Migrations.cs
@@ -1,7 +1,14 @@
+using System.Linq;
+using Microsoft.Extensions.DependencyInjection;
 using OrchardCore.ContentLocalization.Models;
+using OrchardCore.ContentManagement;
 using OrchardCore.ContentManagement.Metadata;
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.ContentLocalization/Migrations.cs
```diff
@@ -71,5 +78,27 @@ public int UpdateFrom2()
...
             return 3;
         }
+
+        // Migrate null LocalizedContentItemIndex Latest column.
+        public int UpdateFrom3()
+        {
+            // Defer this until after the subsequent migrations have succeded as the schema has changed.
+            ShellScope.AddDeferredTask(async scope =>
+            {
+                var session = scope.ServiceProvider.GetRequiredService<ISession>();
```
Skrypt, line 88: Here it cannot find the session in the deferred task.

Skrypt: Nevermind I found the issue

jtkech (commented):
LGTM

Just need to remove the testing code that is commented out

Hmm yes, the only concern is if there is a large amount of localized items ;)

jtkech (approved):

Skrypt:
I'm not sure if YesSQL will batch these or not. But I did a PR to show we can do it. Though, I don't remember how to batch these.

Skrypt:
The only concern I have is about request timeouts with these if there is a large amount of data to be migrated.  Else, the SQL Query is pretty simple to compose.

jtkech:
Yes you're right, hmm don't remember if still in the same transaction, I think so but not sure, but as I remember at least it splits the sql commands in multiple batches.

jtkech:
> Else, the SQL Query is pretty simple to compose.

Yes but may be still expensive on saving, it also depends on how many other index providers will execute

To do this kind of global updates I use custom code (triggered by an user action) that executes batches in separate scopes / transactions

Skrypt:
I don't think there will be any IndexingProviders involved here else than the Localization one.
But yes, just like the Lucene rebuild process we need to come up with a better solution. Never had time to do it and it's fairly simple.

PR #10919, opened 2021-12-23
Update npm packages

agriffard:
typescript `4.5.4`
webpack `5.65.0`


Skrypt:
Not 100% sure about this one as you are also updating other dependencies for some modules at the same time.
Does GraphiQL still compile without issues reported for example?

agriffard:
Yes, GraphiQL still compiles.

I didn't change sensitive packages.

PR #10923, opened 2021-12-24
Optimize SlugService

hishamco:
Addresses #10922

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Liquid/Services/SlugService.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Liquid/Services/SlugService.cs
@@ -8,6 +8,7 @@ namespace OrchardCore.Liquid.Services
     public class SlugService : ISlugService
     {
         private const char Hyphen = '-';
+        private const int MaxLength = 1000;
 
         private static readonly char[] _allowedSymbols = new char[] { Hyphen, '_', '~' };
 
@@ -63,8 +64,9 @@ namespace OrchardCore.Liquid.Services
                     }
                 }
             }
-
-            return slug.ToString().Normalize(NormalizationForm.FormC);
+            return slug
+                .ToString()[..Math.Min(slug.Length, MaxLength)]
+                .Normalize(NormalizationForm.FormC);
         }
     }
 }
--- a/src/OrchardCore.Modules/OrchardCore.Liquid/Services/SlugService.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Liquid/Services/SlugService.cs
@@ -26,7 +26,7 @@ namespace OrchardCore.Liquid.Services
             {
                 var currentChar = normalizedText[i];
 
-                if (CharUnicodeInfo.GetUnicodeCategory(normalizedText[i]) == UnicodeCategory.NonSpacingMark)
+                if (CharUnicodeInfo.GetUnicodeCategory(currentChar) == UnicodeCategory.NonSpacingMark)
                 {
                     continue;
                 }
--- a/src/OrchardCore.Modules/OrchardCore.Liquid/Services/SlugService.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Liquid/Services/SlugService.cs
@@ -2,6 +2,7 @@ using System;
 using System.Globalization;
 using System.Linq;
 using System.Text;
+using Cysharp.Text;
 
 namespace OrchardCore.Liquid.Services
 {
@@ -10,8 +11,6 @@ namespace OrchardCore.Liquid.Services
         private const char Hyphen = '-';
         private const int MaxLength = 1000;
 
-        private static readonly char[] _allowedSymbols = new char[] { Hyphen, '_', '~' };
-
         public string Slugify(string text)
         {
             if (String.IsNullOrEmpty(text))
@@ -19,13 +18,14 @@ namespace OrchardCore.Liquid.Services
                 return text;
             }
 
-            var slug = new StringBuilder();
             var appendHyphen = false;
-            var normalizedText = text.ToLowerInvariant().Normalize(NormalizationForm.FormKD);
+            var normalizedText = text.Normalize(NormalizationForm.FormKD);
+
+            using var slug = ZString.CreateStringBuilder();
 
             for (var i = 0; i < normalizedText.Length; i++)
             {
-                var currentChar = normalizedText[i];
+                var currentChar = Char.ToLowerInvariant(normalizedText[i]);
 
                 if (CharUnicodeInfo.GetUnicodeCategory(currentChar) == UnicodeCategory.NonSpacingMark)
                 {
@@ -38,21 +38,17 @@ namespace OrchardCore.Liquid.Services
 
                     appendHyphen = true;
                 }
-                else if(_allowedSymbols.Contains(currentChar))
+                else if (currentChar is Hyphen)
                 {
-                    if (currentChar == Hyphen)
-                    {
-                        if (appendHyphen && i != normalizedText.Length - 1)
-                        {
-                            slug.Append(currentChar);
-
-                            appendHyphen = false;
-                        }
-                    }
-                    else
+                    if (appendHyphen && i != normalizedText.Length - 1)
                     {
                         slug.Append(currentChar);
+                        appendHyphen = false;
                     }
+                }
+                else if (currentChar == '_' || currentChar == '~')
+                {
+                    slug.Append(currentChar);
                 }
                 else
                 {
@@ -64,9 +60,8 @@ namespace OrchardCore.Liquid.Services
                     }
                 }
             }
-            return slug
-                .ToString()[..Math.Min(slug.Length, MaxLength)]
-                .Normalize(NormalizationForm.FormC);
+
+            return new string(slug.AsSpan()[..Math.Min(slug.Length, MaxLength)]).Normalize(NormalizationForm.FormC);
         }
     }
 }
--- a/src/OrchardCore.Modules/OrchardCore.Liquid/Services/SlugService.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Liquid/Services/SlugService.cs
@@ -1,75 +1,70 @@
+using System;
 using System.Globalization;
+using System.Linq;
 using System.Text;
 
 namespace OrchardCore.Liquid.Services
 {
     public class SlugService : ISlugService
     {
+        private const char Hyphen = '-';
+
+        private static readonly char[] _allowedSymbols = new char[] { Hyphen, '_', '~' };
+
         public string Slugify(string text)
         {
-            if (string.IsNullOrEmpty(text))
+            if (String.IsNullOrEmpty(text))
             {
                 return text;
             }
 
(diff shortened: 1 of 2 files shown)


hishamco:
The benchmark result show that there's a slightly bit improvments:

|     Method |     Mean |     Error |    StdDev | Ratio | RatioSD |  Gen 0 | Allocated |
|----------- |---------:|----------:|----------:|------:|--------:|-------:|----------:|
| NewSlugify | 1.409 us | 0.0174 us | 0.0163 us |  0.97 |    0.02 | 0.2174 |     456 B |
| OldSlugify | 1.453 us | 0.0202 us | 0.0256 us |  1.00 |    0.00 | 0.2327 |     488 B |

test/OrchardCore.Benchmarks/SlugBenchmark.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        private static readonly ISlugService _slugService;
+
+        static SlugBenchmark()
+        {
+            _slugService = new SlugService();
+        }
+
+        /*
+         * Summary 24th November 2021
```
hishamco, line 18: December ;)

src/OrchardCore.Modules/OrchardCore.Liquid/Services/SlugService.cs
```diff
@@ -1,75 +1,70 @@
...
-            }
+                else
+                {
+                    if (appendHyphen)
+                    {
+                        slug.Append(Hyphen);
 
-            if (slug.Length > 1000)
-            {
-                slug = slug.Substring(0, 1000);
```
deanmarcussen, line 67: You appear to have removed this logic?

src/OrchardCore.Modules/OrchardCore.Liquid/Services/SlugService.cs
```diff
@@ -1,75 +1,70 @@
...
         public string Slugify(string text)
         {
-            if (string.IsNullOrEmpty(text))
+            if (String.IsNullOrEmpty(text))
             {
                 return text;
             }
 
-            var sb = new StringBuilder();
+            var slug = new StringBuilder();
```
deanmarcussen, line 21: if you are optimizing it, wouldn't you use `ZString` ?

deanmarcussen (commented):
Looks like you removed features which seems odd?

src/OrchardCore.Modules/OrchardCore.Liquid/Services/SlugService.cs
hishamco: I will try this

src/OrchardCore.Modules/OrchardCore.Liquid/Services/SlugService.cs
hishamco: Could you please let me know what's the use of it? Is it to only take 1000 chars if the slug exceed the limits or is there something special

src/OrchardCore.Modules/OrchardCore.Liquid/Services/SlugService.cs
hishamco: I run this using ZString the results are:

|          Method |     Mean |     Error |    StdDev |   Gen 0 | Allocated |
|---------------- |---------:|----------:|----------:|--------:|----------:|
| EvaluateSlugify | 4.659 us | 0.1532 us | 0.0084 us | 31.2500 |     64 KB |

Also I run it again without

|          Method |     Mean |     Error |    StdDev |  Gen 0 | Allocated |
|---------------- |---------:|----------:|----------:|-------:|----------:|
| EvaluateSlugify | 1.423 us | 0.4246 us | 0.0233 us | 0.2174 |     456 B |

Seems ZString allocate less memory but I don't have much knowlege about the other factors, so @sebastienros could you please let me know which is better and what other factors mean

lahma: Pooling will have its cost, allocations reduced but more machinery running.  The sample you showed is quite short string, is it the normal case. Also if this code is run say 5-10 times a page request the benefits would so with more stress - now CPU is not needed for GC as sample runs so fast and does little - load benchmark would probably show more difference.

hishamco: Thanks for your information Marko, I may create another benchmark for a long string and see how it iss going

lahma: OK now I'm reading the table again (I though they both were B) the ZString allocation looks really odd, can you show the diff for code you use with ZString and without?

hishamco: I just use 

```csharp
var slug = ZString.CreateStringBuilder();
```

instead of
```csharp
var slug = new StringBuilder();
```

hishamco: StringBuilder

|                      Method |     Mean |    Error |   StdDev |  Gen 0 | Allocated |
|---------------------------- |---------:|---------:|---------:|-------:|----------:|
| EvaluateSlugifyWithLongSlug | 16.97 us | 1.529 us | 0.084 us | 2.9907 |      6 KB |

ZString.CreateStringBuilder

|                      Method |     Mean |    Error |   StdDev |   Gen 0 | Allocated |
|---------------------------- |---------:|---------:|---------:|--------:|----------:|
| EvaluateSlugifyWithLongSlug | 20.61 us | 5.802 us | 0.318 us | 32.2571 |     68 KB |

Seems the current implementation is fine, unless there's something I don't know about ZString

lahma:
OK I went through the thing and the problem was that the ZString didn't didn't have `using`, it was leaking all the time thus bad perf. Here are my numbers (I changed the benchmark to have `private static readonly SlugService _slugService;` instead of `ISlugService` because I don't want to test interface dispatch):

``` ini

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.22000
AMD Ryzen 9 5950X, 1 CPU, 32 logical and 16 physical cores
.NET SDK=6.0.101
  [Host]     : .NET 6.0.1 (6.0.121.56705), X64 RyuJIT
  DefaultJob : .NET 6.0.1 (6.0.121.56705), X64 RyuJIT


```

## Original (main branch)

|          Method |     Mean |   Error |  StdDev |  Gen 0 | Allocated |
|---------------- |---------:|--------:|--------:|-------:|----------:|
| EvaluateSlugify | 407.6 ns | 1.76 ns | 1.37 ns | 0.0291 |     488 B |


## StringBuilder

|          Method |     Mean |   Error |  StdDev |  Gen 0 | Allocated |
|---------------- |---------:|--------:|--------:|-------:|----------:|
| EvaluateSlugify | 410.3 ns | 0.97 ns | 0.81 ns | 0.0272 |     456 B |


## ZString

|          Method |     Mean |   Error |  StdDev |  Gen 0 | Allocated |
|---------------- |---------:|--------:|--------:|-------:|----------:|
| EvaluateSlugify | 419.6 ns | 2.07 ns | 1.84 ns | 0.0148 |     248 B |


## ZString v2

|          Method |     Mean |   Error |  StdDev |  Gen 0 | Allocated |
|---------------- |---------:|--------:|--------:|-------:|----------:|
| EvaluateSlugify | 353.6 ns | 1.47 ns | 1.30 ns | 0.0100 |     168 B |

## Code for ZString v2

This was my experiment to show how to optimize the implementation, I might have introduced a bug with ToLowerInvariant change, but it should still be faster even without it.  The output stayed the same though for given example.

I've commented the code to show what's changed. General easy rule (when implicit usings aren't being used): If there's `using System.Linq;` at the top of the performance critical code, try to find a way to get rid of it.

```csharp
public class SlugService : ISlugService
{
    private const char Hyphen = '-';
    private const int MaxLength = 1000;

    public string Slugify(string text)
    {
        if (String.IsNullOrEmpty(text))
        {
            return text;
        }

        // one thing to consider here, if text < 20 or some other threshold should the code do pre-check
        // with foreach if all the chars are valid already (letter, digit or hyphen) and skip whole process?
        // this is why it's important to study common input and optimize usual cases

        // You need to have using directive in order for the pooling to work
        using var slug = ZString.CreateStringBuilder();
        var appendHyphen = false;

        // removed ToLowerInvariant from here, I'm unsure if I break something here
        var normalizedText = text.Normalize(NormalizationForm.FormKD);

        for (var i = 0; i < normalizedText.Length; i++)
        {
            // do ToLowerInvariant for each char so I don't need to create a new string above
            var currentChar = Char.ToLowerInvariant(normalizedText[i]);

            if (CharUnicodeInfo.GetUnicodeCategory(currentChar) == UnicodeCategory.NonSpacingMark)
            {
                continue;
            }

            if (Char.IsLetterOrDigit(currentChar))
            {
                slug.Append(currentChar);
                appendHyphen = true;
            }
            // split the Contains to two different branches, you shouldn't use LINQ method Contains for an
            // array, it's really bad for performance
            else if (currentChar is Hyphen)
            {
                if (appendHyphen && i != normalizedText.Length - 1)
                {
                    slug.Append(currentChar);
                    appendHyphen = false;
                }
            }
            // fast char equality
            else if (currentChar is '_' or '~')
            {
                slug.Append(currentChar);
            }
            else
            {
                if (appendHyphen)
                {
                    slug.Append(Hyphen);
                    appendHyphen = false;
                }
            }
        }

        // old code was doing a ToString() and then a Substring (two string allocations)
        // we can get a span from the builder and construct new string based on that
        return new string(slug.AsSpan()[..Math.Min(slug.Length, MaxLength)]).Normalize(NormalizationForm.FormC);
    }
}

```

src/OrchardCore.Modules/OrchardCore.Liquid/Services/SlugService.cs
lahma: Yes, you were missing `using` for the ZString creation. I've commented my results into PR main discussion.

hishamco:
Thanks a lot @lahma, I need to test your changes with the suggestion that you did in a BiG string, could you please let me know the CLI command that you use to make sure I don't get another results

lahma:
I just do `dotnet run -c Release --framework net6.0` in benchmark project folder, virus scanner disabled and minimal apps running.

hishamco:
> dotnet run -c Release --framework net6.0

I though you mark the `--job short`, anyhow I will made your changes and benchmark again

hishamco:
How this **else if (currentChar is '_' or '~')** working with you, are you changed the lang version?

hishamco:
I got the below results using a normal or condition

|                       Method |      Mean |     Error |    StdDev |  Gen 0 | Allocated |
|----------------------------- |----------:|----------:|----------:|-------:|----------:|
| EvaluateSlugifyWithShortSlug |  1.286 us | 0.0255 us | 0.0389 us | 0.0801 |     168 B |
|  EvaluateSlugifyWithLongSlug | 21.137 us | 0.3821 us | 0.5356 us | 1.1902 |   2,528 B |

sebastienros (approved):

PR #10933, opened 2021-12-30
Fixes index cursors management

### LINKED ISSUE #10937: Index cursor ordering issue
When there are hundreds of thousands rows the results returned by `QueryIndex()` may not be ordered by `Id` by default. In that case services as `AutorouteEntries` using an index cursor on the `Id` column are not working properly, the `_lastIndexId` being not well managed. In my case when first publishing an item having an Autoroute, in place of only adding one Autoroute entry, it was trying to add thousands of Autoroute entries in an unordered way and was taking a very long time.

jtkech:
Fixes #10937

When there are hundreds of thousands rows the results returned by `QueryIndex()` may not be ordered by `Id` by default. In that case services as `AutorouteEntries` using an index cursor on the `Id` column are not working properly, the `_lastIndexId` being not well managed. In my case when first publishing an item having an Autoroute, in place of only adding one Autoroute entry, it was trying to add thousands of Autoroute entries in an unordered way and was taking a very long time.

**Updated**

Also enhances the lookup of Autoroute entries to evict, the ones related to contained items of a given container.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Autoroute/Services/AutorouteEntries.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Autoroute/Services/AutorouteEntries.cs
@@ -81,12 +81,20 @@ namespace OrchardCore.Autoroute.Services
 
         protected void AddEntries(IEnumerable<AutorouteEntry> entries)
         {
+            var entriesByContainer = _paths.Values
+                .Where(x => !String.IsNullOrEmpty(x.ContainedContentItemId))
+                .ToLookup(x => x.ContentItemId);
+
             // Evict all entries related to a container item from autoroute entries.
             // This is necessary to account for deletions, disabling of an item, or disabling routing of contained items.
             foreach (var entry in entries.Where(x => String.IsNullOrEmpty(x.ContainedContentItemId)))
             {
-                var entriesToRemove = _paths.Values.Where(x => x.ContentItemId == entry.ContentItemId &&
-                    !String.IsNullOrEmpty(x.ContainedContentItemId));
+                if (!entriesByContainer.Contains(entry.ContentItemId))
+                {
+                    continue;
+                }
+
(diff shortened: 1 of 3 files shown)


src/OrchardCore.Modules/OrchardCore.Autoroute/Services/AutorouteEntries.cs
```diff
@@ -147,7 +155,10 @@ private async Task RefreshEntriesAsync(AutorouteStateDocument state)
             {
                 if (_stateIdentifier != state.Identifier)
                 {
-                    var indexes = await Session.QueryIndex<AutoroutePartIndex>(i => i.Id > _lastIndexId).ListAsync();
+                    var indexes = await Session
+                        .QueryIndex<AutoroutePartIndex>(i => i.Id > _lastIndexId)
+                        .OrderBy(i => i.Id)
```
deanmarcussen, line 160: Interesting, I'm intrigued at the sql that it produces, as to why it would fail without an order by?

But I believe you :)

jtkech: There is no sql query that fails, but when there are many items the results may not be ordered by the `Id` column, and we rely on this order to manage the lastId cursor that we use to only query new rows.

The result is that when 1st publishing an item, in place of retrieving only one row and process it, you may retrieve many ones. Also, to remove all previous entries of contained items of containers, we were doing a lookup on the whole `_paths` dico values, this on each entry to add, here we now do only one `.ToLookup()`.

For info, in my case a 1st publishing was taking around 5 minutes, after this 1st publishing the lastId was in sync and then no problem on following ones.

deanmarcussen: Ah I see, because you use LastOrDefault() to determine. Makes sense. 
Merge when you are ready.

PR #10939, opened 2021-12-31
Make NullStringLocalizer & HtmlLocalizer Public

hishamco:
This PR solved two issue:

- Make it easier to create `NullStringLocalizer` and `NullHtmlLocalizer` without need to use localizer factory, which quite useful in unit testing
- Adds a generic version of the null localizer, to make the controllers and other types that are using `IStringLocalizer<T>` and/or `IHtmlLocalizer<T>` testable without using mocks

`OrchardCore.Localization.Abstractions` is a suitable place for such types, something similiar to [`NullLogger`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.abstractions.nulllogger?view=dotnet-plat-ext-6.0)

No breaking-changes here, `NullStringLocalizerFactory` and `NullHtmlLocalizerFactory` are still using the above types without any issues


hishamco:
@Skrypt could you please have a look if you have a time, seems Seb is quite busy ;)

Skrypt:
Ok looks good to me. The only thing I've not looked at is the generic classes utility. Else, this is just some moving around of classes for cleaner overall usage.

src/OrchardCore/OrchardCore/Localization/NullHtmlLocalizerFactory.cs
```diff
@@ -14,46 +11,9 @@ namespace OrchardCore.Localization
     public class NullHtmlLocalizerFactory : IHtmlLocalizerFactory
     {
         /// <inheritdocs />
-        public IHtmlLocalizer Create(string baseName, string location) => NullLocalizer.Instance;
+        public IHtmlLocalizer Create(string baseName, string location) => NullHtmlLocalizer.Instance;
 
         /// <inheritdocs />
-        public IHtmlLocalizer Create(Type resourceSource) => NullLocalizer.Instance;
-
-        private class NullLocalizer : IHtmlLocalizer
```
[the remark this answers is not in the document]
hishamco: Localizer factories need to be instantiated to create localizers. In case to use null localizer you could use `NullStringLocalizer.Instance`

hishamco:
@sebastienros could we merge this?

hishamco:
Can we merge this?

hishamco:
@Skrypt any other thing to add here or shall I merge?

src/OrchardCore/OrchardCore/Localization/NullHtmlLocalizerFactory.cs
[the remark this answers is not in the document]
Skrypt: Do we still need NullLocalizer here or should we remove it since the goal is to make this public?

src/OrchardCore/OrchardCore/Localization/NullStringLocalizerFactory.cs
```diff
@@ -13,10 +12,10 @@ namespace OrchardCore.Localization
...
     {
         /// <inheritdocs />
-        public IStringLocalizer Create(Type resourceSource) => NullLocalizer.Instance;
+        public IStringLocalizer Create(Type resourceSource) => NullStringLocalizer.Instance;
 
         /// <inheritdocs />
-        public IStringLocalizer Create(string baseName, string location) => NullLocalizer.Instance;
+        public IStringLocalizer Create(string baseName, string location) => NullStringLocalizer.Instance;
 
         internal class NullLocalizer : IStringLocalizer
```
Skrypt, line 20: Same here do we still need?

src/OrchardCore/OrchardCore.Localization.Abstractions/NullHtmlLocalizerOfT.cs
```diff
@@ -0,0 +1,38 @@
...
+using Microsoft.AspNetCore.Mvc.Localization;
+using Microsoft.Extensions.Localization;
+
+namespace OrchardCore.Localization
+{
+    /// <summary>
+    /// Minimalistic HTML-aware localizer that does nothing.
+    /// </summary>
+    /// <typeparam name="T">The <see cref="Type"/> to provide strings for.</typeparam>
+    public class NullHtmlLocalizer<T> : IHtmlLocalizer<T>
```
Skrypt, line 13: Here, I'm not sure if it's necessary to have a NullHtmlLocalizer of T since in the end it is just something to tell that there is no localizer available. Why would we need to pass a Type? The only usage to me would be if we had non-localizable types?

hishamco: Seems there's no need for `NullHtmlLocalizer<T>` at all

Skrypt (approved):

PR #10941, opened 2022-01-02
Reduce allocations in MediaTokenService and improve performance

lahma:
One has to have hobbies, right? Felt like a fun exercise as Orchard can use all .NET 6 goodness. 

`MediaTokenService` does some unnecessary allocations and is called on every request which has images served.

* ~~tweak constructor to iterate array, DI gives array and it's fastest to traverse - constructor takes considerable amount of time (see scope note below)~~
  * no need after changing service to singleton, shaved off extra 200ns without the constructor invoke
* don't `ToString` `StringValues`, keep them as-is as in dictionary once built
* parse commands and others commands with one pass using constructs that `QueryHelpers` uses
* allocate static command arrays to return from processors, some of them were always returning new ones
* change `GetHash` not to allocation anonymous closure each time by using `Try`/`Set `construct, uses same logic that extension method `GetOrCreate` uses
* change hash calculation to use stackalloc and spans for workloads < 1024 bytes (common case)
* use custom `AddQueryString `method that understands spans and concrete dictionary type

`MediaTokenService` seems to be scoped service and thus known commands need to be initialized for every request, is this a tenancy things and cannot be made singleton?

## Benchmarks

``` ini

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.22000
AMD Ryzen 9 5950X, 1 CPU, 32 logical and 16 physical cores
.NET SDK=6.0.101
  [Host]     : .NET 6.0.1 (6.0.121.56705), X64 RyuJIT
  DefaultJob : .NET 6.0.1 (6.0.121.56705), X64 RyuJIT


```

### Before

|                          Method |     Mean |     Error |    StdDev |  Gen 0 |  Gen 1 | Allocated |
|-------------------------------- |---------:|----------:|----------:|-------:|-------:|----------:|
|                  AddTokenToPath | 1.335 us | 0.0097 us | 0.0091 us | 0.1984 |      - |      3 KB |
|          AddTokenToPath_NoCache | 1.993 us | 0.0089 us | 0.0083 us | 0.2441 |      - |      4 KB |
|         AddTokenToPath_LongPath | 2.474 us | 0.0157 us | 0.0139 us | 0.4005 | 0.0038 |      7 KB |
| AddTokenToPath_LongPath_NoCache | 3.135 us | 0.0098 us | 0.0091 us | 0.4578 | 0.0038 |      7 KB |

### After

Now testing with singleton - unlike main branch!

|                          Method |       Mean |   Error |  StdDev |  Gen 0 | Allocated |
|-------------------------------- |-----------:|--------:|--------:|-------:|----------:|
|                  AddTokenToPath |   696.1 ns | 1.18 ns | 1.05 ns | 0.0715 |      1 KB |
|          AddTokenToPath_NoCache | 1,309.6 ns | 2.28 ns | 1.91 ns | 0.1087 |      2 KB |
|         AddTokenToPath_LongPath | 1,556.8 ns | 4.93 ns | 4.37 ns | 0.2060 |      3 KB |
| AddTokenToPath_LongPath_NoCache | 2,248.9 ns | 6.70 ns | 6.27 ns | 0.2441 |      4 KB |

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenService.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenService.cs
@@ -2,10 +2,12 @@ using System;
 using System.Collections.Generic;
 using System.Security.Cryptography;
 using System.Text;
+using System.Text.Encodings.Web;
 using Cysharp.Text;
 using Microsoft.AspNetCore.WebUtilities;
 using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Options;
+using Microsoft.Extensions.Primitives;
 using SixLabors.ImageSharp.Web.Processors;
 
 namespace OrchardCore.Media.Processing
@@ -15,7 +17,7 @@ namespace OrchardCore.Media.Processing
         private const string TokenCacheKeyPrefix = "MediaToken:";
         private readonly IMemoryCache _memoryCache;
 
-        private readonly HashSet<string> _knownCommands = new HashSet<string>();
+        private readonly HashSet<string> _knownCommands = new HashSet<string>(12);
         private readonly byte[] _hashKey;
 
         public MediaTokenService(
@@ -24,6 +26,7 @@ namespace OrchardCore.Media.Processing
             IEnumerable<IImageWebProcessor> processors)
         {
             _memoryCache = memoryCache;
+
             foreach (var processor in processors)
             {
                 foreach (var command in processor.Commands)
@@ -39,34 +42,24 @@ namespace OrchardCore.Media.Processing
         {
             var pathIndex = path.IndexOf('?');
 
-            var parsed = pathIndex != -1
-                ? QueryHelpers.ParseQuery(path.Substring(pathIndex + 1))
-                : null;
+            Dictionary<string, StringValues> processingCommands = null;
+            Dictionary<string, StringValues> otherCommands = null;
+
+            if (pathIndex != -1)
+            {
+                ParseQuery(path.Substring(pathIndex + 1), out processingCommands, out otherCommands);
+            }
 
             // If no commands or only a version command don't bother tokenizing.
-            if (parsed is null || parsed.Count == 0 || parsed.Count == 1 && parsed.ContainsKey(ImageVersionProcessor.VersionCommand))
+            if (processingCommands is null
+                || processingCommands.Count == 0
+                || processingCommands.Count == 1 && processingCommands.ContainsKey(ImageVersionProcessor.VersionCommand))
             {
                 return path;
             }
 
-            var processingCommands = new Dictionary<string, string>(parsed.Count);
-            Dictionary<string, string> otherCommands = null;
-
-            foreach (var command in parsed)
-            {
-                if (_knownCommands.Contains(command.Key))
-                {
-                    processingCommands[command.Key] = command.Value.ToString();
-                }
-                else
-                {
-                    otherCommands ??= new Dictionary<string, string>();
-                    otherCommands[command.Key] = command.Value.ToString();
-                }
-            }
-
             // Using the command values as a key retrieve from cache
-            var queryStringTokenKey = CreateQueryStringTokenKey(processingCommands.Values);
+            var queryStringTokenKey = CreateQueryStringTokenKey(processingCommands);
             var queryStringToken = GetHash(queryStringTokenKey);
 
             processingCommands["token"] = queryStringToken;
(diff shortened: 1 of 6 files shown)


src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenService.cs
```diff
@@ -15,7 +18,7 @@ public class MediaTokenService : IMediaTokenService
         private const string TokenCacheKeyPrefix = "MediaToken:";
         private readonly IMemoryCache _memoryCache;
 
-        private readonly HashSet<string> _knownCommands = new HashSet<string>();
+        private readonly HashSet<string> _knownCommands = new(12);
```
lahma, line 21: 12 is what default Orchard setup seems to have

src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenService.cs
```diff
@@ -39,34 +44,24 @@ public string AddTokenToPath(string path)
...
-                {
-                    otherCommands ??= new Dictionary<string, string>();
-                    otherCommands[command.Key] = command.Value.ToString();
-                }
+                return path;
             }
 
             // Using the command values as a key retrieve from cache
-            var queryStringTokenKey = CreateQueryStringTokenKey(processingCommands.Values);
+            var queryStringTokenKey = CreateQueryStringTokenKey(processingCommands);
```
lahma, line 64: `.Values` allocates new `ValueCollection`, no need for that so let's just pass the dictionary which has efficient enumeration

src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenService.cs
```diff
@@ -80,12 +75,50 @@ public string AddTokenToPath(string path)
                 }
             }
 
-            return QueryHelpers.AddQueryString(path.Substring(0, pathIndex), processingCommands);
+            return AddQueryString(path.AsSpan(0, pathIndex), processingCommands);
+        }
+
+        private void ParseQuery(
```
lahma, line 81: same as in `QueryHelpers`, but produces two dictionaries based on whether known commands matches

src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenService.cs
```diff
@@ -98,30 +131,97 @@ public bool TryValidateToken(IDictionary<string, string> commands, string token)
             return false;
         }
 
-        private static string CreateQueryStringTokenKey(IEnumerable<string> values)
+        /// <summary>
+        /// Specialized version with fast enumeration and no StringValues string allocation.
+        /// </summary>
+        private static string CreateQueryStringTokenKey(Dictionary<string, StringValues> values)
```
lahma, line 137: a bit of duplication, but with concrete type overload we get fast struct enumerator instead of one wrapped via interface

src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenService.cs
```diff
@@ -98,30 +131,97 @@ public bool TryValidateToken(IDictionary<string, string> commands, string token)
...
             }
             return builder.ToString();
         }
 
         private string GetHash(string queryStringTokenKey)
-            => _memoryCache.GetOrCreate(queryStringTokenKey, entry =>
-               {
-                   entry.SlidingExpiration = TimeSpan.FromHours(5);
+        {
+            if (!_memoryCache.TryGetValue(queryStringTokenKey, out var result))
```
lahma, line 161: old code had capturing closure as extension method took a Func which needed outside context, now using methods actually present in `IMemoryCache` to prevent that

src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenService.cs
```diff
@@ -98,30 +131,97 @@ public bool TryValidateToken(IDictionary<string, string> commands, string token)
...
+            }
+
+            return (string) result;
+        }
+
+        /// <summary>
+        /// Custom version of <see cref="QueryHelpers.AddQueryString(string,string,string)"/> that takes our pre-built
+        /// dictionary, uri as ReadOnlySpan&lt;char&gt; and uses ZString. Otherwise same logic.
+        /// </summary>
+        private static string AddQueryString(
```
lahma, line 195: this ones copes with `ReadOnlySpan<char>` and uses `ZString` instead of allocating `StringBuilder`

deanmarcussen:
> One has to have hobbies, right? Felt like a fun exercise as Orchard can use all .NET 6 goodness.

Indeed :)

We aren't quite there with .NET 6 only yet, still multi targetting. Due to drop 3.1 and 5 after the next release (due shortly)

> `MediaTokenService` seems to be scoped service and thus known commands need to be initialized for every request, is this a tenancy things and cannot be made singleton?

No, it's just Scoped until you find value in making it a Singleton.

lahma:
> > One has to have hobbies, right? Felt like a fun exercise as Orchard can use all .NET 6 goodness.
> 
> Indeed :)
> 
> We aren't quite there with .NET 6 only yet, still multi targetting. Due to drop 3.1 and 5 after the next release (due shortly)

I did a workaround with #if/#else that's the slower path without the zero-allocating query params traversal.

> > `MediaTokenService` seems to be scoped service and thus known commands need to be initialized for every request, is this a tenancy things and cannot be made singleton?
> 
> No, it's just Scoped until you find value in making it a Singleton.

Thanks for the confirmation. I changed it to singleton and tested in PR with singleton instance, which gave 930ns -> 700ns change. Quite big difference I'd say just by singletoning the instance.

src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaTokenService.cs
```diff
@@ -80,12 +74,75 @@ public string AddTokenToPath(string path)
...
+                otherCommands = otherCommandAccumulator.GetResults();
+            }
         }
+#else // slower version until NET 6 is the default
+        private void ParseQuery(
+            string queryString,
+            out Dictionary<string, StringValues> processingCommands,
+            out Dictionary<string, StringValues> otherCommands)
+        {
+            var parsed = QueryHelpers.ParseQuery(queryString);
```
lahma, line 124: basically old logic which generates at least two dictionaries

lahma:
I tweaked existing test case to test both with known and unknown commands and it now checks for complete generated string, results didn't change between main and this branch.

PR #10979, opened 2022-01-07
Drop netcore 3.1 and net 5

deanmarcussen:
@agriffard your other pr regarding the templates and upgrading to program.cs should remain, we will just need to fix the NLog issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Here is a more detailed [roadmap](https://github.com/OrchardCMS/OrchardCore/wiki
 
 - Install the latest version of the .NET SDK from this page <https://dotnet.microsoft.com/download>
 - Next, navigate to `D:\OrchardCore\src\OrchardCore.Cms.Web` or wherever your folder is on the commandline in Administrator mode.
-- Call `dotnet run -f net6.0` (or `dotnet run -f net5.0` depending on your version of .NET SDK).
+- Call `dotnet run`.
 - Then open the `http://localhost:5000` URL in your browser.
 
 ### Visual Studio
--- a/.github/workflows/functional_all_db.yml
+++ b/.github/workflows/functional_all_db.yml
@@ -16,14 +16,6 @@ jobs:
     - uses: actions/setup-node@v2
       with:
         node-version: "15"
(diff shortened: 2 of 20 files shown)


hishamco:
Why we need to drop NetCore 3.1 while still LTS, If I'm not wrong it will be out of support in March. Is there any reason for this change?

Skrypt:
Then stay on 1.2 if you need netcoreapp3.1
Else, the next release of OC will be only in a few months which should drop support for netcoreapp3.1
So, it is fine that the main branch which will feed the preview Nuget packages drops it before time.
It gives time for people to adjust accordingly.

Skrypt (approved):

hishamco:
Thanks for clarification

agriffard (approved):

README.md
```diff
@@ -38,7 +38,7 @@ Here is a more detailed [roadmap](https://github.com/OrchardCMS/OrchardCore/wiki
 
 - Install the latest version of the .NET SDK from this page <https://dotnet.microsoft.com/download>
```
hishamco, line 39: Shall we be explicit about the version 6.0 to avoid unexpected things?

hishamco:
Also these should be removed

https://github.com/OrchardCMS/OrchardCore/blob/066471669ecc1e06c8533b56a50a1cc3ec698a75/src/OrchardCore/OrchardCore.Localization.Core/PortableObject/PortableObjectStringLocalizer.cs#L81-L82

Skrypt:
Obsolete code should be removed before the next release this is not a concern for this PR if it doesn't break the build.

hishamco:
@Skrypt this not OC specific, It's deprecated by ASP.NET Core at the time, we kept there to support multi-targeting

deanmarcussen:
> Also these should be removed

Yes, however there will be other things as well. 
Can be done seperately as we find them.

hishamco (approved):

sebastienros:
@Skrypt anything to do for the vscode launch files?

Skrypt:
It's already set to net6.0 so it should work fine.

Issue #10963, opened 2022-01-05
Improve the OrchardCore.Tenants.FeatureProfiles feature


MikeAlhayek:
### Is your feature request related to a problem? Please describe.
If a profile is renamed, It'll break existing tenants since the name is stored in the appsettings.json file. 

Also, other modules are not able to contribute into the FeatureProfiles setup. For example, I want write a billing module that would calculate the monthly cost for the tenant. I could add `price field` for each profile then sum the total price for billing.

### Describe the solution you'd like
 1. Add an id and store the id in appsettings.json this way renaming a profile does not break anything
 2. Allow the user to have multiple profiles more on than explained in #10962 
 3. Ultimately, user `DisplayDriver` to render the CRUD to allow other modules to contribute into the captured data. 

If the above is acceptable, I can submit a PR to handle this request. I already implemented 1 & 2. I should be able to implement 3 as well.

Skrypt:
Is there no permission required to enable/disable features in a tenant? I suggest you look at that.

ns8482e:
@CrestApps Adding id won't solve the real problem. - IMHO rename should not be allowed IF the profile is assigned to any tenant.

deanmarcussen:
Changing it to an id sounds breaking (and of little benefit).

suggest just making the name non editable (if it wasn’t done that way originally that’s my mistake)

multi select sounds ok as long as it is non breaking. I assume you’ll have to make them comma separated, as we can only support storing strings in the shell sources. It’s a little weird to have a comma separated string stored in json instead of an array, but could live with that.

drivers. Yes, possible, but runs good into the issue of only storing strings, which could make them a bit odd

CrestApps:
@deanmarcussen it should not break anything if we treat existing name as id when id is not stored. This show not break anything and stay backward compatible.

One question I do have, why is `FeatureProfilesValidationProvider` part of `OrcharcCore` project instead of `OrchardCore.Tenants`? I would think it should be also registered in the `OrchardCore.Tenants.FeatureProfiles` startup. Should I move it?

CrestApps:
> @CrestApps Adding id won't solve the real problem. - IMHO rename should not be allowed IF the profile is assigned to any tenant.

[...]

Issue #10962, opened 2022-01-04
In a SaaS setup, how can the SaaS admin/owner manage the tenant features?


MikeAlhayek:
### Is your feature request related to a problem? Please describe.
In a SaaS environment, the service owner may not want the tenants to be able to manage their feature. Instead, I want to manage the features for each tenant so that I control what they have access too.

### Describe the solution you'd like
It would be nice to be able to disable the `OrchardCore.Features` module "At least the UI part" so that the tenant is unable to enable/disable features.

I am aware of the `OrchardCore.Tenants.FeatureProfiles` feature which allows you to define rules for which feature to include/exclude for the tenant. The issue with that is that you can only have a single profile per tenant. What if one tenant should have access to the standard profile, plus access to module A, B and C? Another solution may be to allow the user to select multiple profiles for one tenants to allow more flexible for reusing profiles.

### Describe alternatives you've considered
Revoke access to the features using permissions from a recipe.

deanmarcussen:
Permissions / feature profiles are the good approaches here.

disabling the features feature would make the site unusable / upgradable very easily.

as a saas administrator wanting to restrict the tenant you should just not give them admin rights. Otherwise they’ll still be able to enable features with recipes , upgrade their privileges with workflows etc

PR #10996, opened 2022-01-09
Improve the feature profiles in the Tenants module

CrestApps:
This PR provide few enhancements to the Tenants modules discussed in #10963

Here are some of the highlights.
1. When using `FeatureProfile` add an `id`  to the profile instead of using an editable name. This way renaming profile won't break anything. To keep it backward compatible, we treat current name as the id for old entries. Note renaming old entries won't cause any issue since we consider the id as the name before the new name is saved.
2. Allow the user to have multiple profiles more on than explained in In a SaaS setup, how can the SaaS admin/owner manage the tenant features? #10962
3. Use `DisplayDriver` to render the CRUD to allow other modules to contribute into the captured data.
More on this is discussed in #10963


CrestApps:
One question I do have, why is `FeatureProfilesValidationProvider` part of `OrchardCore` project instead of `OrchardCore.Tenants`? I would think it should be also registered in the `OrchardCore.Tenants.FeatureProfiles` startup. Should I move it?

deanmarcussen (commented):
There is some good things in this pr.

I don't see the code where the existing name is used to as a swap in for the id. 
Please can you add a comment on the line that is doing that check.

I still don't see a lot of value in that change, as @ns8482e says it is a technical name, and doesn't solve all the potential issues, of deleting a profile etc. As I already commented on the issue I suggest to disable editing of the name, if that is not already done.

The ability to select multi profiles looks like a good change, so it would be good to take that as a standalone pr.

The display driver does not look like a great improvment. The controller view was already overridable, and adding a display driver here, is odd, because the `FeatureProfile` does not act as a base class, or is designed for extending, i.e. it's not a `IEntity`, or sectionable.

What I thought you were talking about doing was adding a display driver for the ShellSettings, which would be something we could potentially do for the tenants view, although it has some potential for issues, due to how they are persisted.

I would rather drop that from this pr.

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -602,13 +603,18 @@ private async Task ValidateViewModel(EditTenantViewModel model, bool newTenant)
...
             {
                 var featureProfiles = await _featureProfilesService.GetFeatureProfilesAsync();
-                if (!featureProfiles.ContainsKey(model.FeatureProfile))
+
+                foreach (var featureProfile in model.FeatureProfiles)
                 {
-                    ModelState.AddModelError(nameof(EditTenantViewModel.FeatureProfile), S["The feature profile does not exist.", model.FeatureProfile]);
+                    if (!featureProfiles.ContainsKey(featureProfile))
+                    {
+                        ModelState.AddModelError(nameof(EditTenantViewModel.FeatureProfiles), S["The feature profile does not exist.", model.FeatureProfiles]);
```
sebastienros, line 614: Did you mean to add `{0}` somewhere in the string?

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/FeatureProfilesController.cs
```diff
@@ -109,173 +114,95 @@ public async Task<IActionResult> Create()
                 return Forbid();
             }
 
-            return View(new FeatureProfileViewModel());
+            var shape = await _displayManager.BuildEditorAsync(new FeatureProfile(), _updateModelAccessor.ModelUpdater, true);
```
sebastienros, line 117: Can't controllers be used as `IUpdateModelAccessor`? I think it just need to have the interface added to it, and it might provide better results (I don't remember the details).

src/OrchardCore/OrchardCore.Abstractions/Shell/Models/FeatureProfile.cs
```diff
@@ -1,11 +1,20 @@
...
     public class FeatureProfile
     {
+        public string Id { get; set; }
+
+        public string Name { get; set; }
+
         public List<FeatureRule> FeatureRules = new List<FeatureRule>();
+
+        // should we inherit from Entity instead?
+        public JObject Properties { get; set; } = new JObject();
```
sebastienros, line 16: Where is it used?

src/OrchardCore.Modules/OrchardCore.Tenants/Drivers/FeatureProfileDisplayDriver.cs
```diff
@@ -0,0 +1,90 @@
...
+using OrchardCore.DisplayManagement.Views;
+using OrchardCore.Environment.Shell.Models;
+using OrchardCore.Mvc.ModelBinding;
+using OrchardCore.Tenants.Models;
+using OrchardCore.Tenants.Services;
+using OrchardCore.Tenants.ViewModels;
+
+namespace OrchardCore.Tenants.Drivers
+{
+    public class FeatureProfileDisplayDriver : DisplayDriver<FeatureProfile>
```
sebastienros, line 17: Shapes instead of Views ... What is the intent? Scenarios where features can be enhanced by other modules?

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
CrestApps: yes. I'll take care of it.

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/FeatureProfilesController.cs
CrestApps: Not sure. Injecting `IUpdateModelAccessor` in the controller's constructor seems to be hows its' done everywhere

src/OrchardCore/OrchardCore.Abstractions/Shell/Models/FeatureProfile.cs
CrestApps: Currently no where. I just thought it should exists to allow consumer to attach any custom parts to it should any be needed just like we do with Entity.

src/OrchardCore.Modules/OrchardCore.Tenants/Drivers/FeatureProfileDisplayDriver.cs
CrestApps: What if a billing module want to attach a price to a profile? Like if one has a subscription based app, one can charge the tenant based on the available profiles.

deanmarcussen:
> @deanmarcussen In regard to the `FeatureProfile`, I added `public JObject Properties { get; set; } = new JObject();` property to it to allow one to extend it by adding custom parts. I think it should inherit from `Entity` but I did not want to add the dependency to `OrchardCore.Infrastructure.Abstractions`. Also, may be we should move all the FeatureProfiles like (`IFeatureProfilesService`, `NullFeatureProfilesService`, `FeatureProfilesService`, `FeatureProfile` and `FeatureRule`) dependencies away from Shell and into `OrchardCore.Tenants.Core` or `OrchardCore.Tenants`

No, please don't do these things. It's a very specific feature only. `FeatureProfile`, at a deep level of the OrchardCore. Extending it to a `TenantProfile` with drivers is not a good plan.

CrestApps:
> > @deanmarcussen In regard to the `FeatureProfile`, I added `public JObject Properties { get; set; } = new JObject();` property to it to allow one to extend it by adding custom parts. I think it should inherit from `Entity` but I did not want to add the dependency to `OrchardCore.Infrastructure.Abstractions`. Also, may be we should move all the FeatureProfiles like (`IFeatureProfilesService`, `NullFeatureProfilesService`, `FeatureProfilesService`, `FeatureProfile` and `FeatureRule`) dependencies away from Shell and into `OrchardCore.Tenants.Core` or `OrchardCore.Tenants`
> 
> No, please don't do these things. It's a very specific feature only. `FeatureProfile`, at a deep level of the OrchardCore. Extending it to a `TenantProfile` with drivers is not a good plan.

@deanmarcussen I made the change.

src/OrchardCore.Modules/OrchardCore.Tenants/Drivers/FeatureProfileDisplayDriver.cs
CrestApps: removed the shapes and used views

src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Edit.cshtml
```diff
@@ -107,7 +107,7 @@
         </h4>
         <div class="mb-3" asp-validation-class-for="FeatureProfiles">
             <label asp-for="FeatureProfiles">@T["Optionally specify which feature profile should be applied to this tenant."]</label>
-            <select asp-for="FeatureProfiles" asp-items="Model.FeatureProfileItems" class="form-control"></select>
+            <select asp-for="FeatureProfiles" asp-items="Model.FeatureProfileItems" class="form-select"></select>
```
hishamco, line 110: I already fixed these in my last PR

CrestApps:
@hishamco I am closing this PR in favor of a much simpler/more recent PR #11841

hishamco:
No need to close the PR you can push your new commits unless something dramatically changes

PR #11002, opened 2022-01-10
Convert Users ControllersExtensions into Services

hishamco:


rjpowers10:
I know these have been extension methods since the beginning but something I've always wondered is, would these be better as services? It always stuck me as a bit odd that they are extension methods instead of service methods, but maybe there was a good reason for that. Since you are making them public, maybe this would be a good opportunity to refactor them into services. Since they were internal before I don't think moving them to services would be a breaking change.

deanmarcussen:
I would prefer them as services, same with the extension method being added on the tenants pr.

Just seems to make more sense to me, as I've fought around them in the past.

changing how email is sent is a common requirement, and it's quite hard to do with those extensions

and you're quite right @rjpowers10 non breaking.

hishamco:
I will try to convert them as service, but naming is hard ;) while they provide different functionality to the controller according the requirement. Hope to send a PR today

hishamco:
@deanmarcussen please let me know if you have any concern about the latest changes, so I can reflected into tenants PR

hishamco:
Any idea why it fails on Ubuntu?

jtkech:
Just saw that `DefaultControllerService` is registered as a transient service
So it would need to be stateless or only inject singletons (but it injects scoped services)
So `DefaultControllerService` would need to be registered as a scoped service.

src/OrchardCore.Modules/OrchardCore.Users/Controllers/ControllerExtensions.cs
```diff
@@ -1,106 +0,0 @@
...
-using OrchardCore.Modules;
-using OrchardCore.Settings;
-using OrchardCore.Users.Events;
-using OrchardCore.Users.Models;
-using OrchardCore.Users.Services;
-using OrchardCore.Users.ViewModels;
-
-namespace OrchardCore.Users.Controllers
-{
-    internal static class ControllerExtensions
```
Skrypt, line 22: Maybe we should just deprecate it first

src/OrchardCore.Modules/OrchardCore.Users/Services/DefaultControllerService.cs
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        /// <summary>
+        /// Returns the created user, otherwise returns null
+        /// </summary>
+        /// <param name="controller"></param>
+        /// <param name="model"></param>
+        /// <param name="confirmationEmailSubject"></param>
+        /// <returns></returns>
+        public virtual async Task<IUser> RegisterUser(Controller controller, RegisterViewModel model, string confirmationEmailSubject)
```
deanmarcussen, line 62: I would suggest making a context class with an `IUrlHelper` rather than passing the controller class in directly.

I think this should all just go into the `IUserService`

@Skrypt the previous was internal, so no need to deprecate it, as it is unusable by anyone

src/OrchardCore.Modules/OrchardCore.Users/Services/DefaultControllerService.cs
```diff
@@ -0,0 +1,133 @@
...
+
+                        return user;
+                    }
+                }
+            }
+
+            return null;
+        }
+
+        public virtual async Task<bool> SendEmailAsync(string email, string subject, IShape model)
```
sebastienros, line 99: Maybe move to an email service because it knows how to send an email based on a shape. Can be useful somewhere else.

src/OrchardCore.Modules/OrchardCore.Users/Services/DefaultControllerService.cs
```diff
@@ -0,0 +1,133 @@
...
+using OrchardCore.Entities;
+using OrchardCore.Modules;
+using OrchardCore.Settings;
+using OrchardCore.Users.Events;
+using OrchardCore.Users.Models;
+using OrchardCore.Users.ViewModels;
+
+namespace OrchardCore.Users.Services
+{
+    public class DefaultControllerService
```
sebastienros, line 21: UserService?

src/OrchardCore.Modules/OrchardCore.Users/Services/DefaultControllerService.cs
hishamco: > I would suggest making a context class with an IUrlHelper rather than passing the controller class in directly.

I started this PR by using `LinkGenerator` instead, but I don't want to lose any data from the controller http context, I will try to use `IUrlHelper`

> @Skrypt the previous was internal, so no need to deprecate it, as it is unusable by anyone

You are right

src/OrchardCore.Modules/OrchardCore.Users/Services/DefaultControllerService.cs
hishamco: I though about that before, so let me organize this PR

src/OrchardCore.Modules/OrchardCore.Users/Services/DefaultControllerService.cs
hishamco: I think we already have `IUserService`, so I don't want confuse the people ;)

src/OrchardCore.Modules/OrchardCore.Users/Controllers/ControllerExtensions.cs
hishamco: It's `internal`

src/OrchardCore/OrchardCore.Email.Core/Extensions/SmtpServiceExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+using OrchardCore.Environment.Shell.Scope;
+
+namespace OrchardCore.Email
+{
+    public static class SmtpServiceExtensions
+    {
+        public static async Task<bool> SendAsync(this ISmtpService smtpService, string email, string subject, IShape model)
+        {
+            var body = String.Empty;
+            var displayHelper = ShellScope.Services.GetService<IDisplayHelper>();
```
hishamco, line 16: @jtkech is this a valid usage or is there a proper way to get such service

hishamco:
Seems I broke th unit tests ;)

hishamco:
BTW can we rename the service `UserAccountService` while it provides a user account functionalities?

rjpowers10:
Just curious: why no interface for the service? I see that there are virtual methods instead.

hishamco:
@rjpowers10 we can introduce as an interface but I was thinking this services for a controller only. Anyhow all the feedback are welcome

hishamco:
Closed in favor of #15360

PR #11034, opened 2022-01-18
Fix localization and sanitization usages

sebastienros:


deanmarcussen:
Looks good, but the original release was still for core3 so only has lang 8.

PR #11045, opened 2022-01-20
jQuery UI 1.13.1

agriffard:
https://www.npmjs.com/package/jquery-ui/v/1.13.1
https://blog.jqueryui.com/2022/01/jquery-ui-1-13-1-released/


Skrypt:
Was it not causing issues?

agriffard:
Admin menu and menu sortable by drag and drop does not work anymore.

Skrypt:
I'm surprised about the admin menu though. It's not supposed to be using JQuery. It's using Bootstrap 4 collapsible. Probably, the issue is that javascript stops executing because of JQueryUI.

agriffard:
This admin menu:
image

Skrypt:
Admin menu management feature.

deanmarcussen:
Yeah I thought we reverted this back down to fix that issue. Did you try it with the new version?

Skrypt:
I think that the issue is that the JQueryUI sortable is not working anymore. We said that we should refactor this by using sortable.js or something like that if I remember correctly. But don't take me to my words, make a double check, these days my brain needs a replacement 😉

PR #11070, opened 2022-01-25
Taxonomy Term Localization

### LINKED ISSUE #4845: Taxonomies Term Localization

Skrypt:
WIP (just a draft to keep the idea)

Previous attemps to do this is here : #6316
Should Fix #4845 

Not sure at all about the implementation. I don't like the fact that I need to override the LocalizationPart.Edit.cshtml file for the route to the Localize action in the Taxonomy "area" to work. But at the same time, each module could be able to override this view to change its controller actions so that each module could implement its own "Localize" method(s).

Here, the fact is that in some modules we have "Contained" content items that need to be also "Localizable" so we can't use the standard ContentLocalization module controller to localize these as the Localize method in that module exists only to localize common content items that are not "Contained".

In the Taxonomy module each Terms are "Contained" content items that can't be retrieved by doing `var contentItem = await _contentManager.GetAsync(contentItemId, VersionOptions.Latest);`. In Fact, these Terms are "Contained" in the TaxonomyPart of their Taxonomy content item.

*Self note*: Here, I need to make sure that if I'm adding ContentLocalization.Abstractions on the Taxonomy module that this "Localize" method be enabled only when the Content Localization feature is enabled.


src/OrchardCore.Modules/OrchardCore.Taxonomies/Views/LocalizationPart.Edit.cshtml
```diff
@@ -0,0 +1,57 @@
...
+                                    <a class="list-group-item list-group-item-action" asp-action="Edit" asp-controller="Admin" asp-route-area="OrchardCore.Contents" asp-route-contentItemId="@culture.ContentItemId" title=@T["Edit {0}", culture.Culture.DisplayName]><i class="fa fa-edit" aria-hidden="true"></i> @culture.Culture.DisplayName | @culture.Culture.Name </a>
+                                }
+                                else
+                                {
+                                    @if (@Context.Request.RouteValues["containerContentItemId"] != null)
+                                    {
+                                        <a class="list-group-item list-group-item-action" data-url-af="UnsafeUrl" asp-action="Localize" asp-controller="Admin" asp-route-area="OrchardCore.Taxonomies" asp-route-containerContentItemId="@Context.Request.RouteValues["containerContentItemId"]" asp-route-contentItemId="@Model.LocalizationPart.ContentItem.ContentItemId" asp-route-targetCulture="@culture.Culture.Name" title=@T["Create {0}", culture.Culture.DisplayName]><i class="fa fa-plus" aria-hidden="true"></i> @culture.Culture.DisplayName | @culture.Culture.Name</a>
```
Skrypt, line 39: Here I'm changing the AreaName so that it triggers the proper controller action. I tried adding a dynamic AreaName based on the current Request previously and it was not resolving the proper Route/Action. But when I'm overriding this shape view inside the Taxonomy module it returns the proper Route/Action.

Skrypt:
@sebastienros Any comments/suggestions? I noticed that you said we should use a CultureAspect but this would be only useful when needing to retrieve proper records on a specific Taxonomy right?

sebastienros:
Won't work because of indexing for instance.

Skrypt:
Drupal definitely has this per term. So, one idea I just got is to associate a CultureAspect on the Taxonomy and to add a specific Dictionary of culture/term per Taxonomy that could be returned by Taxonomy Alias.

PR #11076, opened 2022-01-26
Use new branding icons for OC projects templates

hishamco:
Seems we forgot to update the branding icons for projects templates


hishamco:
FYI there's also other icons in `.template.config` folder but they are not tracking I'm not sure if it's by design or not

Skrypt (approved):

PR #11083, opened 2022-01-27
Edit permission for Title (Lombiq Technologies: OCORE-76)

Psichorex:
Provides a permission that adds the option to allow or disallow users to edit TitleParts.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Title/Drivers/TitlePartDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Title/Drivers/TitlePartDisplayDriver.cs
@@ -55,13 +55,13 @@ namespace OrchardCore.Title.Drivers
                 model.TitlePart = titlePart;
                 model.ContentItem = titlePart.ContentItem;
                 model.Settings = context.TypePartDefinition.GetSettings<TitlePartSettings>();
-                model.IsEditable = await IsEditableOptionProvider(titlePart, context);
+                model.IsEditable = await IsTitleEditableAsync(titlePart, context);
             });
         }
 
         public override async Task<IDisplayResult> UpdateAsync(TitlePart model, IUpdateModel updater, UpdatePartEditorContext context)
         {
-            if (await IsEditableOptionProvider(model, context) && await updater.TryUpdateModelAsync(model, Prefix, t => t.Title))
+            if (await IsTitleEditableAsync(model, context) && await updater.TryUpdateModelAsync(model, Prefix, t => t.Title))
             {
                 var settings = context.TypePartDefinition.GetSettings<TitlePartSettings>();
                 if (settings.Options == TitlePartOptions.EditableRequired && string.IsNullOrWhiteSpace(model.Title))
--- a/src/OrchardCore.Modules/OrchardCore.Title/Drivers/TitlePartDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Title/Drivers/TitlePartDisplayDriver.cs
@@ -45,7 +45,6 @@ namespace OrchardCore.Title.Drivers
             })
             .Location("Detail", "Header:5")
             .Location("Summary", "Header:5");
-
         }
 
         public override IDisplayResult Edit(TitlePart titlePart, BuildPartEditorContext context)
@@ -62,7 +61,7 @@ namespace OrchardCore.Title.Drivers
 
         public override async Task<IDisplayResult> UpdateAsync(TitlePart model, IUpdateModel updater, UpdatePartEditorContext context)
         {
-            if (await updater.TryUpdateModelAsync(model, Prefix, t => t.Title) && await IsEditableOptionProvider(context))
+            if (await IsEditableOptionProvider(context) && await updater.TryUpdateModelAsync(model, Prefix, t => t.Title))
             {
                 var settings = context.TypePartDefinition.GetSettings<TitlePartSettings>();
                 if (settings.Options == TitlePartOptions.EditableRequired && string.IsNullOrWhiteSpace(model.Title))
--- a/src/OrchardCore.Modules/OrchardCore.Title/Drivers/TitlePartDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Title/Drivers/TitlePartDisplayDriver.cs
@@ -1,4 +1,6 @@
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Localization;
 using OrchardCore.ContentManagement.Display.ContentDisplay;
 using OrchardCore.ContentManagement.Display.Models;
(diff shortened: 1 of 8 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=11083) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=11083) before we can accept your contribution.<br/><br/>:x: Psichorex [sign now](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=11083)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OrchardCMS/OrchardCore?pullRequest=11083) it.</sub>

hishamco:
What's the problem you are trying to solve?

Psichorex:
@hishamco 
Hi! It is not really a problem rather a feature. It was a request that editing of the titles shouldn't be available for every users.
Here I tried to implement this permission.

ns8482e:
TitlePart is an element of content item , if content item is allowed to edit why would you need separate permission to manage?

deanmarcussen:
The problem wtih doing this, is then everyone will want a permission for every part, not just the title part, and then where do we stop?

It's also going to be breaking because on existing sites, no one will have been assigned the new permission. So it'll only for for the site owner, who gets a special pass.

Needs some convincing we should add it as a feature, knowing that it could also just be a custom driver / editor :)

Psichorex:
I was a request that we need a solution for Page titles are constantly being changed by other people. @Piedone came up with the idea that we should introduce a permission that is kind of similar to CreateContent permission that will disallow certain users to touch titles. There is a global permission also that will allow the user to edit all the titles just as before, but there is the option to limit one's title editing ability to certain types only or even to none.

ns8482e:
Considering OC is flexible and extensible to custom needs - IMHO this can be easily handled in custom module by introducing custom permission and with help of alternates and placements.

I agree with @deanmarcussen, why permission is just for `TitlePart` ? Later everyone might ask of every possible parts - and why just parts - they may ask to have permission on every field level - and things will get complicated easily and hard to manage in admin ui.

Psichorex:
As Lombiq previous contribution of introducing Create Content permission, if Im right it is merged already we thought that this one could be also a feature of OC. It can be useful although I can see you point of view that it might be unneccessary for some

hishamco:
> As Lombiq previous contribution of introducing Create Content permission,

This is private for Lombiq If I'm not wrong, please check the issue #10234

src/OrchardCore.Modules/OrchardCore.Title/Permissions.cs
```diff
@@ -0,0 +1,30 @@
...
+                    Name = "Administrator",
+                    Permissions = new[] { EditTitlePart },
+                },
+            };
+        }
+
+        public Task<IEnumerable<Permission>> GetPermissionsAsync()
+        {
+            return Task.FromResult(new[] { EditTitlePart }.AsEnumerable());
+        }
```
Luko6, line 28: Switch the order of these

src/OrchardCore.Modules/OrchardCore.Title/Startup.cs
```diff
@@ -1,24 +1,33 @@
...
     public class Startup : StartupBase
     {
+        static Startup() =>
+            ContentTypePermissionsHelper.PermissionTemplates.Add(
+                Permissions.EditTitlePart.Name,
+                Permissions.EditTitlePartTemplate);
+
         public override void ConfigureServices(IServiceCollection services)
-        {
+        {
```
Luko6, line 30: Remove extra spaces

src/OrchardCore.Modules/OrchardCore.Title/Startup.cs
```diff
@@ -31,9 +40,10 @@ public override void ConfigureServices(IServiceCollection services)
                 .AddHandler<TitlePartHandler>();
 
             services.AddScoped<IContentPartIndexHandler, TitlePartIndexHandler>();
-            services.AddScoped<IContentTypePartDefinitionDisplayDriver, TitlePartSettingsDisplayDriver>();
-
+            services.AddScoped<IContentTypePartDefinitionDisplayDriver, TitlePartSettingsDisplayDriver>();
```
Luko6, line 43: Remove extra spaces

src/OrchardCore.Modules/OrchardCore.Title/Drivers/TitlePartDisplayDriver.cs
```diff
@@ -67,5 +76,14 @@ public override async Task<IDisplayResult> UpdateAsync(TitlePart model, IUpdateM
...
             return Edit(model, context);
         }
+
+        private async Task<bool> IsEditableOptionProvider(TitlePart model, BuildPartEditorContext context)
+        {
+            return await _authorizationService.AuthorizeAsync(
+                _httpContextAccessor.HttpContext.User,
+                Permissions.EditTitlePart,
+                model.ContentItem) ||
+                context.IsNew;
```
Luko6, line 86: Use `context.IsNew` first

Luko6 (changes requested):

src/OrchardCore.Modules/OrchardCore.Title/Drivers/TitlePartDisplayDriver.cs
```diff
@@ -67,5 +76,13 @@ public override async Task<IDisplayResult> UpdateAsync(TitlePart model, IUpdateM
 
             return Edit(model, context);
         }
+
+        private async Task<bool> IsEditableOptionProvider(TitlePart model, BuildPartEditorContext context)
```
barthamark, line 80: Add `Async` suffix to the method name, please. By the way, I don't quite understand the method name itself, wouldn't `IsTitleEditableAsync` be more descriptive?

barthamark (changes requested):

src/OrchardCore.Modules/OrchardCore.Title/Drivers/TitlePartDisplayDriver.cs
Psichorex: committed

Piedone:
Just as the title being editable can be configured from a setting, the idea is to add a corresponding permission too. You can of course add this functionality from a third-party module (though it's a lot cleaner from the Title module) but I figured this would be a good addition to Orchard. Title is already quite a special part (again, disabling its editability is a special thing too) so I doubt people would ask for the same thing for every part; if they do, then, well, there's demand for it so perhaps we should consider them? (Still only talking about edit permissions, not full-fledged field-level permissions.)

That being said, while I'm sure we could have some good debate here, getting the feature out of the door is time-sensitive so let's leave this as it is.

hishamco:
@Piedone what If we create a dynamic permission that allow devs to edit a XXX part or not, this way anyone want to add this feature should attach the part to the permission and do the stuff to disable the fields associated

Piedone:
That would be cool though it's an all or nothing solution then, with each part having to take part in it by checking the permissions. I wouldn't go that far.

hishamco:
I may add this to OCC (or Lombiq), if there's no plan to add it in OC

ns8482e:
The reason not have this in `TitlePart` setting is  - Permission is really comprehensive at content item level and is driven by content `Securable` -  that shows that content item in Role's permission list configurable by admin.

BenedekFarkas:
Can someone reopen this, please? I accidentally deleted the branch from our fork.

Piedone:
It can't be:
image

@Psichorex please push the branch again.

BenedekFarkas:
I missed to mention that I pushed that branch, so it's in our fork once again.

PR #11124, opened 2022-02-04
Make SendOnlineMessage() protected

hishamco:
I created few mail providers, I find myself I implement `ISmtpService` everytime, it would be nice if we can extend the `SmtpService`. The issue I run with is each email provider may has some customization that need to be done at `SmtpClient` level. So making `SendOnlineMessage()` `protected virtual` will make such thing easier.

A side not we could rethink about the method name while it will become visible, I suggest to change it to `SendOnlineMessageAsync()`


Skrypt:
You meant email not mail I assume 😉

hishamco:
When I worked with providers I refer to them to Mail Providers ;)

src/OrchardCore/OrchardCore.Email.Core/Services/SmtpService.cs
```diff
@@ -210,7 +210,8 @@ private bool CertificateValidationCallback(object sender, X509Certificate certif
 
             return false;
         }
-        private async Task<string> SendOnlineMessage(MimeMessage message)
+
+        protected virtual async Task<string> SendOnlineMessage(MimeMessage message)
```
deanmarcussen, line 214:
```suggestion
        protected virtual async Task<string> SendOnlineMessageAsync(MimeMessage message)
```

hishamco: I thought you will suggest that dean ;) I already suggested in the description above

Skrypt (approved):

sebastienros:
Why not introduce another method instead that is called right before the client message is sent which could update it? This way you don't even have to copy this method either in your sub class.

hishamco:
Seb I revised the method even I go with the approach you propsed again, there might be situation where the caller need to using logging or any other stuff, so the better approach is to make this `protected`

I have another option, we can make `SendOnlineMessageAsync` protected, also we could add another method to configure the `SmtpClient` this way the caller has two options either override the configure `SmptClient` in case the changes are easy or override the who send method

hishamco:
Is there anything I need to add here?

sebastienros:
I meant add an `OnMessageSendingAsync` that would be called on line 234, and which would be protected. This way inherited classes don't have to copy the full method to customize the message or the client.

hishamco:
Seems I changed my mind after I watched the last stand up ;) 

I'm not sure if it's what you mean or not, so let us the code describe what I mean

hishamco:
@sebastienros can we merge this?

PR #11147, opened 2022-02-07
Add docs for ThemeSelectorResult

hishamco:
This issue here is to avoid the expectation of `ThemeName` many of the dev may expect the theme name is used in `ThemeSelectorResult` while the identifier is used instead. @agriffard if we need add remark on the docs please let me know where should we add it


hishamco:
What's the wrong with https://nodejs.org/dist/v15.14.0?!!

src/OrchardCore/OrchardCore.DisplayManagement/Theming/ThemeSelectorResult.cs
```diff
@@ -1,8 +1,19 @@
...
     {
+        /// <summary>
+        /// Gets or sets the priority for the selected theme.
+        /// </summary>
         public int Priority { get; set; }
+
+        /// <summary>
+        /// Gets or sets the theme name.
+        /// </summary>
+        /// <remarks>Avoid to use the theme name acciedently. <see cref="ThemeSelectorResult"/> is using the theme identifier instead.</remarks>
```
Skrypt, line 16: Avoid using the theme name as a technical value.

Else fix typo on "acciedently" if you keep the same sentence.

src/OrchardCore/OrchardCore.DisplayManagement/Theming/ThemeSelectorResult.cs
```diff
@@ -13,7 +13,7 @@ public class ThemeSelectorResult
         /// <summary>
         /// Gets or sets the theme name.
         /// </summary>
-        /// <remarks>Avoid to use the theme name acciedently. <see cref="ThemeSelectorResult"/> is using the theme identifier instead.</remarks>
+        /// <remarks>Avoid to use the theme name accidentally. <see cref="ThemeSelectorResult"/> is using the theme identifier instead.</remarks>
```
deanmarcussen, line 16: Gramatically this sentence makes no sense
```suggestion
        /// <remarks>This is the theme identifier. By default this is also the theme name, but if a custom identifier has been specified, please use the identifier.</remarks>
```

I would actually put it in the summary, noone will ever see it in remarks.

hishamco: I'll see the experience in VS & propbably VS Code, how the remarks are show up or they aren't appear at all

hishamco: > This is the theme identifier. By default this is also the theme name, but if a custom identifier has been specified, please use the identifier.

The theme name is not working at all?!!

hishamco: We could summarize it with what Jasmin suggested "Avoid using the theme name as a technical value." plus mentioning to use theme identifier instead

ns8482e: This requires change when enabling theme from admin

as it uses theme name and not id

Piedone (approved):

Issue #6939, opened 2020-08-19
Does the query now support returning the total number of rows? 


hyzx86:
If not, I want to try implement this function, but I don’t know how to start .

hyzx86:
https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL/Types/Relay/ConnectionType.cs

sebastienros:
We don't have that, and I am not sure we should provide it. Unless it takes "paging options" into account such that it can't "scan" the full content item table which can be expensive.

carlwoodhouse:
@sebastienros this is a potential duplicate of https://github.com/OrchardCMS/OrchardCore/issues/5693

hyzx86:
> We don't have that, and I am not sure we should provide it. Unless it takes "paging options" into account such that it can't "scan" the full content item table which can be expensive.

Hi @sebastienros  , I've written some code, but I haven't finished yet. I don't think the calculation of the total number of rows here will result in scanning the whole table

https://github.com/hyzx86/OrchardCore/blob/dfdc6d13f0bf1d018fdcf08d3ee1178073c249a4/src/OrchardCore/OrchardCore.ContentManagement.GraphQL/Queries/ContentItemsFieldType.cs#L105

We need to modify the return result of a resolve, which is similar to this:
```
{
 data:{
    totalcount:100 ,
    items:[
               contentItem1,
               contentItem2
             ]
   }
}
```
But the query results refer to many places. I don't know how to find them to modify  result,
Or is there another way to make all `contentItem` queries support the return total, such as creating a new query.

carlwoodhouse:
IMO if we want more powerful paging we should just implement connections;
this is the accepted standard in graphql.

Both the above and connections are fairly large breaking model changes
though; though with connections you could probably support both by just
having separate connection queries.

hyzx86:
I think this is a very important feature, even if it requires some disruptive updates.
I will be working full time on this feature for a while.
At that time, I hope you can provide some guidance and suggestions. Thank you！

hyzx86:
Hi ，There is a problem here. How can I get a `ConnectionType` from this instance object:`typeType `d ? 
https://github.com/hyzx86/OrchardCore/blob/d755c00fa7aaf4f602b80c717d18db46f42ccc31/src/OrchardCore/OrchardCore.ContentManagement.GraphQL/Queries/ContentTypeQuery.cs#L66

[...]

PR #11052, opened 2022-01-22
Elasticsearch feature :sparkles::boom: (Lombiq Technologies: OSOE-83)

Skrypt:
Fixes #4316

### Install Elasticsearch 7.x with Docker compose

Elasticsearch uses a mmapfs directory by default to store its indices. The default operating system limits on mmap counts is likely to be too low, which may result in out of memory exceptions.

https://www.elastic.co/guide/en/elasticsearch/reference/current/vm-max-map-count.html

For Docker with WSL2, you will need to persist this setting by using a .wslconfig file.

In your Windows `%userprofile%` directory (typically `C:\Users\<username>`) create or edit the file `.wslconfig` with the following:

```
[wsl2]
kernelCommandLine = "sysctl.vm.max_map_count=262144"
```

Then exit any WSL instance, `wsl --shutdown`, and restart.

```cmd
> sysctl vm.max_map_count
vm.max_map_count = 262144
```
Elasticsearch v7.17.5 Docker Compose file : 

[elasticsearch.txt](https://github.com/OrchardCMS/OrchardCore/files/9428195/elasticsearch.txt)

- Copy this file in a folder named Docker somewhere safe.
- Rename this file extension as .yml instead of .txt.
- Open up a Terminal or Command Shell in this folder.
- Execute `docker-compose  up` to deploy Elasticsearch containers.

Advice: don't remove this file from its folder if you want to remove all their containers at once later on in Docker desktop.

You should get this result in Docker Desktop app : 

image

### Set up Elasticsearch in Orchard Core

- Add Elastic Connection in the shell configuration (OrchardCore.Cms.Web appsettings.json file)

```JSON
"OrchardCore_Elastic": {
    "ConnectionType": "SingleNodeConnectionPool",
    "Url": "http://localhost",
    "Ports": "9200"
}
```

- Start an Orchard Core instance with VS Code debugger
- Go to Orchard Core features, Enable Elasticsearch.

## Implementation details

**_Analyzed and Stored Properties are not very meaningful in context of Elasticsearch._**

### Analyzed 

Analyzed is default for strings in Elasticsearch.
Because of automatic mapping, by default, all string fields are indexed twice in Elasticsearch as a "Text" Field and a "Keyword" field.

So we will have a field called ContentItemId(Text) analyzed and another called ContentItemId.Keyword to match on exact values using TermQuery for fields like ContentItemId or emails (Elastic Stores text fields in 2 fields analyzed vs not analyzed, a field ContentItemId.Keyword is created automatically)

[...]

PR #11159, opened 2022-02-11
Enable Graphql queries to support total when using Lucene queries

hyzx86:
Change the query schema to the following: add 'hasTotal' :true

```
{
  "type":"contentItem/TableProducts",
 "hasTotal":true
}
```

image

https://github.com/OrchardCMS/OrchardCore/issues/6939


hyzx86:
Regarding the `"hasTotal":true`  option, I want it to be the default
I notice this PR is in the works, maybe it can be released with this feature?
https://github.com/OrchardCMS/OrchardCore/pull/11052

aminmohammadi05:
What if when there are multiple content types? 
```
[{
  "type":"contentItem/TableProducts",
 "hasTotal":true
},
{
  "type":"contentItem/Type2",
 "hasTotal":true
}]
```

hyzx86:
@aminmohammadi05  have you try  graphql ?

hishamco (changes requested):
This is a quick review, please check the format in all the places

aminmohammadi05:
> @aminmohammadi05 have you try graphql ?
Yes but it does not shown in the graphql

hyzx86:
> > @aminmohammadi05 have you try graphql ?
> > Yes but it does not shown in the graphql

cuz , you haven't  input the schema field

aminmohammadi05:
> > > @aminmohammadi05 have you try graphql ?
> > > Yes but it does not shown in the graphql
> 
> cuz , you haven't input the schema field

How to do that?

hyzx86:
> > > > @aminmohammadi05 have you try graphql ?
> > > > Yes but it does not shown in the graphql
> > 
> > 
> > cuz , you haven't input the schema field
> 
> How to do that?

image

aminmohammadi05:
> > > > > @aminmohammadi05 have you try graphql ?
> > > > > Yes but it does not shown in the graphql
> > > 
> > > 
> > > cuz , you haven't input the schema field
> > 
> > 
> > How to do that?
> 
> image
I mean when I put type Product in type field it shows just Product type field in Graphiql but I need it to show Blog type field also
image
image

hyzx86:
image

aminmohammadi05:
> image

This is a Lucene search query which I need to provide parameters for it to search the parameters in all the selected types: Product, Blog

hyzx86:
image

aminmohammadi05:
> image

I know, I want to set the schema field of the Lucene query in a way that the "searchAll" query shows fields of the both Product and Blog types

hyzx86:
https://github.com/OrchardCMS/OrchardCore/pull/11052

I noticed that this function was merged into the main branch yesterday

hyzx86:
> I know, I want to set the schema field of the Lucene query in a way that the "searchAll" query shows fields of the both Product and Blog types

Lucene query should not do what you said. It stores some types in a table containing many fields
If you want to query two types at the same time, you may need to perform secondary filtering on the client

aminmohammadi05:
> > I know, I want to set the schema field of the Lucene query in a way that the "searchAll" query shows fields of the both Product and Blog types
> 
> Lucene query should not do what you said. It stores some types in a table containing many fields If you want to query two types at the same time, you may need to perform secondary filtering on the client

No problem how can I query two types at the same time. As you said I can handle separating the result based on its type.

hyzx86:
> No problem, how can I query two types at the same time. As you said I can handle separating the result based on its type.


have you try  object type?

https://docs.orchardcore.net/en/latest/docs/reference/modules/Queries/#graphql

image

aminmohammadi05:
> > No problem, how can I query two types at the same time. As you said I can handle separating the result based on its type.
> 
> have you try object type?
> 
> https://docs.orchardcore.net/en/latest/docs/reference/modules/Queries/#graphql
> 
> image

Yeah I saw that, I wonder how to query two types at the same time. As you see the document only shows one type: "BlogPost"

hyzx86:
Not ready , I'm trying to implement more flexible schema configuration, such as

``` json
{
 {"type":"Array","ofType":"Blog","name":"items"},
 {"type":"Integer","name":"total"},
}
```

aminmohammadi05:
> Change the query schema to the following: add 'hasTotal' :true
> 
> ```
> {
>   "type":"contentItem/TableProducts",
>  "hasTotal":true
> }
> ```
> 
> image
> 
> #6939

I tried this one to return total records but it did not work. What else should I do to get the result?

hyzx86:
@aminmohammadi05 
> I tried this one to return total records but it did not work. What else should I do to get the result?

I have updated the code. Would you like to try again?

image

Schema:

```
{
    "type":"contentitem/AmisSchema",
    "hasTotal":true
}
```
Query
``` Liquid
{
    "query": {
        "bool": {
            "must": [
                { "term": { "Content.ContentItem.ContentType": "AmisSchema" } },
                { "term": { "Content.ContentItem.Published": 'true' } },
                { "term": { "Content.ContentItem.Latest": "true" } }
            ],
            {% if filters %}
            "filter":[
                  {% for filter in filters %} 
              		{%if filter.key %}
      					{"{{filter.method}}": { "{{filter.key}}": "{{filter.value}}" }},
                    {%else%}
                    	{"{{filter.method}}":"{{filter.value}}"}},
                    {%endif%}
				  {% endfor %} 
            	]
            {% endif %}
        }
    }
  {% if from %}
    ,"from":{{from}}
  {% endif %}
  {%if size %}
   ,"size":{{size}}
  {%endif%}  	
}
```

graphql
``` graphql
query MyQuery {
  defaultPagingQuery(parameters: "{'from':0,'size':2}") {
    items {
      displayText
      modifiedUtc
      publishedUtc
      createdUtc
      render
    }
    total
  }
}

```

jtkech:
@hyzx86 There is the following warning to fix in the GitHub Actions / Build & Test ;)

'LuceneQueryResults' is obsolete: 'This class has been deprecated and will be removed in the next major release, please use OrchardCore.Search.Lucene.Abstractions instead.'

hyzx86:
@hyzx86  Yes , I noticed that ,And I haven‘t received such a message  since I update the code

aminmohammadi05:
What if I don't want to update my orchard package? I mean what approach did you use before this modification to return total number of records?

hyzx86:
> What if I don't want to update my orchard package? I mean what approach did you use before this modification to return total number of records?

same as above ， Have you pull my PR before yesterday？

aminmohammadi05:
> > What if I don't want to update my orchard package? I mean what approach did you use before this modification to return total number of records?
> 
> same as above ， Have you pull my PR before yesterday？

I am using version 1.4 of OrchardCore.Application.Cms.Core.Targets in .NET 6 project but this approach didn't work

hyzx86:
> > > What if I don't want to update my orchard package? I mean what approach did you use before this modification to return total number of records?
> > 
> > 
> > same as above ， Have you pull my PR before yesterday？
> 
> I am using version 1.4 of OrchardCore.Application.Cms.Core.Targets in .NET 6 project but this approach didn't work

Please provide more information, such as error logs, Schema and Query

aminmohammadi05:
image
 this is the schema

aminmohammadi05:
image
 And the GraphQl output

hyzx86:
Looks like the code is not work ,Do you use my branch directly?

aminmohammadi05:
No I downloaded it via Nuget package manager in Visual Studio 2022 3 months ago

hyzx86:
> No I downloaded it via Nuget package manager in Visual Studio 2022 3 months ago

This feature only work on my branch...

aminmohammadi05:
So how can I get total records without using you branch? cuz I am in the middle of development and feel it risky to change the Orchard

hyzx86:
> So how can I get total records without using you branch? cuz I am in the middle of development and feel it risky to change the Orchard


You can directly use the WebAPI : `api/queries/{name}`

https://docs.orchardcore.net/en/latest/docs/reference/modules/Queries/#breaking-changes

aminmohammadi05:
Thanks a lot.

aminmohammadi05:
Would you please tell me except bearer token which I provide in postman where should I put the query parameter which in my case is 
```
{
    "from" : 0,
    "size" : 6,
    "fulltext" : "'sf6'"
}
```

hyzx86:
@hishamco 

I thought, is it possible to change the UI of LuceneQuery to add a checkbox, which can togdle the total field 

What do you think?

Piedone:
I think this would be useful, but only if you implement parity with the Elasticsearch queries too. @Skrypt could you please chime in here?

sebastienros:
Why not always have a `count` property and if the provider supports it return the value or `-1` if it's not available. `0` would mean no items when the query supports it.

hyzx86:
Of course that would be best, but that would introduce a disruptive change. 
Because the current query api returns a collection type

hyzx86:
In addition, we have discussed quantity statistics implemented on SQL at this level
https://github.com/OrchardCMS/OrchardCore/issues/6939

PR #11172, opened 2022-02-12
Introduce OC.XmlRpc.Core

hishamco:
This was in my Git Stashes for awhile, it's a time to push it


Piedone (commented):
While I don't see any issues with the code but please bring it up on the Thursday meeting whether there's something that would make this not viable.

sebastienros:
File an issue, we'll discuss it, we'll agree that it doesn't help anybody to reference a project to use these services outside of a module, and we'll close the PR ;)

PR #11283, opened 2022-02-28
Fix NRE with TaxonomyIndex

Skrypt:
When a Taxonomy Field is removed from a Content Type. Cloning the content item may fail with an NRE.

/cc @ns8482e

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Taxonomies/Indexing/TaxonomyIndex.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Taxonomies/Indexing/TaxonomyIndex.cs
@@ -87,7 +87,8 @@ namespace OrchardCore.Taxonomies.Indexing
                             continue;
                         }
 
-                        var jField = (JObject)jPart[fieldDefinition.Name] ?? null;
+                        var jField = jPart[fieldDefinition.Name] ?? null;
+                        jField = (JObject)jField;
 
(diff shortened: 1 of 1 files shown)


ns8482e (approved):
LGTM

netwavebe:
Hi @Skrypt, turns out it has nothing to do with the removed taxonomy field.

The index runs while cloning, right after an empty content item was created (so before the.actual cloning). The fields are there, but all ´null´. Casting this to JObject fails, as null is a JValue.

I locally fixed this by changing
```
var jField = (JObject)jPart[fieldDefinition.Name];
```
to
```
var jField = jPart[fieldDefinition.Name] as JObject;
```

Skrypt:
Yes, I think this would work still with less lines.

netwavebe (approved):
All good!

mwpowellhtx:
My two cents, does not look as though it is an `as` issue so much as potentially `fieldDefinition` itself could be `null`? That or perhaps an indexing issue, depending on the part indexer implementation handles keys, lookups, etc. i.e. mentioned removed taxonomy, i.e. `fieldDefinition` which to me says `fieldDefinition?.Name` is more appropriate? Again, my two cents. Good hunting!

Skrypt:
We are looping inside a collection of fieldDefinition so it is impossible that the object be null else it would not be part of the IEnumerable.

PR #11287, opened 2022-03-02
Make AutoroutePart handlers execute last

### LINKED ISSUE #11285: Execution order of AutoroutePartHandler vs TitlePartHandler
OC 1.2.2

### Describe the bug
I have an autoroute pattern that is based on the content item's DisplayText, `foo/{{ ContentItem.DisplayText | slugify }}`. I'm also trying to programmatically set the TitlePart (and DisplayText) in a custom TitlePartHandler. The problem is there doesn't seem to be a good way to control the ordering of those handlers, so the AutoroutePartHandler runs before my CustomTitlePartHandler and therefore the DisplayText is not yet set.

### To Reproduce
1. Create a new OC site using the blog recipe.
2. Set breakpoints in AutoroutePartHandler.UpdatedAsync and TitlePartHandler.UpdatedAsync
3. Create a new "page" content item
4. Save the content item and see that the AutoroutePartHandler breakpoint is hit before the TitlePartHandler breakpoint.

### Expected behavior
I think it's common for the Autoroute pattern to incorporate the DisplayText, and that works fine as long as you provide a title before the AutoroutePartHandler runs (for example, the user simply providing the title text on the edit page). But I think there is a hole when the TitlePart is being generated in a content part handler, like my own CustomTitlePartHandler or the built-in OC TitlePartHandler.

### Screenshots
Looking at ContentPartHandlerCoordinator, there doesn't seem to be any guarantee to the order in which the content parts are handled. In my case, the AutoroutePart is going to be handled before the TitlePart.
image

Here's the built-in OC functionality to set the TitlePart based on a pattern. But if my Autoroute pattern depends on the DisplayText, I think there is a hole here if it's not guarnateed for the TitlePartHandler to run first.
image

### Workaround
I can probably get around this by using a different event (for example, UpdatingAsync, CreatedAsync, etc.). But TitlePartHandler and AutorouteHandler use UpdatedAsync, so that's what I went with for my CustomTitlePartHandler.

Skrypt:
Fixes #11285 

Needs to be tested. Just opened a draft PR to remember to do it.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Autoroute/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Autoroute/Startup.cs
@@ -34,7 +34,7 @@ namespace OrchardCore.Autoroute
     public class Startup : StartupBase
     {
         public override int ConfigureOrder => -100;
-
+        public override int Order => 999999;
         public override void ConfigureServices(IServiceCollection services)
         {
             services.Configure<TemplateOptions>(o =>


rjpowers10:
I don't think setting the Order on the Startup is going to fluence the order the handlers execute in _across different content parts_.

I tried setting the Order to 999999 and redid my test from #11285 

My title pattern
image

My autoroute pattern
image

Leave everything blank and publish
image

After the first publish, the title part is populated according to the pattern.
image

After the second publish, the autoroute part is populated according to the pattern.
image

In ContentPartHandlerCoordinator
image

One (maybe bad) idea would be to add a special case directly in ContentPartHandlerCoordinator, like so

```
foreach (var typePartDefinition in contentTypeDefinition.Parts
    .OrderBy(p => p.PartDefinition.Name == "AutoroutePart" ? 1 : 0))
```

Of course, the TitlePart _also_ allows for a Liquid pattern, so it might be a good idea to delay that one, too.

```
foreach (var typePartDefinition in contentTypeDefinition.Parts
    .OrderBy(p => p.PartDefinition.Name == "AutoroutePart" ? 2 : p.PartDefinition.Name == "TitlePart" ? 1 : 0))
```

### Concerns

1. Changing the ordering could have unintended consequences for existing sites.
2. Technically, the TitlePart _could_ reference the AutoroutePart in its pattern, so this ordering wouldn't help that. But it's more likely that the AutoroutePart pattern will reference the TitlePart pattern, so this is probably the right ordering.

Skrypt:
We need to try with Manifest dependencies then.

Skrypt:
I just tested this and I don't get the same result concerning the order.

image

jtkech:
Okay just tried it with the Article type of the Blog theme, so for info

@rjpowers10 Yes, I can repro your test, then yes setting an high Order to Autoroute changes the order of the module startup registration. So yes, for the same given part the execution order of its handlers will depend on the above

But but but, for 2 different parts as Title and Autoroute, the order of handler executions depends on the order of parts in the ContentType Definition, so here if it doesn't work setting an higher Order will not fix it.

If you look at the blog theme recipe for the Article ContentType Definition, you will see that the AutoroutePart is defined before the TitlePart. For testing, in the Article type definition (in the Admin UI), just to change the part ordering, I removed the AutoroutePart and then re-added it. Then I redo the same test and it was working.

Skrypt:
ohh my god. 🙊

ns8482e:
@Skrypt this is same ordering @CrestApps was asking on gitter  for drivers- The order of handler/driver is effective only if it's either on contents or within same part not between parts.

Here the issue is `TitlePart` and `AutoroutePart` are really not targeting their Part's properties but instead it's targeting ContentItem properties so  It would make logical sense to inherit both from `ContentsHandlerBase` 

Here if `AutoRouteHandler` inherits from `ContentsHandlerBase` then it will honor the dependency order and run after all part's handler executed.

Skrypt:
Yes, but we might break everyones handlers order by doing so. That why I'm 🙊

sebastienros:
It is too risky to change it this way. So many places could be impacted, like cloning, localization, import/export. Even if this hack is fixing an issue, what will be done when we discover something else is broken? Revert? Add another set of hacks?

This can't be a solution unless all other features have been verified to work. But not even checking the issue it's fixing, I don't think it's a good idea. By experience this kind of hack has always been hard to deal with afterwards.

Skrypt:
At first, I did not remember that the Admin UI was affecting this order. Then the proper fix would be to be able to set the handlers to order separately from the UI order. Because UI order should only affect UI. Then, this will also break existing implementations from people relying on that UI order. Though, I could want to display the AutoroutePart UI first and still want the handler to execute last.

PR #11336, opened 2022-03-08
Added `where` to GraphQL query

devlife:
I noticed that the GraphQL query in the docs doesn't actually work.
The screenshot is the error I get when I use the `contentItemId` parameter.

image

And here is a successful query using `where`:

image


devlife:
@sebastienros I added a screenshot of the error.

sebastienros:
I trusted you ;) I am just surprised it was broken, or that it might have never worked.

sebastienros:
My opinion is we might want to understand if it was broken, because it sounds like it used to work. So if it worked it should be fixed, and not just update the documentation.

src/docs/reference/core/Apis.GraphQL.Abstractions/README.md
```diff
@@ -16,7 +16,7 @@ In this example, the `blog` is the type, and the `displayText` is the return val
 
 ```json
 {
-  blog(contentItemId: "4k5df0kadp9asy1n2ejzs1rz4r") {
+  blog(where {contentItemId: "4k5df0kadp9asy1n2ejzs1rz4r"}) {
```
jptissot, line 19:
```suggestion
  blog(where: {contentItemId: "4k5df0kadp9asy1n2ejzs1rz4r"}) {
```

PR #11366, opened 2022-03-12
Code Mirror SRI

agriffard:
codemirror.css and codemirror.js SRIs changed, compared to those I generated for the last PR:

https://github.com/OrchardCMS/OrchardCore/pull/11247/files#diff-596f484f166194b186319679b8d4d9ffbeb229b566217781278ba98d0d4f9f6b

.css are back to the previous one and .js to another one.

I don't really understand what has happened. 
It could have been my mistake but why 2 times and with different causes?

I figured it out by testing this page: /Admin/Settings/userRegistration, whose `Script Reset` button does not work BTW.


Skrypt:
I found another one

image

PR #11382, opened 2022-03-14
Fix OpenId prune operations

jersiovic:
Current prune operations of authorizations and tokens used at openid background tasks have a bug.
They do a loop with an offset for query the database that is incremented. But in fact it is a mistake, cause on each iteration those records are removed from the db, so on the query of the next iteration it has no sense to use an offset for the query.

By the way I've harnessed the fix to reduce que pagination to only 100 records per query, cause I've experienced problems with saving chagnes of 1000 records.

Finally, I've added code to finish the loop if the query doesn't return records. I didn't find any reason to continue looping in that case.


sebastienros:
I don't know why it's limiting to 100000 items, but I assume it's in case there are two many, then one can invoke it multiple times instead of blocking the server.

sebastienros (approved):

PR #11399, opened 2022-03-17
Fixes #11360 Liquid IHtmlContent

### LINKED ISSUE #11360: Liquid can't seem to handle IHtmlContent
### Describe the bug

I've tried to use the INotifier interface, where you're supposed to enter a LocalizedHtmlString, which derives from IHtmlContent. Then I tried to customize the message presentation by creating a Message.liquid. Alas, try as I might to convince the {{ Model.Message }} to actually print the message, it was all for naught.

(And of course, it's annoying to do this in Razor because IHtmlContent doesn't have a way to simply return the string so that I can javascript-escape it. And I can't easily just cast it, since it becomes a HtmlString instead of a LocalizedHtmlString if it's been serialized...)

### To Reproduce
1. Create a Message.liquid in your theme, containing "{{ Model.Message }}" and "{{ Model.Message | json }}"
2. Do something that creates a notification. I used custom code, but some workflow could probably do it. (Not sure if it would behave differently with the HtmlString, I've only fought this problem with the LocalizedHtmlString.)
3. The json version should show that there's a Value, but {{ Model.Message }} only prints "Microsoft.AspNetCore.Mvc.Localization.LocalizedHtmlString" (and {{ Model.Message.Value }} prints nothing, it seems)

### Expected behavior
I should be able to print a IHtmlContent-derived object from Liquid (especially you can apparently generate such objects with Liquid).

jtkech:
Fixes #11360

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/OrchardCoreBuilderExtensions.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/OrchardCoreBuilderExtensions.cs
@@ -57,10 +57,11 @@ namespace Microsoft.Extensions.DependencyInjection
                         {
                             return new ObjectValue(s);
                         }
-                        if (!(o is IShape) && o is IHtmlContent c)
+                        else if (x is IHtmlContent c)
                         {
                             return new HtmlContentValue(c);
                         }
+
                         return null;
                     });
 


hishamco (approved):

Skrypt:
Seems like an appropriate fix. Have you tested the result to assert that it fixes the issue though? Just making sure. Else, lgtm.

jtkech:
> Have you tested the result to assert that it fixes the issue though?

Yes by using a Message.liquid view, first I could repro then checked that it fixes the issue

sebastienros (approved):

PR #11425, opened 2022-03-22
Fixing html error in the admin UI (Lombiq Technologies: OCORE-87)

BanzragchUchral:
I was writing a UI test for one of our Lombiq modules, and I came across some UI errors/warnings in the admin page.

I am attaching the validation report, and here are the rules which got validated:

- https://html-validate.org/rules/prefer-native-element.html
- https://html-validate.org/rules/no-dup-id.html
- https://html-validate.org/rules/text-content.html

[HtmlValidationReport.txt](https://github.com/OrchardCMS/OrchardCore/files/8327963/HtmlValidationReport.txt)

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.AuditTrail/Navigation/AuditTrailSettingsAdminMenu.cs
+++ b/src/OrchardCore.Modules/OrchardCore.AuditTrail/Navigation/AuditTrailSettingsAdminMenu.cs
@@ -26,7 +26,7 @@ namespace OrchardCore.AuditTrail.Navigation
                  .Add(S["Configuration"], configuration => configuration
                      .Add(S["Settings"], settings => settings
                         .Add(S["Audit Trail"], S["Audit Trail"].PrefixPosition(), settings => settings
-                            .AddClass("audittrail").Id("audittrail")
+                            .AddClass("audittrail").Id("audittrailSettings")
                             .Action("Index", "Admin", new { area = "OrchardCore.Settings", groupId = AuditTrailSettingsGroup.Id })
                             .Permission(AuditTrailPermissions.ManageAuditTrailSettings)
                             .LocalNav())));
--- a/src/OrchardCore.Modules/OrchardCore.Users/Views/UserMenu.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Views/UserMenu.cshtml
@@ -1,9 +1,9 @@
 @inject IAuthorizationService AuthorizationService
 <div class="dropdown">
     <form asp-route-area="OrchardCore.Users" asp-controller="Account" asp-action="LogOff" method="post" class="form-inline" class="no-multisubmit">
-        <a class="ml-3 dropdown-toggle" id="navbarDropdown" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" data-toggle-tooltip="tooltip" title="@User.Identity.Name">
+        <button class="ml-3 dropdown-toggle" id="navbarDropdown" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" data-toggle-tooltip="tooltip" title="@User.Identity.Name">
             <i class="fa fa-fw fa-user" aria-hidden="true"></i>
-        </a>
+        </button>
         <div class="dropdown-menu dropdown-menu-right" aria-labelledby="navbarDropdown">
             <h6 class="dropdown-header">@T["Signed in as"]<span class="ml-1 font-weight-bold">@User.Identity.Name</span></h6>
             <div class="dropdown-divider"></div>
--- a/src/OrchardCore.Modules/OrchardCore.Users/Views/UserMenu.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Views/UserMenu.cshtml
@@ -1,7 +1,7 @@
 @inject IAuthorizationService AuthorizationService
 <div class="dropdown">
     <form asp-route-area="OrchardCore.Users" asp-controller="Account" asp-action="LogOff" method="post" class="form-inline" class="no-multisubmit">
-        <a class="ml-3 dropdown-toggle" id="navbarDropdown" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" role="button" data-toggle-tooltip="tooltip" title="@User.Identity.Name">
+        <a class="ml-3 dropdown-toggle" id="navbarDropdown" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" data-toggle-tooltip="tooltip" title="@User.Identity.Name">
             <i class="fa fa-fw fa-user" aria-hidden="true"></i>
         </a>
         <div class="dropdown-menu dropdown-menu-right" aria-labelledby="navbarDropdown">
--- a/src/OrchardCore.Modules/OrchardCore.Users/Views/UserMenu.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Views/UserMenu.cshtml
@@ -1,7 +1,7 @@
 @inject IAuthorizationService AuthorizationService
 <div class="dropdown">
-    <form asp-route-area="OrchardCore.Users" asp-controller="Account" asp-action="LogOff" method="post" class="form-inline" class="no-multisubmit">
-        <button class="ml-3 dropdown-toggle" id="navbarDropdown" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" data-toggle-tooltip="tooltip" title="@User.Identity.Name">
+    <form asp-route-area="OrchardCore.Users" asp-controller="Account" asp-action="LogOff" method="post" class="form-inline no-multisubmit">
+        <button type="button" class="ml-3 dropdown-toggle" id="navbarDropdown" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" data-toggle-tooltip="tooltip" title="@User.Identity.Name">
             <i class="fa fa-fw fa-user" aria-hidden="true"></i>
             <span>@T[User.Identity.Name]</span>
         </button>
@@ -10,9 +10,9 @@
             <div class="dropdown-divider"></div>
             @if (await AuthorizationService.AuthorizeAsync(User, Permissions.ManageOwnUserInformation))
             {
-                <a class="dropdown-item" asp-area="OrchardCore.Users" asp-action="Edit" asp-controller="Admin" asp-route-returnUrl="@FullRequestPath"><span><i class="far fa-address-card" aria-hidden="true"></i></span>&nbsp;<span> @T["Profile"]</span></a>
+                <a class="dropdown-item" asp-area="OrchardCore.Users" asp-action="Edit" asp-controller="Admin" asp-route-returnurl="@FullRequestPath"><span><i class="far fa-address-card" aria-hidden="true"></i></span>&nbsp;<span> @T["Profile"]</span></a>
             }
-            <a class="dropdown-item" asp-area="OrchardCore.Users" asp-controller="Account" asp-action="ChangePassword" asp-route-returnUrl="@FullRequestPath"><span><i class="fas fa-key" aria-hidden="true"></i></span>&nbsp;<span> @T["Change password"]</span></a>
+            <a class="dropdown-item" asp-area="OrchardCore.Users" asp-controller="Account" asp-action="ChangePassword" asp-route-returnurl="@FullRequestPath"><span><i class="fas fa-key" aria-hidden="true"></i></span>&nbsp;<span> @T["Change password"]</span></a>
             <button type="submit" class="dropdown-item"><span><i class="fa fa-sign-out-alt" aria-hidden="true"></i></span>&nbsp;<span> @T["Log off"]</span></button>
         </div>
     </form>
--- a/src/OrchardCore.Modules/OrchardCore.Users/Views/UserMenu.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Views/UserMenu.cshtml
@@ -3,6 +3,7 @@
     <form asp-route-area="OrchardCore.Users" asp-controller="Account" asp-action="LogOff" method="post" class="form-inline" class="no-multisubmit">
         <button class="ml-3 dropdown-toggle" id="navbarDropdown" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" data-toggle-tooltip="tooltip" title="@User.Identity.Name">
             <i class="fa fa-fw fa-user" aria-hidden="true"></i>
+            <span>@T[User.Identity.Name]</span>
         </button>
         <div class="dropdown-menu dropdown-menu-right" aria-labelledby="navbarDropdown">
             <h6 class="dropdown-header">@T["Signed in as"]<span class="ml-1 font-weight-bold">@User.Identity.Name</span></h6>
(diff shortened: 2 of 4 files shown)


src/OrchardCore.Modules/OrchardCore.AuditTrail/Navigation/AuditTrailSettingsAdminMenu.cs
```diff
@@ -26,7 +26,7 @@ public Task BuildNavigationAsync(string name, NavigationBuilder builder)
                  .Add(S["Configuration"], configuration => configuration
                      .Add(S["Settings"], settings => settings
                         .Add(S["Audit Trail"], S["Audit Trail"].PrefixPosition(), settings => settings
-                            .AddClass("audittrail").Id("audittrail")
+                            .AddClass("audittrail").Id("audittrailSettings")
```
Piedone, line 29: Where is the other "audittrail" ID coming from?

src/OrchardCore.Modules/OrchardCore.Users/Views/UserMenu.cshtml
```diff
@@ -1,7 +1,7 @@
 @inject IAuthorizationService AuthorizationService
 <div class="dropdown">
     <form asp-route-area="OrchardCore.Users" asp-controller="Account" asp-action="LogOff" method="post" class="form-inline" class="no-multisubmit">
-        <a class="ml-3 dropdown-toggle" id="navbarDropdown" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" role="button" data-toggle-tooltip="tooltip" title="@User.Identity.Name">
+        <a class="ml-3 dropdown-toggle" id="navbarDropdown" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" data-toggle-tooltip="tooltip" title="@User.Identity.Name">
```
Piedone, line 4: Why not use a `<button>` instead? Also in src/OrchardCore.Themes/TheAdmin/Views/Layout.cshtml.

Also see: https://stackoverflow.com/questions/57398579/when-and-when-not-to-use-role-button

src/OrchardCore.Themes/TheAdmin/Views/Layout.cshtml
```diff
@@ -65,7 +65,7 @@
         <div id="ta-left-sidebar" class="d-flex flex-column justify-content-between align-items-stretch">
             @await RenderSectionAsync("Navigation", required: false)
             <div class="footer">
-                <button class="leftbar-compactor" type="button" title="@T["Collapse / expand menu"]"></button>
+                <a class="leftbar-compactor" title="@T["Collapse / expand menu"]"></a>
```
Piedone, line 68: This was fine as a `<button>`, just it shouldn't have the redundant `role` but have text.

src/OrchardCore.Themes/TheAdmin/wwwroot/css/TheAdmin.css
```diff
@@ -22137,6 +22137,7 @@ a[dir]:root[data-theme=default] .badge-dark:focus, a[data-theme=default][dir]:ro
 :root[data-theme=default] .ta-navbar-top a {
   color: #495057;
   text-decoration: none;
+  cursor: pointer;
```
Piedone, line 22140: Why are these CSS changes necessary?

Piedone (changes requested):

src/OrchardCore.Modules/OrchardCore.AuditTrail/Navigation/AuditTrailSettingsAdminMenu.cs
BanzragchUchral: First [instance](https://github.com/OrchardCMS/OrchardCore/blob/ef1e17bbe573a7c00c1ec355c47a334b9e94addc/src/OrchardCore.Modules/OrchardCore.AuditTrail/Navigation/AuditTrailAdminMenu.cs#L27) and [second ](https://github.com/OrchardCMS/OrchardCore/blob/ef1e17bbe573a7c00c1ec355c47a334b9e94addc/src/OrchardCore.Modules/OrchardCore.AuditTrail/Navigation/AuditTrailSettingsAdminMenu.cs#L29).

src/OrchardCore.Modules/OrchardCore.Users/Views/UserMenu.cshtml
BanzragchUchral: I was thinking about using `<button>` here, but I need to change a lot of CSS for that, and I went with this because the change would be minimal.

I will replace it with buttons. If I understand it correctly, [this](https://github.com/OrchardCMS/OrchardCore/blob/67256b454fe9cdb3f14ffc91bd3ed8ce9bcfbbd1/src/OrchardCore.Themes/TheAdmin/Views/Layout.cshtml#L57) can stay like this, right? It is a link, after all.

src/OrchardCore.Themes/TheAdmin/Views/Layout.cshtml
BanzragchUchral: I wasn't sure what to put between the buttons as a text. What should I do?

src/OrchardCore.Modules/OrchardCore.Users/Views/UserMenu.cshtml
Piedone: Yes, that looks good. The principle is that if it has button behavior, which this element does, then it's best if it's actually a `button`.

src/OrchardCore.Themes/TheAdmin/Views/Layout.cshtml
Piedone: Just the same: `@T["Collapse / expand menu"]`.

BanzragchUchral: Are you sure? The button looks like this:
image

I don't think we just add a simple text there.
image

Piedone: I'm sure (this is how a `<button>` works) but you have to adjust the CSS, of course.

BanzragchUchral: What should I do with the CSS? Should I hide the text? Or should I make the text fit there, even if the text wasn't there before?

Piedone: Hide it. Screen readers should be able to process it nevertheless.

src/OrchardCore.Modules/OrchardCore.Users/Views/UserMenu.cshtml
```diff
@@ -1,9 +1,9 @@
 @inject IAuthorizationService AuthorizationService
 <div class="dropdown">
     <form asp-route-area="OrchardCore.Users" asp-controller="Account" asp-action="LogOff" method="post" class="form-inline" class="no-multisubmit">
-        <a class="ml-3 dropdown-toggle" id="navbarDropdown" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" role="button" data-toggle-tooltip="tooltip" title="@User.Identity.Name">
+        <button class="ml-3 dropdown-toggle" id="navbarDropdown" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false" data-toggle-tooltip="tooltip" title="@User.Identity.Name">
```
Piedone, line 4: The rule of a button needing an inner text applies here as well, even if axe is confused by the `i` tag. Same in Layout.

src/OrchardCore.Themes/TheAdmin/Views/Layout.cshtml
```diff
@@ -65,7 +65,7 @@
         <div id="ta-left-sidebar" class="d-flex flex-column justify-content-between align-items-stretch">
             @await RenderSectionAsync("Navigation", required: false)
             <div class="footer">
-                <button class="leftbar-compactor" type="button" title="@T["Collapse / expand menu"]"></button>
+                <button class="leftbar-compactor">@T["Collapse / expand menu"]</button>
```
Piedone, line 68: It's good to keep the `title` as well for the tooltip.

src/OrchardCore.Themes/TheAdmin/wwwroot/css/TheAdmin.css
```diff
@@ -23213,6 +23219,7 @@ a[dir]:root[data-theme=default] .badge-dark:focus, a[data-theme=default][dir]:ro
   cursor: pointer;
   transition-duration: 0.2s;
   transform: scale(2, 1);
+  font-size: 0;
```
Piedone, line 23222: Is this really enough to make the text invisible? Check for Bootstrap classes that we may use instead, which are safer.

src/OrchardCore.Themes/TheAdmin/wwwroot/css/TheAdmin.css
```diff
@@ -46699,6 +46714,7 @@ a[dir]:root[data-theme=darkmode] .badge-dark:focus, a[dir]:root[data-theme=darkm
 }
 :root[data-theme=darkmode] .leftbar-compactor:before {
   content: "⟪";
+  font-size: initial;
```
Piedone, line 46717: Why is this necessary?

Piedone (changes requested):
Run it though the https://validator.w3.org/ validator as well, please.

src/OrchardCore.Themes/TheAdmin/wwwroot/css/TheAdmin.css
BanzragchUchral: The ⟪ would be invisible too otherwise. But I will try to find another way to hide the text.

hishamco: What about RTL? coz the direction is matter

src/OrchardCore.Themes/TheAdmin/wwwroot/css/TheAdmin.css
BanzragchUchral: I am thinking about using `display: none` but the problem I have at the moment is that this removes even the icon too.

Can I put the button text in a `<span>.` This way, I can easily target the correct part of the button. I don't think this would go against the rule.

src/OrchardCore.Themes/TheAdmin/wwwroot/css/TheAdmin.css
BanzragchUchral: Does RTL matter here? The going to be text hidden anyway, won't it?

BanzragchUchral: Does RTL matter here? The text going to be hidden anyway, won't it?

hishamco: I though the arrows will be shown, while it's in the content

Piedone: The `font-size` or other changes of Uchral here don't affect RTL.

src/OrchardCore.Themes/TheAdmin/wwwroot/css/TheAdmin.css
BanzragchUchral: @Piedone What do you think?

Piedone: Just please do whatever you think is suitable, you needn't check before trying out little things like this.

Piedone (changes requested):
Looks good but actually, not the W3C validator, but please run it through `html-validate` (it's a CLI tool, you can just use it with local files) so we can get the consistent results. Doesn't it show anything new?

BanzragchUchral:
I wasn't running it properly before. I found some mistakes, and most errors in the "UserMenu.cshtml" and "Layout. cshtml" should be fixed.

I didn't find any error in the Admin.css.

UserMenu:
- Attribute "class" duplicated
- Attribute "asp-route-returnurl" should be lowercase
- `<button>` is missing required "type" attribute

Layout:
- `<resources>` must not be self-closed
- `<button>` is missing required "type" attribute
- Anchor link must have a text describing its purpose

src/OrchardCore.Modules/OrchardCore.Users/Views/UserMenu.cshtml
```diff
@@ -10,9 +10,9 @@
             <div class="dropdown-divider"></div>
             @if (await AuthorizationService.AuthorizeAsync(User, Permissions.ManageOwnUserInformation))
             {
-                <a class="dropdown-item" asp-area="OrchardCore.Users" asp-action="Edit" asp-controller="Admin" asp-route-returnUrl="@FullRequestPath"><span><i class="far fa-address-card" aria-hidden="true"></i></span>&nbsp;<span> @T["Profile"]</span></a>
+                <a class="dropdown-item" asp-area="OrchardCore.Users" asp-action="Edit" asp-controller="Admin" asp-route-returnurl="@FullRequestPath"><span><i class="far fa-address-card" aria-hidden="true"></i></span>&nbsp;<span> @T["Profile"]</span></a>
```
Piedone, line 13: Please check if this still works and you're correctly redirected to the URL if you run OC via the dotnet CLI, not from VS via IIS Express, since casing can matter.

src/OrchardCore.Themes/TheAdmin/Views/Layout.cshtml
```diff
@@ -22,7 +22,7 @@
     <script asp-name="font-awesome-v4-shims" at="Head" version="6"></script>
     <script asp-name="js-cookie" at="Foot"></script>
     <script asp-name="admin" version="1" at="Foot"></script>
-    <resources type="Header" />
+    <resources type="Header"></resources>
```
Piedone, line 25: Why is this change necessary? I hoe `<resources>` doesn't end up in the HTML output though? Same for the Footer.

Piedone (changes requested):

Piedone (approved):

src/OrchardCore.Themes/TheAdmin/Views/Layout.cshtml
```diff
@@ -65,7 +67,9 @@
         <div id="ta-left-sidebar" class="d-flex flex-column justify-content-between align-items-stretch">
             @await RenderSectionAsync("Navigation", required: false)
             <div class="footer">
-                <button class="leftbar-compactor" type="button" title="@T["Collapse / expand menu"]"></button>
+                <button type="button" class="leftbar-compactor" title="@T["Collapse / expand menu"]">
+                    <span>@T["Collapse / expand menu"]</span>
```
Skrypt, line 71: Next time use class="d-none"

src/OrchardCore.Themes/TheAdmin/wwwroot/css/TheAdmin.css
Skrypt: Ok, we need to change this because I'm getting admin UI errors.

First thing. Don't make the changes directly in the .css files. It needs to be added in the SCSS files and then transpiled so that it is also included in the minified version. I will create a Pull Request to fix this and remove the changes dones directly in this file by re-transpiling the SCSS assets.

Piedone: Argh, I didn't notice that in this project we have SCSS...

PR #11435, opened 2022-03-25
Use ShellSettings.IsDefaultShell()

hishamco:


Piedone (changes requested):
Looks good! Please also check for other uses of `ShellHelper.DefaultShellName` and standardize it to the same `String.Equals(name, ShellHelper.DefaultShellName, StringComparison.OrdinalIgnoreCase)` because there are simple comparisons too.

Piedone:
Also, the Tenants `AdminController` has one occasion too.

hishamco:
> Also, the Tenants AdminController has one occasion too.

I didn't see any place other that the one you mentioned

Piedone:
Look in `Disable()`.

hishamco:
Is there anything left here?

hishamco:
Is there anything left here?

hishamco:
Is there anything left here?

Piedone:
You didn't address this:

> Looks good! Please also check for other uses of ShellHelper.DefaultShellName and standardize it to the same String.Equals(name, ShellHelper.DefaultShellName, StringComparison.OrdinalIgnoreCase) because there are simple comparisons too.

hishamco:
The `IsDefaultShell()` is already did the exact thing, could please refer to a place where this not applied

Piedone:
Look up the references:
image

hishamco:
Please let me check, may some project are not loaded in my solution. Coz sometimes I unload some project to save some memory ;)

hishamco:
As I expect, some of the project was unloaded, I just push a new commit, there remaining things is unrelated to the comparison

CrestApps:
@hishamco, @jtkech  just a thought here “may need a separate PR”, but, why don’t we add another extension method to the ShellSettings class where we can validate if the setting equal other shell by name? 

For example,

```
public bool IsEqualsTo(string compareTo) =>  !String.Equals(Name, compareTo, StringComparison.OrdinalIgnoreCase));
```
This would clean up some of the code where we can’t use `ShellHost.TryGetSettings`. It would centralize the lookup logic into one extension method instead of repeating the logic everywhere.

jtkech:
About `IsDefaultShell()` in ShellSettings that was before in internal methods defined as needed in a given context which doesn't mean these methods were originally done in the best way and with best names.

- Not at the right place in ShellSettings as it is not aware of the context. Checking for the Default Shell Name is sufficient only if it is a registered shellSettings, e.g. injected or retrieved from shellHost.

-  Doesn't do the best most of the time . Still related to the context, do we check for the Default shell before acting on it (we may want to be case insensitive) or before allowing it to do something (most frequent case), here I would have already done a case sensitive check in the old methods.

    Anyway most of the time we are checking registered shellSettings, so checking for the Default tenant could have already been done in a case sensitive way. Moreover now shellHost dictionaries are case insensitive so having "Default" and "default" tenants would have already failed on a duplicate key.

- For info at some point I thought about a shellHost extension to fully check that a shellSettings instance is really related to the current Default shell (knowing that the inner dicos now are case insensitive).

      public static bool IsDefaultShellSettings(
          this IShellHost shellHost,
          ShellSettings shellSettings) =>
              shellHost.TryGetSettings(ShellHelper.DefaultShellName, out var settings) &&
              shellSettings == settings;

    But this is too much when we know that we work with registered shellSettings (injected or retrieved from shellHost) where checking the name is sufficient and in a case sensitive way.

- The only one helper I can think about that could be defined in `ShellSettings`, would have a name that only describes what it does, maybe with an optional `ignoreCase` param false by default, if we still think that it could be useful, which is not the case when checking on registered shellSettings.

      public bool HasDefaultShellName() => Name == ShellHelper.DefaultShellName;

    But then not sure it is worth to have this helper, for example why not also having

      public bool IsRunning() => State == TenantState.Running;

    But why not, but in a separate ShellSettingsExtensions.cs

So in summary for `IsDefaultShell()` I think that we can always do a case sensitive check so just by using the `==` operator, and not sure we need an helper defined in ShellSettings. Maybe just a stringExtensions `IsDefaultShellName()` and/or a ShellSettingsExtensions `HasDefaultShellName()` but in all cases with a different name than `IsDefaultShell()`, and maybe with an optional `ignoreCase` param (false by default).

---

Now that we have oob case insensitive dicos, we can also update all the places where we get all settings to retrieve a tenant in a case insensitive way which now can be done oob in one pass by using `_shellHost.TryGetSettings()`, as @CrestApps started to do in one of its PR.

Piedone:
For the record, note the convo under https://github.com/OrchardCMS/OrchardCore/pull/12131 about why keep going here.

PR #11452, opened 2022-03-29
Make OrchardCore.Resources module always enabled

### LINKED ISSUE #11451: Disabling Resources Module Broke Site
### Describe the bug
I was able to disable the Resources module without any warning and hosed my site. I know it's a dumb mistake, it was kind of a "turn it off and back on" thing I was thinking I could do. Please don't make fun of me. 😂

InvalidOperationException: Unable to resolve service for type 'OrchardCore.Recipes.Services.IRecipeMigrator' while attempting to activate 'OrchardCore.Contents.Deployment.ExportContentToDeploymentTarget.ExportContentToDeploymentTargetMigrations'.
Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteFactory.CreateArgumentCallSites(Type implementationType, CallSiteChain callSiteChain, ParameterInfo[] parameters, bool throwIfCallSiteNotFound)

### To Reproduce
1. Go to modules
2. Search for Resources
3. Click Disable
4. Cry

### Expected behavior
I expect that either 1) there is no option to disable a required module, or 2) a very mean error.

I see this discussed regarding[ another issue in the past](https://github.com/OrchardCMS/OrchardCore/issues/3403)

Total random thought - it would be nice if there was a "safe mode" if someone does disable things they shouldn't, or somehow they're corrupted or missing, there's still some kind of basic interface to allow for fixing.

Skrypt:
Fixes #11451

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Resources/Manifest.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Resources/Manifest.cs
@@ -6,5 +6,6 @@ using OrchardCore.Modules.Manifest;
     Website = ManifestConstants.OrchardCoreWebsite,
     Version = ManifestConstants.OrchardCoreVersion,
     Description = "The Resources module allows you to declare scripts and stylesheets.",
-    Category = "Theming"
+    Category = "Theming",
+    IsAlwaysEnabled = true
 )]


hishamco (commented):
I don't think we should enable resources always. I know there are some modules are critical if they disabled they well break the site such as resources, roles .. etc

jtkech:
@hishamco 

Here `IsAlwaysEnabled` only means that you can"t disable the feature once it is enabled, so you can still not use it if not defined from a setup recipe, from configuration or from the app with our helpers and so on. We already use it for OC.Scripting, OC.Recipes, OC.Settings ...

        /// <summary>
        /// Once enabled, check whether the feature can't be disabled. Defaults to false.
        /// </summary>
        public bool IsAlwaysEnabled { get; set; } = false;

jtkech (approved):

Issue #11455, opened 2022-03-29
Issue with MongoBD store for OpenId Applications


vitalybrandes:
I am not sure if this is the right place to open bug,
I am using Orchard core cms and adding MongoDB nuget to store app and token in mongo collections based on @kevinchalet blogpost https://kevinchalet.com/2020/10/03/using-the-orchardcore-openid-management-feature-with-an-existing-openiddict-deployment/

ITEM : [#11371](https://github.com/OrchardCMS/OrchardCore/discussions/11371)

Mostly its work fine except "Client Credentials Roles". Its not saved to the application document.

 Orchard Core
![image](https://user-images.githubusercontent.com/61165619/160587703-d919d214-a608-4147-b947-2a2596695f4f.png)
---------------------------
Mongo
![image](https://user-images.githubusercontent.com/61165619/160588179-8efb89a4-cd1d-4d48-8d38-bec33996ee58.png)

### To Reproduce
Use MongoDB store in orchard core.
add application client credentials with predefined roles.

kevinchalet:
@vitalybrandes thanks for reporting this bug.

I sent https://github.com/OrchardCMS/OrchardCore/pull/11456 to fix it. It will likely be integrated to the nightly builds shorty after it's merged, but you'll need to check with @sebastienros what's the current policy for bug fix releases if you need a "public" patched version.

vitalybrandes:
@kevinchalet
Thank you for update.
Question,
is there any way to store openid secret not encrypted?

PR #11456, opened 2022-03-29
Fix the untyped properties fallback logic that prevents persisting roles with non-Orchard OpenID stores

kevinchalet:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/11455.

Note: this PR also cleans things up by using the `JsonSerializer.SerializeToElement()` API introduced in .NET 6.0.

/cc @vitalybrandes
/cc @sebastienros I targeted the main branch but feel free to select a different branch if you think it's worth shipping a hotfix for this bug.


sebastienros (approved):

vitalybrandes:
@kevinchalet 
Thank you for update.
Question,
is there any way to store openid secret not encrypted?

PR #11489, opened 2022-04-04
Fix invalid cookie name caused by parenthesis in path name

jardg:
Hi when the root path is a in a Windows directory that contains parenthesis I get the following antiforgery error
Error

I believe his happens since the project upgraded to a .NET Core version that no longer scapes characters in the antiforgery cookie (3 or 5 I don't remember the version). I encounter this error while deploying release version on Windows Server or while developing with Visual Studio when the project is inside a folder with parenthesis.

Repro steps:

1. Download OrchardCore project in a folder with parenthesis(or a parent folder) and launch OrchardCore.Cms.Web project. Then the error appears inmediately in the setup screen.

This pr fixes this error. I don't know if it could be more problematic characters since HttpUtility.UrlEncode method can get rid of them.

### DIFF
--- a/src/OrchardCore/OrchardCore/Modules/Extensions/ServiceCollectionExtensions.cs
+++ b/src/OrchardCore/OrchardCore/Modules/Extensions/ServiceCollectionExtensions.cs
@@ -312,7 +312,7 @@ namespace Microsoft.Extensions.DependencyInjection
                 var settings = serviceProvider.GetRequiredService<ShellSettings>();
                 var environment = serviceProvider.GetRequiredService<IHostEnvironment>();
 
-                var cookieName = "orchantiforgery_" + HttpUtility.UrlEncode(settings.Name + environment.ContentRootPath);
+                var cookieName = "orchantiforgery_" + HttpUtility.UrlEncode(settings.Name + environment.ContentRootPath.Replace("(", "").Replace(")", ""));
 
                 // If uninitialized, we use the host services.
                 if (settings.State == TenantState.Uninitialized)


src/OrchardCore/OrchardCore/Modules/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -312,7 +312,7 @@ private static void AddAntiForgery(OrchardCoreBuilder builder)
                 var settings = serviceProvider.GetRequiredService<ShellSettings>();
                 var environment = serviceProvider.GetRequiredService<IHostEnvironment>();
 
-                var cookieName = "orchantiforgery_" + HttpUtility.UrlEncode(settings.Name + environment.ContentRootPath);
+                var cookieName = "orchantiforgery_" + HttpUtility.UrlEncode(settings.Name + environment.ContentRootPath.Replace("(", "").Replace(")", ""));
```
hishamco, line 315: Can we define a `static` array of the invalid characters, then we can remove them from `ContentRootPath`

Skrypt: Let's use the slugify method here.

sebastienros, line 315: I think the issue is with the encoding logic, `UrlEncode` is apparently not adapted. Base64 or Uri.EscapeDataString (something close).

Also we can use a Guid.ToString("n") instead of the `ContentRootPath`. 
Details: https://github.com/OrchardCMS/OrchardCore/pull/5803

PR #11491, opened 2022-04-04
Move ISlugService to OrchardCore.Autoroute.Core

Skrypt:
Introduces new assembly OrchardCore.Autoroute.Core.

I moved some code into this assembly to clean up the OrchardCore.Autoroute module.
SlugService is now registered as an application-level singleton and it is moved to the OrchardCore assembly directly along with the LocalClock service for example.


Skrypt:
Ok, this breaks because the ISlugService is not registered before doing AddAntiForgery(). Also, probably need to be registered like the ShellSettings are.

hishamco:
I quote from the title

> Move ISlugService to OrchardCore.Autoroute.Core

I think `ISlugService` is better to be places in the abstractions not the Core

Skrypt:
That's what I did in the next commit.

src/OrchardCore/OrchardCore/OrchardCore.csproj
```diff
@@ -21,6 +21,7 @@
   <ItemGroup>
     <ProjectReference Include="..\OrchardCore.Abstractions\OrchardCore.Abstractions.csproj" />
     <ProjectReference Include="..\OrchardCore.Autoroute.Abstractions\OrchardCore.Autoroute.Abstractions.csproj" />
+    <ProjectReference Include="..\OrchardCore.Autoroute.Core\OrchardCore.Autoroute.Core.csproj" />
```
hishamco, line 24: I don't think it's a right reference, I presume that the service located in `OrchardCore.Autoroute.Abstractions` which is a right reference not the implementation one

hishamco (changes requested):

src/OrchardCore/OrchardCore/OrchardCore.csproj
Skrypt: Actually, I used the class directly here instead of the interface. I'm not injecting this service because of the fact that `ServiceCollectionExtensions` is where I use it and it is where we register our services with the DI. I need to make a direct reference to the class to have an instance that I can use. Also, not using the abstraction project ISlugService because I want to be sure I'm using the proper implementation of that SlugService.

hishamco: Why the proper implementation is not registered by the Autoroute module as we did for all the services?

Skrypt: The ISlugService is registered by the OrchardCore.Liquid module because we have a Liquid filter. If we enable the Slugify service on the Autoroute module it may cause issues if the Autoroute module is disabled. That's mainly why.

hishamco: Ya, I forgot it's registered inside `OC.Liquid`

jtkech:
@Skrypt 

> Ok, this breaks because the ISlugService is not registered before doing AddAntiForgery()

Yes we are in a `builder.ConfigureServices()` so while building a tenant container, so it would have to be an app level singleton as `IHostEnvironment`, the tenant singleton `ShellSettings` is a special case, we eagerly register it in the pre-built container we use to resolve module startups an then run their `ConfigureServices()`.

So would need to be an app level singleton registered for example in our `AddDefaultServices`, so defined in `OrchardCore` and the interface in `OC.Abstractions`, OR don't change anything and just make a static Slug component/helper that does the same and that could be used here, maybe the default `ISlugSrvice` could use this static component (but not necessarily), as I did at some point for the `IdGenerator` in `OC.Abstractions`.

Skrypt:
I tried in the `AddDefaultServices()` without success. Maybe it was not a singleton though. I will try again.

Skrypt:
Thanks @jtkech it works now as a singleton. I tested the Liquid filter on a Liquid page and it works so far.

Skrypt:
Changed the dependency of `OrchardCore.Autoroute.Core` to `OrchardCore.Autoroute.Abstractions` on the `OrchardCore` project. Created a specific implementation of the SlugService in the `OrchardCore` project which is now registered through DI in the `OrchardCore` project separately from the `SlugService` implementation in the `OrchardCore.Autoroute.Core` project which is getting registered through the `OrchardCore.Liquid` module.

Now, we register the "Autoroute SlugService implementation" in the Liquid module because it is required by the Slugify Liquid filter but we should probably just move the Liquid filter to the Autoroute module. I'm keeping it as it was for now because we may think about something else for these later on.

jtkech:
I would have just created a new static component helper in `OC.Abstractions` for what you need, OR if you still want to always inject `ISlugService`, just move the implementation from `OC.Liquid` to `OrchardCore` and the interface definition to `OC.Abstractions`, not in `OC.Autoroute.***`.

Skrypt:
Ok, now the SlugService is registered only as an Application level singleton and it works for all OC modules. Removed the OrchardCore.Autoroute.Abstraction project. Moved the ISlugService to OrchardCore.Abstractions.

I think it makes sense that we register this as an Application level singleton. It doesn't need to be a scoped service like it was before.

jtkech (approved):

src/OrchardCore.Modules/OrchardCore.Liquid/Services/SlugService.cs
```diff
@@ -63,5 +64,10 @@ public string Slugify(string text)
 
             return new string(slug.AsSpan()[..Math.Min(slug.Length, MaxLength)]).Normalize(NormalizationForm.FormC);
         }
+
+        public string Slugify(string text, char hyphen)
+        {
+            throw new NotImplementedException();
```
hishamco, line 70: Is this by intention?

src/OrchardCore/OrchardCore.Abstractions/Modules/Services/ISlugService.cs
```diff
@@ -0,0 +1,21 @@
...
+        string Slugify(string text);
+
+        /// <summary>
+        /// Transforms specified text to a custom form generally not suitable for URL slugs.
+        /// Allows you to use a specified separator char.
+        /// </summary>
+        /// <param name="text">The text to transform.</param>
+        /// <param name="hyphen">The separator char</param>
+        /// <returns>The slug created from the input text.</returns>
+        string Slugify(string text, char hyphen);
```
hishamco, line 19: hyphen -> separator

src/OrchardCore/OrchardCore/Modules/Services/SlugService.cs
```diff
@@ -0,0 +1,71 @@
...
+using Cysharp.Text;
+
+namespace OrchardCore.Modules.Services
+{
+    public class SlugService : ISlugService
+    {
+        private const char Hyphen = '-';
+        private const int MaxLength = 1000;
+
+        public string Slugify(string text)
```
hishamco, line 13:
```csharp
public string Slugify(string text) => Slugify(text, Hyphen);
```

hishamco (changes requested):

src/OrchardCore.Modules/OrchardCore.Liquid/Services/SlugService.cs
Skrypt: Yes, I added a method in the interface but we deprecate this class so no need to implement the method here.

hishamco: We deprecate the class but it's still anyone can use it unless you use `true` in the `Obsolete` attribute

Skrypt: Yes, just like we do in other places we give time for people to migrate their code.

hishamco: Sure, but my comment is the deprecated class and the new one should be functional, unless you enforce the people to migrate their code immediately by settings `[Obsolete("", true)]`

Skrypt: They are both functional, I'm not sure what you mean. Here, the SlugService class was only having one method to return a slug with a hyphen. We don't need to implement the new method with the custom separator because this is an addition that I just made. The real new SlugService is in the OrchardCore project which contains an implementation of the Slugify method with a custom separator. Here, I don't want to implement this method in a class that we will deprecate, which may lead to encouraging people using this class when they shouldn't.

hishamco (approved):

Skrypt:
Okay waiting on @sebastienros approval on this one.

sebastienros:
I am ok to look at the PR but not with the intent of fixing the linked issue, it's not the correct solution IMO.

agriffard:
Conflicts to solve.

hishamco:
Resolve the conflict

Skrypt:
2 approvals. I'm leaving someone else to merge this one. I fixed the remaining conflicts.

Skrypt:
Oops, don't merge yet. Unit tests failing, that's a new one.

Skrypt:
Done.

Skrypt:
Fixed build again.

PR #10782, opened 2021-11-30
Update graphql.net to 4.6.1

MikeKry:
Gatsby in version 3 and 4 is not working with current version of GraphQL in OrchardCore, so I have tried to update GraphQL.NET package. 

I am currently testing it, so it does not break anything, right now it looks ok, but it definitely needs some more checking. I will be testing it further on existing projects. 

It was quite a lot of work, so I decided to create a PR right away, so we can potentially merge it.

related to #10781
related also to #10760

PR #11499, opened 2022-04-07
GraphQL4 - resolve async write using System.Text.Json

MikeKry:
Replace newtonsoft.json with System.Text.Json to allow async write directly to response

handles - buffered stream issue from (https://github.com/OrchardCMS/OrchardCore/pull/10782)


sebastienros (approved):

sebastienros:
I like it

Skrypt (approved):

PR #11502, opened 2022-04-08
MiniProfiler visibility based on permissions

DotCat1985:
Hi,
I committed a change related to issue [11144](https://github.com/OrchardCMS/OrchardCore/issues/11144).
Please let me know what you think.
Thank you for attention.


Skrypt (approved):
So basically, if a role has the permission then all its users will see the MiniProfiler. Maybe we should also have a user-specific configuration to enable/disable the MiniProfiler too. Else, the PR is fine.

DotCat1985:
> So basically, if a role has the permission then all its users will see the MiniProfiler. Maybe we should also have a user-specific configuration to enable/disable the MiniProfiler too. Else, the PR is fine.

I agree. If the user-specific configuration is necessary to enable/disable the MiniProfiler, please let me know so I will develop it soon.

sebastienros:
Current PR is correct. 
I think with this change we need a "switch" since otherwise it appears all the time (disabling the feature is not the way to hide it anymore)
So OK to add a user setting. You can look at how TimeZone per user is implemented (driver + view) here https://github.com/OrchardCMS/OrchardCore/tree/main/src/OrchardCore.Modules/OrchardCore.Users/TimeZone

PR #11528, opened 2022-04-12
Unpublishing and editing content item with an uninitialized ContentPickerField breaks (Lombiq Technologies: OCORE-92)

### LINKED ISSUE #11527: Unpublishing and editing content item with an uninitialized ContentPickerField breaks
### Describe the bug

If you have a custom ContentPart that contains a `ContentPickerField` and it's not initialized (so `null`) and you try to unpublish the content type it will say it was unpublished but that won't be true. If you try to edit and publish the same content item you will get "An unhandled exception occurred while processing the request." error.

I realize this can be avoided by giving the field a default value (`public ContentPickerField Related { get; set; } = new();`), but it's easy to miss and the resulting symptoms are not easy to connect with the true problem.

### To Reproduce
1. Create a content type via migration.
2. Start and set up the webapp.
3. Create a new content item of the content type.
4. Add part (e.g. `public class RelatedPart : ContentPart { public ContentPickerField Related { get; set; } }`) via migration update step.
5. Start up the webapp again.
6. Try to unpublish or edit the Content item created in step 3.
7. Experience the bug described above.

### Expected behavior
The unpublishing or editng should be successful.

### Screenshots
The problem is here:
image
image
The `if (jPart == null)` and `if (jField == null)` will never be true because if the value is JSON null, it will be a JValue type so the cast on the preceding line will cause the exception you can see on the screenshot. This can be fixed if type matching is used instead of casting. 

PR will shortly follow.

DAud-IcI:
Fixes #11527

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Extensions/ContentPartFieldDefinitionExtensions.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Extensions/ContentPartFieldDefinitionExtensions.cs
@@ -13,7 +13,7 @@ public static class ContentPartFieldDefinitionExtensions
         ContentItem contentItem)
         where TField : ContentField
     {
-        if (contentItem.Content[fieldDefinition.PartDefinition.Name] is not JObject jPart ||
+        if (((JObject)contentItem.Content)[fieldDefinition.PartDefinition.Name] is not JObject jPart ||
             jPart[fieldDefinition.Name] is not JObject jField)
         {
             return null;
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Extensions/ContentPartFieldDefinitionExtensions.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Extensions/ContentPartFieldDefinitionExtensions.cs
@@ -34,7 +34,10 @@ public static class ContentPartFieldDefinitionExtensions
         foreach (var fieldDefinition in fieldDefinitions)
         {
             var field = fieldDefinition.GetContentField<TField>(contentItem);
-            if (field is not null) yield return (fieldDefinition, field);
+            if (field is not null)
+            {
+                yield return (fieldDefinition, field);
+            }
         }
     }
 }
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Extensions/ContentPartFieldDefinitionExtensions.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Extensions/ContentPartFieldDefinitionExtensions.cs
@@ -33,8 +33,8 @@ public static class ContentPartFieldDefinitionExtensions
     {
         foreach (var fieldDefinition in fieldDefinitions)
         {
-            if (fieldDefinition.GetContentField<TField>(contentItem) is not { } field) continue;
-            yield return (fieldDefinition, field);
+            var field = fieldDefinition.GetContentField<TField>(contentItem);
+            if (field is not null) yield return (fieldDefinition, field);
         }
     }
 }
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Extensions/ContentPartFieldDefinitionExtensions.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Extensions/ContentPartFieldDefinitionExtensions.cs
@@ -11,12 +11,16 @@ public static class ContentPartFieldDefinitionExtensions
     public static TField GetContentField<TField>(
         this ContentPartFieldDefinition fieldDefinition,
         ContentItem contentItem)
-        where TField : ContentField =>
-        contentItem.Content[fieldDefinition.PartDefinition.Name] is JObject jPart &&
-        jPart[fieldDefinition.Name] is JObject jField &&
-        jField.ToObject<TField>() is { } field
-            ? field
-            : null;
+        where TField : ContentField
+    {
+        if (contentItem.Content[fieldDefinition.PartDefinition.Name] is not JObject jPart ||
+            jPart[fieldDefinition.Name] is not JObject jField)
+        {
+            return null;
+        }
+
+        return jField.ToObject<TField>();
+    }
 
     /// <summary>
     /// Returns each field from <paramref name="fieldDefinitions"/> that exists in <paramref name="contentItem"/> in a
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Extensions/ContentPartFieldDefinitionExtensions.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Extensions/ContentPartFieldDefinitionExtensions.cs
@@ -1,0 +1,36 @@
+﻿using System.Collections.Generic;
+using Newtonsoft.Json.Linq;
+
+namespace OrchardCore.ContentManagement.Metadata.Models;
+
+public static class ContentPartFieldDefinitionExtensions
+{
+    /// <summary>
+    /// Returns the value of the defined content field from the <paramref name="contentItem"/>.
+    /// </summary>
+    public static TField GetContentField<TField>(
+        this ContentPartFieldDefinition fieldDefinition,
+        ContentItem contentItem)
+        where TField : ContentField =>
+        contentItem.Content[fieldDefinition.PartDefinition.Name] is JObject jPart &&
+        jPart[fieldDefinition.Name] is JObject jField &&
+        jField.ToObject<TField>() is { } field
+            ? field
+            : null;
+
+    /// <summary>
+    /// Returns each field from <paramref name="fieldDefinitions"/> that exists in <paramref name="contentItem"/> in a
+    /// tuple along with its <see cref="ContentPartFieldDefinition"/>.
+    /// </summary>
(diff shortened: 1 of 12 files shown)


src/OrchardCore.Modules/OrchardCore.ContentFields/Extensions/ContentPartFieldDefinitionExtensions.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Collections.Generic;
+using Newtonsoft.Json.Linq;
```
Skrypt, line 2: Just some thought here. Could we do the same but without using Newtonsoft.Json and replace it with System.Text.Json?

DAud-IcI: The `ContentItem.Content` points to `OrchardCore.ContentManagement.ContentElement.Data` which is a Newtonsoft `JObject`. This class doesn't do any serializing, just access this existing property so trying to use  System.Text.Json here would require conversion. Better table it until `ContentElement` itself is ported to STJ, I think.

Skrypt: Ok, forget it for now then.

ns8482e: Will Linq be concern here for allocation and performance?

DAud-IcI: We pass on the enumerables without using an intermediary list like before, so that should improve allocation. I'm not sure about the performance, but we don't do anything complex so it should be fine. Also these providers are only called when you update a content item's field; that's relatively rare so you shouldn't be concerned.

jtkech:
Why not just replacing `(JObject)` castings by `as JObject`.

Hmm, maybe from the beginning would have been better to use the ContentElement extensions already using `as JObject` and cached Elements, this in place of using directly the dynamic Content.

So for now can you try the above suggestion and see if it fixes the issue.

src/OrchardCore.Modules/OrchardCore.ContentFields/Extensions/ContentPartFieldDefinitionExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// <summary>
+    /// Returns the value of the defined content field from the <paramref name="contentItem"/>.
+    /// </summary>
+    public static TField GetContentField<TField>(
+        this ContentPartFieldDefinition fieldDefinition,
+        ContentItem contentItem)
+        where TField : ContentField =>
+        contentItem.Content[fieldDefinition.PartDefinition.Name] is JObject jPart &&
+        jPart[fieldDefinition.Name] is JObject jField &&
+        jField.ToObject<TField>() is { } field
```
sebastienros, line 17: does `is { }` mean `!= null`. If so then `!= null` is easier to comprehend.

sebastienros:
Apart from the new c# `is { }` usages this is fine

src/OrchardCore.Modules/OrchardCore.ContentFields/Extensions/ContentPartFieldDefinitionExtensions.cs
DAud-IcI: It's a combined operation for pattern matching introduced in C# 8. It does `!= null` check and assignment to the `field` variable. So it makes the code more compact and lowers the cognitive complexity (once you are used to it anyway). Back in C# 7 we'd use `jField.ToObject<TField>() is var field` for the same, but starting C# 8 this empty object pattern is the norm.

If I replaced this with `!= null` then it would look like this:
```csharp
        jField.ToObject<TField>() != null
            ? jField.ToObject<TField>()
            : null;
```
I think that's less efficient and visually noisier. Please comment if you want C# 8 style, C# 7 style, or the old fashioned null check and explicit assignment.

sebastienros: > and lowers the cognitive complexity (once you are used to it anyway)

Like pain, after a while you don't realize it's there, so pain isn't real actually.
We had a meeting today and nobody could explain what it does. Everyone could explain `!= null` though.

Please keep the old style, and add more lines if necessary for readability or prevent two calls to ToObject.

DAud-IcI: It's not like pain, more like learning a new technique. First it's weird and awkward, but it you give it a chance you start seeing the utility in practice and it grows on you. Change is good.
(also we use this pattern matching in Lombiq everywhere)

But it's you guys' repo, so I abide by your decision and have rewrote the method.

ns8482e: May be you can use `jField is not null` it’s more readable

DAud-IcI: Ok. I have already eliminated the null check here, but I applied your suggestion to the null check in the other method in this file.

However this is starting to get a bit confusing. Does the OC repo have a style guide? I couldn't find it. If not, do you have plans to make one?

ns8482e: https://github.com/OrchardCMS/OrchardCore/blob/main/CONTRIBUTING.md

DAud-IcI: That links to [ASP.Net Core's engineering guidelines](https://github.com/dotnet/aspnetcore/wiki/Engineering-guidelines#coding-guidelines). If you can edit that, it would be good to add something about null checking conventions based on this discussion. If not, an OC-specific guide would still be valuable to keep a record of such decisions. Just a thought.

DAud-IcI:
Do I need to do anything else or is this ready to be merged?

jtkech:
As said I would have just replaced the `(JObject)` castings with `as JObject` to fix the issue.

    var jPart = contentItem.Content[fieldDefinition.PartDefinition.Name] as JObject;

Also, @sebastienros why not making the ContentElement `JObject Data` getter public (only keep the setter internal), this in place of dealing with the `dynamic Content`, as we do in `ContentExtensions`.

    var jPart = contentItem.Data[fieldDefinition.PartDefinition.Name] as JObject;

Or at least do a cast on the `contentItem.Content` itself before accessing its properties, less reflections.

    var jPart = ((JObject)contentItem.Content)[fieldDefinition.PartDefinition.Name] as JObject;

But this PR LGTM so I will approve.

Hmm, as suggested above, maybe still add a cast on the `contentItem.Content` itself, e.g. for this line.

    if (((JObject)contentItem.Content)[fieldDefinition.PartDefinition.Name] is not JObject ...)

Or directly use `.Data` if we make its getter public.

But yes all the above is not crucial as here we are updating items, but making the `.Data` getter public may be useful for less reflections in other places using `.Content`, maybe another PR if worth to do ;)

So as said I will still approve this PR ;)

jtkech (approved):

Skrypt:
I agree with the `as JObject;` comment.

DAud-IcI:
> As said I would have just replaced the (JObject) castings with as JObject to fix the issue.

If possible I'd prefer to keep using `is not JObject` because it allows the type checks to be chained with boolean operators. After all it's one big type checking operation to reach the `JField` anyway so sticking them together seems logical.

> Hmm, as suggested above, maybe still add a cast on the contentItem.Content itself, e.g. for this line.

Good point, I have applied the suggested cast.

> But yes all the above is not crucial as here we are updating items, but making the .Data getter public may be useful for less reflections in other places using .Content, maybe another PR if worth to do ;)

If this comment gets two thumbs ups I will do it :)
(I mean make the getter public and replace `.Content` with `.Data` where applicable in a separate PR.)

jtkech:
@DAud-IcI Okay, all good then for me, thanks for the change.

src/OrchardCore.Modules/OrchardCore.ContentFields/Extensions/ContentPartFieldDefinitionExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// </summary>
+    public static IEnumerable<(ContentPartFieldDefinition Definition, TField Field)> GetContentFields<TField>(
+        this IEnumerable<ContentPartFieldDefinition> fieldDefinitions,
+        ContentItem contentItem)
+        where TField : ContentField
+    {
+        foreach (var fieldDefinition in fieldDefinitions)
+        {
+            var field = fieldDefinition.GetContentField<TField>(contentItem);
+            if (field is not null) yield return (fieldDefinition, field);
```
jptissot, line 37: brakcets

DAud-IcI: Thanks, fixed.

PR #11529, opened 2022-04-13
Widgets overwrite zone tag helper - log warning when the Zone isn't a Shape. (Lombiq Technologies: OCORE-90)

### LINKED ISSUE #11481: Widgets overwrite zone tag helper.
### Describe the bug
I can't use both `<zone>` tag helpers and widgets for the same zone. If you have a widget in the zone, the content in zone tag helpers are ignored.

### To Reproduce
1. Create a new Orchard Core CMS Web App solution.
2. Add a new Orchard Core Theme project
2. Replace the _Views/Layout.liquid_ with _Views/Layout.cshtml_ having the following content to keep it simple:
```razor

<zone name="Footer">
<p>Hello from zone tag helper!</p>
</zone>

@await RenderSectionAsync("Content", required: false)
@await RenderSectionAsync("Footer", required: false)
```
3. Launch and set up the webapp.
4. Go to /Admin/Themes, and select the newly created theme.
5. Confirm "Hello from zone tag helper!" is visible on the home page.
6. Create a zone "Footer" and a layer "Always" (boolean true condition).
8. Create a widget content type with Html Body and add a new item with "Hello from Widget!" to the Footer zone.
9. Go back to the home page. You only see `Hello from Widget!` and any tag helpers are ignored.

### Expected behavior
You should see
```
Hello from zone tag helper!
Hello from Widget!
```
in the footer.

DAud-IcI:
Fixes #11481

Add warning logging when ZoneTagHelper can't be applied.

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement/TagHelpers/ZoneTagHelper.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/TagHelpers/ZoneTagHelper.cs
@@ -47,7 +47,7 @@ namespace OrchardCore.DisplayManagement.TagHelpers
             else
             {
                 _logger.LogWarning(
-                    "Unable to add shape to the zone using the <zone2> tag helper because the zone's type is " +
+                    "Unable to add shape to the zone using the <zone> tag helper because the zone's type is " +
                     "\"{ActualType}\" instead of the expected {ExpectedType}",
                     zone.GetType().FullName,
                     nameof(Shape));
--- a/src/OrchardCore/OrchardCore.DisplayManagement/TagHelpers/ZoneTagHelper.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/TagHelpers/ZoneTagHelper.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Razor.TagHelpers;
+using Microsoft.Extensions.Logging;
 using OrchardCore.DisplayManagement.Layout;
 using OrchardCore.DisplayManagement.Shapes;
 
@@ -13,10 +14,12 @@ namespace OrchardCore.DisplayManagement.TagHelpers
         private const string NameAttribute = "name";
 
         private readonly ILayoutAccessor _layoutAccessor;
+        private readonly ILogger<ZoneTagHelper> _logger;
 
-        public ZoneTagHelper(ILayoutAccessor layoutAccessor)
+        public ZoneTagHelper(ILayoutAccessor layoutAccessor, ILogger<ZoneTagHelper> logger)
         {
             _layoutAccessor = layoutAccessor;
+            _logger = logger;
         }
 
         [HtmlAttributeName(PositionAttribute)]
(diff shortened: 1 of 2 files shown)


src/OrchardCore/OrchardCore.DisplayManagement/TagHelpers/ZoneTagHelper.cs
```diff
@@ -41,6 +44,14 @@ public async override Task ProcessAsync(TagHelperContext context, TagHelperOutpu
             {
                 await shape.AddAsync(childContent, Position);
             }
+            else
+            {
+                _logger.LogWarning(
+                    "Unable to add shape to the zone using the <zone2> tag helper because the zone's type is " +
```
Skrypt, line 50: `<zone>` instead of `<zone2>`

ns8482e: @jtkech should we take a step further and actually allow shapes in position wrapper for rendering?

src/OrchardCore/OrchardCore.DisplayManagement/TagHelpers/ZoneTagHelper.cs
```diff
@@ -41,6 +44,14 @@ public async override Task ProcessAsync(TagHelperContext context, TagHelperOutpu
             {
                 await shape.AddAsync(childContent, Position);
             }
+            else
+            {
+                _logger.LogWarning(
+                    "Unable to add shape to the zone using the <zone> tag helper because the zone's type is " +
+                    "\"{ActualType}\" instead of the expected {ExpectedType}",
+                    zone.GetType().FullName,
+                    nameof(Shape));
```
hishamco, line 53: How the `{ExpectedType}` will be different while the `nameof(Shape)` returns "Shape" all the time?!!

hishamco (changes requested):

src/OrchardCore/OrchardCore.DisplayManagement/TagHelpers/ZoneTagHelper.cs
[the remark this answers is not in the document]
DAud-IcI: `{ExpectedType}` is always supposed to be `"Shape"`. I only made it a parameter because I think it's clearer this way. Do you want me to remove the parameter and just type "Shape" in the message string instead?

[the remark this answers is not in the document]
hishamco: > {ExpectedType} is always supposed to be "Shape"

No need to remove it, it's fine to keep it this way

DAud-IcI: Not everyone knows about the pre-rendering and if you are developing a theme then it can have logical uses. (In my case I had a shape in the base theme's layout's header that the derived theme can override to inject stuff). 
I didn't know about PositionWrapper earlier and I thought that something else broke. The purpose of this PR is to be more developer friendly by warning about the irregular behavior.

hishamco (approved):

sebastienros (approved):

PR #11542, opened 2022-04-14
Add missing permission registration in OC.ReverseProxy module

hishamco:


Skrypt:
Ok, I agree with the changes since it looks like something that has been forgotten.

Though, I personally don't like the fact that the Permission name is ReverseProxySettings which is the same as an actual Class name. I would rename it to ManageReverseProxySettings.

hishamco:
I will rename it on other PR, while the intend here is to add the missing registration

Skrypt (approved):

Issue #9330, opened 2021-04-30
GraphQL session MARS/DataReader is open problem


tropcicstefan:
### Describe the bug
Resolving graphQLFields of content part through DataLoader causes MARS or Datareader is open exception.
### To Reproduce
Latest dev branch
Make content Part with two taxonomy fields(JournalistCategory, ArticleCategory). Make custom GraphQLObjectType that exposes those data as two `Field<ListGraphType<ContentItemInterface>, ContentItem[]>` and resolve those fields querying DB for linked taxonomy and parsing it to get that specific ContentItem. It appears as having two or more graphQLFields when resolving through ISession is too much
    
`    

    public class CustomCategoryObjectType : ObjectGraphType<CustomCategoryPart>
    {
        public CustomCategoryObjectType(IStringLocalizer<CustomCategoryObjectType> S)
        {
            Name = "CustomCategory";
            Description = S["Alternative part definition"];
            Field<ListGraphType<ContentItemInterface>, ContentItem[]>()
                .Name("journalistMetadata")
                .Description("Exposed journalist taxonomy metadata")
                .PagingArguments()
                .ResolveAsync(context =>
                {
                    var serviceProvider = context.ResolveServiceProvider();
                    var accessor = serviceProvider.GetRequiredService<IDataLoaderContextAccessor>();
                    var session = serviceProvider.GetService<ISession>();

                    var contentItemLoader = accessor.Context.GetOrAddBatchLoader<string, ContentItem>("GetJournalistsById",
                        ci => LoadTaxContentItems(ci, session, Const.Taxonomy.Journalist));
                    return contentItemLoader.LoadAsync(context.Page(context.Source.JournalistCategory.TermContentItemIds));
                });
            Field<ListGraphType<ContentItemInterface>, ContentItem[]>()
                .Name("categoryMetadata")
                .Description("Exposed category taxonomy metadata")
                .PagingArguments()
                .ResolveAsync(context =>
                {
                    var serviceProvider = context.ResolveServiceProvider();

                    var accessor = serviceProvider.GetRequiredService<IDataLoaderContextAccessor>();
                    var session = serviceProvider.GetService<ISession>();

                    var contentItemLoader = accessor.Context.GetOrAddBatchLoader<string, ContentItem>("GetCategoryById",
                        ci => LoadTaxContentItems(ci, session, Const.Taxonomy.Category));
[...]

PR #11548, opened 2022-04-16
Revert "Non awaited task in GraphQL DataLoaderExtensions (#11536)"

jtkech:
This reverts commit 5ba889e56d9e6e8ff817ae24f48d0135094143cf.

@sebastienros I reverted #11536 because first I didn't updated the right method, the one that uses a Task.WhenAll(), anyway If I remove the Task.WhenAll in the right extension method it works for a CPField with only one item (only CPField uses this method) otherwise it was blocking, you already tried to do exactly the same in #9769.

I noticed that the Task.WhenAll is not used to do concurrent queries but to grab all needed Ids, maybe this is here that concurrent accesses are done on our ContentItem Elements, notice that I was not able to repro #9330. If I remove the Task.WhenAll() and add some await, it tries to do a query for each Id and it is blocking on the 2nd one.

As I read GraphQL is more intended to provide a single item or a collection but both by an unique key, to support a multiple keys input they introduced an extension that exactly does what we do, i.e. they also use a Task.WhenAll().

For ListPart we use a CollectionBatchLoader that returns a collection but each group set being related to an unique key, the `ContainedPart.ListContentItemId`. For a CPField I could mimic the behavior of a ListPart by scope caching a relation between a guid and the item Ids of a given CPField, then I used this uid to retrieve the items. It works but for now if 2 CPFields reference the same item, this item is rendered once.

Let me know if it is worth to work on it, maybe not as it seems that there is an incoming version upgrade of GraphQL.

### DIFF
--- a/src/OrchardCore/OrchardCore.Apis.GraphQL.Abstractions/Extensions/DataLoaderExtensions.cs
+++ b/src/OrchardCore/OrchardCore.Apis.GraphQL.Abstractions/Extensions/DataLoaderExtensions.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using GraphQL.DataLoader;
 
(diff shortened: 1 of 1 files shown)


tropcicstefan:
Don't think that is a solution either. Because the essence of a problem is in parallel resolving of graphql field. So if a graphql query contains listpart and cpf on same level they will execute in parallel and cause MARS problem.

jtkech:
@tropcicstefan

Ah okay, do you mean that we have the same issue with ListPart? Hmm, when I will have time I'll try to see more precisely what we are doing and also what GraphQL is doing on its side, interesting ;)

PR #11551, opened 2022-04-17
SelectedContentType query param added multiple times

### LINKED ISSUE #11550: SelectedContentType query param added multiple times
### Describe the bug

Note: Just added to the 1.6 milestone but no problem if you move it to 1.x

### To Reproduce
1. Go to the Admin Content Items page.
2. Click on the type info (provided by the summary view) of a given item.
3. You see the items of the same type, then click again and again on the type info.
4. The SelectedContentType param is added again and again to the query string.

### Expected behavior
The SelectedContentType param is added once to the query string..

### Screenshots

Another related issue is that when we add the SelectedContentType param, we assume that the `?` query separator is already there, which is not the case when listing items of a ListPart under the Admin e.g. BlogPosts, so here we just append the SelectedContentType to the current path, the result is that when clicking on the type info it returns a 404.

This in place of staying on the same page because when we list items of a ListPart under the Admin, anyway we just ignore the SelectedContentType param. So, because the summary view is not aware of this, we can still add this param even if it may be useless, but correctly and only once as described above.

jtkech:
Fixes #11550


Skrypt:
Is there not a way to prevent this from the controller itself?

sebastienros:
I agree with @Skrypt I think there is a way to get the current url and "update" the `Options.SelectedContentType` route value to rebuild the current url.

jtkech:
Content options are submitted through query string params, we mainly bind to filter results, the term type option can be overridden by the old `SelectedContentType` that can be overridden by the `contentTypeId` which is used by the admin menu.

Yes we could use a route value to generate the url but we may loose other current options, here we have no access to the `ContentOptionsViewModel` and options are not in ambient route values. We would need to parse the current query string to retrieve current contextual options and add them to route values before generating the url.

The admin menu uses the `contentTypeId` as a route value to generate an url to link to the list of items of a given type, but here there is no other contextual options to retrieve.

Skrypt:
I will take a look with you when I get some time.

jtkech:
What about this one?

- It fixes 2 issues on the type info links, generating bad links if there is not yet a contentType option param in the query string, or adding duplicates contentType params.

- As said, we don"t use route values for content options, so we can't use them to generate a link.

- Unless when we come from a left admin menu dedicated to a given content type, in that case we use a route value, we can see it in the url where it is no more in the query string but as a path segment.

    So I now also check this route value, if it exists nothing to do, just use the current full request path. This use case was already fixed by this PR but now we don't add an useless query string param.

jtkech:
Just for infos and as a reminder.

- Another approach is to generate a link by using the `contentTypeId` route value, this way the type info link will act as if you click from the left admin menu e.g. `Content > Content Types > Article`.

    Make sense but I didn't do that, because when listing items of a ListPart, you would allow to list all items of this type, even if the type is not Listable,  and alongside items of other ListParts.

    That said we did the above for a client to allow him to retrieve all filters capabilities and bulk actions that are not present when listing items in the context of a ListPart.

- Another use case is when you are listing items of all types, let's say on pageNum 3, then if you click on a type info link, the result may be nothing if the selected type has less than 3 pages of items.

- Note: Excluding LisParts, there are 3 ways to pass the content type, with the route value `contentTypeId` (used by the admin menu), the old `Options.SelectedContentType` query param (used by the type info link), and with the `?q=...` query param embedding list filter values (binded from list filters).

So many interesting things to do but out of the scope of this PR, but we will be able to retrieve the above info if we plan to do a more complete PR. In the meantime we have to live with some 404s and duplicate query string params, but not so hard to override some views.

So for now I'm closing this PR.

PR #11579, opened 2022-04-22
ImageSharp.Web 2.0.0

agriffard:
https://www.nuget.org/packages/SixLabors.ImageSharp.Web/2.0.0
https://github.com/SixLabors/ImageSharp.Web/releases/tag/v2.0.0

### DIFF
--- a/src/OrchardCore.Build/Dependencies.props
+++ b/src/OrchardCore.Build/Dependencies.props
@@ -43,7 +43,7 @@
     <PackageManagement Include="PdfPig" Version="0.1.5" />
     <PackageManagement Include="Serilog.AspNetCore" Version="5.0.0" />
     <PackageManagement Include="Shortcodes" Version="1.3.3" />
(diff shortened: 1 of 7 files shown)


agriffard:
Not sure how to fix this issue with the type of the `commands` parameter that changed to a `CommandCollection`:
https://github.com/OrchardCMS/OrchardCore/blob/9f963e0ebccb562aa6545497eb84100844cc24d5/src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaImageSharpConfiguration.cs#L50

https://github.com/OrchardCMS/OrchardCore/blob/9f963e0ebccb562aa6545497eb84100844cc24d5/src/OrchardCore/OrchardCore.Media.Abstractions/IMediaTokenService.cs#L12

Do we need to add a reference to ImageSharp in OrchardCore.Media.Abstractions?

mariojsnunes:
Is there a way to override the CacheHashLength or make it smaller by default? I've had issues of container startup time due to this.

agriffard:
See #11585

PR #11582, opened 2022-04-23
Add password to user creating form

### LINKED ISSUE #11080: Creating a new user with password
By default apps use basic-authentication to authenticate users into their site including Microsoft Identity based apps. No user is able to authenticate/login without providing "username" and "password". Alternatively, an admin could enable some sort of external authentication providers like Azure AD or any other providers to avoid the need to manage the authentication by the app directly. These external providers are optional and not enabled by default in OC so creating a user with no password by default adds no value since no the user can't login without one. 

Currently, when adding a new user using the admin UI, the admin has to create a user and then go to a second view to set/reset their password. This is a very awkward workflow specially if you want to add many users. Since a username/password are required by default for a user to login, I think the admin should be required to provide a password rather than creating a useless user with no password. Note, the point here is that password is required to login by default without having to enable optional additional authentication providers.

I think we should add a security settings something like "IsPasswordRequired" in addition to [PasswordOptions ](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.identity.passwordoptions?view=aspnetcore-6.0) setting for the user to control. If no external authentication provider is provided, the Password should be required "otherwise the user can't login without password". Now in the [CreatePost](https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore.Modules/OrchardCore.Users/Controllers/AdminController.cs#L319-L345) action we can check if a password is required then validate password and create a user with password. Otherwise, we create the user with no password since password is externally managed. I would think that the controller code will then look something like this

```
public async Task<IActionResult> CreatePost(string password)
{
    var user = new User();

    if (!await _authorizationService.AuthorizeAsync(User, Permissions.ViewUsers, user))
    {
        return Forbid();
    }

    var shape = await _userDisplayManager.UpdateEditorAsync(user, updater: _updateModelAccessor.ModelUpdater, isNew: true);

    if (!ModelState.IsValid)
    {
        return View(shape);
    }

    if (setting.IsPasswordRequiredForNewUsers)
    {
        // I am guessing the password will be passed to Identity for validation before the user is created
        // the ModelState would get updated if the password does not meet the standard defined using PasswordOptions in identity
        await _userService.CreateUserAsync(user, password, ModelState.AddModelError);
    } 
    else 
    {
        await _userService.CreateUserAsync(user, null, ModelState.AddModelError);
    }

    if (!ModelState.IsValid)
    {
        return View(shape);
    }

    await _notifier.SuccessAsync(H["User created successfully."]);

    return RedirectToAction(nameof(Index));
}
```

At this point the process becomes more robust. If a user want to generate a random password, then the password shape will generate a random password or any other password requirement is required. And if the user want to require the user to set new password after the first login, then he/she can create a display driver `DisplayDriver<User>` that would allow them to inject the option and require the user to change their password after they login,

I would like to add "Password" field on the Create User screen and create the new user with a password instead of null.

I think the only way I can do this is to create a new module that depends on `OrchardCore.User` and override the `/admin/user/create` route with my own code. Now sure if this wold even work

CrestApps:
Fix #11080 by adding password field during the creating of new user


src/OrchardCore.Modules/OrchardCore.Users/Views/UserFields.Edit.cshtml
```diff
@@ -1,6 +1,58 @@
...
-@using OrchardCore.Entities
-@using OrchardCore.Users.Models
+@using OrchardCore.Entities;
+@using OrchardCore.Users.Models;
+@using Microsoft.AspNetCore.Identity;
+@using Microsoft.AspNetCore.Html;
+@inject Microsoft.Extensions.Options.IOptions<IdentityOptions> IdentityOptions
+@{
+    var options = IdentityOptions.Value;
+    var passwordOptions = new HtmlContentBuilder();
```
ns8482e, line 9: this code  can be moved to shape

src/OrchardCore.Modules/OrchardCore.Users/Views/UserFields.Edit.cshtml
```diff
@@ -1,6 +1,58 @@
...
+
+    passwordOptions.AppendHtml(separator).AppendHtml(T["{0} characters in total", options.Password.RequiredLength]);
+    passwordRegex += $".{{{options.Password.RequiredLength},}}";
+
+    if (options.Password.RequiredUniqueChars > 1)
+    {
+        passwordOptions.AppendHtml(T[", with {0} unique characters", options.Password.RequiredUniqueChars]);
+    }
+
+    var passwordTooltip = T["Password must have at least {0}.", passwordOptions];
```
ns8482e, line 54: Why this logic needs to be different than ChangePassword screen and Register screen?

src/OrchardCore.Modules/OrchardCore.Users/Views/UserFields.Edit.cshtml
```diff
@@ -14,8 +66,63 @@
...
+            let password = document.getElementById('User_Password');
+            togglePassword.addEventListener('click', function(){
+                toggleField(password);
+            });
+
+        });
+
+    </script>
+
+}
```
ns8482e, line 128: new line

src/OrchardCore.Modules/OrchardCore.Users/Views/UserFields.Edit.cshtml
ns8482e: IMO this should not be part of View - and Let Identity handle the messages as one could implement their own `IPasswordValidator`

ns8482e:
Script changes in assets folder are missing in PR

src/OrchardCore.Modules/OrchardCore.Users/Views/UserFields.Edit.cshtml
CrestApps: I agree with you. I tried to reproduce the code used in [OrchardCore.Setup](https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore.Modules/OrchardCore.Setup/Views/Setup/Index.cshtml#L18-L66) this logic should be extracted into some sort of service. also the `strength.js` resource should be registered separated into resource so it can be reused here too. I just did not want to add many changes here to keep the PR minimal

src/OrchardCore.Modules/OrchardCore.Users/Views/UserFields.Edit.cshtml
CrestApps: fixed

src/OrchardCore.Modules/OrchardCore.Users/Views/UserFields.Edit.cshtml
CrestApps: I changed the view without the logic in the Setup module since the password change views that logic isn't the same

src/OrchardCore.Modules/OrchardCore.Users/Views/UserFields.Edit.cshtml
CrestApps: removed

CrestApps:
> Script changes in assets folder are missing in PR
none is needed

Skrypt:
Just an idea but we could have something to generate a password.

CrestApps:
@Skrypt password generator was added

image

hishamco:
Why it's generated twice? Also we could simplify the UI by adding a checkbox for auto generation, it could save the generated password too at once

CrestApps:
@hishamco its generated once but placed in both fields. When user generates a password, the next logical step in to copy it. Not sure if a checkbox is a good idea here.
Currently we have 2 fields for the user to provide a password with the option to generate one along with the copy button.

ns8482e:
@CrestApps  Either include  changes from corresponding Assets folder otherwise  the changes in `OrchardCore.Media\wwwroot` and `OrchardCore.Setup\wwwroot`

src/OrchardCore.Modules/OrchardCore.Users/Views/UserFields.Edit.cshtml
```diff
@@ -14,8 +14,117 @@
...
+        </div>
+
+        <div class="pt-2">
+            <div class="btn btn-warning btn-sm" id="generatePassword">@T["Generate strong password"]</div>
+            <div class="btn btn-info btn-sm" id="copyPassword">@T["Copy password"]</div>
+        </div>
+    </div>
+
+
+    <script>
```
ns8482e, line 54: use script at="Foot" and `asp-name`

src/OrchardCore.Modules/OrchardCore.Users/Controllers/AdminController.cs
```diff
@@ -316,7 +316,7 @@ public async Task<IActionResult> Create()
 
         [HttpPost]
         [ActionName(nameof(Create))]
-        public async Task<IActionResult> CreatePost()
+        public async Task<IActionResult> CreatePost([Bind(Prefix = "User.Password")] string password)
```
ns8482e, line 319: why `User` prefix?

CrestApps: The `UserDisplayDriver` build the form with the `User` prefix. So in order for the binding engine to bind the password correctly, we have to use the same prefix. Is there a better way to do this?

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
```diff
@@ -55,18 +60,19 @@ public override IDisplayResult Display(User user)
             );
         }
 
-        public override IDisplayResult Edit(User user)
+        public override Task<IDisplayResult> EditAsync(User user, BuildEditorContext context)
```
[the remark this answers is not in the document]
CrestApps: There is no non-async override that would provide you the `BuildEditorContext`. We need access to it to check if the request is for a new user or update of an existing user (i.e., `context.IsNew`)

src/OrchardCore.Modules/OrchardCore.Users/Views/UserFields.Edit.cshtml
CrestApps: done

src/OrchardCore.Modules/OrchardCore.Users/Views/UserFields.Edit.cshtml
```diff
@@ -1,6 +1,6 @@
 @model EditUserViewModel
-@using OrchardCore.Entities
-@using OrchardCore.Users.Models
+@using OrchardCore.Entities;
+@using OrchardCore.Users.Models;
```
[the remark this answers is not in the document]
CrestApps: done

src/OrchardCore.Modules/OrchardCore.Users/Views/UserFields.Edit.cshtml
```diff
@@ -14,8 +14,117 @@
...
+                    return;
+                }
+
+                element.classList.remove('fa-eye-slash');
+                element.classList.add('fa-eye');
+            }
+
+            function toggleFieldType(element) {
+
+                console.log('toggle fields...');
```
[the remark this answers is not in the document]
CrestApps: done

src/OrchardCore.Modules/OrchardCore.Users/Views/UserFields.Edit.cshtml
```diff
@@ -14,8 +14,117 @@
...
+                copyPassword(password.value);
+            });
+
+            passwordToggle.addEventListener('click', togglePasswordFieldState);
+            passwordConfirmationToggle.addEventListener('click', togglePasswordFieldState);
+        }, false);
+    </script>
+}
+
+
```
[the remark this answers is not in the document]
CrestApps: done

CrestApps:
> @CrestApps Either include changes from corresponding Assets folder otherwise revert the changes in `OrchardCore.Media\wwwroot` and `OrchardCore.Setup\wwwroot`

@ns8482e finally removed :)

src/OrchardCore.Modules/OrchardCore.Users/Views/UserFields.Edit.cshtml
```diff
@@ -14,8 +14,97 @@
...
+                passwordConfirmation.value = newPassword;
+            });
+            copy.addEventListener('click', function() {
+                copyPassword(password.value);
+            });
+            passwordToggle.addEventListener('click', togglePasswordFieldState);
+            passwordConfirmationToggle.addEventListener('click', togglePasswordFieldState);
+        });
+    </script>
+}
```
ns8482e, line 110: new line

CrestApps: done.

ns8482e (approved):

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
```diff
@@ -36,7 +39,8 @@ public class UserDisplayDriver : DisplayDriver<User>
             ILogger<UserDisplayDriver> logger,
             IEnumerable<IUserEventHandler> userEventHandlers,
             IAuthorizationService authorizationService,
-            IHtmlLocalizer<UserDisplayDriver> htmlLocalizer)
+            IHtmlLocalizer<UserDisplayDriver> htmlLocalizer,
+            IStringLocalizer<UserDisplayDriver> s)
```
hishamco, line 43: stringLocalizer

hishamco (changes requested):
LGTM after renaming the constructor param

src/OrchardCore.Modules/OrchardCore.Users/Drivers/UserDisplayDriver.cs
CrestApps: done

hishamco (approved):

hishamco:
BTW do really need to copy password?

CrestApps:
@hishamco I think so. When a new password is generated, it’s a random and long password.

PR #11598, opened 2022-04-26
Add an overload to ShapeViewModel

### LINKED ISSUE #11597: ShapeViewModel should have an overload that would accept the ShapeType
I need to create a shape manually using [ShapeViewModel<>](https://github.com/OrchardCMS/OrchardCore/blob/339d74b251b53393dd510f8ad3027aa17d78500d/src/OrchardCore/OrchardCore.DisplayManagement/Views/ShapeViewModelOfT.cs#L5) directly. I had to do that in 2 steps
1. initialize mt shape 
2. Set the Type on the Metadata object.

I think `ShapeViewModelOfT` should have an overload constructor that would allow the consumer to provider a custom `shapeType`. I think we may want to event set the `Metadata.Type` on the current constructor so the Type is set by default.

CrestApps:
Resolve #11597


src/OrchardCore/OrchardCore.DisplayManagement/Views/ShapeViewModelOfT.cs
```diff
@@ -3,6 +3,12 @@ namespace OrchardCore.DisplayManagement.Views
     public class ShapeViewModel<T> : ShapeViewModel
     {
         public ShapeViewModel(T value)
+            : this(typeof(T).Name, value)
+        {
+        }
+
+        public ShapeViewModel(string shapeType, T value)
+            : base(shapeType)
```
ns8482e, line 11: Add param check  should not be null or empty

CrestApps: added in the base class

sebastienros (approved):

PR #10762, opened 2021-11-27
OrchardCore.Cms.Web, Mvc.Web and Templates with .NET 6

agriffard:
Use only `Program.cs`

Options in csproj:
```
    <Nullable>enable</Nullable>
    <ImplicitUsings>enable</ImplicitUsings>
```

Docs updated.
Templates updated.

Fixes #11932

PR #11604, opened 2022-04-27
UseNLogHost() to be used on IHostBuilder

jtkech:
For example with the minimal hosting model in 6.0.

    ...
    using OrchardCore.Logging;

    var builder = WebApplication.CreateBuilder(args);
    builder.Logging.ClearProviders();
    builder.Host.UseNLogHost();
    ...


agriffard:
Now that it is merged, we need to also change the Cms.Web template or merge #10762 for release 1.4

Issue #11422, opened 2022-03-22
Lucene: Document contains at least one immense term in field


devlife:
### Describe the bug
Getting the below `ArgumentException` when trying to reset/rebuild a lucene index where one of the content items has very long text.

```
ArgumentException: Document contains at least one immense term in field="BlogPost.HtmlBody" (whose UTF8 encoding is longer than the max length 32766), all of which were skipped. Please correct the analyzer to not produce such terms. The prefix of the first immense term is: '[3c 64 69 76 20 63 6c 61 73 73 3d 22 69 6e 73 69 67 68 74 2d 61 72 74 69 63 6c 65 73 2d 77]...' at Lucene.Net.Index.DocInverterPerField.ProcessFields(IIndexableField[] fields, Int32 count)
```

### To Reproduce
1. Create a site for the blog theme
2. If one does not already exist, create an index for `Blog Post`
3. Create a query which will use the index in step 2
```
{
    "query": {
          "bool": {
                  "must": [
                          {
                              "match": {
                                  "BlogPost.Tags": {
                                      "query":"{% for tag in tags %}{{tag}} {% endfor %}",
                                      "operator": "or"
                                  }
                              }
                          },
                  ],
              }
    },
    "sort": {
      "BlogPost.PostedDate": {
        "order": "desc",
        "type": "double"
      }
  	}
}
```
5. Make sure you have an extremely long blog post
6. Click Reset or Rebuild on the index and see the error

### Expected behavior
I would like to know if there is any way around this issue and/or if there is a way to identify which blog post or posts are responsible for this.

Skrypt:
Of course, everything has its limits. Maybe we need to use a TextField here instead of a StringField. But, of course, if you change a Field from one type to the other then you need to rebuild the index so that it uses the proper Lucene Type in the Lucene document.

devlife:
@Skrypt how do I go about using a TextField? Are you talking about changing the field type of the HtmlBody?

sebastienros:
When we index the value in lucene we need to validate the length. I am very surprised that is a limit.
If there is really one then we can split it into multiple entries on the same field.

devlife:
The current type is `HtmlField`. To be clear, are you saying I need to convert it to a `TextField`? I don't think the client would want to write blog posts in plain text.

sebastienros:
[...]

PR #11630, opened 2022-05-02
Add multiple documents for lucene indexing if HtmlField.Html is longer than max length

devlife:
Add multiple documents for lucene indexing if HtmlField.Html is longer than max length.
Covers request in issue https://github.com/OrchardCMS/OrchardCore/issues/11422


src/OrchardCore.Modules/OrchardCore.ContentFields/Indexing/HtmlFieldIndexHandler.cs
```diff
@@ -1,18 +1,25 @@
...
             var options = context.Settings.ToOptions();
 
             foreach (var key in context.Keys)
             {
-                context.DocumentIndex.Set(key, field.Html, options);
+                field.Html.Chunk(maxStringLength).Select(chunk =>
+                {
+                    context.DocumentIndex.Set(key, new string(chunk), options);
+                    return Task.CompletedTask;
+                });
```
sebastienros, line 22:
```suggestion
                field.Html.Chunk(maxStringLength).Select(chunk =>
                    context.DocumentIndex.Set(key, new string(chunk), options)
                );
```

src/OrchardCore.Modules/OrchardCore.ContentFields/Indexing/HtmlFieldIndexHandler.cs
```diff
@@ -1,18 +1,25 @@
...
+﻿using System.Linq;
+using System.Threading.Tasks;
 using OrchardCore.ContentFields.Fields;
 using OrchardCore.Indexing;
 
 namespace OrchardCore.ContentFields.Indexing
 {
     public class HtmlFieldIndexHandler : ContentFieldIndexHandler<HtmlField>
     {
+        private const int maxStringLength = 32766;
```
sebastienros, line 10:
```suggestion
        private const int MaxStringLength = 32766;
```

devlife:
@sebastienros I committed your suggested changes.

sebastienros:
Does it solve the problem?

Also this is for Html field only, the same issue should appear with TextField, HtmlBodyPart and MarkdownBody

devlife:
> Does it solve the problem?
I will test this later today.
> Also this is for Html field only, the same issue should appear with TextField, HtmlBodyPart and MarkdownBody
Assuming it works, I'll make the same change for those fields as well.

dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=11630) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=11630) before we can accept your contribution.<br/><br/>:x: devlife [sign now](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=11630)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OrchardCMS/OrchardCore?pullRequest=11630) it.</sub>

devlife:
Ah I see. I'll try deleting all indexes and content under the lucene folder and try again. Otherwise I'll look into the unit test.

devlife:
The plot thickens! I was able to repro after unchecking `Stored` and `Analyzed` on the field. I'm not entirely sure what these two fields do but this is the error message I get with the original code:

```
ArgumentException: Document contains at least one immense term in field="BlogPost.Html" (whose UTF8 encoding is longer than the max length 32766), all of which were skipped. Please correct the analyzer to not produce such terms. The prefix of the first immense term is: '[3c 64 69 76 20 63 6c 61 73 73 3d 22 69 6e 73 69 67 68 74 2d 61 72 74 69 63 6c 65 73 2d 77]...'
```

Adding in the new code results in the index being rebuilt successfully. I'll look into adding one or more unit tests.

Skrypt:
Would make sense that an Analyzed field ends up with less content since it will also remove non-meaningful words like "a, at, the ... or".

devlife:
@Skrypt There also is a lot of markup in this field. If it removes "a, at, the ... or" does it remove markup as well?
@sebastienros I can test more thoroughly with `analyzed` and `stored` and report back.

Skrypt:
"Stored" should store "as is". But Analyzed should be removing HTML markup since it is not relevant. If it doesn't then there might be an issue that needs to be fixed. I don't remember all the details but normally it should be.

devlife:
I confirmed that the issue is only occurring when `Stored` is checked. The length of the field is 281384 chars.

Skrypt:
Here, as a reference, we sanitize when "Analyzed" on an HtmlBodyPart.

https://github.com/OrchardCMS/OrchardCore/blob/d138a9b94e013bf724ac9f47c7bbd265fd938599/src/OrchardCore.Modules/OrchardCore.Html/Indexing/HtmlBodyPartIndexHandler.cs#L12

devlife:
Should I make the update only when `Stored` is selected, then?

Skrypt:
No, I think it is safer to have it for both.

devlife:
So is there anything else you're waiting on? Just a unit test?

sebastienros:
I'd like you to confirm something. The issue only happens when "Stored" is selected. The intent of this property is to be able to read the original value that was indexed. In that case can you confirm that splitting a big stored value doesn't impact the original value when you try to read it again?

Can be a unit test if you are able to do so. Should be easy if there is a unit test that checks for "stored" values already.

devlife:
@sebastienros It looks like the recipe being used by the tests is using `Stored` on `HtmlBodyPart` as seen [here](https://github.com/OrchardCMS/OrchardCore/blob/main/test/OrchardCore.Tests/Apis/Lucene/Recipes/luceneQueryTest.json#L59). If that's the case, then I should be able to make [this](https://github.com/OrchardCMS/OrchardCore/blob/main/test/OrchardCore.Tests/Apis/Lucene/Recipes/luceneQueryTest.json#L102) content very long to properly test this change. I would actually just add another article to the recipe.

However, to test against the stored value, are you saying that if the original value is split on the word "apple", for example, so that the first document ends in "ap" and the second document begins with "ple" that we would simply need to make sure that we get back the result if we search for "apple" assuming that word only appears in that one specific place in the document?

If that is your intent then I should be able to do that this week.

sebastienros:
Yes, if the document is split, we should still be able to read the whole original content.
To test it, index such a big content, even without a recipe (in the UI), and in a custom page try to query the index and render the "stored" content.

devlife:
@sebastienros @Skrypt Unfortunately this does not work. I added a very long string in a blog post where the document gets split on the word "washington" with the first chunk ending in "wa" and the second chunk starting in "shington" but the query does not return any results.

This returns no results:
```json
{
  "from": 0,
  "size": 1,
  "query": {
    "simple_query_string": {
      "analyze_wildcard": false,
      "fields": [
        "BlogPost.Html"
      ],
      "query": "washington"
    }
  }
}
```

This returns the expected result:
```
{
  "from": 0,
  "size": 1,
  "query": {
    "simple_query_string": {
      "analyze_wildcard": false,
      "fields": [
        "BlogPost.Html"
      ],
      "query": "wa"
    }
  }
}
```

This returns the expected result:
```
{
  "from": 0,
  "size": 1,
  "query": {
    "simple_query_string": {
      "analyze_wildcard": false,
      "fields": [
        "BlogPost.Html"
      ],
      "query": "shington"
    }
  }
}
```

devlife:
@jtkech that's an interesting thought. We could grab the default chunk and then back track to the previous white space. Although, would that then prevent someone from searching for something like "washington*state" or something. Or perhaps we can do chunk overlap? Have the chunks overlap by, say, 100 characters?

devlife:
@jtkech @Skrypt @sebastienros Is that an approach you'd like to take?

Skrypt:
I will try to test this. Also, I'm not 100% sure about the query you used to test this. Normally a simple wildcard query should be enough to see if it will return any result.

Skrypt:
Elasticsearch doesn't like when we pass multiple documents for a single field. I needed to pass the FullText as a single string for it to work.

devlife:
@Skrypt I'd be happy to try a query you supply. I just grabbed an existing query from the unit tests.

devlife:
@Skrypt Where do we go from here? My client is going live in January and I'd love to get this fixed before then if possible.

Skrypt:
I will need to test it with both Lucene and Elasticsearch enabled at the same time. I will try to make it a priority.

devlife:
Great. Please let me know if there is anything I can do to help.

sebastienros:
> like "washington*state" 

You can also split on the last non-alphanumeric char.

Skrypt:
Why did it get removed from 1.6?

sebastienros:
Not blocking.

sebastienros:
Actually the PR was removed, not the issue

sebastienros:
@Skrypt do you want to update the PR to remove the methods we mentioned and use substring?

Skrypt:
I can't say when I will have the time to work on this. I'm already in a 2 weeks sprint to get things done at work.

devlife:
I'm happy to help if you can tell me what needs to be done.

devlife:
My client is going live in March, I believe and still experiencing this issue. So anything I can do to move the process forward please let me know.

image

sebastienros:
@devlife I made some comments a while ago that could be addressed. I see them as pending still, can you see that?

jtkech:
I would like to help you but I'm not aware of the last triage on this PR, maybe, as suggested by @sebastienros, we could ensure that we split on a non alphanumeric char to not break any word, maybe use `subString()` and for example with `StringSegment` to reduce allocations ...

If we can have more info I will try to help you ;)

devlife:
> You can also split on the last non-alphanumeric char, and non-whitespace
That is all that I see. If there are other things I can give those a try. I was not able to get a successful unit test. The code ran but was unable to find the value if it was split across multiple documents.

jtkech:
@devlife 

> I confirmed that the issue is only occurring when Stored is checked

Naive question ;) Is it necessary for you to use `Stored` ? Is it possible to use `Analyzed` instead?

devlife:
@jtkech I'm not really sure what the difference is. As long as the blog post body can be both searched and returned I'm good. I guess if that isn't possible I could use two search functions. One to get the post ids and then get the posts via C# instead of lucene.

jtkech:
> As long as the blog post body can be both searched and returned I'm good

Can you try it?

We did a dedicated search engine for a customer with dedicated filters / analyzers / fuzzy search ... but not an expert on Lucene/Elastic themselves ;)

Unless @Skrypt has time to look at it, I will try to work on it this week end

jtkech:
Okay I understand

> Stored => Check to be able to retrieve the **original value from the index**.

jtkech:
@devlife 

Just started some simple tests by creating 2 Articles with large bodies as in your test.

Okay, it doesn't work if I check `Stored` in the `HtmlBodyPart` settings, but if I let it unchecked the following query works

    {
      "query": {
        "simple_query_string" : {
            "query": "\"washington\"",
            "fields": ["Content.BodyAspect.Body"]
        }
      }
    }

Then through the Admin, from the above I created a lucene query named `LargeBody` where I checked the `Return Content Items` checkbox. Then in one of the liquid file I used this query and was able to render the full items, meaning that the query was returning full content items.

    {% assign largeBodies = Queries.LargeBody | query %}
    {% for item in largeBodies %}
    {{ item | shape_build_display | shape_render }}
    {% endfor %}

Hmm, make sense that too long string should not be stored as is, but in separate terms (maybe normalized and after having been sanitized). Hmm, after a sanitization the whole is reduced, so just tried by copying the body that you are using in your test, and then pasted it 10 times in the body of the items I'm testing, the term washington being present 10 times per item, it still works.

jtkech:
@devlife 

Wait wait, in fact when I checked `Stored`, first I didn't have any exception, then when testing the query in the Admin I thought it was only returning only one item of the both I was testing, in fact because on the left there was a very long body field, I was not seing the second item but it was there.

So I updated my `LargeBody` Lucene query by unchecking the `Return Content Items` checkbox.

And then used the following in a liquid file, just using the search result objects

    {% assign results = Queries.LargeBody | query %}
    {% for result in results %}
    {{ result }}
    {% endfor %}

And it was working as expected, so with bodies 10 times longer than the one we use here in the tests.

jtkech:
@devlife 

Okay, I could get the `Document contains at least one immense term in field="Article.HtmlBodyPart"` but only if I also check the `Keyword` checkbox for the `HtmlBody`.

    `Keyword` checkbox => Check to index as a keyword value (indexed as a single token).

jtkech:
@devlife 

I'm using the main branch, sorry I thought this PR was in sync with the main branch.

Yes I remember @Skrypt said to me that the UI would need to be changed, as I remember because an option may be used even if not checked if another option is checked, or something like that. So yes, if you only use `Analyzed` with your version, maybe quite the same as only using `Stored` with the new UI.

Here the options I used, only `Stored`, and you can see that the new `Keyword` option is now available.

Index-options

jtkech:
So maybe we can close this PR, @devlife let me know

Skrypt:
By default, the Field is set to "Analyzed" when we check the "Include this element in the index". So, before my changes it was the opposite, it was indexed as a "Keyword". So, I reversed this because it makes more sense that people want to analyze first and specify explicitly when they want to have a field used as a technical term (keyword).

devlife:
@jtkech @Skrypt That's great! I guess I'll leave it up to you if we want to proceed with the PR to prevent that exception or add additional logging or anything else. For my specific case, it appears to be handled by using `Analyzed` only.

Skrypt:
Yes, that makes sense because everything stored is now tokenized which will take less space when it removes extra words that are not indexed. Also, I guess sanitizing the HTML would prevent this from happening to some degree. Though, I agree that we keep the changes and move forward with it because we never know when we will hit the limit. This change should prevent this from happening. I need to put my head around the details but I trust @jtkech if he wants to take a shot at it. 😄

jtkech:
@Skrypt 

I think all is okay because in my tests I still used `Stored` (and only) and with a body 10 time latger that the one used for testing in this PR. I only got the excetion if `Keyword` is checked, which make sense for ma as it means indexed as a single token.

Maybe before if only `Analyzed` => still `Stored` but not indexed as a single token as we can now do by letting `Keyword` unchecked.

I may be wrong but for me all seems good and we could close this PR, if you agree ;)

Skrypt:
Well the idea is to prevent the exception. So, maybe we should prevent from indexing a Keyword that is too large. We need to fix a limit on it just like Elasticsearch does. That, would make sense.

jtkech:
Okay then

Skrypt:
https://stackoverflow.com/questions/47177163/what-is-the-maximum-length-for-keyword-type-in-elasticsearch

jtkech:
But then maybe a simpler check, we could just shrink the value and add a notification/log message, hmm but as Lucene already does. I let you see what is better to do, here I only suggested my help because it was "blocking" the 1.6 release ;)

Skrypt:
The difference with Elasticsearch is that there is a setting for Keyword max-length set by default on the index. Which is normally 255 chars. But here, we don't have mappings with Lucene so I guess we could add a setting to the Index and leave it to people decide the max-length by default used on that index.

jtkech:
For me it's no longer an issue, Lucene already shrink the value an log a warning.

Yes a setting is a good idea.

Skrypt:
We can close the issue then. Open a ticket for adding a setting on the max-length of a Keyword and done! 😄

jtkech:
Okay I'm closing it, @devlife when you will have time can you open a new issue with the enhancement label, you will better describe the goal of that new size limit setting than me ;)

devlife:
Awesome thanks very much for your help guys!

PR #11640, opened 2022-05-04
Translate ReCaptcha TagHelper

Skrypt:
The ReCaptcha TagHelper is currently always using the tenant default culture. By using the "language" param of the TagHelper we can then use the current request culture.


hishamco:
Looks good, can you share a screenshot to see how the translation is look like

Issue #11667, opened 2022-05-09
Disabling a user should invalidate the users security stamp


deanmarcussen:
Disabling a user should invalidate the users security stamp

When the security stamp validator revalidates the cookie, the user will then be expired.

PR #11668, opened 2022-05-09
Set security stamp when user is disabled

deanmarcussen:
fixes https://github.com/OrchardCMS/OrchardCore/issues/11667


sebastienros:
@hishamco usually it means there is a circular dependency and we need to use the lazy pattern

PR #11680, opened 2022-05-10
Add OrchardCore.ContentField.Core project

### LINKED ISSUE #11678: Create OrchardCore.ContentField.Core project
In many cases I create "ProjectName.ModuleName.Core" project where I use `ContentFields` since this is a core modules I don't really need anything more than the class derived from `ContentField` and the corresponding settings. So why do I need to depend on `OrchardCore.ContentFields` in my core project.

I think we should introduce a project called `OrchardCore.ContentField.Core` or `OrchardCore.ContentField.Abstractions` where we move all the fields and setting and extensions classes into.

CrestApps:
This pull request Fix #11678 

Do we need to move the `setting` classes along with the view-models to this project too? Or should we just leave them where there are at? Perhaps, we can place the fields into `.Abstraction` project where the fields would live while settings and viewmodel would live in `.Core` project.


sebastienros:
I assume this wouldn't be a breaking change since the module would still reference the Core module?

CrestApps:
@sebastienros yes this should not break anything since all namespaces are the same the the OrchardCore.ContentFields still depends on projects that has all the dependencies.

@Skrypt I just made a change where I moved all the field and settings models to a new `.Abstractions` project and moved services to the `.Core` project. At this point, I am not sure if we need `.Core` project but may not be a bad idea to keep it. With this change any user that wants to use fields in their module, they would only install `OrchardCore.ContentFields.Abstractions` this would provide the content fields model and the settings needed.

hishamco:
Please fix the conflict

src/OrchardCore.Modules/OrchardCore.ContentFields/OrchardCore.ContentFields.csproj
```diff
@@ -16,6 +18,8 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\OrchardCore\OrchardCore.Admin.Abstractions\OrchardCore.Admin.Abstractions.csproj" />
+    <ProjectReference Include="..\..\OrchardCore\OrchardCore.ContentFields.Abstractions\OrchardCore.ContentFields.Abstractions.csproj" />
```
hishamco, line 21: We need to reference the `OC.ContentFields.Core` ONLY

src/OrchardCore/OrchardCore.ContentFields.Abstractions/OrchardCore.ContentFields.Abstractions.csproj
```diff
@@ -0,0 +1,24 @@
...
+
+  <PropertyGroup>
+    <RootNamespace>OrchardCore.ContentFields</RootNamespace>
+
+    <!-- NuGet properties-->
+    <Title>OrchardCore ContentFields Abstractions</Title>
+    <Description>
+      $(OCCMSDescription)
+
+      Abstractions for ContentFields Module.
```
hishamco, line 11: Abstractions for ContentFields

hishamco (changes requested):

Skrypt:
Please fix conflicts by merging latest main branch.

CrestApps:
@Skrypt I am not sure how to do this one. If this is something we are planing on merging I can just create a new PR for this.

image

hishamco:
No need to create another PR, try to rebase the `main` branch

CrestApps:
@Skrypt 

Closing this PR in favor of PR #11990.

CrestApps:
@hishamco I tried but could not pull it off. I already pushed a cleaner PR as a replacement to this.

PR #11684, opened 2022-05-12
Add OC version to X-Powered-By http header

hishamco:
PHP for instance adding a version alongside with the product name, I think it would be nice to add the OC version


sebastienros:
We have decided not to display the version when we created the header. Some people are already scared by the header, adding the version would just have more people remove it.

sebastienros:
The header could inform an attacker that a website is vulnerable because it's using OC. Showing the version might let them know a vulnerability exists on this site. IMO it doesn't make it safer to "not tell" which version it is, but some security audits recommend to remove it. I want to give fewer reasons as possible for people to remove the header completely.

hishamco:
> I want to give fewer reasons as possible for people to remove the header completely.

The ASP already added that header too, I will close this, while a decision made at the beginning

PR #11699, opened 2022-05-13
Cleanup markdown by moving models to the Abstractions project

### LINKED ISSUE #11698: Move MarkdownField and MarkdownFieldSettings to the OC.MarkdownField.Abstractions project
I think both `MarkdownField` and `MarkdownFieldSettings` classes should be moved to the OrchardCore.Abstractions project to allow other projects to use them without the need to install `OrchardCore.Markdown` project.

MikeAlhayek:
Fix #11698


Skrypt (approved):

hishamco:
Can anyone give me  proper reason why we need this before we merge? https://github.com/OrchardCMS/OrchardCore/issues/11698#issuecomment-1179579984

CrestApps:
If you are creating an abstraction or core project where you define your models, while the model is using MarkdownField, there is no need to add dependency on the OC.Markdown on the core or abstraction model .

for example, I want to create a custom model in my project called xyz. I created xyz.Core or xyz.Abstraction where I keep a model that uses MarkdownField. In this case I won’t have to depend on OC.Markdown, instead all I would add is a reference to the abstractions

hishamco:
Still I didn't understand, when you need a `MarkdownField` you should reference it's project. If that's the case `HtmlField` should have an abstractions

Again I didn't see a value unless the implementation should be different for a particular abstractions project

hishamco:
I got what you want to explain, but this something special in your case, If we want plan to accept this, what about `HtmlField` and other fields?

CrestApps:
IMO, I think all field ContentFields, Markdown, MediaField, HtmlField all should be moved to their abstraction project. All these models and their settings can be used in Abstraction and Core project.

hishamco:
If it's like that do we really treat fields as abstractions? Which I don't think so, let us hear @sebastienros thought about this one?

Skrypt:
It is just about reusability/extensibility. If you want to use a NumericField in a different module right now you need to reference the entire OrchardCore.ContentField project which is wrong to me. And yes, ContentFields needs to be reusable.

Skrypt:
Yes, I agree with you @hishamco but it is part of the tasks that we need to do to make OC more like a Framework. And these tasks will take time. It is probably better to do the work progressively.

hishamco:
@CrestApps just one question if we use something like this

```csharp
public MarkdownField Content { get; set; }
```
How the UI will be generated for such field in the referenced module?!!

CrestApps:
@hishamco the field will be added to a model to a `CustomModule.Abstraction` project. So it will not need to render the field. This would just be part of abstraction/core project. The module that would consume this project would then need to depend on the Markdown feature.  The idea here is structuring the code is much reusable way.

hishamco:
> I guess the question is, do you see an issue with abstracting the model and the setting?

Nope, but wondering while any field should has its own UI, so is that mean any implementation should has to redefine the UI. I never mind if all the guys are agreed with the idea

CrestApps:
@hishamco another good reason for this is to prevent the use of internal classes.

For example, if you look at the [Taxonomies migration](https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore.Modules/OrchardCore.Taxonomies/Migrations.cs#L167-L176) you'll see internal classes AliasPartSettings and AutoroutePartSettings created as shadow to the actual settings class, just because AliasPartSettings and AutoroutePartSettings are not part of the OrchardCore.Alias.Abstractions or OrchardCore.Autoroute.Core. The problem with creating internal models like that is what if the actual setting model change? I see no reason for doing such an approach “creating shadows classes” where we can simply move the model and setting into the abstraction class and everything will just fall into place as it should.

The Markdown and other modules falls into the same problem as TaxonomiesMigration. In fact settings and models classes are already part of many abstraction projects in OC. So at the very least there is a consistency problem with not moving the model and the settings into the abstraction project. 

I think abstraction project should contain, interfaces, core/domain models including setting models.

Core project will include implementations and services and will depend on abstraction project. Then, the CMS module project will have everything else and will depend on the core project.

PR #11705, opened 2022-05-15
Fix message view to avoid exception

CrestApps:
In some case, the `Message.cshtml` view return error when Model.Type is null. This view added a null check before utilizing the property to avoid exception


src/OrchardCore.Themes/TheAdmin/Views/Message.cshtml
```diff
@@ -1,4 +1,8 @@
 @{
+    if (Model.Type == null)
```
hishamco, line 2: IMHO I think we know what's the reason that make it `null` instead of `return`

hishamco (changes requested):

src/OrchardCore.Themes/TheAdmin/Views/Message.cshtml
CrestApps: I don't think I understand the ask.

hishamco: Sorry I miss "need to" ;)

CrestApps: I don't know the real reason why it becomes null. but I have encounter the null reference exception couple times and I think a null check is a valid catch to prevent the exception.

hishamco: Please steps to reproduce, otherwise we cann't accept this as a proper solution

jtkech: Hmm, normally the model is built from a `NotifyEntry` and `Type` is an **enum field**.

It's like a "Message" shape whose dynamic model would not be built from a `NotifyEntry`.

src/OrchardCore.Themes/TheAdmin/Views/Message.cshtml
```diff
@@ -1,4 +1,8 @@
...
+    {
+        return;
```
jtkech, line 4: Can we directly return from a piece of razor code? Never did this ;)

CrestApps: I guess we learn something new everyday :)   I use return in view which help clean up logic a bit

src/OrchardCore.Themes/TheAdmin/Views/Message.cshtml
CrestApps: Not sure why it happens for me every now and then. Next time it happen, I’ll try to see if I can come up with steps to reproduce

hishamco:
I think we should close this PR while there's no steps to reproduce the issue, moreover it doesn't resolve the issue properly

PR #11538, opened 2022-04-14
Add Security Module  (Lombiq Technologies: OCORE-91)

hishamco:
Fixes #854

- [x] Strict-Transport-Security
- [x] Content-Security-Policy
- [x] ~~X-Frame-Options~~ (Obsoleted by `frame-ancestor` in CSP)
- [x] X-Content-Type-Options
- [x] Referrer-Policy
- [x] Permissions-Policy

PR #11706, opened 2022-05-16
Work on security module

jtkech:
Related to and target #11538 

Oh yes, I saw that we don't edit the `report-uri` policy, maybe intended


hishamco:
> Oh yes, I saw that we don't edit the report-uri policy, maybe intended

Right, I remembered there's a reason that prevent me to add it, may be it's experimental

PR #11714, opened 2022-05-17
Fix content item display coordinator when a GroupId is provided

### LINKED ISSUE #11718: When building a ContentItem Editor using GroupId, the parts driver never get called in 1.4-preview
### Describe the bug
When using `IContentItemDisplayManager` build-editor for a specific GroupId, the drivers for each of the corresponding part never gets called. **This was not an issue in version 1.3 however it is an issue in 1.4-preview.**

### To Reproduce
1. Create a content part that generates special result when the `content.GroupId` is set to "test"
2. Create a content type and attach the content part you previously created
3. Now create a new controller to render the content-items under a specific GroupId
4. in the controller caller the driver like so

```
_displayManager.BuildEditorAsync(contentItem, _accessor.Updater, true, "true", "some-prefix")
```

### Expected behavior
The driver for each attached part would get called. Its then the driver responsibility to return something, nothing or something different when the groupId is set.

CrestApps:
The `ContentItemDisplayCoordinator` has an issue where it was not rendering the parts on Edit when a groupId is provided. Explained in #11718 

This PR calls `shapeResult.OnGroup(context.GroupId);` to ensure the shape is rendered with a groupId is provided fixed #11718

Also, slightly cleans up the code.


sebastienros:
Hard to understand the changes since there are indentations changes that mess-up with the diff.

ns8482e:
just looked at following loop should continue and not retrun

```cs
  if (partTypeName == partName)
                        {
                            return;
                        }
```

Issue #11717, opened 2022-05-17
OpenId Server Hybrid Flow does not work


MichaelPetrinolis:
### Describe the bug
When you enable Hybrid Flow on the server and configure a client to use Hybrid Flow, it does not work.

### Expected behavior
The OpenId Application should persist Hybrid Flow configuration and the client should be able to connect

kevinchalet:
> @kevinchalet I could make a PR and remove the HybridFlow when configuring server (mentioning in hint that Hybrid flow will be available for applications if both basic and implicit flows are enabled)

This was the logic used in OpenIddict 1.x/2.x: for the hybrid flow to be considered allowed, it was enough to enable both the implicit and authorization code grants (the hybrid flow doesn't have a dedicated `grant_type` so it's often treated as a combination of both).

OpenIddict 3.x introduced "response type permissions" to make things much more granular and it seems I missed a few things when I updated the OpenID module to support them.

> and allow configuration of the Response Types CodeToken, CodeIdToken and CodeIdTokenToken for the application, if you agree.

Making the allowed `response_type` combinations is something we decided against to keep things simpler for the users (instead, we just have checkboxes for the different flows)

> The current code enables CodeToken and CodeIdTokenToken only for public clients. Is it because they get the access_token through the front_channel ?

Public clients can get a backchannel access token by making a token request too. That check is actually there to *prevent* confidential clients (or a malicious actor trying to downgrade the `response_type`) from getting a frontchannel access token without having to authenticate first. It's not prohibited by the spec, but it's a security design decision we made for OC :smile:

kevinchalet:
@MichaelPetrinolis if you're interested, there's another issue to fix with the recipe, that doesn't correctly remove permissions on updates: https://github.com/OrchardCMS/OrchardCore/issues/11720 😃

MichaelPetrinolis:
Thanx for the clarifications @kevinchalet , I will check  the other bug.

PR #11719, opened 2022-05-17
Fix various issues with the hybrid flow

kevinchalet:
Fixes https://github.com/OrchardCMS/OrchardCore/issues/11717.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/ApplicationController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/ApplicationController.cs
@@ -204,9 +204,25 @@ namespace OrchardCore.OpenId.Controllers
 
             var model = new EditOpenIdApplicationViewModel
             {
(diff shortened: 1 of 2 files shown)


kevinchalet:
@MichaelPetrinolis when you have a moment, could you please give it a try? 😃

MichaelPetrinolis:
@kevinchalet tested the scenarios I used and works as expected, thanx! 
I also verified that when we change the response type to codetoken or codeidtokentoken for a client on a confidential app, the idp does not allow it. I will open a a bug for the OpenId client, because when we change the application from confidential to public(to check the other two response types), we cannot clear the client secret in the client configuration and the idp rejects the request because client_secret is not expected.

sebastienros:
Now it's conflicting ...

kevinchalet:
> Now it's conflicting ...

That's what happens when you merge PRs in the wrong order... :trollface: 

I'll rebase it and update @MichaelPetrinolis' new extension to use the same fix.

sebastienros (approved):

PR #11721, opened 2022-05-17
Use extension method to update OpenId application descriptor

### LINKED ISSUE #11720: Fix OpenIdApplicationStep to remove permissions for flows that are no longer enabled
https://github.com/OrchardCMS/OrchardCore/pull/10136 updated `OpenIdApplicationStep` to support updates but unfortunately, permissions removal was not implemented: if you remove an allowed flow from an already configured recipe, the permissions corresponding to that flow (grant type permissions, endpoint permissions and response types permissions) must be removed too (otherwise, the flow is still considered allowed).

Basically, the same logic present in `ApplicationController.Edit()` should be used in the recipe step.

MichaelPetrinolis:
reuse logic used in ```ApplicationController.Edit()``` also in ```ApplicationController.Create()``` and ```OpenIdApplicationStep```

closes #11720

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/ApplicationController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/ApplicationController.cs
@@ -157,91 +157,27 @@ namespace OrchardCore.OpenId.Controllers
                 return View(model);
             }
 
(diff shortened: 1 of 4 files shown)


MichaelPetrinolis:
I don't now why the tests fail, my changes should not be the reason.

PR #11666, opened 2022-05-09
bootstrap-select 1.14.0-beta3

agriffard:
https://www.npmjs.com/package/bootstrap-select/v/1.14.0-beta3

PR #11723, opened 2022-05-18
Revert Bootstrap-Select to beta 2

Skrypt:
Full Gulp Rebuild. Update Sortable.js to 1.15.0

Closes #11415

Beta 3 has issues related with doing `element.selectpicker('refresh')` which will duplicate elements instead of updating the list.

See comments added here : https://github.com/OrchardCMS/OrchardCore/pull/11666#issuecomment-1129331993


Skrypt:
Doing a gulp rebuild command relevated an issue that we have when using package.json dependencies with versions that uses ^

Here sortable.js version is "^1.14.0" which will get updated to 1.15.0 anytime we drop the package.lock.json file to update also its dependency versions. If we want to have reproducible builds in this case we are forced to also update sortable.js version unless we never use these kind of versions in the package.json files.

PR #11726, opened 2022-05-18
Add Automatically QueryString Params to PagerSlim

### LINKED ISSUE #10384: Clicking older or newer posts button not show list of blog post but show all categories instead
### Describe the bug
- On blog posts in a category page, if there are pagination buttons, clicking older or newer posts button bring users to all categories page which is not expected behavior.
- Given we have 6 blog posts and set page size to 2.
- We have 3 pages (2 post items per page)
- We can't get 3rd page by clicking **Older Posts** button but we get all categories page instead.
- I am not sure if it happens when we are about to go to first or last page only.

### To Reproduce
- Setup or new Orchard Core CMS website with a blog recipe or use my existing demo website:
  - https://orchard-core-demo.azurewebsites.net/
  - username: admin
  - password: 12345Abc$
  - source code https://github.com/aaronamm/OrchardCoreDemo
  - Orchard version 1.0
  - database SQLite
- Login in as admin.
- Go to blog menu.
- Clone an existing blog post to get 6 post items in total.
- Publish all content items.
- Go to Configuration > Settings > General
- Change page size to **2** to get **3 pages** on frontend.
- Go back to home page and browse all blog posts in travel category (e.g. https://orchard-core-demo.azurewebsites.net/categories/travel)
- You will find a simple pagination with **Older Posts** button.
- Click **Older Posts** button to go to 2nd page.
- Click **Older Post** button again to go to 3rd page.
- **You will be redirect to all categories page that list all categories (e.g. https://orchard-core-demo.azurewebsites.net/categories) which is not what we want.**
- On a page that has a Newer Posts button, try to click it and you will get the same behavior.

### Expected behavior
- Show 3rd page and 1st page instead of showing all categories page
- Do not show "Older Posts" or "Newer Posts" button if there is no blog post, I think it is existing  behavior.

### Screenshots

blog posts in a category page

1st page
image

2nd page
image

Not go to 3rd page but get all categories page instead

image

Skrypt:
Fixes #10384 

https://user-images.githubusercontent.com/3228637/169114830-73b4ad5a-38d5-4074-bd8e-540887e38fb7.mp4


src/OrchardCore.Modules/OrchardCore.Navigation/PagerShapes.cs
```diff
@@ -398,14 +373,23 @@ public Task<IHtmlContent> Pager(Shape shape, DisplayContext displayContext)
...
-                    routeData.Add(item.Key, item.Value);
+                    if (!routeData.ContainsKey(item.Key))
+                    {
+                        routeData.Add(item.Key, item.Value);
+                    }
+                    else
+                    {
+                        // We update the value since it has been explicitly passed on the Pager
+                        routeData[item.Key] = item.Value;
+                    }
```
sebastienros, line 392:
```suggestion
                    routeData[item.Key] = item.Value;
```

src/OrchardCore.Modules/OrchardCore.Navigation/PagerShapes.cs
```diff
@@ -431,6 +407,13 @@ public Task<IHtmlContent> Pager(Shape shape, DisplayContext displayContext)
                 await shape.AddAsync(previousItem);
                 shape.Properties["FirstClass"] = PreviousClass;
             }
+            else
+            {
+                if (routeData.ContainsKey("before"))
+                {
+                    routeData.Remove("before");
```
jtkech, line 414: The route dictionary checks for you if it contains the key.

src/OrchardCore.Modules/OrchardCore.Navigation/PagerShapes.cs
```diff
@@ -454,6 +437,13 @@ public Task<IHtmlContent> Pager(Shape shape, DisplayContext displayContext)
                 await shape.AddAsync(nextItem);
                 shape.Properties["LastClass"] = NextClass;
             }
+            else
+            {
+                if (routeData.ContainsKey("after"))
+                {
+                    routeData.Remove("after");
```
jtkech, line 444: Idem

src/OrchardCore.Modules/OrchardCore.Navigation/PagerShapes.cs
```diff
@@ -398,14 +373,15 @@ public Task<IHtmlContent> Pager(Shape shape, DisplayContext displayContext)
...
+
+            var routeData = GetRouteData(shape, displayContext, Html);
 
             // Allows to pass custom url params to PagerSlim
             if (UrlParams != null)
             {
                 foreach (var item in UrlParams)
                 {
-                    routeData.Add(item.Key, item.Value);
+                    routeData[item.Key] = item.Value;
```
jtkech, line 384: Okay unless we would prefer `routeData.TryAdd()` to check if the key already exists, I'm talking about this just because I saw that in `GetRouteData()` this is what we do when we update route data from the request query.

But not sure at all what would be better to do here, Hmm, before with `.Add()`, it was failing if the key already exists.

src/OrchardCore.Modules/OrchardCore.Navigation/PagerShapes.cs
```diff
@@ -572,5 +562,37 @@ private static IHtmlContent CoerceHtmlString(object value)
...
+            var routeData = new RouteValueDictionary(Html.ViewContext.RouteData.Values);
+
+            if (httpContext != null)
+            {
+                var queryString = httpContext.Request.Query;
+                if (queryString != null)
+                {
+                    foreach (var key in from string key in queryString.Keys where key != null && !routeData.ContainsKey(key) let value = queryString[key] select key)
+                    {
+                        routeData[key] = queryString[key];
```
jtkech, line 580: That's okay as it is from existing code.

But I would rename `queryString` to `query`, hmm here I don't think it can be null and can have any null key, then we do `let value` but don't use it. So the suggested code would be, but would need to be tried, so as you want ;)

                var query = httpContext.Request.Query;
                foreach (var key in query.Keys)
                {
                    routeData.TryAdd(key, query[key]);
                }

src/OrchardCore.Modules/OrchardCore.Navigation/PagerShapes.cs
Skrypt: It needs to be removed in case that we are on the first page or the last page. Here I am adding any QueryString params that might be passed and I found some places where the before or the after is passed even if it's not necessary which cause issues.

src/OrchardCore.Modules/OrchardCore.Navigation/PagerShapes.cs
Skrypt: idem

src/OrchardCore.Modules/OrchardCore.Navigation/PagerShapes.cs
Skrypt: I've not tested this yet but if it is a dictionary I'm not sure that affecting a key that doesn't exist will create the key/value. This is why I was using the `.Add()` and affecting only if it was existing. These are code optimizations that only will give the same result in the end. I'm trying to fix the issues we have here with the Pagers before anything. We have an SEO issue with pages that can have 2 different URLs as an example. We talked about this in this afternoon's meeting.

src/OrchardCore.Modules/OrchardCore.Navigation/PagerShapes.cs
Skrypt: Was working before ... code optimization. Idem. 😉

src/OrchardCore.Modules/OrchardCore.Navigation/PagerShapes.cs
jtkech: Okay no problem, what I said is that `routeData.Remove()` checks internally if the key exists https://github.com/dotnet/aspnetcore/blob/c481a38000428c50efcbd2fc075f6084cf7a3227/src/Http/Http.Abstractions/src/Routing/RouteValueDictionary.cs#L491, so can be done in one line without having to do `ContainsKey()` line 412.

src/OrchardCore.Modules/OrchardCore.Navigation/PagerShapes.cs
jtkech: Okay no problem just because in `GetRouteData()` I saw it doesn't do the same ;)

`.Add()` adds if not exists but throw if exists.
`TryAdd()` only set if not exists and never throws.
`[key] = value` set or replace existing value and never throws.

In `GetRouteData()` we were using a kind of `TryAdd()` by 1st checking for the key.

Skrypt: If I'm doing `.TryAdd()` it will add a value only if it doesn't exist. Though it probably won't update the value if it exists and that is what I want to do. So, I think it `.TryAdd()` is not appropriate.

jtkech: Okay so `[key] = value` is the best, as you already did ;)

Hmm, but then maybe this is what we also want in `GetRouteData()`.

src/OrchardCore.Modules/OrchardCore.Navigation/PagerShapes.cs
Skrypt: image

It is called a QueryString. Probably was shorter to just name it `Request.Query` but it is always called a QueryString.

jtkech: No problem for the naming

The `Request` has 2 different properties `QueryString` and `Query`, but yes the last has been populated by parsing the first.

Skrypt:
The more I read code in that Class the more I find optimizations to do.

Skrypt:
While working on this I found out that we use also datetimes values on after and before params which makes this a little more complicated to fix. We should probably drop the datetimes unless there was a reason behind it. Also, even if we don't know the amount of total records that we have we could easily use a pageNum param instead of after and/or before. Knowing that a pagesize == x having a PageNum implies that a page is After X and Before Y.

jtkech:
Don't remember all details but these after/before are integers, unique per item and representing their ordering, if a list part is orderable items have configurable Order values, otherwise we use the createdUtc ticks. In our project we have a content type having a numeric field (so editable per item) that we use to order items by using this field to set the pager after/before.

That said we also had to use the full Pager in the front end, in that case there is no after/before, only had to pass the page number, but also a `TotalItemsCount` property.

Now that you already did a number of changes, maybe worth to get rid of the Linq query in `GetRouteData()`, this for perf of pages using Pager(s).

Skrypt:
I'm just trying to find a way to use consistent params because here a page can have 2 valid URI. For example.
Let's say I have a page size == 2.

?after=2 and ?before=5 will get you to the same page. This cause issues with SEO.

Another example : 

?before=3 OR ?&after=0 OR / 

These are all pointing to the first page which normally shouldn't require a param to be used. That would be more appropriate to not use a param for SEO.

So, if we are using a timestamp we can't always just display one param. I would have used only "after" in that case since we don't need the "before" param if we are always passing int in those params.

So, I think we will need to pass a canonical URL in the header meta of the page to make sure SEO is consistent.

jtkech:
Yes, that's why with the 1st page of the full pager we don't add `?pagenum=1` for SEO.

Yes, with the full pager we do paging, the query skip n * size and then take a page size of items, it also needs the total count.

With the pagerSlim we have a kind of chain with afters/befores to go forward/backward, maybe to not have to do a additional query to get the total count of items.

I will think about this SEO problem in that case.

Skrypt:
And the issue with timestamps is that we cannot evaluate if we are on the first page or not with them.

Skrypt:
I'm going to merge this PR and open another one for the SEO issue since it fixes the issue with the QueryString params missing.

PR #11733, opened 2022-05-20
Added placement example for dynamic parts

brunoAltinet:


src/docs/reference/core/Placement/README.md
```diff
@@ -319,6 +319,60 @@ We also specify that the `Content` column will take 9 columns, of the default 12
...
+
+Alternatively, if you find this complicated for dynamic parts, you can display your dynamic part like so:
+
+=== Content-Product.Summary.html
+``` html
+@await DisplayAsync(Model.Content.GalleryPart)
+```
+
+!!! note
+    Directly displaying parts like shown above without placement is a bit hacky. If for example we would call both `@await DisplayAsync(Model.Content.GalleryPart)` and `@await DisplayAsync(Model.Content)` inside the same template, we would expect both calls to show gallery. However, only the first call will show the gallery, and the other would ignore the gallery shape (since it's already rendered).
```
ns8482e, line 375: IMHO  - In documentation, objective is to clear the doubt - we should avoid wording that creates doubts in readers mind.

brunoAltinet: Yeah, i'm not completely for keeping this in. Although i like the fact you can do that.

ns8482e (approved):

src/docs/reference/core/Placement/README.md
```diff
@@ -319,6 +319,50 @@ We also specify that the `Content` column will take 9 columns, of the default 12
...
+  "ContentPart": [
+    {
+      "place": "MyGalleryZone",
+      "differentiator": "GalleryPart"
+    }
+  ]
+}
+```
+
+This setup would then show your template (e.g. `GalleryPart.cshtml` or `GalleryPart.Detail.cshtml`) where DisplayAsync was called.
```
hishamco, line 344: `DisplayAsync`

src/docs/reference/core/Placement/README.md
```diff
@@ -319,6 +319,50 @@ We also specify that the `Content` column will take 9 columns, of the default 12
...
+
+If we would like to show the same part in a summary display content template (or any other that isn't `Detail` display type):
+
```
hishamco, line 347: Remove extra line

src/docs/reference/core/Placement/README.md
```diff
@@ -319,6 +319,50 @@ We also specify that the `Content` column will take 9 columns, of the default 12
...
+{
+  "ContentPart_Summary": [
+    {
+      "place": "MyGalleryZone",
+      "differentiator": "GalleryPart"
+    }
+  ]
+}
+```
+This setup would then show your template  (e.g. `GalleryPart.cshtml` or `GalleryPart.Summary.cshtml`) where DisplayAsync was called.
```
hishamco, line 365: Add line before and after code snippet

hishamco (changes requested):
LGTM after fixing the format

src/docs/reference/core/Placement/README.md
hishamco: `DisplayAsync`

hishamco (changes requested):

hishamco (approved):

PR #11763, opened 2022-05-24
Add user Edit and Delete permission to the users UI

### LINKED ISSUE #11762: Expand the User management permission to allow for Edit,Delete or view only scenarios
I have a need where some users need to be able to view and edit other users but not to be able to delete users. Currently, ManageUser permission is all that we have around editing/deleting users which is limiting the ability to grant the users the needed permissions.

I think we should add 
 1. EditUsers: allows the user to edit other users
 2. DeleteUsers: allows the user to delete other users

Now, a user with EditUsers permission would be able to add users or edit any users. The existing `ManageOwnUserInformation` would only allow the user to edit their own profile. Also, any user with ManageUsers would be able to also edit, delete to prevent breaking existing permission.

Now, this would give us more flexibility/control over the secured user UI.

MikeAlhayek:
This would provide a way to enable/disable permission to Edit or Delete users explicitly. 

Fixes #11762


src/OrchardCore/OrchardCore.Users.Core/CommonPermissions.cs
```diff
@@ -11,13 +11,23 @@ public class CommonPermissions
...
+
         /// <summary>
         /// View users only allows listing or viewing a users profile.
         /// </summary>
-        public static readonly Permission ViewUsers = new Permission("View Users", "View Users", new[] { ManageUsers });
+        public static readonly Permission ViewUsers = new Permission("View Users", "View All Users", new[] { ManageUsers, EditUsers, DeleteUsers });
 
         // Dynamic permission template.
-        private static readonly Permission ManageUsersInRole = new Permission("ManageUsersInRole_{0}", "Manage Users in Role - {0}", new[] { ManageUsers });
+        private static readonly Permission ManageUsersInRole = new Permission("ManageUsersInRole_{0}", "Manage Users in Role - {0}", new[] { EditUsers });
```
sebastienros, line 30: I am concerned it would allow a less restricted user to change people in specific roles. I'd keep it like before for safety.

CrestApps: @sebastienros Thanks for looking at this PR.

`ManageUsers` is a higher permission than `EditUsers` so we should be fine to use `EditUsers`. If we keep it to `ManageUsers`, then it we'll eventually override `EditUsers` and `DeleteUsers`. Perhaps, we should rename "Manage Users in Role - {0}" to "Edit Users in Role - {0}"?

However, currently I think we list all users "weather the user can manage them or not". This may be a problem even before this PR. Also before this PR, the `ManageUsersInRole_{0}` without ViewUser is useless since you can't see the users to even manage them. Also, even when a listed in X role and the logged user does not have ManageUsersInRole_X, he is still able to edit that user. I think this is an Existing security issue with that `ManageUsersInRole_` permission.

**A potential fix:** Add a `UserRoleIndex` that would give store the user to role relation "so we can query it" during listing the users. Now, we could add the following to fix this issues.

 1. Add a new service `IUsersAdminListFilter` that would allow us to query the new `UserRoleIndex` to see if a user can manage or edit 
 2. On [this line](https://github.com/OrchardCMS/OrchardCore/blob/c8b2954e038aaa54963b60bc6cfe193cfea54799/src/OrchardCore.Modules/OrchardCore.Users/Services/DefaultUsersAdminListQueryService.cs#L33) add a call to a new service like this `await _usersAdminListFilters.InvokeAsync((filter, model, query, updater) => filter.FilterAsync(model, query, updater), model, query, updater, _logger);`.

src/OrchardCore.Modules/OrchardCore.Users/Controllers/AdminController.cs
```diff
@@ -224,7 +224,7 @@ public async Task<ActionResult> IndexPOST(UserIndexOptions options, IEnumerable<
             // Check a dummy user account to see if the current user has permission to manage it.
             var authUser = new User();
 
-            if (!await _authorizationService.AuthorizeAsync(User, Permissions.ManageUsers, authUser))
+            if (!await CanEditContextAsync(authUser))
```
ns8482e, line 227: At first I didn't understand what is `CanEditContextAsync` 

I guess keeping `_authorizationService.AuthorizeAsync` keeps code more in flow and readable rather than moving logic in private functions

ns8482e (commented):
In PR,  Permissions are changed from `ManageUsers`  to  `EditUsers`, so what's the real difference between these two?

src/OrchardCore.Modules/OrchardCore.Users/Controllers/AdminController.cs
CrestApps: @ns8482e 
Here `CanEditContextAsync` is performing two check. First it checking to see if the user has EditUsers, or if the user has `ManageOwnUserInformation` to the given user.  changing it to `await _authorizationService.AuthorizeAsync(User, Permissions.EditUsers, user)` would be a different logic. We could rename the `CanEditContextAsync` method to `CanEditUserAsync` to make it more readable. Thoughts?

ns8482e:
As @sebastienros  suggested in meeting - changing username/password is functionally equivalent of deleting so there is not real different between two distinct permission edit/delete , hence IMO permission `ManageUsers` is enough for managing identity/permission related feature set provided OOTB. 

However @CrestApps I also understand your requirement to have ability to add/edit users by roles  other than Administrator that may need custom permission ( in your case in this PR EditUser/DeleteUser) where you might want to provide some additional functionality other than identity/permission for the user being created/edit.


For example -  Building a CRM, where in  a Sales Person may onboard new Account/Contact where he's adding sales related information that would create a user's login account in  OC as part of sales account creation process but the sales person really do not have permission to create/manage user in OC as Administrator role has.

Above can be achieved  by defining your own permission/controller and using UserService/Workflow - keeping OOTB user management unchanged.

src/OrchardCore/OrchardCore.Users.Core/CommonPermissions.cs
sebastienros: The ImpliedBy in line 20 might be removable completely.

CrestApps: I removed it. same effect. So we list all users for a user with ViewUsers permission. if a user is able to manage users with in X role, the user will see the edit button.

Skrypt:
Fix conflict by merging latest main branch.

MikeAlhayek:
This PR was replaced by #12407

PR #11766, opened 2022-05-25
Removes duplicated email validation in user registration.

### LINKED ISSUE #11765: Duplicated validation messages when registering as new user
### Describe the bug
The user registration is validating the email address of new users in two places, which will cause two validation messages on the summary view:

1. In the `RegisterViewModel` class which is an `IValidatableObject`
2. In the `RegistrationController` with the addition of checking for duplicated email addresses.

### To Reproduce
Enable user registration and leave the email field empty.

gvkries:
Fixes #11765

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Users/Controllers/RegistrationController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Controllers/RegistrationController.cs
@@ -7,7 +7,6 @@ using Microsoft.AspNetCore.Mvc.Localization;
 using Microsoft.Extensions.Localization;
 using Microsoft.Extensions.Logging;
 using OrchardCore.DisplayManagement.Notify;
-using OrchardCore.Email;
 using OrchardCore.Entities;
 using OrchardCore.Modules;
 using OrchardCore.Settings;
@@ -23,7 +22,6 @@ namespace OrchardCore.Users.Controllers
         private readonly IAuthorizationService _authorizationService;
         private readonly ISiteService _siteService;
         private readonly INotifier _notifier;
-        private readonly IEmailAddressValidator _emailAddressValidator;
         private readonly ILogger _logger;
         private readonly IStringLocalizer S;
         private readonly IHtmlLocalizer H;
@@ -33,7 +31,6 @@ namespace OrchardCore.Users.Controllers
             IAuthorizationService authorizationService,
             ISiteService siteService,
             INotifier notifier,
-            IEmailAddressValidator emailAddressValidator,
             ILogger<RegistrationController> logger,
             IHtmlLocalizer<RegistrationController> htmlLocalizer,
             IStringLocalizer<RegistrationController> stringLocalizer)
@@ -42,7 +39,6 @@ namespace OrchardCore.Users.Controllers
             _authorizationService = authorizationService;
             _siteService = siteService;
             _notifier = notifier;
-            _emailAddressValidator = emailAddressValidator ?? throw new ArgumentNullException(nameof(emailAddressValidator));
             _logger = logger;
             H = htmlLocalizer;
             S = stringLocalizer;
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=11766) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=11766) before we can accept your contribution.<br/><br/>:x: gvkries [sign now](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=11766)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OrchardCMS/OrchardCore?pullRequest=11766) it.</sub>

src/OrchardCore.Modules/OrchardCore.Users/Controllers/RegistrationController.cs
```diff
@@ -74,12 +70,7 @@ public async Task<IActionResult> Register(RegisterViewModel model, string return
                 return NotFound();
             }
 
-            if (string.IsNullOrEmpty(model.Email))
-            {
-                ModelState.AddModelError("Email", S["Email is required."]);
-            }
-
-            if (_emailAddressValidator.Validate(model.Email))
```
hishamco, line 82: The `IEmailValidator` is useful for checking the email syntax

hishamco:
You break the build

src/OrchardCore.Modules/OrchardCore.Users/Controllers/RegistrationController.cs
gvkries: That is already done in the `RegisterViewModel`: https://github.com/OrchardCMS/OrchardCore/blob/40d3f65663c8582a9c76c59d4f5bdfaf781d0bea/src/OrchardCore.Modules/OrchardCore.Users/ViewModels/RegisterViewModel.cs#L35

Skrypt:
No, we can't assume that the View that is posting to that form is using that ViewModel. It is important that the controller makes a double verification to make sure it is a valid email.

gvkries:
Sorry, but that statement isn't correct, validation takes place in the controller. It does it twice!

Skrypt:
It is fine to validate in the Controller. The ViewModel validation is just for that View. What if you use a different form to post to that controller action and that you don't validate the email?

Skrypt:
Add validation on the ViewModel itself to make sure it is never NULL on that controller action before doing anything else. That might fix the issue I was referring to.

gvkries:
Maybe I'm not getting your point, but it's already doing all of that in https://github.com/OrchardCMS/OrchardCore/blob/40d3f65663c8582a9c76c59d4f5bdfaf781d0bea/src/OrchardCore.Modules/OrchardCore.Users/ViewModels/RegisterViewModel.cs#L31 
It doesn't make sense to redo the validation a second time IMHO.

Skrypt:
Here, when we call the action : 

https://github.com/OrchardCMS/OrchardCore/blob/cf16293360e3acce4c5eed7e7a679039a24210ea/src/OrchardCore.Modules/OrchardCore.Users/Controllers/RegistrationController.cs#L64-L67

The `RegisterViewModel model` is passed to the action by doing databinding magical stuff. It is not given that the model here will not fail data binding if we are using a RazorView that has an `AlternateRegistrationViewModel` that has a completely different structure. The model will then be "null" in that case. So, we should drop the request.

Sp basically, make sure that the model passed is never null.

gvkries:
That's the normal way it is handled everywhere in asp.net and I didn't change that here. If model is null, an `ArgumentNullException` will be thrown by `TryValidateModel` now, before it would be an NRE by accessing its properties. If you override `RegisterViewModel` with your own type and remove the validation, you are also going to fail in other places, e.g. `this.RegisterUser()`, which also doesn't recheck every property.

Skrypt:
We need to try to make sure. Because else keeping the validation in the controller makes sense. I don't want to rely on an NRE to say that it is safe.

gvkries:
This is what model binding and model validation is build for, no need to reinvent the wheel here in a single instance.

Skrypt:
Well for this PR I assume that it is ok. I don't mind. if you omit to send only an email param from an AJAX POST request then it should throw an NRE at line 73. We could instead on the first line do a redirect with a notification message if model == null.
But I guess this should be something handled by the Validator at that point.

Need to try I guess.

Skrypt:
By the way, if the request throws an NRE from an AJAX request it can be used for a DDOS attack. So while we have the Antiforgery Token I feel less concerned but we should probably still take some time to think about these on known routes like these.

PR #11799, opened 2022-06-02
Workflow - content for each task - allows looping through content items using a query or contentype

PBMikeW:
Allows looping through content items either of a single type or using a query, many possible uses I.e. loop through form entries and insert into CRM. Update content item lists etc etc.

image
image


src/docs/reference/modules/Workflows/README.md
```diff
@@ -85,6 +85,11 @@ Although many activities support multiple outcomes, they typically return only o
 For example, the _Send Email_ activity has two possible outcomes: "Done" and "Failed".  
 When the email was sent successfully, it yields "Done" as the outcome, and "Failed" otherwise.
 
+For activities that have an iterator outcome e.g _Content For Each_, return the outcome of the next activity back to the irrerate task for the irritation to continue either through the done outcome or via a fork.
```
Skrypt, line 88: irrerate == iterate?

PBMikeW: Ah yes, knew my spelling was off so spell checked but then did not double check the result :/

src/OrchardCore.Modules/OrchardCore.Contents/Workflows/Activities/ContentForEachTask.cs
```diff
@@ -0,0 +1,95 @@
...
+            return Outcomes(S["Iterate"], S["Done"]);
+        }
+        public override bool CanExecute(WorkflowExecutionContext workflowContext, ActivityContext activityContext)
+        {
+            return true;
+        }
+        public override async Task<ActivityExecutionResult> ExecuteAsync(WorkflowExecutionContext workflowContext, ActivityContext activityContext)
+        {
+            var contentItems = await _session
+                .Query<ContentItem, ContentItemIndex>(index => index.ContentType == ContentType && index.Published == true && index.Latest == true)
```
sebastienros, line 45: Unbounded query, could load all the content items and OOF.

sebastienros, line 45: The filter doesn't seem right, Published AND Latest? Means if there is a draft we are not loading the Published version of it.

src/OrchardCore.Modules/OrchardCore.Contents/Workflows/Activities/ContentForEachTask.cs
```diff
@@ -0,0 +1,95 @@
...
+                .ListAsync();
+            List<ContentItem> items = contentItems.ToList();
+            if (contentItems.Count() == 0)
```
sebastienros, line 48: If there is no content item of this type it is throwing an Exception?

src/OrchardCore.Modules/OrchardCore.Contents/Workflows/Activities/ContentForEachTask.cs
```diff
@@ -0,0 +1,95 @@
...
+            {
+                throw new InvalidOperationException($"The '{nameof(ContentForEachTask)}' failed to evaluate the 'ContentType'.");
+            }
+
+            if (Index < contentItems.Count())
+            {
+                var contentItem = await _contentManager.LoadAsync(items[Index]);
```
sebastienros, line 55: So you are loading all content items of a type, in a loop, then a single item?

sebastienros: I would suggest to at least use `LIMIT 1 WHERE ID > LastItemId` which would be store in the state. OR even better, load `LIMIT 2 WHERE ID >= NextItemId` then you store the second one as NextItemId. Each loop then only load 2 documents max. You could even just return the ids, and load the first one in a separate query to optimize the bandwidth.

PBMikeW: > So you are loading all content items of a type, in a loop, then a single item?
It was my understanding that LoadAsync triggers the load events and does not make a db call.
image

src/OrchardCore.Modules/OrchardCore.Contents/Workflows/Activities/ContentForEachTask.cs
PBMikeW: Correct, all other content Tasks throw exceptions when content can't be found so was following the module conventions. The below is from RetriveContentTask. I do agree the workflow error handling could be better.
image

Happy to change based on your advice.

src/OrchardCore.Modules/OrchardCore.Contents/Workflows/Activities/ContentForEachTask.cs
PBMikeW: My use case was for published only, I will add options to the view model.

src/OrchardCore.Modules/OrchardCore.Contents/Workflows/Activities/ContentForEachTask.cs
sebastienros: But it's a for loop, so it doesn't mean you expect a specific item, might be different semantic it.

src/OrchardCore.Modules/OrchardCore.Contents/Workflows/Activities/ContentForEachTask.cs
sebastienros: Still, if there is a published item and there is also a draft, you are not loading the published one.

PBMikeW: My misunderstanding of the filters, will sort.

PBMikeW: Will implement paging and take with defaults.

src/OrchardCore.Modules/OrchardCore.Contents/Workflows/Activities/ContentForEachTask.cs
PBMikeW: I get what you mean now, I will implement the exception on only incorrect contenttype and have a select for contenttype.

src/OrchardCore.Modules/OrchardCore.Contents/Workflows/Activities/ContentForEachTask.cs
PBMikeW: Take parameter added

Piedone:
@sebastienros @Skrypt do you want to get back to finish the review here?

Piedone:
Please fix the build errors. Since then, OC moved from Json.NET to System.Text.Json.

PBMikeW:
I'll tidy this up early July.

Piedone:
Thank you

PBMikeW:
@sebastienros @Piedone - Looking at this two years on, should this be split to fit more with the other workflows. I.e. 1) RetrieveContentFromQuery and then have a dropdown for query type? It can then be processed by the foreach task. On the flip side to that, I do find the workflows overwhelming for people and iterating in the one task would be way simpler.

Piedone:
I think having a dropdown would be nicer than a textbox, yes. Also, yes, I'd imagine this working similarly to the For Each task, in that it'll execute the connected task for each item (like the Iterate path). I don't think it makes sense to have to loop this into For Each.

MikeAlhayek:
for instructions on how to resolve the merge conflicts due to #16572 please follow the step listed in [this comment](https://github.com/OrchardCMS/OrchardCore/issues/16572#issuecomment-2291836365).

PR #11805, opened 2022-06-05
Re-add tenant level AFT on setup

### LINKED ISSUE #11501: Unexpected behavior when "ARR affinity" cookies is disabled

jtkech:
Fixes #11501 

Before we were using fixed cookie names, and to prevent some logged errors on setup, in place of registering tenant ATF while the tenant is not initialized, we just remove a possible existing cookie, and let the host level registrations being used.

- We don't need this anymore as we now use fixed cookie names based on the settings VersionId.

- Also, only for the default tenant, when first creating the setup context the VersionId was null, this leading up to an `__orchantiforgery_` cookie name which was no more unique.

Useful to still register tenant level AFT even if the tenant is in a setup context, in a multi instances env to setup tenants we may want to use a tenant level shared data protection, as tried in #11501 from the app by using `AddSetupFeatures()`, and then on setup we also need a tenant level AFT that will use this DP.

### DIFF
--- a/src/OrchardCore/OrchardCore/Shell/ShellHost.cs
+++ b/src/OrchardCore/OrchardCore/Shell/ShellHost.cs
@@ -378,6 +378,7 @@ namespace OrchardCore.Environment.Shell
                 // Creates a default shell settings based on the configuration.
                 var shellSettings = _shellSettingsManager.CreateDefaultSettings();
                 shellSettings.Name = ShellHelper.DefaultShellName;
+                shellSettings.VersionId = IdGenerator.GenerateId();
                 shellSettings.State = TenantState.Uninitialized;
                 defaultSettings = shellSettings;
             }
--- a/src/OrchardCore/OrchardCore/Shell/ShellHost.cs
+++ b/src/OrchardCore/OrchardCore/Shell/ShellHost.cs
@@ -366,7 +366,7 @@ namespace OrchardCore.Environment.Shell
         /// <summary>
         /// Creates a transient shell for the default tenant's setup.
         /// </summary>
-        private Task<ShellContext> CreateSetupContextAsync(ShellSettings defaultSettings)
+        private async Task<ShellContext> CreateSetupContextAsync(ShellSettings defaultSettings)
         {
             if (_logger.IsEnabled(LogLevel.Debug))
             {
@@ -378,12 +378,13 @@ namespace OrchardCore.Environment.Shell
                 // Creates a default shell settings based on the configuration.
                 var shellSettings = _shellSettingsManager.CreateDefaultSettings();
                 shellSettings.Name = ShellHelper.DefaultShellName;
-                shellSettings.VersionId = IdGenerator.GenerateId();
                 shellSettings.State = TenantState.Uninitialized;
                 defaultSettings = shellSettings;
+
+                await UpdateShellSettingsAsync(defaultSettings);
             }
 
-            return _shellContextFactory.CreateSetupContextAsync(defaultSettings);
+            return await _shellContextFactory.CreateSetupContextAsync(defaultSettings);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/OrchardCore/OrchardCore/Shell/ShellHost.cs
```diff
@@ -378,6 +378,7 @@ private Task<ShellContext> CreateSetupContextAsync(ShellSettings defaultSettings
                 // Creates a default shell settings based on the configuration.
                 var shellSettings = _shellSettingsManager.CreateDefaultSettings();
                 shellSettings.Name = ShellHelper.DefaultShellName;
+                shellSettings.VersionId = IdGenerator.GenerateId();
```
Skrypt, line 381: Will this generate a new Id every time I go to the setup page from, let's say, different browsers?

jtkech: `CreateSetupContextAsync()` is only used for the Default tenant.

For all other tenants, each time we create/update settings, the settings are persisted so you will retrieve the same Id from any browser and any app app instance if the settings store is shared.

Here for the default tenant, the settings for setup are only set in memory, so you will get the same Id from any browser if you target the same app instance. The Id will change if you restart the app instance or target another one.

But it is so simple to also call here `UpdateShellSettingsAsync()` which will set once a new VersionId, as for other tenants, I did the update and tried it, now you will get the same Id from any browser and any targetted instance, also for the Default tenant.

PR #11813, opened 2022-06-06
Fix Layers module "sortable" behavior.

### LINKED ISSUE #11812: Zones - Edit Widgets and Layers - Widgets reorder drag/drop not working - v1.4
### Describe the bug
v1.4. 
On the admin 'Setting - Zones' page, drag and drop of widgets doesn't work. 

### To Reproduce
1. Go to 'Design - Settings - Zones'
2. Create two or more widgets in a zones'
3. Try to drag a widget to a space above or below another widget. The multi-arrow icon appears but you cannot move order of widget

### Expected behavior
Tested and works on v1.1. 
You should be able to move a widget to a different position/zone

Skrypt:
Fixes #11812


Skrypt:
Merged, because ... regressions. Nothing new here.

hishamco:
I expect it's Sortable issue

Skrypt:
Yeah, I guess it didn't get much of my attention because it needs to be migrated to sortable.js so that we can remove the JQuery UI dependency eventually.

PR #11816, opened 2022-06-07
Use form-select class for select element instead of form-control

### LINKED ISSUE #11815: Some <select> tags have "form-control" class, but really should be 'form-select'
In most instances, from a UI/UX perspective, a `<select>` tag (drop down list) should have a 'form-select' class, not a 'form-control' class. There are several `<select>` tags that still have the 'form-control' class.

e.g. Taxonomy Edit > Term Content Type 

The result is that there is no indication it is a drop down (normally with the down caret) until focus/action in control

Update the select tag class where appropriate so the representation is a drop down list

hishamco:
Fixes #11815

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Media/Views/MediaProfiles/Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Views/MediaProfiles/Edit.cshtml
@@ -43,7 +43,7 @@
     <div class="form-g mb-3" asp-validation-class-for="SelectedHeight">
         <div class="w-100 w-md-50">
             <label asp-for="SelectedHeight">@T["Height"]</label>
-            <select asp-for="SelectedHeight" asp-items="@Model.AvailableHeights" class="form-selectv content-preview-select" asp-validation-class-for="SelectedHeight"></select>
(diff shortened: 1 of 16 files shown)


src/OrchardCore.Modules/OrchardCore.Media/Views/MediaProfiles/Edit.cshtml
```diff
@@ -43,7 +43,7 @@
     <div class="form-g mb-3" asp-validation-class-for="SelectedHeight">
         <div class="w-100 w-md-50">
             <label asp-for="SelectedHeight">@T["Height"]</label>
-            <select asp-for="SelectedHeight" asp-items="@Model.AvailableHeights" class="form-control content-preview-select" asp-validation-class-for="SelectedHeight"></select>
+            <select asp-for="SelectedHeight" asp-items="@Model.AvailableHeights" class="form-selectv content-preview-select" asp-validation-class-for="SelectedHeight"></select>
```
agriffard, line 46: selectv => select

hishamco: Paste issue ;)

PR #11713, opened 2022-05-16
Provide a way to override YesSql configuration

CrestApps:
currently there is no way to change the configuration of YesSql in OC. This would provided option to allow one to override the default.

In my case, I need a way to change the of [ITableNameConvention](https://github.com/sebastienros/yessql/blob/1d04a007d893f0287ecc721f087d4f9764f9c90f/src/YesSql.Core/Services/DefaultTableNameConvention.cs)

PR #11822, opened 2022-06-07
Validate database connection before allowing a tenant to be added or setup

### LINKED ISSUE #11803: Validate table prefix before allowing new tenant to be added
### Describe the bug
When adding a new tenant, we allow the user to set a prefix to the tables to enable reusing the same database for multiple tenets. However, we don't validate to ensure that the prefix isn't already used first. The problem with this is that it throws an exception

Screenshot_6

### To Reproduce
1. Create 2 tenant on same database connection with the same prefix or both with no prefix.
2. Set up both tenants. you should see the exception when you set up the second tenant.

### Suggestion 
When creating the tenant, we should group existing connections by prefix. Then check to make sure the prefix isn't already used before allowing it.

I think this logic should be part of the `ITenantValidator` But I think the validation would have to be done per type/servername/databasename/prefix

### LINKED ISSUE #11863: Invalid serial number for shell descriptor during tenant setup. Validate database connection before allowing new tenant to be added
When a user is trying to add a new tenant while the database already exists, the app throw the following exception

image

The above exception is valid, but I think we should prevent it from happening but adding check to prevent the user from getting to this state.

Here are some scenarios that could lead to this exception
 1. On a local environment
   - The developer adds a new tenant (with or without prefix). For argument sake, we'll use `site1` prefix.
   - The developer deletes the App_Data folder from the project
   - The developer adds another tenant with the prefix `site1`
   - The UI will allow adding the tenant, but the setup of the tenant will fail with the exception listed above. IMO, The tenant should not be added in this case
 2. A user adds a new tenant from the UI
   - The user adds a new tenant (with or without prefix). For argument sake, we'll use `site1` prefix.
   - The user adds another tenant (and unintentionally uses the `site1` prefix.)
   - Unfortunately, the UI will allow adding the tenant, but the setup of the tenant will fail with the exception listed above. IMO, The tenant should not be added in this case

Additionally, there is no way of telling if the provided database connection is valid when a new tenant is added. Currently, the only way is to attempt to setup the tenant, and hope that all goes well. Otherwise, you'll need to update the connection string and attempt to setup the site all over again. IMO, this isn't a good user experience and is error-prone.

I would think some sort connection validation should take place to ensure that the database does not exists before allowing the tenant to be added. Since OrchardCore uses Yessql which require the `Document` table, I think we should run the following checks against this table prior allowing adding/updating the database connection
 1. Try to open a connection to the provided database connection. If the connection is open successfully, we then would know that the database connection is valid.
 2. When the connection is valid, we can try to execute select query on the `Document` table, if the select query runs, then we know that the table already exists and we'll prevent the user from using the provided prefix. Even if the user does not provide a prefix and `Document` table already used, we'll force the user to use a prefix.

### LINKED ISSUE #11876: Validate database connection before setting up a tenant
In some cases, while trying to setup a new tenant on the "Setup" screen, the user is able to provide connection settings if the connection is not per-configured. 

In other case, the setup could fail and only some of the tables are created on the connection which leads to incomplete setup. If we run into an in-complete setup, the user should be notified that the tables already exists so they can change the connection or manually delete incomplete tables. Not sure if possible, if a setup fail, may be find out the created table and drop them manually? We would have to find a way to track created tables during setup before dropping table otherwise we could unintentionally drop wrong tables.

Both cases above will lead to this exception

image

I think we should validate the database connection before allowing the user to proceed the setup.

Bonus, If possible, track a list new tables, and if the setup fail, drop them

This issue is related to #1183 and #11863 but targets a different scenarios.

CrestApps:
1. Fix #11803 
 2. Fix #11863 
 3. Fix #11876
 4. Make the table prefix separator a configurable option to allow a user to user prefix separator other than `_`


hishamco:
Thanks again Mike, seems still you have a Git issue ;)

hishamco:
I may share simpler way to fix the issue, let me try it tonight

CrestApps:
> Thanks again Mike, seems still you have a Git issue ;)

Yes I see that. Please check #11825

> I may share simpler way to fix the issue, let me try it tonight

Can't wait to see a simpler way.

Thank you

src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/IConnectionFactoryProvider.cs
```diff
@@ -0,0 +1,9 @@
+using OrchardCore.Environment.Shell;
+using YesSql;
+
+namespace OrchardCore.Data;
+
+public interface IConnectionFactoryProvider
```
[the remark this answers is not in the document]
CrestApps: done `IConnectionValidator` now provide info about the connection. This is good

src/OrchardCore/OrchardCore.Data.YesSql/ConnectionValidator.cs
```diff
@@ -0,0 +1,88 @@
...
+        selectBuilder.Take("1");
+
+        try
+        {
+            var selectCommand = connection.CreateCommand();
+            selectCommand.CommandText = selectBuilder.ToSqlString();
+
+            using var result = await selectCommand.ExecuteReaderAsync();
+
+            // at this point the query succeeded and the table exists
```
sebastienros, line 52: Looks like a comment I would have written, copy-pasted?

src/OrchardCore/OrchardCore.Data.YesSql/ConnectionValidator.cs
```diff
@@ -0,0 +1,88 @@
...
+            return ConnectionValidatorResult.ValidDocumentExists;
+        }
+        catch (Exception)
+        {
+            // at this point we know that the document table does not exists;
```
sebastienros, line 57:
```suggestion
            // at this point we know that the document table does not exist
```

src/OrchardCore/OrchardCore.Data.Abstractions/DatabaseHelper.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+
+namespace OrchardCore.Data;
+public class DatabaseHelper
+{
+    public static string GetStandardPrefix(string prefix)
```
sebastienros, line 6: Keep this in the implementation project, and make it internal. It's only used in this project.

src/OrchardCore/OrchardCore.Data.YesSql/ConnectionValidator.cs
CrestApps: @sebastienros  Actually this is exactly how I write my comments and it is not copy-paste. do you suggest an edit to it?

src/OrchardCore/OrchardCore.Data.Abstractions/DatabaseHelper.cs
CrestApps: done.

CrestApps:
If the pr #11713 merged, the code in `ConnectionValidator` should get updated by injecting [ITableNameConvention](https://github.com/sebastienros/yessql/blob/1d04a007d893f0287ecc721f087d4f9764f9c90f/src/YesSql.Core/Services/DefaultTableNameConvention.cs) in the constructor and using it instead of manually creating an instance during validation

src/OrchardCore/OrchardCore.Data.YesSql/ConnectionValidator.cs
```diff
@@ -0,0 +1,88 @@
...
+
+            return ConnectionValidatorResult.ValidDocumentDoesNotExists;
+        }
+    }
+
+    private IConnectionFactory GetFactory(string providerName, string connectionString)
+    {
+        return providerName.ToLower() switch
+        {
+            "sqlconnection" => new DbConnectionFactory<SqlConnection>(connectionString),
```
hishamco, line 67: I dislike the magic strings, probably we could add an enum-like class that contains all the providers

hishamco (changes requested):
Please add variant unit tests to ensure that it works for both valid & invalid use cases

hishamco:
Please HOLD ON, I'm working on similar PR to make it easier to achieve this goal, also we need more unit tests to make sure nothing broken

If my attempt failed probably I could add some modifiaction into this PR

src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/IConnectionValidator.cs
```diff
@@ -0,0 +1,16 @@
+using System.Threading.Tasks;
+
+namespace OrchardCore.Data;
+
+public interface IConnectionValidator
+{
+    public Task<ConnectionValidatorResult> ValidateAsync(string providerName, string connectionString, string tablePrefix);
```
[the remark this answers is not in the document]
CrestApps: done.

src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/IConnectionValidator.cs
```diff
@@ -0,0 +1,16 @@
...
+}
+
+public enum ConnectionValidatorResult
```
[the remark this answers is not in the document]
CrestApps: Done.

src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/IConnectionValidator.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    Unknown,
+    ValidDocumentDoesNotExists,
+    ValidDocumentExists,
```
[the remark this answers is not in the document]
CrestApps: updated.

src/OrchardCore/OrchardCore.Data.YesSql/ConnectionValidator.cs
```diff
@@ -0,0 +1,88 @@
...
+using YesSql.Services;
+using YesSql.Sql;
+
+namespace OrchardCore.Data;
+
+public class ConnectionValidator : IConnectionValidator
+{
+    public async Task<ConnectionValidatorResult> ValidateAsync(string providerName, string connectionString, string tablePrefix)
+    {
+        ArgumentNullException.ThrowIfNull(providerName);
```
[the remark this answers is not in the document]
CrestApps: Done.

src/OrchardCore/OrchardCore.Data.YesSql/ConnectionValidator.cs
CrestApps: Agreed! I changed this to `Enum` and removed the string from multiple places. Good suggestion

src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/IConnectionValidator.cs
```diff
@@ -0,0 +1,16 @@
+using System.Threading.Tasks;
+
+namespace OrchardCore.Data;
+
+public interface IConnectionValidator
```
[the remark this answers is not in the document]
CrestApps: done.

hishamco:
FYI I labled this PR as `dontmerge` because I mentioned at the beginning before this PR started that I have a simpler way to achieve the goal

After exchanging messages and voice notes with @CrestApps I realize there are two issues:
1. Avoid adding duplicate table prefix in the same database
2. Checking table existance (or check if Documents table is there) to avoid invali serial number issue

Based on that I created the PR #11846 which is fixes the issue without hitting the database. Backing to this PR I see the connection validations is good for certain scenarios such as checking Documents table, database is LIVE. So, if we add the validation during tenant creation that we be a behavior breaking-change because in the past the user can create a tenant without the database exists.

I suggest to :

- Open another issue explaining point 2, so this PR can be modified little bit and fine to fix the issue
- The PR I created can target the old issue because as described above it should avoid adding a duplicate table prefix on the same database.
- Add connection validation if the provider has connectionString (If we agreed that we can accept the behavior breaking changes)

/cc @sebastienros @jtkech @Skrypt @agriffard yours feedback is very important here

hishamco:
The build is broken :(

CrestApps:
> The build is broken :(

Very strange the build failed to pull the required packages which is why it failed. I faked a commit to force it to build again

hishamco:
I will make a quick review, but can your PR target the issue no. 2 to make it clear what each PR is target

CrestApps:
> I will make a quick review, but can your PR target the issue no. 2 to make it clear what each PR is target

@hishamco I created another issue to describe the other scenarios

hishamco:
@CrestApps can this target #11863 and related to the prefix issue, coz invalid serial could be happen if we didn't specify a table prefix

src/OrchardCore.Modules/OrchardCore.Tenants/Services/TenantValidator.cs
```diff
@@ -79,21 +84,64 @@ public async Task<IEnumerable<ModelError>> ValidateAsync(TenantViewModel model)
...
+                }
+
+                if (selectedProvider != null && hasConnectionString)
+                {
+                    await ValidateConnectionAsync(selectedProvider.Value, model.ConnectionString, model.TablePrefix, errors);
+                }
+
+                return errors;
+            }
+
```
sebastienros, line 108: `else`

src/OrchardCore.Modules/OrchardCore.Tenants/Services/TenantValidator.cs
```diff
@@ -79,21 +84,64 @@ public async Task<IEnumerable<ModelError>> ValidateAsync(TenantViewModel model)
...
             return errors;
         }
 
+        private async Task ValidateConnectionAsync(string databaseProvider, string connectionString, string tablePrefix, List<ModelError> errors)
+        {
+            var result = await _dbConnectionValidator.ValidateAsync(databaseProvider, connectionString, tablePrefix);
+
+            if (result == DbConnectionValidatorResult.UnsupportedProvider)
```
sebastienros, line 131: switch

sebastienros (approved):

hishamco:
@sebastienros hope we can accept this PR to resolve the invalid serial issue, but again I see table prefix issue can be easily resolved using settings

src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs
```diff
@@ -150,16 +155,32 @@ private async Task<string> SetupInternalAsync(SetupContext context)
...
+            switch (await _dbConnectionValidator.ValidateAsync(shellSettings["DatabaseProvider"], shellSettings["ConnectionString"], shellSettings["TablePrefix"]))
+            {
+                case DbConnectionValidatorResult.NoProvider:
+                    context.Errors.Add(String.Empty, S["DatabaseProvider is required."]);
+                    break;
+                case DbConnectionValidatorResult.UnsupportedProvider:
+                    context.Errors.Add(String.Empty, S["The provided database provider is not supported."]);
+                    break;
+                case DbConnectionValidatorResult.InvalidConnection:
+                    context.Errors.Add(String.Empty, S["The provided connection string is invalid or unreachable."]);
```
Skrypt, line 174: or server is unreachable.

src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs
```diff
@@ -150,16 +155,32 @@ private async Task<string> SetupInternalAsync(SetupContext context)
...
                 shellSettings["DatabaseProvider"] = context.Properties.TryGetValue(SetupConstants.DatabaseProvider, out var databaseProvider) ? databaseProvider?.ToString() : String.Empty;
                 shellSettings["ConnectionString"] = context.Properties.TryGetValue(SetupConstants.DatabaseConnectionString, out var databaseConnectionString) ? databaseConnectionString?.ToString() : String.Empty;
                 shellSettings["TablePrefix"] = context.Properties.TryGetValue(SetupConstants.DatabaseTablePrefix, out var databaseTablePrefix) ? databaseTablePrefix?.ToString() : String.Empty;
             }
 
-            if (String.IsNullOrWhiteSpace(shellSettings["DatabaseProvider"]))
+            switch (await _dbConnectionValidator.ValidateAsync(shellSettings["DatabaseProvider"], shellSettings["ConnectionString"], shellSettings["TablePrefix"]))
+            {
+                case DbConnectionValidatorResult.NoProvider:
+                    context.Errors.Add(String.Empty, S["DatabaseProvider is required."]);
```
Skrypt, line 168: DatabaseProvider setting is required.

src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs
```diff
@@ -192,7 +213,7 @@ private async Task<string> SetupInternalAsync(SetupContext context)
                         // unless the recipe is executing?
 
                         _logger.LogError(e, "An error occurred while initializing the datastore.");
-                        context.Errors.Add("DatabaseProvider", S["An error occurred while initializing the datastore: {0}", e.Message]);
+                        context.Errors.Add(String.Empty, S["An error occurred while initializing the datastore: {0}", e.Message]);
```
Skrypt, line 216: Why String.Empty here?

Skrypt:
Fix conflicts by merging latest main branch.

src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs
CrestApps: If the tenant database is preconfigured, we never show the connection fields even if the database provider has an error. In this case, we make it empty string and then we render that error to the user so they can see the error. Otherwise, the model state will have an error, but the user will not be able to see it

src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs
CrestApps: updated

src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs
CrestApps: done

CrestApps:
@Skrypt done.

src/OrchardCore/OrchardCore.Data.YesSql/DatabaseHelper.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+
+namespace OrchardCore.Data;
+internal class DatabaseHelper
+{
+    public static string GetStandardPrefix(string prefix)
```
Skrypt, line 6: Is it necessary? Maybe it can be retrieved from the `YesSql.Configuration`?
Just some thought.

CrestApps: @Skrypt I think so. the same logic is used into multiple places and it isn't int `YesSql.Configuration` it is being defined in `[OrchardCoreBuilderExtensions.cs](https://github.com/OrchardCMS/OrchardCore/pull/11822/files/aa461200d3106c54f93ede74cd7c5e7c06a741c9#diff-b499711d6aba6467e68b5d1377792bdadf5074011067c7ecedfd8e166458ec70) I thought it would be best to abstract that into common helper so it defined in one place and used multiple places

Skrypt: Ok, but if someone uses a different prefix with the OrchardCoreBuilderExtension then this won't work for him.

CrestApps: @Skrypt yes I get it. But the `YesSql.Configuration` does not store the separator `_` separately. Int he validator we need to know the used separator to be able to see if the table `prefix` is already in use or not.

Another way we could add a new property called `PrefixSeparator` to new YesSqlOptions  with `_` being the default. Then we could remove the Helper method and instead `prefix + YesSqlOptions.PrefixSeparator ?? String.Empty`. Do you want me to make this change?

CrestApps: @Skrypt check out the change now. I removed the DatabaseHelper. Now `TablePrefixSeparator` is an option that is available on `YesSqlOptions`. During the site setup, we store `TablePrefixSeparator` in the site settings so if the `YesSqlOptions.TablePrefixSeparator` change after tenants were setup, it wont break anything. For backward compatibility, if the TablePrefixSeparator isn't set, we use _ as the default value.

Let me know if you see an issue or requesting more changes.

Skrypt: Ok, I like it.

Skrypt (approved):

src/OrchardCore.Modules/OrchardCore.Tenants/Services/TenantValidator.cs
```diff
@@ -64,36 +56,63 @@ public async Task<IEnumerable<ModelError>> ValidateAsync(TenantViewModel model)
                 errors.Add(new ModelError(nameof(model.Name), S["Invalid tenant name. Must contain characters only and no spaces."]));
             }
 
-            if (!_shellSettings.IsDefaultShell() && String.IsNullOrWhiteSpace(model.RequestUrlHost) && String.IsNullOrWhiteSpace(model.RequestUrlPrefix))
+            if (!ShellHelper.DefaultShellName.Equals(model.Name, StringComparison.OrdinalIgnoreCase) && String.IsNullOrWhiteSpace(model.RequestUrlHost) && String.IsNullOrWhiteSpace(model.RequestUrlPrefix))
```
CrestApps, line 59: @jtkech This is the change you request I think. Replacing `_shellSettings.IsDefaultShell()` since it add no value being injected.

jtkech: Yes, but in fact these lines can be just removed as this check is useless as any tenant can have an empty host and prefix, the only requirement is that only one tenant can have this including the Default tenant.

Note: But I agree that it is not recommended that a non Default tenant has both empty host and url prefix, maybe the specifications has changed (needs to be checked).

jtkech: If we want to keep this test (if recommended) but make it working (I tried it, currently we can create a tenant with both empty host and url prefix e.g. if the default tenant has a non empty url prefix) we just need to remove `_shellSettings.IsDefaultShell()` from the if condition.

    if (String.IsNullOrWhiteSpace(model.RequestUrlHost) &&
        String.IsNullOrWhiteSpace(model.RequestUrlPrefix))

CrestApps: I think that rule exists just to reserve the “catch all/localhost” to the default-tenant/host. Otherwise, when running locally every time the app runs, we launch a non-default tenant. I am not sure if we should change that logic. I’ll await clear guidance form you or other to change it or leave it. Maybe @hishamco can provide more feedback on that matter since he created the TenantValidator originally.

jtkech: Okay no problem, I'm not against preserving the intended behavior, but if you read my comments this intended behavior is not working.

Note: Yes if the default tenant is not `DefaultIsCatchAll()` we then call `TryMatchInternal("", "", "/", out result)` letting the ability to another tenant to serve on empty host/urlPrefix.

Anyway, tomorrow I will start a new PR to show what I mean and it will be easier for me to let people explain to me why I should not do this or that ;)

Skrypt:
This one for next triage

PR #11824, opened 2022-06-07
Fix docs for how script is executed from code

### LINKED ISSUE #11814: Executing some script docs is outdated
### Describe the bug
The example in the docs on how to [evaluate code on this page](https://docs.orchardcore.net/en/latest/docs/reference/modules/Scripting/) seems to be outdated or does not work.

The example shows this way is to get current date using script.
```
var scriptingManager = serviceProvider.GetService<IScriptingManager>();
var date = scriptingManager.Evaluate("js: new Date().toISOString()");
```

However, the `IScriptingManager` does not expose a method that accepts `string` only.

Maybe the example should be as follow? Not sure if there is an easier way.

```
var scriptingManager = serviceProvider.GetService<IScriptingManager>();
var engine = scriptingManager.GetScriptingEngine("js");
var scope = engine.CreateScope(_scriptingManager.GlobalMethodProviders.SelectMany(x => x.GetMethods()), serviceProvider, null, null);
var date = engine.Evaluate("js: new Date().toISOString()");
```

Also it may be helpful to explain how to evaluate parameters using code? For example, how can I get the `AdminUsername`?

executing `engine.Evaluate(scope, "js: parameters('AdminUsername')"`  throws an exception `parameters is not defined.

CrestApps:
Fix #11814


hishamco (changes requested):

CrestApps:
@hishamco I made the requested changes

src/docs/reference/modules/Scripting/README.md
```diff
@@ -8,42 +8,48 @@ The scripting module provides an API allowing you to evaluate custom scripts in
 
 ### Executing some script
 
-The main interface is `IScriptingManager`.
+The main interface is [IScriptingManager](https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore/OrchardCore.Infrastructure.Abstractions/Scripting/IScriptingManager.cs).
 
 ```csharp
 public interface IScriptingManager
```
hishamco, line 14: Why we need the class here while you reference it above?

hishamco (changes requested):

src/docs/reference/modules/Scripting/README.md
CrestApps: I think it's helpful to see the docs that are in the file. Originally I left the reference only so if the code/docs changes, we don't have to maintain the code in multiple places. I don't know what is acceptable to you, but the PR is to make it where I though would be useful.

hishamco: The reference for `IScriptManager` is good, my comment was about `IScriptingEngine` no need to reference the code.

CrestApps: To make sure I understand the ask, you want me to:
1) Remove the GitHub reference to `IScriptingEngine`
2) Keep the GitHub reference to `IScriptingManager`? 
3) Keep the following code as part of this doc?
```csharp
{
    IScriptingEngine GetScriptingEngine(string prefix);
    object Evaluate(string directive, IFileProvider fileProvider, string basePath, IEnumerable<IGlobalMethodProvider> scopedMethodProviders);
    IReadOnlyList<IGlobalMethodProvider> GlobalMethodProviders { get; }
}
```

hishamco: Exactly except you need to remove `IScriptManager` code because you referenc to it

CrestApps:
@hishamco 
Closing this request in a favor of #11829

PR #11829, opened 2022-06-08
Update the Scripting docs to reflect the correct way to use IScriptingManager and IScriptingEngine

### LINKED ISSUE #11814: Executing some script docs is outdated

CrestApps:
Fix #11814

### DIFF
--- a/src/docs/reference/modules/Scripting/README.md
+++ b/src/docs/reference/modules/Scripting/README.md
@@ -14,18 +14,31 @@ The main interface is `IScriptingManager`.
 public interface IScriptingManager
 {
     IScriptingEngine GetScriptingEngine(string prefix);
-    object Evaluate(string directive);  
-    IList<IGlobalMethodProvider> GlobalMethodProviders { get; }
+
+    object Evaluate(string directive, IFileProvider fileProvider, string basePath, IEnumerable<IGlobalMethodProvider> scopedMethodProviders);
+
+    IReadOnlyList<IGlobalMethodProvider> GlobalMethodProviders { get; }
 }
 ```
 
 To evaluate an expression using a scripting engine, you must know which ones are available in the system.  
 For instance, a JavaScript one is available by default and its prefix is `js`.  
-To return the current date and time as a string we could do something like this:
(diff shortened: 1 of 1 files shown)


src/docs/reference/modules/Scripting/README.md
```diff
@@ -14,18 +14,31 @@ The main interface is `IScriptingManager`.
 public interface IScriptingManager
 {
     IScriptingEngine GetScriptingEngine(string prefix);
-    object Evaluate(string directive);  
-    IList<IGlobalMethodProvider> GlobalMethodProviders { get; }
```
hishamco, line 18: I think we agreed to reference the file and remove the code instead of update it each time when there's a change

CrestApps: Done.

CrestApps:
> BTW finally I see 1 commit instead of +10 ;)

This was by far harder than the commit in this PR :)

hishamco (approved):

PR #11838, opened 2022-06-09
localize AdminTemplates 

ProstoMimoProhodill:
I ran into a problem when localizing admin templates.
It is necessary to use "AdminTemplatesAdminMenu" intead  "AdminMenu" in Localizer.


hishamco (approved):
I mentioned this to @Piedone long time back, we need an nalyzer to help us on this, we faced similar issue on many places

PR #11841, opened 2022-06-09
Allow Default tenant’s admin to assign multiple feature-profiles to the same tenant

### LINKED ISSUE #10962: In a SaaS setup, how can the SaaS admin/owner manage the tenant features?

### LINKED ISSUE #10963: Improve the OrchardCore.Tenants.FeatureProfiles feature

MikeAlhayek:
This PR provide few enhancements to the Tenants modules discussed in #10963

Fix #10963

Here are some of the highlights.

 1. When using `FeatureProfile` add an `id` to the profile instead of using the existing editable name. This way renaming profile won't break anything. To keep it backward compatible, we treat current name as the id for old entries. Note renaming old entries won't cause any issue since we consider the id as the name before the new name is saved.
 2. Allow the user to have multiple profiles more on than explained in In a SaaS setup, how can the SaaS admin/owner manage the tenant features? Fix #10962

This PR to replace #10996


MikeAlhayek:
@sebastienros this has been setting for a while. Can you please take a look?

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -301,9 +301,8 @@ public async Task<IActionResult> Create()
             // Creates a default shell settings based on the configuration.
             var shellSettings = _shellSettingsManager.CreateDefaultSettings();
 
-            var currentFeatureProfile = shellSettings["FeatureProfile"];
-
-            var featureProfiles = await GetFeatureProfilesAsync(currentFeatureProfile);
+            var currentProfiles = shellSettings["FeatureProfile"]?.Split(',');
```
sebastienros, line 304:
```
static readonly separators = new [] {' ', ','};
Split(separators, StringSplitOptions.RemoveEmptyEntries)
```

src/OrchardCore.Modules/OrchardCore.Tenants/Services/FeatureProfilesManager.cs
```diff
@@ -21,17 +21,17 @@ public class FeatureProfilesManager
...
+            document.FeatureProfiles.Remove(id);
             await _documentManager.UpdateAsync(document);
         }
 
-        public async Task UpdateFeatureProfileAsync(string name, FeatureProfile profile)
+        public async Task UpdateFeatureProfileAsync(string id, FeatureProfile profile)
         {
             var document = await LoadFeatureProfilesDocumentAsync();
-            document.FeatureProfiles[name.ToLower()] = profile;
+            document.FeatureProfiles[id.ToLower()] = profile;
```
sebastienros, line 34: Remove Tolower (anywhere), and ok to make the dictionary case insensitive (if necessary)

src/OrchardCore.Modules/OrchardCore.Tenants/ViewModels/TenantViewModel.cs
```diff
@@ -22,7 +22,7 @@ public class TenantViewModel
 
         public string RecipeName { get; set; }
 
-        public string FeatureProfile { get; set; }
+        public string[] FeatureProfile { get; set; }
```
sebastienros, line 25: Rename the view model at least, it's not breaking anything

sebastienros (approved):

src/OrchardCore.Modules/OrchardCore.Tenants/ViewModels/TenantViewModel.cs
MikeAlhayek: I renamed the property like this `public string[] FeatureProfile { get; set; }` is this enough @sebastienros?

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -303,7 +301,7 @@ public async Task<IActionResult> Create()
             // Creates a default shell settings based on the configuration.
             var shellSettings = _shellSettingsManager.CreateDefaultSettings();
 
-            var currentProfiles = shellSettings["FeatureProfile"]?.Split(ProfileSeparators);
+            var currentProfiles = shellSettings["FeatureProfile"]?.Split(',', StringSplitOptions.TrimEntries | StringSplitOptions.RemoveEmptyEntries);
```
jtkech, line 304: Better I think to keep the 2 separators including the space as it allows to input `value1,value2` but also `value1, value2`, then `RemoveEmptyEntries` is no more necessary, as we do to split the tenant RequesturlHost in some places.

MikeAlhayek: @jtkech using the two separators will introduce a breaking change. Currently, a space can be part of the existing name which is stored in the settings. That would break. For example, if you have a profile with the name `test profile`, that would break the use if we trim using `',',' '` but using `.Split(',', StringSplitOptions.TrimEntries | StringSplitOptions.RemoveEmptyEntries)` will not be a breaking change.

jtkech: Ah okay if they are profiles names that may contain spaces, yes hosts don't have spaces, so maybe define a static string[] to split on `"," and ", "`, allowing at least one space after a comma.

MikeAlhayek: @jtkech I am now using `", "`

MikeAlhayek: @jtkech you good with the last changes?

jtkech:
@MikeAlhayek 

Oops sorry didn't have time to review this one again, but first need to complete #13040 as it is related to a breaking change, but there are many scenarios so it may take some times, I will try this week end.

MikeAlhayek:
@jtkech just a reminder to look at this when you have time. This one should be ready.

MikeAlhayek:
@jtkech just a reminder to look at this when you have time. This one should be ready.

MikeAlhayek:
@jtkech do you have any further update here?

jtkech:
Okay, will look at it next night

src/OrchardCore.Modules/OrchardCore.Tenants/TenantsConstants.cs
```diff
@@ -0,0 +1,6 @@
+namespace OrchardCore.Tenants;
+
+public class TenantsConstants
```
MikeAlhayek, line 3: I think it may be good idea to define something like this class at an upper level project like  `OrchardCore.Infrastructure.Abstractions`. This would allow us to reuse it anytime we want to split a string. Similar logic is defined in many places in the project

I don't know if `Abstractions` project is the best for this. But, it may not be a bad idea since it is used may most projects.

```C#
public class OrchardCoreConstants
{
    public const string StringSeperator = ", ";
}

public class OrchardCoreHelper
{
    public static string[] Split(string value)
    {
        return value?.Split(OrchardCoreConstants.StringSeperator, StringSplitOptions.TrimEntries | StringSplitOptions.RemoveEmptyEntries)
            ?? Array.Empty<string>();
    }

    public static string Join(param string[] values)
    {
        if(values == null) 
        {
            return null;
        }

        return String.Join(OrchardCoreConstants.StringSeperator, values);
    }
}
```
@jtkech thoughts?

jtkech: Yes if the comma and space are often used as separators, maybe a `StringExtensions` and `StringArrayExtensions` but not sure, the good thing being to use the same static separators.

Okay for joining, but for splitting better to use a static array `new[] { ',', ' ' }` so that you can separate by comma with or whithout a following space(s). And in that case you don't need `TrimEntries`, only `RemoveEmptyEntries`.

Having the right split `Separators` and join `Separator`, the usage being.

    someString?.Split(Seperators, StringSplitOptions.RemoveEmptyEntries) ?? Array.Empty<string>();
    String.Join(Seperator, someArray ?? Array.Empty<string>();

I did this in ShellSettings but dedicated to the list of urlHosts and only for the splitting part.

Hmm, but why not having helpers just for the null checks, `OrchardCoreHelper` is more intended to be used in razor views, so maybe extensions and if often used maybe in `OC.Abstractions`, maybe under the `Modules` folder.

MikeAlhayek: @jtkech  in this case, we can't use `new[] { ',', ' ' }` since previously we could have used a space int he id. For example `this is an old name` in this case `new[] { ',', ' ' }` would break which is why we need to use `", "` I guess in this PR we don't need to add `OrchardCoreHelper` but maybe to use in other PR with  `new[] { ',', ' ' }` under a different name.

src/OrchardCore/OrchardCore/Shell/FeatureProfilesValidationProvider.cs
```diff
@@ -49,12 +50,15 @@ public async ValueTask<bool> IsFeatureValidAsync(string id)
 
                     var feauterProfiles = await featureProfilesService.GetFeatureProfilesAsync();
```
jtkech, line 51: Typo feauterProfiles => featureProfiles

MikeAlhayek: Fixed. good catch

src/OrchardCore.Modules/OrchardCore.Tenants/TenantsConstants.cs
jtkech: Oops, okay I understand, but even if the comma separated list is not inputed by the user directly (if I understood correctly), can still be provided by a model in the ApiController, from a configuration source e.g. appsettings.json globally or for the AutoSetup, so I was thinking about the case for example if a comma is not followed by a space.

But maybe okay if the convention would be to use `, `.

Hmm, I think that in that case better to use a more dedicated separator, e.g. `|` and then we could split on `new[] { '|', ' ' }`, and document that we now can specify multiple profiles by separating them with `|`.

Sorry need to leave, will try to complete the review soon.

MikeAlhayek: @jtkech  Yes I understand. I think for consistency we should keep using a comma not a bar just because that is the command used separator. But if you want that change, we can do it. I'll wait your final review and we'll change anything needed. 

Currently everything is working as expect and is backward compatible.

MikeAlhayek:
@jtkech fixed conflicts we had.

src/OrchardCore.Modules/OrchardCore.Tenants/TenantsConstants.cs
jtkech: Anyway I was wrong, even if we use a `|` we still need to not split on space because of the old names, so my suggestion would be to split on `,` without space and then trim entries. About joining no matter so why not using this same separator. Then yes no need for an helper just to split / join on a single char, neither we need to use a const for a single char, only useful when we use an array.

> because that is the command used separator

Sorry I don't understand.

MikeAlhayek: @jtkech that is correct. Changing the separator in this case won't add any value. 

Let you me know you have any other requests or if you approve this PR

jtkech: What I meant is that you can split on a single char `','`, not a string having a space `", "`, e.g. if a source of data provides `"profile1,profile2"` in place of `"profile1, profile2"`.

Idem for joining by using the same single char `","` if it is just to persist the data as a comma separated string, still representing an array.

This everywhere and without defining a `TenantsConstants.FeatureProfileSeperator` const string.

    someString?.Split(',', StringSplitOptions.TrimEntries | StringSplitOptions.RemoveEmptyEntries)
        ?? Array.Empty<string>();

    String.Join(',', someArray ?? Array.Empty<string>());

Also, what you meant when saying: `because that is the command used separator`?

MikeAlhayek: > because that is the command used separator


There is a typo in there "because that is the common used separator"

I'll update the code when I get back to use a single comma instead of the constant.

jtkech: Okay cool then, for other things I trust @sebastienros which approved the PR.

MikeAlhayek: Yes I understand. I have not merged it because you had comments here so I want to make sure you are good with the changes.

MikeAlhayek: @jtkech I made the last changes. If I'll wait on merging this until tomorrow just in case you have any additional feedback.

thank you

PR #11842, opened 2022-06-09
Add descriptions to content-type Fix #11827

### LINKED ISSUE #11827: Add description to content types
I created `StudentGoal`, `StrengthGoal`, and few other content types, where each have an attached `BagPart` that would allow the user to add multiple notes to each goal. I also created `StudentGoalNote`, `StrengthGoalNote`, and other notes to allow a user with permission add notes "different kind of notes". The reason I have multiple different note types is because I want to different roles to have different permission (some roles my be able to add a specific note but not all type of notes.) However, since these `StudentGoalNote`, `StrengthGoalNote`, and others are nothing but note to the user, I want the display name to be "Note" so it makes sense on the UI for the user.  Below is a screenshot from the admin panel 
image

The above looks good and does exactly what I want. However, since all these notes have the same display name, this makes the content-type UI not friendly. Note this is not a duplicate problem, each of these note serves a different purpose but should be called "Note"

image

If the Content-type have description like `ContentPart` does, this would solve the problem and I would be able to add a description to the Content-Type describing what each note actually is used for.

If this something would be accepted, I would submit a PR.

CrestApps:
Add content description to the content-type just like we have for content-parts.

Accommodate #11827 

Here are some screenshots of the change

Screenshot_1
7
Screenshot_2

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentTypes/Views/ContentTypeSettings.Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.ContentTypes/Views/ContentTypeSettings.Edit.cshtml
@@ -1,6 +1,14 @@
 @model ContentTypeSettingsViewModel
 
 <div class="mb-3">
+    <div class="w-md-75 w-xl-50">
+        <label asp-for="Description">@T["Description"]</label>
+        <input asp-for="Description" type="text" class="form-control">
+    </div>
+    <span class="hint">@T["(Optional) Description of the type"]</span>
+</div>
+
+<div class="mb-3">
     <div class="form-check">
         <input type="checkbox" class="form-check-input" asp-for="Creatable">
         <label class="form-check-label" asp-for="Creatable">@T["Creatable"]</label>
--- a/src/OrchardCore.Modules/OrchardCore.ContentTypes/Views/ContentTypeSettings.Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.ContentTypes/Views/ContentTypeSettings.Edit.cshtml
@@ -5,7 +5,7 @@
         <label asp-for="Description">@T["Description"]</label>
         <input asp-for="Description" type="text" class="form-control">
     </div>
-    <span class="hint">@T["(Optional) Description of the type"]</span>
+    <span class="hint">@T["(Optional) Description of the content type"]</span>
 </div>
 
 <div class="mb-3">
--- a/src/OrchardCore.Modules/OrchardCore.ContentTypes/Views/Shared/DisplayTemplates/EditTypeViewModel.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.ContentTypes/Views/Shared/DisplayTemplates/EditTypeViewModel.cshtml
@@ -3,18 +3,28 @@
 @{
     var settings = Model.TypeDefinition.GetSettings<ContentTypeSettings>();
 }
-<div class="w-100">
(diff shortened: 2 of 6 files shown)


hishamco:
I'm not sure if there's a need to display a technical name?!! But agree with adding a description when it's needed

CrestApps:
@hishamco I thought it would be helpful. In my case, I could just look at the technical name without having to even add description. During the call today @sebastienros mention to add it.

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Metadata/Settings/ContentTypeSettings.cs
```diff
@@ -31,5 +31,10 @@ public class ContentTypeSettings
         /// Used to determine if this content type supports custom permissions
         /// </summary>
         public bool Securable { get; set; }
+
+        /// <summary>
+        /// Gets or sets the description name of the type.
```
hishamco, line 36: content type

src/OrchardCore.Modules/OrchardCore.ContentTypes/Views/Shared/DisplayTemplates/EditTypeViewModel.cshtml
```diff
@@ -3,18 +3,28 @@
...
+        <a id="btn-create-@Model.Name" asp-route-area="OrchardCore.Contents" asp-route-action="Create" asp-route-id="@Model.Name" asp-route-returnUrl="@FullRequestPath" class="btn btn-success btn-sm">@T["Create New {0}", Model.DisplayName]</a>
     }
+    <a role="btn-list-@Model.Name" asp-route-action="List" asp-route-area="OrchardCore.Contents" asp-route-contentTypeId="@Model.Name" class="btn btn-secondary btn-sm">@T["List Items"]</a>
+    <a role="btn-edit-@Model.Name" asp-route-action="Edit" asp-route-area="OrchardCore.ContentTypes" asp-route-id="@Model.Name" class="btn btn-primary btn-sm">@T["Edit"]</a>
+
 </div>
+
+@if (!string.IsNullOrWhiteSpace(settings.Description))
+{
+    <div class="row"><div class="col hint"> @settings.Description</div></div>
```
hishamco, line 29:
```html
@if (!string.IsNullOrEmpty(entry.Description))
{
    <div class="row">
        <div class="col">
            <p class="mb-0"><small class="text-muted">@settings.Description</small></p>
        </div>
    </div>
}
```

hishamco (changes requested):

src/OrchardCore.Modules/OrchardCore.ContentTypes/Views/ContentTypeSettings.Edit.cshtml
```diff
@@ -1,5 +1,13 @@
 @model ContentTypeSettingsViewModel
 
+<div class="mb-3">
+    <div class="w-md-75 w-xl-50">
+        <label asp-for="Description">@T["Description"]</label>
+        <input asp-for="Description" type="text" class="form-control">
+    </div>
+    <span class="hint">@T["(Optional) Description of the type"]</span>
```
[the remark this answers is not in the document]
CrestApps: done

hishamco (approved):

Skrypt (approved):

sebastienros (approved):

PR #11869, opened 2022-06-14
(OLD) Provide the Default Tenant’s admin a way to manage tenant’s features

### LINKED ISSUE #11870: Ability to enable/disable feature for Tenant from Host
I want to create a tenant with the following features disabled `OrchardCore.Features` , `OrchardCore.Workflows` and `OrchardCore.Deployments`. I have a SaaS environment and I want to be able to manage the tenant's features from the host.

Create a new feature that would allow the host admin to enable/disable features for tenants.

MikeAlhayek:
Add a new feature that would allow the host to manage tenant's features 
Fix #11870 

This PR enhanced the `OrchardCore.Features` module by allowing to accept `{tenant?}` is the route. This tenant value would only be evaluated if the request is running on the host. When a tenant value is present, we'll execute the code in the scope of the given tenant.

Here is a screen shot from the Tenants list in the host the "Manage Features" button was added
image

listing features for site 1 example
image

This is how the features show up when it is running for a tenant and not the host "no changes to the outcome"
image


OrchardCore.sln
```diff
@@ -464,9 +464,11 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Examples.Themes.AssyAttrib.
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Examples.OrchardCoreThemes.Alpha", "test\OrchardCore.Tests.Themes\Examples.OrchardCoreThemes.Alpha\Examples.OrchardCoreThemes.Alpha.csproj", "{ECCBDB6C-BF57-43B8-9A58-5F3D60357D0D}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OrchardCore.FileStorage.AmazonS3", "src\OrchardCore\OrchardCore.FileStorage.AmazonS3\OrchardCore.FileStorage.AmazonS3.csproj", "{38F43FA0-5BA8-4D6B-8F66-C708D590EF76}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OrchardCore.FileStorage.AmazonS3", "src\OrchardCore\OrchardCore.FileStorage.AmazonS3\OrchardCore.FileStorage.AmazonS3.csproj", "{38F43FA0-5BA8-4D6B-8F66-C708D590EF76}"
```
sebastienros, line 467: Revert these lines

OrchardCore.sln
```diff
@@ -1236,6 +1238,10 @@ Global
 		{FF1C550C-6D30-499A-AF11-68DE7C8B6869}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{FF1C550C-6D30-499A-AF11-68DE7C8B6869}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{FF1C550C-6D30-499A-AF11-68DE7C8B6869}.Release|Any CPU.Build.0 = Release|Any CPU
+		{A6563050-EE6D-4E7C-81AA-C383DB3ED124}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
sebastienros, line 1241: Also unexpected changes

src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Index.cshtml
```diff
@@ -110,61 +77,20 @@
...
-                                }
-                                @if (!entry.IsDefaultTenant && entry.ShellSettings.State == TenantState.Running)
-                                {
-                                    <a asp-action="Disable" asp-route-id="@entry.Name" class="btn btn-danger btn-sm disable" data-url-af="RemoveUrl UnsafeUrl" data-title="@T["Disable"]" data-message="@T["Are you sure you want to disable this tenant?"]" data-ok-text="@T["Ok"]" data-cancel-text="@T["Cancel"]">@T["Disable"]</a>
-                                }
-                                @if (entry.ShellSettings.State == TenantState.Disabled)
-                                {
-                                    <a asp-action="Enable" asp-route-id="@entry.Name" class="btn btn-success btn-sm enable" data-url-af="UnsafeUrl">@T["Enable"]</a>
+                                @{
+                                    var buttonShape = await DisplayManager.BuildDisplayAsync(entry, UpdateModelAccessor.ModelUpdater, String.Empty, "ActionButtons");
```
sebastienros, line 90: The controller is an updated so it doesn't have to be injected. Or better, move the calls to the display manager in the controller, and pass the shapes to the view model.

sebastienros:
Can you also check if the current tenant controller/views can be adapted to support any tenant instead of the default one only? This way the new feature could use them too instead of duplicating the management UI.

src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Index.cshtml
CrestApps: @sebastienros I changed this by creating a shape instead of using Display Drivers. Let me know if you see a better way to doing this

OrchardCore.sln
CrestApps: I am still trying to figure out why Visual Studio keeps on changing this. I'll update this one I find a solutions.

CrestApps: @sebastienros I don't know why this is happening. But VS 2022 seems to always want to change this. I manually edited the `.sln` file and reverted these changes and committed it outside of VS "otherwise VS will revert it back". Any way, this should be fixed.

hishamco:
Is it a shortcut for features, instead of going to each tenant and configure its features? If YES, I suggest to consider improving the UI/UX the buttons take a BIG room. May be dropdown menu can make the UI more cleaner

CrestApps:
@hishamco yes. This is a shortcut to features for the given tenant. I don’t mind changing the UI but not sure if it will be acceptable. We can make it a drop down menu, but what functions would you want to move into the menu? Eventually, “continuous PR after this one” we would want to also add a shortcut to “execute recipes” and “recipes” too. Would you want a button called “Manage” and that button would have “Reload Site”, “Features”, then eventually we can add “Execute recipe”…?

Either way, I think we may want to leave the button as is for now, and I’ll take the drop down button into the second PR when more links are added.

Skrypt:
The UI enhancement is a non-blocking issue. We can take a second PR later on.
I would use a dropdown menu as we do in the Content Items list. Name it "Actions".
I would keep the "Edit, View" buttons as single items and move everything else inside that "Actions" button.

hishamco:
Believe it or not I hate the colored buttons when they take more room ;) I think we need to review the color scheme from UI/UX perspective

CrestApps:
@Skrypt @sebastienros  @hishamco  are you able to revisit this PR please?

hishamco:
We 're watching, sometimes the PRs takes time to review. Don't forgot we have limited resources and almost the guys might have a FT job but we are struggling to help to community

Just a little note for myself and others don't ping too much the team got notified when something change on any issue or pull request

jtkech:
I think that the `Manage Features` buttons should be rendered only if their related tenant is `Running`, not for example if `Uninitialized` or `Disabled`. And also for tenants having the `OC.Features` enabled. 

Hmm, then why not just generating urls pointing to the `OC.Features` index page of each tenant, here as above only for `Running` tenants and having the `OC.Features` enabled.

Then while in the `OC.Features` page of a given tenant, provide a link to return back to the `OC.Tenants` index page of the `Default` tenant if the `OC.Tenants` feature is enabled in the `Default` tenant.

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -103,7 +103,6 @@ public async Task<IActionResult> Index(TenantIndexOptions options, PagerParamete
                        Description = x["Description"],
                        Name = x.Name,
                        ShellSettings = x,
-                       IsDefaultTenant = x.IsDefaultShell(),
```
jtkech, line 106: Why?

jtkech: Hmm yes because you now test it in a view

MikeAlhayek: Yup

MikeAlhayek:
@jtkech i’ll add the condition to show the button only when the tenant is running. Also I’ll check in the OC.Features if the given tenant is running too. 

I don’t think we need to check if OC.Features is enabled on the given tenant. The idea here is to be able to manage a tenants feature on the behalf of the tenant. Sometimes we don’t want the tenant to be able to do that at all.

jtkech:
@MikeAlhayek 

Okay, I'll try to look at it more in depth if I have time.

But what you think about the main part of my previous comment? Here in summary.

> ... why not just generating urls pointing to the OC.Features index page of each tenant .... Then while in the OC.Features page of a given tenant, provide a link to return back to the OC.Tenants index page of the Default tenant ...

MikeAlhayek:
> why not just generating urls pointing to the OC.Features index page of each tenant .... Then while in the OC.Features page of a given tenant, provide a link to return back to the OC.Tenants index page of the Default tenant ...

Wouldn't that require the host-admin to login to the tenants website as admin to perform the change then? What if the `OC.Features` is disabled on the tenant site?

Currently, the operation is handler directly on tenants website from the host admin without the need to login to the tenant or event have the `OC.Features` enabled for that given tenant.

MikeAlhayek:
@jtkech I added the state check in both the `OC.Tenants` view and also the `OC.Features` controller so now the requests can't be processed on a not-running tenants.

However, I think we need `returnUrl` here. because `OC.Tenants` has `List` action with no submit form like "Publish and continue" scenario. One the user is on the List admin, then can enable/disable features. But they they stay on the List action. From the list action, we can't really redirect them back to where they came from (i.e. `OC.Tenants`) since there is no real function that would use that currently on `OC.Tenants. 

Here is a screenshot after the update
image

jtkech:
Okay good points, I will try your branch when I will have time.

In the meantime, maybe a shorter text for the action buttons `Manage Features` => maybe just `Features`.

Yes, we have at least the `Default` tenant and we already have functionalities that rely on it, so it is not exactly like other tenants, but for many other things it is like other tenants, for example how its isolated tenant container and pipeline are built.

Just to say that maybe not right to talk about the `Default` tenant as the `Host` tenant, maybe the `Main` tenant or just always the `Default` tenant, for me `Host` refers for example to all services that are registered at the host level from the application.

MikeAlhayek:
I’ll rename that button. But eventually we’ll add “Action” button and move “Features”, “Disable”, “Reload”, and “Disable” button in that action menu. I was thinking to do that with a follow up PR after this one is merged.

MikeAlhayek:
Closing this in favor of PR #12360 which has no conflicts

PR #11881, opened 2022-06-19
Log DocumentManager.UpdateAsync() exception

### LINKED ISSUE #11873: DocumentManager hides exception
DocumentManager.UpdateAsync is currently hiding exceptions that occur when the distributed cache throws an exception.  

It would be helpful if in the code below the current exception was provided to InvalidOperationException as an InnerException.

```
                try
                {
                    _ = await _distributedCache.GetStringAsync(_options.CacheIdKey);
                }
                catch
                {
                    await DocumentStore.CancelAsync();

                    throw new InvalidOperationException($"Can't update the '{typeof(TDocument).Name}' if not able to access the distributed cache");
                }

```

hishamco:
Fixes #11873

### DIFF
--- a/src/OrchardCore/OrchardCore.Infrastructure/Documents/DocumentManager.cs
+++ b/src/OrchardCore/OrchardCore.Infrastructure/Documents/DocumentManager.cs
@@ -169,6 +169,8 @@ namespace OrchardCore.Documents
                 {
                     await DocumentStore.CancelAsync();
 
+                    _logger.LogError(e, "Failed to update the '{DocumentName}'", typeof(TDocument).Name);
+                    
                     throw new InvalidOperationException($"Can't update the '{typeof(TDocument).Name}' if not able to access the distributed cache");
                 }
             }
--- a/src/OrchardCore/OrchardCore.Infrastructure/Documents/DocumentManager.cs
+++ b/src/OrchardCore/OrchardCore.Infrastructure/Documents/DocumentManager.cs
@@ -165,7 +165,7 @@ namespace OrchardCore.Documents
                 {
                     _ = await _distributedCache.GetStringAsync(_options.CacheIdKey);
                 }
-                catch
+                catch(Exception e)
                 {
                     await DocumentStore.CancelAsync();
 
--- a/src/OrchardCore/OrchardCore.Infrastructure/Documents/DocumentManager.cs
+++ b/src/OrchardCore/OrchardCore.Infrastructure/Documents/DocumentManager.cs
@@ -165,13 +165,13 @@ namespace OrchardCore.Documents
                 {
                     _ = await _distributedCache.GetStringAsync(_options.CacheIdKey);
                 }
-                catch(Exception e)
+                catch
                 {
                     await DocumentStore.CancelAsync();
 
-                    _logger.LogError(e, "Failed to update the '{DocumentName}'", typeof(TDocument).Name);
+                    _logger.LogError("Can't update the '{DocumentName}' if not able to access the distributed cache", typeof(TDocument).Name);
                     
-                    throw new InvalidOperationException($"Can't update the '{typeof(TDocument).Name}' if not able to access the distributed cache");
+                    throw;
                 }
             }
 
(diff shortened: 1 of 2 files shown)


src/OrchardCore/OrchardCore.Infrastructure/Documents/DocumentManager.cs
```diff
@@ -165,10 +165,12 @@ public async Task UpdateAsync(TDocument document, Func<TDocument, Task> afterUpd
...
                 }
-                catch
+                catch(Exception e)
                 {
                     await DocumentStore.CancelAsync();
 
+                    _logger.LogError(e, "Failed to update the '{DocumentName}'", typeof(TDocument).Name);
+                    
                     throw new InvalidOperationException($"Can't update the '{typeof(TDocument).Name}' if not able to access the distributed cache");
                 }
```
jtkech, line 175:
```suggestion
                catch (Exception ex)
                {
                    await DocumentStore.CancelAsync();

                    throw new InvalidOperationException(
                        $"Can't update the '{typeof(TDocument).Name}' if not able to access the distributed cache",
                        innerException: ex);
                }
```

Then we can do the same in `VolatileDocumentManager.cs`.

hishamco: +1, but log an exception may need too

jtkech: Logging the error would be needed if we do a catch and then don't re-throw, but here the whole exception that we throw will be naturally logged.

Here I needed to catch to do a `CancelAsync()`, and then the idea was to not overfill the log file with the full exception message when you can't connect to redis, because in that case you have the same logged exception from other places, but maybe still useful when the issue is intermitent.

So another suggestion would be to just do the following, where we do a `CancelAsync()` as needed, and where the exception is re-thrown and then will be logged as is.

                catch
                {
                    await DocumentStore.CancelAsync();

                    throw;
                }

Or do the following if still useful to have an higher specific message, knowing that the initial goal of this specific message was to replace the long exception message by a shorter one.

                catch
                {
                    await DocumentStore.CancelAsync();

                    _logger.LogError("Can't update the '{DocumentName}' if not able to access the distributed cache", typeof(TDocument).Name);

                    throw;
                }

Note: In some rare context (not in a given request/bg task), e.g. a kind of background job, an exception may not be naturally logged, in that case this is at an higher level (e.g. a job manager) that the exception needs to be catched and logged, and then re-thrown or not depending on the usage.

jtkech: Hmm, maybe the last suggestion is the best compromise, in the catch block we do the cancel that we need, then we are sure to have a looged error as we do it explicitly, then we just re-throw the original exception.

hishamco: Can we expose the exception object to the logger?

jtkech: But then it will be logged twice, one time here explicitly, another time as we re-throw, like any other exception that we don't catch ourselves but are still logged as is at the end of the request.

src/OrchardCore/OrchardCore.Infrastructure/Documents/DocumentManager.cs
```diff
@@ -169,7 +169,9 @@ public async Task UpdateAsync(TDocument document, Func<TDocument, Task> afterUpd
                 {
                     await DocumentStore.CancelAsync();
 
-                    throw new InvalidOperationException($"Can't update the '{typeof(TDocument).Name}' if not able to access the distributed cache");
+                    _logger.LogError("Can't update the '{DocumentName}' if not able to access the distributed cache", typeof(TDocument).Name);
+                    
+                    throw;
```
jtkech, line 174: Usually we don't log when we throw because when we throw it will be logged.

hishamco: This is based on your suggestion above ;)

jtkech: Which suggestion? Where?

hishamco: [Here](https://github.com/OrchardCMS/OrchardCore/pull/11881#discussion_r901043702)

jtkech: I re-read the comments, okay then ;)

PR #11905, opened 2022-06-27
Fixes #11904 TypeFeatureProvider throws if a Razor Page has no PageModel

### LINKED ISSUE #11904: TypeFeatureProvider throws if a Razor Page has no PageModel
### Describe the bug
We now can use the `[Feature()]` attribute on a razor `PageModel` and we use the `ITypeFeatureProvider` to retrieve the feature related to a given `PageModel`. The problem is that when no `PageModel` is defined for a razor page, e.g. only `@functions` handlers, the related page `ModelType` is then of type `Object`.

In that case `ITypeFeatureProvider.GetFeatureForDependency(pageModelType)` throw an exception.

### To Reproduce
1. Create a Razor @page without defining a `PageModel` (e.g. only `@functions` handlers).
2. Start the app and see `_typeFeatureProvider` failing.

### Expected behavior
A Razor @page without defining a `PageModel` should work.

Fixed on my side, will push a PR

jtkech:
Fixes #11904 

We now can use the `[Feature()]` attribute on a razor `PageModel` if it is a **component**, meaning a non abstract public class, e.g. not an internal class when defined in a razor page itself, and we use the `ITypeFeatureProvider` to retrieve the feature related to a given `PageModel`.

The problem is that when no `PageModel` is defined for a razor page, e.g. only `@functions` handlers, the related page `ModelType` is then of type `Object` which is non abstract and public, so considered as a **component**.

In that case `ITypeFeatureProvider.GetFeatureForDependency(pageModelType)` throw an exception.

A quick solution was to change the following by also checking the `Object` type.

        private bool IsComponentType(Type type)
        {
            return type.IsClass && !type.IsAbstract && type.IsPublic &&
                type != typeof(Object); <= The change
        }

But I think it is better and simpler to just check if the page `ModelType` has a custom attribute of type `FeatureAttribute`, and if so just use it to retrieve the related feature.

@ns8482e if you can take a look ;)

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Demo/Pages/HelloWorld.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Demo/Pages/HelloWorld.cshtml
@@ -1,0 +1,25 @@
+@page "/helloworld"
+
+@functions
+{
+    public Task<IActionResult> OnGetAsync()
+    {
+        return Task.FromResult<IActionResult>(Page());
+    }
+}
+
+@{
+    var title = "This is a razor 'Page' coming from OrchardCore.Demo that doesn't define any 'PageModel'.";
+    var message = $"Date and time on the server: { DateTime.Now.ToString() }";
+}
+
+@await Html.PartialAsync("Header", new { Title = title })
+
+<h2>Hello World from @ViewContext.RouteData.Values["page"]</h2>
+
+<p>
+    @message
+</p>
+<p>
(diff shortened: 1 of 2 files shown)


ns8482e:
Nice - Can you add sample feature in OrchardCore.Demo?

jtkech:
@ns8482e Okay will do

ns8482e (approved):
LGTM

jtkech:
@ns8482e Okay, done and tried ;)

Skrypt (approved):

PR #11921, opened 2022-06-29
ReadAsByteArray support for Workflows

### LINKED ISSUE #4634: Orchard Core - Uploading Image from Code.
I have recently started learning Orchard Core. Everything is working fine if I use the Admin Console to create the content to upload Images. Here I have Created
Untitled
This is my UI from where I am able to save Images using default orchard functionality.

Here are my questions:-
1. Can we Upload this Image from code instead?
2. How Can I Upload Image from a form using workflow? (Consider the attached Image, I want to post this form to the workflow and save all the form fields in DB)
image
Below is my workflow form which is not currently working. (I was trying posting these form data to be saved from my API controller but It failed. Noticed - I have sent a POST request to my API)
image

lampersky:
it would be great to have additional switch inside **HttpRequestTask** and **HttpResponseTask**, which will allow users to choose if content is binary or not, currently both tasks allow us to only use string there, which is quite a big limit

this will allow us to:
- download binary data like images etc. using **HttpRequestTask**,
- return binary content like gifs, pdfs using **HttpResponseTask**,
- return binary content using scripting **rawResponseWrite** method.

Here you can check those extended tasks in action:

anim

Fixes #4634


Skrypt:
Do you agree with me here that binary could be anything?

lampersky:
> Do you agree with me here that binary could be anything?

yep, it could be anything

lampersky:
but this small extension is not related with media module...

Skrypt:
What you want is to add the option to use a `<form enctype="multipart/form-data">` with the Workflows.

Skrypt:
Can you replace your demo with a .mpg file produced with ScreenToGif? I'm having a hard time following the .gif if I can't pause it.

lampersky:
@Skrypt before this PR you were allowed to only return **string** from the **HttpResponseTask**, which is great when you are playing with html, json, in general with text files, but sometimes you may want to return non-text file with some binary content like PDF or as mentioned before images

same with **HttpRequestTask** currently you can only perform request to endpoint which returns text content (json, xml, text, html...) but when you will try to perform request to some endpoint which returns binary content, you will find problem with encodings etc. because current solution is built to work only with endpoints which are returning text content

instead of preparing mpg, I've extracted all frames from gif for you:









4a







7a







10a

src/OrchardCore.Modules/OrchardCore.Workflows/Http/Activities/HttpRequestTask.cs
```diff
@@ -185,14 +191,28 @@ public override async Task<ActivityExecutionResult> ExecuteAsync(WorkflowExecuti
             var responseCodes = ParseResponseCodes(HttpResponseCodes);
             var outcome = responseCodes.FirstOrDefault(x => x == (int)response.StatusCode);
 
-            workflowContext.LastResult = new
+            if (IsBinary)
```
Skrypt, line 194: I don't think it is necessary to have an option on the HttpRequestTask for this. The form post itself should be set with an enctype="multipart/form-data" to indicate that it is sending data.

lampersky: @Skrypt this PR is not related to form and enctype.

it is just extending HttpRequestTask to allow us to read binary content, as mentioned before, without those changes you can only fetch and read text files

lampersky: I can't find example for dotnet, but here you can read about javascript equivalent for reading binary data:

https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Sending_and_Receiving_Binary_Data

Skrypt: Ok, but your XMLHttpRequest here is defining the response content type. Just like a form can set the content type of its data sent. Here, you are then missing the point that this is a value that should be taken from the HttpContext itself to set the Body of the workflowContext.LastResult. Else, we will need an option for Binary, blob, document, JSON, string, empty.

lampersky: > Here, you are then missing the point that this is a value that should be taken from the HttpContext itself to set the Body of the `workflowContext.LastResult`

it is there, few lines below

image

Skrypt: Yeah, I saw that. But you still don't get my point.

lampersky: true, I can't get your point, but I'm sure you will help me to do so :P

again, this PR is just about adding another way of reading body passed through Http request, now you can choose if body should be read as string (text) or as byte array (binary)

> Also, you need to take into consideration that this HttpRequest can also come from an HTML form with an enctype set to "multipart/form-data".

usually, this type of requests are returning some text body like json etc.

Skrypt: When you are using an `XMLHttpRequest` it is documented that it is normally possible to use the XMLHttpRequest.responseType but it seems that ASP.NET core doesn't care... it won't affect the "Accept" header of the HttpRequest.

The Accept header is meant to say that this HttpRequest will Accept a response of X content-type.

So, another way that would work is to set the "Accept" header to "application/octet-stream".

```javascript
var xhr = new XMLHttpRequest();
xhr.open('GET', '/workflows/Invoke?token=CfDJ8On1YqmVp8VLlfa7K1eDOI3SK1AVDWzbsdHVsZA6kFIS7u2v2Goh6Qjm9GzWgbEGAG_0SPBYm6tYrHR-PeyLiKQ-POdFrq3ZVOh8TIIAXfZnDuBL3wW-k78kw_jrHR-3WHGUZ6RQxE_cCbeWnetndefB8TQ4oEKIkKfvUbmdTx34EVxNeMFkyGXYx4V3NhpnVURQHYTHTH_mB4HeoD0TeW4mdamtC766Xxnzyav8-eV-', true);
xhr.responseType = 'arraybuffer'; // doesn't work with .NET
xhr.setRequestHeader("Accept", "application/octet-stream");
xhr.send(null);
```

Then, the only thing you have to is:

```c#
var acceptHeader = HttpContext.Request.Headers.TryGetValue("Accept", out var acceptValue);
```

Then return the appropriate response.content based on the value of the acceptHeader.

Skrypt: And also, using this method, you can also now support multiple accepted Content Types.

Skrypt: Here, also, as documented, the XMLHttpRequest.responseType will prevent that data be anything else than binary when set to "arrayBuffer". It will return null if the data sent from the server is not binary.

http://man.hubwiz.com/docset/JavaScript.docset/Contents/Resources/Documents/developer.mozilla.org/en-US/docs/Web/API/XMLHTTPRequest/responseType.html

Skrypt: Or in this case here maybe we need to use "Content-Type" header if we want to only read the value of the HttpRequest. But then the "Accept" header would be for the HttpResponse.

src/OrchardCore.Modules/OrchardCore.Workflows/Views/Items/HttpRequestTask.Fields.Edit.cshtml
```diff
@@ -55,6 +55,16 @@
     <span class="hint">@T["A comma-separated list of HTTP response status codes to handle as activity outcomes. For example: 200, 404, 500"]</span>
 </div>
 
+<div class="mb-3 form-group">
+    <div class="custom-control custom-checkbox">
+        <input type="checkbox" class="custom-control-input" asp-for="IsBinary" checked="@Model.IsBinary" />
+        <label class="custom-control-label" asp-for="IsBinary">@T["Is Binary"]</label>
+        <span class="hint dashed">
+            @T["Is Binary"]
```
Piedone, line 63: If this remains (but reading a bit your convo I'm not sure), then add an explanation of why this option is useful and when would you use it. Also for `HttpResponseTask`

Piedone (commented):
Is this something you'd like to revisit any time soon @lampersky or should we close?

src/OrchardCore.Modules/OrchardCore.Workflows/Views/Items/HttpRequestTask.Fields.Edit.cshtml
lampersky: There are many use cases:
- creating a zip file in your workflow script and return it,
- creating an image in your workflow script and return it,
- accessing an external API from your workflow script which results binary response.

If this PR doesn't have any value for others, let's close it. I've just shared a working solution that handles a few use cases I've encountered in the past.

Piedone: I'm not arguing against the usefuleness of being able to send a binary response. By "this" I meant the commented "Is Binary" checkbox. My comment is about if this checkbox remains in its current form, as opposed to, e.g., having a dropdown with more options, or some auto-detection.

Issue #11922, opened 2022-06-29
Display the feature name instead of the id when listing the permission in admin


CrestApps:
### Is your feature request related to a problem? Please describe.
When the permissions are listed during role edit/create, currently we show the feature Id instead of a name. Displaying the feature id is not user friendly.

![image](https://user-images.githubusercontent.com/24724371/176494086-2a884a31-3a28-4685-8bed-7193527cfa64.png)

### Describe the solution you'd like
I think the permissions should be grouped by feature id but we should show the feature name if one is available

CrestApps:
@Skrypt and yes the id will be more unique which is why I suggest we still group by id but display the name. we could show the id in a lighter color if needed.

Skrypt:
Ok, if we want to make this right. We need to display both, I agree. Then something like : 

### OrchardCore.Media (id)
#### Media Feature (name)

CrestApps:
@Skrypt How does this look?
![image](https://user-images.githubusercontent.com/24724371/176502525-9605eb39-2a38-49ed-b86f-97cdc646ef25.png)

Skrypt:
![image](https://user-images.githubusercontent.com/3228637/176503118-365c06e7-767e-4a02-9a61-627e8ac0f7a9.png)

Skrypt:
If we keep grouping by Feature Id then the Feature Id should be displayed before the Feature Name.

Skrypt:
Yes, I don't think it will break anything.

PR #11923, opened 2022-06-29
Change how the permissions are grouped and sorted

CrestApps:
Accommodate #11922 

image

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Roles/ViewModels/EditRoleViewModel.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Roles/ViewModels/EditRoleViewModel.cs
@@ -9,7 +9,11 @@ namespace OrchardCore.Roles.ViewModels
     {
         public string Name { get; set; }
         public string RoleDescription { get; set; }
-        public IDictionary<string, IEnumerable<Permission>> RoleCategoryPermissions { get; set; }
+
+        [BindNever]
+        public IDictionary<PermissionGroupKey, IList<Permission>> RoleCategoryPermissions { get; set; }
+
+        [BindNever]
         public IEnumerable<string> EffectivePermissions { get; set; }
 
         [BindNever]
--- a/src/OrchardCore.Modules/OrchardCore.Roles/ViewModels/PermissionGroupKey.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Roles/ViewModels/PermissionGroupKey.cs
@@ -1,0 +1,40 @@
+using System;
+
+namespace OrchardCore.Roles.ViewModels;
+
+public class PermissionGroupKey
+{
+    private readonly string _key;
+
+    public string Title { get; set; }
+
+    public string Source { get; set; }
+
+    public PermissionGroupKey(string key)
+    {
+        if (String.IsNullOrWhiteSpace(key))
+        {
+            throw new ArgumentException(nameof(key));
+        }
+
+        _key = key;
+    }
+
+    public PermissionGroupKey(string key, string title)
+        : this(key)
+    {
+        Title = title;
+    }
+
+    public override int GetHashCode()
+    {
+        return _key.GetHashCode();
+    }
+
+    public override bool Equals(object obj)
+    {
+        var other = obj as PermissionGroupKey;
+
+        return _key.Equals(other?._key);
+    }
+}
--- a/src/OrchardCore.Modules/OrchardCore.Roles/Views/Admin/Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Roles/Views/Admin/Edit.cshtml
@@ -3,7 +3,7 @@
 
 <zone Name="Title"><h1>@RenderTitleSegments(T["Edit Role {0}", Model.Name])</h1></zone>
 
-<form autocomplete="off">
+<form autocomplete="off" class="mb-3">
 
     <div class="row">
         <div class="mb-3 col-xl-4 col-sm-6">
@@ -15,14 +15,14 @@
 </form>
 
 <form asp-action="Edit" method="post" class="no-multisubmit">
-    <h5>@T["Edit Role Detail"]</h5>
+    <h2>@T["Edit Role Detail"]</h2>
     <div class="mb-3" asp-validation-class-for="RoleDescription">
         <label asp-for="RoleDescription">@T["Role description"]</label>
         <input asp-for="RoleDescription" class="form-control" />
     </div>
-    <hr>
+    <hr class="my-3">
 
-    <h5>@T["Role permissions"]</h5>
+    <h2>@T["Role permissions"]</h2>
 
     <div class="w-100 ms-2">
         @T["Allow"] <span class="hint dashed">@T["Permission is granted explicitly."]</span>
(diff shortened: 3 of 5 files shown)


Skrypt:
Looks good. Will wait for triage on this one.

Skrypt (approved):

src/OrchardCore.Modules/OrchardCore.Roles/ViewModels/PermissionGroupKey.cs
```diff
@@ -0,0 +1,40 @@
...
+        _key = key;
+    }
+
+    public PermissionGroupKey(string key, string title)
+        : this(key)
+    {
+        Title = title;
+    }
+
+    public override int GetHashCode()
```
hishamco, line 29:
```csharp
public override int GetHashCode() => _key.GetHashCode();
```

src/OrchardCore.Modules/OrchardCore.Roles/ViewModels/PermissionGroupKey.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        if (String.IsNullOrWhiteSpace(key))
+        {
+            throw new ArgumentException(nameof(key));
+        }
+
+        _key = key;
+    }
+
+    public PermissionGroupKey(string key, string title)
```
hishamco, line 23: public PermissionGroupKey(string key, string title) : this(key) => Title = title;

src/OrchardCore.Modules/OrchardCore.Roles/ViewModels/PermissionGroupKey.cs
```diff
@@ -0,0 +1,40 @@
+using System;
+
+namespace OrchardCore.Roles.ViewModels;
+
+public class PermissionGroupKey
+{
+    private readonly string _key;
+
+    public string Title { get; set; }
```
hishamco, line 9: Make it readonly

src/OrchardCore.Modules/OrchardCore.Roles/ViewModels/PermissionGroupKey.cs
```diff
@@ -0,0 +1,40 @@
...
+
+    public string Source { get; set; }
```
hishamco, line 11: Where this is used in the class?!!

src/OrchardCore.Modules/OrchardCore.Roles/ViewModels/PermissionGroupKey.cs
```diff
@@ -0,0 +1,40 @@
...
+    public override int GetHashCode()
+    {
+        return _key.GetHashCode();
+    }
+
+    public override bool Equals(object obj)
+    {
+        var other = obj as PermissionGroupKey;
+
+        return _key.Equals(other?._key);
```
hishamco, line 38: This is fine if the key is supplied, but what if the second constructor called, I think title should be considered here as well

src/OrchardCore.Modules/OrchardCore.Roles/Views/Admin/Edit.cshtml
```diff
@@ -15,14 +15,14 @@
 </form>
 
 <form asp-action="Edit" method="post" class="no-multisubmit">
-    <h5>@T["Edit Role Detail"]</h5>
```
hishamco, line 18: Why you change the header size?

src/OrchardCore.Modules/OrchardCore.Roles/Views/Admin/Edit.cshtml
```diff
@@ -15,14 +15,14 @@
...
+    <h2>@T["Edit Role Detail"]</h2>
     <div class="mb-3" asp-validation-class-for="RoleDescription">
         <label asp-for="RoleDescription">@T["Role description"]</label>
         <input asp-for="RoleDescription" class="form-control" />
     </div>
-    <hr>
+    <hr class="my-3">
 
-    <h5>@T["Role permissions"]</h5>
```
hishamco, line 25: Why you change the header size?

src/OrchardCore.Modules/OrchardCore.Roles/ViewModels/EditRoleViewModel.cs
```diff
@@ -9,7 +9,11 @@ public class EditRoleViewModel
     {
         public string Name { get; set; }
         public string RoleDescription { get; set; }
-        public IDictionary<string, IEnumerable<Permission>> RoleCategoryPermissions { get; set; }
+
+        [BindNever]
+        public IDictionary<PermissionGroupKey, IList<Permission>> RoleCategoryPermissions { get; set; }
```
hishamco, line 14: Could be `IEnumerable<Permission>`?

src/OrchardCore.Modules/OrchardCore.Roles/ViewModels/PermissionGroupKey.cs
CrestApps: The class is used in the controller for grouping

src/OrchardCore.Modules/OrchardCore.Roles/ViewModels/PermissionGroupKey.cs
CrestApps: the second constructor calls the first one. Key would never be null

src/OrchardCore.Modules/OrchardCore.Roles/Views/Admin/Edit.cshtml
CrestApps: because we changed how the title appears. I thought it would look better if the header has a different size than the actual legend labels

src/OrchardCore.Modules/OrchardCore.Roles/ViewModels/PermissionGroupKey.cs
hishamco: I mean the `Source` property

src/OrchardCore.Modules/OrchardCore.Roles/ViewModels/PermissionGroupKey.cs
hishamco: Right, I was talking about `Title`, which is not taking into account here

src/OrchardCore.Modules/OrchardCore.Roles/ViewModels/PermissionGroupKey.cs
```diff
@@ -4,9 +4,9 @@ namespace OrchardCore.Roles.ViewModels;
 
 public class PermissionGroupKey
 {
-    private readonly string _key;
+    public readonly string Key;
 
-    public string Title { get; set; }
+    public readonly string Title;
```
hishamco, line 9: It would be nice to make it readonly property

```csharp
public string Title { get; }
```

src/OrchardCore.Modules/OrchardCore.Roles/ViewModels/PermissionGroupKey.cs
```diff
@@ -17,24 +17,20 @@ public PermissionGroupKey(string key)
...
-        return _key.GetHashCode();
-    }
+    public override int GetHashCode() => Key.GetHashCode();
 
     public override bool Equals(object obj)
     {
         var other = obj as PermissionGroupKey;
 
-        return _key.Equals(other?._key);
+        return other != null && Key.Equals(other.Key);
```
hishamco, line 34: Is the equality by key only?

PR #11925, opened 2022-06-29
Add Manage Steps button and description for Deployment. Fix #11924

### LINKED ISSUE #11924: Add Manage Steps to the button to the deployemnt and add description to the feature
When viewing deployment plans on the UI, clicking the title directs the user manage steps page which is great. However, there is no other indication about managing steps. In other words, the user does not know the different between edit or manage steps.

I think we should add a button called "Manage Steps" that also direct the user to the manage deployment plans. This way the user can easily understand that there are 2 edit function here (i.e., one for Editing the the steps and the other one for editing the deployment plant itself.)

Also the Deployment feature should have some sort of description like `Provides a way to import and export packages`

CrestApps:
Fix #11924

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Deployment/Manifest.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Deployment/Manifest.cs
@@ -5,5 +5,6 @@ using OrchardCore.Modules.Manifest;
     Author = ManifestConstants.OrchardCoreTeam,
     Website = ManifestConstants.OrchardCoreWebsite,
     Version = ManifestConstants.OrchardCoreVersion,
-    Category = "Deployment"
+    Category = "Deployment",
+    Description = "Provides a way to import and export packages"
 )]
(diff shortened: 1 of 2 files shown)


src/OrchardCore.Modules/OrchardCore.Deployment/Manifest.cs
```diff
@@ -5,5 +5,6 @@
     Author = ManifestConstants.OrchardCoreTeam,
     Website = ManifestConstants.OrchardCoreWebsite,
     Version = ManifestConstants.OrchardCoreVersion,
-    Category = "Deployment"
+    Category = "Deployment",
+    Description = "Provides a way to import and export packages"
```
Skrypt, line 9: "packages" should be replaced by "recipes".

CrestApps: @Skrypt  The menu item is called "Package Import" which is why I referred to it as package. May be the menu should be also renamed. May be even combine "JSON Import" and Recipe Import into a single page with option to upload a file of just provide a string.
Any way, the request change is done.

Skrypt (approved):

CrestApps:
@Skrypt agreed. May be we can create a another issue to add buttons where needed.

hishamco (approved):

PR #11926, opened 2022-06-29
Add the list of content items being displayed on the list-page to the…

### LINKED ISSUE #11591: Add DisplayName to the user profile to allow the admin to control how user's name appear
OC display the username field everywhere though the app like in content item list and the `UserPickerField` this may be great for performance, but IMO, it isn't great from user experience. Usernames are just not human friendly and should used for identification only.

One way to do this is by creating a "Display field" in the user screen that would allow the to type is a display name instead of a username. By default, we could save the username into the display name.

A better option may be to add First, Middle and Last name field in the user, then in the settings provide an option for the user how to display the user with the following possible options
  1. Username
  2. First Middle then Last name. For example, "John M Smith"
  3. Last Name, First then Middle name. For example, "Smith, John M"

For reporting and placed where I have control over, I use a the following contract to to get the display name. The implementation reads the settings to determine the format. I also attach a part on the `User` shape that would capture First, Middle, and Last name.

```
public interface IDisplayNameProvider
{
    public Task<string> GetAsync(User user);
}
```

Meanwhile on the content list/display views, I want to make the username as a clickable link so when the user clicks it a popover opens up with the user's info. Although this isn't ideal, this is the best I can do with the current OC setup.

MikeAlhayek:
Fixed #11591 

### Summary
There is a need to change how the owner name appears on the Content List/index views. But in order to do that, we need a way to determine a list of the content owners then query the Users table to get info about each owner while avoiding select N+1 case. 
Adding the content items to the feature will provider a full list of the contents which contains the owner id.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Controllers/AdminController.cs
@@ -257,6 +257,8 @@ namespace OrchardCore.Contents.Controllers
             // Load items so that loading handlers are invoked.
             var pageOfContentItems = await query.Skip(pager.GetStartIndex()).Take(pager.PageSize).ListAsync(_contentManager);
 
+            _httpContextAccessor.HttpContext.Features.Set(new ContentItemListFeature(pageOfContentItems));
+
             // We prepare the content items SummaryAdmin shape
             var contentItemSummaries = new List<dynamic>();
             foreach (var contentItem in pageOfContentItems)
--- a/src/OrchardCore/OrchardCore.Contents.Core/Services/ContentItemListFeature.cs
+++ b/src/OrchardCore/OrchardCore.Contents.Core/Services/ContentItemListFeature.cs
@@ -1,0 +1,13 @@
+using System.Collections.Generic;
+using OrchardCore.ContentManagement;
+
+namespace OrchardCore.Contents.Services;
+public class ContentItemListFeature
+{
+    public IEnumerable<ContentItem> ContentItems { get; set; }
+
+    public ContentItemListFeature(IEnumerable<ContentItem> contentItems)
+    {
+        ContentItems = contentItems;
+    }
+}


CrestApps:
@Skrypt When the content items are in the request, then we can easily create a display driver to get a distinct list of owners and get the User object for every owner using a **single query**. Then we change how the owner name is being displayed using something like this

```
public class UserContentsDriver : ContentDisplayDriver
{
    private readonly IHttpContextAccessor _httpContextAccessor;
    private readonly IUserForContentItemsProvider _userForContentItemsProvider;

    public UserContentsDriver(
        IUserForContentItemsProvider userForContentItemsProvider)
    {
        _userForContentItemsProvider = userForContentItemsProvider;
    }

    public override Task<IDisplayResult> DisplayAsync(ContentItem contentItem, BuildDisplayContext context)
    {
        if (!context.DisplayType.Equals("SummaryAdmin", System.StringComparison.OrdinalIgnoreCase))
        {
            return null;
        }

        var contentList = _httpContextAccessor.HttpContext.Features.Get<ContentItemListFeature>();

        if (contentList == null)
        {
            return null;
        }

        return Initialize<ContentItemWithUserViewModel>("ContentsMeta_SummaryAdmin_User", async viewModel =>
        {
            viewModel.ContentItem = contentItem;
            viewModel.User = await _userForContentItemsProvider.UserAsync(contentList.ContentItems, contentItem.Owner);
        }).Location("SummaryAdmin", "Meta:25");
    }
}
```

MikeAlhayek:
@sebastienros can you please review this one?

MikeAlhayek:
@sebastienros any objection on adding this?

src/OrchardCore.Modules/OrchardCore.Contents/Controllers/AdminController.cs
```diff
@@ -257,6 +257,8 @@ public class AdminController : Controller
             // Load items so that loading handlers are invoked.
             var pageOfContentItems = await query.Skip(pager.GetStartIndex()).Take(pager.PageSize).ListAsync(_contentManager);
 
+            _httpContextAccessor.HttpContext.Features.Set(new ContentItemListFeature(pageOfContentItems));
```
Skrypt, line 260: This is not really what Features are meant for.

MikeAlhayek: Alternatively, we can invoke an event that would let's other hook into it. Do you have a better suggestion?

Skrypt:
Looks like we missed that one.

sebastienros:
1. Create a `UserName` shape in which you render the username by default. (like it's done today).
3. Use this shape everywhere in the admin.
4. You can update the shape in your app to load the user's display name (or anything else), and add caching metadata to it. Probably add a cache dependency for users when they are updated https://docs.orchardcore.net/en/latest/docs/reference/modules/DynamicCache/#well-known-cache-dependencies

PR #11928, opened 2022-06-30
Display all users in the UserPickerField by default unless the user select specific role

### LINKED ISSUE #11878: You can't pick users when you only have permissions to edit your own content.
You can't pick users when you only have permissions to edit your own content.

### To Reproduce
I have a content type  "Project" with a user content picker field. I'm only authorizated to edit my own created projects. When I create a new project I currently can't select users with the user picker field.

### Expected behavior
I should be able to select users when I have the permission to create the content item (in this example "Project").

### Proposed fix
https://github.com/OrchardCMS/OrchardCore/commit/f8fab3c2f8c5a606080b5333f70f949dcbb7a790

### Screenshots
select users bug

CrestApps:
Fix #11878

When a user attaches a `UserPickerField` to a content type, the field settings does not have default roles selected or have the `DisplayAllUsers` set to true so in this case, the search for a user return no records. The fix is to display all users in the `UserPickerField`  by default to allow all users to show up. If the user want to override that behavior, then they can specific roles to select from. Without this, no users will show up in  `UserPickerField` unless the user edit the field and set the settings.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Settings/UserPickerFieldSettingsDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Settings/UserPickerFieldSettingsDriver.cs
@@ -26,10 +26,14 @@ namespace OrchardCore.ContentFields.Settings
                 model.Hint = settings.Hint;
                 model.Required = settings.Required;
                 model.Multiple = settings.Multiple;
-
                 var roles = (await _roleService.GetRoleNamesAsync())
                     .Except(new[] { "Anonymous", "Authenticated" }, StringComparer.OrdinalIgnoreCase)
-                    .Select(x => new RoleEntry { Role = x, IsSelected = settings.DisplayedRoles.Contains(x, StringComparer.OrdinalIgnoreCase) }).ToArray();
+                    .Select(roleName => new RoleEntry
+                    {
+                        Role = roleName,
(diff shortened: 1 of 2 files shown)


Skrypt (approved):

Skrypt:
Can you confirm that you can still remove all roles from the UserPickerField settings?

src/OrchardCore.Modules/OrchardCore.ContentFields/Settings/UserPickerFieldSettingsDriver.cs
```diff
@@ -43,14 +51,17 @@ public override async Task<IDisplayResult> UpdateAsync(ContentPartFieldDefinitio
...
                 {
-                    settings.DisplayedRoles = Array.Empty<String>();
+                    // No selected role should have the same effect as display all users
+                    settings.DisplayedRoles = Array.Empty<string>();
+                    settings.DisplayAllUsers = true;
                 }
                 else
                 {
-                    settings.DisplayedRoles = model.Roles.Where(x => x.IsSelected).Select(x => x.Role).ToArray();
+                    settings.DisplayedRoles = selectedRoles;
```
Skrypt, line 64: settings.DisplayAllUsers = false;

Skrypt: Because here it is a UserPickerFieldSettings which always defaults to true from a new instance.

PR #11960, opened 2022-07-05
Fix Media Modal

### LINKED ISSUE #11897: In mediaFile in admin theme no hide modal windows
image
If you click the "edit" button, a modal window will open.
image
If you do nothing and click on any button in this window, it will close. If you rename the file and click "OK", the changes will be applied, but after that the buttons do not work, the modal window does not close. It can be closed only by pressing the "Esc" key on the keyboard. The "Cantel" and "X" buttons on the top right of the modal window also do not work after applying the changes. How do I fix this behavior?

Skrypt:
For Rename and Create folder.

Fixes #11897


hishamco (approved):
LGTM, but still I'm not sure why it's happen now?!! Is it a part of the new Bootstrap update? May  be

Skrypt:
Yes, I think it simply doesn't bind the events to the buttons because it doesn't create a proper Modal instance as it should.

Skrypt:
Yes, the code could be simplified by having a unique global instance for each of these modals.

hishamco:
So, let us merge this and open another issue for enhancements

PR #11968, opened 2022-07-06
Redirect the user to edit role when a new role is created

### LINKED ISSUE #11959: When a user create a new role, redirect the use to the edit page instead of index
When a new role is created, currently we redirect the user to the [Index](https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore.Modules/OrchardCore.Roles/Controllers/AdminController.cs#L116) view. After a user adds a new role, they would likely want to assign permissions to it. from a user experience, I think it would be more logical to redirect the user to the Edit screen of that role instead.

After a new role is successfully created, redirect the user to the edit view `return RedirectToAction(nameof(Edit), new { id = role.RoleName});`

CrestApps:
Fix #11959

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Roles/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Roles/Controllers/AdminController.cs
@@ -113,7 +113,7 @@ namespace OrchardCore.Roles.Controllers
                 if (result.Succeeded)
                 {
                     await _notifier.SuccessAsync(H["Role created successfully."]);
-                    return RedirectToAction(nameof(Index));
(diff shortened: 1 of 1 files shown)


hishamco:
It's not compulsary to assign permissions after creation directly

CrestApps:
@hishamco its not. But 9 time out of 10 you will add permission.

ns8482e:
That not good UX

IMO on save of add or edit should return to list.

 If you have save and continue button then staying on edit is fine.

I guess you want to provide `save and continue` as an additional option

hishamco:
@ns8482e reads my mind, I will push a PR to supports save and continue as we already did in many places. We could add nother option `Save and edit` because roles are little bit different because they have permissions

PR #11978, opened 2022-07-09
Add ContainedStereotypes BagPartSettings to allow a user to include content-types by steryotype

### LINKED ISSUE #11977: Add ContainedStereotypes BagPartSettings to allow a user to include content-types by steryotype
When attaching `BagPart` to a content type, the user must explicitly set [ContainedContentTypes](https://github.com/OrchardCMS/OrchardCore/blob/339d74b251b53393dd510f8ad3027aa17d78500d/src/OrchardCore.Modules/OrchardCore.Flows/Models/BagPartSettings.cs#L7) with an array of content-types to be include in the BagPart. This is good for most case, however, it would be great to allow for setting the contained content-types using Stereotype. Stereotype would be in addition to not in place of `ContainedContentTypes`.

For example, I want to group all Contact Methods (phone number, address, email address... etc) by a stereotype type called `ContactMethod`. All of these content types share similar functionally which is a way to contact a person. Now I created a `Person` content-type and attached BagPart to it. In this case, I had to explicitly specific each content types that in the `ContainedContentTypes`. But if a new content-type was added later "from other feature/module", the user would have manually to edit the BagPart settings every time/every where ContactMethods are using to add the new content-type which it isn't efficient. 

**Additional benefit here** if any module want to create a new content type that adhere to my Stereotype would be welcome to do so without having the need to modify the `BagPart` Settings. So now my BagPart is more flexible. 

Add a way to specific a stereotypes in the `BagPartSettings` in addition to `ContainedContentTypes` so the user is able to explicitly able to set contained-content-types or explicitly set the stereotype.

We can achieve this by changing the `BagPartSettings` to something like this
```
    public class BagPartSettings
    {
        public string[] ContainedContentTypes { get; set; } = Array.Empty<string>();
        public string[] ContainedStereotypes { get; set; } = Array.Empty<string>();
        public string DisplayType { get; set; }
    }
```
Then the line [BagPartDisplayDriver](https://github.com/OrchardCMS/OrchardCore/blob/339d74b251b53393dd510f8ad3027aa17d78500d/src/OrchardCore.Modules/OrchardCore.Flows/Drivers/BagPartDisplayDriver.cs#L95) will be changed to something like this

```
if(settings.ContainedStereotypes != null && settings.ContainedStereotypes.Length > 0)
{
    return _contentDefinitionManager.ListTypeDefinitions()
                   .Where(contentType => contentType.HasStereotype() && settings.ContainedStereotypes.Contains(contentType.GetStereotype(), StringComparer.OrdinalIgnoreCase));
}

 return settings.ContainedContentTypes
                .Select(contentType => _contentDefinitionManager.GetTypeDefinition(contentType))
                .Where(contentType => contentType != null);
```
Then these the first 2 extensions would be added to clean up the code whereas the rest of the helpers will be nice add to help with other usage
```

public static class ContentTypeDefinition
{
    public static bool HasStereotype(this ContentTypeDefinition contentTypeDefinition)
    {
        return !String.IsNullOrEmpty(contentTypeDefinition.GetSettings().Stereotype);
    }

    public static string GetStereotype(this ContentTypeDefinition contentTypeDefinition)
    {
        return contentTypeDefinition.GetSettings().Stereotype;
    }
}
```

CrestApps:
Fix #11977

Here is a visual of how the edit view of the bag part.

bagpart settings


CrestApps:
This PR is also related to PR#12103

src/OrchardCore.Modules/OrchardCore.Flows/Drivers/BagPartDisplayDriver.cs
```diff
@@ -213,10 +213,20 @@ private static bool IsSecurable(IContentDefinitionManager contentDefinitionManag
...
-
-            var contentTypes = settings.ContainedContentTypes
-                .Select(contentType => _contentDefinitionManager.GetTypeDefinition(contentType))
-                .Where(contentType => contentType != null);
+            var contentTypes = Enumerable.Empty<ContentTypeDefinition>();
+            
+            if (settings.ContainedStereotypes != null && settings.ContainedStereotypes.Length > 0)
+            {
+                contentTypes = _contentDefinitionManager
+                            .ListTypeDefinitions()
```
hishamco, line 221: Extra whitespaces

src/OrchardCore.Modules/OrchardCore.Flows/Drivers/BagPartDisplayDriver.cs
```diff
@@ -213,10 +213,20 @@ private static bool IsSecurable(IContentDefinitionManager contentDefinitionManag
...
+                            .Where(contentType => contentType.HasStereotype() && settings.ContainedStereotypes.Contains(contentType.GetStereotype(), StringComparer.OrdinalIgnoreCase));
+            } 
+            else if (settings.ContainedContentTypes != null && settings.ContainedContentTypes.Length > 0)
+            {
+                contentTypes = settings.ContainedContentTypes
+                               .Select(contentType => _contentDefinitionManager.GetTypeDefinition(contentType))
```
hishamco, line 227: Same here

hishamco (changes requested):

src/OrchardCore.Modules/OrchardCore.Flows/Drivers/BagPartDisplayDriver.cs
CrestApps: done

src/OrchardCore.Modules/OrchardCore.Flows/Settings/BagPartSettingsDisplayDriver.cs
```diff
@@ -44,22 +49,56 @@ public override async Task<IDisplayResult> UpdateAsync(ContentTypePartDefinition
...
+            {
+                ContainedContentTypes = Array.Empty<string>(),
+                ContainedStereotypes = model.Stereotypes.Split(',', StringSplitOptions.TrimEntries | StringSplitOptions.RemoveEmptyEntries),
+                DisplayType = model.DisplayType
+            });
+        }
+
+        private void SetContentTypes(UpdateTypePartEditorContext context, BagPartSettingsViewModel model)
+        {
+            if (model.ContainedContentTypes == null || model.ContainedContentTypes.Length == 0)
```
hishamco, line 89: I think null `coalescing` operator could be used in many places in your PR

CrestApps: I agree with you. I like using the null coalescing when it make sense. But I don’t like using it inside an of condition. Either way, it’s a preference

hishamco: If you look there are many places that you are checking both nullability and length while you can use the available C# operators for that

CrestApps: @hishamco I am aware. As mentioned previously, it's a preference. Additionally, I don't think I ever saw something like this in OC code.

```
if ((model.ContainedContentTypes?.Length) ?? 0 == 0)
{

}
```
ugggggh, that hurt my eyes.

We could do something like this instead, but that is just a wast of memory allocation

```
var totalContainedTypes = model.ContainedContentTypes?.Length ?? 0;

if (totalContainedTypes  == 0)
{
}
```

hishamco: What about

```csharp
if (model.ContainedContentTypes?.Length == 0)
{

}
```

jtkech: It doesn't have the same behavior if `ContainedContentTypes` is null because you get a nullable int without any value, so it is neither == 0, nor != 0, nor >=0 and so on.

You would need to use the following which is not simpler.

    if (!(model.ContainedContentTypes?.Length).HasValue ||
        model.ContainedContentTypes.Length == 0)
    {

    }

I think the following would work but it is not so clear.

    if (!(model.ContainedContentTypes?.Length > 0))
    {

    }

CrestApps: @hishamco as @jtkech mentioned that it's not the same. 

As you can see the the original code in the PR is not the shortest, but is the most readable.
```
if (model.ContainedContentTypes == null || model.ContainedContentTypes.Length == 0)
```

hishamco:
```csharp
if (!(model.ContainedContentTypes?.Any() ?? false))
{
}
```

This should work ;) or we can keep as you did
```

Comment (CrestApps) [reply src/OrchardCore.Modules/OrchardCore.Flows/Settings/BagPartSettingsDisplayDriver.cs:89]: @hishamco we all can agree that this can be written in many different ways. I don't want to complicate the code for no good reason (other than making it too fancy :) ). I would rather leave it as is for simple readability

Comment (hishamco) [reply src/OrchardCore.Modules/OrchardCore.Flows/Settings/BagPartSettingsDisplayDriver.cs:89]: Leave it as its, then I can refactor this later

File: src/OrchardCore.Modules/OrchardCore.Flows/Views/BagPartSettings.Edit.cshtml [hunk 4]
```diff
@@ -1,15 +1,67 @@
...
     <div class="w-sm-50">
         <label asp-for="DisplayType">@T["Display Type"]</label>
         <input asp-for="DisplayType" placeholder="Summary" class="form-control medium" />
     </div>
     <span class="hint">@T["The display type to use when rendering the content items. Default is Summary."]</span>
 </div>
+
+<script at="Foot" asp-name="bagPartSettings">
+    document.addEventListener('DOMContentLoaded', function() {
+
```
hishamco, line 45: Remove the extra line here

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Metadata/Models/ContentTypeExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+public static class ContentTypeExtensions
+{
+    public static bool HasStereotype(this ContentTypeDefinition contentTypeDefinition)
+    {
+        return !String.IsNullOrEmpty(contentTypeDefinition.GetSettings().Stereotype);
+    }
+
+    public static string GetStereotype(this ContentTypeDefinition contentTypeDefinition)
+    {
+        return contentTypeDefinition.GetSettings().Stereotype;
```
hishamco, line 15: Need to check if it has stereotype

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Metadata/Models/ContentTypeExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public static bool IsListable(this ContentTypeDefinition contentTypeDefinition)
```
hishamco, line 18: I don't think these change should belong to this PR

hishamco (changes requested):
Looks good, except few comments

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Metadata/Models/ContentTypeExtensions.cs
CrestApps: I don't this this is needed. This extension simply return `Stereotype` or `null`. If you need to check if `Stereotype` exists, you can use `HasStereotype`

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Metadata/Models/ContentTypeExtensions.cs
CrestApps: I think these extensions are related in a way. They should clean up some of the code used in may places. Additionally, some of these methods are used in this PR. I recommend we keep them and start using them in other RPs

src/OrchardCore.Modules/OrchardCore.Flows/Views/BagPartSettings.Edit.cshtml
CrestApps: Done.

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Metadata/Models/ContentTypeExtensions.cs
hishamco: If it returns stereotype or `null`, `GetStereotypeOrDefault()` could make sense. It's similar to `Single()`, `SingleOrDefault` the first throws while the last one returns `null`

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Metadata/Models/ContentTypeExtensions.cs
hishamco: I see

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Metadata/Models/ContentTypeExtensions.cs
CrestApps: Okay. I renamed it.

src/OrchardCore.Modules/OrchardCore.Flows/Views/BagPartSettings.Edit.cshtml
```diff
@@ -1,15 +1,66 @@
...
 <div class="mb-3">
     <div class="w-sm-50">
         <label asp-for="DisplayType">@T["Display Type"]</label>
         <input asp-for="DisplayType" placeholder="Summary" class="form-control medium" />
     </div>
     <span class="hint">@T["The display type to use when rendering the content items. Default is Summary."]</span>
 </div>
+
+<script at="Foot" asp-name="bagPartSettings">
+    document.addEventListener('DOMContentLoaded', function() {
```
hishamco, line 44: Just a final comment here, AFAIK we usually using `$(function() { })` in our code base, I'd like to use jQuery like we normally doing

hishamco, line 44: Just a final comment here, AFAIK we usually using `$(function() { })` in our code base, I'd like to use jQuery like we normally doing

src/OrchardCore.Modules/OrchardCore.Flows/Views/BagPartSettings.Edit.cshtml
```diff
@@ -1,15 +1,66 @@
...
+        var sources = document.getElementsByName('@Html.NameFor(x => x.Source)');
+        var contentTypesContainer = document.getElementById('ContentTypesContainer');
+        var stereotypeContainer = document.getElementById('StereotypesContainer');
+
+        for (var i = 0; i < sources.length; i++) {
+            var source = sources[i];
+            source.addEventListener('change', function(src) {
```
hishamco, line 51: I think we need to be consistent with other code base in terms of dealing with javascripr events

hishamco, line 51: I think we need to be consistent with other code base in terms of dealing with javascripr events

src/OrchardCore.Modules/OrchardCore.Flows/Views/BagPartSettings.Edit.cshtml
CrestApps: We discussed the same exact thing in different PR’s.  There is no point of including jQuery just to make the code couple of lines shorter.

sebastienros:
Let's try to not use jQuery anymore when not necessary since BS5 is not using it. And remove it piece by piece.

src/OrchardCore.Modules/OrchardCore.Flows/Views/BagPartSettings.Edit.cshtml
```diff
@@ -1,15 +1,66 @@
...
+                if (!src.target.checked) {
+                    return;
+                }
+                if (src.target.value == '@(BagPartSettingType.Stereotypes)') {
```
sebastienros, line 55: Something like this:
`var contentTypesContainer = document.getElementById('ContentTypesContainer');`

sebastienros (approved):

src/OrchardCore.Modules/OrchardCore.Flows/Views/BagPartSettings.Edit.cshtml
CrestApps: @sebastienros actually `BagPartSettingType` here is an enum so I don't think the requested change is needed since this isn't coming from a user input.

hishamco:
> Let's try to not use jQuery anymore when not necessary since BS5 is not using it. And remove it piece by piece.

Is this mean we will remove the jQuery dependency from everywhere?

Skrypt:
> Is this mean we will remove the jQuery dependency from everywhere?

Yes, at some point. But we are not there yet. We don't encourage using it in the meantime.

hishamco:
I never mind, but we could introduce a helper function to reduce of amount code needed for basic stuff

Skrypt:
That would be equivalent to using JQuery right? I have no issue with JQuery but JQueryUI is what we need to get rid of. And at the same time, I don't see any issue using plain javascript.

hishamco:
> That would be equivalent to using JQuery right?

Right

> I don't see any issue using plain javascript

Me too except a lot of code to do a small thing ;)

Issue #11920, opened 2022-06-29
Add IsSystem property to IRole interface to allow identifing the system roles easily.


MikeAlhayek:
## Is your feature request related to a problem? Please describe.
In many instances in OrchardCore's code, we repetitively use the following code
```
var allRoles = (await _roleService.GetRoleNamesAsync()).Except(new[] { "Anonymous", "Authenticated" }, StringComparer.OrdinalIgnoreCase);
```
These two roles are system roles; IMO, both should be grouped somehow.

## Describe the solution you'd like
First it may be a good idea to add the following class to reduce the use of these terms all over the place.

```
public class CommonRoleConstants
{
    public const string Administrator = "Administrator";
    public const string Editor = "Editor";
    public const string Moderator = "Moderator";
    public const string Author = "Author";
    public const string Contributor = "Contributor";
    public const string Authenticated = "Authenticated";
    public const string Anonymous = "Anonymous";
}
```
I can think of two approaches to clean up the code

### Option 1: Flagging the System Role
I think this may be a better option even if it require few changes. But the idea is to add `bool IsSystem { get; }` property to the [IRole](https://github.com/OrchardCMS/OrchardCore/blob/339d74b251b53393dd510f8ad3027aa17d78500d/src/OrchardCore/OrchardCore.Infrastructure.Abstractions/Security/IRole.cs) interface Then,
 - Either, set the `IsSystem` flag to true when adding these 2 roles. Perhaps we should also add `IsSystem` flag to [PermissionStereotype](https://github.com/OrchardCMS/OrchardCore/blob/339d74b251b53393dd510f8ad3027aa17d78500d/src/OrchardCore.Modules/OrchardCore.ContentLocalization/Permissions.cs#L53-L60) also since that is used to create the role internally via [RoleUpdater](https://github.com/OrchardCMS/OrchardCore/blob/339d74b251b53393dd510f8ad3027aa17d78500d/src/OrchardCore.Modules/OrchardCore.Roles/Services/RoleUpdater.cs) 
otherwise, we would have to add a check if the new role is system based or not 
```
role = new Role 
{ 
    RoleName = stereotype.Name, 
    RoleDescription = stereotype.Name + " role",
    IsSystem = new[] { CommonRoleConstants.Anonymous, CommonRoleConstants.Authenticated }.Contains(stereotype.Name, StringComparer.OrdinalIgnoreCase)
};
```
 - Or, define the `bool IsSystem` property in Role like this 
```
[...]

PR #11981, opened 2022-07-09
Add BuiltInRole

hishamco:
Addresses #11920

### DIFF
--- a/src/OrchardCore/OrchardCore.Infrastructure.Abstractions/Security/BuiltInRole.cs
+++ b/src/OrchardCore/OrchardCore.Infrastructure.Abstractions/Security/BuiltInRole.cs
@@ -1,0 +1,22 @@
+using System.ComponentModel.DataAnnotations;
+
+namespace OrchardCore.Security
+{
+    public static class BuiltInRole
+    {
+        [Display(Name = nameof(Administrator))]
+        public const string Administrator = "Administrator";
+
+        public const string Anonymous = "Anonymous";
+
+        public const string Authenticated = "Authenticated";
+
+        public const string Author = "Author";
+
+        public const string Contributor = "Contributor";
+
+        public const string Editor = "Editor";
+
+        public const string Moderator = "Moderator";
+    }
+}
--- a/src/OrchardCore.Modules/OrchardCore.Admin/Permissions.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Admin/Permissions.cs
@@ -1,5 +1,6 @@
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using OrchardCore.Security;
 using OrchardCore.Security.Permissions;
 
 namespace OrchardCore.Admin
@@ -19,23 +20,23 @@ namespace OrchardCore.Admin
             {
                 new PermissionStereotype
                 {
-                    Name = "Administrator",
+                    Name = BuiltInRole.Administrator,
                     Permissions = GetPermissions()
                 },
                 new PermissionStereotype {
-                    Name = "Editor",
+                    Name = BuiltInRole.Editor,
                     Permissions = GetPermissions()
                 },
                 new PermissionStereotype {
-                    Name = "Moderator",
+                    Name = BuiltInRole.Moderator,
                     Permissions = GetPermissions()
                 },
                 new PermissionStereotype {
-                    Name = "Author",
+                    Name = BuiltInRole.Author,
                     Permissions = GetPermissions()
                 },
                 new PermissionStereotype {
-                    Name = "Contributor",
+                    Name = BuiltInRole.Contributor,
                     Permissions = GetPermissions()
                 }
             };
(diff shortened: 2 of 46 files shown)


src/OrchardCore/OrchardCore.Infrastructure.Abstractions/Security/BuiltInRole.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        public const string Authenticated = "Authenticated";
+
+        public const string Author = "Author";
+
+        public const string Contributor = "Contributor";
+
+        public const string Editor = "Editor";
+
+        public const string Moderator = "Moderator";
```
Skrypt, line 20: Not sure that "Author, Contributor, Editor, Moderator" should be considered built-in roles. Sometimes people want to remove them because they have no use for them.

hishamco: Right, but they are builtin once in OC,. We could add a method say `GetBuiltInRoles()` that can be `virtual`

The PR still WIP, hope to update it frequently ..

src/OrchardCore/OrchardCore.Infrastructure.Abstractions/Security/BuiltInRole.cs
```diff
@@ -0,0 +1,22 @@
+using System.ComponentModel.DataAnnotations;
+
+namespace OrchardCore.Security
+{
+    public static class BuiltInRole
```
CrestApps, line 5: @hishamco How about changing the name to `SecurityConstants` instead? 
Also, in #11920 you said you were going to provide ways to localize these constants, I don't see how localization is will be supported here.

src/OrchardCore/OrchardCore.Infrastructure.Abstractions/Security/BuiltInRole.cs
CrestApps: @Skrypt I tried to go down that path of removing these roles at one point, but there is no way to remove unwanted roles now since all these roles are populated using `IPermissionProvider.GetDefaultStereotypes()` from the IoC container.

One possible way to be able to remove these roles would be to using a service that would provide role types. In `IRole` we can add a property `public string Type { get; set; }`

Then in `IRoleService` we can `Task<IEnumerable<IRole>> GetRolesAsync(string type);`

Then we'll change `IPermissionProvider` from something like this

```
public class UserPermissionProvider : IPermissionProvider
{
    public IEnumerable<PermissionStereotype> GetDefaultStereotypes() =>
        new[]
        {
                new PermissionStereotype
                {
                    Name = "Administrator",
                    Permissions = new[]
                    {
                        UserPermissions.ImportUsingExcel,
                    },
                },
        };

    public Task<IEnumerable<Permission>> GetPermissionsAsync() =>
        Task.FromResult(new[]
        {
                UserPermissions.ImportUsingExcel,
        }
        .AsEnumerable());
}
```

to something like this

```
public class UserPermissionProvider : IPermissionProvider
{
	public IRoleService _roleService;

	public UserPermissionProvider(IRoleService roleService)
	{
		_roleService = roleService
	}

    public IEnumerable<PermissionStereotype> GetDefaultStereotypes()
    {
    	return _roleService.GetRolesAsync("administrator")
    					   .GetAwaiter()
    					   .GetResult()
						   .Select(role => new PermissionStereotype
				           {
				               Name = x.RoleName,
				               Permissions = new[]
				               {
				                   UserPermissions.ImportUsingExcel,
				               },
				           }).ToArray();
    }

    public Task<IEnumerable<Permission>> GetPermissionsAsync() =>
        Task.FromResult(new[]
        {
                UserPermissions.ImportUsingExcel,
        }
        .AsEnumerable());
}
```

Then we would add Delete button to the `Roles` UI.

CrestApps: @Skrypt I tried to go down that path of removing these roles at one point, but there is no way to remove unwanted roles now since all these roles are populated using `IPermissionProvider.GetDefaultStereotypes()` from the IoC container.

But I think we should be able to remove default roles. I pitched one way to allow us to delete roles [here](https://github.com/OrchardCMS/OrchardCore/issues/11920#issuecomment-1179769462)

src/OrchardCore/OrchardCore.Infrastructure.Abstractions/Security/BuiltInRole.cs
hishamco: `BuiltInRole` is much cleaner and inspired by .NET enum

> Also, in https://github.com/OrchardCMS/OrchardCore/issues/11920 you said you were going to provide ways to localize these constants, I don't see how localization is will be supported here.

It's still WIP don't expect all the things in the first commits ;)

src/OrchardCore/OrchardCore.Infrastructure.Abstractions/Security/BuiltInRole.cs
hishamco: Please hold on until the PR is almost ready

PR #11983, opened 2022-07-09
Extend graphQL API

hyzx86:
Some graphql APIs in orchardcore will be extended in this pr

- [X] support contentpickerfield, which can return the first value instead of an array, which is very useful when building list queries
- [ ] graphql mutation (for now  in my EasyOC project  I created an API method ，witch acept an Dicionary Model )
- [ ] support  query  `UserPickerField`
- [ ] support paging query of default index


hyzx86:
# support paging query of default index

In JSON filter, we can use the parameters of Lucene query and specify a Lucene query, which only returns a contentitemid
This screenshot is the implementation in the current easyoc project. It uses the dynamic index generated based on FreeSQL, but now I don't think it should be used here

image

hyzx86:
@Skrypt Maybe I should create a issue first , Can you directly convert it into issue?

hishamco:
Please format your code

Skrypt:
@hyzx86 No, I can't create an issue from a PR.

PR #12002, opened 2022-07-12
Fix count type error

hyzx86:


src/OrchardCore.Modules/OrchardCore.Search.Elastic/ViewModels/AdminQueryViewModel.cs
```diff
@@ -12,7 +12,7 @@ public class AdminQueryViewModel
         public string Parameters { get; set; }
 
         [BindNever]
-        public int Count { get; set; }
+        public long Count { get; set; }
```
hishamco, line 15: Why?

Skrypt: There are not a lot of chances a Query reaches a count that is more than what an int allows but I don't mind using a long either.

PR #12005, opened 2022-07-12
Don't use HttpContext as a field

hishamco:
https://docs.microsoft.com/en-us/aspnet/core/performance/performance-best-practices?view=aspnetcore-6.0#do-not-store-ihttpcontextaccessorhttpcontext-in-a-field


hishamco:
Thanks for @jtkech to mention this in his comment [here](https://github.com/OrchardCMS/OrchardCore/pull/11569#discussion_r919064662)

jtkech (approved):

PR #12013, opened 2022-07-14
Hide TablePrefix for SQLite while editing a tenant

### LINKED ISSUE #12010: Hide TablePrefix for SQLite while editing a tenant.
### Describe the bug

Minor issue but the Tenant Create Editor doesn't allow to define a TablePrefix for Sqlite.
Only after the tenant has been created by re-editing it through the Tenant Edit Editor.

### To Reproduce
1. Create a teant and select Sqlite.
2. You can't define a Table Prefix.
3. Save and re-edit the tenant.
4. You now can define a prefix.

@hishamco I think this is an issue that you can easily fix.

hishamco:
Fixes #12010

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Edit.cshtml
@@ -75,7 +75,7 @@
                         <option value="" data-connection-string="false">@T["None"]</option>
                         @foreach (var provider in DatabaseProviders)
                         {
-                            <option value="@provider.Value" data-connection-string="@provider.HasConnectionString.ToString().ToLower()" data-table-prefix="@provider.HasTablePrefix.ToString().ToLower()" data-connection-string-sample="@T["The connection string to your database instance. e.g., {0}", provider.SampleConnectionString]">@provider.Name</option>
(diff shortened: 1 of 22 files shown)


src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Edit.cshtml
```diff
@@ -80,7 +80,7 @@
                     </select>
                 </div>
             }
-            <div class="mb-3 col-md-6" asp-validation-class-for="TablePrefix">
+            <div class="mb-3 col-md-6 tablePrefix" asp-validation-class-for="TablePrefix">
```
jtkech, line 83: Okay, just saw that we need the class so that the script can hide it.

jtkech (approved):

src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Edit.cshtml
hishamco: Exactly

hishamco: So I think it's fine to merge this

Issue #12006, opened 2022-07-12
UI Refresh & Enhancements


hishamco:
As we all know, the UI in many places need to refresh and consistent. One of the main UI issues that I notice using bootstrap colors extensively, while some colors are not make sense to the actions.

I can take the initiative an start prettify the UI in one page, then we can apply these changes after we approve it

CrestApps:
Just a thought, we should find a way to create reusable components for each of these button. This way we have one reusable component for all these actions. Component view or a shape.

hishamco:
It could be doable if we introduce `OC.UI` that contains all the predefine components that could be reused and override. Believe it or not I was thinking in the past about UI project that remove the dependency from bootstrap or any other UI framework. May be one day this could be happen even as a community project

hishamco:
Could be, but base UI components should be abstracted

hishamco:
I just created a quick prototype something similar to Azure DevOps dropdown menu. The final output make a lot of room, which makes the list item clear and clean

https://user-images.githubusercontent.com/3237266/179314511-f90a6757-2440-4787-950e-54b07bb94fc1.mp4

CrestApps:
@hishamco although this look nice, it requires 2 click to get to the common functions. IMO, I think the "View" and "Edit" should be kept as buttons. 

Either way, this should be constant across the entire admin.

hishamco:
I don't think we need the buttons and random colors anymore ;) regarding the clicks that how it works ;)

You can check Azure DevOps to see it in action, I will submit a PR soon

> Either way, this should be constant across the entire admin.

Sure

sebastienros:
Is it about creating reusable components (not a fan) or refreshing some items that could be made better (always good).

hishamco:
This about refreshing some items. But still I'm planning about creating a resuable components as a separated project, probably OSS if there's no resources interested on this

hishamco:
Closing this while no one interested :)

PR #12028, opened 2022-07-18
Move tenant buttons into dropdown menu

hishamco:
Address a part of #12006 

https://user-images.githubusercontent.com/3237266/179314511-f90a6757-2440-4787-950e-54b07bb94fc1.mp4


Skrypt:
You are missing this : `id="btn-setup-@entry.Name"`
On the setup button.

Skrypt:
You will need to add a click trigger on the dropdown menu before the click on the setup button.

Skrypt:
The functional tests are not executed under Windows that's why.

src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Index.cshtml
```diff
@@ -146,25 +146,29 @@
...
-                                    <a asp-action="Enable" asp-route-id="@entry.Name" class="btn btn-success btn-sm enable" data-url-af="UnsafeUrl">@T["Enable"]</a>
-                                }
-                                @if (entry.ShellSettings.State == TenantState.Uninitialized)
-                                {
-                                    <a class="btn btn-info btn-sm setup" id="btn-setup-@entry.Name" href="@GetEncodedUrl(entry, originalPathBase)">@T["Setup"]</a>
-                                }
-                                <a asp-action="Reload" asp-route-id="@entry.Name" class="btn btn-secondary btn-sm reload" data-url-af="UnsafeUrl">@T["Reload"]</a>
+                            <div class="dropdown float-end">
+                                <a id="tenantDropdownMenuButton" class="btn text-secondary" data-bs-toggle="dropdown" aria-expanded="false"><i class="fa fa-ellipsis-v"></i></a>
+                                <ul class="dropdown-menu" aria-labelledby="tenantDropdownMenuButton">
```
ns8482e, line 151: aria-labelledby Ref doesn’t have any text

hishamco:
The states are clearly shown as a badge. The current UI change:

- Makes a lot of room in each row
- Makes the UI clean & clear
- The actions should react based on some state
- No more colored buttons which are sometime confuses

FYI such menu actions already shown in GitHub & AzureDevOps for better UX. May be the issue we are looking from developer point of you ;)

Skrypt:
Ok, we need to make sure we are consistent here. While I don't dislike entirely the UI change; here we need to take into consideration that the Content Items list uses an "Action" button for these. This is why I recommended using the same UI concept.
I'm sure DevOps uses also single item buttons in some places.

hishamco:
Actions button

ns8482e:
> Actions button

Personally I don’t like devops UI,  I like GitHub’s UI better

hishamco:
GitHub

It quite similar except the action button uses dropdown menu similar to what we have in filters

Skrypt:
The inline action buttons are simply icons here on Github.

MikeAlhayek:
@hishamco thanks for putting together this PR. Can you please fix the conflict? 

I personally do not like hiding all the buttons for all actions. Very common functions like "Edit" should be visible all the time. I don't want to have to make 2 click to perform a common functions. The idea here is improve the UI not complicating it like devops :). I am also not a big fan of DevOps UI as @ns8482e mentioned. 


I think in this UI, we should show "Edit" and "View" buttons, then move the "Reload", "Enable", "Disable", "Features", and "Remove" button should be in the "Actions" button.

hishamco:
> @hishamco thanks for putting together this PR. Can you please fix the conflict?

Sure

>  personally do not like hiding all the buttons for all actions. Very common functions like "Edit" should be visible all the time

May be but the issue is everyone of us has it's own common functionality ;)

hishamco:
Let me check this with a UI expert and see what I can come up with

Piedone:
Putting the buttons under a nondescript three-dot menu is a hard no from me.

MikeAlhayek:
... is a hard no for me to. I am okay with "Actions" drop down button if that is an option

hishamco:
> Putting the buttons under a nondescript three-dot menu is a hard no from me.

It's much better for UI clarity

Piedone:
How does hiding buttons clarify intent?

hishamco:
From UI/UX perspective when there are many buttons like what we have, it would be nice to introduce an action menu instead

Piedone:
This is how the current UI looks like if all the possible buttons are displayed, in a common full HD resolution:

image

Ultimately, design is subjective until we do large sample size research for this with focus groups (which would probably be excessive for this) but this doesn't seem to be too many buttons for me. Nor there is any lack of screen real estate that we'd use better. Same if you have 10 tenants on a page.

Did you encounter somebody complaining about this UI?

hishamco:
> Did you encounter somebody complaining about this UI?

While I saw many web framework applications, & CMSs I never see something like this :) Also the coloring is another story lol

Piedone:
Do you want to bring this up at the meeting or something so we can get a closure? Here we have 2-3 against votes, and other remarks that necessitate changes even if the basic direction is agreed on.

hishamco:
Probably yes, I have this and Culture Picker in localization part #13354

MikeAlhayek:
@hishamco @Piedone I think we should have 3 buttons are we do in contents. The buttons should be View, Edit and Actions (drop down menu). But I do not like the ... button.

This way we can simplify all the buttons we have today and keep them consistent with other pages like controller.

Piedone:
I'm OK with that.

hishamco:
> But I do not like the ... button

Sure everyone will like and dislike some of the UI, but I was using things from UI/UX perspective

hishamco:
I will close this while there's not much interest in UI improvements, so I will move this to Orchard Core Contrib (OCC), while I'm already some improvements to the admin UI

PR #12030, opened 2022-07-18
Add DisplayAsync() overloads to render when a given feature is enabled

hishamco:
Related to https://github.com/OrchardCMS/OrchardCore/discussions/12014

### DIFF
--- a/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Html;
 using Microsoft.AspNetCore.Mvc.Localization;
@@ -8,6 +9,7 @@ using Microsoft.Extensions.DependencyInjection;
 using OrchardCore.DisplayManagement.Html;
 using OrchardCore.DisplayManagement.Title;
 using OrchardCore.DisplayManagement.Zones;
+using OrchardCore.Environment.Shell;
 using OrchardCore.Settings;
 
 namespace OrchardCore.DisplayManagement.Razor
@@ -22,6 +24,7 @@ namespace OrchardCore.DisplayManagement.Razor
         private IDisplayHelper _displayHelper;
         private IShapeFactory _shapeFactory;
         private IOrchardDisplayHelper _orchardHelper;
+        private IShellFeaturesManager _shellFeaturesManager;
         private ISite _site;
 
         public override ViewContext ViewContext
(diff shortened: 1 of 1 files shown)


Skrypt:
The only concern I have is that it makes that DisplayAsync call a little more expensive. Should we be concerned?

src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
```diff
@@ -404,6 +446,16 @@ public ISite Site
...
             }
         }
+
+        private async Task<bool> IsFeatureEnabledAsync(string featureId)
+        {
+            EnsureShellFeaturesManager();
+
+            var feature = (await _shellFeaturesManager.GetEnabledFeaturesAsync())
+                .FirstOrDefault(f => f.Id == featureId);
+
```
Skrypt, line 456: You need to also look for `GetAlwaysEnabledFeaturesAsync()` here.

ns8482e:
@hishamco Technically this is not useful as all views are tied to module and not a feature

jtkech:
Not sure it's a good thing, normally to validate/register things relying on a feature we have the manifest dependencies and the usage of the `[Features()]` attrubute. Or we have the `[RequireFeatures()]` to automatically validate things if a given feature(s) is enabled, e.g to execute a startup. Or a mix of both.

hishamco:
I knew, @jtkech please refer to my PR #12017 to see the use case on action. Please let me know if there's a better way to display the admin culture

src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
hishamco: I presume it should get all enabled features, I saw similar call in features admin controll

jtkech:
I don't think we have an extension point allowing features to collaborate to this admin zone/section, as we do to shape compose a content item with drivers, but also as we do for admin menus, dashboard widgets and so on.

Hmm, maybe just do an extension helper to know if a given feature is enabled and use it in the admin view, if so I would made an extension on ShellDescriptor that you can inject, so that you just have to lookup in ShellDescriptor.Features so without calling _extensionManager.GetFeatures() as it is done by ShellFeaturesManager, this because you only need the feature names, not full IFeatureInfo.

Maybe just a local extension, or no extension at all, just use in a view

    @inject ShellDescriptor ShellDescriptor
    ...
    var isEnabled = shellDescriptor.Features.Any(sf => sf.Id == "OC.Localization");

And yes, as @ns8482e said a shape is not tied to an individual feature but to a module, that may have only one feature, itself ;) But yes, a shape may be only used by the components of a given feature.

hishamco:
```
@inject ShellDescriptor ShellDescriptor
...
var isEnabled = shellDescriptor.Features.Any(sf => sf.Id == "OC.Localization");
```

if we can utilize this as extension method to render a shape optionally it will be good

src/OrchardCore/OrchardCore.DisplayManagement/Razor/RazorPage.cs
```diff
@@ -51,6 +54,14 @@ private void EnsureShapeFactory()
             }
         }
 
+        private void EnsureShellFeaturesManager()
```
CrestApps, line 57: I think this method would be better if it returned the cached instance if `IShellFeaturesManager` and maybe name name `GetFeatureManager()`

hishamco: You can suggest this into a new issue. I'm not sure if we still need this PR ;)

PR #12031, opened 2022-07-18
Use file scoped namespace

hishamco:
Fixes #12029

PR #12032, opened 2022-07-18
Code cleanup

Skrypt:
WIP


Skrypt:
That makes a big Pull Request. I feel concerned about code changes history since it changes almost everything in these files.

jtkech:
Hmm, personally I would keep it as is and let people using it progressively in new PRs. And maybe when a PR update an old file, but not sure as it would pollute the git diff.

hishamco:
Is this including the namespace file scope? coz I already done!!

jtkech:
Oops my comment was more related to #12031 in the first place.

Skrypt:
I don't know, to be honest, if we want to change it then it is necessary. At the same time, it will eventually break everyone's PR. Since @hishamco did it I will close this one.

Issue #11703, opened 2022-05-14
Radio button groups are not working after migrating TheAdmin theme from BS4 to BS5


MikeAlhayek:
### Describe the bug
When using [Radio button group](https://getbootstrap.com/docs/5.1/components/button-group/#checkbox-and-radio-button-groups) in the Admin theme, selecting a radio button does not select.

### To Reproduce
Create a view anywhere in admin using TheAdmin theme and place the following HTML code

```
<form>
            <div class="btn-toolbar" role="toolbar" aria-label="Toolbar with button groups">
               <div class="btn-group btn-group-sm me-2" role="group" aria-label="Sort">
                  <input type="radio" class="btn-check" value="Profit" id="Sort_ByProfit" autocomplete="off" checked="checked" data-val="true" data-val-required="The By field is required." name="Sort.By">
                  <label class="btn btn-outline-secondary" for="Sort_ByProfit">Profit</label>
                  <input type="radio" class="btn-check" value="Year" id="Sort_ByYear" autocomplete="off" name="Sort.By">
                  <label class="btn btn-outline-secondary" for="Sort_ByYear">Year</label>
                  <input type="radio" class="btn-check" value="Cost" id="Sort_ByCost" autocomplete="off" name="Sort.By">
                  <label class="btn btn-outline-secondary" for="Sort_ByCost">Cost</label>
                  <input type="radio" class="btn-check" value="Odometer" id="Sort_ByOdometer" autocomplete="off" name="Sort.By">
                  <label class="btn btn-outline-secondary" for="Sort_ByOdometer">Odometer</label>
                  <input type="radio" class="btn-check" value="Status" id="Sort_ByStatus" autocomplete="off" name="Sort.By">
                  <label class="btn btn-outline-secondary" for="Sort_ByStatus">Status</label>
                  <input type="radio" class="btn-check" value="DaysInStock" id="Sort_ByDaysInStock" autocomplete="off" name="Sort.By">
                  <label class="btn btn-outline-secondary" for="Sort_ByDaysInStock">Days In Stock</label>
               </div>
               
            </div>
</form>
```

Now run the app and click on any of the radio buttons, you'll notice that the button do not appear selected. This works perfectly fine in other themes like TheTheme theme from the front end. 

### Expected behavior
When you click on the radio button, it should appear darker "selected"

### Screenshots
Here is a screen cast of TheAdmin theme where it does not stay selected as well as TheTheme theme in the front end with the expected behavior 

[...]

PR #12046, opened 2022-07-20
Bootstrap 5.2

agriffard:
https://www.npmjs.com/package/bootstrap/v/5.2.3
https://www.npmjs.com/package/bootstrap-scss/v/5.2.3
https://github.com/twbs/bootstrap/releases/tag/v5.2.3

https://getbootstrap.com/docs/5.2/migration/

- [ ]  Check modals with cards.
- [ ]  Check package.json dependencies in OrchardCore.Workflows.
- [ ]  Replace div with class `dropdown-menu` by ul li everywhere needed.
- [ ]  Replace split buttons with class `dropdown-menu` by ul li where needed.


Skrypt:
Updating the Workflow module Bootstrap components will break the popovers and modals. The module probably needs a refresh because these javascript components have been changed in BS 5.

agriffard:
The searchable dropdown lists used int the filters of the content items admin page should also be adapted to this 5.2.2 changes.

They rely on [bootstrap-select](https://www.npmjs.com/package/bootstrap-select) and I'm afraid we will have to wait a long time before this BS version is supported.

MikeAlhayek:
@agriffard @Skrypt

Just a thought on this upgrade. I think we can take advantage of the new `text-bg-dark` and `text-bg-dark` to make it easy to change between dark and light theme. Currently we use `bg-primary` class for card to change the background. I think we should `text-bg-light` and `text-bg-dark` instead.

MikeAlhayek:
@Skrypt will this PR fix #11703 ?

PR #12050, opened 2022-07-20
Prevent Trumbowyg from converting div to p element

### LINKED ISSUE #12049: WYSIWYG/Trumbowyg editors are changing div element to p
### Describe the bug
When using HtmlBodyPart or HTML Field with either WYSIWYG or Trumbowyg, the editor convert `div` elements to `p`.

Also, when the user switch into code insert mode, the textarea box does not take up the full size of the editor. 
image

### To Reproduce
1. Add `HtmlBodyPart` or `HtmlField` to a content type, use either `WYSIWYG` or `Trumbowyg` editors
2. Create new or edit existing content item. Click on `<>` of the editor to switch into code insert mode and insert the following

Here is the issue in action
wysiwyg

CrestApps:
Fix #12049

From the semantic section in [Trumbowyg docx](https://alex-d.github.io/Trumbowyg/documentation/#semantic)

Since 2.12.0 you can also put an object to customize the semantic tag mapping for each one of these tags: `<b>, <i>, <s>, <strike>, <div>`.

The `semantic` option is set to `true` by default which is equivalent to:
```
$('.trumbowyg').trumbowyg({
    semantic: {
        'b': 'strong',
        'i': 'em',
        's': 'del',
        'strike': 'del',
        'div': 'p'
    }
});
```

To prevent the `div` from being converted into `p` we should set the semantic settings like this

```
$('.trumbowyg').trumbowyg({
    semantic: {
        'b': 'strong',
        'i': 'em',
        's': 'del',
        'strike': 'del'
    }
});
```

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Views/HtmlField-Wysiwyg.Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Views/HtmlField-Wysiwyg.Edit.cshtml
@@ -18,12 +18,12 @@
     @if (culture.IsRightToLeft())
     {
         <div style="text-align:right">
-            <textarea asp-for="Html" class="form-control"></textarea>
+            <textarea asp-for="Html"></textarea>
         </div>
     }
     else
     {
-        <textarea asp-for="Html" class="form-control"></textarea>
+        <textarea asp-for="Html"></textarea>
     }
 
     @if (!String.IsNullOrEmpty(settings.Hint))
@@ -57,7 +57,13 @@
         }
 
         var settings = { 
-            lang: '@culture.TwoLetterISOLanguageName'
+            lang: '@culture.TwoLetterISOLanguageName',
+            semantic: {
+                    'b': 'strong',
+                    'i': 'em',
+                    's': 'del',
+                    'strike': 'del'
+                }
         }
 
         $('#@Html.IdFor(m => m.Html)').trumbowyg(settings).on('tbwchange', function () {
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Views/HtmlField-Wysiwyg.Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Views/HtmlField-Wysiwyg.Edit.cshtml
@@ -33,40 +33,35 @@
 </div>
 
 <script at="Foot">
-    $(function () {
-        @* set dir while keeping trumbowg translations dictionary intact *@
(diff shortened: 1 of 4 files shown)


Skrypt:
I don't agree to add these semantic options if they are already applied.
The idea is that the "Wysiwyg" editor option should not alter the "Tryumbowyg" editor default behavior.
This is why we have the "Trumbowyg" editor option that allows changing its behavior for your own needs.

sebastienros:
`semantic: false` on Wysiwyg templates
Don't change anything on Trumbowyg templates

CrestApps:
Updated

Skrypt (approved):

src/OrchardCore.Modules/OrchardCore.Html/Views/HtmlBodyPart-Wysiwyg.Edit.cshtml
```diff
@@ -16,46 +16,47 @@
...
         if (defaultButtons[6] !== "insertShortcode") {
             defaultButtons.splice(6, 0, "insertShortcode");
         }
-                
-        var settings = { 
-            lang: '@culture.TwoLetterISOLanguageName'
+
+        var settings = {
+            lang: '@culture.TwoLetterISOLanguageName',
+            semantic: false
```
hishamco, line 56: This is the only change needed, right?

src/OrchardCore.Modules/OrchardCore.ContentFields/Views/HtmlField-Wysiwyg.Edit.cshtml
```diff
@@ -18,12 +18,12 @@
     @if (culture.IsRightToLeft())
     {
         <div style="text-align:right">
-            <textarea asp-for="Html" class="form-control"></textarea>
```
hishamco, line 21: Why the bootstrap class removed, is it because Trumbowyg will override the textarea?

CrestApps: Yes

src/OrchardCore.Modules/OrchardCore.Html/Views/HtmlBodyPart-Wysiwyg.Edit.cshtml
CrestApps: Yes. The rest was auto formatted by VS

hishamco (approved):

PR #12065, opened 2022-07-22
Fix select arrow color in dark theme

hishamco:


hishamco:
I presume the same issue should happen in modal, which I will try to fix it next

hishamco:
Can we merge this before Bootstrap update?

Skrypt:
Can we see a screenshot?

hishamco:
Before

Before

After

After

Skrypt (approved):

PR #12067, opened 2022-07-23
GetCurrentThemeNameAsync() -> GetSiteThemeNameAsync()

hishamco:
This change makes `ISiteThemeService` similar to `IAdminThemeService` in terms of APIs

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Themes/Deployment/ThemesDeploymentSource.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Themes/Deployment/ThemesDeploymentSource.cs
@@ -29,7 +29,7 @@ namespace OrchardCore.Themes.Deployment
 
             result.Steps.Add(new JObject(
                 new JProperty("name", "Themes"),
-                new JProperty(nameof(ThemeStepModel.Site), await _siteThemeService.GetCurrentThemeNameAsync()),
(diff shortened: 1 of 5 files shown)


Skrypt:
The difference here is that `SiteTheme` is a tenant-scoped service. Maybe it would be better that the name reflects it.
Just an idea. Or add a comment somewhere about it.

hishamco:
> The difference here is that SiteTheme is a tenant-scoped service.

Also the AdminTheme is tenant-scoped, while we can set admin theme per tenant, Am I right?

The context here is to make both classes has similar APIs while the difference is one sets the site theme and the other sets the admin theme

Skrypt:
You are right, my brain collapsed. 😄

hishamco:
No problem it happen ;)

Skrypt:
I approve once we agree on the new name.

hishamco:
@sebastienros are you agree with the name? We should be consistent in terms of naming APIs

MikeAlhayek (approved):

hishamco:
Finally merged :)

MikeAlhayek:
you're welcome lol. It pays to join the meeting

hishamco:
Hope to join you on the upcoming meetings ..

PR #12069, opened 2022-07-24
Add missing alert in recipes page

hishamco:


src/OrchardCore.Modules/OrchardCore.Recipes/Views/Admin/Index.cshtml
```diff
@@ -3,6 +3,8 @@
 
 <zone Name="Title"><h1>@RenderTitleSegments(T["Recipes"])</h1></zone>
 
+<p class="alert alert-warning">@T["The current tenant will be reloaded when the recipe is executed."]</p>
```
ns8482e, line 6: Does it reload always?

hishamco: https://github.com/OrchardCMS/OrchardCore/blob/c255456bf1d9093d13447e17485a23a1253f3e9b/src/OrchardCore.Modules/OrchardCore.Recipes/Controllers/AdminController.cs#L122

According to the above line it reloaded after the recipe execution

ns8482e (approved):

PR #12102, opened 2022-07-26
Allow invariant type/part definition lookup in ContentDefinitionManager

### LINKED ISSUE #12101: Allow case insensitive definition lookup in ContentDefinitionManager
When using the `AdminMenu` module, the content-type is provided in the URL. For example, `/Admin/Contents/ContentItems/Event` where `Event` is the content type. Changing that to `event` will return "The page could not be found". Also, when creating a new content item using the `Contents` module, the URL will be wrong in the UI if the case changed.

Similar issue when editing content types or content parts in the UI.

The `ContentDefinitionManager` should do invariant definition lookup  to prevent this issue.

CrestApps:
Fix #12101

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentTypes/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentTypes/Controllers/AdminController.cs
@@ -249,7 +249,7 @@ namespace OrchardCore.ContentTypes.Controllers
             {
                 Type = typeViewModel,
                 PartSelections = _contentDefinitionService.GetParts(metadataPartsOnly: false)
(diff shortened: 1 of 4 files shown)


ns8482e:
Not sure if this request might add breaking changes on instantiating type systems are case sensitive - in C# event and Event can be represented  differently and technical names are tied to C# types

CrestApps:
> Not sure if this request might add breaking changes on instantiating type systems are case sensitive - in C# event and Event can be represented differently and technical names are tied to C# types

[The line](https://github.com/OrchardCMS/OrchardCore/blob/c02fedd8c3bf1bb1b892e3c5fc0c000d27a9ca80/src/OrchardCore.Modules/OrchardCore.ContentTypes/Controllers/AdminController.cs#L129) prevent the user from creating the name using UI. The good news is that this isn’t going to be a breaking change. It’s more of making all the rules the same :)

PR #12108, opened 2022-07-28
Reduce memory when reloading settings of a given tenant from database.

### LINKED ISSUE #11965: ShellSettings/DatabaseShellsSettingsSources memory leak
### Describe the bug
There appears to be a memory leak caused by the usage of IConfiguration as the backing for ShellSettings.  What I've found is that memory used by each ShellSettings instance is not being freed for garbage collection due to some fairly crazy gcroots even if you intentionally release the ShellSetting.  

In our case drilling further into the problem shows most of the memory is being used by Strings related to the DatabaseShellsSettingsSources with the VersionId field being the largest consumer of memory.   The problem is that the DatabaseShellsSettingSources contains data for ALL tenants (which in our case is 2000+) even if you're only loading/reloading a single tenant.  

I've considered either two possible solutions but the entire configuration subsystem to safely do either.

1. Make the source use a singleton internally that was shared for each call and updated as appropriate.  This is preferred since it keeps memory cost low even with many tenants but is a pretty big change in behavior since currently once you get a ShellSetting it's not live updated.  I'm also not 100% sure of how to pull this one off since I haven't studied how the builder works internally (for instance maybe AddInMemoryCollection could be provided with the same list of KeyValuePairs across different calls?

1. Figure out how to properly dispose of IConfigurationRoots when they are no longer needed.  It looks like work was done in .NET Core 3 to support this but it also appears to depend on disposing of the HostingEnvironment but I'm assuming the base work could potentially be used here? maybe?

jtkech:
Fixes #11965 

Among our tenant config/settings sources, one of them is by default the `tenants.json` file. In fact **when we reload a given tenant** we don't need all data from this source but only those related to this tenant.

When we add this source to the config stack of a given tenant, if we still use `builder.AddJsonFile()` we have no choice, we can't split this file in smaller parts per tenant.

But when this source comes from the database, we do a `builder.AddJsonStream()` on a `MemoryStream` that holds all data of this source, so here **the idea is to only hold in memory the data related to this tenant**.

To do so I added a method to the interface of this source allowing to pass a tenant name.

So that when the source comes from the database, each time we reload a specified tenant we still create a `MemoryStream` but that only holds the data related to this tenant (not all tenants).

When calling this new method on the other implementations of this source, they just ignore the provided tenant name by calling the existing method without this parameter.

---

Note: Maybe something to do in `BlobShellsSettingsSources` when we do `builder.AddJsonStream()`.

---

@ShaneCourtrille Can you give it a try if you have time?

### DIFF
--- a/src/OrchardCore/OrchardCore.Abstractions/Shell/Configuration/IShellsSettingsSources.cs
+++ b/src/OrchardCore/OrchardCore.Abstractions/Shell/Configuration/IShellsSettingsSources.cs
@@ -7,6 +7,7 @@ namespace OrchardCore.Environment.Shell.Configuration
     public interface IShellsSettingsSources
     {
         Task AddSourcesAsync(IConfigurationBuilder builder);
+        Task AddSourcesAsync(string tenant, IConfigurationBuilder builder);
         Task SaveAsync(string tenant, IDictionary<string, string> data);
     }
 
@@ -15,6 +16,12 @@ namespace OrchardCore.Environment.Shell.Configuration
         public static async Task<IConfigurationBuilder> AddSourcesAsync(this IConfigurationBuilder builder, IShellsSettingsSources sources)
         {
             await sources.AddSourcesAsync(builder);
+            return builder;
+        }
+
+        public static async Task<IConfigurationBuilder> AddSourcesAsync(this IConfigurationBuilder builder, string tenant, IShellsSettingsSources sources)
+        {
+            await sources.AddSourcesAsync(tenant, builder);
             return builder;
         }
     }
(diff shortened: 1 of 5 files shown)


ShaneCourtrille:
@sebastienros I'm taking a look at it now

ShaneCourtrille:
Thanks @jtkech!  This had a significant improvement on memory usage when provisioning a new client.  In my tests we went from 37 MB per new tenant down to 11 MB so thanks for that!

jtkech:
@ShaneCourtrille Okay cool, good to know, thanks

PR #11890, opened 2022-06-21
Tenant removal (Lombiq Technologies: OCORE-96)

jtkech:
Fixes #1212, related to #8471
Fixes #11989.

- [X] Remove button in Tenants admin.
- [X] Removing event handlers (tenant IModularTenantEvents and new host handlers).
- [X] Database handler for sqlite => delete the .db file (we clear the pool to unlock the file).
- [X] Database => Drop tables by prefix (from a central place by replaying tenant migrations).
- [X] Uninstall => Drop index table in modules (indirectly done by the above).
- [X] ShellsSettingsSources and ShellConfigurationSources.
- [X] DatabaseShellsSettingsSources and DatabaseShellConfigurationSources.
- [X] BlobShellsSettingsSources and BlobShellConfigurationSources.
- [X] Sitemaps-> Delete cache files under wwwroot (now in separated folder per tenant).
- [X] Media -> delete tenant folder and cache files under wwwroot (idem as above).
- [x] Azure Media storage => Remove Container config option, remove cache files.
- [x] Amazone Media storage => Remove Bucket config option, remove cache files.
- [X] Lucene => delete folder (the folder is unlocked when the tenant is released).
- [X] Keep in sync removed tenants in a distributed environment.
- [X] Allow to delete an Uninitialized tenant.
- [X] Add Disable, Enable and Remove Web API.
- [X] Allow to delete a tenant that failed on setup/migration.

PR #12114, opened 2022-07-29
Provide case-insensitive tenant names in ShellHost

### LINKED ISSUE #12106: Allow case-insensitive tenant names in ShellHost
In a mobile version, I created a page when a user can login to a specific tenant by providing the `tenant name`, `username`, `password`.  To identify which tenant the user is trying to access I user `_shellHost.TryGetSettings(tenantName, out settings)` which looks up the tenant by name. The issue here is if the user does not provide a case sensitive tenant name, the `ShellHost` does not find anything. 

It would be helpful if user case insensitive to find tenants. We can easily achieve this by storing the tenant key as lowercase and when we get/remove the tenant we will also use lowercase so no matter what the value is, we treat it the same. 

@jtkech is there an issue from making this change? I can push a PR if no issue with doing this.

For example the line [if (!_shellContexts.TryAdd(settings.Name, new ShellContext.PlaceHolder { Settings = settings }))] (https://github.com/OrchardCMS/OrchardCore/blob/51f0db7ac0d621fe3ecc415e707df58f81883c00/src/OrchardCore/OrchardCore/Shell/ShellHost.cs#L197) will become something like this instead `if (!_shellContexts.TryAdd(settings.Name.ToLower(), new ShellContext.PlaceHolder { Settings = settings }))`

Obviosly, we have to make the same change on `TryGetValue` and `TryRemove`
I did the following

```
private ShellSettings GetSettings(string tenantName)
{
    if (_shellHost.TryGetSettings(tenantName, out ShellSettings settings))
    {
        return settings;
    }

    var shells = _shellHost.ListShellContexts();
    var tenant = shells.FirstOrDefault(x => x.Settings != null && x.Settings.Name.Equals(tenantName, StringComparison.OrdinalIgnoreCase));

    if (tenant != null)
    {
        return tenant.Settings;
    }

    return null;
}
```

CrestApps:
Fix #12106

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/ScopeController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/ScopeController.cs
@@ -142,7 +142,7 @@ namespace OrchardCore.OpenId.Controllers
 
             descriptor.Resources.UnionWith(model.Tenants
                 .Where(tenant => tenant.Selected)
-                .Where(tenant => !String.Equals(tenant.Name, _shellSettings.Name))
+                .Where(tenant => !String.Equals(tenant.Name, _shellSettings.Name, StringComparison.OrdinalIgnoreCase))
                 .Select(tenant => OpenIdConstants.Prefixes.Tenant + tenant.Name));
 
             await _scopeManager.CreateAsync(descriptor);
--- a/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/ScopeController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/ScopeController.cs
@@ -187,7 +187,7 @@ namespace OrchardCore.OpenId.Controllers
             {
                 model.Tenants.Add(new EditOpenIdScopeViewModel.TenantEntry
                 {
-                    Current = String.Equals(tenant.Name, _shellSettings.Name, StringComparison.OrdinalIgnoreCase),
+                    Current = String.Equals(tenant.Name, _shellSettings.Name),
                     Name = tenant.Name,
                     Selected = resources.Contains(OpenIdConstants.Prefixes.Tenant + tenant.Name)
                 });
@@ -244,7 +244,7 @@ namespace OrchardCore.OpenId.Controllers
 
             descriptor.Resources.UnionWith(model.Tenants
                 .Where(tenant => tenant.Selected)
-                .Where(tenant => !String.Equals(tenant.Name, _shellSettings.Name, StringComparison.OrdinalIgnoreCase))
+                .Where(tenant => !String.Equals(tenant.Name, _shellSettings.Name))
                 .Select(tenant => OpenIdConstants.Prefixes.Tenant + tenant.Name));
 
             await _scopeManager.UpdateAsync(scope, descriptor);
--- a/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/ScopeController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/ScopeController.cs
@@ -96,11 +96,11 @@ namespace OrchardCore.OpenId.Controllers
 
             var model = new CreateOpenIdScopeViewModel();
 
-            foreach (var tenant in _shellHost.GetSettingsByState(TenantState.Running))
+            foreach (var tenant in _shellHost.GetAllSettings().Where(s => s.State == TenantState.Running))
             {
                 model.Tenants.Add(new CreateOpenIdScopeViewModel.TenantEntry
                 {
-                    Current = String.Equals(tenant.Name, _shellSettings.Name),
+                    Current = string.Equals(tenant.Name, _shellSettings.Name),
                     Name = tenant.Name
                 });
             }
@@ -135,19 +135,19 @@ namespace OrchardCore.OpenId.Controllers
                 Name = model.Name
             };
 
-            if (!String.IsNullOrEmpty(model.Resources))
+            if (!string.IsNullOrEmpty(model.Resources))
             {
                 descriptor.Resources.UnionWith(model.Resources.Split(' ', StringSplitOptions.RemoveEmptyEntries));
             }
 
             descriptor.Resources.UnionWith(model.Tenants
                 .Where(tenant => tenant.Selected)
-                .Where(tenant => !String.Equals(tenant.Name, _shellSettings.Name))
+                .Where(tenant => !string.Equals(tenant.Name, _shellSettings.Name))
                 .Select(tenant => OpenIdConstants.Prefixes.Tenant + tenant.Name));
 
             await _scopeManager.CreateAsync(descriptor);
 
-            if (String.IsNullOrEmpty(returnUrl))
+            if (string.IsNullOrEmpty(returnUrl))
             {
                 return RedirectToAction("Index");
             }
@@ -178,16 +178,16 @@ namespace OrchardCore.OpenId.Controllers
 
             var resources = await _scopeManager.GetResourcesAsync(scope);
 
-            model.Resources = String.Join(" ",
+            model.Resources = string.Join(" ",
                 from resource in resources
-                where !String.IsNullOrEmpty(resource) && !resource.StartsWith(OpenIdConstants.Prefixes.Tenant, StringComparison.Ordinal)
+                where !string.IsNullOrEmpty(resource) && !resource.StartsWith(OpenIdConstants.Prefixes.Tenant, StringComparison.Ordinal)
                 select resource);
 
-            foreach (var tenant in _shellHost.GetSettingsByState(TenantState.Running))
+            foreach (var tenant in _shellHost.GetAllSettings().Where(s => s.State == TenantState.Running))
             {
                 model.Tenants.Add(new EditOpenIdScopeViewModel.TenantEntry
                 {
-                    Current = String.Equals(tenant.Name, _shellSettings.Name),
+                    Current = string.Equals(tenant.Name, _shellSettings.Name),
                     Name = tenant.Name,
                     Selected = resources.Contains(OpenIdConstants.Prefixes.Tenant + tenant.Name)
                 });
@@ -214,7 +214,7 @@ namespace OrchardCore.OpenId.Controllers
             if (ModelState.IsValid)
             {
                 var other = await _scopeManager.FindByNameAsync(model.Name);
-                if (other != null && !String.Equals(
+                if (other != null && !string.Equals(
                     await _scopeManager.GetIdAsync(other),
                     await _scopeManager.GetIdAsync(scope)))
                 {
@@ -237,19 +237,19 @@ namespace OrchardCore.OpenId.Controllers
 
             descriptor.Resources.Clear();
 
-            if (!String.IsNullOrEmpty(model.Resources))
+            if (!string.IsNullOrEmpty(model.Resources))
             {
                 descriptor.Resources.UnionWith(model.Resources.Split(' ', StringSplitOptions.RemoveEmptyEntries));
             }
 
             descriptor.Resources.UnionWith(model.Tenants
                 .Where(tenant => tenant.Selected)
-                .Where(tenant => !String.Equals(tenant.Name, _shellSettings.Name))
+                .Where(tenant => !string.Equals(tenant.Name, _shellSettings.Name))
                 .Select(tenant => OpenIdConstants.Prefixes.Tenant + tenant.Name));
 
             await _scopeManager.UpdateAsync(scope, descriptor);
 
-            if (String.IsNullOrEmpty(returnUrl))
+            if (string.IsNullOrEmpty(returnUrl))
             {
                 return RedirectToAction("Index");
             }
--- a/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/ScopeController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/ScopeController.cs
@@ -96,11 +96,11 @@ namespace OrchardCore.OpenId.Controllers
 
             var model = new CreateOpenIdScopeViewModel();
 
-            foreach (var tenant in _shellHost.GetAllSettings().Where(s => s.State == TenantState.Running))
+            foreach (var tenant in _shellHost.GetAllSettings(TenantState.Running))
             {
                 model.Tenants.Add(new CreateOpenIdScopeViewModel.TenantEntry
                 {
-                    Current = string.Equals(tenant.Name, _shellSettings.Name),
+                    Current = String.Equals(tenant.Name, _shellSettings.Name, StringComparison.OrdinalIgnoreCase),
                     Name = tenant.Name
                 });
             }
@@ -135,19 +135,19 @@ namespace OrchardCore.OpenId.Controllers
                 Name = model.Name
             };
 
-            if (!string.IsNullOrEmpty(model.Resources))
+            if (!String.IsNullOrEmpty(model.Resources))
             {
                 descriptor.Resources.UnionWith(model.Resources.Split(' ', StringSplitOptions.RemoveEmptyEntries));
             }
 
             descriptor.Resources.UnionWith(model.Tenants
                 .Where(tenant => tenant.Selected)
-                .Where(tenant => !string.Equals(tenant.Name, _shellSettings.Name))
+                .Where(tenant => !String.Equals(tenant.Name, _shellSettings.Name))
                 .Select(tenant => OpenIdConstants.Prefixes.Tenant + tenant.Name));
 
             await _scopeManager.CreateAsync(descriptor);
 
-            if (string.IsNullOrEmpty(returnUrl))
+            if (String.IsNullOrEmpty(returnUrl))
             {
                 return RedirectToAction("Index");
             }
@@ -178,16 +178,16 @@ namespace OrchardCore.OpenId.Controllers
 
             var resources = await _scopeManager.GetResourcesAsync(scope);
 
-            model.Resources = string.Join(" ",
+            model.Resources = String.Join(" ",
                 from resource in resources
-                where !string.IsNullOrEmpty(resource) && !resource.StartsWith(OpenIdConstants.Prefixes.Tenant, StringComparison.Ordinal)
+                where !String.IsNullOrEmpty(resource) && !resource.StartsWith(OpenIdConstants.Prefixes.Tenant, StringComparison.Ordinal)
                 select resource);
 
-            foreach (var tenant in _shellHost.GetAllSettings().Where(s => s.State == TenantState.Running))
+            foreach (var tenant in _shellHost.GetAllSettings(TenantState.Running))
             {
                 model.Tenants.Add(new EditOpenIdScopeViewModel.TenantEntry
                 {
-                    Current = string.Equals(tenant.Name, _shellSettings.Name),
+                    Current = String.Equals(tenant.Name, _shellSettings.Name, StringComparison.OrdinalIgnoreCase),
                     Name = tenant.Name,
                     Selected = resources.Contains(OpenIdConstants.Prefixes.Tenant + tenant.Name)
                 });
@@ -214,7 +214,7 @@ namespace OrchardCore.OpenId.Controllers
             if (ModelState.IsValid)
             {
                 var other = await _scopeManager.FindByNameAsync(model.Name);
-                if (other != null && !string.Equals(
+                if (other != null && !String.Equals(
                     await _scopeManager.GetIdAsync(other),
                     await _scopeManager.GetIdAsync(scope)))
                 {
@@ -237,19 +237,19 @@ namespace OrchardCore.OpenId.Controllers
 
             descriptor.Resources.Clear();
 
-            if (!string.IsNullOrEmpty(model.Resources))
+            if (!String.IsNullOrEmpty(model.Resources))
             {
                 descriptor.Resources.UnionWith(model.Resources.Split(' ', StringSplitOptions.RemoveEmptyEntries));
             }
 
             descriptor.Resources.UnionWith(model.Tenants
                 .Where(tenant => tenant.Selected)
-                .Where(tenant => !string.Equals(tenant.Name, _shellSettings.Name))
+                .Where(tenant => !String.Equals(tenant.Name, _shellSettings.Name, StringComparison.OrdinalIgnoreCase))
                 .Select(tenant => OpenIdConstants.Prefixes.Tenant + tenant.Name));
 
             await _scopeManager.UpdateAsync(scope, descriptor);
 
-            if (string.IsNullOrEmpty(returnUrl))
+            if (String.IsNullOrEmpty(returnUrl))
             {
                 return RedirectToAction("Index");
             }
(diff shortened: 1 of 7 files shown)


CrestApps:
@jtkech here is the PR we discussed in #12106. Let me know your thoughts

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -392,11 +392,7 @@ public async Task<IActionResult> Edit(string id)
                 return Forbid();
             }
 
-            var shellSettings = _shellHost.GetAllSettings()
-                .Where(x => String.Equals(x.Name, id, StringComparison.OrdinalIgnoreCase))
-                .FirstOrDefault();
-
-            if (shellSettings == null)
+            if (!_shellHost.TryGetSettings(id, out var shellSettings))
```
jtkech, line 395: Yes

That's funny, I already did the same in my PR #11890 for tenant removing, I did it because anyway there are many places where it is intended to pass the exact shell settings Name. By changing the inner dicos to be case insensitive we retrieve somewhere the old behavior where it was already not allowed to create 2 "Foo" and "foo" tenants.

You are breaking my PR ;)

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -453,11 +449,7 @@ public async Task<IActionResult> Edit(EditTenantViewModel model)
                 await ValidateViewModelAsync(model, false);
             }
 
-            var shellSettings = _shellHost.GetAllSettings()
-                .Where(x => String.Equals(x.Name, model.Name, StringComparison.OrdinalIgnoreCase))
-                .FirstOrDefault();
-
-            if (shellSettings == null)
+            if (!_shellHost.TryGetSettings(model.Name, out var shellSettings))
```
jtkech, line 452: yes

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -561,11 +549,7 @@ public async Task<IActionResult> Enable(string id)
                 return Forbid();
             }
 
-            var shellSettings = _shellHost.GetAllSettings()
-                .Where(x => String.Equals(x.Name, id, StringComparison.OrdinalIgnoreCase))
-                .FirstOrDefault();
-
-            if (shellSettings == null)
+            if (!_shellHost.TryGetSettings(id, out var shellSettings))
```
jtkech, line 552: Also look at `TenantValidator.cs` for another place to use `_shellHost.TryGetSettings()`.

https://github.com/OrchardCMS/OrchardCore/blob/af69cdb09a2959a76583f6538f20b9ae9cd185ac/src/OrchardCore.Modules/OrchardCore.Tenants/Services/TenantValidator.cs#L80-L85

src/OrchardCore/OrchardCore.Abstractions/Shell/Extensions/ShellHostExtensions.cs
```diff
@@ -50,5 +53,13 @@ public static ShellSettings GetSettings(this IShellHost shellHost, string tenant
 
             return settings;
         }
+
+        /// <summary>
+        /// Retrieves all shell settings with the given state
+        /// </summary>
+        public static IEnumerable<ShellSettings> GetAllSettings(this IShellHost shellHost, TenantState tenantState)
+        {
+            return shellHost.GetAllSettings().Where(x => x.State == tenantState);
```
jtkech, line 62: Maybe `GetSettingsByState()`

The comments could be `/// Retrieves the shell settings being in the specified state.` <= Use a dot at the end.

In the delegate at least use `settings` in place of `x`.

src/OrchardCore/OrchardCore/Shell/Distributed/DistributedShellHostedService.cs
```diff
@@ -103,7 +103,7 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
                     string shellChangedId;
                     try
                     {
-                        shellChangedId = await distributedCache.GetStringAsync(ShellChangedIdKey);
+                        shellChangedId = await distributedCache.GetStringAsync(ShellChangedIdKey, stoppingToken);
```
jtkech, line 106: You are really breaking my PR ;)

src/OrchardCore/OrchardCore/Shell/Distributed/DistributedShellHostedService.cs
```diff
@@ -141,7 +141,7 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
                     {
                         // Retrieve all new created tenants that are not already loaded.
                         var names = (await _shellSettingsManager.LoadSettingsNamesAsync())
-                            .Except(allSettings.Select(s => s.Name))
+                            .Except(allSettings.Select(s => s.Name), StringComparer.OrdinalIgnoreCase)
```
jtkech, line 144: No, same remark as for openId, I prefer to not change it here.

src/OrchardCore/OrchardCore/Shell/Distributed/DistributedShellHostedService.cs
```diff
@@ -185,7 +185,7 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
                             }
 
                             // Try to retrieve the reload identifier of this tenant from the distributed cache.
-                            var reloadId = await distributedCache.GetStringAsync(ReloadIdKey(settings.Name));
+                            var reloadId = await distributedCache.GetStringAsync(ReloadIdKey(settings.Name), stoppingToken);
```
jtkech, line 188: So, I prefer to not change this file.

I will add the stoppingToken in my PR #11890.

I already saw the analyzer suggesting it but was too lazy ;)

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
CrestApps: Yes, in the TeantValiadator we need all settings anyway to check for existing prefixes

src/OrchardCore/OrchardCore/Shell/Distributed/DistributedShellHostedService.cs
CrestApps: Got it! I’ll revert that change.

src/OrchardCore/OrchardCore/Shell/Distributed/DistributedShellHostedService.cs
CrestApps: I’ll revert my change

src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/ScopeController.cs
```diff
@@ -96,11 +96,11 @@ public async Task<IActionResult> Create(string returnUrl = null)
 
             var model = new CreateOpenIdScopeViewModel();
 
-            foreach (var tenant in _shellHost.GetAllSettings().Where(s => s.State == TenantState.Running))
+            foreach (var tenant in _shellHost.GetAllSettings(TenantState.Running))
             {
                 model.Tenants.Add(new CreateOpenIdScopeViewModel.TenantEntry
                 {
-                    Current = string.Equals(tenant.Name, _shellSettings.Name),
+                    Current = String.Equals(tenant.Name, _shellSettings.Name, StringComparison.OrdinalIgnoreCase),
```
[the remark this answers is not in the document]
CrestApps: Are you saying no need to check for variant here?

src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/ScopeController.cs
```diff
@@ -135,19 +135,19 @@ public async Task<IActionResult> Create(CreateOpenIdScopeViewModel model, string
...
-            if (!string.IsNullOrEmpty(model.Resources))
+            if (!String.IsNullOrEmpty(model.Resources))
             {
                 descriptor.Resources.UnionWith(model.Resources.Split(' ', StringSplitOptions.RemoveEmptyEntries));
             }
 
             descriptor.Resources.UnionWith(model.Tenants
                 .Where(tenant => tenant.Selected)
-                .Where(tenant => !string.Equals(tenant.Name, _shellSettings.Name))
+                .Where(tenant => !String.Equals(tenant.Name, _shellSettings.Name, StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
CrestApps: And here no need to check for the variant?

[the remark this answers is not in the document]
jtkech: Yes no need

src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/ScopeController.cs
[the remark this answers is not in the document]
jtkech: Yes

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
jtkech: Okay as you want

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
CrestApps: I see. You change will break existing code since it no longer check for invariant case. I think this change should be made to this PR since it has the complete fix. Let me know if you want me to recent this change anyway.

jtkech: First yes, you can do these kind of changes here, no problem.

> I think the change in the other PR will break existing code

It was intentional as in some place the passed tenant name is intended to be the exact tenant name, e.g the Tenants view generate action link based on existing shellSettings.Names, idem for the openId where tenants has been selected from existing shellSettings.Names.

Mainly I mean we only need to update the places where we were doing ignore case lookup which is now useless, but not the opposite, hmm I think we only need to update the tenants admin controller.

CrestApps: @jtkech can you kindly review the code change one more time? I reverted some of the original changes as per your request.

jtkech: No problem, will do it tomorrow.

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -521,16 +506,12 @@ public async Task<IActionResult> Disable(string id)
...
             {
                 return NotFound();
             }
 
-            if (String.Equals(shellSettings.Name, ShellHelper.DefaultShellName, StringComparison.OrdinalIgnoreCase))
+            if (shellSettings.IsDefaultShell())
```
jtkech, line 514: Yes as the related extension does the same currently, but not sure I like this extension.

1. To be more strict before doing an action **not allowed on the Default tenant**, we have better to do a case insensitive comparison. So here okay as this is what the extension currently does.

2. But in other places, before doing an action only **allowed by the Default tenant**, I would say that we have better to do a case sensitive comparison, in that case `IsDefaultShell()` is less restrictive and not right for me.

We could remove `IsDefaultShell()` and just use `==` or `String.Equals(...OrdinalIgnoreCase)` as here if it still makes sense ;) but removing it is a breaking change, or just make this extension cases sensitive (see below).

But `IsDefaultShell()` is always used on a registered shellSettings, we can create any non registered shell with any name, but we can't register e.g. a `dEfAuLt` tenant if the tenant validator was used, and moreover with the new case insensitive dicos it would have already failed on a duplicate key.

So anyway I think that's okay if `IsDefaultShell()` is case sensitive or not.

---

So first we could change `IsDefaultShell()` like this as it makes more sense for me, in most of the places that we are using it we are in the above case number `2.`.

    public bool IsDefaultShell() => Name == ShellHelper.DefaultShellName;

Then because it would be case sensitive, revert the usage of `IsDefaultShell()` **here and around line 254/248** where we are in the case number `1.`. OR because of the above analysis, keep this change **here and line 254/248** just because the code is cleaner.

I'm hesitating ;)

**In any case you can use** `IsDefaultShell()` **to init the** `ShellSettingsEntry.IsDefaultTenant` **line 108**, which is checked in the index view and where we are in the case number `2.`. At least it would be as before if we don't change the extension, and better if we change it.

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -235,23 +235,17 @@ public async Task<IActionResult> Index(BulkActionViewModel model)
...
+                if (!_shellHost.TryGetSettings(tenantName, out var shellSettings))
                 {
                     break;
                 }
 
                 switch (model.BulkAction.ToString())
                 {
                     case "Disable":
-                        if (String.Equals(shellSettings.Name, ShellHelper.DefaultShellName, StringComparison.OrdinalIgnoreCase))
+                        if (shellSettings.IsDefaultShell())
```
jtkech, line 248: We will keep this change or not, see my other and more detailed review comment.

jtkech:
@CrestApps 

Okay looks good, only some thoughts that could be dealed here or on separate issues/PRs.

See my too long review comment on the `IsDefaultShell()` extension.

---

Just looked at the TenantValidator, with the new case insensitive dicos we could do around line 80.


    if (_shellHost.TryGetSettings(model.Name, out var settings) && model.IsNewTenant)
    {
        errors.Add(new ModelError(nameof(model.Name), S["...."]));
    }

    var allOtherSettings = _shellHost.GetAllSettings().Where(s => s != settings);

    if (allOtherSettings.Any(tenant => ...))
    {
        errors.Add(new ModelError(nameof(model.RequestUrlPrefix), S["..."]));
    }

---

Also for info in TenantValidator I think we can just remove the following lines. TenantValidator is injected and used by the default tenant so here `!_shellSettings.IsDefaultShell()` is always false. And a tenant which is not the Default tenant can have an empty Host and empty url prefix if the Default tenant has a non empty url prefix (which is allowed and works). We only need to check that 2 tenants don't have the same Host and same prefix, which is checked afterwards. But maybe in a separate issue/PR ;)

    if (!_shellSettings.IsDefaultShell() &&
        String.IsNullOrWhiteSpace(model.RequestUrlHost) &&
        String.IsNullOrWhiteSpace(model.RequestUrlPrefix))
    {
        errors.Add(new ModelError(nameof(model.RequestUrlPrefix),
            S["Host and url prefix can not be empty at the same time."]));
    }

CrestApps:
> @CrestApps
> 
> Okay looks good, only some thoughts that could be dealed here or on separate issues/PRs.
> 
> See my too long review comment on the `IsDefaultShell()` extension.
> 
> Just looked at the TenantValidator, with the new case insensitive dicos we could do around line 80.
> 
> ```
> if (_shellHost.TryGetSettings(model.Name, out var settings) && model.IsNewTenant)
> {
>     errors.Add(new ModelError(nameof(model.Name), S["...."]));
> }
> 
> var allOtherSettings = _shellHost.GetAllSettings().Where(s => s != settings);
> 
> if (allOtherSettings.Any(tenant => ...))
> {
>     errors.Add(new ModelError(nameof(model.RequestUrlPrefix), S["..."]));
> }
> ```
> 
> Also for info in TenantValidator I think we can just remove the following lines. TenantValidator is injected and used by the default tenant so here `!_shellSettings.IsDefaultShell()` is always false. And a tenant which is not the Default tenant can have an empty Host and empty url prefix if the Default tenant has a non empty url prefix (which is allowed and works). We only need to check that 2 tenants don't have the same Host and same prefix, which is checked afterwards. But maybe in a separate issue/PR ;)
> 
> ```
> if (!_shellSettings.IsDefaultShell() &&
>     String.IsNullOrWhiteSpace(model.RequestUrlHost) &&
>     String.IsNullOrWhiteSpace(model.RequestUrlPrefix))
> {
>     errors.Add(new ModelError(nameof(model.RequestUrlPrefix),
>         S["Host and url prefix can not be empty at the same time."]));
> }
> ```

@jtkech 
I need to avoid making any change to the tenant validator as I have another big PR that is making lots of changes to it. I don’t want to break the other change now. I’ll review this more and can make necessary changes in the other PR. Actually you may also want to review that PR https://github.com/OrchardCMS/OrchardCore/pull/11822.  I’ll review more of your other feedback tomorrow and respond.  Thank you for your detailed feedback.

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
hishamco: I think I already did a PR for this couple months ago, let me find it

hishamco: It's here #11435, so please don't re-invent the wheel, furthermore check @jtkech there ;)

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
CrestApps: I am not sure that I agree, or may be I do but could be missing the point. 

First, `IsDefaultShell()` checks if the settings is a host settings. Since this PR is allowing invariant tenant, then `Default`, `dEfAuLt`, `default` or any other form should be treated the same. So I think, `IsDefaultShell() => String.Equals(Name, ShellHelper.DefaultShellName, StringComparison.OrdinalIgnoreCase)` is valid.

In fact, I think we should add another extension method to `ShellSettings` to clean up some of the other code. I think we should add `IsShell(string tenantName) => String.Equals(Name, tenantName, StringComparison.OrdinalIgnoreCase);`

I don't know that I like the name `IsShell` perhaps you can come up with a better name if you agree. But at the end the combined code will look like this

```
public bool IsDefaultShell() => IsShell(ShellHelper.DefaultShellName);
public bool IsShell(string tenantName) => String.Equals(Name, tenantName, StringComparison.OrdinalIgnoreCase);
```

Now, instead of having to write code like `if (String.Equals(shellSettings.Name, "name", StringComparison.OrdinalIgnoreCase))` we would simplify it to `shellSettings.IsShell("name")`.

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
CrestApps: @hishamco I see that now. Unfortunately, this change is in 3 different PR which is why I hate when PR site in the queue for so long. So merging any of these 3 PRs, will cause a conflict in two.

Unless you mind, I would like to keep the change here too and when the first PR is merged, we'll just fix this simple conflict.

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
jtkech: > Since this PR is allowing invariant tenant, then Default, dEfAuLt, default or any other form should be treated the same.

I don't think so, this PR only allows to retrieve a tenant by name in a case insensitive way, then for example the name of the default registered tenant is still always equal to the const name `ShellHelper.DefaultShellName` which currently is `Default`.

Anyway I think it is better to cleanup this PR, maybe open a new one to only focus on the new case insensitive shell inner dictionaries. I will comment it in the main thread.

jtkech:
@CrestApps 

Okay, I was not aware of all these mixed PRs e.g. #11435 where I retrieved one of my comments https://github.com/OrchardCMS/OrchardCore/pull/11435#issuecomment-1133115849 where I already didn't totally agree.

So I suggest that you close this PR, then start a new one which only focus on the new shell inner dictionaries that would be case insensitive.

Then I will try to do separate PRs for the other little concerns we discussed here, maybe better after this PR if it get merged because the context would not be the same.

CrestApps:
Closing this in favor of #12120

CrestApps:
@jtkech 
> Just looked at the TenantValidator, with the new case insensitive dicos we could do around line 80.

I made the suggested changes to the `TenantValidator` in PR #11822. Actually, with this change, the `TenantValidatorTests` fail because it was using `GetAllSettings()` and manually compare names. I think by using `TryGetSettings()` the code is more accurate and more bullet proof to changes in the ShellHost instead of just comparing names.

jtkech:
Yes I think you're right but normally the TenantValidatorTests should have still worked.

Would need to look at it more in depth.

Also as said in TenantValidator there is a wrong usage of `_shellSettings.IsDefaultShell()` as normally the injected `_shellSettings` is always the Default shell so no need to inject and use it. And it is used to check if a non default tenant has both an empty host and prefix but this is allowed for any tenant, so this check can be removed if afterwards we check for host/prefix duplicates.

It works in the unit test as we create a validator that we let be used by a non default tenant, so in this case `_shellSettings.IsDefaultShell()` is not always the same.

Also I didn't take a look at the AutoSetup as it would need to be consistent, I don't think it already uses the tenant validator.

Hmm, I'm afraid that we already saw too many things (all simple things) that would need to be done before adding new functionalities.

I suggest to first do separate PR(s) to tweak the existing code now that we have case insensitive dictionaries, then let decide what we do with `IsDefaultShell()`, then tweak the TenantValidator and make the AutoSetup consistent with these new things.

Then and only after that, even if there are new conflicts in PRs, restart to add new features on this base.

CrestApps:
@jtkech i am not sure I follow you on what duplicate check that should be removed. if you can, please leave a review directly on that PR so I can correct it.

As for the `TeantValiadatorTest`, the reason it failed is because we mocked GetAllSettings to return in memory collection. Meanwhile. The TryGetSettings looks at a different/internal collection to validate if the given tenant exists. So in this case TryGetSettings() will always compare an empty collection. The change I made, is using an instance of ShellHost now which comparing same internal collection is both placed which is why the update test works.

CrestApps:
``

jtkech:
Okay cool then

> please leave a review directly on that PR so I can correct it.

The lines to remove (that don't appear in your PR changes) are

            if (!_shellSettings.IsDefaultShell() && String.IsNullOrWhiteSpace(model.RequestUrlHost) && String.IsNullOrWhiteSpace(model.RequestUrlPrefix))
            {
                errors.Add(new ModelError(nameof(model.RequestUrlPrefix), S["Host and url prefix can not be empty at the same time."]));
            }

The tenant validator is only intended to be used in the context of the `Default` tenant, so `!_shellSettings.IsDefaultShell()` always have the same false value, so no need to inject `_shellSettings`. It works in the test as we let the validator be used by non default tenants.

To fix this check we would need to check the `model.Name` instead I think and see if it is equal to the DefaultShellName but here directly on a string and here in a case insensitive way as it doesn't come from a registered shellSettings.

But anyway this check prevent from having both an empty host and prefix which in fact is allowed, so we could just remove the above lines and don't inject anymore `_shellSettings`.

But need to check step by step.

PR #12120, opened 2022-07-30
Can Get Shell By Case Insensitive Name.

### LINKED ISSUE #12106: Allow case-insensitive tenant names in ShellHost

CrestApps:
Fix #12106

@jtkech this PR replaces #12114 as per your request

### DIFF
--- a/src/OrchardCore/OrchardCore/Shell/ShellHost.cs
+++ b/src/OrchardCore/OrchardCore/Shell/ShellHost.cs
@@ -64,11 +64,14 @@ namespace OrchardCore.Environment.Shell
             await _initializingSemaphore.WaitAsync();
             try
             {
-                await PreCreateAndRegisterShellsAsync();
-                _initialized = true;
+                if (!_initialized)
+                {
+                    await PreCreateAndRegisterShellsAsync();
+                }
             }
             finally
             {
+                _initialized = true;
                 _initializingSemaphore.Release();
             }
         }
@@ -470,8 +473,6 @@ namespace OrchardCore.Environment.Shell
             {
                 shell.Dispose();
             }
-
-            GC.SuppressFinalize(this);
         }
     }
 }
--- a/src/OrchardCore/OrchardCore/Shell/ShellHost.cs
+++ b/src/OrchardCore/OrchardCore/Shell/ShellHost.cs
@@ -30,10 +30,10 @@ namespace OrchardCore.Environment.Shell
         private readonly ILogger _logger;
 
         private bool _initialized;
-        private readonly ConcurrentDictionary<string, ShellContext> _shellContexts = new ConcurrentDictionary<string, ShellContext>();
-        private readonly ConcurrentDictionary<string, ShellSettings> _shellSettings = new ConcurrentDictionary<string, ShellSettings>();
-        private readonly ConcurrentDictionary<string, SemaphoreSlim> _shellSemaphores = new ConcurrentDictionary<string, SemaphoreSlim>();
-        private SemaphoreSlim _initializingSemaphore = new SemaphoreSlim(1);
+        private readonly ConcurrentDictionary<string, ShellContext> _shellContexts = new(StringComparer.OrdinalIgnoreCase);
+        private readonly ConcurrentDictionary<string, ShellSettings> _shellSettings = new(StringComparer.OrdinalIgnoreCase);
+        private readonly ConcurrentDictionary<string, SemaphoreSlim> _shellSemaphores = new();
+        private readonly SemaphoreSlim _initializingSemaphore = new(1);
 
         public ShellHost(
             IShellSettingsManager shellSettingsManager,
@@ -55,22 +55,21 @@ namespace OrchardCore.Environment.Shell
 
         public async Task InitializeAsync()
         {
-            if (!_initialized)
+            if (_initialized)
             {
-                // Prevent concurrent requests from creating all shells multiple times
-                await _initializingSemaphore.WaitAsync();
-                try
-                {
-                    if (!_initialized)
-                    {
-                        await PreCreateAndRegisterShellsAsync();
-                    }
-                }
-                finally
-                {
-                    _initialized = true;
-                    _initializingSemaphore.Release();
-                }
+                return;
+            }
+
+            // Prevent concurrent requests from creating all shells multiple times
+            await _initializingSemaphore.WaitAsync();
+            try
+            {
+                await PreCreateAndRegisterShellsAsync();
+                _initialized = true;
+            }
+            finally
+            {
+                _initializingSemaphore.Release();
             }
         }
 
@@ -183,10 +182,8 @@ namespace OrchardCore.Environment.Shell
             }
 
             var count = 0;
-            while (count < ReloadShellMaxRetriesCount)
+            while (count++ < ReloadShellMaxRetriesCount)
             {
-                count++;
-
                 if (_shellContexts.TryRemove(settings.Name, out var context))
                 {
                     _runningShellTable.Remove(settings);
@@ -449,7 +446,7 @@ namespace OrchardCore.Environment.Shell
         /// <summary>
         /// Whether or not a shell can be activated and added to the running shells.
         /// </summary>
-        private bool CanRegisterShell(ShellSettings shellSettings)
+        private static bool CanRegisterShell(ShellSettings shellSettings)
         {
             return
                 shellSettings.State == TenantState.Running ||
@@ -473,6 +470,8 @@ namespace OrchardCore.Environment.Shell
             {
                 shell.Dispose();
             }
+
+            GC.SuppressFinalize(this);
         }
     }
 }
--- a/test/OrchardCore.Tests/Shell/ShellHostTests.cs
+++ b/test/OrchardCore.Tests/Shell/ShellHostTests.cs
@@ -1,7 +1,6 @@
 using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
 using OrchardCore.Environment.Shell;
-using OrchardCore.Environment.Shell.Builders;
 using OrchardCore.Environment.Shell.Models;
 using OrchardCore.Tests.Apis.Context;
 using Xunit;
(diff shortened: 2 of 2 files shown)


src/OrchardCore/OrchardCore/Shell/ShellHost.cs
```diff
@@ -473,6 +470,8 @@ public void Dispose()
             {
                 shell.Dispose();
             }
+
+            GC.SuppressFinalize(this);
```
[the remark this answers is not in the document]
CrestApps: So no need for the finalizer here?

src/OrchardCore/OrchardCore/Shell/ShellHost.cs
```diff
@@ -55,22 +55,21 @@ public class ShellHost : IShellHost, IDisposable
...
-                    if (!_initialized)
-                    {
-                        await PreCreateAndRegisterShellsAsync();
-                    }
-                }
-                finally
-                {
-                    _initialized = true;
-                    _initializingSemaphore.Release();
-                }
```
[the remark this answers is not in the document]
CrestApps: Maybe we should move the second check in  `PreCreateAndRegisterShellsAsync` instead? So if that method is ever called again it won’t be executed

test/OrchardCore.Tests/Shell/ShellHostTests.cs
```diff
@@ -0,0 +1,63 @@
...
+            Name = exactName,
+        };
+
+        await ShellHost.GetOrCreateShellContextAsync(testSettings);
+
+        ShellHost.TryGetSettings(exactName, out var foundExactShellSettings);
+        ShellHost.TryGetSettings(exactName, out var foundExactShellShellContext);
+
+        ShellHost.TryGetSettings("test", out var foundMixedShellSettings);
+        ShellHost.TryGetSettings("test", out var foundMixedShellShellContext);
```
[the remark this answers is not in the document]
CrestApps: Using `foundMixedShellSettings` asserts that we can get the same setting using case-insensitive (i.e, find using “test” when the name is “Test”) where as `foundShellSettings` just assets that it’s able to get the same exact name. Perhaps asserting that we can get the same exact this is not too important?

test/OrchardCore.Tests/Shell/ShellHostTests.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        var testSettings = new ShellSettings()
+        {
+            Name = exactName,
+        };
+
+        await ShellHost.GetOrCreateShellContextAsync(testSettings);
+
+        ShellHost.TryGetSettings(exactName, out var foundExactShellSettings);
+        ShellHost.TryGetSettings(exactName, out var foundExactShellShellContext);
```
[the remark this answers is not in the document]
CrestApps: Both shellSettings and ShellContext are now case insensitive. So I want to assert both get using case-insensitive name

src/OrchardCore/OrchardCore/Shell/ShellHost.cs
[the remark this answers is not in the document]
CrestApps: Fixed

test/OrchardCore.Tests/Shell/ShellHostTests.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            Name = ShellHelper.DefaultShellName,
+        };
+
+        await ShellHost.GetOrCreateShellContextAsync(defaultSettings);
+
+        ShellHost.TryGetSettings("dEFaUlT", out var defaultTenant);
+
+        Assert.NotNull(defaultTenant);
+        Assert.True(defaultTenant.IsDefaultShell());
```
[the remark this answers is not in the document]
CrestApps: What about the updated test now?

test/OrchardCore.Tests/Shell/ShellHostTests.cs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    [Fact]
+    public static async Task FindCaseInsensitiveShellSettings()
+    {
+        const string exactName = "Test";
+
+        var testSettings = new ShellSettings()
+        {
+            Name = exactName,
```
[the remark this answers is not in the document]
CrestApps: Done

test/OrchardCore.Tests/Shell/ShellHostTests.cs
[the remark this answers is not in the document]
CrestApps: My fault, I was using TryGetSettings instead of using TryGetShellContext. Please checkout the updated test

jtkech:
@CrestApps 

Okay will work on it tomorrow.

Just did some little tweaks locally for the unit tests,

Will push it tomorrow to let you see them.

hishamco:
@CrestApps why you closed #12114 it is not a good idea to close a PR and replace it unless there's a value. I had seen a lot of comments in the previous one during the review. Hope to consider this next time and thanks again for your contribution

test/OrchardCore.Tests/Shell/ShellHostTests.cs
```diff
@@ -0,0 +1,95 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using OrchardCore.Environment.Shell;
+using OrchardCore.Environment.Shell.Models;
+using OrchardCore.Tests.Apis.Context;
+using Xunit;
+
+namespace OrchardCore.Tests.Shell;
+public class ShellHostTests : SiteContext
+{
+    public static IShellHost ShellHost { get; }
```
hishamco, line 11: Move it down after the constructor

test/OrchardCore.Tests/Shell/ShellHostTests.cs
```diff
@@ -0,0 +1,95 @@
...
+
+    static ShellHostTests()
+    {
+        ShellHost = Site.Services.GetRequiredService<IShellHost>();
+    }
+
+    [Fact]
+    public static async Task FindCaseInsensitiveShellSettings()
```
hishamco, line 19: Can we use `Theory` for the first two unit tests while they are similar?

test/OrchardCore.Tests/Shell/ShellHostTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        };
+
+        await ShellHost.GetOrCreateShellContextAsync(defaultSettings);
+
+        ShellHost.TryGetSettings("dEFaUlT", out var defaultTenant);
+
+        Assert.NotNull(defaultTenant);
+    }
+
+
```
hishamco, line 56: Remove extra lines

test/OrchardCore.Tests/Shell/ShellHostTests.cs
```diff
@@ -0,0 +1,95 @@
...
+            State = TenantState.Uninitialized,
+        };
+
+        await ShellHost.GetOrCreateShellContextAsync(defaultSettings);
+
+        ShellHost.TryGetShellContext("dEFaUlT", out var defaultTenant);
+
+        Assert.NotNull(defaultTenant);
+    }
+}
```
hishamco, line 95: add new line after

test/OrchardCore.Tests/Shell/ShellHostTests.cs
hishamco, line 11: Why `ShellHost` static?

CrestApps:
I did that based on the request from @jtkech

test/OrchardCore.Tests/Shell/ShellHostTests.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        Assert.NotNull(foundExactShellShellContext);
+        Assert.NotNull(foundMixedShellShellContext);
+        Assert.Null(missingShellShellContext);
+    }
+
+    [Fact]
+    public static async Task FindCaseInsensitiveDefaultShellContext()
+    {
+        var defaultSettings = new ShellSettings()
```
hishamco, line 83: You can simplify by introducing `GetOrCreateShellContext()` as private method

hishamco (changes requested):
I will leave the Shell related stuff to @jtkech, just a little notes about the unit tests

test/OrchardCore.Tests/Shell/ShellHostTests.cs
CrestApps: @hishamco  Can you please explain why you are requesting to use theory? It's my understanding that `[Thoery]` would be used when there are parameters involved and `[Fact]` where there are no parameters.

CrestApps: I think I understood your request. Please check the updated tests

test/OrchardCore.Tests/Shell/ShellHostTests.cs
```diff
@@ -0,0 +1,80 @@
...
+        var shellSettings = new ShellSettings()
+        {
+            Name = name,
+            State = TenantState.Uninitialized,
+        };
+
+        return await _shellHost.GetOrCreateShellContextAsync(shellSettings);
+    }
+
+    private static string GetDifferentVariant(string name)
```
hishamco, line 69: I'm not sure why we need this, you can supply various casing values in `Theory` args

CrestApps: I thought this way we only provide one variable and the logic will compare it to another variant. This way we don’t have to provide to parameters. Is there a reason we should provide two parameters instead?

hishamco: Adding values with different cases will be more obvious to everyone instead of adding logic that doing something we don't know until we dig into the test

test/OrchardCore.Tests/Shell/ShellHostTests.cs
```diff
@@ -38,13 +34,12 @@ public async Task FindCaseInsensitiveShellSettings(string shellName)
     }
```
hishamco, line 34:
```csharp
Assert.NotEqual(shellName, anotherVariantOfShellName);
```

This will never fail while the values are different. Please let us know which values should be tested?

CrestApps: It will fail if the user accident provided equal parameters in the InlineData. I would leave it just in case someone made a change to the test. If both parameters are equal the test will pass but won't really test anything.

jtkech: Cool, you are testing the test ;)

CrestApps: Yup

jtkech:
@hishamco

I already said to @CrestApps that I will update the tests with some suggestions, so maybe was too early to review them. The goal being to keep this PR as clean as possible because it acts on a sensitive  area.

> Just did some little tweaks locally for the unit tests,
> Will push it tomorrow to let you see them.

But that's okay as we still only have 2 files changed, so the PR is still "clean" ;)

I will do my suggestions soon.

Anyway the main point is to make `_shellContexts` and `_shellSettings` dictionaries case insensitive, I just need to think about it a little more and if I still think it is safe I will merge it (and if unit tests still pass).

Otherwise will wait for more feedback.

jtkech:
@CrestApps 

I pushed my suggestions, feel free to revert or update what you don't like ;)

test/OrchardCore.Tests/Shell/ShellHostTests.cs
```diff
@@ -10,54 +9,32 @@ namespace OrchardCore.Tests.Shell;
 
 public class ShellHostTests : SiteContext
 {
-    private readonly IShellHost _shellHost;
+    public static IShellHost ShellHost { get; }
```
hishamco, line 12: @jtkech why you mark it as static? @CrestApps already did that at the beginning but I asked him to change it ;)

jtkech: Yes I saw this, because Site is a static thing so all can be static

hishamco:
Seems the build is failed

jtkech:
I saw it, I will fix it

Fail intermitently while running all tests in parallel and only for the Default tenant

PR #12132, opened 2022-08-02
added better description for partname being null

brunoAltinet:
I'm seeing a lot of these errors when developing with orchard and there's not much help about it. So i decided to add a safeguard and some context when error is thrown

image

I also refactored a lot of this repetitive code into one method.

### DIFF
--- a/src/OrchardCore/OrchardCore.ContentManagement/Handlers/ContentPartHandlerCoordinator.cs
+++ b/src/OrchardCore/OrchardCore.ContentManagement/Handlers/ContentPartHandlerCoordinator.cs
@@ -120,23 +120,6 @@ namespace OrchardCore.ContentManagement.Handlers
                 return;
             await InvokeHandlersForParts(contentTypeDefinition, context, (handler, context, part) => handler.InitializedAsync(context, part));
         }
-        private async Task InvokeHandlersForParts<TContext>(ContentTypeDefinition contentTypeDefinition, TContext context, Func<IContentPartHandler, TContext, ContentPart, Task> dispatch) where TContext : ContentContextBase
-        {
-            foreach (var typePartDefinition in contentTypeDefinition.Parts)
-            {
-                var partName = typePartDefinition.PartDefinition.Name;
-                if (string.IsNullOrEmpty(partName))
-                    throw new ArgumentNullException("partName", $"Part name for content type {contentTypeDefinition.Name} cannot be null");
-                var activator = _contentPartFactory.GetTypeActivator(partName);
-                var part = context.ContentItem.Get(activator.Type, partName) as ContentPart;
-
-                if (part != null)
-                {
-                    var partHandlers = _contentPartHandlerResolver.GetHandlers(partName);
-                    await partHandlers.InvokeAsync(dispatch, context, part, _logger);
-                }
-            }
-        }
 
         public override async Task LoadingAsync(LoadContentContext context)
         {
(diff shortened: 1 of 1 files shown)


brunoAltinet:
I realize that this refactor might not be what you want to see since it's a critical part of the system, so let me know if i went overboard:)

src/OrchardCore/OrchardCore.ContentManagement/Handlers/ContentPartHandlerCoordinator.cs
```diff
@@ -575,41 +343,36 @@ public override async Task GetContentItemAspectAsync(ContentItemAspectContext co
...
                 return;
+            await InvokeHandlersForParts(contentTypeDefinition, context, (handler, context, part) => handler.CloningAsync(context, part));
+        }
 
+        private async Task InvokeHandlersForParts<TContext>(ContentTypeDefinition contentTypeDefinition, TContext context, Func<IContentPartHandler, TContext, ContentPart, Task> dispatch) where TContext : ContentContextBase
+        {
             foreach (var typePartDefinition in contentTypeDefinition.Parts)
             {
                 var partName = typePartDefinition.PartDefinition.Name;
+                if (string.IsNullOrEmpty(partName))
```
CrestApps, line 367: I could be wrong, but you may want to log the error and continue instead of throwing exception. I would think this mean that the part does not exists.

Also, please use Framework type instead of language `String.IsNullOrEmpty(partName)`

CrestApps:
You may want to create an issue and explain the exact issue and how to reproduce it.

src/OrchardCore/OrchardCore.ContentManagement/Handlers/ContentPartHandlerCoordinator.cs
brunoAltinet: The line afterwards throws the error anyway and it gets logged like in description, i'm just making the error a bit more descriptive. So String instead of string?

brunoAltinet:
From my understanding, this happens when a content type has a part whose definition is either missing or has been removed from the system.

brunoAltinet:
I also noticed slight inconsistency in how some of these methods were previously implemented.
Some of them fetch the part like this
image
while other do it like this:
image
ie. one is using part type, the other is using the part name. I think that content type part is more correct

src/OrchardCore/OrchardCore.ContentManagement/Handlers/ContentPartHandlerCoordinator.cs
```diff
@@ -223,6 +115,8 @@ public override async Task LoadingAsync(LoadContentContext context)
...
```
CrestApps, line 118: Use Framework language and add curly bracket

src/OrchardCore/OrchardCore.ContentManagement/Handlers/ContentPartHandlerCoordinator.cs
```diff
@@ -223,6 +115,8 @@ public override async Task LoadingAsync(LoadContentContext context)
...
+                    throw new System.ArgumentNullException("partName cannot be null for contentType:" + contentTypeDefinition.Name);
```
CrestApps, line 119: Instead of throwing an exception, log it and continue "so it skips the execution" instead.

src/OrchardCore/OrchardCore.ContentManagement/Handlers/ContentPartHandlerCoordinator.cs
```diff
@@ -575,26 +271,17 @@ public override async Task GetContentItemAspectAsync(ContentItemAspectContext co
...
-                }
-            }
+            await InvokeHandlersForParts(context, (handler, context, part) => handler.ClonedAsync(context, part));
         }
         public override async Task CloningAsync(CloneContentContext context)
+        {
+            await InvokeHandlersForParts(context, (handler, context, part) => handler.CloningAsync(context, part));
+        }
+
+        private async Task InvokeHandlersForParts<TContext>(TContext context, Func<IContentPartHandler, TContext, ContentPart, Task> dispatch) where TContext : ContentContextBase
```
CrestApps, line 284: May be `InvokePartHandlers` for the name?

src/OrchardCore/OrchardCore.ContentManagement/Handlers/ContentPartHandlerCoordinator.cs
```diff
@@ -603,13 +290,15 @@ public override async Task CloningAsync(CloneContentContext context)
             foreach (var typePartDefinition in contentTypeDefinition.Parts)
             {
                 var partName = typePartDefinition.PartDefinition.Name;
+                if (String.IsNullOrEmpty(partName))
```
CrestApps, line 293: Use curly bracket. Log the error and continue execution instead of throwing an exception.

CrestApps:
Looks good to me. We’ll see how the member feel about the change. I like the refactoring part, but, I am sure some will dislike it.

brunoAltinet:
thanks. Do note about that discrepancy that i considered a bug where 
OnActivated a part is fetched with 
                var part = context.ContentItem.Get(activator.Type, partName) as ContentPart;
while most of them do it with
                var part = context.ContentItem.Get(activator.Type, typePartDefinition.Name) as ContentPart;
 
 so i used the same private method and went with typePardDefinition.Name

sebastienros:
This file has been changed tremendously since this PR

PR #12165, opened 2022-08-09
Decouple OC.Tenants from an ISiteService registration

jtkech:
Otherwise the `AdminController` fails to inject `ISiteService`.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Tenants/Manifest.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Tenants/Manifest.cs
@@ -11,6 +11,10 @@ using OrchardCore.Modules.Manifest;
     Id = "OrchardCore.Tenants",
     Name = "Tenants",
     Description = "Provides a way to manage tenants from the admin.",
+    Dependencies = new[]
+    {
+        "OrchardCore.Settings",
+    },
     Category = "Infrastructure",
     DefaultTenantOnly = true
 )]
--- a/src/OrchardCore.Modules/OrchardCore.Tenants/Manifest.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Tenants/Manifest.cs
@@ -11,10 +11,6 @@ using OrchardCore.Modules.Manifest;
     Id = "OrchardCore.Tenants",
     Name = "Tenants",
     Description = "Provides a way to manage tenants from the admin.",
-    Dependencies = new[]
-    {
-        "OrchardCore.Settings",
-    },
     Category = "Infrastructure",
     DefaultTenantOnly = true
 )]
--- a/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
@@ -93,7 +93,6 @@ namespace OrchardCore.Tenants.Controllers
             var dataProtector = _dataProtectorProvider.CreateProtector("Tokens").ToTimeLimitedDataProtector();
 
             var pageSize = 10;
-
             var siteService = HttpContext.RequestServices.GetService<ISiteService>();
             if (siteService != null)
             {
--- a/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
@@ -8,6 +8,7 @@ using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Localization;
 using Microsoft.AspNetCore.Mvc.Rendering;
 using Microsoft.AspNetCore.Routing;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Localization;
 using OrchardCore.Data;
 using OrchardCore.DisplayManagement;
@@ -37,7 +38,6 @@ namespace OrchardCore.Tenants.Controllers
         private readonly IDataProtectionProvider _dataProtectorProvider;
         private readonly IClock _clock;
         private readonly INotifier _notifier;
-        private readonly ISiteService _siteService;
         private readonly ITenantValidator _tenantValidator;
 
         private readonly dynamic New;
@@ -55,7 +55,6 @@ namespace OrchardCore.Tenants.Controllers
             IDataProtectionProvider dataProtectorProvider,
             IClock clock,
             INotifier notifier,
-            ISiteService siteService,
             ITenantValidator tenantValidator,
             IShapeFactory shapeFactory,
             IStringLocalizer<AdminController> stringLocalizer,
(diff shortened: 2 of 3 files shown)


CrestApps:
@jtkech I think this is because `ResourceDebugMode ` and `ResourceOptions ` which is addressed in pr#11569

do you this this is going to still be needed!

src/OrchardCore.Modules/OrchardCore.Tenants/Manifest.cs
```diff
@@ -11,6 +11,10 @@
     Id = "OrchardCore.Tenants",
     Name = "Tenants",
     Description = "Provides a way to manage tenants from the admin.",
+    Dependencies = new[]
+    {
+        "OrchardCore.Settings",
```
hishamco, line 16: Why we need settings in Tenants? Any use case require are to add this dependency?

hishamco:
Seems you need it in your PR #11890 ;) If it's like that I'm suggesting to do the changes in the other PR, so everyone will know what's the need for `OC.Settings`

src/OrchardCore.Modules/OrchardCore.Tenants/Manifest.cs
CrestApps: I think it’s because `ISiteService` gets registered when OC.Settings is enabled. so if OC.Settings we’re not enabled, `ISiteService` won’t exists yes OC.Tenants depends on it.

hishamco: Right, but I'm not sure if we are using `ISiteService` already used in `OC.Tenants`

CrestApps: we are [here](https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs#L58) and [here](https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/FeatureProfilesController.cs#L40)

hishamco: I'm wondering how it works before ;)

hishamco (approved):

src/OrchardCore.Modules/OrchardCore.Tenants/Manifest.cs
CrestApps: I am guessing because `OC.Settings` is enabled by default with all of the default recipes. will it work if you create a recipe without enabling `OC.Settings`, probably not. once `OC.Settings` is enabled once, it's always enabled.

hishamco: Absolutly right, its always enabled

https://github.com/OrchardCMS/OrchardCore/blob/21c897c26db6336059e93cbd7cdadd91c9b347f7/src/OrchardCore.Modules/OrchardCore.Settings/Manifest.cs#L10

So, I don't think this dependecy is necessary, unless @jtkech has a reason for that

MajdAlhayek: @hishamco `IsAlwaysEnabled` will only take effect *after* a the module is actually enabled. This means, if you create a site without enabling `OC.Settings` then it will never be enabled. So for assurance, the dependency to OC.Settings is still needed.

CrestApps: @hishamco IsAlwaysEnabled will only take effect after a the module is actually enabled. This means, if you create a site without enabling OC.Settings then it will never be enabled. So for assurance, the dependency to OC.Settings is still needed.

hishamco: Right, but it's enabled by default in almost if not all the recipes

hishamco: Right, but it's enabled by default in almost if not all the recipes

CrestApps: what if someone want to use their own recipe and did not enable OC.Settings?

hishamco: I never mind to accept this PR while I already approved ;)

Skrypt (approved):

CrestApps:
@jtkech the last change is great! But wondering why aren’t we using something like `IOptions<PagerOptions>` instead? Similar approach was done in PR #11569 which remove the need of ISiteService altogether. This could probably be beneficial in other modules as well.

jtkech:
I'm looking at #11569

jtkech:
@CrestApps 

Yes I added some comments in your PR, open to any suggestion but there are many scenario, see my comments in your PR.

For now, here it works as before and just fixes an issue I encountered.

jtkech:
Also, just to say that here it only uses `ISiteService` for the pager `PageSize`.

CrestApps:
@jtkech I am not against what you did, but I think it’s a workaround not a fix. I think a fix could be by adding `IOptions<PagerOptions>` which will provides easy way to override options without having to implement `ISiteService`. My suggestion  will involve a lot more work that you may want to avoid for now. Perhaps my suggestion could be a second PR at some point “unless you disagree with the approach”

jtkech:
Yes I understand.

I also thought about a default value in the `Pager` class itself.

As said there are many scenario, for example do we want all admin settings configurable? Not sure, maybe would need to be discussed more globally, or not I don't know, maybe better to consider each specific case, this is not because a pattern is good that we have to use it everywhere.

For now I think that managing `IOptions<PagerOptions>` here is too much, but yes I understand what you mean.

CrestApps:
Yea I agree with you. We don’t use a pattern just because. Just for future reference, the reason I suggested introducing `PagerOptions` is because in many places in OC, we inject `ISiteService` only to get pager info like is done [here](https://github.com/OrchardCMS/OrchardCore/blob/9d8e75f5caaa0540e6206228f098980350dad244/src/OrchardCore.Modules/OrchardCore.Contents/Controllers/AdminController.cs#L102) and many other places like it. Having `PagerOptions` will fix your issue and help in many other cases. But, the workaround you provided will defiantly work for the time being.

jtkech:
Okay good point ;)

Will think about it tomorrow

jtkech:
That said for the example related to `OC.Contents` module.

No problem as it depends on `OC.Settings`

    [assembly: Feature(
        Id = "OrchardCore.Contents",
        Name = "Contents",
        Description = "The contents module enables the edition and rendering of content items.",
        Dependencies = new[]
        {
            "OrchardCore.Settings",
            "OrchardCore.Liquid"
        },
        Category = "Content Management"
    )]

hishamco (approved):

CrestApps:
Yes I saw that. Not looking at the code now to find out the real reason, but probably they ran into the same problem you ran into which lead you to this PR :)

jtkech:
Yes, maybe also for info because OC.Contents is part of the CMS modules meta package, whereas other modules are more intended to be standalone modules that only "depend" on the OC infrastructure.

sebastienros:
Didn't read the comments, can the SiteSettings module be a dependency of Tenants?

jtkech:
Yes, that was my first way to do this

But I thought that it would be useful to not define this dependency for some scenario

But as you want, easy to do

PR #12174, opened 2022-08-10
Default Shell State Consistency

jtkech:
In a multi instances env with tenant settings coming from a shared source as the database, if it fails due to a bad connection or the absence of the needed table, the `Default` tenant is in an inconsistent state.

Not a big issue when we work with only one OC instance (just needs to be restarted), and this is what we recommend before having setup the `Default` tenant on which we rely e.g. to keep in sync tenants.

But while testing the Tenant Removal with a local K8S deployment allowing multiple OC replicas, when I forget to specify only one OC replica before a fresh setup, it is less easy to find the OC pods to restart.

So here the main point is when `_shellSettingsManager.LoadSettingsAsync()` fails from different places.

- When it fails from `ShellHost.InitializeAsync()`, here we still let the exception be thrown but we don't mark the `ShellHost` as `_initialized`, so that when the database connection/table are okay, on a new request we can create a setup context and render the setup form in place of a blank page.

- We also call `_shellHost.InitializeAsync()` In `ModularBackgroundService` if the `.ShellWarmup` option is true, here we use a try/catch block to log an error but without interrupting the background service, so that when the Default tenant is finally setup the background service will start its periodic loop.

- TODO: In a multi instances env, the `Default` setup form may be rendered by one instance, but the setup executed by another one. In that case the `Default` tenant of the first instance will always be `Uninitialized` because tenant syncing rely on the `Default` tenant being running.

    Here the solution, in `DistributedShellHostedService`, would be to periodically check directly from the shared tenants settings source if the `Default` tenant was setup by another instance.

    Maybe here or in a separate PR. **Updated**: Okay done here.

### DIFF
--- a/src/OrchardCore/OrchardCore/Shell/Distributed/DistributedShellHostedService.cs
+++ b/src/OrchardCore/OrchardCore/Shell/Distributed/DistributedShellHostedService.cs
@@ -22,9 +22,6 @@ namespace OrchardCore.Environment.Shell.Distributed
         private const string ReleaseIdKeySuffix = "_RELEASE_ID";
         private const string ReloadIdKeySuffix = "_RELOAD_ID";
 
-        // The syncing period in seconds of the default tenant while it is uninitialized.
-        private const int DefaultTenantSyncingPeriod = 25;
-
         private static readonly TimeSpan MinIdleTime = TimeSpan.FromSeconds(1);
         private static readonly TimeSpan MaxRetryTime = TimeSpan.FromMinutes(1);
         private static readonly TimeSpan MaxBusyTime = TimeSpan.FromSeconds(2);
@@ -67,6 +64,9 @@ namespace OrchardCore.Environment.Shell.Distributed
         /// </summary>
         protected override async Task ExecuteAsync(CancellationToken stoppingToken)
         {
+            // The syncing period in seconds of the default tenant while it is 'Uninitialized'.
+            const int DefaultTenantSyncingPeriod = 25;
+
             stoppingToken.Register(() =>
             {
                 _logger.LogInformation("'{ServiceName}' is stopping.", nameof(DistributedShellHostedService));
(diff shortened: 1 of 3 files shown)


src/OrchardCore/OrchardCore/Shell/Distributed/DistributedShellHostedService.cs
```diff
@@ -94,15 +94,19 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
-                    // If the default tenant is uninitialized, check its shared state periodically.
-                    if (defaultContext.Settings.State == TenantState.Uninitialized &&
-                        defaultTenantSyncingPeriod++ > DefaultTenantSyncingPeriod)
+                    // Manage the syncing period of the default tenant while it is 'Uninitialized'.
+                    defaultTenantSyncingPeriod = defaultContext.Settings.State == TenantState.Uninitialized
+                        ? defaultTenantSyncingPeriod++
+                        : 0;
+
+                    // Check periodically if the default tenant is no more 'Uninitialized'.
+                    if (defaultTenantSyncingPeriod++ > DefaultTenantSyncingPeriod)
```
hishamco, line 103: The naming is confusing while both are default, can we use `DefaultDefaultTenantSyncingPeriod` or `DefaultTenantSyncingPeriodDefault`

May be `DefaultSyncingPeriodForDefaultTenant`

jtkech: Yes, they should not have the same name, one is a period the other is a second counter.

hishamco: The PR is looks good except this tiny comment ;)

hishamco (approved):

PR #12188, opened 2022-08-14
Tenant names checking tweaks

jtkech:
@Piedone @hishamco 

I re-opened a PR on this branch just to show what I just did because I wanted to update it before forgetting the details around this, and also because I didn't see any update on the related PR #11435.

So here what I did

- Re-introduce `IsDefaultShell()` but make it case sensitive and then use it where it can be.

And what this branch was already doing as a reminder.

- Take into account that the shellHost dicos now are case insensitive.

- Fix a little validation issue in TenantValidator (which now also uses `IsDefaultShell()`).

All of the above being related to tenant names checking.

PR #12207, opened 2022-08-16
Add PagerOptions

CrestApps:
Add `PagerOptions` options.
Replace `ISiteService` use when possible by `IOptions<PagerOptions>` 
Removed `ISiteService` when it's not used.


src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -91,35 +92,26 @@ public async Task<IActionResult> Index(TenantIndexOptions options, PagerParamete
...
+                    Category = x["Category"],
+                    Description = x["Description"],
+                    Name = x.Name,
+                    ShellSettings = x,
+                    IsDefaultTenant = String.Equals(x.Name, ShellHelper.DefaultShellName, StringComparison.OrdinalIgnoreCase)
+                };
+
+                if (x.State == TenantState.Uninitialized && !String.IsNullOrEmpty(x["Secret"]))
                 {
-                    var entry = new ShellSettingsEntry
```
CrestApps, line 107: I am not sure what happened here. I’ll revert some of these changes

jtkech: No problem, you formatted it like this which is good

            var entries = allSettings.Select(x =>
            {
                var entry = new ShellSettingsEntry
                {
                    Category = x["Category"],

While it was formatted like this

            var entries = allSettings.Select(x =>
                {
                    var entry = new ShellSettingsEntry
                    {
                        Category = x["Category"],

jtkech: As a side note, not now but here it will be `IsDefaultTenant = x.Name == ShellHelper.DefaultShellName`.

I think I will merge soon #12188 as there are too many PRs related to the same files since a too long time.

So that we will have the same base on which we will be able to add new things, even if it creates some PR conflicts.

src/OrchardCore/OrchardCore.Navigation.Core/PagerOptions.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace OrchardCore.Navigation;
+
+public class PagerOptions
+{
+    [DefaultValue(10)]
+    public int PageSize { get; set; } = 10;
+
+    [DefaultValue(100)]
+    public int MaxPageSize { get; set; } = 100;
```
jtkech, line 11: Good for the default values

src/OrchardCore.Modules/OrchardCore.Navigation/Startup.cs
```diff
@@ -6,13 +7,21 @@ namespace OrchardCore.Navigation
...
+        }
+
         public override void ConfigureServices(IServiceCollection services)
         {
             services.AddNavigation();
 
             services.AddScoped<IShapeTableProvider, NavigationShapes>();
             services.AddScoped<IShapeTableProvider, PagerShapesTableProvider>();
             services.AddShapeAttributes<PagerShapes>();
+            services.Configure<PagerOptions>(options => _configuration.GetSection("OrchardCore:OrchardCore_Pager").Bind(options));
```
jtkech, line 24: Hmm, maybe `OrchardCore:OrchardCore_Navigation` to follow the section convention = module/feature name.

Then maybe the following if we need other options still related to the Navigation module.

    "OrchardCore_Navigation": {
        "PagerOptions": {
            "PageSize": 10,
            "MaxPageSize": 100,
            "MaxPagedCount": 100
        }
    },

If so, as I remember we may not need an additional `.GetSection("PagerOptions")` to the above `Bind()` as it will see the matching `PagerOptions` property, but needs to be tried.

CrestApps: @jtkech I made the requested changed

FYI the both of the following worked
```
_configuration.GetSection("OrchardCore:OrchardCore_Navigation").GetSection("PagerOptions").Bind(opt);
```

and 
```
_configuration.GetSection("OrchardCore:OrchardCore_Navigation:PagerOptions").Bind(opt);
```

but this did NOT work
```
_configuration.GetSection("OrchardCore:OrchardCore_Navigation:PagerOptions").Bind(opt);
```

jtkech: Okay cool

Don't see any difference in your 2 last examples

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
CrestApps: I fixed the conflict.

src/OrchardCore.Modules/OrchardCore.Users/Controllers/AccountController.cs
```diff
@@ -631,7 +631,6 @@ public async Task<IActionResult> RegisterExternalLogin(RegisterExternalLoginView
         [ValidateAntiForgeryToken]
         public async Task<IActionResult> LinkExternalLogin(LinkExternalLoginViewModel model, string returnUrl = null)
         {
-            var settings = (await _siteService.GetSiteSettingsAsync()).As<RegistrationSettings>();
```
jtkech, line 634: Oops, it was not used!

CrestApps: Yup! Couple places like that so I removed them since I was cleaning up the use `ISiteService`

src/OrchardCore.Modules/OrchardCore.Navigation/Startup.cs
```diff
@@ -6,13 +7,21 @@ namespace OrchardCore.Navigation
...
+        }
+
         public override void ConfigureServices(IServiceCollection services)
         {
             services.AddNavigation();
 
             services.AddScoped<IShapeTableProvider, NavigationShapes>();
             services.AddScoped<IShapeTableProvider, PagerShapesTableProvider>();
             services.AddShapeAttributes<PagerShapes>();
+            services.Configure<PagerOptions>(options => _configuration.GetSection("OrchardCore:OrchardCore_Navigation:PagerOptions").Bind(options));
```
jtkech, line 24: Maybe you meant that this is the following that was not working, right?

    _configuration.GetSection("OrchardCore:OrchardCore_Navigation").Bind(opt);

Maybe this works

    _configuration.GetSection("OrchardCore:OrchardCore_Navigation").Get<PagerOptions>();

But there is an overload of `services.Configure<PagerOptions>()` that takes an `IConfiguration` param without having to do a bind.

    services.Configure<PagerOptions>(
        _configuration.GetSection("OrchardCore:OrchardCore_Navigation").GetSection("PagerOptions"));

CrestApps: Right! I'll get that changed. much cleaner 
```
services.Configure<PagerOptions>("OrchardCore:OrchardCore_Navigation:PagerOptions", _configuration);
```

CrestApps: That did not work.

jtkech: Me too I always have problems to remember all the overloads

No problem will test it tomorrow if you don't fix it in the meantime

Also I prefer to not have a too long key with here 2 `:` separator that it will always have to parse, better I think to provide to it the right section, or as a compromise the last parent section and then use `PagerOptions` as a key.

CrestApps: humm, would you rather have it like this

```
_configuration.GetSection("OrchardCore:OrchardCore_Navigation").GetSection("PagerOptions").Bind(options)
```

jtkech: I'm testing them

Just tried the following that works

            services.Configure<PagerOptions>(_configuration
                .GetSection("OrchardCore:OrchardCore_Navigation")
                .GetSection("PagerOptions"));

Also I think that you don't need the `[DefaultValue(10)]` in `PageOptions.cs`.

CrestApps: I guess you are right we don’t need the defaultValue attribute since we are not serializing it into json. 

maybe this would even be better? 
```
  services.Configure<PagerOptions>(_configuration
            .GetSection("OrchardCore")
            .GetSection(“Navigation”)
            .GetSection("PagerOptions"));
```

jtkech: Good also to use a variable for the config so that the delegate closure don't have to reference the startup class.


            var configuration = _configuration
                .GetSection("OrchardCore")
                .GetSection("OrchardCore_Navigation")
                .GetSection("PagerOptions");

            services.Configure<PagerOptions>(configuration);

CrestApps: Okay. I would probably set it to the code below.

I am out at the moment. I’ll try to do it in the next hour or tomorrow. Unless you want to merge it now, feel free to make this last update and merge it. 
```
var pagerOptions = _configuration
            .GetSection("OrchardCore")
            .GetSection("OrchardCore_Navigation")
            .GetSection("PagerOptions");

        services.Configure<PagerOptions>(pagerOptions);
```

jtkech: Or this one, useful if we will have other options, but the above one is okay

            var configuration = _configuration
                .GetSection("OrchardCore")
                .GetSection("OrchardCore_Navigation");

            services.Configure<PagerOptions>(configuration.GetSection("PagerOptions"));

            // As an example for future options.
            services.Configure<PagerOptions>(configuration.GetSection("OtherOptions"));


Note: Yes the overload taking a string key and a config doesn't work because the key is intended to point to an individual field, not a nested object.

jtkech: No problem will see tomorrow, I let you choose one of my 2 examples.

E.g. what you just suggested, but don't use `var pagerOptions` as it is not a `PagerOptions` but a `IConfigurationSection` which inherit of `IConfiguration`.

So e.g. `var configuration` or maybe `configurationSection` or just `section` as you want.

CrestApps: @jtkech I ended up doing this
```
            var navigationConfig = _configuration
                .GetSection("OrchardCore")
                .GetSection("OrchardCore_Navigation");

            services.Configure<PagerOptions>(navigationConfig.GetSection("PagerOptions"));
```
It bugged me until I added the `navigation` in the variable. But everything should be updated now. I also tested the above and it works as expected.

jtkech: Okay will look at it toonight but looks good!

jtkech:
@CrestApps 

So I committed some last tweaks

- Re-use `IShellConfiguration` for `PagerOptions` and also `ResourceOptions` because both related projects still reference (maybe transitively) OC.Abstractions, so better to use `IShellConfiguration`.

- Renamed to `PagerOptionsConfiguration` and `ResourceOptionsConfiguration` to follow the naming we use most of the time (not always) for `IPostConfigureOptions<>`.

- Commented out the `OrchardCore_Navigation` example in `appsettings.json`.

I'm waiting for your review ;)

Skrypt:
Looks good to me by doing a quick vertical check.

jtkech:
Okay cool then ;)

PR #12233, opened 2022-08-20
Use underscore (_) instead of (this.) when creating new fields in VS

### LINKED ISSUE #12232: Change project editorconfig to use underscore (_) instead of (this.) when creating new fields
### Describe the bug

When injecting new service in a class constructor, VisualStudio suggest the use `this.` syntax to initialize the variable as you can see in the following screenshot.

image

When using the above suggestion, the output will look like this

```
public class OrchardCoreExample
{
    private readonly IContentManager contentManager;

    public OrchardCoreExample(IContentManager contentManager)
    {
        this.contentManager = contentManager;
    }
}
```
The above is not the style use in the OrchardCore solutions. I suggest we change the editor to the following instead

image

So the new suggested code will look like this by default

```
public class OrchardCoreExample
{

    public OrchardCoreExample(IContentManager contentManager)
    {
        _contentManager = contentManager;
    }
}
```

MikeAlhayek:
Fix #12232

image


.editorconfig
```diff
@@ -51,3 +51,12 @@ dotnet_style_explicit_tuple_names                          = true : suggestion
...
 insert_final_newline = true
+
+# Suggest to use "_" instead of "this." when creating new private fields
+dotnet_naming_style.camel_case_leading_underscore.capitalization = camel_case
+dotnet_naming_style.camel_case_leading_underscore.required_prefix = _
+dotnet_naming_symbols.private_fields.applicable_kinds = field
+dotnet_naming_symbols.private_fields.applicable_accessibilities = private
+dotnet_naming_rule.private_fields_should_be_camel_case_leading_underscore.severity = suggestion
+dotnet_naming_rule.private_fields_should_be_camel_case_leading_underscore.symbols = private_fields
+dotnet_naming_rule.private_fields_should_be_camel_case_leading_underscore.style = camel_case_leading_underscore
```
Skrypt, line 62: add a new line all the time at the end of your files for Git.

MikeAlhayek: done

Skrypt (approved):

PR #12235, opened 2022-08-21
Added dotnet format

ns8482e:
Added dotnet format verification (`--verify-no-changes`) in PR pipeline.

Developer can fix the the reported errors by running `dotnet format`


agriffard:
Is there a way to apply the recommendations?

MikeAlhayek:
Build failed

Skrypt:
Probably because it is chasing its own tail. It adds a CI `dotnet format` check while there are issues needed to be fixed which this PR fixes. The issues will need to be fixed before adding `dotnet format`  check.

ns8482e:
> Build failed

it’s failed because due to line ending in Linux.  Git config takes care of line ending based on OS. However editorconfig has defined it crlf 

May be just validate it in windows only

Skrypt:
The EOL is for Git I believe. Could have worked with a warn-only message. Need to see if we could force squash and merge still.

Piedone (changes requested):
We should rather guard against such changes as part of the .NET build, treating warnings as errors, not with a separate CI step. Also see https://github.com/OrchardCMS/OrchardCore/issues/7950.

PR #12240, opened 2022-08-21
Apply MaxPageSize for PageSize if PageSize  > MaxPageSize

### LINKED ISSUE #12239: meaning between  "MaxPageSize" and   "MaxPagedCount"
### Expected behavior
under namespace OrchardCore.Contents.Controllers
{
    public class AdminController : Controller

line 243:
  if (_pagerOptions.MaxPagedCount > 0 && pager.PageSize > _pagerOptions.MaxPagedCount)
            {
                pager.PageSize = _pagerOptions.MaxPagedCount;
            }
It seems that MaxPagedCount  should be MaxPageSize

infofromca:
fix #12239

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Controllers/AdminController.cs
@@ -240,6 +240,10 @@ namespace OrchardCore.Contents.Controllers
 
             var routeData = new RouteData(options.RouteValues);
 
+            if (_pagerOptions.MaxPageSize > 0 && pager.PageSize > _pagerOptions.MaxPageSize)
+            {
+                pager.PageSize = _pagerOptions.MaxPageSize;
+            }
             if (_pagerOptions.MaxPagedCount > 0 && pager.PageSize > _pagerOptions.MaxPagedCount)
             {
                 pager.PageSize = _pagerOptions.MaxPagedCount;


hishamco:
@MikeAlhayek I think you worked on this last few days

src/OrchardCore.Modules/OrchardCore.Contents/Controllers/AdminController.cs
```diff
@@ -240,6 +240,10 @@ public class AdminController : Controller
 
             var routeData = new RouteData(options.RouteValues);
 
+            if (_pagerOptions.MaxPageSize > 0 && pager.PageSize > _pagerOptions.MaxPageSize)
+            {
+                pager.PageSize = _pagerOptions.MaxPageSize;
```
ns8482e, line 245: what's difference between MaxPageSize and MaxPagedCout? both seems to set PageSize  line 245/249

jtkech: @ns8482e  I think MaxPageSize is the max of items per pager page, MaxPagedCount is the max of total items of all pages. So if the max per page is > total max, meaning that in one page you can have more than the allowed total max, it makes sense to limit MaxPageSize to MaxPagedCount.

jtkech: Hmm, maybe there are many other places where we don't take MaxPagedCount into account.

Maybe create a `PagerOptions.GetPageSize()` method that would check the MaxPagedCount limit.

jtkech: For now PagerOptions MaxPagedCount is only used in this admin controller and PagerOptions MaxPageSize never used, so makes sense to add it here.

jtkech (approved):

src/OrchardCore.Modules/OrchardCore.Contents/Controllers/AdminController.cs
MikeAlhayek: @jtkech I actually don't think there is a need for `MaxPagedCount`. Reason I am saying that is because `MaxPageSize` isn't editable in the UI. so `MaxPageSize` should be checked.

So the condition should be something like

```
if (_pagerOptions.MaxPageSize > 0 && pager.PageSize > _pagerOptions.MaxPageSize )
{
        pager.PageSize = _pagerOptions.MaxPageSize;
}
```
I made another PR #12257 that would enforce using that condition throughout the app.

Can't we just remove `MaxPagedCount` altogether? `MaxPageSize` would only be populated from the `appsettings` file and the `PageSize` would be the admin settings.

infofromca: there is another situation:
if MaxPagedCount < MaxPageSize, my code defend this situation

infofromca: move all those logic to PagerOptions.GetPageSize() or property is also good

MikeAlhayek: I would not create `GetPageSize()` unless you want to change everywhere PageSize is currently use. I would just put that logic in the PageSize getter

src/OrchardCore.Modules/OrchardCore.Contents/Controllers/AdminController.cs
```diff
@@ -240,6 +240,10 @@ public class AdminController : Controller
 
             var routeData = new RouteData(options.RouteValues);
 
+            if (_pagerOptions.MaxPageSize > 0 && pager.PageSize > _pagerOptions.MaxPageSize)
```
MikeAlhayek, line 243: I suggest you remove this check from the controller. Instead, add it to the getter of PagerOption.PageSize so that this check is used everywhere in the app.

PR #12253, opened 2022-08-23
Fix SmtpService to validate email address

### LINKED ISSUE #12252: Unexpected expection thrown when testing the email settings with an invalid email
### Describe the bug
When testing the email settings with an invalid email format, the SmtpService display the following error

```
An error occurred while sending an email: 'Unexpected 'A' token at offset 5'
```

### To Reproduce
1. Enable `OC.Email` feature
2. Click on Configuration >> Settings >> Email
3. Make sure the email settings is configured. then click on "Test settings" button
4. Provide invalid formatted email address in the "sender" field.
5. Provide "To" email, "Subject" and "Body" and click send.

MikeAlhayek:
Fix #12252

### DIFF
--- a/src/OrchardCore/OrchardCore.Email.Core/Services/SmtpService.cs
+++ b/src/OrchardCore/OrchardCore.Email.Core/Services/SmtpService.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Net.Security;
 using System.Security.Cryptography.X509Certificates;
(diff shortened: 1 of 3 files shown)


src/OrchardCore/OrchardCore.Email.Core/Services/SmtpService.cs
```diff
@@ -71,7 +72,14 @@ public async Task<SmtpResult> SendAsync(MailMessage message)
                     message.From = senderAddress;
                 }
 
-                var mimeMessage = FromMailMessage(message);
+                var errors = new List<LocalizedString>();
+
+                var mimeMessage = FromMailMessage(message, errors);
```
ns8482e, line 77: Not sure if errors should be part of `FromMailMessage` method? may be just create IsValid method separate from FromMailMessage method

ns8482e (changes requested):

src/OrchardCore/OrchardCore.Email.Core/Services/SmtpService.cs
MikeAlhayek: @ns8482e this is how the class was designed. All we did here is create more checks to ensure we report proper error instead of throwing exception

sebastienros:
"Invalid {0} email." -> "Invalid email address: '{0}'" this way only one string is translated, and users can see what actual address is wrong.

MikeAlhayek:
@sebastienros fixed

PR #12263, opened 2022-08-25
Add settings to form widgets

### LINKED ISSUE #12256: Add settings to the form part for creating label and validation elements

When creating a form using the Form input widgets like (`Input`, `Select`, `Text Area`) there is no easy way to add a label and validation elements to the input. The current approach is to add label widget and then validation widget which is not ideal all the time. Here are some of the issues with the current process
1. Creates too much html code because each widget has its own wrappers.
2. When creating two input with 50% width and two label with 50% width, on a small device the labels are stacked over each other (I.e., label, label, input, input) instead of stacking label, input, label input. 
image

4. It makes editing experience terrible specially if you want to drag/drop one field since you’ll have to drag all 3 related widgets. 

We can make this process much simpler by adding settings to the `Input`, `Select`, `Text Area` widgets with the following properties
 - `LabelOption` an `enum` value with the following values (`None`, `Standard`, `ScreenReaders`.) By default, `None` is selected to keep it backward compatible.
 - `ValidationOptions` an `enum` value with the following values (`None`, `Standard`). By default, `None` is selected to keep it backward compatible.

Now, if the user selects option other than `None` in the settings, we’ll create the label in the same widget block. The same applies to `ValidationOptions`. This will make things like dragging/dropping widgets during edit and controlling the size much easier.  At the same time, we output less HTML code. And if we add the validation rules, we’ll have more validation logic which makes the widgets more useful. 

Lastly, the `Input Widget` should not have "Radio", I don't think that is useful at all. However, the `Select Input` should have `Editor` settings with `Dropdown`, `Radio`, or `Checkbox` style so that the user can render the option different ways.

### LINKED ISSUE #12267: Form SelectInput does not show the options editor when added to the DOM
### Describe the bug
When trying to add "Select Input" while creating a form, the option editor isn't loaded. Also, the edit button, does not open the modal that would allow you to populate the options using json. If the content is saved, and the page is loaded again then modal and the options editor works just fine.

### To Reproduce
1. Enable `OC.Flows` and `OC.Forms` module
2. Create content type `Test` with `FlowPart` and make it creatable
3. Create a new content item of type `Test`, then add "Select Input" widget

### Expected behavior
I should be able to click on the Edit button and the modal should show up. I also should be able to add options when the Select Input is added to the DOM

### Screenshots
Before the content-item is saved
image

After the content-item is saved
image

MikeAlhayek:
Fix #12256
Fix #12267 

Select Input
image

image

image

On input widget, when the user select `button`, `hidden` or `reset` we hide irrelevant fields to clean up the form.
image

image


Skrypt:
Form type : GET, POST.

ns8482e:
@MikeAlhayek  
comments from @sebastienros   https://github.com/OrchardCMS/OrchardCore/issues/4250#issuecomment-535654673

MikeAlhayek:
@ns8482e yes. He is referring to this PR. We agreed during last triage to do this.

src/OrchardCore.Modules/OrchardCore.Forms/Assets/js/formsWidgetEditor.js
```diff
@@ -0,0 +1,55 @@
+formsWidgetEditor = function () {
```
ns8482e, line 1: it's called from event why not just keep it inside event instead of making global function?

src/OrchardCore.Modules/OrchardCore.Forms/Models/InputPart.cs
```diff
@@ -1,11 +1,20 @@
+using System.ComponentModel;
 using OrchardCore.ContentManagement;
 
 namespace OrchardCore.Forms.Models
 {
     public class InputPart : ContentPart
     {
         public string Type { get; set; }
+
```
ns8482e, line 9: remove extra line

src/OrchardCore.Modules/OrchardCore.Forms/Models/InputPart.cs
```diff
@@ -1,11 +1,20 @@
...
         public string DefaultValue { get; set; }
+
         public string Placeholder { get; set; }
+
+        public LabelOptions LabelOption { get; set; }
+
+        public string Label { get; set; }
+
+        public ValidationOptions ValidationOption { get; set; }
     }
```
ns8482e, line 19: remove extra lines

src/OrchardCore.Modules/OrchardCore.Forms/Models/TextAreaPart.cs
```diff
@@ -5,6 +5,13 @@ namespace OrchardCore.Forms.Models
...
         public string DefaultValue { get; set; }
+
         public string Placeholder { get; set; }
+
+        public LabelOptions LabelOption { get; set; }
+
+        public string Label { get; set; }
+
+        public ValidationOptions ValidationOption { get; set; }
     }
```
ns8482e, line 16: remove extra lines

src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/InputPart.Fields.Edit.cshtml
```diff
@@ -1,43 +1,81 @@
...
+            <option value="@ValidationOptions.Standard">@T["Standard"]</option>
+        </select>
+        <span class="hint">@T["The type of validation to create for this field."]</span>
+    </div>
 </div>
+
+<script asp-src="~/OrchardCore.Forms/Scripts/formsWidgetEditor.min.js" debug-src="~/OrchardCore.Forms/Scripts/formsWidgetEditor.js" asp-name="formsWidgetEditor" at="Foot"></script>
+
+<script at="Foot" depends-on="formsWidgetEditor">
+    var wrapper = document.getElementById('@Html.IdFor(m => m)');
```
ns8482e, line 75: call from onloaded event

src/OrchardCore.Modules/OrchardCore.Forms/Models/InputPart.cs
```diff
@@ -1,11 +1,20 @@
...
 {
     public class InputPart : ContentPart
     {
         public string Type { get; set; }
+
         public string DefaultValue { get; set; }
+
         public string Placeholder { get; set; }
+
+        public LabelOptions LabelOption { get; set; }
```
ns8482e, line 14: Similar to `FormElementPart` and `FormInputElementPart` let's have this moved new part `FormElementLabelPart` and provide views for it.

Let's keep `InputPart` and `SelectPart` unchanged and  for label add `FormElementLabelPart` to content type Input and Select using migration

src/OrchardCore.Modules/OrchardCore.Forms/Models/SelectPart.cs
```diff
@@ -5,7 +5,16 @@ namespace OrchardCore.Forms.Models
...
+
         public string DefaultValue { get; set; }
+
+        public SelectEditorOption Editor { get; set; }
+
+        public LabelOptions LabelOption { get; set; }
+
+        public string Label { get; set; }
+
+        public ValidationOptions ValidationOption { get; set; }
```
ns8482e, line 17: same as above 
Similar to `FormElementPart` and `FormInputElementPart` let's have this moved new part `FormElementLabelPart` and provide views for it.

Let's keep `InputPart` and `SelectPart` unchanged and  for label add `FormElementLabelPart` to content type Input and Select using migration

src/OrchardCore.Modules/OrchardCore.Forms/wwwroot/Scripts/formsWidgetEditor.js
```diff
@@ -0,0 +1,68 @@
...
+  return {
+    initilizeFieldType: initilizeFieldType,
+    initilizeLabelOptions: initilizeLabelOptions
+  };
+}();
+
+document.addEventListener('DOMContentLoaded', function () {
+  formsWidgetEditor.initilizeFieldType(document);
+  formsWidgetEditor.initilizeLabelOptions(document);
+});
```
ns8482e, line 68: EOL

src/OrchardCore.Modules/OrchardCore.Forms/Models/TextAreaPart.cs
MikeAlhayek: In other PRs I was asked to add lines between property name. It seems to be the new style

src/OrchardCore.Modules/OrchardCore.Forms/Assets/js/formsWidgetEditor.js
MikeAlhayek: Because it’s being used in multiple places like TextAreaPart, InputPart and SelectPart

src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/InputPart.Fields.Edit.cshtml
MikeAlhayek: Is it called using the loaded event, check out the this. This line will only be called when the widget is added to the DOM using flow part. If we don’t do this, the new widget will not get initialized correctly.

src/OrchardCore.Modules/OrchardCore.Forms/Models/TextAreaPart.cs
ns8482e: I in this PR one place you have no lines and other place you add extra lines.  If you believe extra lines should be there then add them to other classes. or remove from here

src/OrchardCore.Modules/OrchardCore.Forms/Assets/js/formsWidgetEditor.js
MikeAlhayek: I’ll revisit this once I create part for the label options and another for the validation options.

sebastienros (approved):

src/OrchardCore.Modules/OrchardCore.Forms/Views/Widget-Form.cshtml
```diff
@@ -0,0 +1,7 @@
+@{
+    Model.Classes.Add("mb-3");
```
sebastienros, line 2: This is bootstrap, on the front-end, for a default template. Is that acceptable?

Or another option is to use "form-widget-input" as a class instead of "mb-3"

sebastienros: And adding "widget-body" as a class would make it less a breaking change.

MikeAlhayek: as agreed. these files were removed from this PR

hishamco: I disagree to add bootstrap specific CSS class

MikeAlhayek: @hishamco please your comment in PR #12309 since the referenced code here is no longer part of this PR

src/OrchardCore.Modules/OrchardCore.Forms/Models/SelectPart.cs
```diff
@@ -6,6 +6,7 @@ public class SelectPart : ContentPart
     {
         public SelectOption[] Options { get; set; }
         public string DefaultValue { get; set; }
+        public SelectEditorOption Editor { get; set; }
```
ns8482e, line 9: Set Default value to Dropdown, so the exiting items' wont break with NRE

src/OrchardCore.Modules/OrchardCore.Forms/Models/FormElementLabelPart.cs
```diff
@@ -0,0 +1,15 @@
...
+public class FormElementLabelPart : ContentPart
+{
+    public LabelOptions Option { get; set; }
+
+    public string Label { get; set; }
+}
+
+public class FormElementValidationPart : ContentPart
+{
+    public ValidationOptions Option { get; set; }
```
ns8482e, line 14: Set Default to None so exiting content item won't break with NRE

ns8482e:
New Part `FormElementLabelPart` and `FormElementValidationPart` to be attached either via Driver/Migration to existing content item

src/OrchardCore.Modules/OrchardCore.Forms/Models/FormElementLabelPart.cs
MikeAlhayek: @ns8482e `ValidationOptions` is an enum with the default value of `None` so it's always `None` unless a user changes it. so it wont break anything

ns8482e (approved):

src/OrchardCore.Modules/OrchardCore.Forms/Models/SelectPart.cs
MikeAlhayek: Same here `SelectEditorOption` is an enum with `None` is the default value. so it won't break anything

src/OrchardCore.Modules/OrchardCore.Forms/Views/Items/SelectPart.Fields.Edit.cshtml
```diff
@@ -99,32 +100,102 @@
...
+    var wrapper = document.getElementById('@Html.IdFor(m => m)');
+    if (wrapper != null) {
+        // at this point, we know this wrapper was added after page load, let's initilize it
+        initilizeFieldType(wrapper);
     }
-})
+    document.addEventListener('DOMContentLoaded', function() {
+        // at this point, let's initilize any existing SelectPart after the page is loaded
+        initilizeFieldType(document);
+    });
```
ns8482e, line 200: can me moved to separate script tag with `asp-name` to include only once in DOM

MikeAlhayek: @ns8482e done

src/OrchardCore.Modules/OrchardCore.Forms/Models/SelectPart.cs
MikeAlhayek: done.

ns8482e (approved):

PR #12305, opened 2022-08-31
Font awesome 6.2.0

agriffard:
https://www.npmjs.com/package/@fortawesome/fontawesome-free/v/6.2.0
https://github.com/FortAwesome/Font-Awesome/releases/tag/6.2.0
https://fontawesome.com/docs/changelog/


Skrypt:
The class names have changed. We should migrate all these class to new ones as documented here: 

https://fontawesome.com/docs/web/add-icons/how-to#classic-family

PR #12309, opened 2022-09-01
Adding default form templates for TheTheme

### LINKED ISSUE #12296: Render form widgets in a custom template instead of using Widget.cshtml Widget
Currently, all of the form widgets are rendered using `Widget.cshtml` which renders too much html without any added value. Also, by default the create form does not look well formatted by default because of this.

I think we should create default `Widget-Form`, `Widget-Input`, `Widget-Label`, Widget-Select`, `Widget-Validation` and `Widget-ValidationSummary`

Really all these templates should need only this

```
<div class="@String.Join(" ", Model.Classes.ToArray())">
    @await DisplayAsync(Model.Content)
</div>
```
Only the `Widget-Select`, `Widget-TextArea` and `Widget-Input` should have this

```
@{
    Model.Classes.Add("mb-3");
}
<div class="@String.Join(" ", Model.Classes.ToArray())">
    @await DisplayAsync(Model.Content)
</div>
```

MikeAlhayek:
Fix #12296

### DIFF
--- a/src/OrchardCore.Themes/TheTheme/Views/Widget-Form.cshtml
+++ b/src/OrchardCore.Themes/TheTheme/Views/Widget-Form.cshtml
@@ -1,0 +1,7 @@
+@{
+    Model.Classes.Add("widget-body mb-3");
+}
+
+<div class="@String.Join(" ", Model.Classes.ToArray())">
+    @await DisplayAsync(Model.Content)
+</div>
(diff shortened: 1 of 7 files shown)


src/OrchardCore.Themes/TheTheme/Views/Widget-Form.cshtml
```diff
@@ -0,0 +1,7 @@
+@{
+    Model.Classes.Add("widget-body mb-3");
```
hishamco, line 2: Why we need to add them into the model?

MikeAlhayek: The mb-3 is for look, and the widget-body is just to reduce the chance of breaking existing code. Some users may be using widget-body to override style. Widget-body is a class that is added when Widget.cshtml is used.

Skrypt (approved):

PR #12312, opened 2022-09-02
Ignore any explicit enable/disable 'EnabledByDependencyOnly' features

MikeAlhayek:
@jtkech I think this change is needed to prevent someone from explicitly trying to enable or disable a feature that is enabled by dependency only

### DIFF
--- a/src/OrchardCore/OrchardCore/Shell/ShellDescriptorFeaturesManager.cs
+++ b/src/OrchardCore/OrchardCore/Shell/ShellDescriptorFeaturesManager.cs
@@ -45,16 +45,16 @@ namespace OrchardCore.Environment.Shell
                 .ToHashSet();
 
             var installedFeatureIds = enabledFeatureIds
-                .Concat(shellDescriptor.Installed.Select(sf => sf.Id))
+                .Concat(shellDescriptor.Installed.Select(shellFeature => shellFeature.Id))
                 .ToHashSet();
 
-            var alwaysEnabledIds = _alwaysEnabledFeatures.Select(sf => sf.Id).ToArray();
+            var alwaysEnabledIds = _alwaysEnabledFeatures.Select(shellFeature => shellFeature.Id).ToArray();
 
             var byDependencyOnlyFeaturesToDisable = enabledFeatures
                 .Where(feature => feature.EnabledByDependencyOnly);
 
             var allFeaturesToDisable = featuresToDisable
-                .Where(feature => !feature.EnabledByDependencyOnly && !feature.IsAlwaysEnabled && !alwaysEnabledIds.Contains(feature.Id))
+                .Where(feature => !feature.EnabledByDependencyOnly && !alwaysEnabledIds.Contains(feature.Id))
                 .SelectMany(feature => GetFeaturesToDisable(feature, enabledFeatureIds, force))
                 // Always attempt to disable 'EnabledByDependencyOnly' features
                 // to ensure we auto disable any feature that is no longer needed.
--- a/src/OrchardCore/OrchardCore/Shell/ShellDescriptorFeaturesManager.cs
+++ b/src/OrchardCore/OrchardCore/Shell/ShellDescriptorFeaturesManager.cs
@@ -37,11 +37,11 @@ namespace OrchardCore.Environment.Shell
             var featureEventHandlers = ShellScope.Services.GetServices<IFeatureEventHandler>();
 
             var enabledFeatures = _extensionManager.GetFeatures()
-                .Where(f => shellDescriptor.Features.Any(sf => sf.Id == f.Id))
-                .ToList();
+                .Where(feature => shellDescriptor.Features.Any(shellFeature => shellFeature.Id == feature.Id))
+                .ToArray();
 
             var enabledFeatureIds = enabledFeatures
-                .Select(f => f.Id)
+                .Select(feature => feature.Id)
                 .ToHashSet();
 
             var installedFeatureIds = enabledFeatureIds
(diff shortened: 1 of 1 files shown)


src/OrchardCore/OrchardCore/Shell/ShellDescriptorFeaturesManager.cs
```diff
@@ -38,7 +38,7 @@ public class ShellDescriptorFeaturesManager : IShellDescriptorFeaturesManager
 
             var enabledFeatures = _extensionManager.GetFeatures()
                 .Where(f => shellDescriptor.Features.Any(sf => sf.Id == f.Id))
-                .ToArray();
+                .ToList();
```
jtkech, line 41: Keep ToArray() if possible

MikeAlhayek: @jtkech sure! done.

src/OrchardCore/OrchardCore/Shell/ShellDescriptorFeaturesManager.cs
```diff
@@ -54,7 +54,7 @@ public class ShellDescriptorFeaturesManager : IShellDescriptorFeaturesManager
                 .Where(f => f.EnabledByDependencyOnly);
 
             var allFeaturesToDisable = featuresToDisable
-                .Where(f => !alwaysEnabledIds.Contains(f.Id))
+                .Where(f => !f.EnabledByDependencyOnly && !alwaysEnabledIds.Contains(f.Id))
```
jtkech, line 57: Yes I thought about this but let the responsability at the higher level callers, as it is for `IsAlwaysEnabled `, otherwise we would need to do the following, remember we have 2 `AlwaysEnabled ` one at the ShellFeature level and one at the IFeatureInfo level.

    var allFeaturesToDisable = featuresToDisable
        .Where(f => !f.EnabledByDependencyOnly && !f.IsAlwaysEnabled && !alwaysEnabledIds.Contains(f.Id))

jtkech: Just to say that it depends at which level we are doing things, for example currently `IsAlwaysEnabled` is managed by using `GetAlwaysEnabledFeaturesAsync()` at the `ShellFeaturesManager` level, and which is called by the `OC.Features` AdminController.

MikeAlhayek: @jtkech 
IMO, [ShellDescriptorFeaturesManager.cs](https://github.com/OrchardCMS/OrchardCore/pull/12312/files/8c1d90eb9bf9efc1a21241e2691ab5a8b68160ac#diff-5de47caa58f6b04d151aaa3dc5b6485db7b91277285f49d855545036c58e0db8) should be responsibe of assuring no one makes the mistake of trying to enable/disabled what isn't allowed. I agree with your comment about about adding `!f.IsAlwaysEnabled` to the mix. Would you like for me to add it to this PR?

jtkech: let's think about it a little bit.

Then at this same level here why not checking `DefaultTenantOnly` directly (not through the related tenant validator still called at an higher level) if we are not in the Default tenant.

MikeAlhayek: @jtkech I think you missed my updated last comment which suggest that we do that here too. This is the same reason I mentioned we call the validators in this class in the other PR

jtkech: Okay cool then.

Yes since we managed EnabledByDependencyOnly at this level, we now introduce step by step all other properties ;) I already thought about managing these properties here when they were introduced step by step but it may be more complex than how we want to do it for now.

For example if there is an `IsAlwaysEnabled` that has a **dependency** on a non `IsAlwaysEnabled`, we can disable the first by disabling the 2nd dependency. Here idem for `DefaultTenantOnly`.

Also for a `DefaultTenantOnly` that has a a non `DefaultTenantOnly` **dependent**, we can enable the 1st by enabling the 2nd dependent (so that has a dependency on the 1st).

So all dependencies of an `IsAlwaysEnabled` would have to be considered as `IsAlwaysEnabled` too. And all dependencies and also all dependents of a `DefaultTenantOnly` would have to be considered as `DefaultTenantOnly` too.

That's why I kept these properties managed at the level they were intended to be used, even if defined in the manifest (maybe the confusion) but only as driven data to be used at an higher level.

MikeAlhayek: @jtkech maybe we should consider moving all that logic in here and a single point for change. Anyway, what we have in this PR is simple enough to to be added here. But defiantly we should think about consolidating and refactoring these services. Let me know if you want anything changes for this PR

jtkech: But it would be more complex for `DefaultTenantOnly` and `IsAlwaysEnabled` because we would need to take into account the dependency graph, which is already done by nature for `EnabledByDependencyOnly`.

**So my bad as I started the discussion but for now I suggest to not managed here `DefaultTenantOnly` and `IsAlwaysEnabled`, and only keep in this PR what is related to `EnabledByDependencyOnly`, which is simpler and make more sense based on what is in the current dev branch.**

Note: I don't say that `DefaultTenantOnly` and `IsAlwaysEnabled` are currently not well managed if we still consider them more focused on the admin features management where the responsability to not disable a feature is at an higher level. I was just saying that if we manage them here, for consistency with other low level feature states we would need to do what I described above.

MikeAlhayek: @jtkech I updated the change per your request.

jtkech: Okay cool, will look at it toonight.

PR #12314, opened 2022-09-02
Fix model error binding for reset password

### LINKED ISSUE #12313: Reset password form - password errors in validation summary but not validation message
### Describe the bug
OC 1.4

On the reset password form if you submit a password that does not meet the requirements you'll see the error in the validation summary at the top of the page but not the validation message below the new password input.

### To Reproduce
1. Create a new site using OC 1.4 and the blog recipe
2. Enable the Email feature and configure appropriately at /Admin/Settings/email
3. Enable the ResetPassword feature and check "allow the users to reset their password" at /Admin/Settings/userResetPassword
4. Begin the reset password flow at /forgotpassword and enter your email address
5. Click on the link in your email to go to /resetpassword

### Expected behavior
Typically messages are displayed in the validation summary and validation message as shown below
image

But if the password doesn't meet your requirements there is only a validation summary, not a validation message
image

Looking at the "change password" page for reference it does show both. (Although this case is interesting since there are multiple errors but the validation message only displays the first; I'd consider that a separate question.)
image

rjpowers10:
Fixes #12313

Now you get the validation message under the new password input
image

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Users/Controllers/ResetPasswordController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Controllers/ResetPasswordController.cs
@@ -131,7 +131,7 @@ namespace OrchardCore.Users.Controllers
 
             if (TryValidateModel(model) && ModelState.IsValid)
             {
-                if (await _userService.ResetPasswordAsync(model.Email, Encoding.UTF8.GetString(Convert.FromBase64String(model.ResetToken)), model.NewPassword, (key, message) => ModelState.AddModelError(key, message)))
+                if (await _userService.ResetPasswordAsync(model.Email, Encoding.UTF8.GetString(Convert.FromBase64String(model.ResetToken)), model.NewPassword, (key, message) => ModelState.AddModelError(key == "Password" ? nameof(ResetPasswordViewModel.NewPassword) : key, message)))
                 {
                     return RedirectToLocal(Url.Action("ResetPasswordConfirmation", "ResetPassword"));
                 }


src/OrchardCore.Modules/OrchardCore.Users/Controllers/ResetPasswordController.cs
```diff
@@ -131,7 +131,7 @@ public async Task<IActionResult> ResetPassword(ResetPasswordViewModel model)
 
             if (TryValidateModel(model) && ModelState.IsValid)
             {
-                if (await _userService.ResetPasswordAsync(model.Email, Encoding.UTF8.GetString(Convert.FromBase64String(model.ResetToken)), model.NewPassword, (key, message) => ModelState.AddModelError(key, message)))
+                if (await _userService.ResetPasswordAsync(model.Email, Encoding.UTF8.GetString(Convert.FromBase64String(model.ResetToken)), model.NewPassword, (key, message) => ModelState.AddModelError(key == "Password" ? nameof(ResetPasswordViewModel.NewPassword) : key, message)))
```
hishamco, line 134: Can we use `nameof` instead of magic string

hishamco (changes requested):

src/OrchardCore.Modules/OrchardCore.Users/Controllers/ResetPasswordController.cs
rjpowers10: I was using "Password" based on what UserService does [here](https://github.com/OrchardCMS/OrchardCore/blob/c476a3df796cb79d095fe767edbd4e10d4a27c75/src/OrchardCore/OrchardCore.Users.Core/Services/UserService.cs#L243). That's really the main problem here is UserService assumes your field is named "Password".

hishamco (approved):

Skrypt (approved):

PR #12315, opened 2022-09-02
Extract GetLanguageDirection() and IsRightToLeft() to CultureInfoExtensions

hishamco:


src/OrchardCore/OrchardCore.Localization.Abstractions/LanguageDirection.cs
```diff
@@ -1,36 +1,17 @@
-using System.Globalization;
+namespace OrchardCore.Localization;
 
-namespace OrchardCore.Localization
+/// <summary>
+/// Provides a language directions.
+/// </summary>
+public static class LanguageDirection
```
Skrypt, line 6: Why not use a string enumeration here?

hishamco: It's enum like class, coz the enumerations valid for integer based types

Skrypt (approved):

PR #12323, opened 2022-09-04
DbConnectionValidator fails if Shell Settings from database

### LINKED ISSUE #12334: DbConnectionValidator fails if Shell Settings come from same database
If shell settings come from the same database, the document table may already exist before the setup, in that case during the setup the DbConnectionValidator fails because the document table already exits.

jtkech:
Fixes #12334 

If shell settings come from the same database, the document table may already exist before the setup, in that case during the setup the DbConnectionValidator fails because the document table already exits.

**Updated**: So here, in case the Document table exists, we check if the `ShellDescriptor` document exists.

### DIFF
--- a/src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
+++ b/src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
@@ -85,7 +85,7 @@ public class DbConnectionValidator : IDbConnectionValidator
             if (result.HasRows)
             {
                 // The database and prefix are already in use.
-                return DbConnectionValidatorResult.DatabaseAndPrefixInUse;
+                return DbConnectionValidatorResult.DatabaseAndTablePrefixInUse;
             }
         }
         catch
--- a/src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
+++ b/src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
@@ -51,7 +51,7 @@ public class DbConnectionValidator : IDbConnectionValidator
 
         if (provider != null && !provider.HasConnectionString)
         {
-            return DbConnectionValidatorResult.DocumentNotFound;
+            return DbConnectionValidatorResult.Success;
         }
 
         if (String.IsNullOrWhiteSpace(connectionString))
@@ -80,20 +80,19 @@ public class DbConnectionValidator : IDbConnectionValidator
             sqlCommand.CommandText = sqlBuilder.ToSqlString();
 
             using var result = await sqlCommand.ExecuteReaderAsync();
+
+            // If the 'ShellDescriptor' document already exists.
             if (result.HasRows)
             {
-                // At this point we know that the 'ShellDescriptor' document exists.
-                return DbConnectionValidatorResult.DocumentFound;
+                // The database and prefix are already in use.
+                return DbConnectionValidatorResult.DatabaseAndPrefixInUse;
             }
-
-            // The 'Document' table exists but not the 'ShellDescriptor' document.
-            return DbConnectionValidatorResult.DocumentNotFound;
         }
         catch
         {
-            // At this point we know that the document table does not exist.
-            return DbConnectionValidatorResult.DocumentNotFound;
         }
+
+        return DbConnectionValidatorResult.Success;
     }
 
     private ISqlBuilder GetSqlBuilderForShellDescriptorDocument(string tablePrefix, DatabaseProviderName providerName)
--- a/src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
+++ b/src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
@@ -8,7 +8,7 @@ using Microsoft.Extensions.Options;
 using MySqlConnector;
 using Npgsql;
 using OrchardCore.Data.YesSql.Abstractions;
-using OrchardCore.Environment.Shell.Configuration;
+using OrchardCore.Environment.Shell.Descriptor.Models;
 using YesSql;
 using YesSql.Provider.MySql;
 using YesSql.Provider.PostgreSql;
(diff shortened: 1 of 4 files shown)


src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
```diff
@@ -71,6 +75,13 @@ public async Task<DbConnectionValidatorResult> ValidateAsync(string databaseProv
             return DbConnectionValidatorResult.InvalidConnection;
         }
 
+        // If the shell settings come from the same database, the document table may already exist.
+        if (_shellsSettingsSources.GetType().Name == "DatabaseShellsSettingsSources")
```
hishamco, line 79: Use `nameof`

hishamco (approved):
LGTM except a little comment

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
jtkech: Yes would have been better.

But no luck `DatabaseShellsSettingsSources` is defined in `OC` and here the project `OC.Data.YesSql` only references `OC.Abstractions` ;)

Otherwise I would have used `if (_shellsSettingsSources is DatabaseShellsSettingsSources)`

MikeAlhayek: @jtkech At what point would the table exists before the tenant is setup? I don't know if that check belongs in the database validator. Maybe add this check during setup instead?

MikeAlhayek:
@jtkech At what point would the table exists before the tenant is setup? I don't know if that check belongs in the database validator. Maybe add this check during setup instead?

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
hishamco: I see

jtkech:
@MikeAlhayek you know, this is when we use `.AddDatabaseShellsConfiguration()`.

On a first request we try to get the tenant settings from `tenants.json` by default, e.g. to see if we have to do a setup. When they come from a database (that may be the same or not), we retrieve the database info from `IConfiguration`, that's why we can connect to it before a setup, then if there is no tenants settings yet we store a default document in the Document table, which is then created if it doesn't exist yet.

So if `.AddDatabaseShellsConfiguration()` is configured with the same database, the setup fails because the Document table already exists.

jtkech:
Just to say that it is better to now use `DbConnectionValidator` ;)

Just to explain why it fails here.

jtkech:
So in place of checking if we use `DatabaseShellsSettingsSources` (that may use the same database or not), we now check (in case the Document table already exists) if the `ShellDescriptor` document exists

MikeAlhayek:
I don't think that check belongs in the validator. I could be wrong, but I feel it should be done at an upper level instead.

jtkech:
@MikeAlhayek 

Just for info on how we can create an isolated `ShellContext` with an empty `ShellDescriptor`, and then use it to create a scope which allows to use an `ISession` without having to deal with lower level YesSql things.

https://github.com/OrchardCMS/OrchardCore/blob/8b43028b5da9e697f39d2c477ca3916a5fdd94f7/src/OrchardCore/OrchardCore.Infrastructure/Shells.Database/Extensions/DatabaseShellContextFactoryExtensions.cs#L21-L32

https://github.com/OrchardCMS/OrchardCore/blob/8b43028b5da9e697f39d2c477ca3916a5fdd94f7/src/OrchardCore/OrchardCore.Infrastructure/Shells.Database/Configuration/DatabaseShellsSettingsSources.cs#L67-L72

jtkech:
> I don't think that check belongs in the validator. I could be wrong, but I feel it should be done at an upper level instead.

Sorry I don't understand, if the Document table already exists but only for the shells settings/configs, we still want to do the setup and also check e.g. the connection, but don't fail to validate.

Did you see my changes where I only check if the ShellDescriptor document exists in the Document table (if the table exists).

MikeAlhayek:
Yes. That code now makes more sense but still may have a flow.  What if the table was manually created? That would return DocumentNotFound even when it does. 

Maybe we should create a new Result other (maybe "DocumentExistsWithShellSettings") than document found to distinguish between "document table found with shell-settings" from "document table found". This validator is used when adding new tenant to also ensure a connection+prefix does not exists before we allow to use the connection.

jtkech:
> This validator is used when adding new tenant to also ensure a connection+prefix does not exists before we allow to use the connection.

Yes I know and it will still work

jtkech:
> What if the table was manually created? That would return DocumentNotFound even when it does.

Yes here document not found means ShellDescriptor document not found, we could change the name.

MikeAlhayek:
@jtkech here are my thought on this. I understand the need for identifying when a `ShellDescriptor` document exists. But, I don't like the use of `DatabaseAndTablePrefixInUse` or `Success`. I think the term `Success` is subjective here depending of the request needs. For example, Success to the setup call may mean different that to the new tenant request. Also, `DatabaseAndTablePrefixInUse` , may not make sense when prefix is not used.

Here is my suggestion, in `DbConnectionValidatorResult` rename `DocumentFound` to `DocumentTableFound`. Also rename `DocumentNotFound` to `DocumentTableNotFound`. Then add a new value called `ShellDescriptorDocumentFound`.

Finally your if statement would look like this

```
try
{
    var selectCommand = connection.CreateCommand();
    selectCommand.CommandText = selectBuilder.ToSqlString();

    using var result = await selectCommand.ExecuteReaderAsync();

    if (result.HasRows)
    {
        // At this point we know that the Document table and the ShellDescriptor record exists.
        return DbConnectionValidatorResult.ShellDescriptorDocumentFound;
    }

    // At this point the Document table exists with no ShellDescriptor document.
    return DbConnectionValidatorResult.DocumentFound;
}
catch
{
    // At this point we know that the document table does not exist.
    return DbConnectionValidatorResult.DocumentNotFound;
}
```

With the above, we know if an empty `Document` table exists vs `Document` table exists with at least one `ShellDescriptor` record. This is much more helpful to the consumer. In the setup call we validate using`ShellDescriptorDocumentFound` but in the TenantValidatior we validate `ShellDescriptorDocumentFound || DocumentFound`.

I would also keep the name variable name `selectCommand` instead of the proposed `sqlCommand` since the underlining command is strictly a select command.

jtkech:
@MikeAlhayek 

Okay no problem, feel free to update what you want on this branch, if you do not have access rights I suggest that you open a new PR, then I will close this one, the important thing being to fix the issue.

---

>  I think the term Success is subjective here and it's defined by the consumer needs.

Normally a validator succeeds or not, other data being additional info as error messages.

> For example, a success to the setup call means different than the new tenant request.

I don't think so as `OrchardCore_Shells_Database` can be configured with any `TablePrefix`.

> Also, DatabaseAndTablePrefixInUse, may not make sense when prefix is not used.

It just means that the prefix is null or empty.

>  In the setup call we validate using only using ShellDescriptorDocumentFound while in the TenantValidatior we validate ShellDescriptorDocumentFound || DocumentFound.

I don't think so as `OrchardCore_Shells_Database` can be configured with any `TablePrefix`.

> Lastly, I would also keep selectQuery.Take("1") at the end of the select query just to avoid searching for more than one record.

Yes I agree.

MikeAlhayek:
@jtkech I do not have permission to edit this PR. I submitted a PR as per your request. Feel free to close my PR if you want to follow through with this one instead.

jtkech:
Okay no problem I will review yours ASAP.

jtkech:
Close as replaced by #12342

PR #12326, opened 2022-09-04
Fix null check in ContentAnchorTag

hishamco:
This PR fixes a null check in
```csharp
if (metadata.CreateRouteValues == null) // BUG
    {
      if (routeValues != null)
      {
        foreach (var attribute in routeValues)
        {
          metadata.CreateRouteValues.Add(attribute.Key, attribute.Value);     
        }
      }
      ....
    }
```

Also I did a little refactoring

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Liquid/ContentAnchorTag.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Liquid/ContentAnchorTag.cs
@@ -109,14 +109,11 @@ namespace OrchardCore.Contents.Liquid
                 {
                     var metadata = await contentManager.PopulateAspectAsync<ContentItemMetadata>(displayFor);
 
-                    if (metadata.DisplayRouteValues != null)
+                    if (metadata.DisplayRouteValues != null && routeValues != null)
                     {
-                        if (routeValues != null)
+                        foreach (var attribute in routeValues)
                         {
-                            foreach (var attribute in routeValues)
-                            {
-                                metadata.DisplayRouteValues.Add(attribute.Key, attribute.Value);
-                            }
+                            metadata.DisplayRouteValues.Add(attribute.Key, attribute.Value);
                         }
 
                         customAttributes["href"] = urlHelper.Action(metadata.DisplayRouteValues["action"].ToString(), metadata.DisplayRouteValues);
@@ -125,10 +122,9 @@ namespace OrchardCore.Contents.Liquid
             }
             else if (editFor != null)
             {
-                contentItem = editFor;
-                var metadata = await contentManager.PopulateAspectAsync<ContentItemMetadata>(editFor);
+                var metadata = await PopulateAspectForContentItemMetadataAsync(editFor);
 
-                if (metadata.EditorRouteValues != null)
+                if (metadata.EditorRouteValues != null && routeValues != null)
                 {
                     foreach (var attribute in routeValues)
                     {
@@ -140,17 +136,13 @@ namespace OrchardCore.Contents.Liquid
             }
             else if (adminFor != null)
             {
-                contentItem = adminFor;
-                var metadata = await contentManager.PopulateAspectAsync<ContentItemMetadata>(adminFor);
+                var metadata = await PopulateAspectForContentItemMetadataAsync(adminFor);
 
-                if (metadata.AdminRouteValues != null)
+                if (metadata.AdminRouteValues != null && routeValues != null)
                 {
-                    if (routeValues != null)
+                    foreach (var attribute in routeValues)
                     {
-                        foreach (var attribute in routeValues)
-                        {
-                            metadata.AdminRouteValues.Add(attribute.Key, attribute.Value);
-                        }
+                        metadata.AdminRouteValues.Add(attribute.Key, attribute.Value);
                     }
 
                     customAttributes["href"] = urlHelper.Action(metadata.AdminRouteValues["action"].ToString(), metadata.AdminRouteValues);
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.Contents/Liquid/ContentAnchorTag.cs
```diff
@@ -230,6 +214,13 @@ public async ValueTask<Completion> WriteToAsync(List<FilterArgument> argumentsLi
             tagBuilder.RenderEndTag().WriteTo(writer, (HtmlEncoder)encoder);
 
             return Completion.Normal;
+
+            async Task<ContentItemMetadata> PopulateAspectForContentItemMetadataAsync(ContentItem item)
+            {
+                contentItem = item;
+
+                return await contentManager.PopulateAspectAsync<ContentItemMetadata>(item);
```
Skrypt, line 222:
```suggestion
                return await contentManager.PopulateAspectAsync<ContentItemMetadata>(contentItem);
```

hishamco: Both are valid ;)

Skrypt: But the idea is to create a new object so that the original one doesn't get morphed. Here, it is not consistent with what was done before if you use "item". It would be good to see if the original var is mutated too.

hishamco: If is that I will rid of the argument and assign it to `contentItem` directly, unless I did it for a reason

hishamco: It's for a reason

```
contentItem = createFor;
var metadata = await contentManager.PopulateAspectAsync<ContentItemMetadata>(createFor);
```

the `item` should has the `createFor` value

src/OrchardCore.Modules/OrchardCore.Contents/Liquid/ContentAnchorTag.cs
```diff
@@ -140,53 +136,41 @@ public async ValueTask<Completion> WriteToAsync(List<FilterArgument> argumentsLi
...
-                        }
+                        metadata.RemoveRouteValues.Add(attribute.Key, attribute.Value);
                     }
 
                     customAttributes["href"] = urlHelper.Action(metadata.RemoveRouteValues["action"].ToString(), metadata.RemoveRouteValues);
                 }
             }
             else if (createFor != null)
             {
-                contentItem = createFor;
```
Skrypt, line 179: Here, we copy createFor to contentItem because we want to have the original value of contentItem.DisplayText and contentItem.ContentType later on in this method.

src/OrchardCore.Modules/OrchardCore.Contents/Liquid/ContentAnchorTag.cs
```diff
@@ -140,53 +136,41 @@ public async ValueTask<Completion> WriteToAsync(List<FilterArgument> argumentsLi
...
-                var metadata = await contentManager.PopulateAspectAsync<ContentItemMetadata>(createFor);
+                var metadata = await PopulateAspectForContentItemMetadataAsync(createFor);
```
Skrypt, line 167: Here you are passing createFor just like before.

src/OrchardCore.Modules/OrchardCore.Contents/Liquid/ContentAnchorTag.cs
```diff
@@ -230,6 +214,13 @@ public async ValueTask<Completion> WriteToAsync(List<FilterArgument> argumentsLi
             tagBuilder.RenderEndTag().WriteTo(writer, (HtmlEncoder)encoder);
 
             return Completion.Normal;
+
+            async Task<ContentItemMetadata> PopulateAspectForContentItemMetadataAsync(ContentItem item)
+            {
+                contentItem = item;
```
Skrypt, line 220: Now this contentItem copy is useless because it is not used in the context of this method.

src/OrchardCore.Modules/OrchardCore.Contents/Liquid/ContentAnchorTag.cs
```diff
@@ -230,6 +214,13 @@ public async ValueTask<Completion> WriteToAsync(List<FilterArgument> argumentsLi
             tagBuilder.RenderEndTag().WriteTo(writer, (HtmlEncoder)encoder);
 
             return Completion.Normal;
+
+            async Task<ContentItemMetadata> PopulateAspectForContentItemMetadataAsync(ContentItem item)
```
Skrypt, line 218: Also here you define a task inside a method which is something we rarely do.

src/OrchardCore.Modules/OrchardCore.Contents/Liquid/ContentAnchorTag.cs
Skrypt: Ok, I missed something here. You are defining an async task at the end of this method so it is not a different method. I understand now what you are saying.

Skrypt (approved):

src/OrchardCore.Modules/OrchardCore.Contents/Liquid/ContentAnchorTag.cs
hishamco: Via the a local method

src/OrchardCore.Modules/OrchardCore.Contents/Liquid/ContentAnchorTag.cs
hishamco: Sure

PR #12328, opened 2022-09-04
Add missing null check operator in ShellSettingsManager.EnsureConfigurationAsync()

hishamco:


Skrypt (approved):

src/OrchardCore/OrchardCore/Shell/ShellSettingsManager.cs
```diff
@@ -239,7 +239,7 @@ private async Task EnsureConfigurationAsync()
                 .AddConfiguration(_applicationConfiguration)
                 .AddSourcesAsync(_tenantsConfigSources);
 
-            if (lastProviders.Count() > 0)
+            if (lastProviders?.Length > 0)
```
jtkech, line 242: Take care of this kind of changes, here that's okay as we are using `> 0`.

But we could have used `!=0`, in which case it would have been wrong.

            int? test = null;
            if (test > 0) // This is evaluated to false.
            {
                ;
            }
            else if (test != 0) // But this is evaluated to true.
            {
                ;
            }

That's why it's safer to use `test != null && test > 0` or `test.HasValue && test.Value > 0`.

Skrypt: I think actually this is a new code recommendation from Visual Studio to change .Count() to .Length in some places.

jtkech: Yes it is better, no problem about changing Count() to Length, I was talking about keeping the previous behavior when dealing with nullable types.

PR #12332, opened 2022-09-04
Check null for DeploymentStepFactory before step creation

hishamco:


Skrypt (approved):

src/OrchardCore.Modules/OrchardCore.Contents/Deployment/AddToDeploymentPlan/AddToDeploymentPlanController.cs
```diff
@@ -74,7 +74,15 @@ await _authorizationService.AuthorizeAsync(User, OrchardCore.Deployment.CommonPe
...
 
-            var step = (ContentItemDeploymentStep)_factories.FirstOrDefault(x => x.Name == nameof(ContentItemDeploymentStep)).Create();
+            var stepFactory = _factories.FirstOrDefault(x => x.Name == nameof(ContentItemDeploymentStep));
+
+            if (stepFactory == null)
+            {
+                return BadRequest();
+            }
+
+            var step = (ContentItemDeploymentStep)stepFactory.Create();
```
Skrypt, line 84:
```suggestion
            var step = (ContentItemDeploymentStep)stepFactory.Create();
```
```suggestion
            var step = (ContentItemDeploymentStep)stepFactory;
            
            if (step != null)
            {
                stepFactory.Create();
            }
```

hishamco: Why?!! we already checked the factory in line 79

Skrypt: Ok, because of the FirstOrDefault but it is not given that an explicit casting will return a value generally.

PR #12333, opened 2022-09-04
React to localization workaround removal for data annotations attributes

hishamco:
Related to #11130

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Email/ViewModels/SmtpSettingsViewModel.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Email/ViewModels/SmtpSettingsViewModel.cs
@@ -1,16 +1,13 @@
-using System;
-using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
-using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Localization;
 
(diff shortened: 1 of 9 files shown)


hishamco:
There is `RegisterExternalLoginViewModel` that needs to follow the others, but some extra properties that need to think properly

Also there are `AutoSetupOptions` and `SmtpSettings` that might need to rid-off `IValidatableObject` if it's possible

src/OrchardCore.Modules/OrchardCore.Email/ViewModels/SmtpSettingsViewModel.cs
```diff
@@ -1,16 +1,13 @@
-using System;
-using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
-using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Localization;
 
 namespace OrchardCore.Email.ViewModels
 {
-    public class SmtpSettingsViewModel : IValidatableObject
```
hishamco, line 9: @MikeAlhayek this the proper way to handle the data attributes validation after localizations support

hishamco:
Oops the build is failed, seems we need to validate the model by call post

PR #12339, opened 2022-09-05
Add a way to retrieve a UTC now datetime in Liquid

Skrypt:
utc: {{ "utc_now" | local | date: "%c" }}
local : {{ "now" | local | date: "%c" }}
local is the default : {{ "now" | date: "%c" }}

image

At the same time I'm questioning the TimeZoneFilter name now. "local" doesn't make sense anymore if we do this. The idea of the TimeZoneFilter should be to allow returning a local DateTime or a DateTime affected by an offset. Here, we should probably consider creating a filter that allows passing a TimeZoneId or an Offset value simply. Need to think about the design. But at the same time `| local` here doesn't make sense to be required to retrieve a UTC DateTime value.

Else, a refactor of this filter might become a breaking change. Maybe better simply create a new one.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Liquid/Filters/LocalTimeZoneFilter.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Liquid/Filters/LocalTimeZoneFilter.cs
@@ -1,0 +1,59 @@
+using System;
+using System.Globalization;
+using System.Threading.Tasks;
+using Fluid;
+using Fluid.Values;
+using OrchardCore.Modules;
+
+namespace OrchardCore.Liquid.Filters
+{
+    public class LocalTimeZoneFilter : ILiquidFilter
+    {
+        private readonly ILocalClock _localClock;
+
+        public LocalTimeZoneFilter(ILocalClock localClock)
+        {
+            _localClock = localClock;
+        }
+
+        public async ValueTask<FluidValue> ProcessAsync(FluidValue input, FilterArguments arguments, LiquidTemplateContext ctx)
+        {
+            DateTimeOffset value;
+
+            if (input.Type == FluidValues.String)
+            {
+                var stringValue = input.ToStringValue();
+
+                if (stringValue == "now" || stringValue == "today")
+                {
+                    value = await _localClock.LocalNowAsync;
+                }
+                else
+                {
+                    if (!DateTimeOffset.TryParse(stringValue, ctx.Options.CultureInfo, DateTimeStyles.AssumeUniversal, out value))
+                    {
+                        return NilValue.Instance;
+                    }
+                }
+            }
+            else
+            {
+                switch (input.ToObjectValue())
+                {
+                    case DateTime dateTime:
+                        value = dateTime;
+                        break;
+
+                    case DateTimeOffset dateTimeOffset:
+                        value = dateTimeOffset;
+                        break;
+
+                    default:
+                        return NilValue.Instance;
+                }
+            }
+
+            return new ObjectValue(await _localClock.ConvertToLocalAsync(value));
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


hishamco:
Why we didn't add UTC into Fluid? Or extend Now filter to support UTC through pipe operator

```{{ "now" | utc | date: "%c" }}```

**Comment by Skrypt** [comment]
image

The first value there is what you are suggesting which doesn't work.

**Comment by Skrypt** [comment]
```liquid
    local: {{ "now" | local | date: "%c" }}<br />
    utc: {{ "now" | utc | date: "%c" }}<br />
    local : {{ "now" | local | date: "%c" }}<br />
    local is the default : {{ "now" | date: "%c" }}<br />
```

image

Skrypt:
After adding UtcTimeZoneFilter:

```liquid
    local is the default: {{ "now" | date: "%c" }}<br />
    local: {{ "now" | local | date: "%c" }}<br />
    utc: {{ "now" | utc | date: "%c" }}<br />
```

image

Skrypt:
> Why we didn't add UTC into Fluid? Or extend Now filter to support UTC through pipe operator

It is not just "now" but all DateTimes that are converted when passed in these filters. So, yes, adding a UTC one makes sense then. It will basically remove or add your tenant timezone offset to these DateTimes.

src/OrchardCore.Modules/OrchardCore.Liquid/Filters/LocalTimeZoneFilter.cs
```diff
@@ -7,11 +7,11 @@
 
 namespace OrchardCore.Liquid.Filters
 {
-    public class TimeZoneFilter : ILiquidFilter
+    public class LocalTimeZoneFilter : ILiquidFilter
```
hishamco, line 10: Breaking change!!

src/OrchardCore.Modules/OrchardCore.Liquid/Filters/UtcTimeZoneFilter.cs
```diff
@@ -0,0 +1,59 @@
+using System;
+using System.Globalization;
+using System.Threading.Tasks;
+using Fluid;
+using Fluid.Values;
+using OrchardCore.Modules;
+
+namespace OrchardCore.Liquid.Filters
+{
+    internal class UtcTimeZoneFilter : ILiquidFilter
```
hishamco, line 10: public

src/OrchardCore.Modules/OrchardCore.Liquid/Filters/UtcTimeZoneFilter.cs
```diff
@@ -0,0 +1,59 @@
...
+                }
+            }
+            else
+            {
+                switch (input.ToObjectValue())
+                {
+                    case DateTime dateTime:
+                        value = dateTime;
+                        break;
+
```
hishamco, line 46: Remove extra line after `break`

hishamco (changes requested):

src/OrchardCore.Modules/OrchardCore.Liquid/Filters/LocalTimeZoneFilter.cs
Skrypt: How can it be? We register the "local" filter in the startup.cs so renaming the class is not a big deal. And it is registered for every tenants. Why would someone use this class and derivate it?

hishamco (approved):

src/OrchardCore.Modules/OrchardCore.Liquid/Filters/LocalTimeZoneFilter.cs
hishamco: It's a module, at least we need to document this as we already did in some other modules

sebastienros:
Relevant PR in Fluid to parse date time inputs (including `"now"`) https://github.com/sebastienros/fluid/pull/431

There is a time_zone filter to convert to any timezone. ,e.g., `{{ "now" | time_zone: 'utc' }}`

PR #12342, opened 2022-09-06
Fix DbConnectionValidator when Shell Settings from database

### LINKED ISSUE #12334: DbConnectionValidator fails if Shell Settings come from same database

MikeAlhayek:
Fix #12334

@jtkech here is the request PR.

### DIFF
--- a/src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
+++ b/src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
@@ -1,7 +1,7 @@
 namespace OrchardCore.Data;
 
 /// <summary>
-/// The result from validating a database connection using <see cref="IDbConnectionValidator"/>
+/// The result from validating a database connection using <see cref="IDbConnectionValidator"/>.
 /// </summary>
 public enum DbConnectionValidatorResult
 {
@@ -16,23 +16,22 @@ public enum DbConnectionValidatorResult
     NoProvider,
 
     /// <summary>
-    /// The connection string is valid. But, the Document table does not exists.
+    /// The connection string is valid and the 'Document' table does not exists.
     /// </summary>
     DocumentTableNotFound,
 
     /// <summary>
-    /// The connection string is valid and the Document table exists.
+    /// The connection string is valid and the 'Document' table exists.
     /// </summary>
     DocumentTableFound,
 
     /// <summary>
-    /// The connection string is valid and the Document table exists with at least one shell-descriptor document.
-    /// This also means that the Document table is used by an existing tenant.
+    /// The 'Document' table exists with no 'ShellDescriptor' document.
     /// </summary>
-    ShellDescriptorDocumentFound,
+    ShellDescriptorDocumentNotFound,
 
     /// <summary>
-    /// Unable to open a connection to the given database connection string.
+    /// Unable to open a connection with the given database connection string.
     /// </summary>
     InvalidConnection,
 
--- a/src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
+++ b/src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
@@ -1,5 +1,8 @@
 namespace OrchardCore.Data;
 
+/// <summary>
+/// The result from validating a database connection using <see cref="IDbConnectionValidator"/>
+/// </summary>
 public enum DbConnectionValidatorResult
 {
     /// <summary>
--- a/src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
+++ b/src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
@@ -1,25 +1,21 @@
 namespace OrchardCore.Data;
 
+/// <summary>
+/// 'Unknown' indicates that the connection string status is unknown or was not validated.
+/// 'NoProvider' indicated that the database provider is missing.
+/// 'DocumentTableNotFound' indicates that the connection string was valid, while the Document table does not exists.
+/// 'DocumentFound' indicates that the connection string was valid, while the Document table exists.
+/// 'ShellDescriptorDocumentFound' indicates that the connection string was valid and the document table exists with at least one shell-descriptor document.
+/// 'InvalidConnection' unable to open a connection to the given connection string.
+/// 'UnsupportedProvider' indicates invalid or unsupported database provider.
+/// </summary>
 public enum DbConnectionValidatorResult
 {
-    // 'Unknown' indicates that the connection string status is unknown or was not yet validated.
     Unknown,
-
-    // 'NoProvider' indicated that the provider is missing.
     NoProvider,
-
-    // 'DocumentTableNotFound' indicates that the connection string was valid, yet the Document table does not exist.
     DocumentTableNotFound,
-
-    // 'DocumentFound' indicates that the connection string was valid, yet the Document table exist.
     DocumentTableFound,
-
-    // 'ShellDescriptorDocumentFound' indicates that the connection string was valid and the document table exists with at least one entry of shell-descriptor document.
     ShellDescriptorDocumentFound,
-
-    // 'InvalidConnection' unable to open a connection to the given connection string.
     InvalidConnection,
-
-    // 'UnsupportedProvider' indicated invalid or unsupported database provider.
     UnsupportedProvider
 }
--- a/src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
+++ b/src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
@@ -34,7 +34,7 @@ public enum DbConnectionValidatorResult
     InvalidConnection,
 
     /// <summary>
-    /// Invalid or unsupported database provider.
+    /// Unsupported database provider.
     /// </summary>
     UnsupportedProvider
 }
--- a/src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
+++ b/src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
@@ -3,38 +3,38 @@ namespace OrchardCore.Data;
 public enum DbConnectionValidatorResult
 {
     /// <summary>
-    /// 'Unknown' indicates that the connection string status is unknown or was not validated.
+    /// The connection string status is unknown or was not validated.
     /// </summary>
     Unknown,
 
     /// <summary>
-    /// 'NoProvider' indicated that the database provider is missing.
+    /// The database provider is missing.
     /// </summary>
     NoProvider,
 
     /// <summary>
-    /// 'DocumentTableNotFound' indicates that the connection string was valid, while the Document table does not exists.
+    /// The connection string is valid. But, the Document table does not exists.
     /// </summary>
     DocumentTableNotFound,
 
     /// <summary>
-    /// 'DocumentTableFound' indicates that the connection string was valid, while the Document table exists.
+    /// The connection string is valid and the Document table exists.
     /// </summary>
     DocumentTableFound,
 
     /// <summary>
-    /// 'ShellDescriptorDocumentFound' indicates that the connection string was valid and the document table exists with at least one shell-descriptor document.
-    /// This also means that the Document table is used by a tenant.
+    /// The connection string is valid and the Document table exists with at least one shell-descriptor document.
+    /// This also means that the Document table is used by an existing tenant.
     /// </summary>
     ShellDescriptorDocumentFound,
 
     /// <summary>
-    /// 'InvalidConnection' unable to open a connection to the given connection string.
+    /// Unable to open a connection to the given database connection string.
     /// </summary>
     InvalidConnection,
 
     /// <summary>
-    /// 'UnsupportedProvider' indicates invalid or unsupported database provider.
+    /// Invalid or unsupported database provider.
     /// </summary>
     UnsupportedProvider
 }
--- a/src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
+++ b/src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
@@ -2,21 +2,24 @@ namespace OrchardCore.Data;
 
 public enum DbConnectionValidatorResult
 {
-    // Unknown indicates that the connection string status is unknown or was not yet validated
+    // 'Unknown' indicates that the connection string status is unknown or was not yet validated.
     Unknown,
 
-    // NoProvider indicated that the provider is missing
+    // 'NoProvider' indicated that the provider is missing.
     NoProvider,
 
-    // DocumentNotFound indicates that the connection string was valid, yet the Document table does not exist 
-    DocumentNotFound,
+    // 'DocumentTableNotFound' indicates that the connection string was valid, yet the Document table does not exist.
+    DocumentTableNotFound,
 
-    // DocumentFound indicates that the connection string was valid, yet the Document table exist
-    DocumentFound,
+    // 'DocumentFound' indicates that the connection string was valid, yet the Document table exist.
+    DocumentTableFound,
 
-    // InvalidConnection unable to open a connection to the given connection string
+    // 'ShellDescriptorDocumentFound' indicates that the connection string was valid and the document table exists with at least one entry of shell-descriptor document.
+    ShellDescriptorDocumentFound,
+
+    // 'InvalidConnection' unable to open a connection to the given connection string.
     InvalidConnection,
 
-    // UnsupportedProvider indicated invalid or unsupported database provider
+    // 'UnsupportedProvider' indicated invalid or unsupported database provider.
     UnsupportedProvider
 }
--- a/src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
+++ b/src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
@@ -1,21 +1,40 @@
 namespace OrchardCore.Data;
 
-/// <summary>
-/// 'Unknown' indicates that the connection string status is unknown or was not validated.
-/// 'NoProvider' indicated that the database provider is missing.
-/// 'DocumentTableNotFound' indicates that the connection string was valid, while the Document table does not exists.
-/// 'DocumentFound' indicates that the connection string was valid, while the Document table exists.
-/// 'ShellDescriptorDocumentFound' indicates that the connection string was valid and the document table exists with at least one shell-descriptor document.
-/// 'InvalidConnection' unable to open a connection to the given connection string.
-/// 'UnsupportedProvider' indicates invalid or unsupported database provider.
-/// </summary>
 public enum DbConnectionValidatorResult
 {
+    /// <summary>
+    /// 'Unknown' indicates that the connection string status is unknown or was not validated.
+    /// </summary>
     Unknown,
+
+    /// <summary>
+    /// 'NoProvider' indicated that the database provider is missing.
+    /// </summary>
     NoProvider,
+
+    /// <summary>
+    /// 'DocumentTableNotFound' indicates that the connection string was valid, while the Document table does not exists.
+    /// </summary>
     DocumentTableNotFound,
+
+    /// <summary>
+    /// 'DocumentTableFound' indicates that the connection string was valid, while the Document table exists.
+    /// </summary>
     DocumentTableFound,
+
+    /// <summary>
+    /// 'ShellDescriptorDocumentFound' indicates that the connection string was valid and the document table exists with at least one shell-descriptor document.
+    /// This also means that the Document table is used by a tenant.
+    /// </summary>
     ShellDescriptorDocumentFound,
+
+    /// <summary>
+    /// 'InvalidConnection' unable to open a connection to the given connection string.
+    /// </summary>
     InvalidConnection,
+
+    /// <summary>
+    /// 'UnsupportedProvider' indicates invalid or unsupported database provider.
+    /// </summary>
     UnsupportedProvider
 }
--- a/src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
+++ b/src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
@@ -93,6 +93,48 @@ public class DbConnectionValidator : IDbConnectionValidator
             // This also means that this table is used for other purposes that a tenant (ex., Database Shells Configuration.)
             return DbConnectionValidatorResult.DocumentTableFound;
         }
+        catch (SqlException e)
+        {
+            for (var i = 0; i < e.Errors.Count; i++)
+            {
+                if (e.Errors[i].Number == 207)
+                {
+                    // This means that the table exists but some expected columns do not exists.
+                    // This likely to mean that the the table was not created using YesSql.
+                    return DbConnectionValidatorResult.DocumentTableFound;
+                }
+                else if (e.Errors[i].Number == 208)
+                {
+                    return DbConnectionValidatorResult.DocumentTableNotFound;
+                }
+            }
+
+            return DbConnectionValidatorResult.InvalidConnection;
+        }
+        catch (MySqlException e)
+        {
+            return e.ErrorCode switch
+            {
+                MySqlErrorCode.NoSuchTable => DbConnectionValidatorResult.DocumentTableNotFound,
+                // This means that the table exists but some expected columns do not exists.
+                // This likely to mean that the the table was not created using YesSql.
+                MySqlErrorCode.BadFieldError => DbConnectionValidatorResult.DocumentTableFound,
+                _ => DbConnectionValidatorResult.InvalidConnection,
+            };
+        }
+        catch (PostgresException e)
+        {
+            return e.SqlState switch
+            {
+                //https://www.postgresql.org/docs/current/errcodes-appendix.html
+                // 'undefined_table'
+                "42P01" => DbConnectionValidatorResult.DocumentTableNotFound,
+
+                // 'undefined_column' this likely to mean that the the table was not created using YesSql.
+                "42703" => DbConnectionValidatorResult.DocumentTableFound,
+                _ => DbConnectionValidatorResult.InvalidConnection
+            };
+        }
         catch
         {
             // At this point we know that the document table does not exist.
@@ -104,7 +146,9 @@ public class DbConnectionValidator : IDbConnectionValidator
     {
         var selectBuilder = GetSqlBuilder(providerName, tablePrefix);
         selectBuilder.Select();
-        selectBuilder.AddSelector("*");
+        // Here we explicitly select the expected column used by YesSql instead of '*'
+        // to ensure that this table can be consumed by YesSql.
+        selectBuilder.AddSelector("Id, Type, Content, Version");
         selectBuilder.Table(_tableNameConvention.GetDocumentTable());
         selectBuilder.WhereAnd($"Type = '{typeof(ShellDescriptor).FullName}, {typeof(ShellDescriptor).Assembly.GetName().Name}'");
         selectBuilder.Take("1");
--- a/src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
+++ b/src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
@@ -85,8 +85,11 @@ public class DbConnectionValidator : IDbConnectionValidator
                 return DbConnectionValidatorResult.DocumentTableFound;
             }
 
-            var columns = Enumerable.Range(0, result.FieldCount).Select(result.GetName);
-            if (!columns.Any(c => c == "Type") || !columns.Any(c => c == "Content") || !columns.Any(c => c == "Version"))
+            var columns = Enumerable.Range(0, result.FieldCount)
+                .Select(result.GetName)
+                .Where(c => c == "Type" || c == "Content" || c == "Version");
+
+            if (columns.Count() != 3)
             {
                 // The 'Document' table exists with another schema.
                 return DbConnectionValidatorResult.DocumentTableFound;
--- a/src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
+++ b/src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
@@ -14,6 +14,7 @@ using YesSql.Provider.MySql;
 using YesSql.Provider.PostgreSql;
 using YesSql.Provider.Sqlite;
 using YesSql.Provider.SqlServer;
+using YesSql.Services;
 using YesSql.Sql;
 
 namespace OrchardCore.Data;
@@ -144,7 +145,7 @@ public class DbConnectionValidator : IDbConnectionValidator
         selectBuilder.Select();
         selectBuilder.Selector("*");
         selectBuilder.Table(_tableNameConvention.GetDocumentTable());
-        selectBuilder.WhereAnd($"Type = '{typeof(ShellDescriptor).FullName}, {typeof(ShellDescriptor).Assembly.GetName().Name}'");
+        selectBuilder.WhereAnd($"Type = '{new TypeService()[typeof(ShellDescriptor)]}'");
         selectBuilder.Take("1");
 
         return selectBuilder;
--- a/src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
+++ b/src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
@@ -35,7 +35,7 @@ public class DbConnectionValidator : IDbConnectionValidator
         _yesSqlOptions = yesSqlOptions.Value;
     }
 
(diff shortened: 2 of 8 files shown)


src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
```diff
@@ -2,21 +2,24 @@ namespace OrchardCore.Data;
...
 
-    // InvalidConnection unable to open a connection to the given connection string
+    // 'ShellDescriptorDocumentFound' indicates that the connection string was valid and the document table exists with at least one entry of shell-descriptor document.
+    ShellDescriptorDocumentFound,
+
+    // 'InvalidConnection' unable to open a connection to the given connection string.
     InvalidConnection,
 
-    // UnsupportedProvider indicated invalid or unsupported database provider
+    // 'UnsupportedProvider' indicated invalid or unsupported database provider.
```
jtkech, line 23: If you use comments, use the `<summary>` tag, also just before the `DbConnectionValidatorResult` declaration.

    /// <summary>
    /// Some comments.
    /// </summary>

MikeAlhayek: @jtkech done.

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
```diff
@@ -80,25 +80,31 @@ public async Task<DbConnectionValidatorResult> ValidateAsync(string databaseProv
...
 
     private ISqlBuilder GetSelectBuilderForDocumentTable(string tablePrefix, DatabaseProviderName providerName)
     {
         var selectBuilder = GetSqlBuilder(providerName, tablePrefix);
-
         selectBuilder.Select();
         selectBuilder.AddSelector("*");
         selectBuilder.Table(_tableNameConvention.GetDocumentTable());
         selectBuilder.Take("1");
+        selectBuilder.WhereAnd("Type = 'OrchardCore.Shells.Database.Models.DatabaseShellsSettings, OrchardCore.Infrastructure'");
```
jtkech, line 107: This is not the right document type, should be

        selectBuilder.WhereAnd($"Type = '{typeof(ShellDescriptor).FullName}, {typeof(ShellDescriptor).Assembly.GetName().Name}'");

src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
```diff
@@ -1,22 +1,21 @@
...
 
+/// <summary>
+/// 'Unknown' indicates that the connection string status is unknown or was not validated.
+/// 'NoProvider' indicated that the database provider is missing.
+/// 'DocumentTableNotFound' indicates that the connection string was valid, while the Document table does not exists.
+/// 'DocumentFound' indicates that the connection string was valid, while the Document table exists.
+/// 'ShellDescriptorDocumentFound' indicates that the connection string was valid and the document table exists with at least one shell-descriptor document.
+/// 'InvalidConnection' unable to open a connection to the given connection string.
+/// 'UnsupportedProvider' indicates invalid or unsupported database provider.
+/// </summary>
```
jtkech, line 11: Sorry I meant to replace the simple comments with a `<summary>` tag + a single line of comment, but still one per value below, then here also at the top a global `<summary>` tag + a single line of comment.

So that you have the intellisense for both the enum type and their individual values.

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
```diff
@@ -80,25 +80,31 @@ public async Task<DbConnectionValidatorResult> ValidateAsync(string databaseProv
...
-            return DbConnectionValidatorResult.DocumentFound;
+            // At this point, the query work and the 'Document' table exists.
+            if (result.HasRows)
+            {
+                // At this point we know that the Document table and the ShellDescriptor record exists.
+                return DbConnectionValidatorResult.ShellDescriptorDocumentFound;
+            }
+
+            // At this point the Document table exists with no ShellDescriptor document.
+            return DbConnectionValidatorResult.DocumentTableFound;
```
jtkech, line 91: This is wrong, when the document table exists but without ShellDescriptor (e.g. if using shell settings from database), the validation should pass, based on the current code here we would need to return DocumentTableNotFound as done line 96 in the catch, that's why when I named it .Success I was just using one line at the very end but not here and in the catch.

Or line 87 return DocumentTableFound so that you don't have to check ShellDescriptorDocumentFound in tenant validator and setup service, so remove this state and here return a new ShellDescriptorDocumentNotFound state that will let the validation pass.

src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs
```diff
@@ -181,7 +181,7 @@ private async Task<string> SetupInternalAsync(SetupContext context)
                 case DbConnectionValidatorResult.InvalidConnection:
                     context.Errors.Add(String.Empty, S["The provided connection string is invalid or server is unreachable."]);
                     break;
-                case DbConnectionValidatorResult.DocumentFound:
+                case DbConnectionValidatorResult.DocumentTableFound:
```
jtkech, line 184: You are missing

                case DbConnectionValidatorResult.ShellDescriptorDocumentFound:

As checked in tenant validator

Or remove this check in both if we apply the suggested change in `DbConnectionValidator`, see my related comment.

MikeAlhayek: Hummm, that was left out on purpose. It's my understanding that the shell document will exists before the default tenant is created. So when the shell settings document exists, we'll allow the setup to proceed during the setup only. Maybe I am missing the issue. Just so we are on the same page, could you please explain the step that encounter the original issue?

jtkech: As mentioned in the above comment, yes you can let it out and remove it also in tenant validator.

> Or remove this check in both if we apply the suggested change in DbConnectionValidator, see my related comment.

When we use shell settings it explains why the document table may exist and with any table prefix, as some other app level services could add data in the same database and also in the document table if they use yesSql.

So  when the document table exists we don't want to check if the shell settings document exists, but if the ShellDescriptor document exists, meaning that the database with a given prefix is in use. So when the ShellDescriptor (not shell settings) is found the validation should fail, if not found the validation can pass.

Yes the Default tenant has specificities but for most of the things it is a tenant as others. For example the Default tenant can have a non empty prefix and in that case before another tenant was allowed to have an empty prefix, it is only quite recently that this is no longer allowed in some validators.

MikeAlhayek: @jtkech I updated the code. So in the SetupService, we don't want to check for `DocumentTableFound` since in this case the table would already exists with the shell settings. However, in the Tenant validator we should check for both `DocumentTableFound` or `ShellDescriptorDocumentFound` to prevent adding new tenants later. I tested it out, but if you see other issue, feel free to directly change the PR.

jtkech:
@MikeAlhayek 

Yes, we are closed to solution.

> since in this case the table would already exists with the shell settings.

With the ShellDescriptor not shells settings, but I saw that you now query ShellDescriptor so that's okay. Yes if there is a shells settings document it only says that we use settings from database (in fact there is another shells configuration document in that case), but it doesn't indicate that the provided pair of database **and prefix** is already in use by a tenant. Note: Any other document can be added to the same database **and prefix** before a setup by another app level functionality as Shells from database does.

So when the document table exists, checking the ShellDescriptor document really means if it is already used or not by a tenant, this for the provided pair of database and table prefix.

> However, in the Tenant validator we should check for both DocumentTableFound or ShellDescriptorDocumentFound

The setup service and Tenant validator can do the same check as `AddDatabaseShellsConfiguration()` (or another app level functionality) can be configured with any database **and prefix**. The only thing that means that a pair of database and prefix is in used is when the document table exists AND the ShellDescriptor document exists in it.

src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs
```diff
@@ -181,7 +181,7 @@ private async Task<string> SetupInternalAsync(SetupContext context)
                 case DbConnectionValidatorResult.InvalidConnection:
                     context.Errors.Add(String.Empty, S["The provided connection string is invalid or server is unreachable."]);
                     break;
-                case DbConnectionValidatorResult.DocumentFound:
+                case DbConnectionValidatorResult.ShellDescriptorDocumentFound:
```
hishamco, line 184: Is this by intention? I see `DocumentFound` replaced by `DocumentTableFound`

src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
```diff
@@ -2,21 +2,39 @@ namespace OrchardCore.Data;
 
 public enum DbConnectionValidatorResult
 {
-    // Unknown indicates that the connection string status is unknown or was not yet validated
+    /// <summary>
+    /// 'Unknown' indicates that the connection string status is unknown or was not validated.
```
hishamco, line 6: No need to start with the enum member

src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
```diff
@@ -2,21 +2,39 @@ namespace OrchardCore.Data;
...
-    // InvalidConnection unable to open a connection to the given connection string
+    /// <summary>
+    /// 'ShellDescriptorDocumentFound' indicates that the connection string was valid and the document table exists with at least one shell-descriptor document.
+    /// This also means that the Document table is used by a tenant.
+    /// </summary>
+    ShellDescriptorDocumentFound,
+
+    /// <summary>
+    /// 'InvalidConnection' unable to open a connection to the given connection string.
+    /// </summary>
```
hishamco, line 33: I'm not sure why its named `Invalid` while the connection string is valid but unable to open!! May be we could rename it to `InUse`

src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
```diff
@@ -2,21 +2,39 @@ namespace OrchardCore.Data;
...
     InvalidConnection,
 
-    // UnsupportedProvider indicated invalid or unsupported database provider
+    /// <summary>
+    /// 'UnsupportedProvider' indicates invalid or unsupported database provider.
```
hishamco, line 37: Indicates that the database provider is not supported

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
```diff
@@ -111,7 +120,7 @@ private static IConnectionFactory GetFactory(DatabaseProviderName providerName,
             DatabaseProviderName.MySql => new DbConnectionFactory<MySqlConnection>(connectionString),
             DatabaseProviderName.Sqlite => new DbConnectionFactory<SqliteConnection>(connectionString),
             DatabaseProviderName.Postgres => new DbConnectionFactory<NpgsqlConnection>(connectionString),
-            _ => throw new ArgumentOutOfRangeException("Unsupported Database Provider"),
+            _ => throw new ArgumentOutOfRangeException(nameof(providerName), "Unsupported Database Provider"),
```
hishamco, line 123: Unsupported database provider

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
```diff
@@ -123,7 +132,7 @@ private ISqlBuilder GetSqlBuilder(DatabaseProviderName providerName, string tabl
             DatabaseProviderName.MySql => new MySqlDialect(),
             DatabaseProviderName.Sqlite => new SqliteDialect(),
             DatabaseProviderName.Postgres => new PostgreSqlDialect(),
-            _ => throw new ArgumentOutOfRangeException("Unsupported Database Provider"),
+            _ => throw new ArgumentOutOfRangeException(nameof(providerName), "Unsupported Database Provider"),
```
hishamco, line 135: Same as above

src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
```diff
@@ -2,21 +2,39 @@ namespace OrchardCore.Data;
...
+    /// 'DocumentTableFound' indicates that the connection string was valid, while the Document table exists.
+    /// </summary>
+    DocumentTableFound,
 
-    // InvalidConnection unable to open a connection to the given connection string
+    /// <summary>
+    /// 'ShellDescriptorDocumentFound' indicates that the connection string was valid and the document table exists with at least one shell-descriptor document.
+    /// This also means that the Document table is used by a tenant.
+    /// </summary>
+    ShellDescriptorDocumentFound,
```
hishamco, line 29: @jtkech are you agree to introduce a new database connection validation status related to ShellDescriptor?

hishamco (changes requested):

src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs
MikeAlhayek: yes

src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
hishamco: Sorry, I miss understood, I though you mean open the database not the connection ;)

jtkech:
@malhayek2014 

For info the new prefix constraint I was talking is about the **url prefix** (not the table prefix), we can still setup the Default tenant with a non empty url prefix, but now in that case we can't create a non default tenant with an empty url prefix (if hostname is also empty) which I think is wrong.

That said, just tried we can still setup the Default tenant with a non empty **table prefix**, and a non default tenant with an empty table prefix. Anyway, even if it was no more allowed I would still keep the concept of the association of a given database and a given table prefix (even if empty).

So for me the setup service and the tenant validator should use the same checks and with the same messages, for example a message stating that the provided association of the database and table prefix is already in use by an OC tenant.

Would need to be retried but based on the current code looks like that we would only need to just check for `ShellDescriptorDocumentFound` in the tenant validator as in the setup service and then use the same message mentioning the database and the provided table prefix.

MikeAlhayek:
@jtkech I disagree here. In the setup context I think all we need to check here whether the ShellDescriptor document exists or not. If the table exists without the ShellDeacriptor document, we'll also continue. 

However, in the tenant validator which is used in other places like OC.Tenants, we should also check if the table exists. We need to prevent the use of the database prefix if a table called Document exists "even if it has a different scheme than YesSql has". For example, what if we are connecting to a database that has that has example1_Document with no records. In that case we get DocumentTableFound and we should prevent the tenant from being added. 

I think we should also add a catch for an exception when the column `Type` does mot exists, we should also return DocumentTableFound.

src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
```diff
@@ -2,21 +2,39 @@ namespace OrchardCore.Data;
 
 public enum DbConnectionValidatorResult
```
jtkech, line 3: Yes I first suggested to just remove all comments here and below, but if we have comments should be summary tags as you did below which is good, but in that case what I meant is to also add a summary comment here just before the type itself, just for info as not so important.

Just for info see here https://github.com/dotnet/aspnetcore/blob/3ea008c80d5cc63de7f90ddfd6823b7b006251ff/src/SignalR/clients/csharp/Client.SourceGenerator/src/HubServerProxyGenerator.SourceGenerationSpec.cs#L45-L53 how to define an enum with power of 2 values so that in a given value we can hold multiple states, and also the `[Flags]` attribute that allows to then use the `.HasFlags()` helper.

MikeAlhayek: @jtkech I’ll add the `[Flags]` attribute and set power of 2 value. But, I am confused about the comment note. Do you want me to change the comments? What do you want the comments to look like?

jtkech: No the `Flags` thing was just for info.

I don't want you to remove the comments but eventually just add another one for the type itself.

    /// <summary>
    /// Bla bla bla. <= Add this comment
    /// </summary>
    public enum DbConnectionValidatorResult
    {
        /// <summary>
        /// The connection string status is unknown or was not validated.
        /// </summary>
        Unknown,

        /// <summary>
        /// The database provider is missing.
        /// </summary>
        NoProvider,

        ....

jtkech:
I totally disagree and I would give up ;)

But good news I may agree but not for the same reason (see the 2nd section).

> I disagree here. In the setup context I think all we need to check here whether the ShellDescriptor document exists or not. If the table exists without the ShellDeacriptor document, we'll also continue.

I agree and the same for any tenant.

> However, in the tenant validator which is used in other places like OC.Tenants, we should also check if the table exists. We need to prevent the use of the database prefix if a table called Document exists "even if it has a different scheme than YesSql has"

I don't agree, I would say also if it also use YesSql.

> what if we are connecting to a database that has that has example1_Document with no records. In that case we get DocumentTableFound and we should prevent the tenant from being added.

I don't agree.

---

But just realized that with the tenant removal feature, when we will remove a given tenant the related Document table will be removed, so the same table prefix should not be also used e.g. by shells from database, otherwise we would lose all shells settings when removing this tenant.

So finally I may agree with you but not for the same reason, not about what we would have to do before creating/setup a tenant but because we will be able to delete a tenant, otherwise I would still think that the validation would have to be the same for any tenant (Default tenant or not).

Most of the time we use an empty table prefix for both the Default tenant and to retrieve shells from database, but fortunally we will not be able to remove the default tenant, so that's okay.

Note: Good to keep as possible the same management for any tenant (Default tenant or not), when we talk about the Default tenant as the Host tenant for me it is wrong, yes it has specificities but most of the things are done in the same way, for example configuring, releasing, reloading, setup and so on.

MikeAlhayek:
@jtkech okay. Here is how I see things

When adding new tenant using `OC.Tenants`, at that point the default site already exists so it does not matter is we check for existing `ShellDescriptor` or not. What matters then is if the table exists or not. So if the user uses `site1` as the prefix and if the table `site1_Document` exists for whatever reason, we should prevent it from being used. This table could be created for other reason that OC... At least, I don't see a good reason to allow it. 

However, since the `Default` tenant can be created on the same `Document` table that the `ShellConfigurations` stores in, we can make an exception for it as only look for the document. So if the table exists with a ShellDescriptor entry, then we allow it to be used. **Which actually create the Document table if it does not exists?**

We should call the DbConnectionValidator on the code that actually created the Document table when shell configuration is used to prevent an error if the connection is invalid at that point.

I think we should also change the query we are executing to something like this instead so ensure that the table that is called document, has the same schema as YesSql would have.
```
  SELECT TOP 1 Content, Version
  FROM [OCTest3].[dbo].[Site1_Document]
  WHERE Type = '....'
```
Anyhow, you know OC better than me by far, so whatever decision you'll come back with I'll proceed with so we can fix this issue.

jtkech:
@MikeAlhayek 

No problem always good to understand and I may be wrong in some places.

Anyway, as said above we may keep this PR as it is because of the incoming tenant removal feature, I'm open to change the query if needed, I will re-think about it and re-do some tests tomorrow.

> When adding new tenant using OC.Tenants, at that point the default site already exists so it does not matter is we check for existing ShellDescriptor or not..

No if I understand what you mean, each tenant has its own ShellDescriptor in its Document table.

> However, since the Default tenant can be created on the same Document table that the ShellConfigurations stores in, we can make an exception for it.

But this is the case of any tenant because ShellConfigurations can be configured with any table prefix. But you may be right because of the incoming tenant removal feature.

> We should call the DbConnectionValidator on the code that actually created the Document table when shell configuration is used to prevent an error if the connection is invalid at that point.

Here it's okay to let it fail and log the error.

MikeAlhayek:
I'll do some more testing myself tomorrow as well and we'll come to a final approach. 

Unfortunately, YesSql always throws `SqlExeption` whether the column does not exists or the table does not exists. We can't really relay on the `where type =...` logic. Because if the table exists but the column `type` does not exists we return DocumentTableNotFound.

MikeAlhayek:
@jtkech I made some updates above like added the missing comments as per your request. Additionally, I added explicit exception validation. 

I also confirmed that in order for the code to work correctly, we should **only** check for `DbConnectionValidatorResult.ShellDescriptorDocumentFound` in the `SetupService` while checking both `DbConnectionValidatorResult.ShellDescriptorDocumentFound` and `DbConnectionValidatorResult.DocumentTableFound` in the `TenantValidator`. 

Since this is the point of argument here, I strongly suggest that you test out my code before we make any additional logic change to it since I am confident that it works as intended.

MikeAlhayek:
@jtkech no I still disagree. SetupContext should not be checking for TableFound. Please test out my code first before making changes to it just so you see why I still saying that TableFound cannot be part of the setup context. The DbValidator validate the connection regardless if you have a prefix or not. It'll only allow one tenant to have no prefix. So if no-prefix is used, no other tenant can have no-prefix. This isn't something new in this PR. This is how we have it in main branch today. 

Feel free to make any changes to the PR. But please test out what I have here first.

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
```diff
@@ -72,32 +73,72 @@ public async Task<DbConnectionValidatorResult> ValidateAsync(string databaseProv
...
         try
         {
             var selectCommand = connection.CreateCommand();
             selectCommand.CommandText = selectBuilder.ToSqlString();
 
             using var result = await selectCommand.ExecuteReaderAsync();
-
-            // at this point the query succeeded and the table exists
-            return DbConnectionValidatorResult.DocumentFound;
+            if (!isDefaultShell)
```
MikeAlhayek, line 82: I don't know why default is treated differently here?

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
```diff
@@ -72,32 +73,72 @@ public async Task<DbConnectionValidatorResult> ValidateAsync(string databaseProv
...
+            {
+                // The 'Document' table exists.
+                return DbConnectionValidatorResult.DocumentTableFound;
+            }
+
+            var columns = Enumerable.Range(0, result.FieldCount).Select(result.GetName);
+            if (!columns.Any(c => c == "Type") || !columns.Any(c => c == "Content") || !columns.Any(c => c == "Version"))
```
MikeAlhayek, line 89: Place all the or statements inside the same Any predict instead of having to check the columns all over again every time

jtkech:
@MikeAlhayek

I tested your code, it works but we can't create some use case by only using the admin UI because of the tenant validator, which is good, but this doesn't imply that we should not be consistent when checking any tenant in the setup service. Please test my code too, if you want.

Anyway I will analyze your last comments and will answer to them.

Hmm, also when I add some info, and I may be wrong, you rarely react to say if you knew it or no, for example when I said that each tenant has its own ShellDescriptor in its Document table.

MikeAlhayek (commented):
Does this solve the original problem?

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
jtkech: Good point

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
jtkech: Because finally we want to check if the ShellDescriptor exists only for the Default tenant. Because for other tenants we will be able to remove them that will remove their Document table.

So for non Default tenant we only check that the Document table exists, otherwise when removing a given tenant we may remove the Document table also used by an app level service, this is only allowed for the Default tenant that we will never be able to remove.

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
MikeAlhayek: I think the logic would be columns.Any(c => c != "Type" && c != …)

jtkech: Done but reverted as it is wrong

MikeAlhayek: @jtkech I noticed it. That is why I posted my previous comment. Please review my last statement. I have one more suggestion I'll level a separate comment in 1 mins

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
```diff
@@ -72,32 +73,72 @@ public async Task<DbConnectionValidatorResult> ValidateAsync(string databaseProv
...
+        {
+            var selectCommand = connection.CreateCommand();
+            selectCommand.CommandText = selectBuilder.ToSqlString();
 
-            return DbConnectionValidatorResult.DocumentNotFound;
+            using var result = await selectCommand.ExecuteReaderAsync();
+            if (!result.HasRows)
+            {
+                // The 'Document' table exists with no 'ShellDescriptor' document.
+                return DbConnectionValidatorResult.ShellDescriptorDocumentNotFound;
```
MikeAlhayek, line 111: It may be better to return ShellDescriptorFound here and flip the above if statement. I think Found is a bit easier to follow.

MikeAlhayek:
@jtkech i do apologize if sometimes I miss commenting on every point you share. Sometime it's hard to follow your chain of thought specially when I am using my phone instead of my PC. I never mean to ignore your comments as they are usually very informative and helpful. 

But yes I know about Flags attribute and that ShellDescriptor document is available for every tenant.

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
jtkech: Yes we could check that any column are equal to one of the names list, but then we would also need to check the `Id`, but I opted to not rely on the creation of a new column, e.g. the Version column was not existing at the beginning.

Otherwise we could check the fieldCount value and check that all columns are there and at the right place.

jtkech:
@MikeAlhayek Okay cool no problem, was just to be sure ;)

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
jtkech: I will look at it

jtkech: Okay I'm not against but then we would need to also check for `ShellDescriptorDocumentFound` in both tenant validator and setup service, to not validate in that case.

Then if the fipped if is false what we will return that will allow the validation to pass, not `DocumentTableNotFound` as you will say it is wrong and you will be right, so we would need to still keep `ShellDescriptorDocumentNotFound`, as you want.

Let me know what you prefer.

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
MikeAlhayek: @jtkech I think the statement `if (columns.Any(c => c != "Type" && c != "Content" && c != "Version"))** does the same thing as what you are doing.

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
MikeAlhayek: I think you don't need to change anything if you code look like this

```
if (result.HasRows)
{
    // The 'Document' table exists with no 'ShellDescriptor' document.
    return DbConnectionValidatorResult.ShellDescriptorDocumentFound;
}
```
So you don't need to change other places

jtkech: If I don't change anything else, if the the ShellDescriptor is not found we would return below DocumentTableFound which will not let pass the validation.

MikeAlhayek: Alight.  I would leave it as is then.

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
jtkech: Not really as we would need to also do `c != "Id"`, and then add others if a new column is added to the Document table, at the beginning I wanted to just check the `Type` column because if then we found a Type value equal to the ShellDescriptor type we could assume that it is the ShellDescriptor document.

But here I have another idea to onlly do one lookup.

MikeAlhayek: You could use intersect and check if the length = 3.

jtkech: Done but not tried yet

jtkech: Yes I did something like this

MikeAlhayek: What you did looks good to me. I would think we should also add "Id" column and make it != 4 but it's up to you.

Either way, the code looks good to me. I have not tested it. But, if you tested it, I say you merge it...

jtkech: Okay I will also check `Id` and retry it.

MikeAlhayek: That looks much better

jtkech (approved):

src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/DbConnectionValidatorResult.cs
jtkech: @hishamco Yes, in case of the Default tenant we allow that the Document table is already used by an host level service, as it happens e.g. when shell settings come from the database in place of `tenants.json`, this by using from the app startup `.AddDatabaseShellsConfiguration()`.

So in that case we allow that the Document table already exists but it should not contain the **ShellDescriptor** document meaning that an initialized tenant already uses it.

We could have done the same check for any tenants by allowing any tenant to use the same table prefix than e.g. the shells settings from database, but in the future we will be able to remove any non Default tenant and then its Document table, so we may lose e.g. all tenant settings.

Fortunally we can't remove the Default tenant.

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
```diff
@@ -72,32 +75,76 @@ public async Task<DbConnectionValidatorResult> ValidateAsync(string databaseProv
...
+    }
+
+    private ISqlBuilder GetSelectBuilderForShellDescriptorDocument(string tablePrefix, DatabaseProviderName providerName)
+    {
+        var selectBuilder = GetSqlBuilder(providerName, tablePrefix);
+
+        selectBuilder.Select();
+        selectBuilder.Selector("*");
         selectBuilder.Table(_tableNameConvention.GetDocumentTable());
+        selectBuilder.WhereAnd($"Type = '{typeof(ShellDescriptor).FullName}, {typeof(ShellDescriptor).Assembly.GetName().Name}'");
```
sebastienros, line 147: Isn't there a helper in yessql to build this? The logic might change.

sebastienros (approved):

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
jtkech: Didn't find it but will check it again

jtkech: Okay I think I found it => `TypeService`

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
```diff
@@ -72,32 +77,76 @@ public async Task<DbConnectionValidatorResult> ValidateAsync(string databaseProv
...
-            return DbConnectionValidatorResult.DocumentFound;
+            if (!isDefaultShell)
+            {
+                // The 'Document' table exists.
+                return DbConnectionValidatorResult.DocumentTableFound;
+            }
+
+            var requiredColumnsCount = Enumerable.Range(0, result.FieldCount)
+                .Select(result.GetName)
+                .Where(c => _requiredDocumentTableColumns.Contains(c))
```
MikeAlhayek, line 94: @jtkech I think here we should add `StringComparer.OrdinalIgnoreCase` to Contains. Some RDBMS has settings to force names to lower case.

jtkech: Done

MikeAlhayek: I think it ready to be merged.

jtkech: Yes I will do, just wait a little for now as I did 2 changes in the last hour ;)

PR #12365, opened 2022-09-10
Allowing multiple INavigationManager

Skrypt:
Provide the ability to create multiple INavigationManagers without needing to create a new SomeMenuShapes.cs. Using NavigationShapes should be sufficient.


src/OrchardCore/OrchardCore.Navigation.Core/ServiceCollectionExtensions.cs
```diff
@@ -1,20 +0,0 @@
...
-    public static class ServiceCollectionExtensions
-    {
-        /// <summary>
-        /// Adds tenant level services.
-        /// </summary>
-        /// <param name="services"></param>
-        /// <returns></returns>
-        public static IServiceCollection AddNavigation(this IServiceCollection services)
-        {
-            services.TryAddScoped<INavigationManager, NavigationManager>();
```
Skrypt, line 15: This cause issues with services resolution when using `scope.ServiceProvider.GetServices<INavigationManager>()` so I needed to remove it for now unless we find a better solution because it should not do that.

@jtkech

src/OrchardCore.Modules/OrchardCore.Navigation/NavigationShapes.cs
```diff
@@ -39,33 +39,38 @@ public void Discover(ShapeTableBuilder builder)
                     var viewContextAccessor = context.ServiceProvider.GetRequiredService<ViewContextAccessor>();
                     var viewContext = viewContextAccessor.ViewContext;
 
-                    var navigationManager = context.ServiceProvider.GetRequiredService<INavigationManager>();
+                    var navigationManagers = context.ServiceProvider.GetServices<INavigationManager>();
```
Skrypt, line 42: Here is where I was not able to get the registered services when using the ServiceCollectionExtensions.cs for registering the admin menu INavigationManager.

src/OrchardCore/OrchardCore.Navigation.Core/ServiceCollectionExtensions.cs
jtkech: @Skrypt 

If you need to register multiple implementations and be able to resolve them with `GetServices()`.

    services.AddScoped<INavigationManager, NavigationManager1>();
    services.AddScoped<INavigationManager, NavigationManager2>();

To prevent to re-add the same implementation

    services.TryAddEnumerable(ServiceDescriptor.Scoped<INavigationManager, NavigationManager1>());
    services.TryAddEnumerable(ServiceDescriptor.Scoped<INavigationManager, NavigationManager2>());

Idem with a generic extension

    public static IServiceCollection AddNavigation<T>(this IServiceCollection services)
    {
        services.TryAddEnumerable(ServiceDescriptor.Scoped<INavigationManager, T>());
    }

Skrypt: What you are suggesting here is that TryAddEnumerable will work better than TryAddScoped in this case inside a ServiceCollectionExtensions method?

jtkech: Yes, it depends on what is needed, `TryAddScoped` prevent to re-register the same `ServiceType` whatever the `ImplementationType`, useful in an extension which may be called multiple time. But when you want to override a service implementation you need to use `AddScoped` (or `Replace`) so that the last registered will be resolved when you inject/get a single service of this `ServiceType`.

Here the `ServiceType` is `INavigationManager`, the `ImplementationType` is `NavigationManager`.

When we need to inject an `IEnumerable<IScopedService>` or use `GetServices`, we should not use `TryAddScoped` to register other implementations, most of the time we just use `AddScoped` and that's okay if we are sure that we use it once per `ImplementationType`, otherwise when resolved e.g. with `Getservices` we would get and use the same implementation multiple times.

Finally to register multiple `ImplementationType` of the same `ServiceType` but also to prevent from registering duplicate `ImplementationType` we can use `TryAddEnumerable `.

src/OrchardCore/OrchardCore.Navigation.Core/ServiceCollectionExtensions.cs
```diff
@@ -12,8 +12,7 @@ public static class ServiceCollectionExtensions
         /// <returns></returns>
         public static IServiceCollection AddNavigation(this IServiceCollection services)
         {
-            services.TryAddScoped<INavigationManager, NavigationManager>();
-
+            services.TryAddEnumerable(ServiceDescriptor.Scoped<INavigationManager, NavigationManager>());
```
hishamco, line 15: Add new line after

src/OrchardCore.Modules/OrchardCore.AdminMenu/Startup.cs
```diff
@@ -36,7 +36,7 @@ public override void ConfigureServices(IServiceCollection services)
             services.AddScoped<IAdminMenuPermissionService, AdminMenuPermissionService>();
 
             services.AddScoped<IAdminMenuService, AdminMenuService>();
-            services.AddScoped<AdminMenuNavigationProvidersCoordinator, AdminMenuNavigationProvidersCoordinator>();
```
hishamco, line 39: Not sure why we did this before ;)

src/OrchardCore.Modules/OrchardCore.Navigation/NavigationShapes.cs
```diff
@@ -38,34 +38,37 @@ public void Discover(ShapeTableBuilder builder)
...
-                        var routeData = new RouteValueDictionary(route.Values);
-                        var query = httpContext.Request.Query;
+                        var menuItems = await navigationManager.BuildMenuAsync(menuName, viewContext);
+                        var httpContext = httpContextAccessor.HttpContext;
 
-                        if (query != null)
+                        if (httpContext != null)
                         {
-                            foreach (var pair in query)
+                            // adding query string parameters
```
hishamco, line 52: Can we extract this into a `private` method?

hishamco (changes requested):

src/OrchardCore.Modules/OrchardCore.AdminMenu/Startup.cs
Skrypt: I can't say :)

src/OrchardCore.Modules/OrchardCore.Navigation/NavigationShapes.cs
Skrypt: There will be no reusability, but I see your point. I don't think it is necessary here.

PR #12368, opened 2022-09-10
Update the action bar on manage contents page

### LINKED ISSUE #12367: Improve the look of the action bar in all the views
### Describe the bug
When using a mobile phone to manage contents (or any view with action bar), the search input on the top shows up very small.

image

Also, on the filter button on the left, can we remove the round corner between the button and the input?

image

MikeAlhayek:
Fix #12367

## Web view

Here is a before screenshot 
image

Here is an after screenshot
image

## Mobile view
Here is a before screenshot
image

Here is an after schreenshot
image


hishamco:
Can we increase the search bar little to make the spacing consistent

MikeAlhayek:
@hishamco I made some more changes to the second action bar along with the action bar spacing. check out the updated screenshots

Skrypt:
Can we see what the searchbar looks like now when width = 100%?

Skrypt:
Also, need to see if these UI fixes need to be applied everywhere else.

MikeAlhayek:
@Skrypt the PR #12368 has the before and after screenshots.

image

image

Skrypt:
can you try and add a w-md-50 or w-lg-50 class there.

src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsAdminListSearch.cshtml
```diff
@@ -2,32 +2,36 @@
...
+    {
+        "dropdown-menu"
+    };
+
+    if (CultureInfo.CurrentUICulture.IsRightToLeft())
+    {
+        dropdownClassList.Add("dropdown-menu-start");
+    }
 }
-<div class="input-group input-group-sm w-50 d-inline-flex has-filter">
```
Skrypt, line 6: here try w-md-50 or w-lg-50 instead.

MikeAlhayek:
I personally think it looks better with 100% width.

Skrypt:
I can't argue with taste. If you have a 4k screen it will make a long search input bar. That's why it was added.

Skrypt:
Here as an example Github does not use full width:

image

Skrypt:
Unless you would need to type long search strings that are equivalent to the full width size of a 4k screen I don't see the added value either.

Skrypt (approved):

hishamco:
I'm not sure if I'm the only one see extra space before the new dropdown

Skrypt:
Add a screenshot. I can't see it either! 😄

hishamco:
I think there's two extra pixels before the new dropdown comparing with the spacing after, needs BIG 👀

Skrypt:
I don't see any extra pixels in his screenshots.

hishamco:
@MikeAlhayek could you please inspect the spacing before and after the new dropdown? It looks good but that's why designer is better than devs because they taking care about each pixel :)

MikeAlhayek:
@hishamco I did not add a drop down menu. Can you show us where are you pointing to? I would say share a screen shot with the develops so that they can see through a designer eyes :)

Skrypt:
The filter button?

hishamco:
I just measure the spaces before and after the new dropdown seems the quality here as @Skrypt mentioned, both are 16 px, so seems I need to close my eyes ;)

hishamco:
@MikeAlhayek could you please check tenants page? I remembered I saw similar issue

MikeAlhayek:
@hishamco yea same problem exists in multiple pages. I want to focus on Contents module for now, once all agree on the change and is merged, I'll apply the same on the rest like OC.Tenant, OC.Features and possibly others.

hishamco:
We already agreed ;) I might help you on this

Thanks

hishamco (approved):

src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsAdminListSearch.cshtml
hishamco: @Skrypt are you fine to merge this without this suggestion?

hishamco:
@MikeAlhayek I just waiting for @Skrypt to reply on my above comment meanwhile could you please file an issue list all the places that we need to modify the action bar, this will make it easy for tracking and I can help you on this especially if it has a long list

src/OrchardCore.Modules/OrchardCore.Contents/Views/ContentsAdminListSearch.cshtml
MikeAlhayek: @Skrypt if you must have a max width set, we could add `w-xxl-75` just for super large screens.

MikeAlhayek:
Combined into #12421

hishamco:
I'm not sure why closed the PRs, small manageable PRs always good

PR #12374, opened 2022-09-11
Add notes in the documents on how to configure YesSql

MikeAlhayek:


hishamco:
Can we describe all the settings in the `YesSqlOptions` like what we did in other modules

MikeAlhayek:
@hishamco added.

PR #12377, opened 2022-09-12
ContentAnchorTag is broken

### LINKED ISSUE #12376: ContentAnchorTag is broken.
@hishamco Looks like ContentAnchorTag is broken since #12326 

Post List Broken

jtkech:
Fixes #12376 

Post List Broken

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Liquid/ContentAnchorTag.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Liquid/ContentAnchorTag.cs
@@ -108,12 +108,14 @@ namespace OrchardCore.Contents.Liquid
                 else
                 {
                     var metadata = await contentManager.PopulateAspectAsync<ContentItemMetadata>(displayFor);
-
-                    if (metadata.DisplayRouteValues != null && routeValues != null)
+                    if (metadata.DisplayRouteValues != null)
                     {
-                        foreach (var attribute in routeValues)
+                        if (routeValues != null)
                         {
-                            metadata.DisplayRouteValues.Add(attribute.Key, attribute.Value);
+                            foreach (var attribute in routeValues)
+                            {
+                                metadata.DisplayRouteValues.Add(attribute.Key, attribute.Value);
+                            }
                         }
 
                         customAttributes["href"] = urlHelper.Action(metadata.DisplayRouteValues["action"].ToString(), metadata.DisplayRouteValues);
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.Contents/Liquid/ContentAnchorTag.cs
```diff
@@ -108,12 +108,14 @@ public async ValueTask<Completion> WriteToAsync(List<FilterArgument> argumentsLi
                 else
                 {
                     var metadata = await contentManager.PopulateAspectAsync<ContentItemMetadata>(displayFor);
-
-                    if (metadata.DisplayRouteValues != null && routeValues != null)
+                    if (metadata.DisplayRouteValues != null)
                     {
-                        foreach (var attribute in routeValues)
+                        if (routeValues != null)
```
hishamco, line 113: In my PR I just moved the inner condition to the upper one, coz it will be the same. I'm still wondering how this causes the bug

jtkech: Try it with the dev branch and then with this PR, I think you will understand.

hishamco: Sure, but I remembered I changed the condition in 3 or 4 places

Skrypt: hmm, I may say the same thing as @hishamco 
What is the logical difference between this and what @hishamco did?

both need to be true to pass.

jtkech: @Skrypt @hishamco 

You are missing the below line 121. Here as before, if `DisplayRouteValues` is not null and the ambient `routeValues` is null the below line 121 is still executed,  which is no more the case in the current dev branch.

    customAttributes["href"] = urlHelper.Action(metadata.DisplayRouteValues["action"] ...);

Skrypt (approved):

src/OrchardCore.Modules/OrchardCore.Contents/Liquid/ContentAnchorTag.cs
hishamco: Seems we should apply the fix for other case

editFor, adminFor, createFor and removeFor

jtkech: Yes indeed, will do

hishamco (approved):

PR #12400, opened 2022-09-13
Update the action bar on manage features page

MikeAlhayek:
This PR depends on #12368. PR #12368 should be merged first as it provides some key css changes.

Here is the before view
image

Here is the after view
image


MikeAlhayek:
Combined into #12421

Issue #12404, opened 2022-09-13
Check that OC works with .NET 7.0


agriffard:
https://dotnet.microsoft.com/en-us/download/dotnet/7.0

Before the November release, to be sure that nothing would be blocking.

PR #12405, opened 2022-09-13
.NET 7.0

agriffard:
https://dotnet.microsoft.com/en-us/download/dotnet/7.0

https://www.nuget.org/packages/Microsoft.AspNetCore.Authentication.AzureAD.UI does not have a `Preview.7` version.


agriffard:
Related to #12404

agriffard:
The build seems to work, except the Cypress Functional tests, which was expected because they explicitly target .NET 6.0, I think.

https://github.com/OrchardCMS/OrchardCore/actions/runs/3048201862/jobs/4913012398

/cc @jptissot

sebastienros:
> except the Cypress Functional tests, which was expected because they explicitly target .NET 6.0

Where ?

sebastienros:
https://github.com/OrchardCMS/OrchardCore/blob/main/test/OrchardCore.Tests.Functional/cypress-commands/src/test-runner.js#L24

agriffard:
I don't know why this package does not have a 7.0.0 version : 
https://www.nuget.org/packages/Microsoft.AspNetCore.Authentication.AzureAD.UI

sebastienros:
Checking...
But I think we should find why we have it and how we could get rid of it.

ns8482e:
https://learn.microsoft.com/en-us/dotnet/core/compatibility/aspnet-core/5.0/authentication-aad-packages-obsolete

agriffard:
`Microsoft.AspNetCore.Authentication.AzureAD.UI` is used in the Microsoft.Authentication module : 
https://github.com/search?q=repo%3AOrchardCMS%2FOrchardCore%20Microsoft.AspNetCore.Authentication.AzureAD.UI&type=code

agriffard:
https://github.com/OrchardCMS/OrchardCore/blob/0f1acae452e6cca2c5fa7f062e4e6c9c71e60540/src/OrchardCore.Modules/OrchardCore.Microsoft.Authentication/Startup.cs#L64

Anyone wants to help to replace those obsolete classes ?

hishamco:
I might help you on this on the weekend

sebastienros:
Still, can you see why we need this package, and if there is a reason add a comment on the nuget reference.[2.0.5-preview](https://www.nuget.org/packages/Microsoft.Identity.Web/2.0.5-preview) targets 6.0 which is nice since we need to target it still, however it's "preview" ...

sebastienros:
@MichaelPetrinolis do you think you could take a look at this PR? We don't know how to update OrchardCore.Microsoft.Authentication with Microsoft.Identity.Web (AzureAD.UI is not shipped in net7)

MichaelPetrinolis:
Sure @sebastienros , will check it.

MichaelPetrinolis:
@agriffard I created the #13042 which compiles, and in a sort test it worked. But needs more test

sebastienros:
We might need two sets of package versions based on 6.0 or 7.0.

PR #12409, opened 2022-09-13
Change the sort order for permission on edit role view

### LINKED ISSUE #12408: Sort permission by Descriptions instead on a name in Roles UI
Currently, we sort permission by `Name` which it isn't something visible to the users. When the description is different than, permissions won't show up in alphabetical order because we sort by `name` by show `description`

To sort them properly, they should be sorted by `Description` property instead since that is what the users see or care about.

The following screenshot is an example of how sorting by `Name` does not work well. As you can see the permission "Assign users...." permission is showing up after the "Manage....". The reason in this case the permissions are not showing up as expected is because the permission "Assign any role" has a name of `AssignRole`. However, `Assign users to ....` has a name of `ManageUsersInRole...`.
image

MikeAlhayek:
Fix #12408

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Roles/Views/Admin/Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Roles/Views/Admin/Edit.cshtml
@@ -52,7 +52,7 @@
                     </tr>
                 </thead>
                 <tbody>
-                    @foreach (var permission in group.Value.OrderBy(x => x.Name))
+                    @foreach (var permission in group.Value.OrderBy(x => x.Description))
                     {
                         <tr data-text="@permission.Description">
                             <td class="col-10">


src/OrchardCore.Modules/OrchardCore.Roles/Views/Admin/Edit.cshtml
```diff
@@ -52,7 +52,7 @@
                     </tr>
                 </thead>
                 <tbody>
-                    @foreach (var permission in group.Value.OrderBy(x => x.Name))
+                    @foreach (var permission in group.Value.OrderBy(x => x.Description))
```
hishamco, line 55: Why it doesn't sort the permissions alphabatically as you mentioned in the issue description?

MikeAlhayek: I don't understand the question

hishamco: > Currently, we sort permission by Name which it isn't something visible to the users. In many cases this does not show the permission in alphabetical order.

Check the last phrase

MikeAlhayek: Currently we sort by "Name" or "ID" which is something that isn't visible to the user on the UI. We show description field. So we should sort by the visible field so that the visible text is correctly sorted.

hishamco: Description could be changed, but I see sorting by name is suited here

MikeAlhayek: Yes name does not change but it's also not visible to the user. Description can change and is visible to the user, so we must sort by what is visible to the user not by hidden values. You should test this out so you can understand the problem I am solving better. Why do you think it's better to sort by hidden value in the UI rather than a visible value?

hishamco: Again description is just description, so I disagree with this change

MikeAlhayek: @hishamco 

Here is how it's sorted today, as you can see "Assign users...." permission is showing up after the 
image

The reason in this case the permissions are not showing up as expected is because the permission "Assign any role" has `AssignRole` name. However, `Assign users to ....` has `ManageUsersInRole...` name.

@Skrypt the sorting is something you and I worked on a while back. Do you agree that it should be description?

MikeAlhayek: @hishamco they are grouped by feature. When this code is executed, we are sorting the permission with in a group. So the sorting is done by feature title, then by category.

hishamco: Ok, then

Skrypt (approved):

Skrypt:
Here is the proof of that.

image

PR #12420, opened 2022-09-14
Prevent Users feature from breaking when Roles feature is disabled

### LINKED ISSUE #12406: When the roles feature is disable the site breaks
### Describe the bug
If the user disable the roles module the site crashes because the Users needs the following services to work
 - IRoleService
 - RoleManager<IRole>
 - IRoleStore<IRole>

### To Reproduce
1. Disable the Roles module

### Expected behavior
No exception when disabling the roles module.

There are 2 possible fixes here. 
 1. In the users module add `OrchardCore.Roles` to the dependency list. But this would force the user to use the roles modules to be able to authenticate users in OC.
 2. Add the following services to the Users startup file
```
services.TryAddScoped<IRoleService, RoleService>();
services.TryAddScoped<RoleManager<IRole>>();
services.TryAddScoped<IRoleStore<IRole>, RoleStore>();
```

But to do we have to move `RolesDocument`, `RoleService` and `RoleStore` from `OrchardCore.Roles` to `OrchardCore.Roles.Core` which I believe where they should live anyway.

MikeAlhayek:
Fix #12406


src/OrchardCore.Modules/OrchardCore.Users/OrchardCore.Users.csproj
```diff
@@ -29,9 +29,9 @@
     <ProjectReference Include="..\..\OrchardCore\OrchardCore.Navigation.Core\OrchardCore.Navigation.Core.csproj" />
     <ProjectReference Include="..\..\OrchardCore\OrchardCore.Recipes.Abstractions\OrchardCore.Recipes.Abstractions.csproj" />
     <ProjectReference Include="..\..\OrchardCore\OrchardCore.ResourceManagement\OrchardCore.ResourceManagement.csproj" />
-    <ProjectReference Include="..\..\OrchardCore\OrchardCore.Roles.Core\OrchardCore.Roles.Core.csproj" />
     <ProjectReference Include="..\..\OrchardCore\OrchardCore.Settings.Core\OrchardCore.Settings.Core.csproj" />
     <ProjectReference Include="..\..\OrchardCore\OrchardCore.Setup.Abstractions\OrchardCore.Setup.Abstractions.csproj" />
     <ProjectReference Include="..\..\OrchardCore\OrchardCore.Workflows.Abstractions\OrchardCore.Workflows.Abstractions.csproj" />
+    <ProjectReference Include="..\OrchardCore.Roles\OrchardCore.Roles.csproj" />
```
sebastienros, line 35: We don't reference modules from modules.

MikeAlhayek: @sebastienros 
The problem here is that the `RoleDocument` field belongs to the `OrchardCore.Roles` project and if we move it to `OrchardCore.Roles.Core` we would have to decorated it with `[YesSql.SimplifiedTypeName("OrchardCore.Roles.Models.RolesDocument, OrchardCore.Roles")]` which is not something you'll consider

src/OrchardCore.Modules/OrchardCore.Users/Services/EmptyRoleStore.cs
```diff
@@ -0,0 +1,63 @@
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Identity;
+using OrchardCore.Security;
+
+namespace OrchardCore.Users.Services;
+
+public class EmptyRoleStore : IRoleStore<IRole>
```
hishamco, line 8: Can we name `NullRoleStore` and move it into `OC.Roles`, then the registration for the services can be done in extension methods

MikeAlhayek: We could rename it. I don't think we should put it in roles. I think it should even be internal to Users.

But either way, I think this fixes the problem on hand but create another problem

hishamco: It's quite similar to `NullLocalizer` to make a default `IStringLocalizer` when the localization feature is disabled

MikeAlhayek: agreed. In this case we are dealing with claims where are managed at the Role level currently. if we don't want to require roles, then we need a service someone that would provide claims/permission when the user first login "including admin" otherwise the site won't be usable

Skrypt: I agree with `NullRoleStore` name. It doesn't return an empty result but a null result.

src/OrchardCore/OrchardCore.Users.Core/Services/RoleBasedUserStore.cs
```diff
@@ -0,0 +1,56 @@
...
+using Microsoft.AspNetCore.Identity;
+using Microsoft.Extensions.Logging;
+using OrchardCore.Security.Services;
+using OrchardCore.Users.Handlers;
+using OrchardCore.Users.Models;
+using YesSql;
+
+namespace OrchardCore.Users.Services
+{
+    public class RoleBasedUserStore : UserStore
```
hishamco, line 16: Do we agree to add such thing, in the related issue?

src/OrchardCore.Modules/OrchardCore.Users/Permissions.cs
```diff
@@ -14,31 +11,14 @@ public class Permissions : IPermissionProvider
...
-        {
-            var list = new List<Permission>
+            return Task.FromResult<IEnumerable<Permission>>(new List<Permission>
             {
                 ManageUsers,
                 ManageOwnUserInformation,
                 ViewUsers
-            };
-
-            var roles = (await _roleService.GetRoleNamesAsync())
```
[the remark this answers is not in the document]
MikeAlhayek: We did not remove it. It was moved it a difference service that only get registered when OC.Roles is available. The same code can be found in [UserRolePermissions.cs](https://github.com/OrchardCMS/OrchardCore/pull/12420/files#diff-fa85ff6a4470ffcee5c5e09c103e4eb39f9cdc9cb2ea944d20f0f3c20af86e69)

src/OrchardCore.Modules/OrchardCore.Users/Services/NullRoleStore.cs
```diff
@@ -0,0 +1,63 @@
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Identity;
+using OrchardCore.Security;
+
+namespace OrchardCore.Users.Services;
+
+public class NullRoleStore : IRoleStore<IRole>
```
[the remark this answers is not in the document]
MikeAlhayek: I think it's best if this is kept in the only module that would ever use it.

ns8482e:
IMO UserStore should stay unchanged, the approach we should take here is 

1. In AddOrchard() after AddIdentity - implement NULLRoleStore  and register `IRoleStore<IRole>, RoleStore`,   
2. In Roles module replace  `NULLRoleStore` with actual RoleStore
3. UserStore should stay unchanged and read roles either from NULLRoleStore or RoleStore depending on UserRole module.
4. Mark UserRole module enable only, means once enabled can't be disabled

MikeAlhayek:
It was changed because it does not need to depend on `IRoleService`. Why should we enforce the user to implement `IRoleService` if that service will not be used? 

Additionally, the new `NoRoleUserService` we depend on `ISiteService` to determine if the user is the super-admin and load admin permissions. With this setup, it much more flexible and we use the services only when needed. Either way, the code is the same in all cases, the only different is where the claims come from.

src/OrchardCore.Modules/OrchardCore.Users/Services/NullRoleStore.cs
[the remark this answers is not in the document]
hishamco: Check the `NullLocalizer` that we implemented previously

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
```diff
@@ -396,15 +455,17 @@ public async Task AddToRoleAsync(IUser user, string normalizedRoleName, Cancella
                 throw new ArgumentNullException(nameof(user));
             }
 
-            var roleNames = await _roleService.GetRoleNamesAsync();
```
sebastienros, line 399: The same way you created a RoleStore that does nothing, you can create a default RoleService that does what is necessary when the Roles module is not enabled.

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
```diff
@@ -507,42 +590,60 @@ public Task RemoveLoginAsync(IUser user, string loginProvider, string providerKe
...
 
-        public Task<IList<Claim>> GetClaimsAsync(IUser user, CancellationToken cancellationToken)
+        public async Task<IList<Claim>> GetClaimsAsync(IUser user, CancellationToken cancellationToken)
         {
             if (user == null)
             {
                 throw new ArgumentNullException(nameof(user));
             }
 
-            return Task.FromResult<IList<Claim>>(((User)user).UserClaims.Select(x => x.ToClaim()).ToList());
```
sebastienros, line 533: Whatever was populating the claims in this collection should have a default implementation in the users module for the administrator role. It could also just assign ALL the permissions (not checking for Administrator).

Another is to have a custom handler (IAuthorizationHandler) that always allow any permission (only for the SuperUser), even if the Claims don't have permissions (empty list). If the implementation of RoleStore is the one from this module (NullRoleStore).

In case the Roles modules is inadvertently disabled, we need to only grant a permission to the SuperUser.

MikeAlhayek:
@sebastienros made the changes we discussed.

MikeAlhayek:
@jtkech any idea why there tests are only failing for ubuntu?

MikeAlhayek:
@Skrypt @hishamco you worked on these failing tests. Any idea why these tests are failing? Can you please help correcting the issue?

hishamco:
I tried to re-run the job but Cypress still failing due class name change, @MikeAlhayek could you try to rebase the `main` branch

MikeAlhayek:
@hishamco same issue

hishamco:
Nah, cypress is annoying ;)

MikeAlhayek:
@jptissot are you able to tell why the cypress tests are failing in this PR on ubuntu or maybe help fix the issue? Thank you

jtkech:
I will look at it when I will have time

src/OrchardCore/OrchardCore.Users.Core/Services/SuperAdminClaimsProviders.cs
```diff
@@ -0,0 +1,58 @@
...
+        var site = await _siteService.GetSiteSettingsAsync();
+
+        if (site.SuperUser != null && String.Equals(site.SuperUser, su.UserId))
+        {
+            var permissionCollections = await Task.WhenAll(_permissionProviders.Select(x => x.GetPermissionsAsync()));
+
+            var permissionClaims = permissionCollections.SelectMany(permissionCollection => permissionCollection)
+                .Select(permission => new Claim(Permission.ClaimType, permission.Name));
+
+            claims.AddClaims(permissionClaims);
```
jtkech, line 55: Not a complete review, just a first comment.

Yes the site service document is cached by the DocumentManager so that's okay.

Maybe replace `x => x.` by `provider => provider.`

I would not use `Task.WhenAll()` as we don't know what `GetPermissionsAsync()` may do.

src/OrchardCore/OrchardCore.Users.Core/Services/SuperAdminClaimsProviders.cs
```diff
@@ -0,0 +1,58 @@
...
+        _roleStore = _serviceProvider.GetService<IRoleStore<IRole>>();
+        _siteService ??= _serviceProvider.GetService<ISiteService>();
+        _permissionProviders ??= _serviceProvider.GetServices<IPermissionProvider>();
+
+        // NullRoleStore is only registered when Roles features is disabled.
+        // So we load all available claims for the super user only to grant them to all available permissions.
+        // These provider should only apply when NullRoleStore is registered.
+        if (
+            _roleStore is not NullRoleStore
+            || user is not User su
```
jtkech, line 38: Ah also, what does the term `su` mean?

src/OrchardCore/OrchardCore.Users.Core/Services/SuperAdminClaimsProviders.cs
MikeAlhayek: actually this class need to be removed. we already have permission handler that grant all permission to super admin.

jtkech: Okay then

src/OrchardCore/OrchardCore.Users.Core/Services/SuperAdminClaimsProviders.cs
jtkech: Maybe here `su` means `super user` but I saw it in many other places.

MikeAlhayek: yes `super user`... bad name

src/OrchardCore/OrchardCore.Users.Core/Services/SuperAdminClaimsProviders.cs
MikeAlhayek: removed this class. The SuperUserHandler already grants the super everything https://github.com/OrchardCMS/OrchardCore/blob/339d74b251b53393dd510f8ad3027aa17d78500d/src/OrchardCore.Modules/OrchardCore.Settings/Services/SuperUserHandler.cs

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
```diff
@@ -171,7 +172,12 @@ public Task<string> GetNormalizedUserNameAsync(IUser user, CancellationToken can
                 throw new ArgumentNullException(nameof(user));
             }
 
-            return Task.FromResult(((User)user).NormalizedUserName);
+            if (user is User su)
+            {
+                return Task.FromResult(su.NormalizedUserName);
```
jtkech, line 177: Okay, so here (and in other places) would need to replace `su` by another name.

Hmm, was it necessary to change all these tests and not use the casting anymore?
Here if user is not null, can it be something else than an `User`?

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
```diff
@@ -201,9 +212,12 @@ public Task SetNormalizedUserNameAsync(IUser user, string normalizedName, Cancel
                 throw new ArgumentNullException(nameof(user));
             }
 
-            ((User)user).NormalizedUserName = normalizedName;
+            if (user is User su)
+            {
+                return Task.FromResult(su.NormalizedUserName);
```
jtkech, line 217: Looks like here this is wrong as you don't set the `NormalizedUserName`.

Shoul be

            su.NormalizedUserName = normalizedName;

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
```diff
@@ -201,9 +212,12 @@ public Task SetNormalizedUserNameAsync(IUser user, string normalizedName, Cancel
...
+            }
 
-            return Task.CompletedTask;
+            return Task.FromResult<string>(null);
```
jtkech, line 220: And here should be

    return Task.CompletedTask;

I will make the change, hoping this is why the tests fails

Was it necessary to change this file?

jtkech:
@MikeAlhayek Okay all tests are now passing

MikeAlhayek:
@jtkech awesome! Feel free to many any changes you see fit. Hoping we can get some of these PR merged soon

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
MikeAlhayek: Any suggestion on name?

jtkech: Not for now ;)

Or just use the old casting, or keep the UserStore unchanged, but as you want.

Hmm, it changes a little the behavior in some places by not throwing if it is not a `User`.

Otherwise I will let you know if I find a better name ;)

Also saw one in `NotifyUserTaskActivity`.

MikeAlhayek: The reason there is change to that file is because there was an actual change... and I made the syntax change then. Now, I don't recall of a change, but since the formatting is much cleaner, I rather keep it. I'll think of another name

jtkech: Maybe


            if (user is User usr)
            {
                usr.NormalizedUserName = normalizedName;
            }

Or why not just


            if (user is User u)
            {
                u.NormalizedUserName = normalizedName;
            }

jtkech: > Now, I don't recall of a change

Would be good to retrieve it to review this actual change, only if possible

MikeAlhayek: okay. I made all the changes around the naming.

jtkech (approved):

src/OrchardCore.Modules/OrchardCore.Roles/Startup.cs
```diff
@@ -35,12 +35,13 @@ public Startup(IOptions<AdminOptions> adminOptions)
 
         public override void ConfigureServices(IServiceCollection services)
         {
+            services.AddScoped<RoleStore>();
+            services.TryAddScoped<IRoleClaimStore<IRole>>(sp => sp.GetRequiredService<RoleStore>());
+            services.AddScoped<IRoleStore<IRole>>(sp => sp.GetRequiredService<RoleStore>());
+
             services.TryAddScoped<RoleManager<IRole>>();
```
jtkech, line 42: If I follow the pattern here we are using some `AddScoped` and in `AddUsers()` we are using `TryAddScoped()`, but this implies that the Users startup should run after the Roles startup, see below.

Line 38 no problem as we are registering different implementations.

Line 39 I think you meant `AddScoped()` in place of `TryAddScoped()`, but I think that for line 39 and 40 this is not the right  pattern, see below.

Line 42 okay as we are trying to register in both places the same `RoleManager` implementation.

For line 38 and 39, because Users and Roles have no dependency on each other, **I think it works** because the Users startup runs last **but only because Users is after Roles in an alphabetic order**.

When we can use dependencies a dependent can just re-register a service because its startup runs after its dependency, more complex when services are resolved by `IEnumerable<T>` but there are helper extensions.

When there is no manifest dependencies we can tweak the Order property of the startup, eventually a new startup class dedicated to a set of services if we don't want to change the order of other services.

 In our case I think better to still use `TryAddScoped()` in `AddUsers()` to register the default implementations, and here use `Replace()` (maybe `ReplaceAll()` will see) to replace the defaults (less registrations) if already registered.

Another solution would be to register both `RoleStore` and `NullRoleStore` with the same new interface e.g. `IRoleAndClaimStore`, then the implementation factory delegates would resolve this new interface, so here we would only need to replace the registration of this new interface, no need to replace `IRoleClaimStore` and `IRoleStore`.

MikeAlhayek: @jtkech I am not aware of a `ReplaceAll()` method. But I can do this for a safer option

```
services.RemoveAll<IRoleClaimStore<IRole>>();
services.RemoveAll<IRoleStore<IRole>>();
services.AddScoped<IRoleClaimStore<IRole>>(sp => sp.GetRequiredService<RoleStore>());
services.AddScoped<IRoleStore<IRole>>(sp => sp.GetRequiredService<RoleStore>());
```
This way we always ensure that only one instance is registered when when roles is called.

MikeAlhayek: @jtkech I agree with you line 39 should have been `AddScoped<>` instead of `TryAddScoped<>`. I also agree that `Replace()` may be a safer option here. I made the change using Replace, If you see no additional issues, please merge it so you can complete test out #13040 with this new change.

```
services.TryAddScoped<IRoleClaimStore<IRole>>(sp => sp.GetRequiredService<RoleStore>());
            services.AddScoped<IRoleStore<IRole>>(sp => sp.GetRequiredService<RoleStore>());
```

jtkech: Okay, I'm just thinking about the new e.g. `IRoleAndClaimStore` interface suggestion that could be used to register both `NullRoleStore` and `RoleStore` and then used in the fatory delegates, so that we only need to replace this registration in the `OC.Roles` startup.

For example in `AddUsers()` somethig like this.

    s.TryAddScoped<IRoleAndClaimStore, NullRoleStore>(); // TryAddScoped() used here.
    s.AddScoped<IRoleClaimStore<IRole>>(sp => sp.GetRequiredService<IRoleAndClaimStore>());
    s.AddScoped<IRoleStore<IRole>>(sp => sp.GetRequiredService<IRoleAndClaimStore>());

Then in the `OC.Roles` startup only.

    s.Replace(ServiceDescriptor.Scoped<IRoleAndClaimStore, RoleStore>());

jtkech: Anyway, that"s okay like this, I will merge it ;)

PR #12426, opened 2022-09-15
Jint 3.0.0-beta-2041

lahma:
Latest Jint has slightly different API surface so tweaked integration to use it and cleaned up a bit.


src/OrchardCore/OrchardCore.Scripting.JavaScript/JavaScriptEngine.cs
```diff
@@ -32,37 +31,21 @@ public IScriptingScope CreateScope(IEnumerable<GlobalMethod> methods, IServicePr
 
         public object Evaluate(IScriptingScope scope, string script)
         {
-            if (scope == null)
+            static void ThrowInvalidScopeTypeException()
```
hishamco, line 34: Use `ArgumentException.ThrowIfNullOrEmpty()` instead

hishamco (changes requested):

src/OrchardCore/OrchardCore.Scripting.JavaScript/JavaScriptEngine.cs
lahma: Haven't heard about that one 😉  there's `ArgumentNullException.ThrowIfNull()` but that won't fit here as it's invalid type.

hishamco: Why it's invalid, FYI many exceptions now have static `Throw()` method, we already use them in quite different places in the code base

lahma: Well the original code checks that it has to be `JavaScriptScope`, if not, it's `ArgumentException`, didn't change the behavior here.

hishamco: The condition will still there, just we need to use the shortcut intead of creating new `ArgumentException` instance

lahma: I still don't understand...

src/OrchardCore/OrchardCore.Scripting.JavaScript/JavaScriptEngine.cs
```diff
@@ -32,37 +31,21 @@ public IScriptingScope CreateScope(IEnumerable<GlobalMethod> methods, IServicePr
...
             {
-                throw new ArgumentNullException(nameof(scope));
+                throw new ArgumentException($"Expected a scope of type {nameof(JavaScriptScope)}", nameof(scope));
```
hishamco, line 36:
```suggestion
                ArgumentNullException.ThrowIfNull(scope, nameof(scope));
```

hishamco (changes requested):

src/OrchardCore/OrchardCore.Scripting.JavaScript/JavaScriptEngine.cs
lahma: Cannot throw null exception as this is case of wrong type of argument. Misleading the same way as FileNotFoundException would be.

Skrypt (commented):
LGTM to me.

src/OrchardCore/OrchardCore.Scripting.JavaScript/JavaScriptEngine.cs
hishamco: What I'm suggested is the same of what you wrote but with the new static version, feel free to accept it or reject it

PR #12430, opened 2022-09-15
Data Provider Values, YesSql Options and Db Validator tweaks.

jtkech:
- Reuse strings for data provider values in place of an Enum so that we don't have to do `Enum.TryParse()` in one direction and an explicitly `.ToString()` in the other direction.

- Don't rely on defined Enum values for supported providers but on the registered providers.

- Ensure that `YesSqlOptions` has a non null `ITableNameConvention` and in sync with yesSql configuration, also allow to not have to register `ITableNameConvention` in the DI.

- We can define any `ShellSettings` value from the config stack, so in the setup better to not mutate `shellSettings["TablePrefixSeparator"]` from `_yesSqlOptions`, just use the options elsewhere. 

- In `DbConnectionValidator`also covers the case of `Sqlite` to check if the Document table exists.

- Update `IndexingTaskManager` to not use an hard coded `_` table prefix separator.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.AutoSetup/Options/TenantSetupOptions.cs
+++ b/src/OrchardCore.Modules/OrchardCore.AutoSetup/Options/TenantSetupOptions.cs
@@ -43,7 +43,7 @@ namespace OrchardCore.AutoSetup.Options
         /// <summary>
         /// Gets or sets the database provider.
         /// </summary>
-        public DatabaseProviderName DatabaseProvider { get; set; }
+        public string DatabaseProvider { get; set; }
 
         /// <summary>
         /// Gets or sets the database connection string.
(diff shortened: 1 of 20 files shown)


src/OrchardCore.Modules/OrchardCore.AutoSetup/Options/TenantSetupOptions.cs
```diff
@@ -43,7 +43,7 @@ public class TenantSetupOptions
         /// <summary>
         /// Gets or sets the database provider.
         /// </summary>
-        public DatabaseProviderName DatabaseProvider { get; set; }
+        public string DatabaseProvider { get; set; }
```
MikeAlhayek, line 46: Why did you change this back to string from Enum?

jtkech: See my first comment https://github.com/OrchardCMS/OrchardCore/pull/12430#issue-1375215225, I have the usage in Tenant Removal where using strings as before is simpler, then just added an helper providing const strings.

PR #12438, opened 2022-09-16
Update SlugService benchmarks

hishamco:


sebastienros:
The comment can be removed because it depends on the machine it runs on. I don't see any value there.
What can be done with the benchmark is run it before and after a change.

hishamco:
> The comment can be removed because it depends on the machine it runs on

Agree, but that's the pattern I saw for other benchmarks

hishamco:
Let us close if there's no value

hishamco:
Closed in favor of #15393

PR #12444, opened 2022-09-17
Add permission Localizer

### LINKED ISSUE #10551: Ability to localize permissions

MikeAlhayek:
Add permission localizer to allow us to localize the description of permission. 

This PR suggest that the new `IPermissionLocalizer` is used in the implementation of `IPermissionProvider` alternatively, we can just use it in the Roles controller to only translate these permission when we are getting ready to render them to the UI instead. This may be a better approach. However, not sure if this would work for the team since one may decide to localize the permission in the `PermissionProvider`. Either way, this `IPermissionLocalizer` should give us what we need to localize the permissions.

Fix #10551

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Roles/Permissions.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Roles/Permissions.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using OrchardCore.Localization;
 using OrchardCore.Security;
 using OrchardCore.Security.Permissions;
 using OrchardCore.Security.Services;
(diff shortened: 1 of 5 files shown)


hishamco:
The build is failed

hishamco:
BTW I though about `IPermissionLocalizer` before I submitted my PR #12442 but I decide to make things simple and easy to migrate in future to data localization, that's why I made it as draft, to welcome all the feedback before move on

So, it would be nice to comment in my PR if you don't mind

PR #12460, opened 2022-09-20
Place custom user settings in cards instead of tabs by default

### LINKED ISSUE #12453: Each Custom User settings should have its own form
Currently, when using `CustomUserSettings`, the custom settings is rendered in a new tab. However, all the tabs are part of the same html form. The problem with this setup is that a user must provide data in all the required fields "in all the tabs before saving". This make for a poor user experience.

In other words, if you have a user-custom-settings that captures user phone number and addresses. where at least one phone number is required, changing the content tab will not work because the user must update the phone number field in the custom tab.

Each custom settings, should have it's own form so when the save button is clicked, only the data in the current page is validated and updated.

MikeAlhayek:
Fix #12453 

## Screenshot after the PR
image

## Screenshot before
image

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Users/Drivers/CustomUserSettingsDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Drivers/CustomUserSettingsDisplayDriver.cs
@@ -57,7 +57,7 @@ namespace OrchardCore.Users.Drivers
                         var contentItem = await GetUserSettingsAsync(user, contentTypeDefinition, () => isNew = true);
                         model.Editor = await _contentItemDisplayManager.BuildEditorAsync(contentItem, context.Updater, isNew);
                     })
-                    .Location($"Content:10#{contentTypeDefinition.DisplayName}")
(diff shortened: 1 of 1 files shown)


ns8482e:
@MikeAlhayek IMO current users are used to with organizing custom settings with tabs

With is change you are forcing them write  placement to switch to current state

not a good user experience for upgrade

MikeAlhayek:
@ns8482e yes this would be a change. But, I believe that the default behavior provides bad user experience and should be changed. If users opted into using OC's defaults then they would see the new default changed. I don't think this is an issue.

ns8482e:
Change in default should always be opt-in/out-out not forceful

MikeAlhayek:
Yes. People choose to use use OC's default. and now we say we are changing OC's default. If we find flaws with default, we should be able to update accordingly otherwise we end up with some unusable functions. In this case it's usable but just provides poor user experience

ns8482e:
Btw, It's UI for `CustomUserSettings` that displays user added custom content item - something that is not available OOTB. Till 1.4 OC displays in tab but after this change it's forced to card ( for user's custom content item) is not good approach to take

PR #12467, opened 2022-09-21
CultureInfo should independent from local computer settings (Lombiq Technologies: OCORE-86)

### LINKED ISSUE #11228: CultureInfo should not depend on local computer settings
### Describe the bug
When the library sets the culture using `new CultureInfo(culture)` you can get unexpected results if this is also the system's culture and data formats were customized. This could be a problem if the server wants one data format for internal administrative use that's different from the culture's canonical format.

For example you have an en-US server with date format set to ISO 8601 (this is actually what we have on our CI - this problem is most painful while doing UI tests).  Then if you want to display a date in OC with the site set to American locale, it will display it using ISO 8601 which is not what you'd expect. The site's internationalization should not depend on the local format settings.

### To Reproduce
This repro assumes you have Windows 10 with en-US region settings:
image

Steps to reproduce the behavior on Windows:
1. Go to _/Admin/Settings/localization_ and make sure your site is en-US.
2. Create a new view that contains `@T["Date: {0:d}", new DateTime(2022, 02, 21, 12, 0)]`
3. Navigate to it, you will see `Date: 2/21/2022`.
4. Go to Start Menu > Settings > Time & Language > Region > Change data format
5. Change the _Short date_ to "2017-04-05".
6. Restart the server.
7. Visit the same page again, you will see `Date: 2022-02-21`.

### Expected behavior
You should still see `Date: 2/21/2022`.

### Speculation

I believe the problem can be solved simply by altering `OrchardCore.Localization.CultureScope.SetCultures` to generate fresh stock cultures. So from this:
```csharp
        private static void SetCultures(CultureInfo culture, CultureInfo uiCulture)
        {
            CultureInfo.CurrentCulture = culture;
            CultureInfo.CurrentUICulture = uiCulture;
        }
```
Into this:
```csharp
        private static void SetCultures(CultureInfo culture, CultureInfo uiCulture)
        {
            CultureInfo.CurrentCulture = new CultureInfo(culture.Name, useUserOverride: false);
            CultureInfo.CurrentUICulture = new CultureInfo(uiCulture.Name, useUserOverride: false);
        }
```

hishamco:
Fixes #11228


hishamco:
Seems I need to rename `LocalizationOptions` to avoid naming collision

src/OrchardCore.Modules/OrchardCore.Localization/CultureLocalizationOptions.cs
```diff
@@ -0,0 +1,6 @@
+namespace OrchardCore.Localization;
+
+public class CultureLocalizationOptions
```
hishamco, line 3: We could rename it to `OrchardCoreLocalizationOptions` naming is hard ;)

src/OrchardCore.Modules/OrchardCore.Localization/CultureSettings.cs
```diff
@@ -0,0 +1,7 @@
+namespace OrchardCore.Localization;
+
+public enum CultureSettings
```
hishamco, line 3: Could be moved to `OC.Abstraction` if we want to add it to `ILocalizationService`

src/OrchardCore/OrchardCore.Abstractions/Localization/CultureScope.cs
```diff
@@ -1,45 +1,55 @@
...
-        }
+    private static void SetCultures(CultureInfo culture, CultureInfo uiCulture)
+    {
+        CultureInfo.CurrentCulture = culture;
+        CultureInfo.CurrentUICulture = uiCulture;
+    }
+
+    private static CultureScope CreateInternal(string culture, string uiCulture)
+    {
+        var useUserSelectedCultureSettings = ShellScope.Services.GetService<ILocalizationService>()
```
hishamco, line 48: @jtkech is there a better way than using `ShellScope.Services`?

jtkech: That's okay for me to resolve a service like this but only when we are sure it is always used in the context of a valid ShellScope.

Then here you have to use .GetAwaiter().GetResult() which is not good, you would need to add an async `CultureScope.CreateAsync()`.

That said here I would have thought about an additional parameter `useUserOverride` that would be passed by the consumer. Also not sure the option should come from admin settings, maybe better to use a configurable IOptions.

hishamco: > but only when we are sure it is always used in the context of a valid ShellScope

How can I verify?

> That said here I would have thought about an additional parameter useUserOverride that would be passed by the consumer. Also not sure the option should come from admin settings, maybe better to use a configurable IOptions.

If you look to the commits history I already used `IOptions<T>` then I move into `ILocalizationService`

Did you think I need to follow the same pattern that I did in `CultureScopeTests` in other broken tests?

@Piedone any though coz `CultureScope` doesn't accept constructor injection

hishamco: > but only when we are sure it is always used in the context of a valid ShellScope

How can I verify?

> That said here I would have thought about an additional parameter useUserOverride that would be passed by the consumer. Also not sure the option should come from admin settings, maybe better to use a configurable IOptions.

If you look to the commits history I already used `IOptions<T>` then I move into `ILocalizationService`

Did you think I need to follow the same pattern that I did in `CultureScopeTests` in other broken tests?

@Piedone any though coz `CultureScope` doesn't accept constructor injection

hishamco: > That said here I would have thought about an additional parameter useUserOverride that would be passed by the consumer

This is could be a simple option, but I was thinking about centeral place that fetch the culture settings either from `appsettings.json` or site settings

Piedone: Yep, I'd use `IOptions` too.

hishamco: If you want to go with `IOptions` as I did at the beginning, I will rollback from the latest changes

src/OrchardCore.Modules/OrchardCore.Localization/Models/LocalizationSettings.cs
```diff
@@ -1,31 +1,36 @@
 using System.Globalization;
 
-namespace OrchardCore.Localization.Models
+namespace OrchardCore.Localization.Models;
```
sebastienros, line 3: Please don't update existing files with this because it's breaking the diff in the PR and we can't see where the differences are. Only do it for new files.

hishamco:
@sebastienros I just use `CultureSettings` enum to make clear, but I never mind to replace it with a boolean value, one more thing I just moved `CultureLocalizationOptions` into `OC.Localization.Abstractions` all the options that we are using are placed in their modules, but while this kind of option could be use in any localization implementation I think it's fine to place it in `OC.Localization.Abstractions` rather than `OC.Abstractions`. Another thing I suggest to use the system culture instead of user-selected culture by default, similar to what `CultureInfo` is using

If sound looks good I will remove everything I added related to `ILocalizationService`

jtkech:
@hishamco

Take care, about `CultureInfo` the user is not an OC user but an user at the machine level, that's why we were talking about System settings.

There are 2 ctors `CultureInfo(string name)` and `CultureInfo(string name, bool useUserOverride)`, the first one is the same as using `CultureInfo(name, useUserOverride: true)`, so if we assume that the default is when using the first ctor, the default is then `useUserOverride = true`.

So the 2 defined values of the current enum `Default, User` is confusing because of the above. Also, I have nothing against using enum but not when it is to retrieve a value to be used as a bool.

So to not be confused with the "user" we could name this option field `UseSystemSettings` whose default would be true, as I remember @sebastienros suggested `IgnoreSystemSettings` but whose default value would be false and we would need to pass `useUserOverride = !option.IgnoreSystemSettings`.

Not sure we need to bind this IOptions to the configuration (e.g. from appsettins.json), if so and if we bind this field individually, if we use `UseSystemSettings` we would need to pass to the `GetValue()` a default value of true, if we bind the whole IOptions class (as we do many time) we just need to specify in the IOptions class a default value of true for this field.

Or if we use `IgnoreSystemSettings`, as said above, we would just need to pass `useUserOverride = !option.IgnoreSystemSettings`

Maybe better to name this IOptions `CultureOptions` and define it in `OC.Abstractions`, where we define `CultureScope`, for example this IOptions would be used each time we use `CultureScope`.

hishamco:
> Take care, about CultureInfo the user is not an OC user but an user at the machine level, that's why we were talking about System settings.

Sure

> There are 2 ctors CultureInfo(string name) and CultureInfo(string name, bool useUserOverride), the first one is the same as using CultureInfo(name, useUserOverride: true), so if we assume that the default is when using the first ctor, the default is then useUserOverride = true.

Oops, seem I made a mistake here, I presumed it's a default culture by default

> Or if we use IgnoreSystemSettings, as said above, we would just need to pass useUserOverride = !option.IgnoreSystemSettings

I never mind with this option as I mentioned above

> Maybe better to name this IOptions CultureOptions and define it in OC.Abstractions, where we define CultureScope, for example this IOptions would be used each time we use CultureScope.

It was there before, but while all the modules adding the options in module project and `CultureScope` is using `CultureOptions` that's why I move it into `OC.Localization.Abstractions`

jtkech (commented):
Mainly params names/logic, otherwise LGTM

src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
```diff
@@ -397,7 +397,10 @@ private async Task GenerateContainerPathFromPatternAsync(AutoroutePart part)
                 _contentManager ??= _serviceProvider.GetRequiredService<IContentManager>();
 
                 var cultureAspect = await _contentManager.PopulateAspectAsync(part.ContentItem, new CultureAspect());
-                using (CultureScope.Create(cultureAspect.Culture))
+
+                var cultureOptions = _serviceProvider.GetService<IOptions<CultureOptions>>().Value;
+
+                using (CultureScope.Create(cultureAspect.Culture, useUserOverride: !cultureOptions.IgnoreSystemSettings))
```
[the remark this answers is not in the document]
hishamco: >  and let the logic be applied at the lowest level in CultureScope and the new extensions helpers.

Do you mean passing the `CultureOption` to the `CultureScope`?

src/OrchardCore.Modules/OrchardCore.Localization/Startup.cs
```diff
@@ -35,6 +43,8 @@ public override void ConfigureServices(IServiceCollection services)
                 AddDataAnnotationsPortableObjectLocalization();
 
             services.Replace(ServiceDescriptor.Singleton<ILocalizationFileLocationProvider, ModularPoFileLocationProvider>());
+
+            services.Configure<CultureOptions>(_shellConfiguration.GetSection("OrchardCore_Localization"));
```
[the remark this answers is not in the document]
hishamco: This way the `CultureOptions` will be there even the localization module is disable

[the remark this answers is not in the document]
hishamco: And that's why it was fine when I use it in `ILocalizationService` before

src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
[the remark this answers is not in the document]
jtkech: If for the options we don't use `UseUserOverride` but `IgnoreSystemSettings`, this to not be confused with the term `User` and to have a default value of `false`, what I meant is to use everywhere params named `IgnoreSystemSettings`.

So here in place of

    using (CultureScope.Create(cultureAspect.Culture, useUserOverride: !cultureOptions.IgnoreSystemSettings))

Would be

    using (CultureScope.Create(cultureAspect.Culture, ignoreSystemSettings: cultureOptions.IgnoreSystemSettings))

So I suggested to apply the logic, here I mean using the `!` operator, only at the lowest level where we use `new CultureInfo()`, here I mean for example in `CultureScope.Create()` and in the new extensions, here I'm reffering to `AddSupportedCultures()` and `AddSupportedUICultures()` that would have the same renamed param `ignoreSystemSettings` in place of `useUserOverride`.

Passing the whole `CultureOptions` is another good idea.

src/OrchardCore.Modules/OrchardCore.Localization/Startup.cs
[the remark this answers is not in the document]
jtkech: No because the above startup only run if the `OC.Localization` module is enabled and this is the one that also registers `LocalizationService`, in `AddDefaultServices()` we only register `DefaultLocalizationService`.

Not sure what you wanted to do in `ILocalizationService` but would need to be in both `LocalizationService` and `DefaultLocalizationService`. And if you need to inject `IOptions<CultureOptions>` in both of them, then same problem, here `services.Configure<CultureOptions>()` is only registered if `OC.Localization` is enabled.

But maybe I didn't understand what you meant.

src/OrchardCore/OrchardCore.Abstractions/Localization/CultureScope.cs
```diff
@@ -3,34 +3,69 @@
...
         private readonly CultureInfo _originalCulture;
         private readonly CultureInfo _originalUICulture;
 
-        private CultureScope(CultureInfo culture, CultureInfo uiCulture)
+        private CultureScope(string culture, string uiCulture, bool useUserOverride)
         {
-            Culture = culture;
-            UICulture = uiCulture;
+            Culture = new CultureInfo(culture, useUserOverride);
+            UICulture = new CultureInfo(uiCulture, useUserOverride);
```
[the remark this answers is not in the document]
jtkech: As said in an above comment

> If for the options we don't use UseUserOverride but IgnoreSystemSettings, this to not be confused with the term User and to have a default value of false, what I meant is to use everywhere params named IgnoreSystemSettings.

So also in the below Create method and then here in this static ctor where you will apply the logic, meaning using the `!` operator, at the lowest level when using `new CultureInfo()` for which we can't change the existing param name.

        private CultureScope(string culture, string uiCulture, bool ignoreSystemSettings)
        {
            Culture = new CultureInfo(culture, useUserOverride: !ignoreSystemSettings)
            UICulture = new CultureInfo(uiCulture, useUserOverride: !ignoreSystemSettings)
            ...

        public static CultureScope Create(CultureInfo culture, bool ignoreSystemSettings = false)
            => Create(culture, culture, ignoreSystemSettings);

src/OrchardCore.Modules/OrchardCore.Autoroute/Handlers/AutoroutePartHandler.cs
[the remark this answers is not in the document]
hishamco: I'm still prefering `UseSystemCulture` or `UserSelectedCultureSetttings` rather than `IgnoreSystemSettings`

> Passing the whole `CultureOptions` is another good idea.

I think passing boolean value is much much better, to make the API simple and easy for testing

jtkech:
@hishamco 

If we assume that `IgnoreSystemSettings` is a good name for the configurable option, because it doesn't contain the term `User` which may be confusing as here it is not an OC user, and because it has a default value of false, here I was just suggesting to use this good name as much as possible, also for the names of method params, for example `AnyMethod(... , goodName: options.GoodName)`.

But no problem, was only a simple naming suggestion, otherwise LGTM.

The only thing is that if we provide a concrete value e.g. in appsettings.json, do we want it to be taken into account even if `OC.Localization` is not enabled? If no that's okay, if yes the config binding should not be kept in the `OC.Localization` startup.

hishamco:
Agree with you @jtkech as I said before, but using the name `UseSystemCultureSettings` or `UseSystemSettings` with default value `true` will be much better if all agree

> do we want it to be taken into account even if OC.Localization is not enabled? If no that's okay, if yes the config binding should not be kept in the OC.Localization startup.

Agree, but let me do some tests and update the PR

src/OrchardCore/OrchardCore/Modules/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -126,6 +126,15 @@ private static void AddDefaultServices(OrchardCoreBuilder builder)
...
                 s.AddSingleton<IDistributedLock>(sp => sp.GetRequiredService<LocalLock>());
+                s.AddSingleton<CultureOptions>(sp =>
+                {
+                    var shellConfiguration = sp.GetService<IShellConfiguration>();
+                    var cultureOptions = shellConfiguration
+                        .GetSection("OrchardCore_Localization_CultureOptions")
+                        .Get<CultureOptions>();
+
+                    return cultureOptions;
+                });
```
jtkech, line 137: `builder.ConfigureServices()` is like a module startup in which we can inject services that are already registered as `ShellSettings` and `IShellConfiguration`, this even if we are building a tenant container.

In case of `builder.ConfigureServices()` you can pass to the delegate an `IServiceProvider` parameter that you can use to resolve `IShellConfiguration` and then use one of the regular `s.Configure<CultureOptions>()`.

            builder.ConfigureServices((s, serviceProvider) =>
            {
                s.AddSingleton<LocalLock>();
                s.AddSingleton<ILocalLock>(sp => sp.GetRequiredService<LocalLock>());
                s.AddSingleton<IDistributedLock>(sp => sp.GetRequiredService<LocalLock>());

                var configuration = serviceProvider.GetRequiredService<IShellConfiguration>();
                s.Configure<CultureOptions>(configuration.GetSection("OrchardCore_Localization_CultureOptions"));
            });

Note: I could have also named it `sp` in place of `serviceProvider`, that's would be okay as other `sp` are used in their own closure (if we click go to definition we can see that it makes the distinction), so I used `serviceProvider` just to make a naming distinction, but you can also use `sp`, as you want.

jtkech (commented):
Just one suggested change, otherwise LGTM

src/OrchardCore/OrchardCore/Modules/Extensions/ServiceCollectionExtensions.cs
hishamco: I'm not sure why I didn't call `Configuure()` instead?!! May be I was sleepy ;)

hishamco: Seems I didn't notice the second overload

src/OrchardCore/OrchardCore.Abstractions/Localization/CultureScope.cs
```diff
@@ -3,34 +3,69 @@
 
 namespace OrchardCore.Localization
 {
+    /// <summary>
+    ///
```
hishamco, line 7: Seems I need to complete the docs

jtkech:
@hishamco Did you tried it? If so I will approve ;)

hishamco:
I just asked you in case if you have anything else, I will add a missing docs then do a final test after latest commit

hishamco:
@jtkech seems we forgot to apply the same settings in `SetDefaultCulture()`, one more thing that I was did it right but you confused me ;) may be because the naming

`UseSystemSettings` != `useUserOverride`
`IgnoreSystemSettings` == `useUserOverride`

For that I remove the negation as it was and everything is fine

CultureInfo

hishamco:
I'm think to do a last tweak for the added extension by adding `OrchardCoreRequestLocalizationOptions` for doing the heavy lifting, then we can use as the following:

```csharp
var requestLocalizationOptions = serviceProvider.GetService<IOptions<OrchardCoreRequestLocalizationOptions>>().Value;

requestLocalizationOptions
    .SetDefaultCulture(defaultCulture)
    .AddSupportedCultures(supportedCultures)
    .AddSupportedUICultures(supportedCultures);

app.UseRequestLocalization(requestLocalizationOptions);
```

jtkech:
@hishamco 

> @jtkech seems we forgot to apply the same settings in SetDefaultCulture()

Yes as I mentioned here https://github.com/OrchardCMS/OrchardCore/issues/11228#issuecomment-1105795875

> one more thing that I was did it right but you confused me ;) may be because the naming

Just saw that you now use `new CultureInfo(culture, ignoreSystemSettings)` which is wrong as by default when just using `new CultureInfo(culture)` it doesn't ignore system settings, so if it was not working it would mean that something else is wrong. But never sure at 100%, will look at it tonight.

jtkech:
Yes I know https://github.com/dotnet/runtime/blob/5481c4bfe6ced9000c05692163a019739c912677/src/libraries/System.Private.CoreLib/src/System/Globalization/CultureInfo.cs#L158-L162

> Because the useUserOverride is true, in our case if ignoreSystemSettings is true it should ignore the default settings

Because the `useUserOverride` is true by default, meaning it will use system settings, and because `ignoreSystemSettings` is false by default (if not configured/overridden), we should use `new CultureInfo(culture, useUserOverride: !ignoreSystemSettings)`, so that it works as before if the option value is not configured (e.g. in appsettings) and not overridden by code (e.g. in a s.Configure<>()).

So if `IgnoreSystemSettings` is configured/overridden to true (which is not the default) the result will be `new CultureInfo(culture, useUserOverride: false)` which is right as here we want to ignore system settings, so we don't want to use user override (meaning we don't want to use system settings).

hishamco:
Believe it or not the `IgnoreSystemSettings is confusing especially we you read the `CultureInfo(string bool)` docs ;)

hishamco:
I need to do a final quick test again

jtkech:
> Believe it or not the IgnoreSystemSettings is confusing especially we you read the CultureInfo(string, bool) docs ;)

Yes I agree ;)

sebastienros (approved):

hishamco:
Everythings looks as expected

jtkech (approved):

src/OrchardCore/OrchardCore.Abstractions/Localization/CultureOptions.cs
```diff
@@ -0,0 +1,12 @@
+namespace OrchardCore.Localization;
+
+/// <summary>
+/// Represents a programmable options for localization culture.
+/// </summary>
+public class CultureOptions
+{
+    /// <summary>
+    /// Gets or sets whether to ignore the system culture settings or not.
```
Piedone, line 9: Link to the issue in the summary for context, otherwise it won't be apparent what we're talking about and why we have the option.

Same in `CultureScope`.

src/OrchardCore/OrchardCore.Localization.Abstractions/OrchardCoreRequestLocalizationOptions.cs
```diff
@@ -0,0 +1,68 @@
...
+{
+    private readonly bool _useUserOverride;
+
+    /// <summary>
+    /// Creates a new <see cref="OrchardCoreRequestLocalizationOptions"/> with adefault values.
+    /// </summary>
+    public OrchardCoreRequestLocalizationOptions() : this(ignoreSystemSettings: false)
+    {
+
+    }
```
Piedone, line 21:
```suggestion
    {
    }
```

src/OrchardCore/OrchardCore.Localization.Abstractions/OrchardCoreRequestLocalizationOptions.cs
```diff
@@ -0,0 +1,68 @@
...
+
+/// <summary>
+/// Represents an options for the <see cref="RequestLocalizationMiddleware"/> that extends <see cref="RequestLocalizationOptions"/>.
+/// </summary>
+public class OrchardCoreRequestLocalizationOptions : RequestLocalizationOptions
+{
+    private readonly bool _useUserOverride;
+
+    /// <summary>
+    /// Creates a new <see cref="OrchardCoreRequestLocalizationOptions"/> with adefault values.
```
Piedone, line 16:
```suggestion
    /// Creates a new <see cref="OrchardCoreRequestLocalizationOptions"/> with default values.
```

src/OrchardCore/OrchardCore.Localization.Abstractions/OrchardCoreRequestLocalizationOptions.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// </summary>
+    public OrchardCoreRequestLocalizationOptions() : this(ignoreSystemSettings: false)
+    {
+
+    }
+
+    /// <summary>
+    /// Creates a new <see cref="OrchardCoreRequestLocalizationOptions"/> with adefault values and ability to ignore system settings.
```
Piedone, line 24:
```suggestion
    /// Creates a new <see cref="OrchardCoreRequestLocalizationOptions"/> with default values and ability to ignore system settings.
```

Piedone (changes requested):
Nice work!

src/OrchardCore/OrchardCore.Localization.Abstractions/OrchardCoreRequestLocalizationOptions.cs
hishamco: Nah, typo ;)

src/OrchardCore/OrchardCore.Abstractions/Localization/CultureOptions.cs
hishamco: We didn't add that frequently, may be we could add extra explanation in remarks

Piedone: That's also good, but an issue link lets people see the full context.

src/OrchardCore/OrchardCore.Abstractions/Localization/CultureScope.cs
```diff
@@ -6,6 +6,7 @@ namespace OrchardCore.Localization
     /// <summary>
     /// Represents a scope that you can change the current culture within.
     /// </summary>
+    /// <remarks>The scope disallow the current culture depends on local computer settings by default. For more information refer to https://github.com/OrchardCMS/OrchardCore/issues/11228</remarks>
```
Piedone, line 9: This is a really long line, add a line break in between. Same for CultureOptions.

src/OrchardCore/OrchardCore.Localization.Abstractions/OrchardCoreRequestLocalizationOptions.cs
Piedone: Did you apply this change? Can't see it.

Piedone (changes requested):

Piedone (approved):

hishamco:
FYI I created a PR in aspnet repo while I'm working on this feature, finally it got merge. I might need to open a draft PR to change the API with the one that will be built-in in .NET 8.0

https://github.com/dotnet/aspnetcore/pull/44282

Piedone:
Great!

alafi:
I believe this PR introduced this issue #12610

hishamco:
Thanks @alafi I will check the related issue

PR #12476, opened 2022-09-22
Change the border color of the drop down filter buttons

### LINKED ISSUE #12475: Change the border color of the drop down filter buttons
Change the border color of the drop down filter buttons

Here is how it currently look
image

Change it so it looks like this instead

image

MikeAlhayek:
Fix #12475

## Before screenshot
image
image

## After screenshot
Change it so it looks like this instead
image

image


Skrypt (approved):

hishamco:
It looks good but how it looks in dark mode?

MikeAlhayek:
@hishamco good point. I fixed it in the dark theme too

hishamco:
Thanks Majd, LGTM

src/OrchardCore.Themes/TheAdmin/Assets/scss/modules/theme/darkmode/_forms.scss
```diff
@@ -0,0 +1,8 @@
+
+.has-filter .btn:not(.show) {
+    border-color: #3C4144 !important;
+}
+
+.has-filter .form-control-feedback, .has-search .form-control-feedback {
```
hishamco, line 6: Can this be global for both theme why the value is same?!!

hishamco (approved):
Looks good except a little comment

src/OrchardCore.Themes/TheAdmin/Assets/scss/modules/theme/darkmode/_forms.scss
MikeAlhayek: The value is currently the same, but also could be changed. So since it a color related, it should be part of the theme files. I was hoping for a variable to use here but did not find one that I can use. I am going to make one more change soon so that the icon color in the same too on the dark theme

PR #12480, opened 2022-09-23
Fixes #12479: Set HttpContext before using ShellScope in Unit Tests

### LINKED ISSUE #12479: Unit Test reporting "Fail" in CI
image

The current CI reports issues when running Unit Tests. See the latest Workflow actions to troubleshoot.
I'm not sure why as of now but the Unit Tests are not set to a failing state afterward either.

jtkech:
Fixes #12479 

The failing report is there for a while, it doesn't break the unit tests as the exception is catched in our invoke extension, the error is only logged. For now what I saw from the trace is on a content manager loading an item.

    Content item loading...
    ...
    HtmlBodyPartHandler => IShortcodeService =>
    IShortcodeProvider => AssetUrlShortcodeProvider  =>
    IOptions<ResourceManagementOptions> =>
    ResourceManagementOptionsConfiguration =>
    NRE  on _pathBase = httpContextAccessor.HttpContext.Request.PathBase;

So I'm adding null operators step by step for testing

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Resources/ResourceManagementOptionsConfiguration.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Resources/ResourceManagementOptionsConfiguration.cs
@@ -25,7 +25,7 @@ namespace OrchardCore.Resources
         {
             _resourceOptions = resourceOptions.Value;
             _env = env;
-            _pathBase = httpContextAccessor.HttpContext?.Request.PathBase ?? "";
+            _pathBase = httpContextAccessor.HttpContext.Request.PathBase;
         }
 
         ResourceManifest BuildManifest()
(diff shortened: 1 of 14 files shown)


jtkech:
Okay it removes the fail report, I will remove null operator step by step.

jtkech:
Okay so this is the `HttpContext` that is null.

Looks like that in unit tests we are getting items => item loading => part coordinator, part handlers including `HtmlBodyPartHandler`, and then the DI injections as described above until the NRE.

Time to sleep, will see tomorrow ;)

jtkech:
Okay some unit tests call `LuceneIndexingService.ProcessContentItemsAsync()` which uses `contentManager.GetAsync()`, and then fail on the null `HttpContext` described above.

So in unit tests we would need to set an `HttpContext` each time we use `shellScope.UsingAsync()` as we do for example in background tasks, but for now I only did it where it was failing.

jtkech:
- Okay, so for `Unit Tests` I opted to make an `UsingTenantScopeAsync()` method that ensures that an `HttpContext` is set (as in a background task) and then calls `ShellScope.UsingAsync()` as before.

- The above method is defined in the `SiteContext` base class where I also defined the `ShellHost` property in place of having to define it in each derived classes.

Skrypt (approved):

PR #12483, opened 2022-09-23
GeoPointField translations

agriffard:
Surround Latitude and Longitude with `@T[""]` in GeoPointField editors.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Spatial/Views/GeoPointField-Leaflet.Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Spatial/Views/GeoPointField-Leaflet.Edit.cshtml
@@ -8,7 +8,7 @@
 
 <div class="form-row" id="GeoPointFields">
     <div class="mb-3 col-md-4 col-lg-3">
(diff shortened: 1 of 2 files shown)


Skrypt (approved):

hishamco (approved):

hishamco:
I'm suggesting to create a TagHelper that can easily localize labels, spans .. etc instead of using T[""] all the time

/cc @sebastienros

agriffard:
As I see it, `@T[""]` is already kind of minimal and it is what is detected by PoExtractor to generate the translations.

hishamco:
Agree, but I'm thinking about a way that localize views without PoExtractor, something framework specific. I remember Damian Edwards suggested that in ASP.NET Core long time ago, I think it will be useful to provide such thing especially for some HTML elements such as label, span, paragraph .. etc

hishamco:
Agree, but I'm thinking about a way that localize views without PoExtractor, something framework specific. I remember Damian Edwards suggested that in ASP.NET Core long time ago, I think it will be useful to provide such thing especially for some HTML elements such as label, span, paragraph .. etc

PR #12489, opened 2022-09-23
Fix content types fields dependency

### LINKED ISSUE #12488: When using ContentTypes without any fields, the "Fields" section should not show up
### Describe the bug
When we enable `OC.ContentTypes` feature but not enable any feature that compose fields like `OC.ContentFields`, the Fields section shows up.

image

### To Reproduce
1. Enable OC.ContentTypes and disable OC.ContentFields.
2. Go to Content Types and edit or add a new content type

### Suggestion
I think we should remove the "Add field" button when there is no fields definitions. Also, if a currently attached field has no definition, we should not be able to edit it, but we should be able to remove it. This can happen if a field was attach to a content type/part while the field definition is available but then the feature that provides the definition was disabled. 

Moreover, we should remove the "Fields" section from the UI if the current content-type has no fields and also there is no field definitions.

MikeAlhayek:
Fix #12488

@jtkech  I am not a fan of the following line but not sure if there is a better way since in the `AdminController` we are using standard ViewModel approach instead of drivers. Can you think of a better way to do the same thing?

```
if (ShellDescriptor.Features.Any(x => x.Id.Equals("OrchardCore.ContentFields", StringComparison.OrdinalIgnoreCase)))
```


hishamco:
You remind me by a discussion that I opened awhile ago asking about simple way to allow show / hide some UI based on feature status

MikeAlhayek:
@hishamco i think the right answer for that is to use DisplayDrives to decouple your features. In the ContentTypes UI, it’s fine to use viewModels as we are doing today. It also may make sense to let ContentTypes to depend on ContentFields using the Mainifest file since these two are directly related. Either way, the way I approached this issue is by not have to depend on field “in case someone wants to not use fields for some reason”.

the only other way I can think to do this is to check if a specific service is registered.  If this is acceptable, may be we add extension to the the FeaturesManager or ShellDescripter to check if a given feature is in enabled or not. But I don’t think it’s a good idea to use this approach everywhere. Because we don’t want others to follow not best practices

jtkech:
> ... asking about simple way to allow show / hide some UI based on feature status

@hishamco Currently `OC.ContentFields` defines most of the fields but not all, for example the `MarkDownField`. So here I think it is still not a good thing to do

@MikeAlhayek Yes `ContentTypes` allows to manage types composed of parts and fields defined in other modules and should not depend on any of them.

Hmm, I looked at the related issue, if it is just to not show `Fields` section with an `Add Field` button, maybe just use `_contentDefinitionService.GetFields()` in some places to see if there is no field definition at all.

MikeAlhayek:
@jtkech great point! I'll revise the PR.

MikeAlhayek:
I replaced this PR with #12492

Thank for your help @jtkech

hishamco:
Agree with @jtkech suggestion, @MikeAlhayek I'm not sure why you didn't update the PR rather than open a new one

PR #12347, opened 2022-09-06
Add wrapper to app content fields

MikeAlhayek:
Fix #12341

Also, fixed spacing issues in some editors.

Updated how the Link is rendered
image

PR #12495, opened 2022-09-24
Fix field wrappers

MikeAlhayek:
In PR #12347 I added css class wrapper to content fields to allow us to use JS to hide/show fields. However, the class name we used for the wrapper is not unique since the same name can be used on multiple parts. To make it unique, I am adding part name to make it unique. This PR makes the class name as `field-wrapper-partname-fieldname` instead of `field-wrapper-fieldname`


MikeAlhayek:
@sebastienros the names were updated based on the feedback from triage

PR #12498, opened 2022-09-25
Fix CulturePickerOptions sample and doc

jtkech:
Only change the sample and doc with the section name used in the code

https://github.com/OrchardCMS/OrchardCore/blob/34b920d63f10a6b93b1da4325f80fa811c11523a/src/OrchardCore.Modules/OrchardCore.ContentLocalization/Startup.cs#L91


hishamco (approved):

jtkech:
I merged it as it only update a commented out section name and the related one in the doc.

PR #12510, opened 2022-09-26
 Move role creation into the recipe instead of feature activation

### LINKED ISSUE #12425: Move role creation into the recipe instead of feature activation
Currently, we create roles based on the registered Permission Providers when a feature is installed. This approach has the following two disadvantages 
 1. There is no way to control which roles to add when the site is setup. Some admins don't want to `Contributor`, `Moderator` or any other role. In today's setup, they can't opt-out of these roles even if they don't want them.
 2. If the site admin delete a role using the UI, this role could come back up when he/she installs the next feature. If an admin delete a role, I don't think we should add it back up just because we enabled a new feature. This also create confusion (like why this role is back?!)

IMO, if an admin choose to opt-out of having a role added to their website because they are handling permission grouping differently, then we should not be forcing them into using some default permission.

Instead of adding roles during feature setup, add the role should be part of the setup recipes. However, we keep today's logic for the permissions. So when the site is first setup, we first execute the [RolesStep](https://github.com/OrchardCMS/OrchardCore/blob/339d74b251b53393dd510f8ad3027aa17d78500d/src/OrchardCore.Modules/OrchardCore.Roles/Recipes/RolesStep.cs) to load the needed role names. Then in the [RoleUpdater](https://github.com/OrchardCMS/OrchardCore/blob/339d74b251b53393dd510f8ad3027aa17d78500d/src/OrchardCore.Modules/OrchardCore.Roles/Services/RoleUpdater.cs#L83-L84), we skip creating the role when the roles does not exists.

We may need to add a new `int Order` property to `IRecipeStepHandler` to ensure that `RolesStep` in migrated before [FeatureStep](https://github.com/OrchardCMS/OrchardCore/blob/4192c145724186c442715414f16d98e883c4b0b1/src/OrchardCore.Modules/OrchardCore.Features/Recipes/Executors/FeatureStep.cs) since [FeatureStep](https://github.com/OrchardCMS/OrchardCore/blob/4192c145724186c442715414f16d98e883c4b0b1/src/OrchardCore.Modules/OrchardCore.Features/Recipes/Executors/FeatureStep.cs) triggers that events used by [RoleUpdater](https://github.com/OrchardCMS/OrchardCore/blob/339d74b251b53393dd510f8ad3027aa17d78500d/src/OrchardCore.Modules/OrchardCore.Roles/Services/RoleUpdater.cs#L83-L84).

MikeAlhayek:
Fix #12425 

@jtkech please check this out when your time permits since we chatted about this in the issue.

Here is a summary of the enhancements this PR brings
1. We can now use recipe step to explicitly list the roles we want to use for the site. We are no longer forced to use the default roles.
2. If a role is deleted, it wont come back at some point after installing new features. Currently, even if we delete a role, it could come back after installing more features.
3. If the Roles feature is disabled, and re-enabled later, we keep track of permissions history to automatically assign any permission that would not have been assigned previously while the Roles feature is disabled.
4. If an enabled feature introduces new permissions, it'll now get assigned to the proper roles by default "if these roles exists".

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Roles/Services/RoleUpdater.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Roles/Services/RoleUpdater.cs
@@ -117,10 +117,7 @@ namespace OrchardCore.Roles.Services
                     .Distinct()
                     .ToList();
 
-                if (!rolesDocument.PermissionGroups.ContainsKey(group.RoleName))
-                {
-                    rolesDocument.PermissionGroups.TryAdd(group.RoleName, new List<string>());
-                }
+                rolesDocument.PermissionGroups.TryAdd(group.RoleName, new List<string>());
 
                 // Get all available permission names that isn't already assigned to the role.
                 var additionalPermissionNames = distinctPermissionNames.Except(currentPermissionNames);
--- a/src/OrchardCore.Modules/OrchardCore.Roles/Services/RoleUpdater.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Roles/Services/RoleUpdater.cs
@@ -1,41 +1,50 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Security.Claims;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Identity;
 using Microsoft.Extensions.Logging;
-using OrchardCore.Environment.Extensions;
+using OrchardCore.Documents;
 using OrchardCore.Environment.Extensions.Features;
 using OrchardCore.Environment.Shell;
+using OrchardCore.Modules;
+using OrchardCore.Roles.Models;
 using OrchardCore.Security;
 using OrchardCore.Security.Permissions;
+using OrchardCore.Security.Services;
 
 namespace OrchardCore.Roles.Services
 {
-    public class RoleUpdater : IFeatureEventHandler
+    public class RoleUpdater : ModularTenantEvents, IFeatureEventHandler
     {
         private readonly RoleManager<IRole> _roleManager;
         private readonly IEnumerable<IPermissionProvider> _permissionProviders;
-        private readonly ITypeFeatureProvider _typeFeatureProvider;
         private readonly ILogger _logger;
+        private readonly IRoleService _roleService;
+        private readonly IDocumentManager<RolesDocument> _rolesDocumentManager;
+
+        private bool _updateInProgress;
 
         public RoleUpdater(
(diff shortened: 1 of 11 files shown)


src/OrchardCore.Modules/OrchardCore.Roles/Startup.cs
```diff
@@ -42,6 +42,7 @@ public override void ConfigureServices(IServiceCollection services)
             services.AddRecipeExecutionStep<RolesStep>();
 
             services.AddScoped<IFeatureEventHandler, RoleUpdater>();
+            services.AddScoped<IModularTenantEvents, RoleUpdater>();
```
sebastienros, line 45: This would create a second instance. You can register `<RoleUpdater>` directly as scoped, then `<IFeatureEventHandler>` and `<IModularTenantEvents>` as `services.AddScoped<...>(sp => sp.GetRequiredService<RoleUpdater>)

MikeAlhayek: @sebastienros updated

MikeAlhayek:
@sebastienros can we review this one again?

MikeAlhayek:
@sebastienros this is ready for a quick review.

src/OrchardCore.Modules/OrchardCore.Roles/Services/RoleUpdater.cs
```diff
@@ -47,68 +56,109 @@ public class RoleUpdater : IFeatureEventHandler
...
+
+                var distinctPermissionNames = currentPermissionNames
+                    .Union(group.PermissionNames)
+                    .Distinct()
+                    .ToList();
+
+                if (!rolesDocument.PermissionGroups.ContainsKey(group.RoleName))
                 {
-                    _logger.LogDebug("No default roles for feature '{FeatureName}'", feature.Id);
+                    rolesDocument.PermissionGroups.TryAdd(group.RoleName, new List<string>());
```
sebastienros, line 122: If you want to replace th List then use `[group.RoleName] = ...` and if you don't want to, keep using `TryAdd` and remove the `ContainsKey` usage

sebastienros (approved):

gvkries:
This PR breaks the permissions feature. Manually removed permissions are added back when the site restarts.

MikeAlhayek:
@gvkries please open a new issue with steps to reproduce the issue. You can reference this PR in the new issue.

hyzx86:
@MikeAlhayek   here: https://github.com/OrchardCMS/OrchardCore/issues/13024

PR #12516, opened 2022-09-27
Prevent ManageOwnMedia from implying ManageMedia

### LINKED ISSUE #12057: Forced effective Manage Media permission
### Issue

Checking **Manage own media** permission for a user implicitly grants that user to **Manage media** which results in his ability to delete other user's media which is unwanted

### Expected behaviour

Isolate the own media management permission from the management of all media.

sebastienros:
Fixes #12057

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Media/Permissions.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media/Permissions.cs
@@ -10,7 +10,7 @@ namespace OrchardCore.Media
         public static readonly Permission ManageMediaFolder = new Permission("ManageMediaFolder", "Manage All Media Folders");
         public static readonly Permission ManageOthersMedia = new Permission("ManageOthersMediaContent", "Manage Media For Others", new[] { ManageMediaFolder });
         public static readonly Permission ManageOwnMedia = new Permission("ManageOwnMediaContent", "Manage Own Media", new[] { ManageOthersMedia });
-        public static readonly Permission ManageMedia = new Permission("ManageMediaContent", "Manage Media", new[] { ManageOwnMedia });
+        public static readonly Permission ManageMedia = new Permission("ManageMediaContent", "Manage Media");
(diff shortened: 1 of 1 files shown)


jtkech:
I will try it if I have time because if I look at the following code, looks like `ManageMedia` is always required to get any folder, and that `ManageOwnMedia` is an additional permission to allow an user to have its own private folder (under `_User/`), but not sure, would need to be tried.

Here `ManageMedia` is checked.

https://github.com/OrchardCMS/OrchardCore/blob/af13b02c502fa1dc16be1c533ab061406ab3dc0b/src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs#L64-L69

Then in the same controller action method `ManageOwnMedia` is checked.

https://github.com/OrchardCMS/OrchardCore/blob/af13b02c502fa1dc16be1c533ab061406ab3dc0b/src/OrchardCore.Modules/OrchardCore.Media/Controllers/AdminController.cs#L81-L86

hishamco:
I already assigned the related issue to me and starting checking the current permissions, some of them are confusing ;) I'm not sure if Seb checked this after pushing the PR?

jtkech:
@hishamco Just for info after doing some quick tests.

So maybe the confusion is that `ManageOwnMedia` allows to add "private" medias for a given user but under a dedicated `_user/{theUserId}` media subfolder. In the other folders any media added by any user can be seen by any other user having the `ManageMedia` permission. `ManageMedia` only means that you can see the related admin menu item without saying that you can do everything on media folders. So `ManageOwnMedia` needs `ManageMedia` just to be able to navigate in medias.

I tried it, for example User1 can see the `_user/{theUser1Id}` subfolder but not `_user/{theUser2Id}`, but other subfolders can be seen by both users. Note: For testing for both User1 and User2 I used the `Author` role that only has the `ManageOwnMedia` by default. And the `Administrator` was able to see "private" medias of both User1 and User2.

So not saying it is good, but maybe it is already working as intended.

sebastienros:
@jtkech do you think the GH issue is not an actual issue but a missunderstanding of what the permissions represent? And there is no way to actually not give access to all media if we want to manage per-user ones?

jtkech:
@sebastienros Yes

PR #12517, opened 2022-09-28
Fix Create button size in ListPart

MikeAlhayek:


hishamco (approved):

hishamco:
Could you please recheck if we are using small or normal buttons in other views, otherwise we should merge this

MikeAlhayek:
All using standard buttons. Even the button that is next to this one is different in size 
image

hishamco:
What I mean previously is if this consist to use normal buttons instead of small one, I will merge this otherwise we should revise other views

PR #12519, opened 2022-09-28
Release the shell context when site settings are updated

### LINKED ISSUE #12518: Changing page size and other site settings from the UI has no affect
### Describe the bug
While the app is running and a user change site settings like "Page Size", the new value does not take effect unless we manually reload/release the shell context.

### To Reproduce
1. Go to general site settings
2. Change the page size from 10 to other value like 1
3. try to list content, you'll still see 10 records per page.

MikeAlhayek:
Fix #12518

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Settings/Drivers/DefaultSiteSettingsDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Settings/Drivers/DefaultSiteSettingsDisplayDriver.cs
@@ -3,6 +3,7 @@ using System.Threading.Tasks;
 using Microsoft.Extensions.Localization;
 using OrchardCore.DisplayManagement.Handlers;
 using OrchardCore.DisplayManagement.Views;
+using OrchardCore.Environment.Shell;
 using OrchardCore.Mvc.ModelBinding;
 using OrchardCore.Settings.ViewModels;
 
@@ -13,10 +14,16 @@ namespace OrchardCore.Settings.Drivers
         public const string GroupId = "general";
 
         private readonly IStringLocalizer S;
+        private readonly IShellHost _shellHost;
+        private readonly ShellSettings _shellSettings;
 
-        public DefaultSiteSettingsDisplayDriver(IStringLocalizer<DefaultSiteSettingsDisplayDriver> stringLocalizer)
+        public DefaultSiteSettingsDisplayDriver(IStringLocalizer<DefaultSiteSettingsDisplayDriver> stringLocalizer,
+            IShellHost shellHost,
+            ShellSettings shellSettings)
         {
             S = stringLocalizer;
+            _shellHost = shellHost;
+            _shellSettings = shellSettings;
         }
 
         public override IDisplayResult Edit(ISite site)
@@ -69,6 +76,11 @@ namespace OrchardCore.Settings.Drivers
                 if (!String.IsNullOrEmpty(site.BaseUrl) && !Uri.TryCreate(site.BaseUrl, UriKind.Absolute, out _))
                 {
                     context.Updater.ModelState.AddModelError(Prefix, nameof(model.BaseUrl), S["The Base url must be a fully qualified URL."]);
+                }
+
+                if (context.Updater.ModelState.IsValid)
+                {
+                    await _shellHost.ReleaseShellContextAsync(_shellSettings);
                 }
             }
 


src/OrchardCore.Modules/OrchardCore.Settings/Drivers/DefaultSiteSettingsDisplayDriver.cs
```diff
@@ -70,6 +77,11 @@ public override async Task<IDisplayResult> UpdateAsync(ISite site, UpdateEditorC
                 {
                     context.Updater.ModelState.AddModelError(Prefix, nameof(model.BaseUrl), S["The Base url must be a fully qualified URL."]);
                 }
+
+                if (context.Updater.ModelState.IsValid)
+                {
+                    await _shellHost.ReleaseShellContextAsync(_shellSettings);
```
hishamco, line 83: We always did that in Post action in the controller

hishamco, line 83: We always did that in Post action in the controller

hishamco (changes requested):

src/OrchardCore.Modules/OrchardCore.Settings/Drivers/DefaultSiteSettingsDisplayDriver.cs
MikeAlhayek: I don't know about that.. here are some examples were we do that in driver

https://github.com/OrchardCMS/OrchardCore/blob/bbe1b14fa448b9653f0523bdc40267af77489325/src/OrchardCore.Modules/OrchardCore.Security/Drivers/SecuritySettingsDisplayDriver.cs#L95

https://github.com/OrchardCMS/OrchardCore/blob/339d74b251b53393dd510f8ad3027aa17d78500d/src/OrchardCore.Modules/OrchardCore.Recipes/Services/RecipeDeploymentTargetHandler.cs#L51

https://github.com/OrchardCMS/OrchardCore/blob/339d74b251b53393dd510f8ad3027aa17d78500d/src/OrchardCore.Modules/OrchardCore.ReCaptcha/Drivers/ReCaptchaSectionDisplayDriver.cs#L71

hishamco: Ok, we can check

hishamco: I remember we did it in Security module when the there's no Controller present, but in case of `OC.Settings` there's a controller if I'm not wrong

MikeAlhayek: yes you are right. but for site settings, we group settings by groupId. If we use the controller, then we force the app to always release shell for all group. Not sure if we want that behavior. In the driver, we only release context only if we pass validation and the group is `general`.

May be we can add `ReleaseShellContextFeature` to the request pipeline in the driver and in the controller we can check to see if `ReleaseShellContextFeature` exists in the pipeline before we release.  @jtkech thoughts?

MikeAlhayek: @hishamco I just checked some settings driver, and we seems to be release the shell context directly from the driver. here is a list of some of the settings drivers that release the shellContext

SmtpSettingsDisplayDriver
FacebookLoginSettingsDisplayDriver
GitHubAuthenticationSettingsDisplayDriver
GoogleAuthenticationSettingsDisplayDriver
HttpsSettingsDisplayDriver
LocalizationSettingsDisplayDriver
AzureADSettingsDisplayDriver
OpenIdClientSettingsDisplayDriver
ReCaptchaSettingsDisplayDriver
ReverseProxySettingsDisplayDriver

jtkech:
Yes all these drivers are ISite Section Drivers, they collaborate to the ISite settings, only those that have settings that needs a shell rebuild call `ReleaseShellContextAsync()`.

The Default Site Settings Driver didn't need to do this until now, one solution would be to use a snapshot IOptions for pager oprions to be evaluated on each scope as I remember, but maybe not so good to force the call of `GetSiteSettings().GetAwaiter()...` on each scope, humm but it is a cached document.

Need to be tried but I think that if you call `ReleaseShellContextAsync()` in the Default driver, we would not need anymore to do it all other section drivers. But then any settings change would trigger a shell release.

MikeAlhayek:
@jtkech yeah I get it. The question is why would we release the shellContext everytime a site settings is saved. We should be able to control when we release. currently we control from each driver.

jtkech:
> We should be able to control when we release. currently we control from each driver.

Yes from each ISite section driver, which was good because no default settings needed a shell rebuild, so no more the case because of the new PagerOptions.

What we could do is to still call `ReleaseShellContextAsync()` in the default driver, but only if a setting that need a rebuild has changed, so for now as I remember only if `PageSize` has changed.

sebastienros:
> I don't know about that.. here are some examples were we do that in driver

but these drivers update methods are called from a POST

src/OrchardCore.Modules/OrchardCore.Settings/Drivers/DefaultSiteSettingsDisplayDriver.cs
sebastienros: I think that's ok to call `ReleaseShellContextAsync` multiple times if we can ensure that subsequent calls in the same request are no-op.

sebastienros (approved):

jtkech:
@sebastienros @MikeAlhayek 

`ReleaseShellContextAsync()` is not totally a no-op at the shellHost level. But that's okay as we check `GroupId` in any driver including the default one. So here we can release the shell, and we can keep the call to `ReleaseShellContextAsync()` in section drivers that was using it.

For example if a section driver don't want to release the shell, it will not be released by the default driver as before because the `GroupId` being not `general`, it will not do any update (and not release the shell).

What is funny is that in the general settings page (for default site settings) it is mentioned.

> The current tenant will be reloaded when the settings are saved.

Which currently is not true.

The only suggestion would be to separate in the default settings the ones that need to release the shell, like a sub section (currently only PageSize), hmm but then not easy to manage the above message ;)

MikeAlhayek:
@jtkech do you want me to change it where we only release the context only if the PageSize change or do you want to leave it as it? Either way, we can all agree that it's safe to call the release from the driver as this PR does

jtkech:
No that's okay because `ReleaseShellContextAsync()` will be called once if it is called in the If block checking the `GroupId`, as it is done in section drivers and here for the default driver.

Yes checking individual setting could be useful but maybe confusing if we add another setting in the future, and as said what message should we show "The tenant may be released" ;)

Would need a separate section but not worth to do only for `PageSize`, that's your fault ;)

So that's okay for me.

jtkech (approved):

PR #12521, opened 2022-09-28
Add Profile Editor and Display mode for ListPart

### LINKED ISSUE #12520: Add Profile editor/display more for ListPart
There is a need to extend the look of the default to a profile like. 

Here is an example of how the default profile look like
image

And I can override the Profile header to look something like this

image

Add new "Profile" editor for ListPart where we have a header.

MikeAlhayek:
Fix #12520 

Here are some screenshots of the default look for the Profile editor

image

## Add a new contained item
image

## Editing existing contained item
image

## Adding new contained item
image

## Edit the container itself "the profile"
image


src/OrchardCore.Modules/OrchardCore.Lists/Views/ProfileHeader.cshtml
```diff
@@ -0,0 +1,11 @@
+@model OrchardCore.Lists.ViewModels.ProfileHeaderViewModel
+@inject OrchardCore.ContentManagement.Display.IContentItemDisplayManager ContentItemDisplayManager
+@inject OrchardCore.DisplayManagement.ModelBinding.IUpdateModelAccessor UpdateModelAccessor
+@{
+    var profileShape = await ContentItemDisplayManager.BuildDisplayAsync(Model.ContainerContentItem, UpdateModelAccessor.ModelUpdater, "Profile");
+}
+<div class="card bg-light mb-3">
```
agriffard, line 7: Does `bg-light` work on dark mode?

MikeAlhayek: @agriffard i change that so it works on the dark mode as well.

PR #12528, opened 2022-09-30
Update YesSQL to 3.2.0

hishamco:
https://github.com/sebastienros/yessql/releases/tag/v3.1.0

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Search.Lucene/Migrations.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Search.Lucene/Migrations.cs
@@ -15,10 +15,12 @@ namespace OrchardCore.Search.Lucene
 {
     public class Migrations : DataMigration
     {
+        private readonly IConfiguration _yesSqlConfiguration;
         private readonly IContentDefinitionManager _contentDefinitionManager;
 
-        public Migrations(IContentDefinitionManager contentDefinitionManager)
+        public Migrations(IConfiguration yesSqlConfiguration, IContentDefinitionManager contentDefinitionManager)
         {
+            _yesSqlConfiguration = yesSqlConfiguration;
             _contentDefinitionManager = contentDefinitionManager;
         }
 
@@ -175,27 +177,27 @@ namespace OrchardCore.Search.Lucene
                                 logger.LogDebug("Updating Lucene indices settings and queries");
                             }
 
-                            var updateCmd = $"UPDATE {dialect.QuoteForTableName(table)} SET Content = REPLACE(content, '\"$type\":\"OrchardCore.Lucene.LuceneQuery, OrchardCore.Lucene\"', '\"$type\":\"OrchardCore.Search.Lucene.LuceneQuery, OrchardCore.Search.Lucene\"') WHERE [Type] = 'OrchardCore.Queries.Services.QueriesDocument, OrchardCore.Queries'";
+                            var updateCmd = $"UPDATE {dialect.QuoteForTableName(table, _yesSqlConfiguration.Schema)} SET Content = REPLACE(content, '\"$type\":\"OrchardCore.Lucene.LuceneQuery, OrchardCore.Lucene\"', '\"$type\":\"OrchardCore.Search.Lucene.LuceneQuery, OrchardCore.Search.Lucene\"') WHERE [Type] = 'OrchardCore.Queries.Services.QueriesDocument, OrchardCore.Queries'";
 
                             await transaction.Connection.ExecuteAsync(updateCmd, null, transaction);
                             
-                            updateCmd = $"UPDATE {dialect.QuoteForTableName(table)} SET Content = REPLACE(content, '\"$type\":\"OrchardCore.Lucene.Deployment.LuceneIndexDeploymentStep, OrchardCore.Lucene\"', '\"$type\":\"OrchardCore.Search.Lucene.Deployment.LuceneIndexDeploymentStep, OrchardCore.Search.Lucene\"') WHERE [Type] = 'OrchardCore.Deployment.DeploymentPlan, OrchardCore.Deployment.Abstractions'";
+                            updateCmd = $"UPDATE {dialect.QuoteForTableName(table, _yesSqlConfiguration.Schema)} SET Content = REPLACE(content, '\"$type\":\"OrchardCore.Lucene.Deployment.LuceneIndexDeploymentStep, OrchardCore.Lucene\"', '\"$type\":\"OrchardCore.Search.Lucene.Deployment.LuceneIndexDeploymentStep, OrchardCore.Search.Lucene\"') WHERE [Type] = 'OrchardCore.Deployment.DeploymentPlan, OrchardCore.Deployment.Abstractions'";
 
                             await transaction.Connection.ExecuteAsync(updateCmd, null, transaction);
 
-                            updateCmd = $"UPDATE {dialect.QuoteForTableName(table)} SET Content = REPLACE(content, '\"$type\":\"OrchardCore.Lucene.Deployment.LuceneSettingsDeploymentStep, OrchardCore.Lucene\"', '\"$type\":\"OrchardCore.Search.Lucene.Deployment.LuceneSettingsDeploymentStep, OrchardCore.Search.Lucene\"') WHERE [Type] = 'OrchardCore.Deployment.DeploymentPlan, OrchardCore.Deployment.Abstractions'";
+                            updateCmd = $"UPDATE {dialect.QuoteForTableName(table, _yesSqlConfiguration.Schema)} SET Content = REPLACE(content, '\"$type\":\"OrchardCore.Lucene.Deployment.LuceneSettingsDeploymentStep, OrchardCore.Lucene\"', '\"$type\":\"OrchardCore.Search.Lucene.Deployment.LuceneSettingsDeploymentStep, OrchardCore.Search.Lucene\"') WHERE [Type] = 'OrchardCore.Deployment.DeploymentPlan, OrchardCore.Deployment.Abstractions'";
 
                             await transaction.Connection.ExecuteAsync(updateCmd, null, transaction);
 
-                            updateCmd = $"UPDATE {dialect.QuoteForTableName(table)} SET Content = REPLACE(content, '\"$type\":\"OrchardCore.Lucene.Deployment.LuceneIndexResetDeploymentStep, OrchardCore.Lucene\"', '\"$type\":\"OrchardCore.Search.Lucene.Deployment.LuceneIndexResetDeploymentStep, OrchardCore.Search.Lucene\"') WHERE [Type] = 'OrchardCore.Deployment.DeploymentPlan, OrchardCore.Deployment.Abstractions'";
+                            updateCmd = $"UPDATE {dialect.QuoteForTableName(table, _yesSqlConfiguration.Schema)} SET Content = REPLACE(content, '\"$type\":\"OrchardCore.Lucene.Deployment.LuceneIndexResetDeploymentStep, OrchardCore.Lucene\"', '\"$type\":\"OrchardCore.Search.Lucene.Deployment.LuceneIndexResetDeploymentStep, OrchardCore.Search.Lucene\"') WHERE [Type] = 'OrchardCore.Deployment.DeploymentPlan, OrchardCore.Deployment.Abstractions'";
 
                             await transaction.Connection.ExecuteAsync(updateCmd, null, transaction);
 
-                            updateCmd = $"UPDATE {dialect.QuoteForTableName(table)} SET Content = REPLACE(content, '\"$type\":\"OrchardCore.Lucene.Deployment.LuceneIndexRebuildDeploymentStep, OrchardCore.Lucene\"', '\"$type\":\"OrchardCore.Search.Lucene.Deployment.LuceneIndexRebuildDeploymentStep, OrchardCore.Search.Lucene\"') WHERE [Type] = 'OrchardCore.Deployment.DeploymentPlan, OrchardCore.Deployment.Abstractions'";
+                            updateCmd = $"UPDATE {dialect.QuoteForTableName(table, _yesSqlConfiguration.Schema)} SET Content = REPLACE(content, '\"$type\":\"OrchardCore.Lucene.Deployment.LuceneIndexRebuildDeploymentStep, OrchardCore.Lucene\"', '\"$type\":\"OrchardCore.Search.Lucene.Deployment.LuceneIndexRebuildDeploymentStep, OrchardCore.Search.Lucene\"') WHERE [Type] = 'OrchardCore.Deployment.DeploymentPlan, OrchardCore.Deployment.Abstractions'";
 
                             await transaction.Connection.ExecuteAsync(updateCmd, null, transaction);
 
-                            updateCmd = $"UPDATE {dialect.QuoteForTableName(table)} SET [Type] = 'OrchardCore.Search.Lucene.Model.LuceneIndexSettingsDocument, OrchardCore.Search.Lucene' WHERE [Type] = 'OrchardCore.Lucene.Model.LuceneIndexSettingsDocument, OrchardCore.Lucene'";
+                            updateCmd = $"UPDATE {dialect.QuoteForTableName(table, _yesSqlConfiguration.Schema)} SET [Type] = 'OrchardCore.Search.Lucene.Model.LuceneIndexSettingsDocument, OrchardCore.Search.Lucene' WHERE [Type] = 'OrchardCore.Lucene.Model.LuceneIndexSettingsDocument, OrchardCore.Lucene'";
 
                             await transaction.Connection.ExecuteAsync(updateCmd, null, transaction);
 
--- a/src/OrchardCore.Modules/OrchardCore.Search.Lucene/Migrations.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Search.Lucene/Migrations.cs
@@ -15,12 +15,10 @@ namespace OrchardCore.Search.Lucene
 {
     public class Migrations : DataMigration
     {
-        private readonly IConfiguration _yesSqlConfiguration;
         private readonly IContentDefinitionManager _contentDefinitionManager;
 
-        public Migrations(IConfiguration yesSqlConfiguration, IContentDefinitionManager contentDefinitionManager)
+        public Migrations(IContentDefinitionManager contentDefinitionManager)
         {
-            _yesSqlConfiguration = yesSqlConfiguration;
             _contentDefinitionManager = contentDefinitionManager;
         }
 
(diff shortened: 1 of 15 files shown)


hishamco:
Seems there's a breaking changes

hishamco:
There's a `schema` parameter has been added

sebastienros:
I might do another update to fix the breaking changes, defaulting to the previous values, and also setting `UseLegacyIdentityColumn` to true by default. Then only do these breaking changes in a major version.

sebastienros:
Also please don't merge without my approval.

hishamco:
Sure, I will not merge until we fix the breaking chnages or using the default schema, also after your approval

sebastienros:
Please use Configuration.Schema when the schema is requested (or null)

hishamco:
Is there a new update released?

sebastienros:
I just pushed 3.2.0, should appear soon.

hishamco:
I was totally busy last few days, but I need to fix it soon

sebastienros:
@MikeAlhayek FYI since you have commit rights now, you can pull Hisham's branch and update it if needed. You can even change the files directly from GH.

MikeAlhayek:
@sebastienros I pinged @hishamco privately. He should have this PR updated with in an hour

hishamco:
Seb, I will update the PR ASAP, since I'm working on other testing API. BTW GH allowed editing the changed files AFAIK, but I need to change other files

hishamco:
Seb, I will update the PR ASAP, since I'm working on other testing API. BTW GH allowed editing the changed files AFAIK, but I need to change other files

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
```diff
@@ -145,7 +148,7 @@ private ISqlBuilder GetSelectBuilderForDocumentTable(string tablePrefix, string
 
         selectBuilder.Select();
         selectBuilder.Selector("*");
-        selectBuilder.Table(_yesSqlOptions.TableNameConvention.GetDocumentTable());
+        selectBuilder.Table(_yesSqlOptions.TableNameConvention.GetDocumentTable(), alias: null, _yesSqlConfiguration.Schema);
```
MikeAlhayek, line 151: Remove ` IOptions<YesSqlOptions> yesSqlOptions,` from being injected and use the following instead

```suggestion
        selectBuilder.Table(_yesSqlConfiguration.TableNameConvention.GetDocumentTable(), alias: null, _yesSqlConfiguration.Schema);
```

src/OrchardCore/OrchardCore.Data.YesSql/OrchardCoreBuilderExtensions.cs
```diff
@@ -166,6 +166,7 @@ private static IConfiguration GetStoreConfiguration(IServiceProvider sp, YesSqlO
                 QueryGatingEnabled = yesSqlOptions.QueryGatingEnabled,
                 ContentSerializer = new PoolingJsonContentSerializer(sp.GetService<ArrayPool<char>>()),
                 TableNameConvention = yesSqlOptions.TableNameConvention,
+                IdentityColumnSize = yesSqlOptions.UseLegacyIdentityColumn ? IdentityColumnSize.Int32 : IdentityColumnSize.Int64
```
MikeAlhayek, line 169: I would suggest you use the same enum used by YesSql `IdentityColumnSize` instead of using a boolean `UseLegacyIdentityColumn`  maybe at somepoint YesSql will support other enum values like string.

hishamco: It's column size not type, I think it's fine to stay with a boolean value, but I never mind to use the `IdentityColumnSize `

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
hishamco: I think we still need `YesSqlOptions` for table prefix separator

hishamco:
@jtkech I expecting `YesSql.IConfiguration` is registered by default, am I right or shall I add the registeration myself to make the tests pass?

src/OrchardCore.Modules/OrchardCore.Search.Lucene/Migrations.cs
```diff
@@ -175,27 +177,27 @@ public int Create()
                                 logger.LogDebug("Updating Lucene indices settings and queries");
                             }
 
-                            var updateCmd = $"UPDATE {dialect.QuoteForTableName(table)} SET Content = REPLACE(content, '\"$type\":\"OrchardCore.Lucene.LuceneQuery, OrchardCore.Lucene\"', '\"$type\":\"OrchardCore.Search.Lucene.LuceneQuery, OrchardCore.Search.Lucene\"') WHERE [Type] = 'OrchardCore.Queries.Services.QueriesDocument, OrchardCore.Queries'";
+                            var updateCmd = $"UPDATE {dialect.QuoteForTableName(table, _yesSqlConfiguration.Schema)} SET Content = REPLACE(content, '\"$type\":\"OrchardCore.Lucene.LuceneQuery, OrchardCore.Lucene\"', '\"$type\":\"OrchardCore.Search.Lucene.LuceneQuery, OrchardCore.Search.Lucene\"') WHERE [Type] = 'OrchardCore.Queries.Services.QueriesDocument, OrchardCore.Queries'";
```
sebastienros, line 180: the schema can be accessed through `session.Store.Configuration.Schema` the same way `tablePrefix` is used a few lines above.

src/OrchardCore/OrchardCore.ContentManagement.GraphQL/Queries/Predicates/PredicateQuery.cs
```diff
@@ -121,15 +127,15 @@ public string GetColumnName(string propertyPath)
                     {
                         // Switch the given alias in the path with the mapped alias.
                         // aliasPart.alias -> AliasPartIndex.Alias
-                        return Dialect.QuoteForTableName($"{tableAlias}") + "." + Dialect.QuoteForColumnName(columnName);
+                        return Dialect.QuoteForTableName($"{tableAlias}", _yesSqlConfiguration.Schema) + "." + Dialect.QuoteForColumnName(columnName);
```
sebastienros, line 130: If `tablePrefix` is not used here, it's probably because this is just an alias, and then schema can/should be null.

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
MikeAlhayek: yeah true. but I would read the `TableNameConvention` from the `IConfiguration` instead

src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/YesSqlOptions.cs
```diff
@@ -19,4 +19,6 @@ public class YesSqlOptions
     public IContentSerializer ContentSerializer { get; set; }
 
     public string TablePrefixSeparator { get; set; } = "_";
+
+    public bool UseLegacyIdentityColumn { get; set; } = true;
```
sebastienros, line 23: I think that where @MikeAlhayek 's suggestion might apply. Store an enum, or the Type of the identity column. Today it is int32/int63 but maybe we'll support Guid someday.

src/OrchardCore/OrchardCore.Data.YesSql/OrchardCoreBuilderExtensions.cs
MikeAlhayek: yes but it is an enum value which can have more than 2 values. Maybe it does not have the best name in YesSql but it should also be an enum in OC not bool.

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
```diff
@@ -26,17 +26,20 @@ public class DbConnectionValidator : IDbConnectionValidator
...
 
     private readonly IEnumerable<DatabaseProvider> _databaseProviders;
+    private readonly IConfiguration _yesSqlConfiguration;
     private readonly YesSqlOptions _yesSqlOptions;
     private readonly SqliteOptions _sqliteOptions;
     private readonly ShellOptions _shellOptions;
 
     public DbConnectionValidator(
         IEnumerable<DatabaseProvider> databaseProviders,
+        IConfiguration yesSqlConfiguration,
```
sebastienros, line 36: Weird to need both "YesSqlOptions" and "YeslSqlConfiguration"

src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/YesSqlOptions.cs
MikeAlhayek: Exactly. We should not use bool in OC and should use the same Enum used in YesSql.

test/OrchardCore.Tests/Orchard.Queries/SqlParserTests.cs
```diff
@@ -31,7 +32,7 @@ private string FormatSql(string sql)
         [InlineData("select distinct a", "SELECT DISTINCT [a];")]
         public void ShouldParseSelectClause(string sql, string expectedSql)
         {
-            var result = SqlParser.TryParse(sql, _defaultDialect, _defaultTablePrefix, null, out var rawQuery, out var messages);
+            var result = SqlParser.TryParse(sql, _configuration, _defaultDialect, _defaultTablePrefix, null, out var rawQuery, out var messages);
```
sebastienros, line 35: Pass the schema directly, this is what is done for dialect and prefix.

sebastienros: And then the test can verify that the schema is used in the result, like we check the prefix.

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
MikeAlhayek: `YesSqlOptions` has the option [TablePrefixSeparator ](https://github.com/OrchardCMS/OrchardCore/blob/8d9ced24fe9c815ccbc309cdf2a48db37db6a1aa/src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/YesSqlOptions.cs#L21) which isn't used in `YesSql`. This is the reason both are needed. I think we should use `IConfiguration` to get the currently configured services instead of using `YesSqlOptions` since `YesSqlOptions` could have null configurations like a null value for `TableNameConvention`

MikeAlhayek:
> @jtkech I expecting `YesSql.IConfiguration` is registered by default, am I right or shall I add the registeration myself to make the tests pass?

You can inject `ISession` and assess it `_session.Store.Configuration` instead of injecting `YesSql.IConfiguration` directly

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
jtkech: I may be wrong but `YesSqlOptions` provides options to configure the yesSql store configuration.

So why not just usin the `IStore` to retrieve what is actually used.

From memory, I did a PR to ensure that `TableNameConvention` is in sync

                // Ensure a non null 'TableNameConvention' to be always used for `YesSql.Configuration` and then in sync with it.
                services.PostConfigure<YesSqlOptions>(o => o.TableNameConvention ??= new YesSql.Configuration().TableNameConvention);

For `TablePrefixSeparator` it has a default value, and it is used indirectly by YesSql through the TablePrefix.

                    if (!String.IsNullOrWhiteSpace(shellSettings["TablePrefix"]))
                    {
                        var tablePrefix = shellSettings["TablePrefix"].Trim() + yesSqlOptions.TablePrefixSeparator;
                        storeConfiguration = storeConfiguration.SetTablePrefix(tablePrefix);
                    }

MikeAlhayek: Yes `TableNameConvention` should be retrieved from the `ISession.Store.Configuration` not from `YesSqlOptions`. But we should still use `YesSqlOptions` to determine the prefix-separator that the tenant is using to determine if the table exists or not

jtkech: I don't agree, normally we could use the table prefix configured on yesSql side, it should have already use the same configured separator on OC side (or its default value `_`).

@hishamco for your question I saw somewhere, yesSql IConfiguration is not registered in the DI.

jtkech: In summary as I can see ATM, `YesSqlOptions` should be only used to configure the store, then we can always retrieve values from the store.Configuration, unless there is no store registered yet, as we use for the tenant validator.

jtkech: Oops, and tenant validator uses `IDbConnectionValidator` so looks like here the `IStore` can't be injected, and that all need to retrived from `YesSqlOptions` that may need to be completed ;)

MikeAlhayek: @jtkech I don't think the prefix table separator is a configurable value in YesSql. AFAIK, you can't retrieve it from YesSql. I think you can get the prefix from YesSql which is not tableSeperator. TableSeperator is something configurable is OC and could be different from one tenant to another. However, in the DbConnectionValidator we assume that the connection being validated will use the currently configured table separator.

jtkech: > I don't think the prefix table separator is a configurable value in YesSql

Yes, I was just meaning that the yesSql store config TablePrefix includes it, but if here we are in the Default tenant to validate the connection of a new tenant, we have to use the tablePrefix param passed to the `ValidateAsync()` and then yes the separator from the YesSql options.

But here we will get the table separator from the YesSqlOptions related to the Default tenant, that's okay if we assume that we only configure these options for the Default tenant, and because we don't use this separator as a tenant settings in the setup form / settings editor.

> TableSeperator is something configurable is OC and could be different from one tenant to another.

Oops ;)

> However, in the DbConnectionValidator we assume that the connection being validated will use the currently configured table separator.

Yes, but you see the little issue.

Just saw that there is a new option field `UseLegacyIdentityColumn`, same remarks, e.g. if it is selectable in the new setup screens, tenant settings editor, okay we will be able to pass it as a new param to `ValidateAsync()`, and in that case not part of the YesSqlOptions (or yes just to show/use a default). Note:  For the separator, one solution would be to do the same. Otherwise here same little issue.

Hmm, maybe here the right solution is to use the dotnet `IConfiguration` to retrieve the `YesSqlOptions` from the config stack but under the specific section of the new tenant we want to create and of which we know the name.

src/OrchardCore/OrchardCore.Data.YesSql/OrchardCoreBuilderExtensions.cs
hishamco: Ok the, `IdentityColumnType` looks much better

src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/YesSqlOptions.cs
hishamco: No problem, but I'm not sure why it's named `IdentityColumnSize` not `IdentityColumnType`?

test/OrchardCore.Tests/Orchard.Queries/SqlParserTests.cs
hishamco: > like we check the prefix

Where?!! we are checking the result and the expected SQL

src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs
hishamco: @jtkech for your [comment](https://github.com/OrchardCMS/OrchardCore/pull/12528#discussion_r993979250) we referring to `TablePrefixSeparator` not `TablePrefix`

jtkech: Okay, so if we can have different option values per tenant, and because here we may validate the connection of a tenant that doesn't exist yet, we need to find a way to retrieve the possible configured values for this tenant.

ShellSettings holds tenant settings and a tenant config (the injectable ShellConfiguration), for a new tenant the config values are pre-populated (here only internally) but not yet from an eventual tenant specific section.

Hmm, it may become a little more complex, if we use `IConfiguration` by specifying an eventual tenant specific section, if this section doesn't exist it will not inherit of the values of the parent section.

So looks like we need to use `IShellSettingsManager.LoadSettingsAsync(string tenant)`, look at the code to see how we stack the config sources. Note: Needs to be tried and see if it doesn't fail.

Note: Hmm, for sqlite options make sense that they are only global options, about using connexion pool or not.

Finally, for options that will be part of the setup form / tenant editor, better to manage them through the tenant settings, the ones we can retrieve by using shellSettings[""] from the injectable ShellSettings. They use the same kind of config sources as pre-configured values e.g. to pre-populate the setup form / tenant editor, and on saving updated in the tenant specific appsettings that will always win.

So in that case we would not retrieve these values from YesSqlOptions but from the provided form data (or any other binding e.g. for Web Api), as I think we already do for the pre-configurable tenant TablePrefix. And then when configuring the store we would need to use shellSettings[""] to retrieve the actual values, as for TablePrefix here also.

Sorry for the long comments, not so obvious as I need to retrieve myself how it works. Yes would be less complex if we could assume that these yesSql options are the same for all tenants, if not no luck ;)

MikeAlhayek: @jtkech @hishamco i think for now so we don’t block this PR, we should just worry about adding this new setting to OC. We can create other PR for later version on OC along with support for Schema. I suggest we only get the new version on YesSql working.
I think the DbConnectionValidator should be refactored one more time to accept ValidationRequestOptions where the consumer would provide the connection string along with other parameters like TablePrefix and TablePrefixSeperator and TableNameConvention.

jtkech: No problem to find a quick compromise to not block this PR.

Then we will have to confirm if YesSqlOptions can be per tenant or not, in both case will be quite easy to do now that we better understand/remember what we already have, this without adding additional options classes.

hishamco: @sebastienros what's the action needed here to merge this PR for `1.5` shall we inject both interfaces for now, then refactor this class later on, or shall we need to do other thing

hishamco: I will pass a `null` for the schema now, and leave everything as it was

src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/YesSqlOptions.cs
```diff
@@ -19,4 +19,6 @@ public class YesSqlOptions
     public IContentSerializer ContentSerializer { get; set; }
 
     public string TablePrefixSeparator { get; set; } = "_";
+
+    public IdentityColumnSize IdentityColumnType { get; set; } = IdentityColumnSize.Int32;
```
MikeAlhayek, line 23: @hishamco 

I think we should remove this from this PR. This can cause a long debate and should not be blocking to this PR.

When this option is added, the value would have to be stored in the `ShellSettings` for each tenant specially when the value is changed to `Int64`. otherwise tenants with long values in their databases could have overflow exception if the options changed again.

IdentitySize + schema option would be part of a separate PR after releasing 1.5.

hishamco: Furthermore I think the name `IdentityColumnSize` is not suited

MikeAlhayek (approved):

jtkech:
Okay, I will tweak the DbConnectionValidator and suggest a PR this week end.

@sebastienros just for confirmation should we support YesSqlOptions (e.g. table separator) per tenant, I assume yes, but if it is not worth to do, having global options for all tenants would be simpler.

This because DbConnectionValidator is also used to validate the connection of a tenant that doesn't exist yet, anyway in both cases not so hard to do but just to confirm if it is worth to support.

hishamco:
I might say that we need to support `YesSqlOptions` as global level, but if we could override it per tenant - if it's required - will be a big plus, because in some cases we might need to use another options for a particular tenant(s)

jtkech:
So okay, as a reminder

Any value that will be selectable in a form (or other data source), as I think the `Schema` string will be, will have to be managed through model <=> tenant settings as we do for `TablePrefix`, and that we will pass to the `IDbConnectionValidator.ValidateAsync()`, if too many params maybe pass the `ShellSettings`.

- So will need to update the setup form, tenant editor, tenant API and so on.

The other options being in `YesSqlOptions` as the table separator, and as I think the new Id column size will be, and then used to configure the yesSql store. Here the problem is that

- If we still want to support different values per tenant and if we still want to validate the connection before a tenant is build, I had no solution ;)

- Hmm maybe I understand the problem, I think we should not allow a module that may be enabled or not to collaborate to the `YesSqlOptions` used to work with the database.

-  So looks like `YesSqlOptions` should come from configuration only as we assume for `SqliteOptions` which we bind oob from configuration, configured when calling `AddDataAccess()`. So I think we need to do the same for `YesSqlOptions`, in place of doing a post configure to let other modules collaborate.

- Then the solution to validate the connection before a tenant is built will be to check if a tenant configuration section exists for a given tenant name and so on until fallback to the parent section.

PR #12532, opened 2022-10-01
Remove hard coded HTML Ids from UserFields.Edit.cshtml to allow reusing it

### LINKED ISSUE #12531: Remove hard coded HTML ids from UserFields.Edit.cshtml view
I am trying to create a feature that would allow an Admin to bulk import users from excel file. After the user uploaded the excel file, I want to build the Edit fields for every user in my file. This means calling the `UserFields.Edit.cshtml` multiple times in the same output. The only problem that I am facing is that the password generator fails because the same html id are used on the same page which is not good. Here are the hard coded ids `generatePassword`, `passwordConfirmationToggle`, `passwordToggle` and `copyPassword`.

We should never hard code fixed ID in views that are used by driver. Instead, we should dynamically generate ID or class name to allow the user to reuse the views if needed like in my case.

### Workaround
I override the view in my own theme, but since this view is driver driven, I think it should be robust.

MikeAlhayek:
Replace hard coded HTML ids with generated ID to allow the reuse of `UserFields.Edit.cshtml`

Fix #12531

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Users/Views/UserFields.Edit.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Views/UserFields.Edit.cshtml
@@ -4,7 +4,13 @@
 @using Microsoft.Extensions.Options
 @using Microsoft.AspNetCore.Identity
 @inject IOptions<IdentityOptions> IdentityOptions
-
+@{
+    var uniqeId = Guid.NewGuid().ToString("n");
+    var generatePasswordId = $"GeneratePassword_{uniqeId}";
+    var passwordConfirmationToggleId = $"PasswordConfirmationToggle_{uniqeId}";
+    var passwordToggleId = $"PasswordToggle_{uniqeId}";
+    var copyPasswordId = $"CopyPassword_{uniqeId}";
+}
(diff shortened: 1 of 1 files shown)


Skrypt:
Looks fine to me.

Skrypt:
No, it is not consistent with what we generally do, I agree. Though, a specific use case here seems to be fine to use. Though, he could have overridden the Razor template in its own Theme for that matter without breaking anything. So, it is debatable.

MikeAlhayek:
@hishamco yes may be this isn't consistent with what we generally do. IMO, since a view is called in a driver, it should be reusable so hard coding HTML ids is not a good idea as I learned :)

@Skrypt yes I can override this view in my own theme "this is what I do today as a workaround". But I thought, we should also make this view robust in OC since it's called from drivers.

Either way, I change the code in the view to make it more robust by removing the hard coded HTML ids completely, and use css classes. This way the Javascript code actually is called with no problem for every instance rendered on the HTML page.

hishamco:
> Though, he could have overridden the Razor template in its own Theme for that matter without breaking anything

That's what supposed to be done, I diagree for adding things that's help user needs, where we provide an extensibility point for that

Let us hear the feedback from others

sebastienros (approved):

PR #12533, opened 2022-10-01
Don't lose an empty option name on saving Predefined List settings

### LINKED ISSUE #12592: Don't lose an empty option name on saving Predefined List settings
### Describe the bug

For a `PredefinedList` Text field it is allowed to define in a Migration an empty Option name.
It works well and when you edit the related settings you can see that the empty Option name is there.
But then if you save the settings and re-edit them, you can see  that you lost the empty Option name.

### To Reproduce
1. Define in a Migration a `PredefinedList` Text field with an empty Option name.
2. Setup the site and go to the settings editor of the above field.
3. You can see that the empty Option name is there, then save the settings.
4. Re-edit the settings to see that you lost the empty Option name.

### Expected behavior
Because it is allowed to define an empty Option name in a Migration and it works well.
Just don't remove this empty Option name when saving the `PredefinedList` settings.

jtkech:
Fixes #12592 

Because it is allowed to define an empty Option name in a Migration and it works well.
Just don't remove this empty Option name when saving the `PredefinedList` settings.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Assets/js/OptionsEditor/optionsEditor.js
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Assets/js/OptionsEditor/optionsEditor.js
@@ -18,7 +18,7 @@ function initializeOptionsEditor(elem, data, defaultValue, modalBodyElement) {
         },
         getOptionsFormattedList: function () {
             if (this.debug) { console.log('getOptionsFormattedList triggered') };
-            return JSON.stringify(this.state.options.filter(function (x) { return !IsNullOrWhiteSpace(x.name) }));
+            return JSON.stringify(this.state.options);
         }
     }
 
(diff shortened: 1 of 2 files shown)


Skrypt:
I would generally think that this is a bad idea. I understand that this would be for a "default" null/string empty value usage. Though, I probably added this string empty/null check for a reason back then.

hishamco:
I'm agree with the proposal but how you can guarantee other than first element is SHOULD non empty?!!

jtkech:
@Skrypt

I agree and this what I said but the customer still want an empty option name ;)

Maybe we could let the choice, currently we define editor settings in Migrations, if you edit them you can see the empty option but on saving we lose it, so to modify the settings we need a migration step.

@hishamco Most of the time the empty option would be the first option, but could be at any position e.g. the last one. Anyway looks like currently we are not filtering duplicate option names.

hishamco:
> Most of the time the empty option would be the first option, but could be at any position e.g. the last one.

From UI/UX perspective the first one is the best choice, so can we avoid the empty for other elements, also you can remove the duplicates as you mentioned

jtkech:
Yes good idea but here I only suggest to fix #12592, I rephrased the issue see my first comment.

We currently use empty option names (defined in migrations) and it work well, the only drawback is that if you edit and save the related settings you lose them. Here I only suggest to fix this.

Issue #12547, opened 2022-10-02
Default value for TextField


BinderDyn:
### Is your feature request related to a problem? Please describe.
Numeric fields and other have default values for their fields. TextField should also have the option to provide a default value e.g. for the color picker editor and so on.

### Describe the solution you'd like
I would like to add a DefaultValue string field to the TextFieldSettings model and display it in the settings view as well as displaying it if set in the text field editor

### Describe alternatives you've considered
Always let the user type in a value in the field and making it required. That's tiresome in many cases when there are multiple text fields that would be satisfied with a default.

I already engineered the proposed solution, forked the repo and sent in a PR

Thank you!

sebastienros:
Example: cloning, localization (more I can't think about right now)

gvkries:
We have this now.

PR #12548, opened 2022-10-02
Add text field default value field to settings model and display it in settings edit view

BinderDyn:
#12547

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Drivers/TextFieldDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Drivers/TextFieldDisplayDriver.cs
@@ -39,7 +39,7 @@ namespace OrchardCore.ContentFields.Drivers
             return Initialize<EditTextFieldViewModel>(GetEditorShapeType(context), model =>
             {
                 var settings = context.PartFieldDefinition.GetSettings<TextFieldSettings>();
-                model.Text = context.IsNew ? settings.DefaultValue : field.Text;
+                model.Text = context.IsNew && field.Text == null ? settings.DefaultValue : field.Text;
                 model.Field = field;
                 model.Part = context.ContentPart;
                 model.PartFieldDefinition = context.PartFieldDefinition;
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Drivers/TextFieldDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Drivers/TextFieldDisplayDriver.cs
@@ -38,7 +38,8 @@ namespace OrchardCore.ContentFields.Drivers
         {
             return Initialize<EditTextFieldViewModel>(GetEditorShapeType(context), model =>
             {
-                model.Text = field.Text;
+                var settings = context.PartFieldDefinition.GetSettings<TextFieldSettings>();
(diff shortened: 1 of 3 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=12548) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=12548) before we can accept your contribution.<br/><br/>:x: BinderDyn [sign now](https://cla.dotnetfoundation.org/OrchardCMS/OrchardCore?pullRequest=12548)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/OrchardCMS/OrchardCore?pullRequest=12548) it.</sub>

Skrypt:
The default value should be optional. The reason is that this field is used by many already without a Default value. You need to make it backward compatible. What happens if I have a TextField that doesn't have a DefaultValue setting in its content definition?

BinderDyn:
Didn't think if that but you're right. I'll rework the existing solution taking that into account.

BinderDyn:
@Skrypt I tested this with the given scenario. It is optional and does not break existing text fields. If the property of the settings was not present it simply returns null for DefaultValue. Because it is a string no harm is done there. It simply stays empty in the UI as there was no default set.

sebastienros:
Please check the issue comments, at least this PR implements the easy case.

src/OrchardCore.Modules/OrchardCore.ContentFields/Drivers/TextFieldDisplayDriver.cs
```diff
@@ -38,7 +38,8 @@ public override IDisplayResult Edit(TextField field, BuildFieldEditorContext con
         {
             return Initialize<EditTextFieldViewModel>(GetEditorShapeType(context), model =>
             {
-                model.Text = field.Text;
+                var settings = context.PartFieldDefinition.GetSettings<TextFieldSettings>();
+                model.Text = context.IsNew ? settings.DefaultValue : field.Text;
```
MikeAlhayek, line 42: If we were to take this PR,  you may want to change this line to 

model.Text = context.IsNew && field.Text == null ? settings.DefaultValue : field.Text;

BinderDyn: Implemented with newest commit

MikeAlhayek:
@sebastienros I think this is a fair ask since we already do similar thing in couple other field drivers.

MikeAlhayek (approved):
@sebastienros I am approving this as we already do the same thing with other fields. I think it is a fair ask. However, if you don't want this merged please close it.

MikeAlhayek (approved):
@sebastienros I am approving this as we already do the same thing with other fields. I think it is a fair ask. However, if you don't want this merged please close it.

sebastienros:
Approved since we already discussed it with you @MikeAlhayek recently for other fields and I believe the same solution is applied here. With a reminder that this won't work for anything that is not driven by the UI (imports, api calls, ...)

sebastienros:
A solution to the limitation might be to be more explicit in the labels/hints, like "will be used to pre-populate the field editor when the item is in creation"

PR #12567, opened 2022-10-04
Prevent AutoroutePartIndex from throwing exception when content definition is not found

### LINKED ISSUE #12575: AutoroutePartIndex from throwing exception when content definition is not found
### Describe the bug
When `Autoroute` feature is enabled, and one is updating content item where content-type is missing, AutoroutePartIndex from throwing exception when content definition is not found

### To Reproduce
1. Enable Autoroute
2. Create Content Type called "Test"
3. Create content item of "Test" called "Test Item"
4. Delete the content type "Test"
5. Edit the "Test Item"
6. Click Save

### Expected behavior
`AutoroutePartIndex` should not throw an exception

MikeAlhayek:
Fix #12575

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Autoroute/Indexes/AutoroutePartIndex.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Autoroute/Indexes/AutoroutePartIndex.cs
@@ -54,8 +54,8 @@ namespace OrchardCore.ContentManagement.Records
     public class AutoroutePartIndexProvider : ContentHandlerBase, IIndexProvider, IScopedIndexProvider
     {
         private readonly IServiceProvider _serviceProvider;
-        private readonly HashSet<ContentItem> _itemRemoved = new HashSet<ContentItem>();
-        private readonly HashSet<string> _partRemoved = new HashSet<string>();
+        private readonly HashSet<ContentItem> _itemRemoved = new();
+        private readonly HashSet<string> _partRemoved = new();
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.Autoroute/Indexes/AutoroutePartIndex.cs
```diff
@@ -92,7 +92,7 @@ public override async Task PublishedAsync(PublishContentContext context)
 
                 // Search for this part.
                 var contentTypeDefinition = _contentDefinitionManager.GetTypeDefinition(context.ContentItem.ContentType);
-                if (!contentTypeDefinition.Parts.Any(ctpd => ctpd.Name == nameof(AutoroutePart)))
+                if (contentTypeDefinition != null && !contentTypeDefinition.Parts.Any(ctpd => ctpd.Name == nameof(AutoroutePart)))
```
Skrypt, line 95: This is an issue because then the AutoroutePart record will stay in the database and thus will keep an active route/url for that content item while it should dispose of the AutoroutePart record of that Content Item. But at the same time if the Content Definition has been removed then the user should not be able to make any change to that content item because it should fail to retrieve its content definition and warn about it beforehand.

MikeAlhayek: I don't recall exact steps that lead me to this case, maybe importing from migration. Regardless, I think we should be doing null check to prevent null reference exception.

Issue #4393, opened 2019-09-25
Recipe steps fail silently when feature enablement isn't prioritized


jrestall:
When I use the Export functionality and by mistake put the 'All Features' step at the bottom of the deployment plan, it isn't immediately obvious why the recipe doesn't work when being used as a setup recipe.

The reason is that the recipe executor runs each step in the recipe in order, so any steps that rely on specific features aren't run since the features are enabled in the last step.

That makes sense to me now, but a user shouldn't have to worry about such technicalities. We should either prioritize the feature enablement step in the recipe or improve messaging and/or logging to make it clearer why a recipe step isn't executed.

deanmarcussen:
Also worth noting that the ContentDefinitions steps likes to be before Contents, so that would be another one to include in the ordering.

jtkech:
Just for infos, some 1st thoughts i had on this and that i commented on gitter.

> about recipe steps ordering, we already have a feature dependency graph that we use to have ordered features, and any recipe step is tied to a feature. So, maybe we could use this graph to order recipe steps. Hmm, all features would need to depend on the Features feature, but we also have a feature Priority property, so in place of adding a dependency on Features everywhere, we could set a very low priority to the Features feature that depends on nothing => to be first in the graph.

jtkech:
What i'm saying is that we already have services to get the enabled features in an order that rely on what we put in the `Manifest.cs` files for each feature about their dependencies and priority. So, maybe we could reuse this ordering (already implemented) to execute each recipe step (that is tied to a feature).

That said, looking in the `Manifest.cs` files, i saw that e.g `OC.Contentypes` depends on `OC.Contents`, meaning it is after in the ordering. So, maybe we would need to use the reverse order for recipe steps, or maybe this is a wrong dependency, maybe this is `Contents` that would need to depend on `Contentypes`. OR maybe one recipe step is in the wrong module but i don't think so.

This to execute `ContentDefinitionStep` before `ContentStep`.

Would need to be analyzed deeper

sebastienros:
The execution should respect the recipe order. But I am not opposed to provide hints in the recipe generation ui to prevent such issues. Now how to define such relative ordering is interesting, because it means steps needs to know about each others. Or we use an `Order` hint with `0` being the default, and then some could have `-1000` like features enabling/disabling, or `Content` be after `ContentType`. 

[...]

PR #12568, opened 2022-10-04
Add explicit order for executing recipe steps

### LINKED ISSUE #12566: Executing "Content" step before "ContentDefinition" caused null reference exceptions
### Describe the bug
The recipe steps are executed in the order they are listed in the the recipe file. so if "Content" step is listed above "ContentDefinition" when it's executed, there could be no content definition which causes a null reference exception. 

### To Reproduce
1. Import the following recipe in a site that had no content type definition named `HomePage` 
```
{
  "steps": [

     // Content
    {
      "name": "Content",
      "data": [
        // Default home page
        {
          "ContentItemId": "4dcz531ajab690tn643jar5rw7",
          "ContentType": "HomePage",
          "DisplayText": null,
          "Latest": true,
          "Published": true,
          "Owner": "[js: parameters('AdminUserId')]",
          "Author": "[js: parameters('AdminUsername')]",
          "HomePage": {},
          "AutoroutePart": {
            "Path": "home-page",
            "SetHomepage": false,
            "Disabled": false,
            "RouteContainedItems": false,
            "Absolute": false
          },
          "HtmlBodyPart": {
            "Html": "<p>Test Page</p>"
          }
        }
      ]
    },


    // Content Definitions
    {
      "name": "ContentDefinition",
      "ContentTypes": [
        {
          "Name": "HomePage",
          "DisplayName": "Landing Page",
          "Settings": {
            "ContentTypeSettings": {
              "Listable": true,
              "Draftable": true,
              "Versionable": true,
              "Securable": true
            },
            "FullTextAspectSettings": {}
          },
          "ContentTypePartDefinitionRecords": [
            {
              "PartName": "HomePage",
              "Name": "HomePage",
              "Settings": {
                "ContentTypePartSettings": {
                  "Position": "0"
                }
              }
            },
            {
              "PartName": "AutoroutePart",
              "Name": "AutoroutePart",
              "Settings": {
                "ContentTypePartSettings": {
                  "Position": "1"
                },
                "ContentIndexSettings": {},
                "AutoroutePartSettings": {
                  "AllowCustomPath": true
                }
              }
            },
            {
              "PartName": "HtmlBodyPart",
              "Name": "HtmlBodyPart",
              "Settings": {
                "ContentTypePartSettings": {
                  "Position": "2"
                },
                "ContentIndexSettings": {},
                "HtmlBodyPartSettings": {
                  "SanitizeHtml": false
                }
              }
            }
          ]
        }
      ],
      "ContentParts": []
    }
  ]
}
```

I think recipe steps should be executed in a logical order to ensure dependencies step are executed first.

For example, we should execute the following key steps in the following order
1. Feature Step
2. Content Definition
3. Content
4. ReplaceContent
5. DeleteContent

MikeAlhayek:
Fix #12566

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Features/Recipes/Executors/FeatureStep.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Features/Recipes/Executors/FeatureStep.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Linq;
 using System.Threading.Tasks;
+using OrchardCore.Abstractions;
 using OrchardCore.Environment.Shell;
 using OrchardCore.Recipes.Models;
 using OrchardCore.Recipes.Services;
@@ -10,7 +11,7 @@ namespace OrchardCore.Features.Recipes.Executors
     /// <summary>
     /// This recipe step enables or disables a set of features.
     /// </summary>
-    public class FeatureStep : IRecipeStepHandler
+    public class FeatureStep : IRecipeStepHandler, IDefineOrder
     {
         private readonly IShellFeaturesManager _shellFeaturesManager;
 
@@ -18,6 +19,8 @@ namespace OrchardCore.Features.Recipes.Executors
         {
             _shellFeaturesManager = shellFeaturesManager;
         }
+
+        public int Order => -100;
 
         public async Task ExecuteAsync(RecipeExecutionContext context)
         {
--- a/src/OrchardCore.Modules/OrchardCore.ContentTypes/RecipeSteps/ContentDefinitionStep.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentTypes/RecipeSteps/ContentDefinitionStep.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Threading.Tasks;
+using OrchardCore.Abstractions;
 using OrchardCore.ContentManagement.Metadata;
 using OrchardCore.ContentManagement.Metadata.Models;
 using OrchardCore.ContentManagement.Metadata.Records;
@@ -11,9 +12,11 @@ namespace OrchardCore.ContentTypes.RecipeSteps
     /// <summary>
     /// This recipe step creates content definitions.
     /// </summary>
-    public class ContentDefinitionStep : IRecipeStepHandler
+    public class ContentDefinitionStep : IRecipeStepHandler, IDefineOrder
     {
         private readonly IContentDefinitionManager _contentDefinitionManager;
+
+        public int Order => -25;
 
         public ContentDefinitionStep(IContentDefinitionManager contentDefinitionManager)
         {
--- a/src/OrchardCore.Modules/OrchardCore.ContentTypes/RecipeSteps/DeleteContentDefinitionStep.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentTypes/RecipeSteps/DeleteContentDefinitionStep.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Threading.Tasks;
+using OrchardCore.Abstractions;
 using OrchardCore.ContentManagement.Metadata;
 using OrchardCore.Recipes.Models;
 using OrchardCore.Recipes.Services;
@@ -9,7 +10,7 @@ namespace OrchardCore.ContentTypes.RecipeSteps
     /// <summary>
     /// This recipe step deletes content definition records.
     /// </summary>
-    public class DeleteContentDefinitionStep : IRecipeStepHandler
+    public class DeleteContentDefinitionStep : IRecipeStepHandler, IDefineOrder
     {
         private readonly IContentDefinitionManager _contentDefinitionManager;
 
@@ -17,6 +18,8 @@ namespace OrchardCore.ContentTypes.RecipeSteps
         {
             _contentDefinitionManager = contentDefinitionManager;
         }
+
+        public int Order => 100;
 
         public Task ExecuteAsync(RecipeExecutionContext context)
         {
--- a/src/OrchardCore.Modules/OrchardCore.ContentTypes/RecipeSteps/ReplaceContentDefinitionStep.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentTypes/RecipeSteps/ReplaceContentDefinitionStep.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Threading.Tasks;
+using OrchardCore.Abstractions;
 using OrchardCore.ContentManagement.Metadata;
 using OrchardCore.ContentManagement.Metadata.Records;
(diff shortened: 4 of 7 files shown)


src/OrchardCore.Modules/OrchardCore.Features/Recipes/Executors/FeatureStep.cs
```diff
@@ -10,7 +11,7 @@ namespace OrchardCore.Features.Recipes.Executors
     /// <summary>
     /// This recipe step enables or disables a set of features.
     /// </summary>
-    public class FeatureStep : IRecipeStepHandler
+    public class FeatureStep : IRecipeStepHandler, IDefineOrder
```
jtkech, line 14: Maybe `IOrderedRecipeStepHandler`

MikeAlhayek: `IDefineOrder` is a generic name that can be used in multiple scenarios as needed. This way we don't have to use it only for recipe steps.

jtkech: Yes, I meant

    public interface IOrderedRecipeStepHandler : IRecipeStepHandler
    {
        int Order { get; }
    }

Then for the handlers we want to order just replace ` : IRecipeStepHandler` by ` : IOrderedRecipeStepHandler`

Otherwise if like e.g. `IDisposable`, maybe `IOrderable`.

MikeAlhayek: @jtkech 
I follow you now. But, why do you recommend that over my approach? I feel that my approach is simpler, no need for inheritance and also, the `IRecipeStepHandler` will never need to depend on `IDefineOrder`

jtkech:
**Just for info**, I don't remember the details but there were already some discussions around this, maybe worth to search for them, e.g. I retrieved this one #4393 where I was suggesting to use as possible inner module/features dependencies.

Hmm, which somewhere is already the case because the order of the handler registrations is based on feature dependencies, but no luck this is "ContentTypes" that depends on "Contents" in the manifests ;) wich maybe is not totally true.

That said, as I remember I already fixed this issue in the "ContentStep" by importing the items in a deferred task so that the migrations are done.

https://github.com/OrchardCMS/OrchardCore/blob/232c6e2620ccadb6d92e2ed2a4f5230ec27df3da/src/OrchardCore.Modules/OrchardCore.Contents/Recipes/ContentStep.cs#L36-L40

But just before we check for the following, so the deferred task is not used if the tenant is actived, for example already setup.

https://github.com/OrchardCMS/OrchardCore/blob/232c6e2620ccadb6d92e2ed2a4f5230ec27df3da/src/OrchardCore.Modules/OrchardCore.Contents/Recipes/ContentStep.cs#L27-L32

MikeAlhayek:
@jtkech why can't we do something like this?

```
private async Task ExecuteStepAsync(RecipeExecutionContext recipeStep)
{
    var shellScope = recipeStep.RecipeDescriptor.RequireNewScope
        ? await _shellHost.GetScopeAsync(_shellSettings)
        : ShellScope.Current;

    await shellScope.UsingAsync(async scope =>
    {
        // Execute non-deferred recipes
        var allHandlers = scope.ServiceProvider.GetServices<IRecipeStepHandler>();
        var recipeStepHandlers = allHandlers.Where(handler => handler is not IDeferExecution ordrededStep)
        .OrderBy(handler => handler is IDefineOrder defineOrder ? defineOrder.Order : default);

        var scriptingManager = scope.ServiceProvider.GetService<IScriptingManager>();
        var recipeEventHandlers = scope.ServiceProvider.GetServices<IRecipeEventHandler>();

        await ExecuteStepAsync(recipeStep, scriptingManager, recipeStepHandlers, recipeEventHandlers);

        scope.AddDeferredTask(deferredScope =>
        {
            var recipeStepHandlers = deferredScope.ServiceProvider.GetServices<IRecipeStepHandler>()
            .Where(handler => handler is IDeferExecution ordrededStep)
            .OrderBy(handler => handler is IDefineOrder defineOrder ? defineOrder.Order : default);

            var scriptingManager = deferredScope.ServiceProvider.GetService<IScriptingManager>();
            var recipeEventHandlers = deferredScope.ServiceProvider.GetServices<IRecipeEventHandler>();

            return ExecuteStepAsync(recipeStep, scriptingManager, recipeStepHandlers, recipeEventHandlers);
        });
    });
}

private async Task ExecuteStepAsync(RecipeExecutionContext recipeStep, 
    IScriptingManager scriptingManager, 
    IEnumerable<IRecipeStepHandler> recipeStepHandlers, 
    IEnumerable<IRecipeEventHandler> recipeEventHandlers)
{
    // Substitutes the script elements by their actual values
    EvaluateJsonTree(scriptingManager, recipeStep, recipeStep.Step);

    if (_logger.IsEnabled(LogLevel.Information))
    {
        _logger.LogInformation("Executing recipe step '{RecipeName}'.", recipeStep.Name);
    }

    await recipeEventHandlers.InvokeAsync((handler, recipeStep) => handler.RecipeStepExecutingAsync(recipeStep), recipeStep, _logger);

    foreach (var recipeStepHandler in recipeStepHandlers)
    {
        await recipeStepHandler.ExecuteAsync(recipeStep);
    }

    await recipeEventHandlers.InvokeAsync((handler, recipeStep) => handler.RecipeStepExecutedAsync(recipeStep), recipeStep, _logger);

    if (_logger.IsEnabled(LogLevel.Information))
    {
        _logger.LogInformation("Finished executing recipe step '{RecipeName}'.", recipeStep.Name);
    }
}
```
using the above, the `ContentStep` would implement the new `IDeferExecution` interface instead. So it would look something like this

```
public class ContentStep : IRecipeStepHandler, IDefineOrder, IDeferExecution
    {

        public ContentStep(IContentManager contentManager)
        {
            _contentManager = contentManager;
        }

        public int Order => 50;

        public bool IsDefeered => true;

        public async Task ExecuteAsync(RecipeExecutionContext context)
        {
            if (!String.Equals(context.Name, "Content", StringComparison.OrdinalIgnoreCase))
            {
                return;
            }

            var model = context.Step.ToObject<ContentStepModel>();
            var contentItems = model.Data.ToObject<ContentItem[]>();

            await _contentManager.ImportAsync(contentItems);
        }
    }

    public class ContentStepModel
    {
        public JArray Data { get; set; }
    }
}
```

So the recipe executor would execute the ordered steps that are not deferred. Then will continue with the ordered and deferred steps.

jtkech:
@MikeAlhayek I thought about it a little, see https://github.com/OrchardCMS/OrchardCore/issues/12566#issuecomment-1271160522

PR #12588, opened 2022-10-08
Reduce ExtensionManager memory allocations

lahma:
TLDR; reduce test suite memory allocations by 40GB (> 90%).

I profiled tests run by OrchardCode suite to check if there's any easy wins to make them go faster. I was quite surprised by the memory allocation for the whole suite when run in Release mode, it was almost 44GB of memory allocated during the suite run. 

After analysis I found out the `ExtensionManager` to be the culprit. It has a lot of closures and allocatey-LINQ usage + enumerators from interface use. Generally, I'd say it would be fine if done once but as it's repeated every time when system initializes in tests, it starts to cost.

 I went through the code and made it less cool by using concrete types, more imperative constructs instead of LINQ and better equality lookups (feature id instead of feature object).

The test suite now allocates 40GB less during run and the `ExtensionManager` should be the same functionality-wise. This helps with the test suite performance and probably also with OrchardCore cold startup speed at least.

Relates to https://github.com/OrchardCMS/OrchardCore/issues/12583

## Before

image

## After

image

### DIFF
--- a/src/OrchardCore/OrchardCore/Extensions/ExtensionManager.cs
+++ b/src/OrchardCore/OrchardCore/Extensions/ExtensionManager.cs
@@ -11,20 +11,25 @@ using OrchardCore.Environment.Extensions.Manifests;
 using OrchardCore.Environment.Extensions.Utility;
 using OrchardCore.Modules;
 
+// ReSharper disable ForeachCanBePartlyConvertedToQueryUsingAnotherGetEnumerator
+// ReSharper disable ParameterTypeCanBeEnumerable.Local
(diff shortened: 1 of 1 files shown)


lahma:
Comparing to other PR "Deprecate ILuceneSearchQueryService in OrchardCore.Search.Lucene" which was run 11 hours ago and it run times, of course very unscientific because might have been different runner etc.

## Other PR

image

## This PR

image

lahma:
Could I get `hacktoberfest-accepted` label for this PR as a reward, it seems to be available in the repo 🙏🏻

Piedone:
Awesome!

PR #12589, opened 2022-10-08
GitHub Actions pr_ci.yml improvements

lahma:
Relates to https://github.com/OrchardCMS/OrchardCore/issues/12583 , checking how to make PR CI build faster on GH actions side.

* NuGet packages probably come faster from local data center than public download interface
* Don't install .NET 6 (comes OOB), don't install Node.JS 15 (v16 comes OOB)
* Add hard-linking to improve build speeds on Windows, [some discussion about the feature](https://github.com/dotnet/msbuild/issues/3788) (not working inside VS but for CI build seems like a good choice)

Probably not a common thing to see Windows build completing faster than the Linux one?

image

I'd so much love to have NUKE used in this project, but I guess that's way too far-fetched...

### DIFF
--- a/.github/workflows/pr_ci.yml
+++ b/.github/workflows/pr_ci.yml
@@ -50,7 +50,7 @@ jobs:
     - name: Restore
       if: steps.cache-nuget.outputs.cache-hit != 'true'
       run: |
-        dotnet restore --verbosity normal /p:DisableImplicitNuGetFallbackFolder=true /p:DisableImplicitLibraryPacksFolder src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
+        dotnet restore /p:DisableImplicitNuGetFallbackFolder=true /p:DisableImplicitLibraryPacksFolder src/OrchardCore.Cms.Web/OrchardCore.Cms.Web.csproj
 
     # for CI build we can use Deterministic = false to enable restore
     - name: Build
--- a/.github/workflows/pr_ci.yml
+++ b/.github/workflows/pr_ci.yml
@@ -33,7 +33,7 @@ jobs:
            **/obj/project.assets.json
            **/obj/project.nuget.cache
            **/**/packages.lock.json
-         key: ${{ runner.os }}-nuget-${{ hashFiles('**/*.props') }}
+         key: ${{ runner.os }}-nuget-${{ hashFiles('**/Dependencies*.props', '**/Directory.Packages.props') }}
          restore-keys: |
             ${{ runner.os }}-nuget-
     - name: Restore
--- a/.github/workflows/pr_ci.yml
+++ b/.github/workflows/pr_ci.yml
@@ -15,10 +15,10 @@ jobs:
     name: Build & Test
     steps:
     - uses: actions/checkout@v3
-    - name: Setup .NET 6.0
-      uses: actions/setup-dotnet@v3
-      with:
-        dotnet-version: 6.0.x
+#    - name: Setup .NET 6.0
+#      uses: actions/setup-dotnet@v3
+#      with:
+#        dotnet-version: 6.0.x
     # https://github.com/actions/cache/blob/main/workarounds.md#improving-cache-restore-performance-on-windowsusing-cross-os-caching
     - if: ${{ runner.os == 'Windows' }}
       name: Use GNU tar
--- a/.github/workflows/pr_ci.yml
+++ b/.github/workflows/pr_ci.yml
@@ -50,7 +50,7 @@ jobs:
            **/obj/project.nuget.cache
            **/*.csproj.nuget.g.targets
            **/packages.lock.json
-         key: ${{ runner.os }}-nuget-${{ hashFiles('**/Dependencies*.props', '**/Directory.Packages.props') }}
+         key: ${{ runner.os }}-nuget-${{ hashFiles('**/packages.lock.json', '**/Dependencies.props', '**/Dependencies.AspNetCore.props', '**/Directory.Packages.props') }}
          restore-keys: |
             ${{ runner.os }}-nuget-
 
--- a/.github/workflows/pr_ci.yml
+++ b/.github/workflows/pr_ci.yml
@@ -43,10 +43,10 @@ jobs:
     - name: Restore
       if: steps.cache-nugets.outputs.cache-hit != 'true'
       run: |
-        dotnet restore --verbosity normal
+        dotnet restore --verbosity normal /p:DisableImplicitNuGetFallbackFolder=true /p:DisableImplicitLibraryPacksFolder
     - name: Build
       run: |
-        dotnet build -c Release --no-restore --framework net6.0 /p:DisableImplicitNuGetFallbackFolder=true /p:DisableImplicitLibraryPacksFolder /p:CreateHardLinksForCopyLocalIfPossible=true /p:CreateHardLinksForCopyFilesToOutputDirectoryIfPossible=true /p:CreateHardLinksForPublishFilesIfPossible=true
+        dotnet build -c Release --no-restore --framework net6.0 /p:CreateHardLinksForCopyLocalIfPossible=true /p:CreateHardLinksForCopyFilesToOutputDirectoryIfPossible=true /p:CreateHardLinksForPublishFilesIfPossible=true
     - name: Unit Tests
       run: |
         dotnet test -c Release --no-restore --no-build  --framework net6.0 ./test/OrchardCore.Tests/OrchardCore.Tests.csproj
--- a/.github/workflows/pr_ci.yml
+++ b/.github/workflows/pr_ci.yml
@@ -57,10 +57,9 @@ jobs:
       run: |
         dotnet restore /p:DisableImplicitNuGetFallbackFolder=true /p:DisableImplicitLibraryPacksFolder
 
-    # for CI build we can use Deterministic = false to enable restore
     - name: Build
       run: |
-        dotnet build -c Release --no-restore --framework net6.0 /p:IncludeSymbols=false /p:Deterministic=false /p:CreateHardLinksForCopyLocalIfPossible=true /p:CreateHardLinksForCopyFilesToOutputDirectoryIfPossible=true /p:CreateHardLinksForPublishFilesIfPossible=true
+        dotnet build -c Release --no-restore --framework net6.0 /p:IncludeSymbols=false /p:CreateHardLinksForCopyLocalIfPossible=true /p:CreateHardLinksForCopyFilesToOutputDirectoryIfPossible=true /p:CreateHardLinksForPublishFilesIfPossible=true
 
     - name: Unit Tests
       run: |
--- a/.github/workflows/pr_ci.yml
+++ b/.github/workflows/pr_ci.yml
@@ -34,7 +34,7 @@ jobs:
             ${{ runner.os }}-nuget-
     - name: Restore
       run: |
-        dotnet restore --verbosity normal
+        dotnet restore
     - name: Build
       run: |
         dotnet build --no-restore --configuration Release --framework net6.0 /p:CreateHardLinksForCopyLocalIfPossible=true /p:CreateHardLinksForCopyFilesToOutputDirectoryIfPossible=true /p:CreateHardLinksForPublishFilesIfPossible=true
--- a/.github/workflows/pr_ci.yml
+++ b/.github/workflows/pr_ci.yml
@@ -1,10 +1,13 @@
 name: PR - CI
+
 on: 
   pull_request:
     branches: [ main, release/** ]
+
 env:
   DOTNET_NOLOGO: true
   DOTNET_CLI_TELEMETRY_OPTOUT: true
+
 jobs:
   build_test:
     runs-on: ${{ matrix.os }}
@@ -18,10 +21,14 @@ jobs:
     - uses: actions/checkout@v3
       name: Checkout source code
 
-#    - name: Setup .NET 6.0
-#      uses: actions/setup-dotnet@v3
-#      with:
-#        dotnet-version: 6.0.x
+   - name: Setup .NET 6.0
+     uses: actions/setup-dotnet@v3
+     with:
+       dotnet-version: 6.0.x
+	   
+    - uses: actions/setup-node@v3
+      with:
+        node-version: "16"
 
     # https://github.com/actions/cache/blob/main/workarounds.md#improving-cache-restore-performance-on-windowsusing-cross-os-caching
     - if: ${{ runner.os == 'Windows' }}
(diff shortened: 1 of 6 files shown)


lahma:
```
Post job cleanup.
/usr/bin/tar --posix --use-compress-program zstdmt -cf cache.tzst --exclude cache.tzst -P -C /home/runner/work/OrchardCore/OrchardCore --files-from manifest.txt
Cache Size: ~558 MB (58[4](https://github.com/OrchardCMS/OrchardCore/actions/runs/3209701554/jobs/5246623608#step:16:4)899487 B)
Cache saved successfully
Cache saved with key: Linux-nuget-
````

**Comment by Skrypt** [comment]
It is the first time I hear about https://nuke.build/ 
Looks nice. Maybe a small POC to see the build times and compare first.

File: .github/workflows/pr_ci.yml [hunk 1]
```diff
@@ -15,17 +15,25 @@ jobs:
...
+# Node.JS 16 is included by default
+#    - uses: actions/setup-node@v2
+#      with:
+#        node-version: "15"
+# NET 6 is included by default
+#    - name: Setup .NET 6.0
+#      uses: actions/setup-dotnet@v1
+#      with:
+#        dotnet-version: 6.0.x
+#        include-prerelease: true
```
Skrypt, line 27: You can remove the comments. I think it is well explained in the PR.

lahma: I've removed the commented-out lines

lahma:
> It is the first time I hear about [nuke.build](https://nuke.build/) Looks nice. Maybe a small POC to see the build times and compare first.

I'll create a PR to show the thing. Just yesterday I just wanted to delete all binaries in OC build folders (netcoreapp3.1 and net5.0) leftovers, `dotnet clean` etc didn't help much so had to do manual search and delete. When using NUKE I would have just said `.\build.cmd clean` on command line to get it cleaned. Having common tasks and automation in c# really helps, but a PR will follow!

Skrypt:
I generally do `git clean -xdf` for that matter but sometimes that's a little too much. 😄

hishamco:
I suffered a lot in the paste, but I did what Jasmin suggested with clean everyting which is too much ;)

hishamco:
I think we need to discuss to support NUKE or not, or we can use programmable MSBuild tasks like what I started long time ago

hishamco:
For reference #9247 and #12364

lahma:
NUKE has the advantage that you can debug it too when needed using the IDE.

hishamco:
Sounds good, can we open a discussion for this one, so others can add thier comments and ideas

.github/workflows/pr_ci.yml
```diff
@@ -15,17 +15,15 @@ jobs:
...
-        node-version: "15"
-    - name: Setup .NET 6.0
-      uses: actions/setup-dotnet@v1
-      with:
-        dotnet-version: 6.0.x
-        include-prerelease: true
+         path: ~/.nuget/packages
+         key: ${{ runner.os }}-nuget-${{ hashFiles('**/packages.lock.json') }}
+         restore-keys: |
+            ${{ runner.os }}-nuget-
```
Piedone, line 23: Could you do some repeated measurements before/after so we can see an average for the total workflow runtime (or cache+build)? Two GHA builds can vary wildly between two runs, so any two builds are not really comparable.

We've experimented around for NuGet and NPM caching with our [OSOCE solution](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions), which uses `OrchardCore.Application.Cms.Targets` (so basically every Orchard package) and some, so it should be comparable in terms of NuGet restore requirements. It turned out that restoring the cache under Windows is so slow that it cancels out the savings.

.github/workflows/pr_ci.yml
```diff
@@ -15,17 +15,15 @@ jobs:
     name: Build & Test
     steps:
     - uses: actions/checkout@v2
-    - uses: actions/setup-node@v2
+    - uses: actions/cache@v3
       with:
-        node-version: "15"
-    - name: Setup .NET 6.0
-      uses: actions/setup-dotnet@v1
-      with:
```
Piedone, line 23: I'm not sure about these. While it's great to save on having to do these, but we also take a dependency on `ubuntu-latest` and `windows-latest` containing the version we want, and if the images are updated, our build can break without us touching it. Perhaps these actions could instead be optimized not to do the installation if it's already available?

Piedone (changes requested):
Interesting things, thank you! I'll check out the hard link build in our workflows too, I'm curious about the impact.

.github/workflows/pr_ci.yml
```diff
@@ -15,17 +15,15 @@ jobs:
...
-        dotnet-version: 6.0.x
-        include-prerelease: true
+         path: ~/.nuget/packages
+         key: ${{ runner.os }}-nuget-${{ hashFiles('**/packages.lock.json') }}
+         restore-keys: |
+            ${{ runner.os }}-nuget-
     - name: Build
       run: |
-        dotnet build --configuration Release --framework net6.0
+        dotnet build --configuration Release --framework net6.0 /p:CreateHardLinksForCopyLocalIfPossible=true /p:CreateHardLinksForCopyFilesToOutputDirectoryIfPossible=true /p:CreateHardLinksForPublishFilesIfPossible=true
```
Piedone, line 26: I tried this out with the OSOCE solution, see the results here: https://github.com/Lombiq/GitHub-Actions/issues/64#issuecomment-1275348134 No luck with performance.

.github/workflows/pr_ci.yml
```diff
@@ -15,17 +15,15 @@ jobs:
...
     - uses: actions/checkout@v2
-    - uses: actions/setup-node@v2
+    - uses: actions/cache@v3
       with:
-        node-version: "15"
-    - name: Setup .NET 6.0
-      uses: actions/setup-dotnet@v1
-      with:
-        dotnet-version: 6.0.x
-        include-prerelease: true
```
Piedone, line 25: v3 of `actions/setup-dotnet` already includes a feature not to install a given version if it already exists on the VM. So, using it should be enough for us to use the version we want if it exists on the image but still get the speedup: https://github.com/Lombiq/GitHub-Actions/pull/67/files#diff-7978921ed7153ba992ebf38332bdd64fe6f3acb10a452b2978859c5a835ea74bR30 (note the comment above, I initially understood the behavior differently).

sebastienros:
> I'd so much love to have NUKE 

Go to hell Marko

sebastienros (approved):

Piedone:
Don't merge, see my comments.

.github/workflows/pr_ci.yml
lahma: Ideally build system would handle this. For example NUKE checks solution's global.json for required SDK version and downloads if needed. Works also outside of GitHub. But probably upgrading actions is the way to go when we can't use the best tools..

.github/workflows/pr_ci.yml
lahma: I'll check with some verbose output...

.github/workflows/pr_ci.yml
lahma: can try them here too, I think these mostly should help with spinning disks, not sure what's the runner like in GHA

lahma:
So because changes requested status I can't get builds to run? Not so convenient...

.github/workflows/pr_ci.yml
Piedone: Well, we can have something fancy, but in the short term, just pointing this action to v3 will optimize this :).

.github/workflows/pr_ci.yml
Piedone: It's an old D2-tier Azure VM, which is supposed to use HDDs (I tested performance on GitHub Actions, not my own machine or else).

Piedone (approved):
I dont know why the build is not running, I think it should (I certainly don't see any way to let it run). I'll add an approval to see if it helps but please note that still, I don't think this should yet be merged.

Piedone:
Still nothing. I don't know... Please try to push again and let's see if the build kicks off.

lahma:
Sorry, by bad. The YAML had indent problem, this is why tools like NUKE generate the YAML for you...

.github/workflows/pr_ci.yml
lahma: Just had a run using v3 and it took 1m 29s, so it still comes with a price...

image

lahma:
So here's some insights I've gathered today:

* For Windows runners, when using cache, [should use GNU tar](https://github.com/actions/cache/blob/main/workarounds.md#improving-cache-restore-performance-on-windowsusing-cross-os-caching), it's still really slow though
* Even with `actions/setup-dotnet@v3`, it can sporadically take more than a minute to run
* It's possible to cache both NuGet packages and generated project level asset and NuGet target artifacts, then checking cache hit allows to skip restore altogether
  * restore is slow as requires a lot of disk I/O even when packages are already there
  * cache key is based on Dependencies file which has the versions so it should be unambiguous, cache is also prepared for Directory.Packages.props
  * needs to have `Deterministic=false` for build to work, but I think it's fine for PR CI
* `DOTNET_NOLOGO` is now the correct environment variable to set instead of `DOTNET_SKIP_FIRST_TIME_EXPERIENCE`


## When cache works as expected (dependencies have not changed)

### Ubuntu

image

### Windows

image

.github/workflows/pr_ci.yml
Piedone: That's supposed to be a lot faster... In practice I've seen <20s for this (vs ~3 minutes for a full install). So, I don't know.

.github/workflows/pr_ci.yml
```diff
@@ -3,7 +3,7 @@ on:
   pull_request:
     branches: [ main, release/** ]
 env:
-  DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
+  DOTNET_NOLOGO: true
```
Piedone, line 6: Are you sure only this is needed? [The docs](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables) doesn't say `DOTNET_SKIP_FIRST_TIME_EXPERIENCE` is deprecated, and from the descriptions is seems to me that `DOTNET_NOLOGO` is a subset of `DOTNET_SKIP_FIRST_TIME_EXPERIENCE`.

.github/workflows/pr_ci.yml
Piedone: The build you're showing was perhaps a one-time fluke. I again checked our builds, and these restores are ~15s when using a preinstalled SDK. See e.g.:

- https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3257120788/jobs/5348182484#step:3:1
- https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3257120788/jobs/5348182424#step:3:1
- https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3256910188/jobs/5347817613#step:3:1
- https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3256910188/jobs/5347817671#step:3:1
- https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3257569220/jobs/5348910654#step:3:1
- https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/3257569220/jobs/5348910715#step:3:1

.github/workflows/pr_ci.yml
Piedone: Could you reply, please?

Piedone (changes requested):
Also, I'm waiting for your reply under these convos:
- https://github.com/OrchardCMS/OrchardCore/pull/12589#discussion_r992758449
- https://github.com/OrchardCMS/OrchardCore/pull/12589#discussion_r992764321
- https://github.com/OrchardCMS/OrchardCore/pull/12589#discussion_r992755543

.github/workflows/pr_ci.yml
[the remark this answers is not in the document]
lahma: I tested it in one commit, end result was weird error about file locking and reverted the suggested ones:

https://github.com/OrchardCMS/OrchardCore/actions/runs/3258530283/jobs/5350690376

.github/workflows/pr_ci.yml
lahma: Definitely no need to install Node 15 when 16 is LTS, seems that for Windows image every unnecessary step will be slow. Using `-latest` already makes building unpredictable - it can be Windows 2032 some day.

.github/workflows/pr_ci.yml
lahma: The comment I made at the end shows the results, with no restore it's faster to have the cache in use. Restore is I/O intensive even when nothing needs to be restored (NuGet cache contains items).

.github/workflows/pr_ci.yml
lahma: 15 seconds is quite long time for no-op?

.github/workflows/pr_ci.yml
lahma: https://github.com/OrchardCMS/OrchardCore/actions/runs/3258551770/jobs/5350727783

image

.github/workflows/pr_ci.yml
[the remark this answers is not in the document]
Piedone: Interesting, thanks.

.github/workflows/pr_ci.yml
Piedone: We definitely shouldn't use v15 of Node, I agree. However, that doesn't necessarily mean we should use the version provided by the image, which is not really something contributors would keep in mind. While using the "latest" images indeed poses a similar issue, I'd say that the chance that a Node or .NET SDK version update brings issues is at least an order of magnitude larger, than that of other image updates, especially OS ones. It's anecdotal, but we even had minor .NET SDK versions breaking the build here (and I've seen that many times with our other apps at Lombiq too), while I can't remember OS updates breaking anything like this.

Similar to `setup-dotnet`, `setup-node`
[won't install anything if the version already exists locally](https://github.com/actions/setup-node#:~:text=The%20action%20will%20first%20check%20the%20local%20cache%20for%20a%20semver%20match.%20If%20unable%20to%20find%20a%20specific%20version%20in%20the%20cache%2C%20the%20action%20will%20attempt%20to%20download%20a%20version%20of%20Node.js.). So if we target v16 (since [the 3 latest LTS versions are pre-installed](https://github.com/actions/runner-images#:~:text=3%20latest%20LTS%20versions)) then this shouldn't be significantly slower than not doing anything, while retaining that the version we use for the build is known.

.github/workflows/pr_ci.yml
Piedone: If you mean [this comment](https://github.com/OrchardCMS/OrchardCore/pull/12589#issuecomment-1279934310) then I only see a single measurement. Sorry for being a pain here, it's just that a +-10% jitter in the performance of GitHub Actions builds is usual, with particularly (un)lucky ones even hitting outside of that, so while this is promising, I'd like to see with multiple runs whether the speedup persists.

.github/workflows/pr_ci.yml
Piedone: Thanks!

.github/workflows/pr_ci.yml
Piedone: That's what I initially thought as well, but [as someone kindly explained in the mentioned thread](https://github.com/actions/setup-dotnet/issues/284#issuecomment-1276250960) it's not a no-op :). It's also checking the latest released `6.0.x` version online, and if it's newer than the pre-installed one, installs that instead.

Piedone:
Is this something you'd like to revisit @lahma or should we close?

PR #12618, opened 2022-10-11
Contents should depend on ContentTypes, not the opposite

### LINKED ISSUE #12617: Contents should depend on ContentTypes, not the opposite
First related to #12566 where we discussed about ordering recipe steps, I was suggesting to use as a 1st default the inner dependencies ordering, so the order in which recipe handlers are registered. But currently `OC.ContentTypes`depends on `OC.Contents` whereas contents needs types definitions.

My 1st feeling was that `OC.Contents` should depend on `OC.ContentTypes`, not the opposite, then I saw that this is `OC.Contents` that registers `ContentDefinitionManager` through `AddContentManagement()`.

I will suggest a PR with more details.

jtkech:
Fixes #12617.

First related to #12566 where we discussed about ordering recipe steps, I was suggesting to use as a 1st default the inner dependencies ordering, so the order in which recipe handlers are registered. But currently `OC.ContentTypes`depends on `OC.Contents` whereas whereas contents needs types definitions.

My 1st feeling was that `OC.Contents` should depend on `OC.ContentTypes`, not the opposite.

- Then I saw that this is `OC.Contents` that registers `ContentDefinitionManager` by calling `AddContentManagement()`, so I moved this registration in the `OC.ContentTypes` startup.

- Then in the manifest of `OC.ContentTypes` I removed the dependency on `OC.Contents`, and in the manifest of `OC.Contents` I added the dependency on `OC.ContentTypes`.

- Then I tried the Saas recipe to see if it still works by only enabling `OC.ContentTypes` that doesn't depend anymore on `OC.Contents`, then by only enabling `OC.Contents` which now depends on `OC.ContentTypes`.

- Finally I updated the other manifests having a dependency on `OC.ContentTypes` to now have a dependency on `OC.Contents` (that now depends on `OC.ContentTypes`).

- **Updated**: Then in the `OC.ContentTypes` project file I was able to remove the references to content management projects, unless `ContentManagement.GraphQL`, will see why tomorrow.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.AdminDashboard/Manifest.cs
+++ b/src/OrchardCore.Modules/OrchardCore.AdminDashboard/Manifest.cs
@@ -6,6 +6,6 @@ using OrchardCore.Modules.Manifest;
     Website = ManifestConstants.OrchardCoreWebsite,
     Version = ManifestConstants.OrchardCoreVersion,
     Description = "Allows to organize widgets in an Admin Dashboard.",
-    Dependencies = new[] { "OrchardCore.ContentTypes", "OrchardCore.Admin" },
+    Dependencies = new[] { "OrchardCore.Contents", "OrchardCore.Admin" },
     Category = "Content Management"
 )]
--- a/src/OrchardCore.Modules/OrchardCore.Alias/Manifest.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Alias/Manifest.cs
@@ -6,6 +6,6 @@ using OrchardCore.Modules.Manifest;
     Website = ManifestConstants.OrchardCoreWebsite,
     Version = ManifestConstants.OrchardCoreVersion,
     Description = "The alias module enables content items to have custom logical identifier.",
-    Dependencies = new[] { "OrchardCore.ContentTypes" },
+    Dependencies = new[] { "OrchardCore.Contents" },
     Category = "Content Management"
 )]
--- a/src/OrchardCore.Modules/OrchardCore.Autoroute/Manifest.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Autoroute/Manifest.cs
@@ -5,6 +5,6 @@ using OrchardCore.Modules.Manifest;
     Author = ManifestConstants.OrchardCoreTeam,
     Website = ManifestConstants.OrchardCoreWebsite,
     Version = ManifestConstants.OrchardCoreVersion,
-    Dependencies = new[] { "OrchardCore.ContentTypes" },
+    Dependencies = new[] { "OrchardCore.Contents" },
     Category = "Navigation"
 )]
--- a/src/OrchardCore.Modules/OrchardCore.ContentFields/Manifest.cs
+++ b/src/OrchardCore.Modules/OrchardCore.ContentFields/Manifest.cs
@@ -13,7 +13,7 @@ using OrchardCore.Modules.Manifest;
     Name = "Content Fields",
     Category = "Content Management",
     Description = "Content Fields module adds common content fields to be used with your custom types.",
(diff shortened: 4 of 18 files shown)


MikeAlhayek:
@jtkech 

> Many projects reference ContentManagement mainly because ContentItemIndex (not the provider) is not defined in ContentManagement.Abstractions (easy to move).

I agree, I think Index should be part of another project. Probably `Core` project not Abstraction since the index will require `YesSql.Abstraction`. This way if someone want to use the Abstraction project without YesSql, they are free to do so.

> Finally I updated the other manifests having a dependency on OC.ContentTypes to now have a dependency on OC.Contents (that now depends on OC.ContentTypes).

The only issue that I see here is that all these module will have the `Content Items` UI. This may not be a problem for all, but may not be needed by all these modules. I think you are adding the Dependencies there just because you need some core services enabled. What if we add a new feature in `OrchardCore.Contents` called something like `OrchardCore.Contents.Foundation` and we make this feature `EnabledByDependencyOnly` and move the registration of the main services into this new on demand feature. Then, all these features will depend on `OrchardCore.Contents.Foundation` instead of `OrchardCore.Contents` unless truly the `OrchardCore.Contents` UI is needed the dependent module. The good thing about this approach is that we decouple the UI from the foundation services.

jtkech:
Yes good ideas, I already thought about this, see this part of my comment https://github.com/OrchardCMS/OrchardCore/issues/12566#issuecomment-1271160522.

> ..., `ContentTypes` is an UI but maybe registers services needed by `Contents`, if so would need another "core" module on which both would depend.

Yes, in O1 we had core projects that were also modules, so being part of the dependency graph, maybe we are missing such core modules.

So yes, for now this PR only suggest as a first step that `OC.Contents` depends on `OC.ContentTypes`.

> The only issue that I see here is that all these module will have the Content Items UI.

Yes I know, not the best we may want, but at least as before because these modules was depending on `OC.ContentTypes` that was depending on `OC.Contents`.

> Probably Core project not Abstraction since the index will require YesSql.Abstraction. This way if someone want to use the Abstraction project without YesSql, they are free to do so.

If you meant `.Core` projects, for now they are only related to modules to not have to reference them. Here I suggest to move `ContentItemIndex` (not the provider) from `OC.ContentManagement` to its `.Abstractions` that already depends on `YesSql.Abstractions`, we would need to split this abstractions.

Even more, all modules using query extensions as `IsIn()` directly or indirectly through `GraphQL`, need to reference `YesSql` because these query extensions are not defined in `YesSql.Abstractions`, but here would have to be done on YesSql side.

MikeAlhayek:
@jtkech instead of creating a new core project, why not created a new feature in OC.Contents using `EnabledByDependencyOnly = true` flag which is responsible of registering the services only? This would fix the problem

jtkech:
Yes that's a good idea, but then we still need to reference the OC.Contents package, but yes could prevent to register unused services.

Yes many things to think about, but here I only suggest a new dependency, I may close this PR soon, was just to try if it still works with this new dependency, but yes would need more work to refactor some projects to optimize project references and registered services.

Hmm but for now maybe still worth to change this dependency.

MikeAlhayek:
@jtkech I think you should keep this PR. But do consider first creating a “OrchardCore.Contents.Infrastructure” feature in the OC.Contents module which should be simple change. Then just use the new feature instead of depending on OC.Contents everywhere. 

Ultimately, we should consider separating UI from infrastructure so allow user to provide their own UI without having to worry about services.

jtkech:
> Then just use the new feature instead of depending on OC.Contents everywhere.

Yes good idea but may be a breaking change as many modules use view components defined in OC.Contents, look at under its `Views/Shared/Components` folder.

> Ultimately, we should consider separating UI from infrastructure

Yes, would be part of a bigger change.

PR #12626, opened 2022-10-11
Change the css to recognize any value other than darkmode as the default

### LINKED ISSUE #12625: Make the AdminTheme default to any theme other than "darkmode"
### Describe the bug
I have a mobile app that behaves as a wrapper which renders an OrchardCore based website. When running this mobile app on IOS, for some odd reason the value of `data-theme=default` in the Admin theme gets changed to `light` instead of default so it becomes `data-theme=light`. Since TheAdmin theme only is aware for `default` and `darkmode` the admin theme does not render correctly in the mobile app.

Unless there is a value not too, I think we can easily make TheAdmin theme supports `darkmode` and for anything else we load the `default` theme. By changing `:root[data-theme='default'] ` to `:root:not([data-theme='darkmode'])`

MikeAlhayek:
Fix #12625


Skrypt:
My only concern is that it will surely break other people's admin customizations. It needs to be marked as a breaking change.

Skrypt (approved):

MikeAlhayek:
@Skrypt it should not break anything really. Basically, we are using any string "other than `darkmode`" as the default so if we have custom value in the theme we still render the default. But if someone want to customize that, they can still target their custom value. So it should not be breaking change.

Skrypt:
You are right, It won't cause any issue because then "default" becomes "any" string. But beware that the default is now "any" ... which means that if you are using something like `data-theme="red"` then it won't be differentiated from "default". But, the feature was not meant to be used like that, although we never know.

PR #12627, opened 2022-10-11
Make all fields and parts in the admin to be responsive to reduce the vertical scrolling

### LINKED ISSUE #12493: Standardizing forms look in TheAdmin theme
Currently in TheAdmin theme in OC, we don't have a standard style across the panel. There is some fields that show up with 50% (like the select menu) and there are others that show up 100% width (like the title part). OC should provide a default standardized look across the board so all of the forms look a like.

Additionally, It would be nice if we place the label on the same row as the input when the screen is large. So this way we can utilize the screen width and lessor the vertical scrolling.

Make all inputs should be 100% width. Also, when the screen is large, we can place the label in the same row. Here is a code example of how I believe a form in OC should look like

```
    <form> 
    <div class="row mb-3">
      <label for="exampleFormControlInput1" class="col-form-label col-lg-2 col-xl-3 text-lg-end">Email address</label>
      <div class="col-lg-10 col-xl-9">     
        <input type="email" class="form-control" id="exampleFormControlInput1" placeholder="name@example.com">
      </div>
    </div>
    <div class="row mb-3">
      <label for="exampleFormControlTextarea1" class="col-form-label col-lg-2 col-xl-3 text-lg-end">Example textarea</label>
      <div class="col-lg-10 col-xl-9">     
        <textarea class="form-control" id="exampleFormControlTextarea1" rows="3"></textarea>
        <div class="hint">
            This is a hint
        </div>
      </div>
    </div>
  </form>
```

Here id a fiddler to allow you to evaluate the form for different screen sizes https://jsfiddle.net/crestapps/rhs5cmqb/13/

here are some screenshots of inconsistent forms

image

image

MikeAlhayek:
Fix #12493 

This PR adds new options for TheAdmin theme where one can change the look of the forms.

Adding the following setting will generate the screencasts below

```json
{
"OrchardCore": {
  "TheAdminTheme": {
    "StyleSettings": {
      "LimitedWidth": "col-md-6 col-lg-5 col-xxl-4",
      "StartClasses": "col-lg-2 col-xl-3",
      "EndClasses": "col-lg-10 col-xl-9",
      "LabelClasses": "col-form-label text-lg-end col-lg-2 col-xl-3",
      "OffsetClasses": "offset-lg-2 offset-xl-3"
    }
  }
}
```

In order for the above settings to work for custom views, you must update your view to use the new introduced helpers.
 1. Call `Orchard.GetLabelCssClasses()` helper to get the proper css classes for your label
 2. Wrap all elements that correspond to the label with an element like `div`, and call `Orchard.GetEndCssClasses()` helper to get the proper css classes for the new div.

For example, The `TitlePart.Edit.cshtml` view was changed from 

```
<div class="mb-3" asp-validation-class-for="Title">
    <label asp-for="Title">@T["Title"]</label>
    <input asp-for="Title" class="form-control content-preview-text content-caption-text" disabled="@(Model.Settings?.Options == TitlePartOptions.GeneratedDisabled)" autofocus="autofocus" dir="@culture.GetLanguageDirection()" />
    <span asp-validation-for="Title"></span>
    @if (!String.IsNullOrWhiteSpace(Model.Settings?.Pattern))
    {
        switch (Model.Settings?.Options)
        {
            case TitlePartOptions.Editable:
                <span class="hint">@T["Leave empty to auto-generate the title using the pattern."]</span>
                break;
            case TitlePartOptions.GeneratedDisabled:
                <span class="hint">@T["The title will be auto-generated using the pattern."]</span>
                break;
        }
    }
</div>
```

to this 

```
<div class="row mb-3" asp-validation-class-for="Title">
    <label asp-for="Title" class="@Orchard.GetLabelCssClasses()">@T["Title"]</label>
    <div class="@Orchard.GetEndCssClasses()">
        <input asp-for="Title" class="form-control content-preview-text content-caption-text" disabled="@(Model.Settings?.Options == TitlePartOptions.GeneratedDisabled)" autofocus="autofocus" dir="@culture.GetLanguageDirection()" />
        <span asp-validation-for="Title"></span>
        @if (!String.IsNullOrWhiteSpace(Model.Settings?.Pattern))
        {
            switch (Model.Settings?.Options)
            {
                case TitlePartOptions.Editable:
                    <span class="hint">@T["Leave empty to auto-generate the title using the pattern."]</span>
                    break;
                case TitlePartOptions.GeneratedDisabled:
                    <span class="hint">@T["The title will be auto-generated using the pattern."]</span>
                    break;
            }
        }
    </div>
</div>
```

- `OrchardCore.GetLimitedWidthCssClasses()` this helper to be used for a field that had limited width like `number` type. As an example, look at `NumericField.Edit.cshtml` view in the PR.
- `GetLabelCssClasses()` this helper to get the css classes for a label element.
- `GetStartCssClasses()` gets the css classes for a stating section. For example, if you want to push your content to the right to light it up with a label.
- `GetEndCssClasses()` get the css classes for the ending section.
- `GetEndCssClasses(bool withOffset)` gets the css classes for the ending section with optionally an offset when 
- `GetStartCssClasses()` or `GetLabelCssClasses()` isn't used.
- `GetOffsetCssClasses()` gets the offset classes to align the right content. 

If no options are provided, the forms will render how they appear today.

### Here is how the content fields looks
content-fields

### Here is how the content parts looks

content-parts


src/OrchardCore/OrchardCore.DisplayManagement/Html/CssOrchardHelper.cs
```diff
@@ -0,0 +1,95 @@
...
+    public static string GetOffsetCssClasses(this IOrchardHelper _, params string[] additionalClasses)
+    {
+        return String.Join(' ', GetOffsetCssClassList(additionalClasses));
+    }
+
+    private static List<string> GetLabelCssClassList(params string[] additionalClasses)
+    {
+        return new List<string>(GetStartCssClassList(additionalClasses))
+        {
+            "col-form-label",
```
Skrypt, line 42: This is something that we should not do. Hard coding Bootstrap classes in a helper class. I need to understand why it is required.

MikeAlhayek: @Skrypt 
It may sounds like a bad idea, but since we maintain TheAdmin theme, it makes things much easier to change down the line or even make it appsettings based so the user provide their own settings. Also, any view can use these helpers to adhere to the TheAdmin theme settings.

MikeAlhayek: @Skrypt I actually change the code so it driven from options. We can also make these options available per tenant by adding the settings to the Site. If people like this approach, I'll add UI settings to change the settings for each settings from the UI.

agriffard:
Can you please add documentation to explain module developers how to change the views of their fields/parts?

MikeAlhayek:
@agriffard I added some docs into this PR.

MikeAlhayek:
@sebastienros Where do we stand with this PR? I don't see any objections on the approach. This can be a tremendous help for my use case. And if no one use the the configuration they will not be impacted by this change.

sebastienros:
We need to decide whether we want to merge it or not, for the reasons explained in the meetings where the feature was demoed. (and also today's meeting, 12/20/2022).

To vote please use 👍 and 👎 on the reactions for this comment (not on a new comment please, but you can still comment on the PR on top of it).

Limit for the vote will be January 3rd 2023.

ns8482e:
IMO Its always better to keep views simple using html and CSS, and avoid C# code for rendering as much as possible.

For example I do don't like to write code as below

```cshtml
<div style='width: @(IsReqestFromMobile(Request) ? "767px" : "992px")'>
``` 
Instead prefer to write responsive media query in CSS thats why like bootstrap. 

Having extension methods like `class="@Orchard.GetEndCssClasses(true)"` forces to rely on C# code that's not good. It also force me to write views in specific way not just using html/css  eg. I can't use these extension methods methods in liquid (if i want to write admins screens in liquid).

May be instead of  using C# extension method - you should fall back to utilize bootstrap mixins and CSS rules instead.

ns8482e:
> @sebastienros Where do we stand with this PR? I don't see any objections on the approach. This can be a tremendous help for my use case. And if no one use the the configuration they will not be impacted by this change.

I am not in favor of having configuration for rendering. That should be managed by SCSS/CSS

julrichkieffer:
In response to Sebastian asking for community opinion on this PR on 20/12/2022 meeting, I vote to include this PR: 👍 

Where I understand Sebastian's reservations on addtional work placed on developers, particularly for existing dependencies on TheAdmin, I wonder if the following my apply:
- As TheAdmin is a module, it's versionable. Existing dependencies rely on the module's current / past versions
- As TheAdmin is a module, it could also include an optional feature.  
In effect, were this PR to be wrapped in a new major version or a separate feature of TheAdmin module, then developers have an opt-in to take a dependency on this "additional work." Using the version approach, TheAdmin may adopt this issue as default from ~v2.0 onwards. A feature may be used in the interim (<v2.0).

Might a separate feature also be created to apply the new setttings, or provide a TheAdmin Theme Settings UI with the following as defaults:
```json
{
"OrchardCore": {
  "TheAdminTheme": {
    "StyleSettings": {
      "LimitedWidth": "col-md-6 col-lg-5 col-xxl-4",
      "StartClasses": "col-lg-2 col-xl-3",
      "EndClasses": "col-lg-10 col-xl-9",
      "LabelClasses": "col-form-label text-lg-end col-lg-2 col-xl-3",
      "OffsetClasses": "offset-lg-2 offset-xl-3"
    }
  }
}
```

jtkech:
To reduce the vertical scrolling we also have Tabs and Columns that can be tweaked with placement files.

ldimitrijevic:
Hello,
i voted yes for this PR but not for technical reasons as i did not have time to check for the impacts. Maybe i am wrong but isn't it possible to create a responsive admin theme and leave the actual theme as is to avoid any impact.
Anyhow i voted yes because i think that the UX is of much importance in such projects. A CMS is not administered by a tech guy but by a user that may have no idea of what responsive is. He just wants something nice that works. 
"You can have the best software ever, if the UX is bad no one will adopt it". I read this once and it is totally true.

MikeAlhayek:
@sebastienros can we please review this PR during Triage in order to merge it hopefully before shipping 1.6?

sebastienros (approved):

sebastienros:
Approved since it's an option that is disabled by default. Once people start complaining about teh inconsistencies, we'll revert it.

PR #12637, opened 2022-10-13
SqlParser & SqlGrammar CTE support

### LINKED ISSUE #12446: SqlParser can't parse SQL queries with CTEs in them
### Describe the bug
There is a line in SqlQuerySource that tries to parse a SQL query:

`SqlParser.TryParse(tokenizedQuery, dialect, _session.Store.Configuration.TablePrefix, parameters, out var rawQuery, out var messages)`

If you try to use a query with a CTE (common table expression), the parse fails with the following error:

Syntax error, expected: SELECT at line:0, col:0

I'm not sure if this is a bug or by design, given Orchard supports storage in different database providers and I'm not sure they all support CTEs. If so, perhaps this could be moved to a feature request for those of us who want to use CTEs and are ok with a requirement that our sites must run on a database that supports them.

### Expected behavior
Ideally, we could use CTEs in SQL queries. Note: I'm not suggesting necessarily that the Orchard UI should support that, but if we use a database connection directly, it would be nice if we could use CTEs (and get the benefits of the SqlParser.TryParse method).

lampersky:
fixes #12446

example sql query, which is now valid:
```
with 
  cte1(abc, def) as (
    select id as abc, id as def from s
    union all
    select id as abc, id as def from t
  ), 
  cte2(ghi,jkl) as (
    select abc as ghi, def as jkl from cte1
  )
select ghi, jkl from cte2;
```

/cc @sebastienros

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Queries/Sql/SqlParser.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Queries/Sql/SqlParser.cs
@@ -808,55 +808,54 @@ namespace OrchardCore.Queries.Sql
 
         private string EvaluateCteStatement(ParseTreeNode cteStatement)
         {
-            var builder = new StringBuilder();
-            builder.Append("WITH ");
+            _builder.Append("WITH ");
             for (var i = 0; i < cteStatement.ChildNodes[1].ChildNodes.Count; i++)
             {
                 var cte = cteStatement.ChildNodes[1].ChildNodes[i];
                 if (i > 0)
                 {
-                    builder.Append(", ");
+                    _builder.Append(", ");
                 }
                 
                 var expressionName = cte.ChildNodes[0].Token.ValueString;
                 var optionalColumns = cte.ChildNodes[1];
-                builder.Append(expressionName);
+                _builder.Append(expressionName);
                 if (optionalColumns.ChildNodes.Count > 0)
                 {
                     var columns = optionalColumns.ChildNodes[0].ChildNodes;
-                    builder.Append("(");
+                    _builder.Append("(");
                     for (var j = 0; j < columns.Count; j++)
                     {
                         if (j > 0)
                         {
-                            builder.Append(", ");
+                            _builder.Append(", ");
                         }
-                        builder.Append(columns[j].Token.ValueString);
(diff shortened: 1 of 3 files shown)


src/OrchardCore.Modules/OrchardCore.Queries/Sql/SqlParser.cs
```diff
@@ -785,6 +806,59 @@ private void EvaluateOverClauseOptional(ParseTreeNode overClauseOpt)
             _builder.Append(")");
         }
 
+        private string EvaluateCteStatement(ParseTreeNode cteStatement)
+        {
+            var builder = new StringBuilder();
```
sebastienros, line 811: Looks like it could use `_builder` instead of creating another one

sebastienros:
Looks great as usual.
After this is merge have you thought about re-implementing with Parlot? We could then remove Irony's dependency and it would be faster. Also it would make me happy.

src/OrchardCore.Modules/OrchardCore.Queries/Sql/SqlParser.cs
lampersky: definitely, I've just fixed it

lampersky:
> After this is merged have you thought about re-implementing with Parlot? 

I will think about it.

hishamco:
> Looks great as usual.
> After this is merged have you thought about re-implementing with Parlot? We could then remove Irony's dependency and it would be faster. Also it would make me happy.

I already started that long time ago, I will share with you some links

hishamco:
FYI 

https://github.com/OrchardCMS/OrchardCore/issues/8214 
https://github.com/hishamco/SqlParser

BTW I did a lot of changes in `SqlParser` in my local but I didn't push them yet, @lampersky if you are interested we can work together on this to make it happen

hishamco:
I will put this into my list, I need to review it because i'm interested in parsers ;)

sebastienros (approved):

hishamco:
@lampersky can we close this in favor of #12645?

lampersky:
> @lampersky can we close this in favor of #12645?

Yes please.

PR #12647, opened 2022-10-15
Use YesSql Named Options based on tenant names

### LINKED ISSUE #12646: If YesSqlOptions per tenant, can't use DbConnectionValidator before a tenant is built
- If `YesSqlOptions` may have different values per tenant, we can't use `DbConnectionValidator` on a shell that is not yet built. I thought about only using configuration so that we could stack a tenant config section.

        var tenantConfigurationOnly = new ConfigurationBuilder()
            .AddConfiguration(_configuration)
            .AddConfiguration(_configuration.GetSection(shellName))
            .Build();

        var yesSqlOptions = tenantConfigurationOnly.Get<YesSqlOptions>();

    But `YesSqlOptions` not only contains simple fields but also implementations as `TableNameConvention`.

- So I'm thinking about an `IConfigureNamedOptions<YesSqlOptions>` returning options by tenant name, the default would always return the same values, and people could register a custom one to collaborate by applying their rules, or if they know by advance some tenant names just use.

      s.Configure<YesSqlOptions>("Default", options => options.TablePrefixSeparator = "_");
      s.Configure<YesSqlOptions>("MyTenant1", options => options.TablePrefixSeparator = "__");
      s.Configure<YesSqlOptions>("MyTenant2", options => options.TablePrefixSeparator = "___");

    People would have to register their config from the application e.g. by using `OCBuilder.ConfigureServices()`, as we do in `OCBuilder.AddDataAccess()`. This to have the same registrations for all tenants, to not rely on any module feature and to be able to retrieve options for a given tenant name before the tenant itself is built.

I will suggest a PR.

jtkech:
Fixes #12646 

- If `YesSqlOptions` may have different values per tenant, we can't use `DbConnectionValidator` on a shell that is not yet built. I thought about only using configuration so that we could stack a tenant config section.

        var tenantConfigurationOnly = new ConfigurationBuilder()
            .AddConfiguration(_configuration)
            .AddConfiguration(_configuration.GetSection(shellName))
            .Build();

        var yesSqlOptions = tenantConfigurationOnly.Get<YesSqlOptions>();

    But `YesSqlOptions` not only contains simple fields but also implementations as `TableNameConvention`.

- So I'm thinking about an `IConfigureNamedOptions<YesSqlOptions>` returning options by tenant name, the default would always return the same values, and people could register a custom one to collaborate by applying their rules, or if they know by advance some tenant names just use.

      s.Configure<YesSqlOptions>("Default", options => options.TablePrefixSeparator = "_");
      s.Configure<YesSqlOptions>("MyTenant1", options => options.TablePrefixSeparator = "__");
      s.Configure<YesSqlOptions>("MyTenant2", options => options.TablePrefixSeparator = "___");

    People would have to register their config from the application e.g. by using `OCBuilder.ConfigureServices()`, as we do in `OCBuilder.AddDataAccess()`. This to have the same registrations for all tenants, to not rely on any module feature and to be able to retrieve options for a given tenant name before the tenant itself is built.

- Maybe we should do the same for `SqliteOptions` but it is there since a longer time I think.

### DIFF
--- a/src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/YesSqlOptions.cs
+++ b/src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/YesSqlOptions.cs
@@ -19,4 +19,6 @@ public class YesSqlOptions
     public IContentSerializer ContentSerializer { get; set; }
 
     public string TablePrefixSeparator { get; set; } = "_";
+
+    public IdentityColumnSize IdentityColumnSize { get; set; } = IdentityColumnSize.Int32;
 }
--- a/src/OrchardCore/OrchardCore.Data.YesSql/Options/YesSqlOptionsConfiguration.cs
+++ b/src/OrchardCore/OrchardCore.Data.YesSql/Options/YesSqlOptionsConfiguration.cs
@@ -1,0 +1,17 @@
+using System;
+using Microsoft.Extensions.Options;
+using OrchardCore.Data.YesSql;
+using YesSql.Services;
+
+namespace OrchardCore.Data
+{
+    public class YesSqlOptionsConfiguration : IConfigureNamedOptions<YesSqlOptions>
+    {
+        public void Configure(YesSqlOptions options) => Configure(String.Empty, options);
+
+        public void Configure(string name, YesSqlOptions options)
+        {
+            options.TableNameConvention ??= new DefaultTableNameConvention();
+        }
+    }
+}
--- a/src/OrchardCore.Cms.Web/appsettings.json
+++ b/src/OrchardCore.Cms.Web/appsettings.json
@@ -88,7 +88,8 @@
     //"OrchardCore_Shells_Database": {
     //  "DatabaseProvider": "SqlConnection", // Set to a supported database provider.
     //  "ConnectionString": "", // Set to the database connection string.
-    //  "TablePrefix": "", // Optionally, configure a table prefix.
+    //  "TablePrefix": "", // Optionally, configure the database table prefix.
+    //  "Schema": "", // Optionally, configure the database schema.
     //  "MigrateFromFiles": true // Optionally, enable to migrate existing App_Data files to Database automatically.
     //},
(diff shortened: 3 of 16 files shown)


src/OrchardCore/OrchardCore.Data.YesSql/Options/YesSqlOptionsConfiguration.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using Microsoft.Extensions.Options;
+using OrchardCore.Data.YesSql;
+using YesSql.Services;
+
+namespace OrchardCore.Data
```
hishamco, line 6: Why not `OC.Data.YesSql`?

jtkech: Yes I agree, it's just a beginning will see tomorrow ;)

For now just because it is OC.Data in the other files under the same folder.

hishamco: I have the same comment for other files in `OC.Data.YesSql` namespace. The PR looks good but we need to fix the unit test

Great job as usual

src/OrchardCore/OrchardCore.Data.YesSql.Abstractions/YesSqlOptions.cs
```diff
@@ -19,4 +19,6 @@ public class YesSqlOptions
     public IContentSerializer ContentSerializer { get; set; }
 
     public string TablePrefixSeparator { get; set; } = "_";
+
+    public IdentityColumnSize IdentityColumnSize { get; set; } = IdentityColumnSize.Int32;
```
hishamco, line 23: I still insist to change the enum name in `YesSQL`, especially if we need to support more types in the future

jtkech:
I close this one as I saw another concern and then opened a new issue.

PR #12677, opened 2022-10-21
Enable / disable RSS feed per list

### LINKED ISSUE #12650: Lists - RSS feeds should be optional
After activation of Feeds module, each List part gets its public url which can be accessed from public and cannot be secured.
List items can be intended to be private / handled in some custom way. If someone activates Feeds module, it can leak those data through newly added endpoint. 

I do not know if there is another endpoit which would show this data. Anonymous access of list on frontend is disabled by removing roles from anonymous user. 

Add option to disable/enable feed generation for given list part. Or permission implementation.

image
even setting custom url does not disable created endpoint

hishamco:
Fixes #12650

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Feeds/Controllers/FeedController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Feeds/Controllers/FeedController.cs
@@ -65,23 +65,23 @@ namespace OrchardCore.Feeds.Controllers
             }
 
             var document = await context.Builder.ProcessAsync(context, async () =>
-           {
-               await bestQueryMatch.FeedQuery.ExecuteAsync(context);
-
-               await _feedItemBuilder.PopulateAsync(context);
-
-               foreach (var contextualizer in context.Response.Contextualizers)
-               {
-                   if (ControllerContext != null)
-                   {
-                       contextualizer(new ContextualizeContext
-                       {
-                           ServiceProvider = _serviceProvider,
-                           Url = Url
-                       });
-                   }
-               }
-           });
+            {
+                await bestQueryMatch.FeedQuery.ExecuteAsync(context);
+            
+                await _feedItemBuilder.PopulateAsync(context);
+            
+                foreach (var contextualizer in context.Response.Contextualizers)
+                {
+                    if (ControllerContext != null)
+                    {
+                        contextualizer(new ContextualizeContext
+                        {
+                            ServiceProvider = _serviceProvider,
+                            Url = Url
+                        });
+                    }
+                }
+            });
 
             return Content(document.ToString(), "text/xml");
         }
--- a/src/OrchardCore.Modules/OrchardCore.Feeds/Controllers/FeedController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Feeds/Controllers/FeedController.cs
@@ -67,9 +67,9 @@ namespace OrchardCore.Feeds.Controllers
             var document = await context.Builder.ProcessAsync(context, async () =>
             {
                 await bestQueryMatch.FeedQuery.ExecuteAsync(context);
-            
+
                 await _feedItemBuilder.PopulateAsync(context);
-            
+
                 foreach (var contextualizer in context.Response.Contextualizers)
                 {
                     if (ControllerContext != null)
--- a/src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListFeedEditViewModel.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListFeedEditViewModel.cs
@@ -4,7 +4,7 @@ namespace OrchardCore.Lists.Feeds
 {
     public class ListFeedEditViewModel
     {
-        public bool EnableFeedProxyUrl { get; set; } = true;
+        public bool DisableRssFeed { get; set; }
 
         public string FeedProxyUrl { get; set; }
 
--- a/src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListFeedEditViewModel.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListFeedEditViewModel.cs
@@ -1,11 +1,15 @@
-﻿using OrchardCore.ContentManagement;
+using OrchardCore.ContentManagement;
 
 namespace OrchardCore.Lists.Feeds
 {
     public class ListFeedEditViewModel
     {
+        public bool EnableFeedProxyUrl { get; set; } = true;
+
         public string FeedProxyUrl { get; set; }
+
         public int FeedItemsCount { get; set; } = ListFeedQuery.DefaultItemsCount;
+
         public ContentItem ContentItem { get; set; }
     }
 }
--- a/src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListFeedQuery.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListFeedQuery.cs
@@ -36,6 +36,14 @@ namespace OrchardCore.Lists.Feeds
                 return null;
             }
 
+            var contentItem = await _contentManager.GetAsync(model.ContentItemId);
+            var feedMetadata = await _contentManager.PopulateAspectAsync<FeedMetadata>(contentItem);
+
+            if (!feedMetadata.EnableFeedProxyUrl)
+            {
+                return null;
+            }
+
             return new FeedQueryMatch { FeedQuery = this, Priority = -5 };
         }
 
--- a/src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListFeedQuery.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListFeedQuery.cs
@@ -39,7 +39,7 @@ namespace OrchardCore.Lists.Feeds
             var contentItem = await _contentManager.GetAsync(model.ContentItemId);
             var feedMetadata = await _contentManager.PopulateAspectAsync<FeedMetadata>(contentItem);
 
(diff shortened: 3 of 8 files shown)


src/OrchardCore.Modules/OrchardCore.Feeds/Controllers/FeedController.cs
```diff
@@ -65,23 +65,23 @@ public async Task<ActionResult> Index(string format)
...
+                    if (ControllerContext != null)
+                    {
+                        contextualizer(new ContextualizeContext
+                        {
+                            ServiceProvider = _serviceProvider,
+                            Url = Url
+                        });
+                    }
+                }
+            });
```
ns8482e, line 84: I don't find anything change in this block other than adding unwanted spaces

hishamco: There was a missing single space for the `document` block, I did it for a reason, but seems 70 and 72 has extra tabs

src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListPartFeedHandler.cs
```diff
@@ -11,8 +11,10 @@ public override Task GetContentItemAspectAsync(ContentItemAspectContext context,
         {
             return context.ForAsync<FeedMetadata>(feedMetadata =>
             {
-                // If the value is not defined, it will be represented as null
+                // Disable the feed if the value is not defined
+                feedMetadata.EnableFeedProxyUrl = part.Content.EnableFeedProxyUrl ?? false;
```
MikeKry, line 15: I think this might influence existing feeds when updating from older version?
They will have null value in this new prop, so it could disable them all? I think it should be true in default

MikeKry:
@hishamco 

Added one comment, otherwise it looks ok. Good work :)

src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListPartFeedHandler.cs
hishamco: Good point, let me think about

hishamco:
@sebastienros the PR is ready to merge

src/OrchardCore/OrchardCore.Feeds.Abstractions/Models/FeedMetadata.cs
```diff
@@ -1,10 +1,12 @@
-﻿namespace OrchardCore.Feeds.Models
+namespace OrchardCore.Feeds.Models
 {
     /// <summary>
     /// Used to gather custom Feed properties for a content item.
     /// </summary>
     public class FeedMetadata
     {
+        public bool EnableFeedProxyUrl { get; set; } = true;
```
sebastienros, line 8: The name here is wrong, the property should be named `DisableRssFeed` such that we "check" the settings to change the default (which is to display the feed).

We need to remove the work `proxy` here because it has nothing to do with proxies.

sebastienros:
@hishamco we made some changes during triage, let us know if you are fine with them

hishamco:
I will check ASAP ..

src/OrchardCore/OrchardCore.Feeds.Abstractions/Models/FeedMetadata.cs
hishamco: This change the intention to disable the feed by default

src/OrchardCore.Modules/OrchardCore.Lists/Views/ListPartFeed.Edit.cshtml
```diff
@@ -18,18 +18,3 @@
     <input asp-for="FeedItemsCount" class="form-control" />
     <span class="hint">@T["The number of elements to display in the feed."]</span>
 </div>
-
-<script at="Foot">
```
hishamco: I'm not sure why its delete!! no need to show the feed URL if the feature is disabled

src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListFeedEditViewModel.cs
```diff
@@ -1,11 +1,15 @@
-﻿using OrchardCore.ContentManagement;
+using OrchardCore.ContentManagement;
 
 namespace OrchardCore.Lists.Feeds
 {
     public class ListFeedEditViewModel
     {
+        public bool DisableRssFeed { get; set; }
```
hishamco, line 7: It woud be better to use `DisableFeed` instead. Coz the feed could be ATOM feed

src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListPartFeedHandler.cs
```diff
@@ -12,7 +12,7 @@ public override Task GetContentItemAspectAsync(ContentItemAspectContext context,
             return context.ForAsync<FeedMetadata>(feedMetadata =>
             {
                 // Enable the feed if the value is not defined. This is handy to migrate the old feeds
-                feedMetadata.EnableFeedProxyUrl = part.Content.EnableFeedProxyUrl ?? true;
+                feedMetadata.DisableRssFeed = part.Content.DisableRssFeed ?? true;
```
hishamco, line 15: This will break the old feeds

src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListFeedQuery.cs
```diff
@@ -39,7 +39,7 @@ public async Task<FeedQueryMatch> MatchAsync(FeedContext context)
             var contentItem = await _contentManager.GetAsync(model.ContentItemId);
             var feedMetadata = await _contentManager.PopulateAspectAsync<FeedMetadata>(contentItem);
 
-            if (!feedMetadata.EnableFeedProxyUrl)
+            if (!feedMetadata.DisableRssFeed)
```
hishamco, line 42: We need to recheck this one

hishamco:
@sebastienros if you fine with above comments I will push the changes

src/OrchardCore/OrchardCore.Feeds.Abstractions/Models/FeedMetadata.cs
sebastienros: No, `EnableXYZ { get; set; } = true;` is the same as `DisableXYZ { get; set; } = false;` in terms of defauts

src/OrchardCore.Modules/OrchardCore.Lists/Views/ListPartFeed.Edit.cshtml
sebastienros: The feed still works when the "disable rss feed" property is checked. Technically we could have another property to actually disable the feed, this one will just remove the link in the HTML, but hitting the url will still work. I was not sure about the change to apply on the JS so for now the safe way was to just remove the hiding logic.

src/OrchardCore.Modules/OrchardCore.Lists/Feeds/ListFeedEditViewModel.cs
sebastienros: But the view actually renders an RSS one
`Type = "application/rss+xml",` and it doesn't really matter

src/OrchardCore.Modules/OrchardCore.Lists/Views/ListPartFeed.Edit.cshtml
```diff
@@ -1,11 +1,16 @@
...
+    <input type="checkbox" class="form-check-input" asp-for="DisableRssFeed">
+    <label class="form-check-label" asp-for="DisableRssFeed">@T["Disable RSS feed link generation."]</label>
+</div>
+
 <div class="mb-3">
     <label asp-for="FeedProxyUrl">@T["Feed proxy Url"]</label>
     <input asp-for="FeedProxyUrl" class="form-control" />
     <span class="hint">@T["Provide a custom public url which will be used to proxy the local rss feed."]</span>
-    <span class="hint">@T["The current feed is available at <a href=\"{0}\" target=\"_blank\">{0}</a>.", Url.Action("Index", "Feed", new { area = "OrchardCore.Feeds", ContentItemId = Model.ContentItem.ContentItemId }, Context.Request.Scheme)]</span>
+    <span id="currentFeedPreview" class="hint">@T["The current feed is available at <a href=\"{0}\" target=\"_blank\">{0}</a>.", Url.Action("Index", "Feed", new { area = "OrchardCore.Feeds", ContentItemId = Model.ContentItem.ContentItemId }, Context.Request.Scheme)]</span>
```
hishamco, line 13:
```suggestion
    <span class="hint">@T["The current feed is available at <a href=\"{0}\" target=\"_blank\">{0}</a>.", Url.Action("Index", "Feed", new { area = "OrchardCore.Feeds", ContentItemId = Model.ContentItem.ContentItemId }, Context.Request.Scheme)]</span>
```

sebastienros (approved):

src/OrchardCore/OrchardCore.Feeds.Abstractions/Models/FeedMetadata.cs
```diff
@@ -1,10 +1,12 @@
-﻿namespace OrchardCore.Feeds.Models
+namespace OrchardCore.Feeds.Models
 {
     /// <summary>
     /// Used to gather custom Feed properties for a content item.
     /// </summary>
     public class FeedMetadata
     {
+        public bool DisableRssFeed { get; set; }
```
hishamco, line 8: We come back to `EnableRssFeed` ;)

```
DisableRssFeed == false <=> EnableRssFeed
```

Issue #12679, opened 2022-10-21
Setup page not work, No error information is recorded when the database connection fails


hyzx86:
First of all, I found that I could not add tenants in my program. I kept prompting that the connection string was incorrect or the server was unavailable. In fact, there was no problem with the database server.
My project uses OC 1.5.0-preview-17106

And Setup page not work.
After filling in all the contents, submit without prompting any error, and still stay on the Setup  page

![image](https://user-images.githubusercontent.com/15613121/197240301-87041513-5488-4770-be3c-5fd984a2e733.png)

To verify this problem, I pulled the latest code and ran the installation program directly using the latest source code. The same problem applies

hyzx86:
Add Tenant
![image](https://user-images.githubusercontent.com/15613121/197241214-1e18ef38-a739-4649-8b44-48a79da9e50d.png)

hyzx86:
Should I write the error information to the log file?
Instead of being ignored when encountering problems?

https://github.com/OrchardCMS/OrchardCore/blob/2385c92dfc3e97324d8031def75e99007c85c80b/src/OrchardCore/OrchardCore.Data.YesSql/DbConnectionValidator.cs#L77-L89

hyzx86:
Solved, after  add Encrypt=false or TrustServerCertificate=True

https://github.com/OrchardCMS/OrchardCore/issues/12653

But I think the error information should be recorded. Not everyone will debug the source code to find problems..

MikeAlhayek:
I am logging the error to provide the admin more info about the error. Can't do much on the UI as this is only impacts SQL Server connection.

hishamco:
Could be related to the TRUSTED connection?

Issue #12657, opened 2022-10-17
YesSqlOptions and DbConnectionValidator concerns.


jtkech:
- `YesSqlOptions` may have different values per tenant, but we use it in `DbConnectionValidator` on a shell that is not yet built so the `.ValidateAsync()` may use wrong values.

    I thought about only using configuration to be able to retrieve the config section of a tenant before it is built, but `YesSqlOptions` not only contains simple fields but also implementations as `TableNameConvention`.

    I started a PR using `IConfigureNamedOptions<YesSqlOptions>` providing options by tenant name but there is another concern if we upgrade the code on an existing application, see below.

- When building an existing tenant we should not use `YesSqlOptions` but the actual values that was used to create, update or setup the tenant. On an existing application if you do the following as documented, you get again the setup screen.

      .ConfigureServices(s => s.Configure<YesSqlOptions>(o => o.TablePrefixSeparator = "__"))

    No problem for e.g. `TablePrefix` that is selectable and then hold by `ShellSettings` but for options coming from `YesSqlOptions` as `TablePrefixSeparator` this is not yet the case.

    Not a big issue, we could also hold all `YesSqlOptions` on `ShellSettings`, but `YesSqlOptions` not only contains simple fields but also implementations as `TableNameConvention`.

    There are solutions, maybe some providers activating implementations by type names, notice that this would also allow to configure all `YesSqlOptions` from configuration sources.

    I can work on it but first I prefer to be sure of what we want.

MikeAlhayek:
@jtkech currently the `YesSqlOptions.TablePrefixSeparator` is stored in each tenant config file. So when the connection is configured for each tenant, then we used the stored value which would provide the correct value.

When we validate a new tenant, then we use the current global value as we should since this global configuration is for defaults. I think the problem with the `DbConnectionValidator` is that it assumes we are only validation new tenants which is wrong. I think it should be changed so that the `ValidateAsync()` accepts a single parameter `DbConnectionValidationContext` which allows the consumer to provide all the value they want to validate. So the `DbConnectionValidationContext` would look something like this

```
public class DbConnectionValidationContext
{
    public string DatabaseProvider { get; set; }
    public string ConnectionString { get; set; }
    public string TablePrefix { get; set; }
    public string ShellName { get; set; }
    public string ShellName { get; set; }
    public string TablePrefixSeparator { get; set; }
    public string Schema { get; set; }  // this is new
[...]

Issue #12653, opened 2022-10-16
After upgrade  getting error : The certificate chain was issued by an authority that is not trusted


giannik:
After recent upgrade with  YesSQL v 3.2.0 (with schema support) I am getting the following error when loading an existing project : 

![image](https://user-images.githubusercontent.com/1516981/196028920-8c8d83db-466c-4888-aded-be1ef4e96b4d.png)

The fix is to append to the connectionstring the setting : 
```
;TrustServerCertificate=True
```
but I am wondering if this could be related to yessql upgrade ?

MikeAlhayek:
I think you need `Encrypt=false` in the connection string 

@sebastienros you may want to shed some light here

jtkech:
You need `TrustServerCertificate=True;`

MikeAlhayek:
Just to clarify, you can use add `Encrypt=false` or `TrustServerCertificate=True` to the connection string to solve this issue. 

The reason this is happening because YesSql is now using the new SQL client which encrypts the connection by default. So if you don't have a trusted certification on the server and want to continue to encrypt the communication using the untrusted server, add `TrustServerCertificate=True` to the connection string. If you want to not encrypt the communication at all, you can use `Encrypt=false`

ns8482e:
This is interesting,  How do we make it self healing on upgrade?

MikeAlhayek:
@ns8482e make sure your server has a valid ssl certificate. I think you would face this issue locally only.

PR #12363, opened 2022-09-09
Refactor DbConnectionValidator

hishamco:

PR #12683, opened 2022-10-21
Add database options like TablePrefixSeparator, Schema, IdentityColumnType, and DocumentTable

MikeAlhayek:
This PR addresses the concerns #12657

Here is a summary of the changes

Both `TableNameConvention` and `TablePrefixSeparator` were removed from `YesSqlOptions`. 
The following properties were added to the `ShellSettings` of each tenant
- `TablePrefixSeparator` when this value does not exists we assume `_` however, if empty string is found, we use no prefix for the tables.
- `Schema` this was added to allow you to provide a specific schema to use.
- `IdentityColumnSize` when this value is empty or invalid, we assume `Int32`
- `DocumentTable` when this value is empty we assume "Document"

If adding a new tenant failed due to missing `Encrypt=false` or `TrustServerCertificate=True`, we add the error to the log file to provide the admin more about the issue. Address issue #12679 and #12653.

@jtkech in the remove tenant code, you can use the new `ITableNameConventionFactory` to create an instance of `ITableNameConvention` with something like this

```C#
var options = DatabaseTableOptions.Create(shellSettingsBeingRemoved);

var tableNameConvention = _tableNameConventionFactory.Create(options);
```
`ITableNameConventionFactory` is a service that we register so you can inject it in `ShellDbTablesInfo` to create the tableNameConvention.

Added the following fields were added to Add/Edit tenant in the tenant UI and also to the setup screen when the database isn't configured.

image

image


hishamco:
Can we merge #12363 before? One more thing I think we discussed before that the above changes could after `1.5`, because the introduced changes might need testing before they got merge. All we know that this time is critical because we are so close to release `1.5` but I will leave this to @sebastienros to decide

I might review this after I finish some of my pending PRs ;)

MikeAlhayek:
@hishamco please hold off of merging PR #12363 since there are few things changes here. As you can see the DbValidation was changed quite bit. I think this should be merged before 1.5. In 1.5 we added `tablePrefixSeparator` which was causing issues so this PR fixes these issue.

Skrypt:
This is getting late for 1.5

jtkech:
@MikeAlhayek Okay, looks now much better ;)

MikeAlhayek:
@jtkech glad you like the change :) If you see no further changes, please approve the PR. Thought on merging this in 1.5 to avoid having to deal with an obsolete YesSqlOptions later?

jtkech:
I will try it this night

jtkech (commented):
As a last comment, so we can select some new options while creating/editing a tenant, then not on the setup form. So how can we define them for the Default tenant on a fresh setup, if not intended that's okay knowing that it is still possible from the regular Configuration.

Yes, some issues quite easy to fix, the main concern being that we can't persist through Shell Settings an empty string, but I prefer this approach even if there are remaining things to fix.

Otherwise LGTM.

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
```diff
@@ -252,6 +256,11 @@ private static async Task<SetupContext> GetSetupContextAsync(TenantSetupOptions
             setupContext.Properties[SetupConstants.DatabaseTablePrefix] = options.DatabaseTablePrefix;
             setupContext.Properties[SetupConstants.SiteName] = options.SiteName;
             setupContext.Properties[SetupConstants.SiteTimeZone] = options.SiteTimeZone;
+            setupContext.Properties[SetupConstants.FeatureProfile] = options.FeatureProfile;
```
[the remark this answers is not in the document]
MikeAlhayek: Yup

src/OrchardCore.Modules/OrchardCore.AutoSetup/Options/TenantSetupOptions.cs
```diff
@@ -81,6 +81,26 @@ public class TenantSetupOptions
...
 
+        /// <summary>
+        /// Gets or sets the database's schema.
+        /// </summary>
+        public string Schema { get; set; }
+
+        /// <summary>
+        /// Gets or sets the database's identity column type.
+        /// </summary>
+        public string IdentityColumnType { get; set; }
```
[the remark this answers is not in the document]
MikeAlhayek: I actually done line the name Size. At some point, I think we'll end up adding string support in YesSql. I think type is a better name for the long run

src/OrchardCore.Modules/OrchardCore.AutoSetup/Options/TenantSetupOptions.cs
```diff
@@ -81,6 +81,26 @@ public class TenantSetupOptions
...
+
+        /// <summary>
+        /// Gets or sets the database's document name.
+        /// </summary>
+        public string DocumentTable { get; set; }
+
+        /// <summary>
+        /// Gets or sets the database's table prefix seperator.
+        /// </summary>
+        public string TablePrefixSeparator { get; set; }
```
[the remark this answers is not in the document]
MikeAlhayek: You are right! I'll rename it

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -346,6 +350,12 @@ public async Task<IActionResult> Create(EditTenantViewModel model)
                 shellSettings["Secret"] = Guid.NewGuid().ToString();
                 shellSettings["RecipeName"] = model.RecipeName;
                 shellSettings["FeatureProfile"] = model.FeatureProfile;
+                // To distinguish between null and empty string, never store null in the shellSettings for TablePrefixSeparator
+                // otherwise old tenants will have invalid table names.
+                shellSettings["TablePrefixSeparator"] = model.TablePrefixSeparator ?? String.Empty;
```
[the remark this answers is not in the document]
MikeAlhayek: You sure about that? I tested it and was able to store empty string. If it is null, then it won't be part of the output. Please double check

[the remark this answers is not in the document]
jtkech: Yes I already checked it many times, if then you setup the tenant you will see that it still uses the '_' separator, not an empty one. Also take care about one of my other comments about the schema that should be passed through e.g. `UseSqlServer()`, otherwise changing it has no effect.

> If it is null, then it won't be part of the output.

It is and then rendered as an empty string

src/OrchardCore/OrchardCore.Data.YesSql/OrchardCoreBuilderExtensions.cs
```diff
@@ -151,21 +151,24 @@ public static OrchardCoreBuilder AddDataAccess(this OrchardCoreBuilder builder)
...
+
             var storeConfiguration = new YesSql.Configuration
             {
                 CommandsPageSize = yesSqlOptions.CommandsPageSize,
                 QueryGatingEnabled = yesSqlOptions.QueryGatingEnabled,
                 ContentSerializer = new PoolingJsonContentSerializer(sp.GetService<ArrayPool<char>>()),
-                TableNameConvention = yesSqlOptions.TableNameConvention
+                TableNameConvention = tableNameFactory.Create(databaseTableOptions),
+                Schema = databaseTableOptions.Schema,
+                IdentityColumnSize = databaseTableOptions.IdentityColumnType,
```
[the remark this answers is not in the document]
hishamco: > I like the comma at the end of the last initializer line, some analyzers suggest it ;)

Still I don't what's the value of it, unless it make it easier when we add a new initialization, super useful for lazy people ;)

src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Create.cshtml
```diff
@@ -1,113 +1,24 @@
...
-        </h6>
-        <div class="row" asp-validation-class-for="FeatureProfile">
-            <div class="mb-3 col-md-6">
-                <select asp-for="FeatureProfile" asp-items="Model.FeatureProfiles" class="form-select"></select>
-            </div>
-        </div>
-    }
+    <partial name="_TenantRecipes" model="Model" />
+    <partial name="_TenantDatabaseInfo" model="Model" />
+    <partial name="_TenantFeatureProfile" model="Model" />
```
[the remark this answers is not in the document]
MikeAlhayek: @jtkech not sure we need to worry about this since the Tenants admin used view/viewmodels there is not display driver being used. so if someone want to override the create/edit views, they would have to define their own fields anyway.

src/OrchardCore.Modules/OrchardCore.Tenants/Views/Shared/_TenantDatabaseInfo.cshtml
```diff
@@ -0,0 +1,81 @@
...
+            <input asp-for="DocumentTable" class="form-control" />
+            <span asp-validation-for="DocumentTable" class="text-danger"></span>
+            <div class="hint">@T["The default value is"] <code>Document</code></div>
+        </div>
+
+        <div class="mb-3 col" asp-validation-class-for="Schema">
+            <label asp-for="Schema">@T["The table schema?"]</label>
+            <input asp-for="Schema" class="form-control" />
+            <span asp-validation-for="Schema" class="text-danger"></span>
+            <div class="hint">@T["When left blank, the default value on the server will be used."] @T["For example, {0} for SQL Server", "dbo"]</div>
```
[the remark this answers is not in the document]
MikeAlhayek: `obo` should not be translated. but I changed it to this `@T["For example, '{0}' for SQL Server", "dbo"]`

src/OrchardCore/OrchardCore.Data.YesSql/OrchardCoreBuilderExtensions.cs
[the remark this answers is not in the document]
MikeAlhayek: @hishamco adding comma at the end will prevent you from seeing red line when you are looking at the diff since you never have to touch the previous line. I also like using the training comma when possible

MikeAlhayek:
@jtkech I made all the changes including adding the same value to the API and to the workflow activities.

jtkech:
Okay cool, I will look at it asap.

src/OrchardCore.Modules/OrchardCore.AutoSetup/Options/TenantSetupOptions.cs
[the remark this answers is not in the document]
jtkech: Okay interesting but how existing services using predicate as `Id > @param` would work? for example related to the `AutoroutePartIndex` and `IndexingTask` tables.

jtkech:
Only did a quick review but LGTM.

Yes for now the `NULL` workaround seems okay when binding from a create/edit/setup form, because we know that the separator field is there and that a value is provided even if it is binded to null.

But in other places we may need to respect the existing usage of APIs, already defined WFs and existing AutoSetup configs, meaning if nothing is changed the `_` separator is still intended to be used.

Yes, this is the only remaining thing I still have to think about.

src/OrchardCore.Modules/OrchardCore.AutoSetup/Options/TenantSetupOptions.cs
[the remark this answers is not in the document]
MikeAlhayek: That is a different battle for a different day :)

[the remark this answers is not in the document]
jtkech: Okay, will see tomorrow ;)

No problem, was just curious, I'm fine with that name.

MikeAlhayek:
@jtkech I made the last suggested changes. Let me know if you see any last items here. Please check the update I made to the workflow and the api controller to make sure I did what you suggested.

jtkech:
@MikeAlhayek

I did a review, sorry if too long but one thing imply two others (exponential), maybe some parts could be managed in separate PR(s). One main point would be to manage new database fields like the existing ones as much as possible, which would be easier if we don't support empty separators, or if all binding sources (behaving dfferently) should provide the same dedicated value for this purpose.

src/OrchardCore/OrchardCore.Abstractions/Setup/SetupConstants.cs
```diff
@@ -11,5 +11,10 @@ public static class SetupConstants
         public const string DatabaseConnectionString = "DatabaseConnectionString";
         public const string DatabaseTablePrefix = "DatabaseTablePrefix";
         public const string SiteTimeZone = "SiteTimeZone";
+        public const string Schema = "Schema";
+        public const string IdentityColumnSize = "IdentityColumnSize";
+        public const string FeatureProfile = "FeatureProfile";
+        public const string DocumentTable = "DocumentTable";
+        public const string TableNameSeparator = "TableNameSeparator";
```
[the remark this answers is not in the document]
MikeAlhayek: Not a good way of these values are stored later like in permission name, migration, saving the value as a key in the settings file…. But yes when the value is not stored, it’s a good idea

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -448,18 +471,29 @@ public async Task<IActionResult> Edit(EditTenantViewModel model)
...
-                    shellSettings["DatabaseProvider"] = model.DatabaseProvider;
-                    shellSettings["TablePrefix"] = model.TablePrefix;
-                    shellSettings["ConnectionString"] = model.ConnectionString;
+
                     shellSettings["RecipeName"] = model.RecipeName;
                     shellSettings["Secret"] = Guid.NewGuid().ToString();
+                    shellSettings["Schema"] = model.Schema?.Trim();
+                    shellSettings["TableNameSeparator"] = model.TableNameSeparator?.Trim() ?? "NULL";
+                    shellSettings["IdentityColumnSize"] = model.IdentityColumnSize.ToString();
+                    shellSettings["DocumentTable"] = model.DocumentTable.Trim();
```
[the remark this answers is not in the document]
MikeAlhayek: I'll add it but null here will never pass the validation which is why it safe

jtkech:
Okay cool no problem

Skrypt:
This is not mandatory for 1.5 as we documented the issue with the connection string. Though, I'm going to discuss it with @jtkech before making any decision.

MikeAlhayek:
@Skrypt no worries. I removed the 1.5 milestone from this PR so we can release 1.5 asap. We'll just have to introduce this as a breaking change in 1.6. Maybe we should put in a quick PR in 1.5 that mark `public ITableNameConvention TableNameConvention { get; set; }` and `public string TablePrefixSeparator { get; set; } = "_";` in 1.5 as obsolete in YesSqlOptions just so no one would use them.

either way, the breaking change should have a minimal impact as I double people will use these options since we are not going to document them.

jtkech:
@MikeAlhayek don't worries too much, maybe just a warning in the doc, but what I experienced with a custom config is that the setup runs again, but when this config is removed it restarts as before.

@Skrypt I worked on it and I think it is now ready to be reviewed and tried, but yes maybe too late e.g. if we want to push the release today, maybe not if we push it the next week.

Only one little thing, `DbConnectionValidatorResult` now makes the difference between 2 tenants having the same `TablePrefix` but not the same `Schema` and/or not the same `TableNameSeparator` which is good. The only little problem is about the `DocumentTable` name that we now can change. So if the `DocumentTable` is not found it doesn't mean that other tables are not already there, but I think that's okay.

Just for info about the `IdentityColumnSize` enum, on YesSql we have `Int32 = 1` and `Int64 = 2`, so if not initialized the default is 0 represented by `"0"`, then on YesSql if not `Int32` it fallbacks to `Int64`. Notice that `Enum.TryParse()` returns true on any integer string as `"3"`. So I needed to use `Enum.IsDefined()`.

jtkech (approved):

src/OrchardCore.Modules/OrchardCore.Tenants/Views/Admin/Create.cshtml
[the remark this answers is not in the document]
jtkech: Okay no problem, so just to say that in a theme, yes we can override a pure mvc view, and if this view trigger a shape with the tag helper, the theme can also override this shape even if not driven by a driver, here by shape type if no alternates was added, which can also be done in the original razor view (also in a liquid view).

Yes I was thinking about someone that just want to not render the new options in the form.

But no problem that's okay, I'm just not used to see `<partial>` in our razor views ;)

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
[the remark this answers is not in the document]
jtkech: I had the case when first created a tenant from a Workflow then edited it and saved. Yes the current regular expression allows an empty value which is then bound to a null value, but that's okay as we then fallback to `Document` later on, and it allows to not have this field in a given data source, e.g. an API model / form / AutoSetup config still not providing this field, but then here we need a null check.

Anyway as I remember here I don't use the `Trim()` anymore, only do it in one place.

sebastienros:
I am concerned with being able to change the document table name and table separator. I would prefer to have these in a configuration file, and not be able to change it per tenant. Schema has to be changed per tenant, if not already set in the configuration or the presets. ID length should also go in the configuration file. 

Also concerned by the security of being able to change part of the query from user entered information (Document table and separator) even if a connection string can be provided, as it could be password-less (impersonation, Key vault based). We can ensure what OC does to the database, and here we would accept user entries to be added to SQL queries (not sure how much it's validated, I don't think YesSql does any validation since it's code-based). 

My suggestion would be to just remove the UI for these 3 fields and have it be from file. The schema option being displayed in preset if not in configuration, and only in setup screen if not defined in preset.

@Piedone ?

urbanit:
I suggest setup screen remain minimal.
Advanced configurations like these should be handled elsewhere - recipe? yaml configuration file?

MikeAlhayek:
@urbanit setup screen is minimal when the connection is preconfigured/set. If the tenant's connection isn't configured, we allow all the options to be editable. 

@sebastienros how about we add a checkbox "Advanced Options" and when checked we can show these options? So by default we collapse them until the users wants to see more.

Piedone:
I would also keep these options off the setup screen, even if it's not pre-configured.

As for the tenant settings editor, I think that's OK, but in addition to environmental configuration (like appsettings, env vars), and the fields would only show if the environment is not already configured. See how these won't show if you have `OrchardCore__DatabaseProvider` or `OrchardCore__ConnectionString` configured, for example.

MikeAlhayek:
@jtkech i'll try to carve out time to complete this PR. I don't know if there is clear guidance on what needs to be done. If we take these options from the setup page, do we leave them on the add new tenant page? If not, how would one set these settings when adding a new tenant?

MikeAlhayek:
@sebastienros @jtkech I carved out time now to work on this. But unfortunately I do not know what we need to change here.

I think the happy medium here to only remove the changes from the Setup screen while leaving everything in the add tenant screen, API and workflow.

Let me know if you guys are good with this before I make any more changes to this PR.

jtkech:
@sebastienros @MikeAlhayek 

First some thoughts I was just thinking about.

- One thing we need to be aware is that we store a tenant setting in a tenant specific source (e.g. appsettings.json under the tenant folder) only if this setting has a different value than the one provided (or not provided) by a previous source (e.g. the application appsettings.json).

- Just to say that there are already ways to "break" an existing tenant when a setting is not persisted in a tenant specific source but provided e.g from the main appsettings, this by changing its value later on in the main appsettings (globaly of under the section of a given tenant).

- We don't see this for `TablePrefix` as we usually persist different prefixes in tenant specific sources, but wth the new `Schema` setting it will be possible to have multiple tenants without any prefix, so it will be possible to "break" them by providing afterwards e.g. a global prefix from the main appsettings.

- For example we are already used to provide from the main source global database settings including a connection string, so we can already "break" existing tenants by changing it. But that's okay, this is how works the regular configuration stack on which we added mutable tenant specific sources.

- So just to be aware of this but I think that's okay, when using config sources people have to know what they are doing. For example the configuration stack allows to provide different connection strings based on the current environment e.g. by also using `appsettings.TheCurrentEnv.json` files.

---

That said what I understood (here I don't talk about configurable factories through `IOptions`).

- Unless for `Schema`, for all other settings different values than the defaults can only be provided by config sources. So not editable (tenant and WF editors) and can't be provided by an API payload.

- The `Schema` can be provided by any config source (as other settings) and is also editable (tenant and WF editors) and can be provided by an API payload, it just has to be managed as `TablePrefix`.

MikeAlhayek:
@jtkech So what do you suggest we change here?

jtkech:
I'm in a meeting, I will comment it soon

sebastienros:
Identifier type; Schema; Table separator; Document table name => in app settings, used when creating a new tenant, and each tenant has these values stores in shell settings.

Schema UI is added wherever Table prefix is already visible.

jtkech:
@MikeAlhayek @sebastienros

No problem I will complete it very soon.

> and each tenant has these values stores in shell settings.

Just to be aware that if you provide a pre-configured `Schema` from the main appsettings and don't specify a different value when creating a new tenant, it will not be saved in the tenant specific appsettings.

So the pre-configured value should not be changed afterwards, or intentionally (e.g. based on the env).

jtkech:
@TFleury for info in #12826 you removed the dependency of `Setup.Core` on `YesSql.Abstractions` by moving `IDbConnectionValidator` to `Data.Abstractions` but here this interface uses a new context param => new options using the `IdentityColumnSize` enum type defined by `YesSql`.

I can make it building by moving other files, defining locally the same `IdentityColumnSize` enum, using this local enum in options and models and a casting in the final usage.

But just saw that `Setup.Core` also depends on `Recipes.Abstractions` that depends on `YesSql.Abstractions`, So it seems that `Setup.Core` still depends on `YesSql.Abstractions`.

So for simplicity and to continue to work on this PR, for now I just moved back `IDbConnectionValidator` in `Data.YesSql.Abstractions`.

Let me know if that's okay for you or if we will have to also remove the dependency of `Recipes.Abstractions` on `YesSql.Abstractions` here or in a future separate PR.

TFleury:
@jtkech thank you for making me aware of this PR.

For me it's OK to have these options in ShellSettings but not in setup UI, because they are quite advanced, specific to YesSql, and maybe with some security concerns.

> But just saw that `Setup.Core` also depends on `Recipes.Abstractions` that depends on `YesSql.Abstractions`, So it seems that `Setup.Core` still depends on `YesSql.Abstractions`.

I know there are already many dependencies over YesSql in many places and I'm trying to remove them as much as possible, so please don't add new ones.

> Let me know if that's okay for you or if we will have to also remove the dependency of `Recipes.Abstractions` on `YesSql.Abstractions` here or in a future separate PR.

Yes, I think we will have to remove it.

> Note: Nothing urgent, because e.g. `IdentityColumnSize` will not be updatable through the UI, maybe at least I will be able easily to move back again `IDbConnectionValidator` to `Data.Abstractions`.

Yes, please move it back to `Data.Abstractions` to keep Setup, Recipies and Tenants independent from data provider implementation.

jtkech:
@TFleury Okay

jtkech:
@TFleury Okay I was able to keep `IDbConnectionValidator` in `Data.Abstractions`.

@sebastienros @MikeAlhayek 

Okay only `Schema` is now editable and managed as `TablePrefix`.

> and each tenant has these values stores in shell settings.

This was my last concern, as said e.g. for `TableSeparator` if we don't provide a different value (e.g. not editable), it will not be persisted in the tenant specific appsettings. So to be used as a preset but only while creating a new tenant we need to use another config as `DefaultTableSeparator`.

We can then use `DefaultTableSeparator` while creating a new tenant to set its `TableSeparator` that will be persisted (if not null) in the tenant specific appsettings. This by still allowing to pre-configure globally `TableSeparator` directly (higher precedence), e.g. to have different values but based on the current env.

I'm working on this last part.

jtkech:
Just for info

So when we create a tenant from the UI, API or WF we call `UpdateShellSettings()`, so easy to mutate just before the shell settings with the new configured database presets (that are not editable).

The problem is that the Default tenant is auto created before a setup by the `ShellHost` which is not aware of these configured database presets.

So maybe a shellSettings update handler but for now I opted to persist these configured database presets only when a tenant is setup, this is easier as it can be done in one central place, and it makes sense to be able to change these configured presets until a tenant is setup successfully.

MikeAlhayek:
@jtkech did you finish all the changes here? I can test it out again once you are done.

jtkech:
Okay no problem, not fully ready, I will let you know soon.

jtkech:
For info

Finally I simplfied the possible configs, I only documented how to configure the presets that are only used before a given tenant is setup. For example `DefaultDocumentTable` for now under the section `OrchardCore_Data_DatabaseTable`.

These presets are used before a setup, on setup they are persisted in the tenant specific config, but by using dedicated keys e.g. `internal_DocumentTable` not intended to be pre-configured from a previous config source. This to ensure that on any setup a valid value is persisted in the tenant specific config.

Doing so it was easier e.g. for the `IdentityColumnSize` to default to `Int64` for any new tenant, and for others to keep `Int32` if nothing was persisted in their specific config, meaning they are old tenants.

---

I also did some tweaks in the tenant validator, e.g. we were assuming that a null url prefix was not the same as an empty one, but both end up to no prefix, I also made host urls checking case insensitive. Hmm, we could also have allowed a non Default tenant to have both an empty host and url prefix, as before because the Default tenant can have a non empty url prefix, but not done as maybe not recommended.

---

Will be ready soon to be reviewed and tried.

jtkech:
@MikeAlhayek 

Didn't have so much time, may need some last tweaks, and not re-tried after the last changes.

But I think it it ready to be tested, here, as documented, how to configure the db table presets.

Schema is managed differently as it is editable, it is managed as TablePrefix.

    "OrchardCore_Data_DatabaseTable": {
      "DefaultDocumentTable": "MyDocument",
      "DefaultTableNameSeparator": "____",
      "DefaultIdentityColumnSize": "Int32"
    }

Just as a reminder, here we want to pre-configure some settings and ensure that they are stored per tenant in their specific mutable config source, even if these settings are not editable.

Also for the IdentityColumnSize, if nothing is configured we want to default to Int64 (the default of the default) for any new tenants but default to Int32 for old existing tenants (created before these changes).

jtkech:
Okay, just saw the last meeting, looks like what I did is in sync with what was suggested.

src/OrchardCore.Cms.Web/appsettings.json
```diff
@@ -15,9 +15,16 @@
     //"OrchardCore_ContentLocalization_CulturePickerOptions": {
     //  "CookieLifeTime": 14 // Set the culture picker cookie life time (in days).
     //},
+    // See https://docs.orchardcore.net/en/latest/docs/reference/core/Data/#sqlite.
     //"OrchardCore_Data_Sqlite": {
     //  "UseConnectionPooling": false
     //},
+    // See https://docs.orchardcore.net/en/latest/docs/reference/core/Data/#database-table to configure database table presets used before a given tenant is setup.
+    //"OrchardCore_Data_DatabaseTable": {
```
MikeAlhayek, line 23: I suggest we name this `OrchardCore_Data_TableOptions` instead, Also I suggest put the default value in the commented out seeing like we do in everything else so something like this instead 
```
//"OrchardCore_Data_TableOptions": {
    //  "DefaultDocumentTable": "Document", // Document table name, defaults to 'Document'.
    //  "DefaultTableNameSeparator": "_", // Table name separator, one or multiple '_', "NULL" means no separator, defaults to '_'.
    //  "DefaultIdentityColumnSize": "Int64" // Identity column size, 'Int32' or 'Int64', defaults to 'Int64'.
    //},
```

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -460,24 +464,25 @@ public async Task<IActionResult> Edit(EditTenantViewModel model)
                 return RedirectToAction(nameof(Index));
             }
 
+            // If we got this far, something failed, reinitialize the model and redisplay form
```
MikeAlhayek, line 467:
```suggestion
            // If we got this far, something failed. Reinitialize the model and re-display form.
```

src/docs/reference/core/Data/README.md
```diff
@@ -26,27 +26,46 @@ See the [`Microsoft.Data.Sqlite` documentation](https://docs.microsoft.com/en-us
...
+| Setting | Description |
+| --- | --- |
+| `DefaultDocumentTable` | Document table name, defaults to 'Document'. |
+| `DefaultTableNameSeparator` | Table name separator, one or multiple '_', "NULL" means no separator, defaults to '_'. |
+| `DefaultIdentityColumnSize` | Identity column size, 'Int32' or 'Int64', defaults to 'Int64'. |
+
+##### `appsettings.json`
+
+```json
+  "OrchardCore_Data_DatabaseTable": {
```
MikeAlhayek, line 62: Please refer to my comment about about suggestion to change the name here and the default values.

MikeAlhayek (commented):
@jtkech

I tested out all the new changes you added to this PR. All looks good to me and it also falls inline with all the feedback we had during the meeting. I have some minor suggestions here.

src/OrchardCore.Cms.Web/appsettings.json
jtkech: Okay

src/docs/reference/core/Data/README.md
jtkech: Okay I will look at it

jtkech:
@MikeAlhayek 

Okay done but wait, maybe some very last little simplifications

MikeAlhayek:
@jtkech thanks. We'll check it tomorrow and ensure there is no more feedback from @sebastienros tomorrow before we merge it.

sebastienros (approved):

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
```diff
@@ -10,6 +10,7 @@
 using OrchardCore.Abstractions.Setup;
 using OrchardCore.AutoSetup.Extensions;
 using OrchardCore.AutoSetup.Options;
+using OrchardCore.Data;
```
hishamco, line 13: Why this is added?

src/OrchardCore.Modules/OrchardCore.AutoSetup/Options/TenantSetupOptions.cs
```diff
@@ -153,6 +158,11 @@ public virtual IEnumerable<ValidationResult> Validate(ValidationContext validati
             {
                 yield return new ValidationResult(String.Format(RequiredErrorMessageFormat, nameof(SiteTimeZone)));
             }
+
+            if (!String.IsNullOrWhiteSpace(DatabaseSchema) && DatabaseSchema.Any(c => !Char.IsLetterOrDigit(c)))
+            {
+                yield return new ValidationResult("The 'DatabaseSchema' should only contain alpha-numeric chars.");
```
hishamco, line 164:
```suggestion
                yield return new ValidationResult($"The '{nameof(DatabaseSchema)}' should only contain alpha-numeric chars.");
```

src/OrchardCore.Modules/OrchardCore.Setup/Controllers/SetupController.cs
```diff
@@ -69,7 +69,7 @@ public async Task<ActionResult> Index(string token)
                 DatabaseProviders = _databaseProviders,
                 Recipes = recipes,
                 RecipeName = defaultRecipe?.Name,
-                Secret = token
+                Secret = token,
```
hishamco, line 72:
```suggestion
                Secret = token
```

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -281,28 +281,32 @@ public async Task<IActionResult> Create()
...
                 RequestUrlPrefix = shellSettings.RequestUrlPrefix,
-                TablePrefix = shellSettings["TablePrefix"],
                 RecipeName = shellSettings["RecipeName"],
+                DatabaseProvider = shellSettings["DatabaseProvider"],
+                ConnectionString = shellSettings["ConnectionString"],
+                TablePrefix = shellSettings["TablePrefix"],
+                Schema = shellSettings["Schema"],
                 FeatureProfile = currentFeatureProfile,
-                FeatureProfiles = featureProfiles
+                FeatureProfiles = featureProfiles,
```
hishamco, line 304:
```suggestion
                FeatureProfiles = featureProfiles
```

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -331,17 +335,17 @@ public async Task<IActionResult> Create(EditTenantViewModel model)
             {
                 // Creates a default shell settings based on the configuration.
                 var shellSettings = _shellSettingsManager.CreateDefaultSettings();
-
+
```
hishamco, line 338:
```suggestion

```

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
```diff
@@ -390,7 +394,7 @@ public async Task<IActionResult> Edit(string id)
                 RequestUrlHost = shellSettings.RequestUrlHost,
                 RequestUrlPrefix = shellSettings.RequestUrlPrefix,
                 FeatureProfile = currentFeatureProfile,
-                FeatureProfiles = featureProfiles
+                FeatureProfiles = featureProfiles,
```
hishamco, line 397:
```suggestion
                FeatureProfiles = featureProfiles
```

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs
```diff
@@ -185,9 +186,7 @@ public async Task<ActionResult> Setup(SetupApiViewModel model)
                 return BadRequest(S["The database provider is not defined."]);
             }
 
-            var selectedProvider = _databaseProviders.FirstOrDefault(provider =>
-                String.Equals(provider.Value, databaseProvider, StringComparison.OrdinalIgnoreCase));
-
+            var selectedProvider = _databaseProviders.FirstOrDefault(provider => provider.Value == databaseProvider);
```
hishamco, line 189: Why the casing in the comparison is removed

src/OrchardCore.Modules/OrchardCore.Tenants/ViewModels/SetupApiViewModel.cs
```diff
@@ -27,8 +27,12 @@ public class SetupApiViewModel
         public string Password { get; set; }
 
         public string RecipeName { get; set; }
+
         public IFormFile Recipe { get; set; }
 
         public string SiteTimeZone { get; set; }
+
+        public string Schema { get; set; }
+
```
hishamco, line 36:
```suggestion

```

src/OrchardCore.Modules/OrchardCore.Tenants/Views/Shared/_TenantDatabaseInfo.cshtml
```diff
@@ -0,0 +1,49 @@
+@model EditTenantViewModel
```
hishamco, line 1: I'm not sure why the partials relying on `EditTenantViewModel`, it would be nice to use the UI components in both Create & Edit views

src/OrchardCore.Modules/OrchardCore.Tenants/Workflows/Activities/CreateTenantTask.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Localization;
+using OrchardCore.Data;
```
hishamco, line 5: Why?

src/OrchardCore/OrchardCore.Data.Abstractions/ShellSettingsExtensions.cs
```diff
@@ -0,0 +1,97 @@
...
+            : shellSettings["IdentityColumnSize"])
+            ?.Trim();
+
+        if (String.IsNullOrEmpty(identityColumnSize))
+        {
+            identityColumnSize = !shellSettings.IsInitialized() ? nameof(Int64) : nameof(Int32);
+        }
+        else if (!_identityColumnSizes.Contains(identityColumnSize))
+        {
+            throw new InvalidOperationException($"The configured identity column size '{identityColumnSize}' is invalid.");
```
hishamco, line 88:
```suggestion
            throw new InvalidOperationException($"The configured identity column size '{inameof(dentityColumnSize)}' is invalid.");
```

hishamco (changes requested):
May be I'm late to review this, but I suffered last two weeks from a damaged hard disk in my previous laptops. I just write down some comments that might be consider

jtkech:
Please wait before merging, need to check the last review / commits, at least fix unit tests.

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
jtkech: We needed it at some point, I think we can remove it, also an additional project ref in the csproj file.

I will check it

src/OrchardCore.Modules/OrchardCore.AutoSetup/Options/TenantSetupOptions.cs
jtkech: Because not yet done in so many other related places

src/OrchardCore.Modules/OrchardCore.Setup/Controllers/SetupController.cs
jtkech: I don't agree

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/AdminController.cs
jtkech: I don't agree, there are analysers that suggests it

src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs
jtkech: Because we check a technical value

src/OrchardCore.Modules/OrchardCore.Tenants/Workflows/Activities/CreateTenantTask.cs
jtkech: Idem, at some point was needed, can be removed

src/OrchardCore/OrchardCore.Data.Abstractions/ShellSettingsExtensions.cs
jtkech: idem, for sure I thought about it, at least the interpolation is no more necessary if we still don't use nameof.

jtkech: Here it's wrong as already we refer to an existing var, so should be kept as is

MikeAlhayek:
@jtkech no worries. Either approve or merge it when you are done making changes to it. Thank you

jtkech:
I already approved it a while ago ;) I'm merging it

PR #12689, opened 2022-10-22
Keep IDisplayManager backward compatible

### LINKED ISSUE #12678: Breaking changes in DisplayManager in 1.5
### Describe the bug

I am using `BuildEditor` of `DisplayManager` to create a shape, It's working well, but when upgraded to 1.5 preview feed started getting following error.

image

ns8482e:
Added method overload to avoid method not found exception in modules that depends on IDisplayManager
//cc @jtkech @MikeAlhayek @hishamco 

Fixes #12678

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.AdminMenu/Controllers/NodeController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.AdminMenu/Controllers/NodeController.cs
@@ -111,7 +111,7 @@ namespace OrchardCore.AdminMenu.Controllers
                 AdminNode = treeNode,
                 AdminNodeId = treeNode.UniqueId,
                 AdminNodeType = type,
-                Editor = await _displayManager.BuildEditorAsync(treeNode, updater: _updateModelAccessor.ModelUpdater, isNew: true)
+                Editor = await _displayManager.BuildEditorAsync(treeNode, updater: _updateModelAccessor.ModelUpdater, isNew: true, "", "")
             };
 
             return View(model);
@@ -140,7 +140,7 @@ namespace OrchardCore.AdminMenu.Controllers
                 return NotFound();
             }
 
-            dynamic editor = await _displayManager.UpdateEditorAsync(treeNode, updater: _updateModelAccessor.ModelUpdater, isNew: true);
+            dynamic editor = await _displayManager.UpdateEditorAsync(treeNode, updater: _updateModelAccessor.ModelUpdater, isNew: true, "", "");
             editor.TreeNode = treeNode;
 
             if (ModelState.IsValid)
@@ -189,7 +189,7 @@ namespace OrchardCore.AdminMenu.Controllers
                 AdminNodeType = treeNode.GetType().Name,
                 Priority = treeNode.Priority,
                 Position = treeNode.Position,
-                Editor = await _displayManager.BuildEditorAsync(treeNode, updater: _updateModelAccessor.ModelUpdater, isNew: false)
+                Editor = await _displayManager.BuildEditorAsync(treeNode, updater: _updateModelAccessor.ModelUpdater, isNew: false, "", "")
             };
 
             model.Editor.TreeNode = treeNode;
@@ -220,7 +220,7 @@ namespace OrchardCore.AdminMenu.Controllers
                 return NotFound();
             }
 
-            var editor = await _displayManager.UpdateEditorAsync(treeNode, updater: _updateModelAccessor.ModelUpdater, isNew: false);
+            var editor = await _displayManager.UpdateEditorAsync(treeNode, updater: _updateModelAccessor.ModelUpdater, isNew: false, "", "");
 
             if (ModelState.IsValid)
             {
--- a/src/OrchardCore/OrchardCore.DisplayManagement/IDisplayManager.cs
+++ b/src/OrchardCore/OrchardCore.DisplayManagement/IDisplayManager.cs
@@ -8,13 +8,13 @@ namespace OrchardCore.DisplayManagement
         Task<IShape> BuildDisplayAsync(TModel model, IUpdateModel updater, string displayType = "", string groupId = "");
 
         Task<IShape> BuildEditorAsync(TModel model, IUpdateModel updater, bool isNew, string groupId = "")
-            => BuildEditorAsync(model, updater, isNew,  groupId, "");
-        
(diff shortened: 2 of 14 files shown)


MikeAlhayek:
@ns8482e I don't know how I feel about this change. your overloading the interface. What if you add the previous overload with as an extension method with the same name space as the interface?

Something like this

```
public static class DisplayManagerExtensions
{
	public static Task<IShape> BuildEditorAsync(this IDisplayManager displayManager, TModel model, IUpdateModel updater, bool isNew, string group = null)
		=> displayManager.BuildEditorAsync(model, updater, isNew, group, String.Empty);
	

	public static Task<IShape> UpdateEditorAsync(this IDisplayManager displayManager, TModel model, IUpdateModel updater, bool isNew, string groupId = "")
            => displayManager.UpdateEditorAsync(model, updater, isNew, groupId, String.Empty, String.Empty);
}
```

ns8482e:
@MikeAlhayek i am not overwriting interface I am reverting change made to interface in PR #11775  to keep interface unchanged between 1.4 and 1.5

MikeAlhayek:
You are adding 2 new signatures to the interface. If anyone is implementing this interface will have the same problem you started with. Except their app will break since the new methods are not implemented. Look into using extensions as I mentioned last and see if that'll fix your issue.

ns8482e:
> What if you add the previous overload with as an extension method with the same name space as the interface?

Extension won’t help as it still requires existing module to recompile, the module compiled against 1.4 expects method in interface

ns8482e:
> You are adding 2 new signatures to the interface. If anyone is implementing this interface will have the same problem you started with. 

we are adding new method in interface in 1.5, if someone is implementing their own flavor of DisplayManager then they will have to implement new methods added  in interface in 1.5

However the real issue here is consumer of IDisplayManager which is injected in client code that using OC provided DisplayManager implementation.

jtkech (approved):

Skrypt (approved):

src/OrchardCore/OrchardCore.DisplayManagement/IDisplayManager.cs
```diff
@@ -6,7 +6,15 @@ namespace OrchardCore.DisplayManagement
     public interface IDisplayManager<TModel>
     {
         Task<IShape> BuildDisplayAsync(TModel model, IUpdateModel updater, string displayType = "", string groupId = "");
-        Task<IShape> BuildEditorAsync(TModel model, IUpdateModel updater, bool isNew, string groupId = "", string htmlPrefix = "");
-        Task<IShape> UpdateEditorAsync(TModel model, IUpdateModel updater, bool isNew, string groupId = "", string htmlPrefix = "");
+
+        Task<IShape> BuildEditorAsync(TModel model, IUpdateModel updater, bool isNew, string groupId = "")
```
sebastienros, line 10: Can we make these obsolete such that old compiled libraries use the obsolete methods, but new code knows it should use the new methods.

ns8482e: No need to make it obsolete as its just an overloading

ns8482e: Also you can't `obsolete` exiting and add new method with default param as you can't compile the code as compiler complains about ambiguous error

sebastienros: Seems to build with this:

```c#
using System;
using System.Threading.Tasks;
using OrchardCore.DisplayManagement.ModelBinding;

namespace OrchardCore.DisplayManagement
{
    public interface IDisplayManager<TModel>
    {
        Task<IShape> BuildDisplayAsync(TModel model, IUpdateModel updater, string displayType = "", string groupId = "");
        Task<IShape> BuildEditorAsync(TModel model, IUpdateModel updater, bool isNew, string groupId, string htmlPrefix);
        Task<IShape> UpdateEditorAsync(TModel model, IUpdateModel updater, bool isNew, string groupId, string htmlPrefix);

        [Obsolete]
        Task<IShape> BuildEditorAsync(TModel model, IUpdateModel updater, bool isNew, string groupId = "")
            => BuildEditorAsync(model, updater, isNew, groupId, "");

        [Obsolete]
        Task<IShape> UpdateEditorAsync(TModel model, IUpdateModel updater, bool isNew, string groupId = "")
            => UpdateEditorAsync(model, updater, isNew, groupId, "");
    }
}

```

There are warning that can be fixed by using the new methods explicitly.

hishamco: @MikeAlhayek which if them you added in `1.4`?

MikeAlhayek: The once’s with the prefix

hishamco: I see, I though the obsolete once

ns8482e: @sebastienros agree, updated as suggested

MikeAlhayek (approved):

ns8482e:
@sebastienros Removed warnings

src/OrchardCore.Modules/OrchardCore.AdminMenu/Controllers/NodeController.cs
```diff
@@ -220,7 +220,7 @@ public async Task<IActionResult> Edit(AdminNodeEditViewModel model)
                 return NotFound();
             }
 
-            var editor = await _displayManager.UpdateEditorAsync(treeNode, updater: _updateModelAccessor.ModelUpdater, isNew: false);
+            var editor = await _displayManager.UpdateEditorAsync(treeNode, updater: _updateModelAccessor.ModelUpdater, isNew: false, "", "");
```
MikeAlhayek, line 223: Can you please use `String.Empty` instead of `""` in all these calls?

MikeAlhayek: 2 seconds too late :)

ns8482e: @ns8482e I'll fix that in dotnet format #12235

PR #12691, opened 2022-10-23
Indexing Latest not working as intended

### LINKED ISSUE #12690: Indexing Latest not working as intended (breaking)
### Describe the bug
1. Rebuild an index with latest items doesn't work and removes all indexed items.
2. Saving items (publish or draft) will index all versions (won't just keep the latest).

### To Reproduce
1. Go to a new project with Blog Recipe
2. Create an index with "Index latest version" option
3. Save an article (publish or draft)
4. Query the index, the article is there (as it should)
5. Rebuild the index, the article is no longer there (breaking issue!)

Also if you save the article multiple times, it will add a version to the index each time.
I believe this was not the case before. Only the latest was kept.

I did some debugging and found out that it never entered the 2nd if as the `contentItem` is always null.
Couldn't figure out why yet

https://github.com/OrchardCMS/OrchardCore/blob/411206031d8d3e7626ed2f80c50c9bbfce958c9d/src/OrchardCore.Modules/OrchardCore.Search.Lucene/Services/LuceneIndexingService.cs#L169-L178

@Skrypt can you please take a look? IMO this is blocking for 1.5 release

Skrypt:
Fixes #12690

This should do it. I thought a StoredField would be enough to find these documents in the indices but seems like it is not.
The other issue is that I refactored the LuceneIndexManager to make it a little faster and made a mistake on the value that should be stored in the dictionary to retrieve a Draft. Seems like we never really use the ContentItemVersionId when indexing. We just take the latest one or the published one and we prune any document indexed related to a ContentItemId before updating with new documents.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Search.Lucene/Services/LuceneIndexManager.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Search.Lucene/Services/LuceneIndexManager.cs
@@ -199,14 +199,17 @@ namespace OrchardCore.Search.Lucene
             var doc = new Document
             {
                 // Always store the content item id and version id
-                new StoredField("ContentItemId", documentIndex.ContentItemId.ToString()),
-                new StoredField("ContentItemVersionId", documentIndex.ContentItemVersionId.ToString())
+                // These fields need to be indexed as a StringField because it needs to be searchable for the writer.DeleteDocuments method.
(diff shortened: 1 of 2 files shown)


Skrypt:
I won't be able to merge that one either. There is an issue with the licence/cla with every new PR.
@sebastienros @agriffard

PR #12709, opened 2022-10-24
Make Themes always enabled

### LINKED ISSUE #12707: Exception due to missing `ThemeTogglerService` when disabling the Themes module
Hello, I was playing around with the features switches and by disabling one of them (not entirely sure which one but I think it was one around Themes), this error started popping up:

InvalidOperationException: No service for type 'OrchardCore.Themes.Services.DarkModeService' has been registered.

Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider, Type serviceType)
Microsoft.AspNetCore.Mvc.Razor.RazorPagePropertyActivator+<>c__DisplayClass8_0.<CreateActivateInfo>b__1(ViewContext context)
Microsoft.Extensions.Internal.PropertyActivator<TContext>.Activate(object instance, TContext context)
Microsoft.AspNetCore.Mvc.Razor.RazorPagePropertyActivator.Activate(object page, ViewContext context)
Microsoft.AspNetCore.Mvc.Razor.RazorPageActivator.Activate(IRazorPage page, ViewContext context)
Microsoft.AspNetCore.Mvc.Razor.RazorView.RenderPageCoreAsync(IRazorPage page, ViewContext context)

I can't access any of the pages or login/logoff, so not entirely sure how to fix this. After searching for a resolution for some time, haven't been able to find this mentioned anywhere.

hishamco:
Fixes #12707

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Themes/Manifest.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Themes/Manifest.cs
@@ -7,5 +7,6 @@ using OrchardCore.Modules.Manifest;
     Version = ManifestConstants.OrchardCoreVersion,
     Description = "The Themes modules allows you to specify the Front and the Admin theme.",
     Dependencies = new[] { "OrchardCore.Admin" },
-    Category = "Theming"
+    Category = "Theming",
+    IsAlwaysEnabled = true
 )]


hishamco:
The admin theme required `DarkModeService`, so if we disable `OC.Themes` module it will break everything

jtkech:
Maybe just make `TheAdmin` dependent on `OC.Themes` in its manifest.

ns8482e:
May be `DarkModeService` should be moved out of theme and defined in `Admin` module then

jtkech:
And make `TheAdmin` dependent on `Admin`, yes looks better.

Skrypt:
I don't agree 100%. The reason why the dark mode is tied to the frontend views is because of the Login views that can be overridden optionally in a theme. The Login Layout needs to register the dark mode service to take it into consideration. Also, the login is the admin login and the frontend login. So, we would need to make a distinction between both then.

jtkech:
Okay didn't remember this ;)

hishamco:
Agree with @Skrypt, dark mode is a theme specific

hishamco:
At least we need a proper way to fix the related issue

PR #12712, opened 2022-10-24
Add Content Field handlers, and update Api Contents to return group keys

### LINKED ISSUE #12696: Content parts/fields are not being validated  against the settings when creating content from the Content API
When creating contents using the Contents API, the content is stored as the user sends it without any kind of validation. For example, if you create a type with a required `TitlePart`, and post a new item with no title, the content item gets created with no title.

I think we need to implement the `ValidatingAsync` method in every part/field handler so that way each part/field is validated against it's own settings. Also, add field handlers to be called so we can validate fields.

MikeAlhayek:
Fix #12696 

- Add a way to attach handlers for fields. For example, `services.AddContentField<TextField>().UseDisplayDriver<TextFieldDisplayDriver>().AddHandler<TextFieldHandler>();`
- *(Breaking Change)* In the interface `IContentPartHandler` the signature `Task ValidatingAsync(ValidateContentContext context, ContentPart part);` was changed to `Task ValidatingAsync(ValidatePartContentContext context, ContentPart part);`. Also, the signature `Task ValidatedAsync(ValidateContentContext context, ContentPart part);` was changed to `Task ValidatedAsync(ValidatePartContentContext context, ContentPart part);`
- `TitlePartHandler` was modified to validate the part against the `TitlePartSettings`
- Now that validation takes place, the post request to `/api/contents` return validation error with key.
- New Handlers were added
  - `TextFieldHandler` to validate a text field
  - One handler for each part/field will be added "if it does not exists" that would implement that `ValidatingAsync(...)` event. This would be part of a follow up PR ensure team is on agreement with the approach and to reduce the code change.

### DIFF
--- a/src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Handlers/ContentPartHandler.cs
+++ b/src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Handlers/ContentPartHandler.cs
@@ -87,14 +87,14 @@ namespace OrchardCore.ContentManagement.Handlers
                 : Task.CompletedTask;
         }
 
-        Task IContentPartHandler.ValidatingAsync(ValidateContentContext context, ContentPart part)
+        Task IContentPartHandler.ValidatingAsync(ValidatePartContentContext context, ContentPart part)
         {
             return part is TPart tpart
                 ? ValidatingAsync(context, tpart)
                 : Task.CompletedTask;
         }
 
-        Task IContentPartHandler.ValidatedAsync(ValidateContentContext context, ContentPart part)
+        Task IContentPartHandler.ValidatedAsync(ValidatePartContentContext context, ContentPart part)
         {
             return part is TPart tpart
                 ? ValidatedAsync(context, tpart)
@@ -205,8 +205,8 @@ namespace OrchardCore.ContentManagement.Handlers
         public virtual Task ImportedAsync(ImportContentContext context, TPart instance) => Task.CompletedTask;
         public virtual Task UpdatingAsync(UpdateContentContext context, TPart instance) => Task.CompletedTask;
         public virtual Task UpdatedAsync(UpdateContentContext context, TPart instance) => Task.CompletedTask;
-        public virtual Task ValidatingAsync(ValidateContentContext context, TPart instance) => Task.CompletedTask;
-        public virtual Task ValidatedAsync(ValidateContentContext context, TPart instance) => Task.CompletedTask;
+        public virtual Task ValidatingAsync(ValidatePartContentContext context, TPart instance) => Task.CompletedTask;
+        public virtual Task ValidatedAsync(ValidatePartContentContext context, TPart instance) => Task.CompletedTask;
         public virtual Task VersioningAsync(VersionContentContext context, TPart existing, TPart building) => Task.CompletedTask;
         public virtual Task VersionedAsync(VersionContentContext context, TPart existing, TPart building) => Task.CompletedTask;
         public virtual Task DraftSavingAsync(SaveDraftContentContext context, TPart instance) => Task.CompletedTask;
--- a/src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Handlers/IContentPartHandler.cs
+++ b/src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Handlers/IContentPartHandler.cs
@@ -19,8 +19,8 @@ namespace OrchardCore.ContentManagement.Handlers
         Task ImportedAsync(ImportContentContext context, ContentPart part);
         Task UpdatingAsync(UpdateContentContext context, ContentPart part);
         Task UpdatedAsync(UpdateContentContext context, ContentPart part);
-        Task ValidatingAsync(ValidateContentContext context, ContentPart part);
-        Task ValidatedAsync(ValidateContentContext context, ContentPart part);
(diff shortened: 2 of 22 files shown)


hishamco:
Thanks for contribution, some of the newly added suggestion need to be discussed before you create a PR, so it would be nice if you file an issue proposing what you want do

hishamco:
> Feel free to discuss anything you want there

What I mean is instead of spend time on a PR that may be accepted or rejected, discussion is much better, so the team will decide how a particular feature should be implemented. With that let us discuss the kind of validation we need, where and how it should go

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Handlers/ContentPartHandler.cs
```diff
@@ -205,8 +220,13 @@ async Task IContentPartHandler.ClonedAsync(CloneContentContext context, ContentP
         public virtual Task ImportedAsync(ImportContentContext context, TPart instance) => Task.CompletedTask;
         public virtual Task UpdatingAsync(UpdateContentContext context, TPart instance) => Task.CompletedTask;
         public virtual Task UpdatedAsync(UpdateContentContext context, TPart instance) => Task.CompletedTask;
+        public virtual Task ValidatingAsync(ValidatePartContentContext context, TPart instance) => Task.CompletedTask;
+        public virtual Task ValidatedAsync(ValidatePartContentContext context, TPart instance) => Task.CompletedTask;
```
ns8482e, line 224: Why adding another overload for `ValidatingAsync` and `ValidatedAsync` we already have these methods below

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Handlers/IContentPartHandler.cs
```diff
@@ -19,8 +20,15 @@ public interface IContentPartHandler
         Task ImportedAsync(ImportContentContext context, ContentPart part);
         Task UpdatingAsync(UpdateContentContext context, ContentPart part);
         Task UpdatedAsync(UpdateContentContext context, ContentPart part);
+        Task ValidatingAsync(ValidatePartContentContext context, ContentPart part);
+        Task ValidatedAsync(ValidatePartContentContext context, ContentPart part);
+
+        [Obsolete("This method will be removed in future releases. Instead, use ValidatingAsync(ValidatePartContentContext context, ContentPart part)")]
```
ns8482e, line 26: Doesn't like an idea to add `Obsolete` for no reason

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Handlers/IContentPartHandler.cs
```diff
@@ -19,8 +20,15 @@ public interface IContentPartHandler
...
         Task ValidatingAsync(ValidateContentContext context, ContentPart part);
+
+        [Obsolete("This method will be removed in future releases. Instead, use ValidatedAsync(ValidatePartContentContext context, ContentPart part)")]
```
ns8482e, line 29: Doesn't like an idea to add Obsolete for no reason

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Handlers/ContentPartHandler.cs
MikeAlhayek: The argument changes. if I replace the parameter, that would be a breaking change and I want to avoid that. so I thought to make the old stuff obsolete and we'll remove them in future releases

ns8482e:
@MikeAlhayek I agree with @hishamco -  any time you feel need to make changes to interface and abstraction, chances are you are introducing breaking changes and/or adding instability to a product so - its better to discuss the design instead of putting efforts on PR.

src/OrchardCore/OrchardCore.ContentManagement.Abstractions/Handlers/IContentPartHandler.cs
MikeAlhayek: @ns8482e 
These methods will add no value since they are replaced by another method. I kept them to not introduce a breaking change as you saw happened in IDisplayManager

ns8482e: It's  better you find a way to use same method, than introducing new one

MikeAlhayek: @ns8482e I made the requested change. Let me know if you see something else that needs updating.

sebastienros:
What would happen if we introduce `ValidateContentPartContext` (same as for fields), and all existing methods and interfaces are not changed, but when the actual validation occurs (example in TextFieldHandler) then we cast to it to get the TypePartDefinition.

sebastienros:
We could then even add handlers with the specialized type for people to overload to make its existence more obvious. The current one's default implementation would call the specialized one. By adding an overload to the concrete class

BaseFieldHandler -> override Validating(ContentContext) -> override Validating(ContentFieldContext)

sebastienros:
@deanmarcussen FYI

deanmarcussen:
> @deanmarcussen FYI

Yeah the idea looks like a good addition

MikeAlhayek:
@deanmarcussen if you have time, can you please review the code?

sebastienros (approved):

PR #12752, opened 2022-11-03
Changing ViewContent to ViewOwnContent

### LINKED ISSUE #12751: GraphQL is not honoring ViewOwnContent for a given content type
### Describe the bug
Let say we have a content type called "Article" and the user have access to view own articles only. The user gets unauthorized error when trying to view the content using GraphQL.

### To Reproduce
1. Create content type called Article
2. Remove ViewContents permissions from Anonymous role
3. give the user permission to view own Articles + ExecuteGraphQL.
4. Query Articles using GraphiQL.

### Expected behavior
I expect GraphiQL to honor the permissions

MikeAlhayek:
Fix #12751

@jtkech we worked on the previous PR for GraphQL together. The permission change in this PR is correct since this will check a lower level permission that is implied by ViewContents.  But the `ShouldReturnBlogsWithViewBlogContentPermission` is failing. For some reason the authenticated user "in the test" has no claims which explain why the test will fail. Any idea on the fix? The change in this PR fixes the issue on hand but breaks the test.

### DIFF
--- a/src/OrchardCore/OrchardCore.ContentManagement.GraphQL/Queries/ContentTypeQuery.cs
+++ b/src/OrchardCore/OrchardCore.ContentManagement.GraphQL/Queries/ContentTypeQuery.cs
@@ -68,7 +68,7 @@ namespace OrchardCore.ContentManagement.GraphQL.Queries
                 };
 
                 query.RequirePermission(CommonPermissions.ExecuteGraphQL);
-                query.RequirePermission(Contents.CommonPermissions.ViewContent, typeDefinition.Name);
+                query.RequirePermission(Contents.CommonPermissions.ViewOwnContent, typeDefinition.Name);
 
                 foreach (var builder in contentTypeBuilders)
                 {
--- a/test/OrchardCore.Tests/Apis/Context/AuthenticationContext.cs
+++ b/test/OrchardCore.Tests/Apis/Context/AuthenticationContext.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Security.Principal;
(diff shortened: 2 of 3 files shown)


jtkech:
Yes about the user claims, maybe that's why we are using fot tests a `PermissionsContext` and a custom  `PermissionContextAuthorizationHandler` and its `HandleRequirementAsync()`. But then yes implied permissions will not be taken into account.

So to make the failing test working I needed to add to the `PermissionsContext` the `ViewOwnContent` which normally is implied by `ViewContent` (see below).

I think that's okay if you tried it and if it works as expected in a real context.

                .WithPermissionsContext(new PermissionsContext
                {
                    UsePermissionsContext = true,
                    AuthorizedPermissions = new[]
                    {
                        GraphQLApi.Permissions.ExecuteGraphQL,
                        Contents.CommonPermissions.ViewContent,
                        Contents.CommonPermissions.ViewOwnContent,
                    }
                });

jtkech:
The other way is to complete the testing authorization handler, I copy pasted the private method `GetGrantingNamesInternal()` from the `DefaultPermissionGrantingService` and used it in the testing handler, and it was working without having to add the implied permissions in the PermissionsContext.

        protected override Task HandleRequirementAsync(AuthorizationHandlerContext context, PermissionRequirement requirement)
        {
            var permissions = (_permissionsContext.AuthorizedPermissions ?? Enumerable.Empty<Permission>()).ToList();

            if (!_permissionsContext.UsePermissionsContext)
            {
                context.Succeed(requirement);
            }
            else if (permissions.Contains(requirement.Permission))
            {
                context.Succeed(requirement);
            }
            else
            {
                // Used here to retrieve ImpliedBy permissions names.
                var grantingNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
                GetGrantingNamesInternal(requirement.Permission, grantingNames);
                if (permissions.Any(p => grantingNames.Contains(p.Name)))
                {
                    context.Succeed(requirement);
                }
                else
                {
                    context.Fail();
                }
            }

            return Task.CompletedTask;
        }

MikeAlhayek:
@jtkech Thank you for that info. I fixed the `HandleRequirementAsync` and the tests are now green.

I also added another test for `ViewOwnContent` so both permission

PR #12757, opened 2022-11-03
Fix OC version used in templates

sebastienros:
The template default settings are generated on build. These templates have an `OrchardVersion` argument that has a default value defined in an msbuild script to be replaced with the `Version` property. In the release pipeline this value will have the `VersionSuffix` applied to it, like `preview`. 

This PR forces the value of the version property to be the one in the tag which will prevent `preview` from being added. This will also fix the templates' default value for `OrchardVersion`.

The `pack` command uses the `Version` property only for the packages, and at that point the templates json files are already built with the wrong version.

### DIFF
--- a/.github/workflows/release_ci.yml
+++ b/.github/workflows/release_ci.yml
@@ -43,7 +43,7 @@ jobs:
       run: echo "BuildNumber=$(( $GITHUB_RUN_NUMBER + 15471 ))" >> $GITHUB_ENV
     - name: Build
       run: |
-        dotnet build --configuration Release --framework net6.0
+        dotnet build --configuration Release --framework net6.0 -p:Version=${{ steps.get_version.outputs.VERSION }}
     - name: Unit Tests
       run: |
         dotnet test -c Release --no-restore --no-build ./test/OrchardCore.Tests/OrchardCore.Tests.csproj 


sebastienros:
@ns8482e this is the issue you mentioned this morning and that I was able to reproduce.
Another option would be to remove the `<VersionSuffix>` value when we ship and put it back after. This is probably something we have forgotten for some releases. With this solution we don't have to change it. Like we don't have to change the `<VersionPrefix>` either.

sebastienros:
https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore.Build/OrchardCore.Commons.props#L7-L9

ns8482e:
@sebastienros this how I have in my modules
```xml
<VersionPrefix>1.5</VersionPrefix>
 <VersionSuffix Condition="'$(GraduatePackage)'!='true'">preview</VersionSuffix>
 <VersionSuffix Condition="'$(VersionSuffix)'!='' AND '$(BuildNumber)' != ''">$(VersionSuffix)-$(BuildNumber)</VersionSuffix>
```

and in ADO

```yml
 - task: DotNetCoreCLI@2
        displayName: "Build OCCMS .net 6.0" 
        inputs:
          command: 'build'
          projects: '$(Agent.BuildDirectory)/code/occms/occms.sln'
          arguments: '--configuration $(buildConfiguration) --no-restore --framework net6.0'
          buildProperties: 'BuildNumber="$(buildNumber)";GraduatePackage="$(GraduatePackage)"'
```

ns8482e (approved):

jtkech:
Too late but LGTM ;)

PR #12760, opened 2022-11-04
Removed popperjs because it was removed from package.json

netwavebe:
popperjs was removed from package.json but it was still used in Assets.json. This throws an exception when building assets on a clean install.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Setup/Assets.json
+++ b/src/OrchardCore.Modules/OrchardCore.Setup/Assets.json
@@ -10,7 +10,6 @@
   {
     "inputs": [
       "node_modules/jquery/dist/jquery.js",
-      "node_modules/@popperjs/core/dist/umd/popper.js",
       "node_modules/bootstrap/dist/js/bootstrap.js",
       "Assets/js/setup.js",
       "Assets/js/strength.js"


Skrypt:
I did not encounter this exception yet while doing `gulp build` on the latest main branch. Could you share a screenshot of the exception?

netwavebe:
Sure, Jasmin!

image

Skrypt:
From what I understand, Popper.js is a dependency of Bootstrap 5. To reproduce the issue I removed the package.lock.json file standing in the Setup module and did a `git clean -xdf` on the entire OC folder. Even when I'm removing the package.lock.json file the popper dependency gets added back.

image

So, this is why I can't repro. Maybe this is an issue related to using an older version of NPM?

netwavebe:
You have popperjs in node_modules because it used to exist in an older version of package.json. To simulate the problem, delete node_modules and reinstall the required packages by running `npm install`. Popperjs won't reappear and building assets will fail.

The package was removed by @agriffard here: https://github.com/OrchardCMS/OrchardCore/commit/0bf04cca19b146ace33d67a3cf3beab8d237f474#diff-4355bceeb8a7bf8d18490bef8420ac02491705d8b7dacc2fd996a422edc3b42e.

Skrypt:
This is what I did by doing `git clean -xdf`; it removes every node_modules folder. So, on my side the `bootstrap` dependency resolves automatically popper as a dependency from a `npm install`.

Skrypt:
I think we should keep popper in there, maybe we should do the opposite and add it explicitly in the package.json file instead.

agriffard:
`bootstrap.bundle.js` contains `popperjs`.

Skrypt (approved):

sebastienros:
Based on the last 3 comments it doesn't look like we should merge this PR.

sebastienros:
Do we agree to close the PR?

PR #12802, opened 2022-11-12
Reject grant_type=client_credentials token requests containing scope=offline_access

kevinchalet:
We recently added the ability to automatically grant the requested scopes for `grant_type=client_credentials` (which requires that the client be granted the corresponding scope permissions, of course) but we forgot to add a check to prevent a refresh token from being requested via `scope=offline_access`, which is a scenario we deliberately don't allow. We have a check preventing blocking `grant_type=refresh_token` requests in this case:

https://github.com/OrchardCMS/OrchardCore/blob/18938aa517fa151f5ac582d03cb80c93ca2f7b27/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/AccessController.cs#L588-L600

Since there's no point returning a refresh token that won't be usable, we should add a check to prevent the `offline_access` scope from being requested in the first place. That's exactly what this PR does 😄

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/AccessController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/AccessController.cs
@@ -442,6 +442,16 @@ namespace OrchardCore.OpenId.Controllers
 
         private async Task<IActionResult> ExchangeClientCredentialsGrantType(OpenIddictRequest request)
         {
+            if (request.HasScope(Scopes.OfflineAccess))
+            {
+                return Forbid(new AuthenticationProperties(new Dictionary<string, string>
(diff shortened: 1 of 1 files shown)


HengzheLi:
So, is that means after this change, the client_credentials granted client have to re-apply a new access token instead of refreshing the access token using refresh token?

kevinchalet:
To get a new access token with `client_credentials`, simply send a new `grant_type=client_credentials` instead of using refresh tokens.

`grant_type=refresh_token` with a refresh token issued after a first `grant_type=client_credentials` has never been supported, this PR just fixes a regression introduced in the last OC patch that allowed a refresh token to be returned (but you can't use it so there's really no point returning it in the first place).

PR #12803, opened 2022-11-12
Don't localize custom errors returned by the userinfo endpoint

kevinchalet:
The OAuth 2.0 specification explicitly requires that errors be composed exclusively of certain USCII characters, which basically prevents localizing them:

image

image

It's an issue we identified and fixed some time ago for the errors returned by `AccessController`, but not by `UserInfoController`, that still returns a localized error. E.g in French:

image

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/UserInfoController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/UserInfoController.cs
@@ -6,7 +6,6 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore;
 using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Mvc;
-using Microsoft.Extensions.Localization;
 using OpenIddict.Abstractions;
 using OpenIddict.Server.AspNetCore;
 using OrchardCore.Modules;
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/UserInfoController.cs
```diff
@@ -6,22 +6,18 @@
...
 using Microsoft.AspNetCore.Mvc;
-using Microsoft.Extensions.Localization;
 using OpenIddict.Abstractions;
 using OpenIddict.Server.AspNetCore;
 using OrchardCore.Modules;
 using static OpenIddict.Abstractions.OpenIddictConstants;
 
 namespace OrchardCore.OpenId.Controllers
 {
+    // Note: the error descriptions used in this controller are deliberately not localized as
```
hishamco, line 16: I think we should move this note before setting the description

kevinchalet: We can certainly do that. It's worth noting that `AccessController` already has the same message at the same place:

https://github.com/OrchardCMS/OrchardCore/blob/18938aa517fa151f5ac582d03cb80c93ca2f7b27/src/OrchardCore.Modules/OrchardCore.OpenId/Controllers/AccessController.cs#L26-L29

If we decide to move it before every returned error, this will result in many copies. Is it fine for you?

PR #12826, opened 2022-11-16
Remove double database connection validation in setup

### LINKED ISSUE #12825: Double database connexion validation in SetupService
It seems to me that there is a double database connexion validation in SetupService.

There is the `IDbConnectionValidator.ValidateAsync` method
https://github.com/OrchardCMS/OrchardCore/blob/2b1771f467fa1fff124acd8f8bc95392b16583f5/src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs#L170-L184

And also this piece of code :
https://github.com/OrchardCMS/OrchardCore/blob/2b1771f467fa1fff124acd8f8bc95392b16583f5/src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs#L201-L223

Isn't this the same verification ?

The idea behind is to remove the dependency over YesSql.

TFleury:
Removes double database connexion validation in SetupService as it's already done by `IDbConnectionValidator.ValidateAsync`.
Removes OrchardCore.Setup.Core dependency on YesSql.Abstractions

Fixes #12825


Skrypt:
If you look at the comments it says that it could throw an exception if someone tries to use the same database with the same prefix. This is not something that the DB connection validator will test unless I'm wrong. Maybe it should then.

TFleury:
It's tested by `DbConnectionValidator` and there is a dedicated validator result for this case : `DocumentTableFound`
I tested the PR on my local env, and already used table prefix is detected.

Skrypt:
I'm ok with this as long as we are sure that it is now impossible for the `IStore` to be null.

TFleury:
I wrapped shell descriptor manager initialization into a try/catch so it will always fail gracefully with error feedback in the UI.

src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs
```diff
@@ -202,32 +199,21 @@ private async Task<string> SetupInternalAsync(SetupContext context)
...
-
-                        // The issue is that the user creation needs the tables to be present,
-                        // if the user information is not valid, the next POST will try to recreate the
-                        // tables. The tables should be rolled back if one of the steps is invalid,
-                        // unless the recipe is executing?
-
-                        _logger.LogError(e, "An error occurred while initializing the datastore.");
-                        context.Errors.Add(String.Empty, S["An error occurred while initializing the datastore: {0}", e.Message]);
+                        _logger.LogError(e, "An error occurred while initializing the shell descriptor manager.");
+                        context.Errors.Add(String.Empty, S["An error occurred while initializing the shell descriptor manager: {0}", e.Message]);
```
jtkech, line 214: Maybe here the error message is too technical.

I suggest to keep the previous message (already translated ;)) because we still do a first data store initialization before a first update of the shell descriptor in this data store that checks if such a table (with a given serialNumber) already exists, and anyway we pass `e.Message`, but as you want.

Otherwise, maybe something like `Failed to update the tenant descriptor to the data store.`

jtkech (approved):
just one suggestion but in the meantime I still approved.

PR #12829, opened 2022-11-16
Remove duplicate ResourceCdnHelperExtensions

TFleury:
Was moved to OrchardCore.ResourceManagement.Core but still present in OrchardCore.ResourceManagement.
OrchardCore.ResourceManagement depends on OrchardCore.ResourceManagement.Core


Skrypt:
Then it needs to be set as [obsolete] with a message describing what should be done.

TFleury:
I don't think it needs to be marked as obsolete because the 2 versions were in conflict.
If you add reference to both OrchardCore.ResourceManagement and OrchardCore.ResourceManagement.Core to your project, you can't use ResourceUrl helper because the call is ambiguous. That's how I found the issue.

Skrypt:
That's what we generally do all the time. Mark the old class or method with an [obsolete] attribute so that when you try to compile it will either fail or warn you.

Skrypt:
It has been moved by @MikeAlhayek 3 months ago so I think my suggestion is to add an [obsolete] attribute here.

TFleury:
As it is actually, the `ResourceUrl` extension method is unusable if you import OrchardCore.ResourceManagement in your project because there are 2 implementations. It raises a CS0121 ambiguous call.
So no code can rely on OrchardCore.ResourceManagement's implementation.

MikeAlhayek:
Oops! Yeah that should have been removed already. I think you should safly be able to remove it since it has the same namespace and also `OrchardCore.ResourceManagement` depends on `OrchardCore.ResourceManagement.Core`

Skrypt (approved):
I have not experienced the issue of the ambiguous call yet but I'm approving this one because of it.

MikeAlhayek (approved):

PR #12831, opened 2022-11-16
Fix NRE on migration from 1.4 -> 1.5 due to Lucene changes

PBMikeW:
Lucene Search Migration from 1.4 to 1.5 fails due to null reference error.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Search.Lucene/Migrations.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Search.Lucene/Migrations.cs
@@ -40,16 +40,16 @@ namespace OrchardCore.Search.Lucene
 
                             if (included != null)
                             {
-                                if (analyzed != null)
+                                if (analyzed == null)
                                 {
-                                    if ((bool)included)
+                                    if ((bool)included && !(bool)analyzed)
                                     {
                                         existingPartSettings["Keyword"] = true;
                                     }
                                 }
                                 else
                                 {
-                                    if ((bool)included && (analyzed == null || !(bool)analyzed))
+                                    if ((bool)included)
                                     {
                                         existingPartSettings["Keyword"] = true;
                                     }
@@ -83,16 +83,16 @@ namespace OrchardCore.Search.Lucene
 
                         if (included != null)
                         {
-                            if (analyzed != null)
+                            if (analyzed == null)
                             {
-                                if ((bool)included)
+                                if ((bool)included && !(bool)analyzed)
                                 {
                                     existingPartSettings["Keyword"] = true;
                                 }
                             }
                             else
                             {
-                                if ((bool)included && (analyzed == null || !(bool)analyzed))
+                                if ((bool)included)
                                 {
                                     existingPartSettings["Keyword"] = true;
                                 }
@@ -121,14 +121,14 @@ namespace OrchardCore.Search.Lucene
                             {
                                 if (analyzed == null)
                                 {
-                                    if ((bool)included)
+                                    if ((bool)included && !(bool)analyzed)
                                     {
                                         existingFieldSettings["Keyword"] = true;
                                     }
                                 }
                                 else
                                 {
-                                    if ((bool)included && (analyzed == null || !(bool)analyzed))
+                                    if ((bool)included)
                                     {
                                         existingFieldSettings["Keyword"] = true;
                                     }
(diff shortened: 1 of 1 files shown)


Skrypt:
Could you change this also in the 2 other places where I did the same.

```
if ((bool)included && !(bool)analyzed)
```

Skrypt (approved):

src/OrchardCore.Modules/OrchardCore.Search.Lucene/Migrations.cs
```diff
@@ -49,7 +49,7 @@ public int Create()
                                 }
                                 else
                                 {
-                                    if ((bool)included && !(bool)analyzed)
+                                    if ((bool)included && (analyzed == null || !(bool)analyzed))
```
jtkech, line 52: In the above line 43 we check `if (analyzed != null)` so here we know that `analyzed` is null and would not need to be checked at all.

@Skrypt maybe the source of the problem is that line 43 you wanted to check `if (analyzed == null)` instead, which is not the same in term of functionality. If so we would only need to update line 43.

src/OrchardCore.Modules/OrchardCore.Search.Lucene/Migrations.cs
```diff
@@ -92,7 +92,7 @@ public int Create()
                             }
                             else
                             {
-                                if ((bool)included && !(bool)analyzed)
+                                if ((bool)included && (analyzed == null || !(bool)analyzed))
```
jtkech, line 95: Idem

src/OrchardCore.Modules/OrchardCore.Search.Lucene/Migrations.cs
```diff
@@ -128,7 +128,7 @@ public int Create()
                                 }
                                 else
                                 {
-                                    if ((bool)included && !(bool)analyzed)
+                                    if ((bool)included && (analyzed == null || !(bool)analyzed))
```
jtkech, line 131: Here this is the opposite, in the above line 122 we check `if (analyzed == null)` so here we know that `analyzed` is not null and then the code could be kept as before without the null check.

src/OrchardCore.Modules/OrchardCore.Search.Lucene/Migrations.cs
Skrypt: I will take a look but that makes sense that I could have made a mistake with `if (analyzed != null)`

Skrypt (dismissed):

src/OrchardCore.Modules/OrchardCore.Search.Lucene/Migrations.cs
```diff
@@ -83,16 +83,16 @@ public int Create()
 
                         if (included != null)
                         {
-                            if (analyzed != null)
+                            if (analyzed == null)
                             {
-                                if ((bool)included)
+                                if ((bool)included && !(bool)analyzed)
```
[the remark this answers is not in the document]
Skrypt: lol, yes but there was a reason why I did it like it was. I need to think more. I'm doing this too quickly.

Skrypt:
Ok, I think I got this. Basically, a field cannot be analyzed and a keyword at the same time.
And previously the settings were:

Included = keyword (Set when the option "Include this element in the index" was checked)
Analyzed = tokenized (Set when the option "Analyzed" was checked)

Now, I believe that we had all the fields as "Keywords" by default which was the opposite of what ElasticSearch does. Also, it was not making any sense to have it set as a Keyword by default because most people will want these fields to be tokenized/analyzed by default because that's the goal of it ... You want to have only fewer specific fields to be set as Keyword for using them as technical values most of the time.

Based on the fact that "Analyzed" is now the option when you check "Include this element in the index".

| old value | new value | comment |
|----|---|----|
| !included + analyzed | included | This was impossible to happen but we never know |
| included + analyzed | included | Here we were specifying that we want the field to be analyzed |
| included + !analyzed | keyword | |
| included | keyword | same as previous but with a null Analyzed value |

So we need to take into consideration that Analyzed is now Included. We don't need to migrate the "Included" to a different value just set the new "Keyword" one.

src/OrchardCore.Modules/OrchardCore.Search.Lucene/Migrations.cs
```diff
@@ -85,14 +85,14 @@ public int Create()
                         {
                             if (analyzed != null)
```
Skrypt, line 86: This looks like a bad merge because here I had this correctly.

jtkech (approved):

Issue #12846, opened 2022-11-20
Make Some Controller Rewriteable


hyzx86:
Just like AccountController , we should abstract some methods to extract the interface to make it easier to replace
At present, if I want to modify some login codes, I have to load the entire UserModule into my project, which will cause trouble for subsequent code updates

sebastienros:
Have you checked if there isn't a way in aspnet to register different controllers dynamically? The current controllers are provided by Application Parts, Orchard has a custom implementation to load them from modules, so there might be a way already to replace existing ones.

hyzx86:
related https://github.com/OrchardCMS/OrchardCore/issues/11644#issuecomment-1399438340

hyzx86:
Sorry, I tried many ways but couldn't rewrite the controller. Maybe I'm not looking the right way,
Can you provide some examples?

hishamco:
@hyzx86 could you please clarify what are you trying to do in details to check if I can help

hyzx86:
First time , I tried to replace `Aaction` with this way
``` C# 
    public class ActionReplaceServiceProvider : IActionInvokerProvider
    {
        private readonly ActionReplaceOption _replaceOptions;

        public ActionReplaceServiceProvider(IOptions<ActionReplaceOption> replaceOptions)
        {
            _replaceOptions = replaceOptions.Value;
        }

        public int Order => 10;

        public void OnProvidersExecuted(ActionInvokerProviderContext context)
        {

        }

        public void OnProvidersExecuting(ActionInvokerProviderContext context)
        {
            var actionDescriptor = context.ActionContext.ActionDescriptor as ControllerActionDescriptor;
            var replaceConfg = _replaceOptions.FindOption(actionDescriptor.MethodInfo);
            if (replaceConfg != null)
            {
// change action methodinfo
                actionDescriptor.MethodInfo = replaceConfg.ActionMapping[actionDescriptor.MethodInfo.Name];
                actionDescriptor.ControllerTypeInfo = replaceConfg.NewController.GetTypeInfo();
            }
        }
    }
```

Then I considered using Filter as an alternative, but it didn't feel like it would work because it would require a lot of work to initialize the object and find the corresponding view

Finally, a possible approach is to use the UrlRewrite function to rewrite a known request path to another path



As for the suggestion that @sebastienros mentioned, I just had a quick look and I don't know where to start

[...]

PR #12855, opened 2022-11-21
Abstract AccountContorller

hyzx86:
Abstracts all actions in the `AccountController`
Enable logic code to be rewritten

link to https://github.com/OrchardCMS/OrchardCore/issues/12846

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Users/Abstractions/IAccountControllerAbstraction.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Abstractions/IAccountControllerAbstraction.cs
@@ -1,0 +1,25 @@
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using OrchardCore.Users.Controllers;
+using OrchardCore.Users.ViewModels;
+
+namespace OrchardCore.Users.Abstractions;
+public interface IAccountControllerAbstraction : IControllerAbstraction<AccountController>
+{
+    IActionResult ChangePassword(string returnUrl = null);
+    Task<IActionResult> ChangePassword(ChangePasswordViewModel model, string returnUrl = null);
+    IActionResult ChangePasswordConfirmation();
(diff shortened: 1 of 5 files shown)


hyzx86:
@sebastienros  @Skrypt  Is there any problem with this?  
At present, they can work, but code naming may not be so standardized
When SPA is upgraded in the future, it may need to be further abstracted to separate the logic code from MVC

hyzx86:
This change allows developers to freely specify views, and is not limited to a certain theme module

hishamco:
This is weird little bit, could you please let us know what functionality need to be customizable? We could use service instead

hyzx86:
> This is weird little bit, could you please let us know what functionality need to be customizable? We could use service instead

1.  I wan't to change login process ，Limit the authorization methods of openid client applications , some openid applications can only use external login (such as Github), and direct access to the login address is unrestricted
2.  Use script to  process  login event 

The login module is too deeply coupled with the MVC. 
Because the Controller cannot be rewritten, I have to load the OC UserModule into my project, if I want to modify these codes
but this will make it difficult to merge with the latest OC code

hishamco:
I think introduce a service for some account functionality will make it easier to extend and customize

hyzx86:
Yes, I admit that these codes are extreme. But it completely solves the problem of controller rewriting.
For example, if I want to debug a piece of code, I can rewrite some OC code to my project for debugging
Instead of copying the entire module to my solution

hyzx86:
> I think introduce a service for some account functionality will make it easier to extend and customize

If services are introduced, they should be decoupled from the controller
But if I want to render some other information on the view, I will not be able to use ViewData

Perhaps developers with such requirements should copy the UserModule into their own solutions separately..

hyzx86:
@hishamco I have entered a misunderstanding. 
Cms modules as a sample of  ocframework.
The functionality I need is no longer limited to this. 
So I should reconstruct some of the key modules.

PR #12927, opened 2022-12-09
Introduce TenantId

### LINKED ISSUE #12926: Redis Cache and default tenant setup
### Describe the bug

The Redis Cache feature will persist even if we reinstall the default tenant. The cache tag elements are prefixed with the tenant name and the default tenant always uses "default" as its tenant technical name. So, if someone reinstalls the default tenant it will happen that the previously cached items will still be persisted in that cache leading to issues.

I suggest that we introduce a TenantId in the tenant appsettings.json file so that we can prefix the cache tags with this UID to differentiate tenants.

Skrypt:
Fixes Redis Cache issue with tenants that are created with the same name over time.
Fixes missing separators on cache tags.

{InstancePrefix_TenantId_ShellName_Tag}

image

Here as you can see having a TenantId would make it possible to have multiple tenants named as Default without having any issues.

Fixes #12926


Skrypt:
I'm trying to find a place where I could add this TenantId from runtime if it is missing. I've tried in some places already but nothing good so far. Let me know if you have an idea.

deanmarcussen:
Ideally the cookie would also use this, preventing conflicts between cookies,
i.e. runing multiple orchard sites in (development) will also break the Default tenant cookie on each site.

This is the related issue https://github.com/OrchardCMS/OrchardCore/issues/1610,  so there are other places as well.

src/OrchardCore.Modules/OrchardCore.Redis/Options/RedisCacheOptionsSetup.cs
```diff
@@ -7,17 +7,18 @@ namespace OrchardCore.Redis.Options
...
         {
-            _tenant = shellSettings.Name;
+            _tenant = shellSettings.TenantId + Separator + shellSettings.Name;
             _redisOptions = redisOptions;
         }
 
         public void Configure(RedisCacheOptions options)
         {
-            options.InstanceName = _redisOptions.Value.InstancePrefix + _tenant;
+            options.InstanceName = _redisOptions.Value.InstancePrefix + Separator + _tenant + Separator;
```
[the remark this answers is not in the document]
Skrypt: This is what I thought after going to sleep last night. I surely missed updating it everywhere else.

src/OrchardCore/OrchardCore.Abstractions/Shell/ShellSettings.cs
```diff
@@ -44,6 +44,12 @@ public string VersionId
             set => _settings["VersionId"] = value;
         }
 
+        public string TenantId
+        {
+            get => _settings["TenantId"];
```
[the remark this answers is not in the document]
Skrypt: Ok for "Id" only. No issue there.

src/OrchardCore.Modules/OrchardCore.AutoSetup/AutoSetupMiddleware.cs
```diff
@@ -211,6 +211,7 @@ public async Task<ShellSettings> CreateTenantSettingsAsync(TenantSetupOptions se
             shellSettings.RequestUrlHost = setupOptions.RequestUrlHost;
             shellSettings.RequestUrlPrefix = setupOptions.RequestUrlPrefix;
             shellSettings.State = TenantState.Uninitialized;
+            shellSettings.TenantId = IdGenerator.GenerateId();
```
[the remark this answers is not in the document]
Skrypt: Does it call the SetupService here?

src/OrchardCore/OrchardCore.Setup.Core/SetupService.cs
```diff
@@ -267,6 +267,7 @@ void reportError(string key, string message)
 
             // Update the shell state
             shellSettings.State = TenantState.Running;
+            shellSettings.TenantId = IdGenerator.GenerateId();
```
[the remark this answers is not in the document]
Skrypt: Makes sense.

src/OrchardCore.Modules/OrchardCore.Redis/Options/RedisCacheOptionsSetup.cs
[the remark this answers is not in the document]
jtkech: Also, most of the time `InstancePrefix` is null, in that case better to not use the first separator.

jtkech:
@deanmarcussen 

> Ideally the cookie would also use this, preventing conflicts between cookies

Just saw that the antiforgery cookie already uses the `VersionId`, maybe there are other places to update.

src/OrchardCore/OrchardCore/Shell/ShellHost.cs
```diff
@@ -144,6 +144,7 @@ public async Task<ShellScope> GetScopeAsync(ShellSettings settings)
         public async Task UpdateShellSettingsAsync(ShellSettings settings)
         {
             settings.VersionId = IdGenerator.GenerateId();
+            settings.TenantId ??= IdGenerator.GenerateId();
```
Skrypt, line 147: This is the only part that I'm going to need some time to think about. I think we need this setting to be generated once and not be able to change it afterward. This means that if I'm passing an arbitrary ShellSettings object without a TenantId taken from the actual .json file it will try and generate a new TenantId. This is tricky. I would somehow prefer that it be set inside the ISetupService only once and that this method never adds any chance to morph it on runtime.

jtkech: When we 1st create the Default tenant (setup context) we call this method. When we create a tenant from the admin UI, API or WF we call this method. When we edit a tenant we load the current settings (including the TenantId) to be updated. When we setup a tenant we load the current settings (uninitialized state but already including the TenantId when it was created) from which we create an internal settings (in the initializing state) but still including the TenantId.

Setting the TenantId as soon as we create a tenant allows to add setup features (enabled in the setup context) that may need the TenantId (e.g. Redis cache).

> This means that if I'm passing an arbitrary ShellSettings object without a TenantId taken from the actual .json file it will try and generate a new TenantId.

If the related tenant is being created we set its TenantId and then first save its settings.

For updating you should 1st load the settings of the related tenant (including the already set TenantId) then update it before passing it here, otherwise it would be inconsistent also for other properties.

jtkech: Forgot to say that also on AutoSetup, we first create a given tenant and save its settings by calling `UpdateShellSettingsAsync()` which will set its `TenantId` (before setup it afterwards)

src/OrchardCore/OrchardCore.Abstractions/Shell/ShellSettings.cs
```diff
@@ -29,11 +29,15 @@ public ShellSettings(ShellConfiguration settings, ShellConfiguration configurati
...
+        /// <summary>
+        /// This construction should only be used for setup purposes as it will generate the immutable TenantId.
+        /// </summary>
+        /// <param name="settings"></param>
         public ShellSettings(ShellSettings settings)
         {
             _settings = new ShellConfiguration(settings._settings);
             _configuration = new ShellConfiguration(settings.Name, settings._configuration);
-            TenantId = settings.TenantId;
+            TenantId = IdGenerator.GenerateId();
```
jtkech, line 40: The passed settings already contains the `TenantId` set as soon as the tenant was created, and that can be preserved.

jtkech: As soon as the tenant was created by the UI, API or WF, also whe the Default tenant is created by code

src/OrchardCore/OrchardCore.Abstractions/Shell/ShellSettings.cs
```diff
@@ -48,7 +52,7 @@ public string VersionId
         public string TenantId
         {
             get => _settings["TenantId"];
-            set => _settings["TenantId"] = value;
+            init => _settings["TenantId"] = value;
```
jtkech, line 55: The init accessor is cool to make the prop read only but we may need to set it in the `UpdateShellSettingsAsync()` central place.

src/OrchardCore/OrchardCore/Shell/ShellHost.cs
```diff
@@ -144,7 +144,6 @@ public async Task<ShellScope> GetScopeAsync(ShellSettings settings)
         public async Task UpdateShellSettingsAsync(ShellSettings settings)
         {
             settings.VersionId = IdGenerator.GenerateId();
-            settings.TenantId ??= IdGenerator.GenerateId();
```
jtkech: Here the central place I was thinking about ;)

src/OrchardCore/OrchardCore.Abstractions/Shell/ShellSettings.cs
Skrypt: The idea is that I added an Init setter on the TenantId so it now needs to be set from a constructor. And it happens that this constructor is the one used for the setup only. The idea is that we should not be able to change the TenantId. Just use the one from the ShellSettings taken from the json file. Of course, anyone could change that TenantId from that json file afterward but that would be a mistake in that case. What I want is that this TenantId be set on setup and afterward nothing should be able to change it.

src/OrchardCore/OrchardCore/Shell/ShellHost.cs
Skrypt: This method is called UpdateShellSettingsAsync so its function is to update the ShellSettings and here it can mutate the TenantId if someone removes the TenantId from the tenant.json file. This is why I removed it from here. This method is surely called in a lot of custom modules to change the ShellSettings for their own need with their extensions. This means that if they don't have a TenantId it will append one to the tenant.json file but I would rather throw an exception if a TenantId is missing instead of having a method that could potentially morph the TenantId.

src/OrchardCore/OrchardCore.Abstractions/Shell/ShellSettings.cs
Skrypt: It makes it read-only and also means that this value can only be set inside one of this class constructors.

src/OrchardCore/OrchardCore.Abstractions/Shell/ShellSettings.cs
jtkech: We may need the TenantId also for an Uninitilized tenant, before the tenant setup is done, e.g. if we add the redis data protection to the Setup Features, or any other feature that would use this TenantId.

src/OrchardCore/OrchardCore/Shell/ShellHost.cs
jtkech: We should update the existing loaded (or 1st created) settings otherwise it is not consistent for any other property. As we do for content items, document manager, the pattern is to cache immutable instance and load mutable instances for updating.

jtkech:
@Skrypt

So as it is done currently by setting the TenantId in the `ShellSettings(ShellSettings settings)` ctor, in all places we create a tenant by using `CreateDefaultSettings()` and then `UpdateShellSettingsAsync()`, the `TenantId` will not be set because `CreateDefaultSettings()` doesn't use the same `ShellSettings` ctor.

So the `TenantId` would not be set as soon as we create an `Uninitialized` tenant, for example we may add `Setup Features` that may need this `TenantId` that would need to be set before a tenant setup.

Then, before a setup we use this `ShellSettings(ShellSettings settings)` ctor to build an internal shell settings, but here we should not generate a new `TenantId` but use the one of the shell settings passed as a parameter, which is related to the previously created `Uninitialized` tenant we want to setup.

So for now the only place I still see to generate the `TenantId` (if not already set) is the central place we use to really persit the `ShellSettings` (a new one or not), which is in `UpdateShellSettingsAsync()`.

So if you agree we would just need to revert back to my last commit.

Skrypt:
Ok, let me take a second look before.

sebastienros:
Ok with the changes that only introduce the tenant id.

jtkech:
> Ok with the changes that only introduce the tenant id.

Yes regarding the last comments in #12946, easy to do with a few lines of code.

PR #12930, opened 2022-12-09
Remove YesSql.Abstractions reference from OC.Users.Core

hishamco:


hishamco:
Nothing much here except an extra project reference

Skrypt (approved):

PR #12931, opened 2022-12-09
Refactor OrchardCore.Roles module

hishamco:
- Move `RoleService` & `RoleStore` to `OC.Roles.Core` similar to `OC.Users`
- Introduce `AddRoles()` extension method


ns8482e:
Same comment as Users, what use of Extension method? Are you adding ability adding roles services by extortion to allow adding such services by modules other than role module?

sebastienros:
Can we do it after the other PR that changes these files too? there would be a merge conflict and it's easier to change this one.

hishamco:
No conflict is there

MikeAlhayek:
@hishamco this PR will break a more important PR #12420 which fixes a real problem. Maybe we should wait here until the other one is looked at and/or merged. I am going to add a do not mere here just until the other PR is merged. I hope you don't mind

hishamco:
@MikeAlhayek your PR is not updated since three months ago, also CI failed, moreover I'm not sure if we going to merge this with proposed changes soon

The changes here is not much complicated, then you can react to it. I never mind at all if we agreed to merge your PR first. Again no need to push your PR to merge, some of PRs might takes months or years ;)

hishamco:
@MikeAlhayek I'm not sure when your PR will get merge, let us merge this, then you can react to the changes

jtkech:
Just for info, moving the `RolesDocument.cs` file is a breaking change, existing tenants would not be able to read their existing `RolesDocument` from the database which is in the `Document` table.

This because in the `Type` column the value include the assembly name.

    OrchardCore.Roles.Models.RolesDocument, OrchardCore.Roles

hishamco:
Good point @jtkech so, can we create migration for the old tenants for backward compatibility or do you suggest something else?

Piedone (changes requested):
I'll be a buzzkill again, and I'm sorry about that, but I'll say the same as under https://github.com/OrchardCMS/OrchardCore/pull/11157#pullrequestreview-1934813242. This is a breaking change for subjective improvements that don't have a clear and widely applicable benefit.

PR #12932, opened 2022-12-09
Introduce AddUsers() extension method

hishamco:


ns8482e:
What's benefit you get with this extension method? Ability to register user services by other modules and just user module? Don't see any reason why some one do that?

hishamco:
This is handy if we use `OC.Users` outside OrchardCore, something similar to `AddPortableObjectLocalization` which registers the required services

sebastienros (approved):

sebastienros:
Better convention and not breaking.

PR #12953, opened 2022-12-15
Add permission check for content type filter 

### LINKED ISSUE #12735: Content Type Filter should only show accessible content types
Content Type Filter should only show accessible content types

The user should only see content-types that he/she can view. currently all of the content types are listed even if you don't have permissions to see it.

image

MikeAlhayek:
Fix #12735

In the process I added `IContentItemFactory` that would allow us to reuse dummy content items that are only used for permission check. This should help reducing the cost of events every time we create content item for just to perform a permission check.


src/OrchardCore/OrchardCore.ContentManagement/ContentItemFactory.cs
```diff
@@ -0,0 +1,46 @@
...
+public class ContentItemFactory : IContentItemFactory
+{
+    private readonly IContentManager _contentManager;
+    private readonly IContentItemIdGenerator _contentItemIdGenerator;
+    private readonly ContentItemFactorySession _contentItemFactorySession;
+
+    public ContentItemFactory(
+        IContentManager contentManager,
+        IContentItemIdGenerator contentItemIdGenerator,
+        ContentItemFactorySession contentItemFactorySession)
```
sebastienros, line 15: Use IMemoryCache

MikeAlhayek: Done!

jtkech:
@MikeAlhayek 

Looks like `IContentItemFactory` is only there to create dummy content items for authorisation checking, if so I think this is not a good name, in some contexts people may use it wrongly. Hmm, would be good to only use extensions / helpers to not have to register this new factory, will see.

Also the usage of `MemoryCache` without invalidation path is not so good, we could use `IContentDefinitionStore` => `.GetContentDefinitionAsync()` returning `ContentDefinitionRecord` which is a `Document` having an `Identifier` that changes when the type definitions are mutated, so we could use this to invalidate the cache entries, we use this technic in `LayerFilter.cs`.

Hmm, I would need to suggest a new PR, will see if I have time.

MikeAlhayek:
@jtkech yes the idea is to prevent having to call all these handler when creating a content item for the sole purpose of doing a permission check. As you can see, we cache a single instance of each content type, then clone it on every request. I don't think we care to remove cache even if the content type definition is changed. 

Feel free to change the name in a separate PR. or let me know what you want changed and I'll adjust in a new PR

jtkech:
Okay, I will let you know.

> I don't think we care to remove cache even if the content type definition is changed.

Are we sure of this? Hmm, beyond the owner if only the content type Id is important, why not using a much more simpler component without calling any handlers to create a very dummy content item. Otherwise if the definition may be important, same invalidation concern.

Also maybe another name for the factory as it should be only used for authorisation checking, not in other regular contexts.

Will see and let you know.

MikeAlhayek:
yeah sure. we can simplify it by not using `IContentManager` at all. At that point, we don't need to cache anything instead we can create an extension method for `IAuthorizationService` that would look something like this instead

```
    public static Task<bool> AuthorizeContentTypeAsync(this IAuthorizationService service, ClaimsPrincipal user, Permission requiredPermission, string contentType, string owner = null)
    {
        if (String.IsNullOrEmpty(contentType))
        {
            throw new ArgumentException($"{nameof(contentType)} cannot be empty.");
        }

        var item = new ContentItem()
        {
            ContentItemId = IdGenerator.GenerateId(),
            ContentType = contentType,
            Owner = owner,
        };

        return service.AuthorizeAsync(user, requiredPermission, item);
    }
```

jtkech:
Cool I like it, just let me check it a little

jtkech:
Yes we have a static `IdGenerator` that uses `String.Create()`, I did it ;)

Okay I agree if the autorisation checking still works

MikeAlhayek:
I don't even think we need to even set an ID to make it even faster

jtkech:
Okay then, I let you try it and update it

PR #12954, opened 2022-12-15
Remove useless placeholder file

TFleury:


Skrypt:
Hmm good one. I think at some point it was necessary to keep it, can't remember why. @jtkech

TFleury:
I think it was necessary before having some code in the project. But since there is a Startup and controller in project, placeholder is useless.

jtkech:
As I remember it is a concern about creating the directory when publishing the `wwwroot` of the web application, but normally not a problemwith embedded files of a module / theme.

But at least should be tried ;)

MikeAlhayek:
@TFleury 

> As I remember it is a concern about creating the directory when publishing the wwwroot of the web application, but normally not a problem with the embedded files of a module / theme.

Did you try removing that file and then publishing to ensure the embedded files of modules and theme don't break?

TFleury:
I tried removing the file and the published directory structure is strictly the same.
I tested the output and didn't encountered any problem creating a new site.

Skrypt (approved):

PR #12980, opened 2022-12-23
Unique TenantId per tenant and per installation.

### LINKED ISSUE #1610: Consider introducing ShellSettings.Id

jtkech:
Fixes #1610, related to #12927 but here we only introduce `TenantId`.


Skrypt:
Merge it when you think it is ready.

sebastienros:
Should we return a default value (tenant name) for existing tenants, and start using TenantId in APIs where it makes sense?

src/OrchardCore/OrchardCore.Abstractions/Shell/ShellSettings.cs
```diff
@@ -44,6 +45,12 @@ public string VersionId
             set => _settings["VersionId"] = value;
         }
 
+        public string TenantId
+        {
+            get => _settings["TenantId"];
+            set => _settings["TenantId"] = value;
```
Skrypt, line 51: @sebastienros I would have preferred that this be an init set and used in the constructor that is used for the setup only. But that seems not possible to be done based on @jtkech comments on my initial PR.

Skrypt:
> Should we return a default value (tenant name) for existing tenants, and start using TenantId in APIs where it makes sense?

I would prefer simply generating a UID in the current appsettings.json file if possible from a migration. But, I can't find where to put it. I don't think we should mix TenantName and TenantId, the latter one should be harder to guess at all times.

jtkech:
@sebastienros @Skrypt Anyway I will think about it again, in the meantime.

> Should we return a default value (tenant name) for existing tenants,

Hmm but `TenantId` is to differentiate a given "installation". Hmm maybe step by step in separate PRs we could think about a good fallback per service that may need it, or expose options to let people using the fallback they want, as discussed about the Redis cache prefix (see below).

> and start using TenantId in APIs where it makes sense?

Okay no problem, need to think about which APIs may need it, maybe in separate PRs.

For info we already use `VersionId` for some cookies, we planned to use `TenantId` to be part of the Redis prefix but we said that people can easily configure the existing `RedisOptions.InstancePrefix`.

> I would have preferred that this be an init set and used in the constructor that is used for the setup only. But that seems not possible to be done based on @jtkech comments on my initial PR.

Yes, here for now I'm thinking about `AddSetupFeatures(...)`, if it adds a feature that needs the `TenantId` before the tenant is setup, while it is still uninitialized (setup context).

> I would prefer simply generating a UID in the current appsettings.json file if possible from a migration.

For now the `TenantId` is persisted in `tenants.json` (as `VersionId`). About using a kind of migration, hmm as we are loading a tenant we would need to update and persist its settings and then reload it, needs to think about it, e.g. in a distributed env, I may have some ideas, will see if we really need it.

src/OrchardCore/OrchardCore.Abstractions/Shell/ShellSettings.cs
```diff
@@ -33,6 +33,7 @@ public ShellSettings(ShellSettings settings)
         {
             _settings = new ShellConfiguration(settings._settings);
             _configuration = new ShellConfiguration(settings.Name, settings._configuration);
+            TenantId = settings.TenantId;
```
jtkech, line 36: Anyway we don't need this line as it comes from the `_settings` `ShellConfiguration` that we init just before by using the `ShellSettings settings` parameter from which we build a new `ShellSettings`.

jtkech:
@sebastienros @Skrypt 

Maybe an idea on the fly, if no `TenantId` we could fallback to the existing `VersionId`, knowing that this `VersionId` will still change on the next call of `UpdateShellSettings()`, but not `TenantId`.

Need to be tried to see if possible, will do this night.

jtkech:
@sebastienros @Skrypt 

Okay done

- For a new tenant, on the 1st save `VersionId` is generated as before and we auto assign `TenantId` with the same value. On a next save `TenantId` is kept as is, a new `VersionId` is generated as before.

- For an existing tenant, `TenantId` fallbacks to the existing `VersionId`, this until the next saving where this `TenantId` fallback is persisted while a new `VersionId` is generated as before.

PR #12984, opened 2022-12-24
Fix spacing around widgets in TheAdmin theme

### LINKED ISSUE #12983: Add margin bottom to widgets in AdminTheme
### Describe the bug
When multiple widgets are added. There is no space between them. I think the issue was caused by PR #12585

MikeAlhayek:
Fix #12983 

## Here is the UI after this PR

image

image

## Here is the UI before this PR.

image

image


hishamco:
From the screenshots above I see the spaces between the first two fields and the block quote are different

MikeAlhayek:
Yes the blockquote is different because that is a FlowPart which add a button between the widgets to insent another widget. That part did not change. The only thing that changed in this PR was around Widgets.

Skrypt (approved):

PR #12986, opened 2022-12-24
Change MappingsViewModel  class to public

hyzx86:
It may affect the view page rewriting


src/OrchardCore.Modules/OrchardCore.Search.Elasticsearch/ViewModels/MappingsViewModel.cs
```diff
@@ -1,6 +1,6 @@
 namespace OrchardCore.Search.Elasticsearch.ViewModels
 {
-    internal class MappingsViewModel
+    public class MappingsViewModel
```
hishamco, line 3: @Skrypt is there any reason that you made it `internal`?

Skrypt: No reason to be internal. Probably automatic code added and I did not double-check because I did add this one pretty quickly.

Skrypt (approved):

PR #12987, opened 2022-12-24
Remove extra semicolon

hyzx86:


src/OrchardCore.Modules/OrchardCore.Media/Fields/MediaFieldAttachedFileNameExtensions.cs
```diff
@@ -12,7 +12,7 @@ public static string[] GetAttachedFileNames(this MediaField mediaField)
         {
             var filenames = mediaField.Content["AttachedFileNames"] as JArray;
 
-            return filenames != null ? filenames.ToObject<string[]>() : Array.Empty<string>(); ;
+            return filenames != null ? filenames.ToObject<string[]>() : Array.Empty<string>();
```
hishamco, line 15:
```suggestion
            return filenames != null
                ? filenames.ToObject<string[]>()
                : Array.Empty<string>();
```

hishamco (changes requested):
Also we can form the condition as I suggested

hishamco (approved):

sebastienros:
And others ...
https://grep.app/search?q=%3B%20%3B&filter[repo][0]=OrchardCMS/OrchardCore

PR #12990, opened 2022-12-24
Remove unused testing data attributes

hishamco:


Skrypt:
Are they functional? I've not tested these environment variables but I believe they were introduced as part of the AutoSetup feature. Removing these tests may be wrong in that case.

hishamco:
> Are they functional?

There are not references at all, FYI they introduced when we have `appveyor.yml` now it's not part of the repo

Skrypt (approved):

PR #13004, opened 2022-12-27
Remove extra semicolon

hishamco:
Thanks for `grep.app` :)


Skrypt:
That would be worth trying to see if the code analysis/cleaner would do the work.

hishamco:
I try it before, I think it's a time to write code analysis for many things that we need. Hope to find time after I finish what I'm working on (Testing)

TFleury (approved):

Skrypt (approved):

PR #13010, opened 2022-12-29
adding returnurl to logoff

DavidStania:
Adding returnUrl to logoff.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Users/Controllers/AccountController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Controllers/AccountController.cs
@@ -243,12 +243,12 @@ namespace OrchardCore.Users.Controllers
 
         [HttpPost]
         [ValidateAntiForgeryToken]
-        public async Task<IActionResult> LogOff()
+        public async Task<IActionResult> LogOff(string returnUrl)
         {
             await _signInManager.SignOutAsync();
             _logger.LogInformation(4, "User logged out.");
 
-            return Redirect("~/");
+            return RedirectToLocal(returnUrl.ToUriComponents());
         }
 
         [HttpGet]
--- a/src/OrchardCore.Modules/OrchardCore.Users/Controllers/AccountController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Controllers/AccountController.cs
@@ -243,7 +243,7 @@ namespace OrchardCore.Users.Controllers
 
         [HttpPost]
         [ValidateAntiForgeryToken]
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.Users/Controllers/AccountController.cs
```diff
@@ -243,12 +243,12 @@ public async Task<IActionResult> Login(LoginViewModel model, string returnUrl =
 
         [HttpPost]
         [ValidateAntiForgeryToken]
-        public async Task<IActionResult> LogOff()
+        public async Task<IActionResult> LogOff(string returnUrl)
```
hishamco, line 246:
```suggestion
        public async Task<IActionResult> LogOff(string returnUrl = null)
```

src/OrchardCore.Modules/OrchardCore.Users/Controllers/AccountController.cs
```diff
@@ -243,12 +243,12 @@ public async Task<IActionResult> Login(LoginViewModel model, string returnUrl =
...
         {
             await _signInManager.SignOutAsync();
             _logger.LogInformation(4, "User logged out.");
 
-            return Redirect("~/");
+            return RedirectToLocal(returnUrl.ToUriComponents());
```
hishamco, line 251:
```suggestion
            return RedirectToLocal(returnUrl);
```

hishamco (changes requested):

sebastienros:
@hishamco did some changes

hishamco:
The build is failed, why there's two arguments instead of one?

jtkech:
@sebastienros added in the private `RedirectToLocal()` defined here our custom `Redirect` extension that uses `.ToUriComponents()` internally, but this extension is defined in `OC.Mvc.Core` that is not referenced by this project.

Hmm I would have used our `LocalRedirect()` extension instead because in `RedirectToLocal()` we test `IsLocalUrl()`  but for the same reason for now we can't.

So need to use one of the mvc methods by using explicitly `ToUriComponents()`.

    Redirect(url.ToUriComponents());

Or I think as we know it is a local url.

    LocalRedirect(localUrl.ToUriComponents());

jtkech:
Also, not important but I don't think we need to use `returnUrl = null` normally if no returnUrl is passed the mvc binder let it to null.

sebastienros:
Did the change in GH directly, please fix the build and merge.

jtkech:
Okay will do

DavidStania:
Nice! Thank you all.

PR #13021, opened 2023-01-01
Use ShellSettings.RequestUrlHosts that uses static separators

jtkech:


test/OrchardCore.Tests/Shell/ShellContainerFactoryTests.cs
```diff
@@ -131,7 +132,13 @@ public void WhenHostSingletonAndScoped_GetServices_Returns_CorrectImplementation
             Assert.IsType<HostScopedOfTheSameTypeAsSingleton>(services.ElementAt(1));
         }
 
-        private ShellBlueprint CreateBlueprint()
+        private static readonly ShellSettings _uninitializedDefaultShell = new()
```
hishamco, line 135: Can we move it to fields area at the top

src/OrchardCore.Modules/OrchardCore.Tenants/Services/TenantValidator.cs
```diff
@@ -151,7 +151,7 @@ private static IEnumerable<string> GetUrlHosts(string combinedUrlHosts)
                 return Enumerable.Empty<string>();
             }
 
-            return combinedUrlHosts.Split(HostsSeparator).Select(h => h.Trim());
+            return combinedUrlHosts.Split(_hostSeparators, StringSplitOptions.RemoveEmptyEntries);
```
hishamco, line 154: Why `Trim()` has been removed?!!

src/OrchardCore.Modules/OrchardCore.OpenId/Services/OpenIdServerService.cs
```diff
@@ -588,7 +588,8 @@ private static X509Certificate2 GenerateCertificate(X509KeyUsageFlags type, RSA
 
             X500DistinguishedName GetSubjectName()
             {
-                try { return new X500DistinguishedName("CN=" + (settings.RequestUrlHost ?? "localhost")); }
+                var host = settings.RequestUrlHosts.FirstOrDefault(host => host != "localhost");
+                try { return new X500DistinguishedName("CN=" + (host ?? "localhost")); }
```
hishamco, line 592: Can you format try .. cache block in your way

hishamco (changes requested):

test/OrchardCore.Tests/Shell/ShellContainerFactoryTests.cs
jtkech: Okay will do

src/OrchardCore.Modules/OrchardCore.Tenants/Services/TenantValidator.cs
jtkech: As we do in other places to split urlHosts, first we not only split on the comma separator but also the space, it allows to have `host1,host2` and for example `host1,   host2`.

So it also splits on space that may result in individual spaces (if any) in the returned array, then because we use the `RemoveEmptyEntries` all these spaces will be removed.

Notice that `StringSplitOptions` also has a `TrimEntries` option that we could have used if we were still splitting only on commas, this in place of adding a Linq select to do it by re-evaluating the array.

Anyway I will tweak it again in #12683 as here we are splitting the model.RequestUrlHost for each tenant which is not good particularly if we have many tenants.

src/OrchardCore.Modules/OrchardCore.OpenId/Services/OpenIdServerService.cs
jtkech: Okay

src/OrchardCore.Modules/OrchardCore.Tenants/Services/TenantValidator.cs
jtkech: Also in #12683 I made the hosts checking case insensitive, and we were assuming in the unit tests for example that an empty url prefix is not the same as a null prefix, but both end up with no prefix.

hishamco: > So it also splits on space that may result in individual spaces (if any) in the returned array, then because we use the RemoveEmptyEntries all these spaces will be removed.

`RemoveEmptyEntries` should remove the empty entries if they are exists, but I don't think they remove the leading or trailing spaces in entries

jtkech: Because we are also splitting on space (as we done e.g. in RunningShellTable.cs), for example if you had 3 consecutive spaces, 3 empty entries will be generated but none of these spaces will be at the beginning or at the end of any entry. So here we only need to remove these empty entries.

hishamco (approved):

Issue #11672, opened 2022-05-10
BUG: 'zh' localization directory not exist


IsQiao:
### To Reproduce
1. when my chrome broswer language is '中文(简体)', the `CurrentUICulture` field value is `zh`,
2. then ModularPoFileLocationProvider()  can't find the `zh` localization dir. because it is not existed.
3. so the website always display `en-US` culture.

### Expected behavior
show Chinese properly

### Screenshots
![image](https://user-images.githubusercontent.com/8536046/167640053-6e40598b-9eca-4229-904e-7f80a2cb3301.png)
![image](https://user-images.githubusercontent.com/8536046/167640564-c259616f-b23f-438e-867b-1ef7c3a7bc32.png)
![image](https://user-images.githubusercontent.com/8536046/167641150-234b2f2b-fcea-4aec-ad34-b66f8d49f26d.png)
![image](https://user-images.githubusercontent.com/8536046/167641484-e085f1e5-cc4f-475c-8b9d-a437ba96b648.png)
![image](https://user-images.githubusercontent.com/8536046/167641642-8c24bbc4-adfb-464d-aa43-f304333d7aaa.png)

Skrypt:
There are discrepancies between different OS with these culture codes. You simply need to rename the folder of your choice to the appropriate culture code for this to work.

IsQiao:
rename the directory can resolve it, but There are same problems in macos and windows 10. 
so i suguest to add `zh` dir to `OrchardCore.Translations` repo, or add some codes to enhance it.

Skrypt:
@agriffard Does Crowdin supports adding the "zh" culture or do we keep only specific culture/regions culture codes?

hyzx86:
It looks like it's just a display problem here, because it works fine on setup page, but once you switch to English, you can't change it back to Chinese

The option `zh-CN` is not listed  here

But I remember that there was no problem with some previous versions, I wonder why `zh-CN` was deleted

@agriffard

hishamco:
FYI Crowdin nowdays have both `zh-CN` and `zh-TW`, so when you choose `zh` can't find the root culture, so you can choose one of the supported cultures or rename the culture folder as workaround

hyzx86:
@hishamco No, that is not a right way.
As a Web application, we should follow the browser Settings, not let the user choose

![image](https://user-images.githubusercontent.com/15613121/210248263-4c2ed708-dbc7-4039-8093-ddec3f5833db.png)

[...]

PR #13023, opened 2023-01-02
Fix zh cn

hyzx86:
fix : https://github.com/OrchardCMS/OrchardCore/issues/11672

In the recent update of System.Runtime, the `CultureInfo.GetCultures(CultureTypes.AllCultures)` 
command does not directly list zh-CN

In order to keep the browser consistent, a patch is made here to get the option of zh-CN, as it is only used for setting up the page, which should not cause performance problems

After fix:

image

image

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Localization/Drivers/LocalizationSettingsDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Localization/Drivers/LocalizationSettingsDisplayDriver.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


hishamco:
I disagree with the change, we need to resolve it in better way, let me check the issue carefully

hishamco:
@hyzx86 check my comment [here](https://github.com/OrchardCMS/OrchardCore/issues/11672#issuecomment-1369011238)

hyzx86:
@hishamco 
You can check this file, it contains a lot of handling of `zh` ,
And the latest version is quite different, so I added  `.Distinct()`


image

[CultureInfo.cs.txt](https://github.com/OrchardCMS/OrchardCore/files/10332232/CultureInfo.cs.txt)

hishamco:
As I mentioned in the original issue the issue is not in .NET, OC just provided two cultures by default I'm not sure if we rename `zh-CN` to `zh` is right or not?!! But I think it's wrong because it targets both Chinese and Taiwan

Still I'm wondering should we have to support `zh`? If yes what about `zh-CN` and `zh-TW`?

hishamco:
Closing this, while this is not a proper fix

hishamco:
Closing this, while this is not a proper fix

hishamco:
Seems we need to modify your PR as temp & quick fix

hishamco:
I wasn't able to commit, so I will close this in favour of #13047

Issue #13024, opened 2023-01-02
Dynamic permission can't save in role


hyzx86:
bug in pull: https://github.com/OrchardCMS/OrchardCore/pull/12510

1. Edit a role 
2. Select some dynamic permissions , like 'ExecuteApi_RecentBlogPosts'
3. Click save .
4. Reopen the role , ExecuteApi_RecentBlogPosts is not checked

cuz the code only can load the static permissions , the program will break on line 235

https://github.com/OrchardCMS/OrchardCore/blob/083d5d7f9885b6ba034f642259cb94326af9e551/src/OrchardCore.Modules/OrchardCore.Roles/Controllers/AdminController.cs#L219-L252

We should load all permissions this way:

https://github.com/OrchardCMS/OrchardCore/blob/083d5d7f9885b6ba034f642259cb94326af9e551/src/OrchardCore.Modules/OrchardCore.Roles/Controllers/AdminController.cs#L184-L185

hyzx86:
https://user-images.githubusercontent.com/15613121/210260619-e0620501-62a0-4229-892b-532e0fd1cee8.mp4

hyzx86:
https://user-images.githubusercontent.com/15613121/210268379-270677fb-5471-476d-8de2-1c8c664938c8.mp4

gvkries:
Hehe, I just wanted to open the same bug. Actually there is a second even bigger issue with this PR. The way the `RoleUpdater` is now implemented causes it to always overwrite any admin settings, whenever the site starts. E.g. the 'Anonymous' role is now always getting the 'ViewContent' permission on startup, which breaks the whole content type permission system. Every content type is then allowed implicitly for everyone.

hyzx86:
It seems that I should turn off PR first  🤣,
I haven't looked at the code carefully, I just let it work
It was embarrassing to find out when the program was released yesterday.. The role permission cannot be changed

jtkech:
@hyzx86 

I think you were in the right path, yes using `GetInstalledPermissionsAsync()` in the controller edit post seems to fix issue because it uses `GetPermissionsAsync()` that also gets dynamic permissions.

For info I could also make it working by just using

            var permissionNames = new List<string>();
            foreach (var permissionProvider in _permissionProviders)
            {
                var permissions = await permissionProvider.GetPermissionsAsync();
                permissionNames.AddRange(permissions.Select(permission => permission.Name));
            }

Maybe you could re-open your PR ;)

[...]

PR #13025, opened 2023-01-02
fix role update

### LINKED ISSUE #13035: Role permission settings are overwritten if a tenant is started
### Describe the bug

Whenever a tenant is activated, the `RoleUpdater` reapplies any default permissions  to the roles. This overwrites any adjustments made from the admin dashboard. This breaks all permissions assigned to specific content types, because e.g. the anonymous role is always reset.

### To Reproduce
Example steps to reproduce the behavior:
1. Go to security -> roles and edit the 'Anonymous' role.
2. Remove 'View' the permission from 'Content' and save.
3. Restart the web app.
4. The removed permission is added back.

### Expected behavior
Permission settings on roles are persisted between site restarts.

This bug was introduced by PR #12510.
See also bug #13024.

hyzx86:
fix : https://github.com/OrchardCMS/OrchardCore/issues/13024

Planned to also fix #13035.

https://user-images.githubusercontent.com/15613121/210268280-d710778f-4781-4c83-be93-2f5a62845831.mp4

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Roles/Controllers/AdminController.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Roles/Controllers/AdminController.cs
@@ -217,7 +217,7 @@ namespace OrchardCore.Roles.Controllers
             rolesDocument.PermissionGroups.TryAdd(role.RoleName, new List<string>());
 
             var installedPermissions = await GetInstalledPermissionsAsync();
-            var allPermissions = installedPermissions.SelectMany(x => x.Value).Select(x => x.Name);
+            var permissionNames = installedPermissions.SelectMany(x => x.Value).Select(x=>x.Name);
 
             // Save
             var rolePermissions = new List<RoleClaim>();
@@ -230,7 +230,7 @@ namespace OrchardCore.Roles.Controllers
                     permissionName = key.Substring("Checkbox.".Length);
                 }
 
-                if (!allPermissions.Contains(permissionName, StringComparer.OrdinalIgnoreCase))
+                if (!permissionNames.Contains(permissionName, StringComparer.OrdinalIgnoreCase))
                 {
                     // The request contains an invalid permission, let's ignore it
                     continue;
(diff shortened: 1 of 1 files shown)


jtkech:
@hyzx86 

Thanks for re-opening this PR but maybe I will open a new PR as I already started to work on it locally to check different things as per my comment https://github.com/OrchardCMS/OrchardCore/issues/13024#issuecomment-1369392407

@MikeAlhayek maybe a different approach for example to not execute the RoleUpdater on each shell activation if possible, I think I will ask you some questions to confirm what was intended to be done.

Hmm, one question on the fly, yes in the controller, dynamic permissions was not taken into account, hmm maybe we need to take them into account also in the RoleUpdater, will see.

MikeAlhayek:
@jtkech I in the `RoleUpdater` I think we have to call `GetDefaultStereotypes()` to group the permissions role this way we know which permissions should belong to Stereotype/role. If the grouped role is known to the current site, we then assign it. Once we assign a permission, we save the reference into `RolesDocument.PermissionGroups` this way we can know which permissions were auto assigned vs the assignments that were manually made "a user checking/unchecking" permissions.

In the Admin I agree we should not be using `GetDefaultStereotypes()`

We should be using the following as per your suggestions.
```
var permissionNames = new List<string>();
foreach (var permissionProvider in _permissionProviders)
{
    var permissions = await permissionProvider.GetPermissionsAsync();
    permissionNames.AddRange(permissions.Select(permission => permission.Name));
}
```

or 

```
var allPermissions = await Task.WhenAll(_permissionProviders.Select(x => x.GetPermissionsAsync()));
```

jtkech:
Even if here all `GetPermissionsAsync()` seem to be synchronous (using Task.FromResult), and in that case it would be useless, I never use `Task.WhenAll()` ;) Unless I'm sure all are thread safe, which is not the case e.g. with `ISession`.

So okay the other breaking change, as stated by @gvkries https://github.com/OrchardCMS/OrchardCore/issues/13035#issuecomment-1370286327, is that on an existing installation not having yet the permissions history, some permissions get back to some roles on startup.

jtkech:
Okay I'm  closing this one as I started to work on it on another branch #13040.

@hyzx86 @gvkries I will let you know when it will be ready to be tested.

PR #13028, opened 2023-01-03
Fix list part navigation on mobile

### LINKED ISSUE #12949: The ListPart navigation is showing up over the Admin menu on mobile device
### Describe the bug
The ListPart had enhancement in v1.5 where we added a Navigation bar over for the container and it's contained items. The navigation has a z-index value of 1001 while the action-bar has a z-index of 1000.  the navigation bar should have a higher z-index than the action-bar to make sure any menu from the ListNaviation to show up over the action-bar. 

### Expected behavior
I expect the admin menu to show up correctly on the top when it is open.

### Screenshots
image

DGDario:
Fixes #12949

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Lists/Views/ListPartNavigationAdmin.cshtml
+++ b/src/OrchardCore.Modules/OrchardCore.Lists/Views/ListPartNavigationAdmin.cshtml
@@ -4,7 +4,7 @@
 @inject OrchardCore.ContentManagement.IContentManager ContentManager
 @inject Microsoft.AspNetCore.Authorization.IAuthorizationService AuthorizationService
 
-<div class="card bg-primary text-right mb-3" style="z-index: 1001">
+<div class="card bg-primary text-right mb-3" style="z-index: 1000">
     <div class="card-body">
         <div class="d-flex justify-content-end">
 
--- a/src/OrchardCore.Themes/TheAdmin/wwwroot/css/TheAdmin.css
+++ b/src/OrchardCore.Themes/TheAdmin/wwwroot/css/TheAdmin.css
@@ -27846,10 +27846,6 @@ html[dir=ltr][data-theme=default] .left-sidebar-compact .ta-navbar-top .nav-item
 html[dir=rtl][data-theme=default] .left-sidebar-compact .ta-navbar-top .nav-item {
   margin-right: 10px;
 }
-html[data-theme=default] .action-bar {
-  top: 3.3rem;
-  z-index: 1050;
-}
 html[data-theme=default] .ta-navbar-top {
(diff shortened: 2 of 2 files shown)


MikeAlhayek:
I don't think this would fix the problem. 

Assign multiple content types to the list part settings so that the "New" button is a drop down menu. Then try to click on the "Add" button to open the menu. Will the menu show up over the action bar?

DGDario:
You are right, you made a huge use of z-index property. With the last css changes it will work (I've also removed the duplicated 
action-bar css definition)

MikeAlhayek:
Looks good to me. 

@Skrypt do you see any issue with changed the `z-index` of the action bar to 999 instead of 1000? I think it is safe

Skrypt:
Hmm, everything is there for a reason. Maybe it will fix the issue on mobile size but break something else somewhere else. The global layout z-indexes have been thinked based on the z-index defined by Bootstrap first and then some custom ones.

src/OrchardCore.Modules/OrchardCore.Lists/Views/ListPartNavigationAdmin.cshtml
```diff
@@ -4,7 +4,7 @@
 @inject OrchardCore.ContentManagement.IContentManager ContentManager
 @inject Microsoft.AspNetCore.Authorization.IAuthorizationService AuthorizationService
 
-<div class="card bg-primary text-right mb-3" style="z-index: 1001">
+<div class="card bg-primary text-right mb-3" style="z-index: 1000">
```
Skrypt, line 7: This should use a class instead of inline styles. Should be also using Bootstrap z-index if we can.

src/OrchardCore.Themes/TheAdmin/wwwroot/css/TheAdmin.css
```diff
@@ -29037,7 +29033,7 @@ html[dir][data-theme=default] .disable.custom-control-input:checked ~ .custom-co
 }
 html[data-theme=default] .action-bar {
   top: 3.3rem;
-  z-index: 1000;
+  z-index: 999;
```
Skrypt, line 29036: This is the wrong place to make that change. It needs to be changed on the .scss file that defines that style. And I would be careful about changing that z-index on a global layout element.

MikeAlhayek: good catch! both css changes need to be done in the `.scss` file then run `npm install && gulp` to generate the new css.

Skrypt:
Duplicate of https://github.com/OrchardCMS/OrchardCore/pull/13062

PR #13046, opened 2023-01-06
Add title to background tasks and improve the UI

### LINKED ISSUE #13045: Add user friendly title to background tasks instead of showing the technical name
Currently, a background name/technical name is the full name of the `C#` class. This is not a user friendly name for a UI. I think the long Technical Name (aka task Id) should not be visible to the UI. I mean why should the user care about the technical name of the background task.

Add a user friendly title to every background tasks. We can continue to use the current "name" in the URL and to locate the task.

MikeAlhayek:
Fix #13045 

Here is a summary of the changes
- New search bar was added to the list page
- Technical Name is no longer visible. Instead we show a title.
- The Create action was removed from the controller as we can never create a background task using a UI.
- The Edit action will now create an settings-entry in the document if an entry does not exists.

## Here are some screenshots after this PR

image

image

## Here are some screenshots prior this PR

image

image


hishamco:
Are we agreed on that?

Is that title comes from the user input? if Yes, how we can guarantee that there's no duplicate. BTW I'm not against the idea, but sometimes we need to discuss before submitting a PR

MikeAlhayek:
The title is something you define in the Attribute. We still identify the task by it type's full name so even if we have name conflict, it should not matter.

sebastienros (approved):

PR #13047, opened 2023-01-06
Add Chinese Alias Cultures for Backward Compatibility

hishamco:


jtkech:
Okay that's another solution to still make them available from the UI in place of adding them dynamically, I have no closed opinion, in both case the aspnet middleware will be able to select them to mutate the thread CurrentCulture.

Then about still using the additional mapping provider and middleware to set again the thread CurrentCulture, I'm not sure, unless it is too annoying to still maintain the old localization folders, but for a full compatinility (even if only temporary) we would need to do it.

hishamco:
@jtkech I deleted the introduced middleware in #13153 according to your recommendation

PR #13049, opened 2023-01-07
Map Chinese Alias Cultures

hishamco:


hishamco:
BTW may be we can map alias to the culture in Localization APIs instead

sebastienros (approved):

PR #13051, opened 2023-01-08
Remove IContentItemFactory

MikeAlhayek:
Instead of using the new `IContentItemFactory` use `IAuthorizationService` extension for simpler approach and use less memory.


MikeAlhayek:
@jtkech I have not had time to test this PR out. I'll test it out tomorrow.

jtkech:
@MikeAlhayek Okay no problem, will see tomorrow

MikeAlhayek:
@jtkech I tested this out and it works as expected with no problems. Feel free to merge it if you are good with everything.

PR #13066, opened 2023-01-10
Fix UserStore by saving/removing the role from the user

### LINKED ISSUE #13065: User to roles are not saving
### Describe the bug
After the last update to the UserStore, something got missed where the `AddToRoleAsync` does not add role and  `RemoveFromRoleAsync` does not remove a role.

I am submitting a PR for this.

MikeAlhayek:
Fix  #13065

@jtkech this is a patch for the PR we last merged. seems that this was missed.

### DIFF
--- a/src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
+++ b/src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
@@ -467,6 +467,8 @@ namespace OrchardCore.Users.Services
                 {
                     throw new InvalidOperationException($"Role {normalizedRoleName} does not exist.");
                 }
+
+                u.RoleNames.Add(normalizedRoleName);
             }
         }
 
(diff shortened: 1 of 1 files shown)


jtkech:
Okay

Hmm, just as a reminder

When you introduced the `is User` directive in place of the `(User)user` casting, I also noticed that it changed a little the behavior, if not a `User` before it would fail on the casting, now it silently do nothing, but seems to not be so important.

Hmm, that said in the following, if the `userToken` is null and user not a `User` in place of failing on the `(User)user` casting, it would now fail at the end on a null `userToken`, if we follow the goal to not throw maybe a missing null check on the `userToken`.

            if (userToken == null && user is User u)
            {
                userToken = new UserToken
                {
                    LoginProvider = loginProvider,
                    Name = name
                };

                u.UserTokens.Add(userToken);
            }

            // Encrypt the token
            userToken.Value = _dataProtectionProvider.CreateProtector(TokenProtector).Protect(value);

PR #13076, opened 2023-01-12
Use zh-Hans-CN & zh-Hant-TW cultures in setup

hishamco:


src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
```diff
@@ -19,8 +19,9 @@ public class Startup : StartupBase
     {
         private readonly string _defaultCulture = CultureInfo.InstalledUICulture.Name;
 
-        private string[] _supportedCultures = new string[] {
-            "ar", "cs", "de", "el", "en", "es", "fa", "fr", "it", "ja", "pl", "pt-BR", "ru", "sv", "tr", "vi", "zh-CN", "zh-TW"
+        private string[] _supportedCultures = new string[]
+        {
+            "ar", "cs", "de", "el", "en", "es", "fa", "fr", "it", "ja", "pl", "pt-BR", "ru", "sv", "tr", "vi", "zh-Hans-CN", "zh-Hant-TW"
```
hishamco, line 24: @sebastienros I think we should add `zh-Hans` & `zh-Hant` after your latest changes, right?

Skrypt (dismissed):

src/OrchardCore.Modules/OrchardCore.Setup/Startup.cs
Skrypt: Should it not be zh-Hans and zh-Hant in these values then?
You need to make sure that it will work with the actual Crowdin generated localization folders. Maybe we should fix that part before.

hishamco: Me & Seb planned to add `zh-Hans-CN` & `zh-hant-TW` at the beggining, then Seb changes his mind to support `zh-Hans` & `zh-Hant` ;)

Skrypt: Yes, that makes more sense. But then, make sure that it will pick up an actual localization folder that works. These cultures listed there are cultures that matches localization folders that we have that comes from Crowdin.

hishamco: > Yes, that makes more sense

But if you look from different angle, all the specific Chinsese cultures will be fallback to their parent which is weird, I'm sure `zh-Hant-TW` is different tham `zh-Hant-SG` for instance, but the introduced why will treat them same as tranditional chinese /cc @sebastienros

jtkech:

PR #13077, opened 2023-01-12
Use ListContent permission instead of ViewContent

### LINKED ISSUE #13053: Use ListContent permission in the Content list action instead of ViewContent
### Describe the bug
Currently, we use `ViewContent` as a min permission to list the contents in the content items UI. We use the same permission to also show the "Content >> Content Items" admin menu. IMO, `List Content` permission should be used not `View Content`. What if someone wants to allow a user to view content using direct link, but does not want them to list the contents in the UI? Currently, this isn't possible.

Additionally, `ViewContent` is granted to all `Anonymous` and `Authenticated` roles which is fine. But, I think access to list contents should by granted by the `ListContent` permission to avoid allowing listing content without explicit permission grant.

Side note: a thought that came to mind, maybe `ViewContent` should also be implied by `ListContent` which isn't the case today. I think we can leave that as is just for added flexibility in case for some reason someone wants to allow listing contents but NOT viewing them which sounds weird at the moment. But, who knows all use cases. 

If I get agreement on this change, I'll submit a PR and document the change in the release notes.

MikeAlhayek:
Fix #13053

### DIFF
--- a/src/docs/releases/1.6.0.md
+++ b/src/docs/releases/1.6.0.md
@@ -12,4 +12,4 @@ If your are using setup recipes and want to enable `Features`, be sure to explic
 
 ## Change Logs
 
-Prior this release, a user with `ViewContent` permission and has access to the Dashboard, has access to "Content Items". Now, to be able to list contents on the dashboard, a user needs `ListContent` permission. 
+Prior this release, a user with `ViewContent` permission and has access to the dashboard, automatically gets access to "Content Items". Now, to be able to list contents on the dashboard, a user needs `ListContent` permission instead. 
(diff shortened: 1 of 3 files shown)


src/docs/releases/1.6.0.md
```diff
@@ -9,3 +9,7 @@ Release date: Not yet released
 `OrchardCore.Features` is no longer auto enabled by `OrchardCore.Recipes` as it doesn't depend on it anymore. If you find yourself using recipes services like `IRecipeMigrator` in your module, be sure to add a dependency on `OrchardCore.Recipes.Core` feature in your module's Manifest file to ensures that the recipe services are available before using them.
 
 If your are using setup recipes and want to enable `Features`, be sure to explicitly enable `OrchardCore.Feature` feature.
+
+## Change Logs
+
+Prior this release, a user with `ViewContent` permission and has access to the Dashboard, has access to "Content Items". Now, to be able to list contents on the dashboard, a user needs `ListContent` permission.
```
MikeAlhayek, line 15:
```suggestion
Prior this release, a user with `ViewContent` permission and has access to the dashboard, automatically gets access to "Content Items". Now, to be able to list contents on the dashboard, a user needs `ListContent` permission instead. 
```

PR #13084, opened 2023-01-14
Prevent the owner editor driver from throwing exception when a user is deleted

### LINKED ISSUE #13083: OwnerEditorDriver throw exception if the user is deleted
### Describe the bug
`CommonPart` allows the site-owner to change the owner of the content item. However, if a content item is owned by a deleted user, the `OwnerEditorDriver` throws a null reference exception. 

Submitting a PR for this soon.

MikeAlhayek:
Fix #13083

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Contents/Drivers/OwnerEditorDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Contents/Drivers/OwnerEditorDriver.cs
@@ -1,4 +1,4 @@
-using System.Security.Claims;
+using System;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http;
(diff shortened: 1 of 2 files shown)


hishamco:
Why not we change the ownership if we delete the user, hmm probably we can disable the user to avoid he log into the system

MikeAlhayek:
Sure. But that would be a different PR. During the removal of a user, you would ask which owner should we assign the content to, and assign accordingly. 

In this PR, I just want to fix the existing issue without creating a new process.

hishamco:
> In this PR, I just want to fix the existing issue without creating a new process.

I understand, but this could be fixed by making the user inactive instead of deletion

MikeAlhayek:
Sure you can do that on your own site. But you can't enforce everyone to follow your business logic. Since OC allows deletion, we should account for that when we try to find a use"aka do the null check"

hishamco:
lol, but what you did here is just a workaround by changing the owner with the current logged user ;)

MikeAlhayek:
We kinda did the same thing before were we assumed to use the current user where the owner is null. If we had an actual user picker instead of text box, we would have required a valid user selection. We could throw an error and ask the user to provide a valid username instead.

hishamco:
We might accept this as workaround, but it's better to hide the deleted used from the system to avoid such behaviour or create such `Unknown` or `Anonymous` user that will be added during setup - which is invisible on the system - just to attach content related for the delete users

hishamco:
We might accept this as workaround, but it's better to hide the deleted used from the system to avoid such behaviour or create such `Unknown` or `Anonymous` user that will be added during setup - which is invisible on the system - just to attach content related for the delete users

MikeAlhayek:
@hishamco I changed it so we do not assume that the new owner is the current users if the user is deleted or invalid. If the username is missing or invalid, we require the user to provide a valid username.

hishamco:
Much better, let us hear others feedback

sebastienros (approved):

PR #13093, opened 2023-01-16
Avoid ignoring existing culture providers

hishamco:


jtkech:
We were already resolving and using `IOptions<RequestLocalizationOptions>` before.

I think it has been broken when you created a derived `OrchardCoreRequestLocalizationOptions` but used it without taking into account the current `RequestLocalizationOptions` values.

    var requestLocalizationOptions = new OrchardCoreRequestLocalizationOptions(...)
        .SetDefaultCulture(defaultCulture)
        .AddSupportedCultures(supportedCultures)
        .AddSupportedUICultures(supportedCultures);

Okay, here you are now reporting `.RequestCultureProviders` but you are still missing other `RequestLocalizationOptions` properties.

One solution would be to have a constructor accepting a `RequestLocalizationOptions` param to init all inner properties of this derived class. Hmm, but maybe better to create extensions acting directly on `RequestLocalizationOptions` in place of this new derived class.

hishamco:
> I think it has been broken when you created a derived OrchardCoreRequestLocalizationOptions but used it without taking into account the current RequestLocalizationOptions values.

Ya we forgot there's a culture provider is set by `OC.ContentLocalization` :)

> One solution would be to have a constructor accepting a RequestLocalizationOptions param to init all inner properties of this derived class

Thanks for your input, I already thought about that before I started this PR, but the important thing is `RequestCultureProviders` coz are properties are set by localization module

> but maybe better to create extensions acting directly on RequestLocalizationOptions in place of this new derived class.

Hmm, let me think about it

hishamco:
BTW I need to keep everything simple without introducing new things - if it's possible - coz `OrchardCoreRequestLocalizationOptions` will be dropped once we use .NET 8 in the upcoming future https://github.com/dotnet/aspnetcore/pull/44282

jtkech:
Okay, what I meant, at least for now, is that we should also pass all properties (other than cultures that we set explicitly) of `RequestLocalizationOptions`, because they can be configured by anybody, I don't know for example if someone configure the `ApplyCurrentCultureToResponseHeaders` property.

This as it was before.

hishamco:
Agree with other than `SupportedCultures` & `SupportedUICultures` coz they will be overridden by `OC.Localization`

hishamco:
> I don't know for example if someone configure the ApplyCurrentCultureToResponseHeaders property.

I forgot that there are some useful options that can be set by other modules such as `FallBackToParentCultures` & `FallBackToParentUICultures`

@jtkech I just added a simple extension method in `OC.Localization` module which is fine IMHO, I tried to don't touch `OrchardCoreRequestLocalizationOptions` which will be remove after .NET 8

jtkech:
So as said because your derived `OrchardCoreRequestLocalizationOptionsExtensions` defines methods returning the base type `RequestLocalizationOptions`, when we chain these methods only the first one executes our custom implementation, the others execute the non virtual ones defined by the base class, so without taking into account the new `ignoreSystemSettings`.

So just before only `.SetDefaultCulture()` was using the new `ignoreSystemSettings`. But because you inserted `WithRequestLocalizationOptions()` that also returns the base `requestLocalizationOptions`, now also `.SetDefaultCulture()` doesn't execute our custom implementation.

So now we never take into account the new `ignoreSystemSettings` ;)

"Worst" than before

hishamco:
Seems this got merge before the PR #13153, @jtkech let us either merge your PR #13158 which have the same changes in #13153 except `WithRequestLocalizationOptions` or I can add it to my PR. Please let us do it ASAP

PR #13096, opened 2023-01-17
Switched to AsyncKeyedLock

MarkCiliaVincenti:


hishamco:
@MarkCiliaVincenti please fix the build

MarkCiliaVincenti:
> @MarkCiliaVincenti please fix the build

Done.

MarkCiliaVincenti:
This PR has two parts:

One that updates LocalLock to a more performant library that also reduces allocations via pooling.

The second is to make ModularTenantRouterMiddleware and DistributedShellHostedService and ShellHost also use the LocalLock instead of having their own implementation.

hishamco:
@MarkCiliaVincenti seems some unit tests are broken

hishamco:
@MarkCiliaVincenti seems some unit tests are broken

MarkCiliaVincenti:
> @MarkCiliaVincenti seems some unit tests are broken


Ah yes, some DI issue. Will fix tomorrow.

MarkCiliaVincenti:
> @MarkCiliaVincenti seems some unit tests are broken

Fixed.

src/OrchardCore/OrchardCore/OrchardCore.csproj
```diff
@@ -14,6 +14,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="AsyncKeyedLock" Version="6.0.5" />
```
hishamco, line 17: Move the version to `Dependencies.props`

MarkCiliaVincenti: Done. No need for Dependencies.AspNetCore.props right?

MarkCiliaVincenti:
Sorry didn't realise that action would remove @jtkech from reviewer.

MarkCiliaVincenti:
@jtkech looking forward to your review :)

jtkech:
Okay, will try to look at it asap.

MarkCiliaVincenti:
> This PR has two parts:
> 
> One that updates LocalLock to a more performant library that also reduces allocations via pooling.
> 
> The second is to make ModularTenantRouterMiddleware and DistributedShellHostedService and ShellHost also use the LocalLock instead of having their own implementation. Please take care of the nuancities: they will now share the dictionary meaning that if they two different classes lock on the same key they would be waiting each other. If this is not the desired behaviour then these 3 classes could create their own instance of AsyncKeyedLocker and use the library directly.
> 
> @jtkech @hishamco

Gentle reminder on these notes @jtkech

MarkCiliaVincenti:
> Okay, will try to look at it asap.

Gentle reminder @jtkech :)

MarkCiliaVincenti:
@jtkech  :)

sebastienros:
Pending tests. Not for 1.6 because it's risky, need to be taken early after 1.6 is released.

MarkCiliaVincenti:
> Pending tests. Not for 1.6 because it's risky, need to be taken early after 1.6 is released.

Thanks @sebastienros. Is there an ETA for that?

MarkCiliaVincenti:
> Pending tests. Not for 1.6 because it's risky, need to be taken early after 1.6 is released.

@sebastienros @jtkech @hishamco I've merged now, is there anything else you need from my end please?

sebastienros (approved):

jtkech:
I added the don't merge label just because I want to review it before, I'll try to do it this night.

jtkech:
Okay, so I removed the don't merge label and did a quick review.

I also took a look at your repo, the way a lock is acquired/released looks quite similar, the main difference being the pool of reusable locks. Maybe worth to integrate pooling but I think it is better to not depend on an external library/repo for this, for example if we need to do some reviews.

Personally, I would prefer to first keep the existing code and that you suggest a PR to add the pooling mechanism but in our repo, that would be easier to review and manage.

Note: I kept the don't merge label because there are some key names conflicts at the host level.

MarkCiliaVincenti:
> Okay, so I removed the don't merge label and did a quick review.
> 
> I also took a look at your repo, the way a lock is acquired/released looks quite similar, the main difference being the pool of reusable locks. Maybe worth to integrate pooling but I think it is better to not depend on an external library/repo for this, for example if we need to do some reviews.
> 
> Personally, I would prefer to first keep the existing code and that you suggest a PR to add the pooling mechanism but in our repo, that would be easier to review and manage.
> 
> Note: I kept the don't merge label because there are some key names conflicts at the host level.

Thank you for your input, [Jean-Thierry](https://github.com/jtkech). The library has a similar method that uses pooling, as well as a completely different one using striped locking. Both methods are more performant and reduce more allocations than what is currently implemented in OrchardCore, and this project would benefit from this PR.

Should you not want to add in a dependency, though, I'm afraid I don't have the time to add pooling within the existing code and testing for all the race conditions possible. The latter can be quite tricky and that's why I suggest going with a tried & tested implementation that is in use in several projects, rather than asking me to reinvent my own wheel and potentially introduce bugs into your codeset.

Let me know what you intend to do.

jtkech:
@MarkCiliaVincenti

No problem I understand and I'm not against using your library ;) Just want to understand.

So currently at the tenant level we only resolve `IDistributedLock` whose default implementation is `LocalLock` but not so much and not in hot paths, for example when we activate a tenant container but only once for the lifetime of a tenant until it is released/reloaded. Also rarely when executing an atomic workflow or an atomic background task, the min period of a background task being one minute.

I did some tests, we can allocate 10000 `SemaphoreSlim` in 6 ms, around 600 nano seconds per semaphore, the allocation being around 80 bytes per semaphore. So at the tenant level I don't think we need pooling.

Where it would make more sense is at the host level where in 3 places we use a concurrent dictionary of `SemaphoreSlim` (one per tenant), in one place we use it once per tenant when we build a tenant container, in another place once per tenant when we build its pipeline, in the last place when a tenant has changed in a distributed environment. Hmm, but most of the time we only have one tenant (the Default one), sometimes a few ones, sometimes tens but much more rarely e.g. hundreds of tenants.

> Both methods are more performant and reduce more allocations than what is currently implemented in OrchardCore.

As said they are not used in hot paths but are you sure about performance ;) About allocations, yes but only significant if we have many tenants. I will continue to think about it.

In the meantime I started to look at your code, yes the `IsNotInUse` management is not so easy to follow, I have one question about the following code in `AsyncKeyedLockDictionary`, if I understood correctly we retrieve a releaser from the dictionary or from the pool, in both case it is ready to use with `IsNotInUse` set to false, so in the while loop at the end I don't understand why `IsNotInUse` is set to true.

            if (PoolingEnabled)
            {
                if (TryGetValue(key, out var releaser) && releaser.TryIncrement(key))
                {
                    return releaser;
                }

                var releaserToAdd = _pool.GetObject(key);
                if (TryAdd(key, releaserToAdd))
                {
                    return releaserToAdd;
                }

                while (true)
                {
                    releaser = GetOrAdd(key, releaserToAdd);
                    if (ReferenceEquals(releaser, releaserToAdd))
                    {
                        return releaser;
                    }
                    if (releaser.TryIncrement(key))
                    {
                        releaser.IsNotInUse = true; // <= Maybe 'releaserToAdd.IsNotInUse' instead.
                        _pool.PutObject(releaserToAdd);
                        return releaser;
                    }
                }
            }

In the above code maybe you meant `releaserToAdd.IsNotInUse` in place of `releaser.IsNotInUse`. If there is a releaser that we can't increment can we continue the loop like this, normally it means that this releaser is being released, hmm maybe we could try to remove it from the dico.

If I understood correctly, maybe we could use something like this.

            if (PoolingEnabled)
            {
                AsyncKeyedLockReleaser<TKey> releaserToAdd = null;
                while (true)
                {
                    if (TryGetValue(key, out var releaser))
                    {
                        if (!releaser.TryIncrement(key))
                        {
                            TryRemove(key, out _);
                            continue;
                        }

                        if (releaserToAdd != null)
                        {
                            releaserToAdd.IsNotInUse = true;
                            _pool.PutObject(releaserToAdd);
                        }

                        return releaser;
                    }

                    releaserToAdd ??= _pool.GetObject(key);

                    if (TryAdd(key, releaserToAdd))
                    {
                        return releaserToAdd;
                    }
                }
            }

Also in `AsyncKeyedLockReleaser` I saw this comment about a race condition, can you elaborate? Does it mean that you had the case where a releaser with a given key was in the dictionary but at a different key?

                if (IsNotInUse || !_key.Equals(key)) // rare race condition

jtkech:
Also when an item can't be returned back to the pool because of the bounded collection capacity, as I can see it is not disposed. Just for info maybe worth to look at this class in aspnetcore

https://github.com/dotnet/aspnetcore/blob/d07f377b6c4c72f44a0eb2810c8abf1ae3fabd40/src/ObjectPool/src/DefaultObjectPool.cs#L15

Here a disposable version

https://github.com/dotnet/aspnetcore/blob/d07f377b6c4c72f44a0eb2810c8abf1ae3fabd40/src/ObjectPool/src/DisposableObjectPool.cs#L9

Here a dedicated pool for cancellation token sources.

https://github.com/dotnet/aspnetcore/blob/d07f377b6c4c72f44a0eb2810c8abf1ae3fabd40/src/Shared/CancellationTokenSourcePool.cs#L8-L13

MarkCiliaVincenti:
> @MarkCiliaVincenti
> 
> No problem I understand and I'm not against using your library ;) Just want to understand.
> 
> So currently at the tenant level we only resolve `IDistributedLock` whose default implementation is `LocalLock` but not so much and not in hot paths, for example when we activate a tenant container but only once for the lifetime of a tenant until it is released/reloaded. Also rarely when executing an atomic workflow or an atomic background task, the min period of a background task being one minute.
> 
> I did some tests, we can allocate 10000 `SemaphoreSlim` in 6 ms, around 600 nano seconds per semaphore, the allocation being around 80 bytes per semaphore. So at the tenant level I don't think we need pooling.
> 
> Where it would make more sense is at the host level where in 3 places we use a concurrent dictionary of `SemaphoreSlim` (one per tenant), in one place we use it once per tenant when we build a tenant container, in another place once per tenant when we build its pipeline, in the last place when a tenant has changed in a distributed environment. Hmm, but most of the time we only have one tenant (the Default one), sometimes a few ones, sometimes tens but much more rarely e.g. hundreds of tenants.
> 
> > Both methods are more performant and reduce more allocations than what is currently implemented in OrchardCore.
> 
> As said they are not used in hot paths but are you sure about performance ;) About allocations, yes but only significant if we have many tenants. I will continue to think about it.
> 
> In the meantime I started to look at your code, yes the `IsNotInUse` management is not so easy to follow, I have one question about the following code in `AsyncKeyedLockDictionary`, if I understood correctly we retrieve a releaser from the dictionary or from the pool, in both case it is ready to use with `IsNotInUse` set to false, so in the while loop at the end I don't understand why `IsNotInUse` is set to true.
> 
> ```
>         if (PoolingEnabled)
>         {
>             if (TryGetValue(key, out var releaser) && releaser.TryIncrement(key))
>             {
>                 return releaser;
>             }
> 
>             var releaserToAdd = _pool.GetObject(key);
>             if (TryAdd(key, releaserToAdd))
>             {
>                 return releaserToAdd;
>             }
> 
>             while (true)
>             {
>                 releaser = GetOrAdd(key, releaserToAdd);
>                 if (ReferenceEquals(releaser, releaserToAdd))
>                 {
>                     return releaser;
>                 }
>                 if (releaser.TryIncrement(key))
>                 {
>                     releaser.IsNotInUse = true; // <= Maybe 'releaserToAdd.IsNotInUse' instead.
>                     _pool.PutObject(releaserToAdd);
>                     return releaser;
>                 }
>             }
>         }
> ```
> 
> In the above code maybe you meant `releaserToAdd.IsNotInUse` in place of `releaser.IsNotInUse`.

Wooops! I think you're right there, though it's too early in the morning for me to think well whether or not that line of code is actually needed. If it does, this might have provided an optimization in rare cases :)

> If there is a releaser that we can't increment can we continue the loop like this, normally it means that this releaser is being released, hmm maybe we could try to remove it from the dico.
> 
> If I understood correctly, maybe we could use something like this.
> 
> ```
>         if (PoolingEnabled)
>         {
>             AsyncKeyedLockReleaser<TKey> releaserToAdd = null;
>             while (true)
>             {
>                 if (TryGetValue(key, out var releaser))
>                 {
>                     if (!releaser.TryIncrement(key))
>                     {
>                         TryRemove(key, out _);
>                         continue;
>                     }
> 
>                     if (releaserToAdd != null)
>                     {
>                         releaserToAdd.IsNotInUse = true;
>                         _pool.PutObject(releaserToAdd);
>                     }
> 
>                     return releaser;
>                 }
> 
>                 releaserToAdd ??= _pool.GetObject(key);
> 
>                 if (TryAdd(key, releaserToAdd))
>                 {
>                     return releaserToAdd;
>                 }
>             }
>         }
> ```

I'm a bit lost in what you are trying to do here.

The current code is that way so that if the first Get/Add is unsuccessful, it starts looping and using the GetOrAdd which is more performant (because in GetOrAdd the hashcode is calculated once rather than twice as what happens with individual TryGetValue and TryAdd).
 
> Also in `AsyncKeyedLockReleaser` I saw this comment about a race condition, can you elaborate? Does it mean that you had the case where a releaser with a given key was in the dictionary but at a different key?
> 
> ```
>             if (IsNotInUse || !_key.Equals(key)) // rare race condition
> ```

Yes, these avoid very rare race conditions and are related to https://github.com/MarkCiliaVincenti/AsyncKeyedLock/issues/5

jtkech:
Thanks for the link to the issue.

Okay I understand, even if the releaser mutations are atomic, the whole starting from the `TryGetValue()` of the concurrent dictionary is not atomic, you could lock the whole on the dictionary itself that could be then a simple dictionary (as we do), but maybe not compatible with the performance you want to achieve.

@sebastienros just for info

- I don't think we need pooling at the Tenant Level, currently `ILocalLock` is only registered at the tenant level and is the default implementation of `IDistributedLock`, it is not so often used and not in hot paths.

- Would make more sense at the Host Level where in 3 services we use dedicated dictionaries with one `SemaphoreSlim` per tenant, but they are not used in hot paths and creations and allocations of `SemaphoreSlim` are quite light.

-  If we really want pooling I think we could do it ourselves to not depend on an external library.

- So I let you decide, at least we would need to fix the key conflicts for the 3 host level services that were using the same keys but from separate dictionaries (they still should not block each other), and in the @MarkCiliaVincenti library would need to fix the issue that I pointed out above (if I was right).

MarkCiliaVincenti:
@sebastienros  are you interested or shall I close this PR?

MarkCiliaVincenti:
@sebastienros kindly revert.

PR #13104, opened 2023-01-17
Don't normalize User.RoleNames

jtkech:
As it was before.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Users/Services/SetupEventHandler.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Users/Services/SetupEventHandler.cs
@@ -29,9 +29,10 @@ namespace OrchardCore.Users.Services
                 UserName = properties.TryGetValue(SetupConstants.AdminUsername, out var adminUserName) ? adminUserName?.ToString() : String.Empty,
                 UserId = properties.TryGetValue(SetupConstants.AdminUserId, out var adminUserId) ? adminUserId?.ToString() : String.Empty,
                 Email = properties.TryGetValue(SetupConstants.AdminEmail, out var adminEmail) ? adminEmail?.ToString() : String.Empty,
-                RoleNames = new string[] { "Administrator" },
                 EmailConfirmed = true
             };
+
+            user.RoleNames.Add("Administrator");
 
             return _userService.CreateUserAsync(user, properties[SetupConstants.AdminPassword]?.ToString(), reportError);
         }
--- a/src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
+++ b/src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
@@ -462,13 +462,14 @@ namespace OrchardCore.Users.Services
             if (user is User u)
             {
                 var roleNames = await _roleService.GetRoleNamesAsync();
+                var roleName = roleNames?.FirstOrDefault(r => NormalizeKey(r) == normalizedRoleName);
 
                 if (!roleNames.Any(r => NormalizeKey(r) == normalizedRoleName))
                 {
                     throw new InvalidOperationException($"Role {normalizedRoleName} does not exist.");
                 }
 
-                u.RoleNames.Add(normalizedRoleName);
+                u.RoleNames.Add(roleName);
             }
         }
 
@@ -482,13 +483,14 @@ namespace OrchardCore.Users.Services
             if (user is User u)
             {
                 var roleNames = await _roleService.GetRoleNamesAsync();
+                var roleName = roleNames?.FirstOrDefault(r => NormalizeKey(r) == normalizedRoleName);
 
                 if (!roleNames.Any(r => NormalizeKey(r) == normalizedRoleName))
                 {
                     throw new InvalidOperationException($"Role {normalizedRoleName} does not exist.");
                 }
 
-                u.RoleNames.Remove(normalizedRoleName);
+                u.RoleNames.Remove(roleName);
             }
         }
 
--- a/src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
+++ b/src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
@@ -462,9 +462,9 @@ namespace OrchardCore.Users.Services
             if (user is User u)
             {
                 var roleNames = await _roleService.GetRoleNamesAsync();
-                var roleName = roleNames?.FirstOrDefault(r => NormalizeKey(r) == normalizedRoleName);
 
-                if (!roleNames.Any(r => NormalizeKey(r) == normalizedRoleName))
+                var roleName = roleNames.FirstOrDefault(r => NormalizeKey(r) == normalizedRoleName);
+                if (String.IsNullOrEmpty(roleName))
                 {
                     throw new InvalidOperationException($"Role {normalizedRoleName} does not exist.");
                 }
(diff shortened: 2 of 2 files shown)


src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
```diff
@@ -462,13 +462,14 @@ public async Task AddToRoleAsync(IUser user, string normalizedRoleName, Cancella
             if (user is User u)
             {
                 var roleNames = await _roleService.GetRoleNamesAsync();
+                var roleName = roleNames?.FirstOrDefault(r => NormalizeKey(r) == normalizedRoleName);
 
                 if (!roleNames.Any(r => NormalizeKey(r) == normalizedRoleName))
```
MikeAlhayek, line 467: if (String.IsNullOrEmpty(roleName))

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
```diff
@@ -462,13 +462,14 @@ public async Task AddToRoleAsync(IUser user, string normalizedRoleName, Cancella
             if (user is User u)
             {
                 var roleNames = await _roleService.GetRoleNamesAsync();
+                var roleName = roleNames?.FirstOrDefault(r => NormalizeKey(r) == normalizedRoleName);
```
MikeAlhayek, line 465:
```suggestion
                var roleName = roleNames.FirstOrDefault(r => NormalizeKey(r) == normalizedRoleName);
```

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
```diff
@@ -482,13 +483,14 @@ public async Task RemoveFromRoleAsync(IUser user, string normalizedRoleName, Can
...
```
MikeAlhayek, line 486:
```suggestion
                var roleName = roleNames.FirstOrDefault(r => NormalizeKey(r) == normalizedRoleName);
```

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
```diff
@@ -482,13 +483,14 @@ public async Task RemoveFromRoleAsync(IUser user, string normalizedRoleName, Can
...
 
                 if (!roleNames.Any(r => NormalizeKey(r) == normalizedRoleName))
```
MikeAlhayek, line 488:
```suggestion
                if (String.IsNullOrEmpty(roleName))
```

src/OrchardCore.Modules/OrchardCore.Users/Services/SetupEventHandler.cs
```diff
@@ -29,10 +29,11 @@ public SetupEventHandler(IUserService userService)
                 UserName = properties.TryGetValue(SetupConstants.AdminUsername, out var adminUserName) ? adminUserName?.ToString() : String.Empty,
                 UserId = properties.TryGetValue(SetupConstants.AdminUserId, out var adminUserId) ? adminUserId?.ToString() : String.Empty,
                 Email = properties.TryGetValue(SetupConstants.AdminEmail, out var adminEmail) ? adminEmail?.ToString() : String.Empty,
-                RoleNames = new string[] { "Administrator" },
                 EmailConfirmed = true
             };
 
+            user.RoleNames.Add("Administrator");
```
MikeAlhayek, line 35: I think it is safer to use `RoleNames = new string[] { "Administrator" },` but up to you.

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
jtkech: Here we check `RoleNames` not `RoleName`

For now I just copy paste the code as it was before.

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
jtkech: Will check, for now here it is as it was before

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
jtkech: idem

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
jtkech: idem

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
MikeAlhayek: @jtkech  yes I understand. you don't need to check names all over again since you are doing the same check in the previous line.

The code should look like this if you want to drop `?` since the `GetRoleNameAsync()` would return collection not nullable.

```
        public async Task AddToRoleAsync(IUser user, string normalizedRoleName, CancellationToken cancellationToken)
        {
            if (user == null)
            {
            }

            if (user is User u)
            {
                var roleNames = await _roleService.GetRoleNamesAsync();

                var roleName = roleNames.FirstOrDefault(name => NormalizeKey(name) == normalizedRoleName);

                if (String.IsNullOrEmpty(roleName))
                {
                    throw new InvalidOperationException($"Role {normalizedRoleName} does not exist.");
                }

                u.RoleNames.Add(roleName);
            }
        }
```

src/OrchardCore.Modules/OrchardCore.Users/Services/SetupEventHandler.cs
jtkech: Here the little fix is to not set the `IList RoleNames` to a string array, otherwise it saves in the json document the following.

    "RoleNames":{"$type":"System.String[], System.Private.CoreLib","$values":["Administrator"]}

In place of

    "RoleNames":["Administrator"]

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
jtkech: But yes better, done

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
jtkech: Done locally

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
jtkech: done

src/OrchardCore/OrchardCore.Users.Core/Services/UserStore.cs
jtkech: done

src/OrchardCore.Modules/OrchardCore.Users/Services/SetupEventHandler.cs
jtkech: Then we add a role to an empty list because the user is created just before.

MikeAlhayek (approved):
@jtkech I think this should be merged ASAP. I am approving this change.

jtkech:
Okay, just waiting for unit tests

jtkech:
Please wait, I'm testing it

PR #13111, opened 2023-01-18
Add missing settings in tenant ApiController

### LINKED ISSUE #13108: Missing Tenant description setting in the Tenant's ApiController
### Describe the bug

Currently, the tenant's description setting is missing in the ApiController [here](https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore.Modules/OrchardCore.Tenants/Controllers/ApiController.cs#L109).

### Expected behavior
We should be able to set a tenant's description via an API call, so if this property is included in the sent model, the tenant should get this description.

hishamco:
Fixes #13108

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Tenants/ViewModels/CreateApiViewModel.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Tenants/ViewModels/CreateApiViewModel.cs
@@ -2,6 +2,6 @@ namespace OrchardCore.Tenants.ViewModels
 {
     public class CreateApiViewModel : TenantViewModel
     {
-
+        public string Category { get; set; }
     }
 }
--- a/src/OrchardCore.Modules/OrchardCore.Tenants/ViewModels/CreateApiViewModel.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Tenants/ViewModels/CreateApiViewModel.cs
@@ -2,6 +2,5 @@ namespace OrchardCore.Tenants.ViewModels
 {
     public class CreateApiViewModel : TenantViewModel
     {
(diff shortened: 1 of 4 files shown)


hishamco:
@MZole seems there are few other than description ;)

hishamco:
What about recipes & feature profiles?

MZole:
> @MZole seems there are few other than description ;)

Yes, good catch, I only noticed the description.

> What about recipes & feature profiles?

I have tested them and they work as intended, they get set from the sent model.

hishamco:
> I have tested them and they work as intended, they get set from the sent model.

Really!! I didn't see any `IRecipeHarvester` injected there

MZole:
Yes, I have tested it in our [project](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions), we have a console application tester for the `ApiClient` which creates and sets up a tenant, and I also tested your solution for the `isNewTenant` with this   :smiley:.

hishamco:
Seems I need to try this one ;)

src/OrchardCore.Modules/OrchardCore.Tenants/ViewModels/CreateApiViewModel.cs
```diff
@@ -2,6 +2,6 @@ namespace OrchardCore.Tenants.ViewModels
 {
     public class CreateApiViewModel : TenantViewModel
     {
-
+        public string Category { get; set; }
```
jtkech, line 5: Maybe better to put it in the base `TenantViewModel` if needed in both derived `CreateApiViewModel ` and `EditTenantViewModel`.

PR #13112, opened 2023-01-18
Refactor tenant view models

hishamco:
Seems `EditTenantViewlModel` was exist when there's `CreateTenantViewModel`, I don't think there's a value from being there


jtkech:
@hishamco 

Okay, so In #10586 you added `Category` in `EditTenantViewModel` while it was not derived yet from `TenantViewModel`. Then in #10934 you created `TenantViewModel` and made `EditTenantViewModel` and `CreateApiViewModel` inheriting from it, notice that you kept `Category` but not `Description` in `EditTenantViewModel`. And here you are only keeping `TenantViewModel` containing all properties.

So before we had different view models, then okay good to inherit from a common one to not have to redefine shared properties, but here with only one model we can't tweak anymore the properties for different contexts, e.g. `CanEditDatabasePresets` is only used for editing.

So maybe better to keep it as is and only tweak the models if there is a current issue.

hishamco:
@MikeAlhayek we didn't do that usually, coz some PRs may take time, so no rush, so I will close this PR based on @jtkech feedback

Issue #12762, opened 2022-11-05
Bugs in displaying the localized calendar!


asemandeveloper:
In version 1.5.0, when we define the date field and change its type to Localized. In the data entry form, the calendar is not displayed correctly.
The calendar is also not localized. For example, the Arabic calendar does not work.

![image](https://user-images.githubusercontent.com/13555275/200100704-3a950a45-0d7b-4b8d-b769-c72970f669c5.png)

![image](https://user-images.githubusercontent.com/13555275/200100562-edad0751-7c66-4b3d-a95d-e0d583be6971.png)

hishamco:
@Skrypt is there something broken in the bootstrap?

asemandeveloper:
> @Skrypt is there something broken in the bootstrap?

Thank you
Maybe.
Also, when I have selected the Arabic culture, it does not show the Hijri (Arabic) calendar.

hishamco:
Hijri calndar is another thing, I don't think we support this in terms of UI, but it would be a big plus for Arabian ;)

micrub-it:
I have the same problem with IT localzation

stoyandimov:
@hishamco, I'm don't think localized calendar renders with bootstrap. Looking at the resource manager, it's probably still `jQuery-ui`.

[In this PR](https://github.com/OrchardCMS/OrchardCore/pull/13115), I ensure the `jQuery-ui` resources are included when the editor is displayed . I didn't want to switch to `bootstrap-datepicker` without discussing it first, but I hope this PR will be good enough fix for v1.6.

hishamco:
> @hishamco, I'm don't think localized calendar renders with bootstrap

Sure this will not be rendered properly

asemandeveloper:
Hello to all
As with Orchard CMS, it is better to display the calendar based on the selected culture. For example, if the Arabic culture is selected, the calendar will also show Hijri.
In your work, only the title of the months changes and not the dates and days of the calendar.

hishamco:
https://github.com/OrchardCMS/OrchardCore/pull/13115#issuecomment-1449687574

PR #13115, opened 2023-01-19
Fix localized datepicker for DateField

stoyandimov:
* Add the jQuery UI when date filed editor is displayed (+ i18n)
* Add jQuery UI image for the default template to ensure month arrows display correctly.

Fixes: https://github.com/OrchardCMS/OrchardCore/issues/12762


hishamco:
Please screenshot

stoyandimov:
Screenshot with Bulgarian localization
 image

hishamco:
Do we have `jQuery-ui-i18n` in our resource before?!! If Yes I'm not sure why we didn't use this from awhile ;)

hishamco:
Could you please try `ar` culture and share a screenshot

stoyandimov:
Not perfect:
image

hishamco:
I dislike the months names ;)

LGTM

hishamco:
BTW do we need a variations for those attached colorized images

stoyandimov:
I initially tried with only one image. However, the hover effect loads icons from a different image. Then I wasn't sure which ones are required so I added them all (as per the docs of jquery-ui).

stoyandimov:
> Do we have `jQuery-ui-i18n` in our resource before?!! If Yes I'm not sure why we didn't use this from awhile ;)

I checked that. I believe ([here](https://github.com/OrchardCMS/OrchardCore/commit/050d766f0d980c20229095b59333e214b1464ab9#diff-0dbbec37c3f8b3e751ec99d345c69464351a552a508221b3bd47ea275e297fa2)) it was moved from the layout.cshtml to the editor of the DateFiled, but it somehow never made it in the localized editor.

Skrypt:
I think the appropriate fix would be to use a different date picker that doesn't use JQuery UI. But for now, I guess that it can work. Though, we should really try to move away from JQueryUI.

stoyandimov:
This PR is a bug fix for an issue scheduled for 1.6. Looking at the code and commit history, the bug is that `jQuery-ui` is being used but the library (though present in the Resource Manager) is not included.

@Skrypt please, feel free to close this PR if you really find this fix inappropriate or if it is decided that the calendar fix should be replacing jQuery-ui all together.

Skrypt:
Nah, I guess we should merge it for now since it fixes the issue with the current implementation. Thanks.

hishamco:
LGTM after remove the extra icons PNGs, I'm sure we aren't using other than black icons, right?

hishamco:
@stoyandimov could you please react to the latest comment, then we could merge

hishamco:
@stoyandimov if you are busy let us know, so I can start a new PR or update yours if it's possible. We need to merge this before realease `1.6`

Skrypt:
@hishamco Use Gitextensions and pull the Pull Request fork locally. Then make the changes.

hishamco:
I will try, coz I remembered that I did it in OC Extractor, but for unknown reason I become the only commiter

sebastienros:
Note for later: the ui-icons_* files are defined in the jquery-ui.css file. These files need to be in a ./images subfolder of the css.

hishamco:
I just checkout the PR into a local branch, seems Seb merge it quickly :)

I will add a new PR

asemandeveloper:
Hello to all
As with Orchard CMS, it is better to display the calendar based on the selected culture. For example, if the Arabic culture is selected, the calendar will also show Hijri.
In your work, only the title of the months changes and not the dates and days of the calendar.

hishamco:
> As with Orchard CMS, it is better to display the calendar based on the selected culture. For example, if the Arabic culture is selected, the calendar will also show Hijri.

This is valid for `ar-SA` coz not all the Arabian countries are using Hijri calendar

asemandeveloper:
> 

But by choosing the Persian culture (fa), the calendar does not become Jalali calendar!

hishamco:
There's another service that you choose use to manage the calendar(s) `ICalendarSelector` but I don't think it's align with the jQuery datapicker

PR #13120, opened 2023-01-19
Make Notifications module publicly available 

### LINKED ISSUE #13119: Unable to find `OrchardCore.Notifications` feature when preview packages are installed in a provate project
### Describe the bug
I have a custom solution where I am using OC preview. If I search for `Notifications` module, it isn't listed as a feature unless I explictly install `OrchardCore.Notifications` directly in my host project.

I think the issue is the `OrchardCore.Notifications` project is included in `OrchardCore.Application.Cms.Core.Targets` project without `PrivateAssets="none"`

MikeAlhayek:
I think this will fix #13119 . But, if it does not, this will organize the `.csproj` as per the comment block.

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Notifications/Drivers/ListNotificationOptionsDisplayDriver.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Notifications/Drivers/ListNotificationOptionsDisplayDriver.cs
@@ -55,8 +55,8 @@ public class ListNotificationOptionsDisplayDriver : DisplayDriver<ListNotificati
         m.Statuses = model.Statuses;
         m.BulkActions = model.BulkActions;
         m.StartIndex = model.StartIndex;
-        m.EndIndex = m.EndIndex;
-        m.TotalItemCount = m.TotalItemCount;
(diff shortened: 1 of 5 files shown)


jtkech:
@MikeAlhayek 

Yes `PrivateAssets="none"` should be use when modules/themes are not referenced directly by the main application but transitively through another package, otherwise some msbuild targets scripts are not run on building.

Hmm, did you fixed other issues here?

MikeAlhayek:
@jtkech yes indeed it did. It's good to see the notification module in action outside the main project specially when multiple methods like Email and mobile/push notifications is used :)

jtkech:
Yes, I meant we also need to use `PrivateAssets="none"` in our solution because the web app references modules indirectly through the `OrchardCore.Application.Cms.Core.Targets` project.

If `PrivateAssets="none"` is not used, it still works if we are working with the source code, meaning referencing projects (not packages), msbuild scripts of the module are still executed on building.

But if we are using OC by referencing packages (not projects), If `PrivateAssets="none"` is not used, module scripts are not executed and the module is not harvested as a module.

Issue #13126, opened 2023-01-21
Unable to print property value using Liquid.


MikeAlhayek:
### Describe the bug
Unable to print property value using Liquid when using a for loop like this
```
<h4>Here is the responses</h4>
<ul>
{% for item in Workflow.Input["ThresholdSummaries"] %}
<li>The response to '{{ item.Title }}' was '{{ item.Respond }}' while threshold is '{{ item.Threshold }}'</li>
{% endfor %}
</ul>
```

### To Reproduce

I created the following workflow event

```
public class InterviewBelowThresholdEvent : EventActivity
{
    private readonly IStringLocalizer S;

    public override string Name => nameof(InterviewBelowThresholdEvent);

    public override LocalizedString DisplayText => S["Interview Below Threshold"];

    public override LocalizedString Category => S["Interview"];

    public InterviewBelowThresholdEvent(IStringLocalizer<InterviewBelowThresholdEvent> stringLocalizer)
    {
        S = stringLocalizer;
    }

    public override IEnumerable<Outcome> GetPossibleOutcomes(WorkflowExecutionContext workflowContext, ActivityContext activityContext)
    {
        return Outcomes(S["Done"]);
    }

    public override ActivityExecutionResult Resume(WorkflowExecutionContext workflowContext, ActivityContext activityContext)
    {
        return Outcomes("Done");
    }
}

public class ThresholdSummary
{
    public string Title { get; set; }

    public double Respond { get; set; }

    public double Threshold { get; set; }
}
```

From with in a handler I triggered the workflow like this

```
var workflowInput = new Dictionary<string, object>();
var summaries = new List<ThresholdSummary>()
{
    new ThresholdSummary()
    {
        Title = "Test1",
        Respond  = 2,
        Threshold = 3,
    },
    new ThresholdSummary()
    {
        Title = "Test2",
        Respond  = 1,
        Threshold = 3,
    }
};

workflowInput.Add("User", owner);
workflowInput.Add("ContentItem", interview);
workflowInput.Add("ThresholdSummaries", summaries);

await _workflowManager.TriggerEventAsync(nameof(InterviewBelowThresholdEvent), input: workflowInput, correlationId: interview.ContentItemId);
```

Then in a notify workflow, I put the following template

```
<h4>Here is the responses</h4>
<ul>
{% for item in Workflow.Input["ThresholdSummaries"] %}
[...]

PR #13129, opened 2023-01-21
Update workflow docs

MikeAlhayek:
Based on finding in issue #13126, update the workflow docs.


src/docs/reference/modules/Workflows/README.md
```diff
@@ -257,6 +257,29 @@ Developing custom activities involve the following steps:
...
+{
+    // Here we are passing custom data to the workflow's input.
+    { "data", customData}
+};
+
+await workflowManager.TriggerEventAsync("CustomTaskActivity", input);
+```
+
+You may passing an instance of a custom object to the workflow's input by adding it to the input collection. If you are looking to use liquid to access the member of the custom object, you must register a member access strategy. The following example for definitng a custom type
+
```
jtkech, line 275: Typo `definitng` => `defining` ;) and maybe a dot or a colon at the end.

MikeAlhayek: I'll correct it

MikeAlhayek: @jtkech fixed in #13198

PR #13134, opened 2023-01-22
Use IgnoreSystemCulture in DefaultCultureAliasProvider

hishamco:
Related to #13131

### DIFF
--- a/src/OrchardCore/OrchardCore.Localization.Abstractions/OrchardCoreRequestLocalizationOptions.cs
+++ b/src/OrchardCore/OrchardCore.Localization.Abstractions/OrchardCoreRequestLocalizationOptions.cs
@@ -10,8 +10,6 @@ namespace OrchardCore.Localization;
 /// </summary>
 public class OrchardCoreRequestLocalizationOptions : RequestLocalizationOptions
 {
-    private readonly bool _useUserOverride;
-
     /// <summary>
     /// Creates a new <see cref="OrchardCoreRequestLocalizationOptions"/> with default values.
     /// </summary>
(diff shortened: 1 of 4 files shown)


jtkech:
@hishamco 

In the module startup Configure() we create a local instance of `OrchardCoreRequestLocalizationOptions` (that uses `CultureOptions`) and that is passed to the middleware, so you can't inject it elsewhere, you would need to inject `CultureOptions` instead to retrieve `IgnoreSystemSettings`.

But as it is done currently, as I remember `DefaultCultureAliasProvider` is a host/app level service, so you cant inject options that are registered at the tenant level. Anyway looks like the alias middleware does nothing, see https://github.com/OrchardCMS/OrchardCore/pull/13131#issuecomment-1399628689, but maybe I misunderstood something ;)

hishamco:
First of all we have 4 localization PRs some are related if not all ;) let us revise them, merge what it should and reject what are not necessary - if there -

> In the module startup Configure() we create a local instance of OrchardCoreRequestLocalizationOptions (that uses CultureOptions) and that is passed to the middleware, so you can't inject it elsewhere, you would need to inject CultureOptions instead to retrieve IgnoreSystemSettings.

You 're right, nothing but you can't inject `CultureOptions` there ;)

hishamco:
Seems if move the registration of `DefaultCultureAliasProvider` to tenant level, we can inject `CultureOptions` then everything should work as expected

jtkech:
> First of all we have 4 localization PRs some are related if not all ;) let us revise them, merge what it should and reject what are not necessary

No problem, personnally I only open one PR ;)

> Seems if move the registration of DefaultCultureAliasProvider to tenant level, we can inject CultureOptions

Yes

> then everything should work as expected

Unless if the mapping is still never done, see https://github.com/OrchardCMS/OrchardCore/issues/11672#issuecomment-1401146843

hishamco:
@jtkech please review this once you make sure that the `CultureAlias` is works as expected according to my comment on the issue

jtkech:

hishamco:
Closed in favor of #13153

PR #13143, opened 2023-01-25
Fix CultureAliasConversionMiddleware

hishamco:
After serveral tests I realize there's a major issue on `CultureAliasConversionMiddleware` which is happen with unknown reason, until today - after @jtkech comment - I realize there's an extra culture exist on the document but not show up in the admin cultures

The current solution SHOULD fix the issue permenantly


jtkech:

Issue #13144, opened 2023-01-25
GraphQL errors when content part only contains custom fields


kevinsmgov:
### Describe the bug

When a ContentPart only contains custom fields, not one or more of the standard field types in Orchard, an error is raised when attempting too use GraphQL. The error reported is similar to this one (where _CustomPart_ represents a ContentPart with only custom fields)

```
[...]

PR #13145, opened 2023-01-25
fix error condition for custom fields

kevinsmgov:
#13144


kevinsmgov:
Also see discussion #13136

kevinsmgov:
This fix excludes ContentParts that contain only custom fields GraphQL doesn't know how to use.

PR #13147, opened 2023-01-26
Move YesSql Initialize outside the service configuration

### LINKED ISSUE #13061: Thread pool Starvation caused by sync over async pattern
We are currently working with Microsoft to resolve issues we're seeing with our production environment using OrchardCore.  They've identified a problem with thread starvation caused by sync over async code within the OrchardCore.Data.YesSql.OrchardCoreBuilderExtensions.AddData method associated with this line.

image

The problem I can see is there doesn't appear to be a way to fix this as there isn't a way to make the call in the service provider an async one instead but I'm wondering if anyone else has an idea?

MikeAlhayek:
Fix #13061

@jtkech you may want to check out the changes in `ShellScope` as it may not be a good idea. 

I'll work on this tomorrow. But feel free to mess around with the PR if you like.

### DIFF
--- a/src/OrchardCore/OrchardCore.Abstractions/Shell/IShellContextEvents.cs
+++ b/src/OrchardCore/OrchardCore.Abstractions/Shell/IShellContextEvents.cs
@@ -1,0 +1,9 @@
+using System.Threading.Tasks;
+using OrchardCore.Environment.Shell.Builders;
+
+namespace OrchardCore.Abstractions.Shell;
+
+public interface IShellContextEvents
+{
+    Task CreatedAsync(ShellContext context);
+}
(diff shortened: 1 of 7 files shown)


jtkech:
Yes maybe something like this but the goal of `_serviceScopeOnly` is to build a light scope e.g. without any tenant event, for example `ActivatingAsync()` triggering data migrations.

But `Store` is a tenant singleton so I'm more thinking about a tenant level registration. At the scope level we already `RegisterBeforeDispose()` a delegate to commit the session, so maybe something like this at the tenant level to ensure the `Store` initialization. E.g. in `CreateDescribedContextAsync()` we already call `settings.EnsureConfigurationAsync()` to init lazily shell config in an async way.

Will see tomorrow ;)

sebastienros (approved):

jtkech:
Yes, this is the right fix as I did locally but ATM with a singleton resolving `IStore` and `IOptions<StoreCollectionOptions>` itself.

Also I had an use case where I needed to check if the resolved `IStore` is null. In my case I inject it so it's like a `.GetRequiredService()` but even if the resolution doesn't fail, we have cases where we explicitly return null.

Wait before merging so that I can try it and do a last review.

sebastienros:
@jtkech since I approved, please merge it when you are ok with it.

jtkech:
Okay

PR #13148, opened 2023-01-26
fix error condition for custom fields

kevinsmgov:
#13144 
#13136 
When a ContentPart only contains custom fields, that GraphQL doesn't know how to use, the ContentPart is excluded.

### DIFF
--- a/src/OrchardCore/OrchardCore.ContentManagement.GraphQL/Queries/Types/DynamicContentTypeBuilder.cs
+++ b/src/OrchardCore/OrchardCore.ContentManagement.GraphQL/Queries/Types/DynamicContentTypeBuilder.cs
@@ -53,7 +53,7 @@ namespace OrchardCore.ContentManagement.GraphQL.Queries.Types
                     continue;
                 }
 
-                if (!(part.PartDefinition.Fields.Any(field=> contentFieldProviders.Any(fieldProvider=>fieldProvider.GetField(field)!=null))))
+                if (!(part.PartDefinition.Fields.Any(field => contentFieldProviders.Any(fieldProvider => fieldProvider.GetField(field)!=null))))
                 {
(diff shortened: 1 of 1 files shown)


src/OrchardCore/OrchardCore.ContentManagement.GraphQL/Queries/Types/DynamicContentTypeBuilder.cs
```diff
@@ -53,6 +53,11 @@ public void Build(FieldType contentQuery, ContentTypeDefinition contentTypeDefin
                     continue;
                 }
 
+                if (!(part.PartDefinition.Fields.Any(field=> contentFieldProviders.Any(fieldProvider=>fieldProvider.GetField(field)!=null))))
```
MikeAlhayek, line 56: This line will be more readable if you use a regular foreach loop. Also, fix the spacing

kevinsmgov: It's not really a loop. It's a boolean expression testing that no standard field types are included in the current ContentPart.

MikeAlhayek: well, Any loop over the records.

kevinsmgov: That would be duplicating the logic already in the fieldProvider and, IMO, it would make the process a lot more complicated that it is right now.

That being said, if the Orchard Core maintainers think it would improve the project, their welcome to create all that extra logic. This commit fixes the issue so I'm good with it as is.

PR #13151, opened 2023-01-26
Refactor social AUTH modules

hishamco:
After @sebastienros comments on the last standup meeting, there's no need to have interface for social AUTH settings services, this PR make everything consistent plus abstract almost common implementation details


Piedone:
I won't be able to review this for quite a while, so if anybody else can, please do.

src/OrchardCore.Modules/OrchardCore.Facebook/Services/IFacebookService.cs
```diff
@@ -1,16 +0,0 @@
-using System.Collections.Generic;
-using System.ComponentModel.DataAnnotations;
-using System.Threading.Tasks;
-using OrchardCore.Facebook.Settings;
-
-namespace OrchardCore.Facebook.Services
-{
-    public interface IFacebookService
```
Piedone, line 8: I'm OK with the `OAuthSettingsService<TAuthenticationSettings>` base class, but why is it good to remove the service interfaces? We lose the ability to override them (for extensibility and testing).

src/OrchardCore.Modules/OrchardCore.GitHub/Settings/GithubAuthenticationSettings.cs
```diff
@@ -1,12 +1,16 @@
 using Microsoft.AspNetCore.Http;
+using OrchardCore.Security;
 
 namespace OrchardCore.GitHub.Settings
 {
-    public class GitHubAuthenticationSettings
+    public class GitHubAuthenticationSettings : OAuthSettings
     {
         public string ClientID { get; set; }
+
```
Piedone, line 9: I don't think these new line breaks are needed. Same in `GoogleAuthenticationSettings`, `AzureADSettings`, `MicrosoftAccountSettings`, `MicrosoftAccountSettings`.

src/OrchardCore/OrchardCore.Infrastructure.Abstractions/Security/Services/OAuthSettingsService.cs
```diff
@@ -0,0 +1,49 @@
...
+    public async Task UpdateSettingsAsync(TAuthenticationSettings settings)
+    {
+        if (settings == null)
+        {
+            throw new ArgumentNullException(nameof(settings));
+        }
+
+        var oldSettings = await _siteService.LoadSiteSettingsAsync();
+
+        oldSettings.Properties[nameof(TAuthenticationSettings)] = JObject.FromObject(settings);
```
Piedone, line 43:
```suggestion
        oldSettings.Put(nameof(TAuthenticationSettings), settings);
```

Piedone (changes requested):

src/OrchardCore.Modules/OrchardCore.GitHub/Settings/GithubAuthenticationSettings.cs
hishamco: It's good for clarity & readability

src/OrchardCore.Modules/OrchardCore.Facebook/Services/IFacebookService.cs
hishamco: > We lose the ability to override them (for extensibility and testing).

Totally agree, I just react toSeb feedback on the meeting, but I think it's fine to leave it as it is

Piedone: @sebastienros please elaborate, then.

src/OrchardCore.Modules/OrchardCore.GitHub/Settings/GithubAuthenticationSettings.cs
Piedone: There's a reason we don't do this in every single place though :).

src/OrchardCore.Modules/OrchardCore.Facebook/Services/IFacebookService.cs
hishamco: May be because each service has default implementation, but still I'm agree with you :)

hishamco: May be because each service has default implementation, but still I'm agree with you :)

src/OrchardCore.Modules/OrchardCore.GitHub/Settings/GithubAuthenticationSettings.cs
hishamco: That's why I surrprised in many places in the code base ;)

Piedone:
Any news here, @hishamco?

hishamco:
@Piedone I need to revise all the pending PRs, seems it's the time make my PRs ready, thanks for reminder

src/OrchardCore.Modules/OrchardCore.Facebook/Services/IFacebookService.cs
Piedone: I still don't think we should remove the service interfaces. It's a step back.

src/OrchardCore.Modules/OrchardCore.GitHub/Settings/GithubAuthenticationSettings.cs
Piedone: I still think these line breaks are not needed. It doesn't add any readability any more than adding line breaks between any two lines of code.

Piedone (changes requested):

hishamco:
Adding lines between properties and methods is a good practice I have seen. Even ASP.NET Core team use it in the code base

Piedone:
Please don't move threaded convos out because it makes things harder to follow.

hishamco:
Believe it or not, I do it because I can't reply to your comment above :(

src/OrchardCore.Modules/OrchardCore.GitHub/Settings/GithubAuthenticationSettings.cs
hishamco: Adding lines between properties and methods is a good practice I have seen. Even ASP.NET Core team uses it in the code base

src/OrchardCore.Modules/OrchardCore.Facebook/Services/IFacebookService.cs
hishamco: I'm not sure why we need if we have an abstract class?!!

hishamco:
GitHub surpised me, seems we you comment on the code it duplicate it

src/OrchardCore.Modules/OrchardCore.GitHub/Settings/GithubAuthenticationSettings.cs
Piedone: If the properties have anything else on them, like attributes, or documentation, or they are multi-line, then sure. But otherwise, why? Why don't we add line breaks between private fields of a class too?

hishamco:
1st

Screenshot 2024-03-14 021426

2nd

Screenshot 2024-03-14 021441

src/OrchardCore.Modules/OrchardCore.Facebook/Services/IFacebookService.cs
Piedone: The same reason as we do it for services generally: loose coupling. However, these classes aren't even abstract. All are concrete types that are injected, removing the option to override them.

src/OrchardCore.Modules/OrchardCore.GitHub/Settings/GithubAuthenticationSettings.cs
hishamco: That's the code lines I know from years :)

src/OrchardCore.Modules/OrchardCore.Facebook/Services/IFacebookService.cs
hishamco: I didn't get you, I'm not sure why we should keep interfaces with the same methods while we can introduce a new abstract class to do the same thing

Piedone:
We talked about this under the ACS PR. This is how GH's PR comments work for some reason. It's not good, and I've actually written their support about it (weeks ago, no reply yet) but we should still scroll up and reply to the threads.

src/OrchardCore.Modules/OrchardCore.Facebook/Services/IFacebookService.cs
hishamco: I will bring the interfaces again

hishamco: Zoltan while we have many interfaces, I think we need only one for all services that could be implemented in the abstract class, right?

Piedone: Not really. If I want to override `FacebookService`, I need a specific interface for that, i.e. `IFacebookService` as it is in `main`. A common interface won't work for this.

hishamco: Could you please let me know how all the interfaces will interact with the newly added abstract class?

Piedone: They don't need to, just e.g. `FacebookService` needs to implement `IFacebookService`, as it was. All these interfaces can remain as they were, and since the abstract class implements them, in e.g. `FacebookService` you don't need to add anything else.

hishamco: That means nothing to be removed except a few types have been added

Piedone: Well, up to you how to go forward. The interfaces should stay because they have a use, they can't just be removed, as it will break people's code.

hishamco: Return all interfaces will make things harder because the abstract class should implement all the auth interfaces which do not make sense, all are equivalent except the settings type. If we introduce a generic type we will find ourselves using a single interface instead which is why I moved the implementation into an abstract class

Piedone: I can only repeat myself. Refactoring that breaks all consumers of these authentication services is not suitable. We have more than 400 interfaces in OC for such services, for a good reason. I really don't want to debate further why we should keep them. By looking at the code it seems to me that the only thing you need to do is re-add the interfaces and revert the dependency registrations, nothing else to change. It's something really simple.

Without the interfaces, I don't approve this PR. It's your call whether you want to rework the PR to keep them, or close it.

hishamco: As I said before my PR is based on @sebastienros suggestion nothing else. Moreover, I never mind reverting all the removed interfaces but the thing that I still do not get is how the interfaces will be used in the introduced abstract class

If you think this will take so long, please close the PR, I don't want to continue in something I wasn't convinced with

Thanks

Piedone: I suggest you just re-add the interfaces to see in action that nothing else with the base class needs to change.

hishamco: I just reverted the interfaces back as you said but again a generic interface make sense to me rather than have 4 - 6 interfaces with the same methods

src/OrchardCore.Modules/OrchardCore.GitHub/Recipes/GithubAuthenticationSettingsStep.cs
```diff
@@ -27,7 +27,7 @@ public async Task ExecuteAsync(RecipeExecutionContext context)
                 return;
             }
             var model = context.Step.ToObject<GitHubLoginSettingsStepModel>();
-            var settings = await _githubAuthenticationService.LoadSettingsAsync();
+            var settings = await _githubAuthenticationService.GetSettingsAsync();
```
Piedone, line 30: `GetSettingsAsync()` returns an immutable settings object that won't be possible to update in the next lines, so this should stay Load*. Same for all the *Step classes.

Piedone (changes requested):
Please test all logins, as well as imports, if you haven't done already.

sebastienros:
Closing as stale and not much value. There is also a new settings helper that is now used in these classes.

PR #13153, opened 2023-01-26
Fix Culture Alias

hishamco:


hishamco:
@jtkech could you please have a look if you have a time

src/OrchardCore.Modules/OrchardCore.Localization/Services/LocalizationService.cs
```diff
@@ -40,10 +46,31 @@ public async Task<string[]> GetSupportedCulturesAsync()
...
+
+            foreach (var supportedCulture in supportedCultures)
+            {
+                foreach (var provider in _cultureAliasProviders)
+                {
+                    if (provider.TryGetCulture(supportedCulture, out var culture))
+                    {
+                        if (!supportedCultures.Contains(culture.Name))
+                        {
+                            mappedSupportedCultures.Add(culture.Name);
```
jtkech, line 63: Why do you want to add for example `zh-CN` if `zh-Hans-CN` is there, if now `zh-CN` is available again and can be selected in the admin settings, joined to `CultureInfo.GetCultures()` in `LocalizationSettingsDisplayDriver`?

Yes we could have used the same logic to add dynamically these alias when populating the supported cultures of the aspnet options (as I did at some point), maybe only here that already built culture infos would be useful.

Now, we could use the same logic but in the settings driver to complete the available culture names showed in the admin settings, hmm but here only providing a list of culture aliases we want to support as you did looks sufficient.

hishamco:
Closed in favor of #13174

PR #13158, opened 2023-01-27
'IgnoreSystemSettings' culture option is not used.

jtkech:
This currently as our our custom `SetDefaultCulture()` and `AddSupported[UI]Cultures()` are not called.

Repro: Enable Localization and put breakpoints in the methods intended to be executed.


hishamco:
Already fixed in #13153

hishamco (approved):

hishamco:
Let us merge this, then we can review #13153

PR #13163, opened 2023-01-27
Replace CultureOptions with OrchardCoreRequestLocalizationOptions

hishamco:
There's few limitations in the current `OrchardCoreRequestLocalizationOptions` no one can access `ignoreSystemSettings` unless you are using `CultureOptions` but if we look closely we could add a new property to `OrchardCoreRequestLocalizationOptions` , then it could be accessible everywhere through option pattern, this way no need for `CultureOptions` anymore.


jtkech:
@hishamco 

Maybe better to keep them separate, we can need culture options without needing other OC options if Localization is not enabled. Also, OC options and aspnet options having the same properties, would be both configurable, even if at the end the other options will only come from the aspnet options.

Personally I would have not created the derived OC options but just extensions on the aspnet options to set cultures. We saw that "ignore settings" was not used, then that we forgot to report the other aspnet options (naturally done before), then here another little breaking change by changing the config section name.

---

Just for info, just saw that the aspnet `AddSupportedCultures()` uses an initial capacity to build the list.

            var supportedCultures = new List<CultureInfo>(cultures.Length);

Just saw that the setup doesn't use `IgnoreSystemSettings`, maybe intended.

hishamco:
> Maybe better to keep them separate, we can need culture options without needing other OC options if Localization is not enabled.

I don't think so, because `CultureOptions` is only has `IgnoreSystemSettings` which is available in `OrchardCoreRequestLocalizationOptions`, further more when .NET 8 is shipping we can easily use original `RequestLocalizationOptions` which has equivalent to `IgnoreSystemSettings`

I'm open for any suggestion but no need to have an option which has a single property that already somewhere else

jtkech:
Okay, so the aspnet Options will have a `CultureInfoUseUserOverride` which is not right for me because  it is not used by the middleware, but only if we use their helper methods to add cultures.

For me culture options are global for all tenants, request localization options are more feature based.

I still think that it would have been better to not create a new OC options, but keep things as before and just add extensions on aspnet Options to add cultures with an `useUserOverride` param (as aspnet could have done) whose value comes from the more global `CultureOptions`.

Then in a future version just re-use the aspnet Options methods, and before just set the new property with the value deduced from the global `CultureOptions`, can be done in a `services.Configure<>()` (or `IConfigureOptions<>`) in place of the setup `Configure()`, so that people can override it.

---

Otherwise it has been and will be more error prone.

At some point we were not using "ignore system settings", then not reporting some options, here we change a config section name, if in a future version we will bind to the new aspnet Options directly, we will also have to change the config field name to `CultureInfoUseUserOverride`. And with the opposite logic, and true by default so we will have to check if the config value exists.

hishamco:
> Okay, so the aspnet Options will have a CultureInfoUseUserOverride which is not right for me because it is not used by the middleware, but only if we use their helper methods to add cultures.

Which middleware do you mean? Sure we need to use normal `RequestLocalizationOptions`, that's why we introduced `OrchardCoreRequestLocalizationOptions`

> For me culture options are global for all tenants, request localization options are more feature based.

Don't forget that we introduce the `CultureOptions`, but it's something related to localization options

> I still think that it would have been better to not create a new OC options, but keep things as before and just add extensions on aspnet Options to add cultures with an useUserOverride param (as aspnet could have done) whose value comes from the more global CultureOptions.

It could be too, but it will introduce a breaking-changes

Anyhow, I'm open to any suggestion, your suggestions is welcome too ;)

jtkech:
@hishamco Okay no problem, look at #13174 to see what I mean.

hishamco:
If we agreed to make `CultureOptions` as global settings we can make it as it is, but I'm still see there's no need to introduce extension methods for `RequestLocalizationOptions`, when .NET 8 shipped we simplify remove the `OrchardCoreRequestLocalizationOptions` and set `CultureInfoUseUserOverride` from our `CultureOptions`. Moreover this will not conflict with the changes introduced in PR #13153

jtkech:
> but I'm still see there's no need to introduce extension methods for `RequestLocalizationOptions`

Because this is only what we need in place of having introduced `OCRequestLocalizationOptions`, and less prone of errors and breaking changes as we already had. In fact you used this new options as an helper, and here you planned to configure it and use it directly which would be another little breaking change.

> when .NET 8 shipped we simplify remove the `OrchardCoreRequestLocalizationOptions` and set `CultureInfoUseUserOverride`.

With #13174 we would just have to re-use the `RequestLocalizationOptions` methods.

hishamco:
I agree with you to don't remove `CultureOptions`, so I will close this PR then

But still I see a value with the PR I did before, and make things as its without introducing a new extension methods, then removing `OrchardCoreRequestLocalizationOptions` after .NET 8, this way `CultureOptions` is there, `IgnoreSystemCulture` is working as well as culture mapping working as well through the `LocalizationService`

PR #13167, opened 2023-01-28
Enable Build Acceleration in VS

drewnoakes:
Visual Studio 17.5 supports "Build Acceleration" which can dramatically improve incremental build times. This PR turns the feature on via the `AccelerateBuildsInVisualStudio` property, for all projects. It adds a `Directory.Build.props` file to the solution root to be picked up by all projects, and chains it into any existing `Directory.Build.props` files.

As an example of the performance improvement possible with this feature.

1. Build the entire solution, making everything up-to-date.
2. Make a non-API change to `IdGenerator` in `OrchardCore.Abstractions` (add a comment, for example).
3. Build the solution again (build, not rebuild).

Looking at the log output you'll see something resembling:

```
========== Build: 1 succeeded, 0 failed, 199 up-to-date, 3 skipped ==========
========== Build started at 9:43 PM and took 07.308 seconds ==========
Visual Studio accelerated 175 project(s), copying 525 file(s). See https://aka.ms/vs-build-acceleration.
```

(Note: You may see a larger number of builds due to [this issue](https://github.com/dotnet/project-system/blob/main/docs/up-to-date-check.md#discrepancies-between-futdc-logging-and-build-summary). Reviewing the earlier log output will show that only a single project actually built.)

Without this PR, the same steps produce:

```
========== Build: 176 succeeded, 0 failed, 24 up-to-date, 3 skipped ==========
========== Build started at 9:51 PM and took 01:03.484 minutes ==========
```

Instead of just building the one project that changed (`OrchardCore.Abstractions`), VS would previously build all 175 referencing projects as well, taking 63 seconds. With Build Acceleration enabled, that time drops to around 7 seconds.

Note that developers using earlier versions of Visual Studio will continue to see correct builds. The previous behaviour remains as is.

For more information on the feature, see https://aka.ms/vs-build-acceleration


OrchardCore.sln
```diff
@@ -216,6 +216,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "build", "build", "{184139CF
 	ProjectSection(SolutionItems) = preProject
 		src\OrchardCore.Build\Dependencies.AspNetCore.props = src\OrchardCore.Build\Dependencies.AspNetCore.props
 		src\OrchardCore.Build\Dependencies.props = src\OrchardCore.Build\Dependencies.props
+		Directory.Build.props = Directory.Build.props
```
hishamco, line 219: You could hosted in the `OrchareCore.Build` folder instead

drewnoakes: That's `src/OrchardCore.Build` but this file also applies to files in `test`, so that doesn't feel like the right level to me.

hishamco: `Dependencies.prop` for instance has been used in test project too

drewnoakes: The convention for Directory.Build.props is that it applies to subfolders. The probing stops at the first round file, hence the other changes here that chain the files to one another when multiple levels exist.

This way, all future projects added will pick this property up by default.

drewnoakes: https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022#directorybuildprops-and-directorybuildtargets

drewnoakes:
Build Acceleration is now GA with the release of Visual Studio 17.5.

Is there anything else you need from me to get this change merged?

Skrypt (approved):

Skrypt:
Thanks @drewnoakes 
Good getting contribs from @microsoft people!

PR #13169, opened 2023-01-29
Startup Async Initializers on Tenant Container created.

jtkech:
Useful to init any Tenant Singleton Service in an async way when a tenant container is built.

Without having to define and register another service as this is integrated by the system.

But by defining in any module startup an `InitializeServicesAsync(IServiceProvider serviceProvider)`.

Or by using the new `OCBuilder` helper `builder.InitializeServices(async serviceProvider => ...`.

In this PR we use the new `OCBuilder` extension helper to init the YesSql `IStore`.

    builder.ConfigureServices(services =>
    {
        services.AddScoped<IDbConnectionValidator, DbConnectionValidator>();
        ...
        // Configuring data access
        services.AddSingleton(sp =>
        {
            ...
            var store = StoreFactory.Create(storeConfiguration);
            ...
        });
        ...
    })
    .InitializeServices(async serviceProvider =>
    {
        var store = serviceProvider.GetService<IStore>();
        if (store == null)
        {
            return;
        }

        await store.InitializeAsync();

        ...
    });

Note: The usage of the `OCBuilder` helper allows to do this kind of thing from the web application.


src/OrchardCore/OrchardCore.Abstractions/Modules/Builder/OrchardCoreBuilder.cs
```diff
@@ -56,6 +57,26 @@ public OrchardCoreBuilder ConfigureServices(Action<IServiceCollection> configure
             return ConfigureServices((s, sp) => configure(s), order);
         }
 
+        /// <summary>
+        /// This async method gets called for each tenant. Use this method to initialize container services.
```
MikeAlhayek, line 61:
```suggestion
        /// Use this method to initialize tenant services asynchronously.
```

src/OrchardCore/OrchardCore.Abstractions/Modules/Builder/OrchardCoreBuilder.cs
```diff
@@ -56,6 +57,26 @@ public OrchardCoreBuilder ConfigureServices(Action<IServiceCollection> configure
...
+        /// </summary>
+        /// <param name="initializeAsync">The async task to execute when initializing the services for a tenant..</param>
```
MikeAlhayek, line 63:
```suggestion
        /// <param name="initializeAsync">The task to execute when initializing tenant services.</param>
```

src/OrchardCore/OrchardCore.Abstractions/Modules/Builder/OrchardCoreBuilder.cs
```diff
@@ -104,7 +125,7 @@ public OrchardCoreBuilder EnableFeature(string id)
                 {
                     var service = services[index];
                     if (service.ImplementationInstance is ShellFeature feature &&
-                        string.Equals(feature.Id, id, StringComparison.OrdinalIgnoreCase))
+                        String.Equals(feature.Id, id, StringComparison.OrdinalIgnoreCase))
```
MikeAlhayek, line 128: I don't think we need `StringComparison.OrdinalIgnoreCase` here since `EnableFeature` is called from code not user input.

src/OrchardCore/OrchardCore.Data.YesSql/OrchardCoreBuilderExtensions.cs
```diff
@@ -151,6 +149,22 @@ public static OrchardCoreBuilder AddDataAccess(this OrchardCoreBuilder builder)
                 services.AddScoped<IDocumentStore, DocumentStore>();
                 services.AddSingleton<IFileDocumentStore, FileDocumentStore>();
                 services.AddTransient<IDbConnectionAccessor, DbConnectionAccessor>();
+            })
+            .InitializeServices(async serviceProvider =>
```
sebastienros, line 153: The name doesn't make it easy to understand what it does and when it's called.

jtkech: @sebastienros 

I'm open to any name ;) It is executed when a shell container is built, so after all `ConfigureServices()` and of course before all `Configure()`used to build the tenant pipeline. It allows to init some services in an async way, here I use it to initialize the `IStore`, see my 1st comment https://github.com/OrchardCMS/OrchardCore/pull/13169#issue-1561147273

    .InitializeServices(async serviceProvider =>
    {
        var store = serviceProvider.GetService<IStore>();
        if (store == null)
        {
            return;
        }

        await store.InitializeAsync();

        ...
    });

src/OrchardCore/OrchardCore.Abstractions/Modules/Builder/OrchardCoreBuilder.cs
jtkech: Yes, I always hesitate to be consistent with what is already there to not have to do too much updates, for example here in the same file we have the following comments including this one.

        /// This method gets called for each tenant. Use this method to add services to the container.
        /// This async method gets called for each tenant. Use this method to initialize container services.
        /// This method gets called for each tenant. Use this method to configure the request's pipeline.

If I apply the suggestion we would have

        /// This method gets called for each tenant. Use this method to add services to the container.
        /// Use this method to initialize tenant services asynchronously.
        /// This method gets called for each tenant. Use this method to configure the request's pipeline.

That's why I hesiated between keeping the same kind of comments or updating all of them ;)

src/OrchardCore/OrchardCore.Abstractions/Modules/Builder/OrchardCoreBuilder.cs
jtkech: Idem, for example just before we already have this existing comment.

        /// <param name="configure">The action to execute when configuring the services for a tenant.</param>

I'm open to any suggestion but maybe a separate PR to review all of these comments.

At least there is an extra dot to remove.

src/OrchardCore/OrchardCore.Abstractions/Modules/Builder/OrchardCoreBuilder.cs
jtkech: Yes I was tempted to do it, just to not change the existing code here.

Hmm, anyway the provided id should be the exact feature id, otherwise it would not work.

Okay I will change it

jtkech: Done

PR #13174, opened 2023-01-29
LocalizationOptions cleanup

jtkech:
@hishamco 

- So here most of the changes was to revert back to how it was before.

- Then I just created `RequestLocalizationOptionsExtensions` to add cultures by using our new `IgnoreSystemSettings`, this without using a custom `OCRequestLocalizationOptions`.

    In a future version we will just need to set the new `CultureInfoUseUserOverride` of the aspnet options, and then re-use their methods.

- Then about cultures as `zh-CN`, the only thing we needed was to add them back to the supported cultures, by adding them dynamically or making them available again in the admin UI.


src/OrchardCore.Modules/OrchardCore.Localization/RequestLocalizationOptionsExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+            string defaultCulture)
+        {
+            options.DefaultRequestCulture = new RequestCulture(new CultureInfo(defaultCulture, useUserOverride));
+
+            return options;
+        }
+
+        public static RequestLocalizationOptions AddSupportedCultures(
+            this RequestLocalizationOptions options,
+            bool useUserOverride,
```
hishamco, line 23: I expect you will add `useUserOverride` as first param, but it's weird from API design perspective

No problem if it's hosted in the module itself

hishamco (commented):
Removing `OrchardCoreRequestLocalizationOptions` is a breaking-change, right?

hishamco:
BTW I need to know what's the wrong with #13153? I'm still see it's better to provide the cultures in `ILocalizationService` implementation for consistency, so it's available everywhere

src/OrchardCore.Modules/OrchardCore.Localization/RequestLocalizationOptionsExtensions.cs
jtkech: `useUserOverride` can't be the first param which is `this options` on which the extension is applied, and can't be the last param because the `string[] params` should be at the end.

hishamco: Yes I know, that's one reason, which change my mind at the beginning  - before introducing `OrchardCoreRequestLocalizationOptions` - when you suggested the extensions

jtkech:
I think `LocalizationService` is okay, anyway `GetDefaultCultureAsync()` should return the supported cultures that are set in the admin settings, and because we opted to add the newly missing cultures as `zh-CN` and `zh-TW` (maybe there are others) to the available cultures in the admin, they will be there.

Then yes to work, any of the above "missing" culture we want to still support, or any newly added culture, would need a related localization folder. Here I let you decide about the strategy, I'm not a culture expert, we can keep some folders, create new ones and/or let people creating their custom folders.

Not a fan but maybe this is here that some mapping could be done, when we harvest files for a given culture, so that Culture1 and Culture2 would use the same localization folder of Culture1, or of Culture2 depending on the strategy and the translations we want to provide from crowdin.

But maybe better to just support the cultures we show up in the admin or say to people to create custom folders. As I remember there are different locations/folders from where *.po files are harvested, and some locations are not cleared on each building and then can be used by people for custom translations.

---

Note: As I remember, at some point one concern was if the culture fails to be loaded, but even if we want to map it, it would have failed before when first selected by the aspnet request localize middleware.

src/OrchardCore.Modules/OrchardCore.Localization/RequestLocalizationOptionsExtensions.cs
jtkech: Ah okay, but that's okay if internal to the module and only temporary ;)

hishamco: > No problem if it's hosted in the module itself

That's what I intended here ;)

hishamco:
> Then yes to work, any of the above "missing" culture we want to still support, or any newly added culture, would need a related localization folder. Here I let you decide about the strategy, I'm not a culture expert, we can keep some folders, create new ones and/or let people creating their custom folders.

The folder names are still exist in Crowdin for backward compatibility, the aim for the PR #13185 is to include `zh-CN`, `zh-TW` to the supported list

> Not a fan but maybe this is here that some mapping could be done, when we harvest files for a given culture, so that Culture1 and Culture2 would use the same localization folder of Culture1, or of Culture2 depending on the strategy and the translations we want to provide from crowdin.

According what @sebastienros suggested to have `zh-Hans` and `zh-Hant` so the fallback will be there for `zh-Hans-CN` and `zh-Hant-TW` which is good and bad coz I'm thinking Hans and Hant cultures they will sure fallback into their parent we can't use a specific localization string for a particular culture

> Note: As I remember, at some point one concern was if the culture fails to be loaded, but even if we want to map it, it would have failed before when first selected by the aspnet request localize middleware.

Could you please ellaborate your intend here to understand you very well :)

jtkech:
> The folder names are still exist in Crowdin for backward compatibility

Yes I saw cool.

> the aim for the PR https://github.com/OrchardCMS/OrchardCore/pull/13185 is to include zh-CN, zh-TW to the supported list

No need if we add them to the available cultures in the driver, but another new PR to check, after a quick look I saw that you now add them through the `ILocalizationService` interface which looks wrong because if `OC.Localization` is not enabled we use `DefaultLocalizationService` which is intended to only return the installed culture.

Personally I would have made e.g. a `CultureInfo.GetCulturesWithAliases()` extension to be used in the right places, for example in the search controllers if we want the same behavior as when the aliases was returned from `CultureInfo.GetCultures()`, but not sure. Note: Also, maybe `_chineseAliasCultures` is too specific, we may need to add other aliases so maybe `_culturesAliases`.

Maybe you are doing too many PRs too soon, maybe better to first return to a stable version. For me about `IgnoreSystemSettings` we only needed new extensions to add cultures, and for the culture aliases maybe we only needed a new `CultureInfo` extension, maybe I will update this PR for this.

For example in your last PR where we return `await NullProviderCultureResult`, maybe I will update this PR to just `return default;`, see https://github.com/OrchardCMS/OrchardCore/pull/13181/files#r1093054022.

> According what @sebastienros suggested to have zh-Hans and zh-Hant so the fallback will be there for zh-Hans-CN and zh-Hant-TW

Okay cool

> which is good and bad coz I'm thinking Hans and Hant cultures they will sure fallback into their parent we can't use a specific localization string for a particular culture

Not sure I understand here, if at the end a culture is selected (fallback or not) and if the folder is there, normally that would be okay, anyway here I let you and Sebastien decide.

> Could you please ellaborate your intend here to understand you very well :)

What I meant is that at some point it was said that we need to map the culture because the alias one may not be correctly loaded, but as it is done in the main branch, in our new middleware we map the culture alias only if the aspnet middleware has already set the thread culture to this alias, so it would have already failed. And now that these culture aliases are available again (listed in the admin and with the related folder), no need to do any mapping which prevents from setting the thread culture twice.

hishamco:
> Maybe you are doing too many PRs too soon, maybe better to first return to a stable version

Still I'm not sure what's the wrong with culture alias provider, the PR #13153 fix it, while your PR trying to cleanup everything and do the changes in the localization module which is might wrong IMHO

> For example in your last PR where we return await NullProviderCultureResult, maybe I will update this PR to just return default;

How they are different?!!

hishamco:
BTW I see you did similar changes like what I did in #13185, instead you did it as extension method. If we accept that PR regardless we use extension method of interface static method, we could rid off the culture alias, right?

hishamco:
We could meet together if you don't mind, or let us try to consider #13185, then it's fine to rid off the culture alias stuff that we introduce

jtkech:
First for the new `GetAllCulturesWithAliases()` I was tempted to only return culture names (not full `CultureInfo`) because this is only when we populate `SupportedCultures` (that may be used by the aspnet middleware to set the thread culture) and as I remember with `CultureSCope` (where we set explicitly the thread culture) that we really need to built a full `CultureInfo` and by using `useUserOverride`.

The other places for now is just because in some views we not only want to display the culture `Name` but also the `DisplayName`, and the regular `GetCultures()` return cultures without using `useUserOverride`, so I think we just need to complete `GetCultures()` without having to map some cultures to other full `CultureInfo` using `useUserOverride`.

> Still I'm not sure what's the wrong with culture alias provider, the PR https://github.com/OrchardCMS/OrchardCore/pull/13153 fix it

Don't remember all details but I let you a review comment in this PR where in `LocalizationService` you added a `GetSupportedCulturesAsync()` to add aliases while `_localizationSettings.SupportedCultures` is intended to already contain them. About the alias provider see may above comment.

> How they are different?!!

I agree not so important but for me better to just return `default` than `await NullProviderCultureResult` which ends up to `await Task.FromResult(default(ProviderCultureResult))`, I think this helper is only useful when you don't have any await in your method to not write all the `Task.FromResult()` that you don't need if you have some await (when your method has the `async` modifer).

Need to leave, will see for the other comments ;)

> We could meet together

No problem if needed, just ping me, jtkech on skipe, but not free before around 23h pm UTC ;)

---

Just saw that you added the label `breaking change` to this PR while you already introduced some, not so important but you remember with the new OC options where `IgnoreSystemSettings` was not always used, then never used, the other options not reported, and that we fixed recently.

Also when you wanted to bind the new OC options and use it directly when registering the aspnet middleware, it would have been another little breaking change because `OC.ContentLocalization` (if enabled) still configures the aspnetcore options not our derived one.

    services.Configure<RequestLocalizationOptions>(options =>
        options.AddInitialRequestCultureProvider(new ContentRequestCultureProvider()));

So here I tried to use the simplest way to have a stable version, without introducing any new breaking change, but yes I may be wrong in some places, errare humanum est, "only" perseverare "is" diabolicum ;) Then after I'm not against any improvment, e.g resolving alias providers (maybe only names) to allow people to add some alias names, or on demand we could complete the current list, will see.

hishamco:
> Just saw that you added the label breaking change to this PR while you already introduced some, not so important but you

I just added for a little note because we breaking a public API nothing else

I'm suggesting to meet on Teams if it's possible ot chatting in Gitter and agree what we should do regarding options, aliases as also the Chinese cultures which bring all this stuff ;)

jtkech:
Okay let's start a private chat on skipe or gitter if you prefer, I'm always connected on skipe, let me know if I have to connect to gitter.

hishamco:
I will be on gitter with 20-30 mins

hishamco:
After a long discussion with @jtkech on Gitter we decided to integrate different things in different PRs

jtkech:
Close in favor of #13185

PR #13178, opened 2023-01-30
fix: #13177 - changed AswFileStore initialization

### LINKED ISSUE #13177: OrchardCore_Media_AmazonS3 - BasePath folder name issue with multi-tenant
### Describe the bug
When configuring the Feature `OrchardCore_Media_AmazonS3` I have configured the `BasePath` value as below:  
```
"BasePath": "{{ ShellSettings.Name }}/Media",
```
My expectation was that for each Tenant, there would be a subfolder created for each Tenant with a unique name.  What is happening though is that the folder in the S3 bucket is actually being generated as below:
image
The Liquid value is never being interpreted.  The files still get uploaded, but they all get dropped in that same incorrectly named (and shared) folder.

### Expected behavior
If I have 2 tenants named `Tenant01` and `Tenant02` I expect two subfolders be created off of the root of the site.  The Media for those tenant sites will be dropped in the relevant folder for the tenant.

### Possible solution
I made two changes locally to test a theory
- `OrchardCore\OrchardCore.FileSTorage.AmazonS3\AwsFileStorage.cs` : constructor is modified to use `IOptions<AwsStorageOptions> instead of just `AwsStorageOptions`
```
    public AwsFileStore(IClock clock, IOptions<AwsStorageOptions> options, IAmazonS3 amazonS3Client)
    {
        _options = options.Value;
```
- `OrchardCore.Modules\OrchardCore.Media.AmazonS3\Startup.cs : alter the initialization of the `AwsFileStore` as follows:  
```
                IOptions<AwsStorageOptions> options = serviceProvider.GetRequiredService<IOptions<AwsStorageOptions>>();
                var fileStore = new AwsFileStore(clock, options, amazonS3Client);
```

kylgarmor:
- fluid interpreter was not being used to parse options

Fixes #13177

### DIFF
--- a/src/OrchardCore.Modules/OrchardCore.Media.AmazonS3/Startup.cs
+++ b/src/OrchardCore.Modules/OrchardCore.Media.AmazonS3/Startup.cs
@@ -101,7 +101,7 @@ public class Startup : Modules.StartupBase
                 var logger = serviceProvider.GetRequiredService<ILogger<DefaultMediaFileStore>>();
                 var amazonS3Client = serviceProvider.GetService<IAmazonS3>();
                 
-                IOptions<AwsStorageOptions> options = serviceProvider.GetRequiredService<IOptions<AwsStorageOptions>>();
+                var options = serviceProvider.GetRequiredService<IOptions<AwsStorageOptions>>();
                 var fileStore = new AwsFileStore(clock, options.Value, amazonS3Client);
 
                 var mediaUrlBase = $"/{fileStore.Combine(shellSettings.RequestUrlPrefix, mediaOptions.AssetsRequestPath)}";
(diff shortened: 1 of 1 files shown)


src/OrchardCore.Modules/OrchardCore.Media.AmazonS3/Startup.cs
```diff
@@ -100,8 +100,10 @@ public override void ConfigureServices(IServiceCollection services)
                 var clock = serviceProvider.GetRequiredService<IClock>();
                 var logger = serviceProvider.GetRequiredService<ILogger<DefaultMediaFileStore>>();
                 var amazonS3Client = serviceProvider.GetService<IAmazonS3>();
+                
+                IOptions<AwsStorageOptions> options = serviceProvider.GetRequiredService<IOptions<AwsStorageOptions>>();
```
jtkech, line 104:
```suggestion
                var options = serviceProvider.GetRequiredService<IOptions<AwsStorageOptions>>();
```

jtkech (approved):

MikeAlhayek:
Can we move line 41 to 96?

jtkech:
Not really, as it is done line 41 `storeOptions` is just bound from the configuration to do some validations, whereas in the implementation factory we resolve `IOptions<AwsStorageOptions>` so that some Liquid templating is applied by `AwsStorageOptionsConfiguration`, which fixes #13177.

kylgarmor:
Just curious, does this get added automatically to the "preview" nuget package for 1.6? 

https://cloudsmith.io/~orchardcore/repos/preview/packages/detail/nuget/OrchardCore.Application.Cms.Targets/latest/

MikeAlhayek:
Yes. Immediately after the merge the CLI would have trigger to create and publish the nuget packages.

